PR #313, opened 2014-09-17
Dump option added --dump <on/off>:

romanman:
+ Dump a full state to a file in a json format after each block import


chfast:
I think this might be outdated...

PR #401, opened 2014-10-20
Exception safety of Web3 API - first shot (please read comment!)

CJentzsch:
At many places I am not sure if it is ok to do nothing, or if one should exit or the other way around. Please review carefully and read my comments.
I used noexcept way more than I should. Because boost::current_exception_diagnostic_information() theoretically might throw, they are not 100% noexcept. But it is extremly helpful to label those functions so one does not have to check them over and over again.
This is no final pull request, since some noexcept as well as comments designed for the review process will be removed. 
Although my goal was/is to catch handle exceptions as low level as possible, in order to strongly guarantee exceptions safety for the web3 API a lot of try/catch blocks are in the Client class. But hopefully almost no exceptions will make it that fare.


CJentzsch:
this so outdated that I will have to redo it completely, when it becomes a higher priority than testing for consensus.

PR #656, opened 2014-12-18
p2p: peer discovery integration

subtly:
Also updated/renamed member properties in Host to reflect future support for multiple listeners and to clarify what current listeners provide.


libp2p/NodeTable.h
```diff
@@ -0,0 +1,363 @@
...
+class NodeTable: UDPSocketEvents, public std::enable_shared_from_this<NodeTable>
+{
+	friend struct Neighbors;
+	using NodeSocket = UDPSocket<NodeTable, 1280>;
+	using TimePoint = std::chrono::steady_clock::time_point;
+	using EvictionTimeout = std::pair<std::pair<Address,TimePoint>,Address>;
+
+	struct NodeDefaultEndpoint
+	{
+		NodeDefaultEndpoint(bi::udp::endpoint _udp): udp(_udp) {}
```
subtly: This struct will hold additional information and the node may have other endpoints thanks to ipv6.

test/net.cpp
```diff
@@ -0,0 +1,207 @@
...
+{
+	UDPDatagram d(bi::udp::endpoint(boost::asio::ip::address::from_string("127.0.0.1"), 30300), bytes({65,65,65,65}));
+	TestUDPSocket a; a.m_socket->connect(); a.start();
+	a.m_socket->send(d);
+	sleep(1);
+	BOOST_REQUIRE_EQUAL(true, a.success);
+}
+
+BOOST_AUTO_TEST_SUITE_END()
+
```
subtly: When I went to soapui.org the only thing I saw was advertisements trying to sell me certificates and training sessions for ~400eur and several links to wikipedia.

There aren't any 'optimization features' to remove. The optimizations mentioned in todos are not mandatory and they happen to be minor tasks.

subtly:
@chriseth Re: separate network operations. I understand that is a priority after this sprint.

subtly:
Next steps:
1/ update connect() methods so Host can be started with default bootstrap node and join()
2/ cryptographic handshake (log success via insert to CapDesc peer table)
3/ solidify interfaces
4/ dry/refactor node/peer tables
5/ double-check thread-saftey

sveneh:
Do you already want reviews? Asking because all builds are still failing.

subtly:
Not quite ready for codereview but I wanted to update PR w/current status.

This is getting close, and is working. Creating two hosts with different keypairs and then adding host A's node info to host B causes host B to discover host A, which then causes host A to discovery host B, resulting in host A and B connecting to each other. They both connect to each other at the same time because they have no way of knowing that they're finding out about each other at the same time (with only two hosts, and an ideal peer count of 5, they're both ideal for each other). The host which tries to connect last fails connecting, of course, as it'll already be connected; this is intended functionality.

subtly:
It would be great to eyes on this even though it is not complete. The remaining tasks will double PR, so it'd be great to have a review first. Before merging, import/export needs updated, endpoint logic must be revisited, and peer/session information is to be consolidated.

libp2p/Host.cpp
```diff
@@ -167,94 +172,50 @@ void Host::registerPeer(std::shared_ptr<Session> _s, CapDescs const& _caps)
...
+void Host::seal(bytes& _b)
 {
-	RecursiveGuard l(x_peers);
-	Nodes ret;
-
-	auto ns = (m_netPrefs.localNetworking ? _known : (m_private + _known)).inverted();
-	for (auto i: ns)
-		ret.push_back(*m_nodes[m_nodesList[i]]);
-	return ret;
+	_b[0] = 0x22;
```
sveneh: use constants?

libp2p/Host.cpp
```diff
@@ -609,34 +491,17 @@ void Host::run(boost::system::error_code const&)
...
-			if (auto pp = p.second.lock())
-				if (chrono::steady_clock::now() - pp->m_lastReceived > chrono::seconds(60))
-					pp->disconnect(PingTimeout);
-		pingAll();
-	}
+	for (auto p: m_sessions)
+		if (auto pp = p.second.lock())
+			pp->serviceNodesRequest();
+	
+	if (chrono::steady_clock::now() - m_lastPing >= chrono::seconds(30))	// ping every 30s.
```
sveneh: Should be a constant.

libp2p/Host.cpp
```diff
@@ -702,19 +575,19 @@ bytes Host::saveNodes() const
...
-		RecursiveGuard l(x_peers);
-		for (auto const& i: m_nodes)
+		RecursiveGuard l(x_sessions);
+		for (auto const& i: m_peers)
 		{
-			Node const& n = *(i.second);
+			PeerInfo const& n = *(i.second);
 			// TODO: PoC-7: Figure out why it ever shares these ports.//n.address.port() >= 30300 && n.address.port() <= 30305 &&
-			if (!n.dead && chrono::system_clock::now() - n.lastConnected < chrono::seconds(3600 * 48) && n.address.port() > 0 && n.address.port() < /*49152*/32768 && n.id != id() && !isPrivateAddress(n.address.address()))
+			if (chrono::system_clock::now() - n.lastConnected < chrono::seconds(3600 * 48) && n.address.port() > 0 && n.address.port() < /*49152*/32768 && n.id != id() && !isPrivateAddress(n.address.address()))
```
sveneh: convert to constants

alethzero/MainWin.cpp
```diff
@@ -942,18 +942,17 @@ void Main::refreshNetwork()
...
+		for (p2p::PeerInfo const& i: ns)
+			ui->nodes->insertItem(clients.count(i.id) ? 0 : ui->nodes->count(), QString("[%1 %3] %2 - ( =%5s | /%4s%6 ) - *%7 $%8")
+						   .arg(QString::fromStdString(i.id.abridged()))
+						   .arg(QString::fromStdString(toString(i.address)))
+						   .arg(i.id == web3()->id() ? "self" : clients.count(i.id) ? clients[i.id] : i.secondsSinceLastAttempted() == -1 ? "session-fail" : i.secondsSinceLastAttempted() >= (int)i.fallbackSeconds() ? "retrying..." : "retry-" + QString::number(i.fallbackSeconds() - i.secondsSinceLastAttempted()) + "s")
+						   .arg(i.secondsSinceLastAttempted())
+						   .arg(i.secondsSinceLastConnected())
+						   .arg(i.isOffline() ? " | " + QString::fromStdString(reasonOf(i.lastDisconnect)) + " | " + QString::number(i.failedAttempts) + "x" : "")
+						   .arg(i.rating)
+						   .arg(0 /* (int)i.idOrigin */)
```
sveneh: .arg(0 /* (int)i.idOrigin */) Is that a temporary change? Leftover?

libp2p/Common.h
```diff
@@ -116,7 +116,7 @@ typedef std::pair<std::string, u256> CapDesc;
 typedef std::set<CapDesc> CapDescSet;
 typedef std::vector<CapDesc> CapDescs;
 
-struct PeerInfo
+struct PeerSessionInfo
```
sveneh: If this is to be code reviewed by external auditors, at least the .h files should have more comments. I assume an auditor would ask himself at least two questions: What is a PeersSessionInfo used for? What are "notes".

libp2p/Host.h
```diff
@@ -50,38 +51,38 @@ namespace p2p
...
-	Perfect,
-};
-
-struct Node
+struct PeerInfo
 {
 	NodeId id;										///< Their id/public key.
-	unsigned index;									///< Index into m_nodesList
+	
+	// p2p: move to NodeIPEndpoint
```
sveneh: Are p2p: comments internal TODOs for yourself?

libp2p/Host.h
```diff
@@ -120,6 +136,9 @@ class Host: public Worker
 	/// Will block on network process events.
 	virtual ~Host();
 	
+	/// Default host for current version of client.
+	static std::string pocHost();
```
sveneh: does poc come from Proof-of-concept? Then this is just a temporary name?

libp2p/Host.h
```diff
@@ -129,27 +148,20 @@ class Host: public Worker
...
-	void connect(bi::tcp::endpoint const& _ep);
-	void connect(std::shared_ptr<Node> const& _n);
-
-	/// @returns true iff we have a peer of the given id.
-	bool havePeer(NodeId _id) const;
-
+	
+	bool havePeerSession(NodeId _id) { RecursiveGuard l(x_sessions); if (m_sessions.count(_id)) return !!m_sessions[_id].lock(); else return false; }
+	
+	/// Add node.
```
sveneh: a comment like this does not add any value ;-)

libp2p/Session.h
```diff
@@ -39,7 +39,7 @@ namespace dev
 namespace p2p
 {
 
-struct Node;
+struct PeerInfo;
 
 /**
  * @brief The Session class
```
sveneh: this is not a very helpful comment. Either change or delete.

libp2p/Session.h
```diff
@@ -110,14 +107,12 @@ class Session: public std::enable_shared_from_this<Session>
 	std::array<byte, 65536> m_data;			///< Buffer for ingress packet data.
 	bytes m_incoming;						///< Read buffer for ingress bytes.
 
-	PeerInfo m_info;						///< Dynamic information about this peer.
-
 	unsigned m_protocolVersion = 0;			///< The protocol version of the peer.
```
sveneh: Host::protocolVersion() const returns 3, this returns 0. Is that correct?

test/peer.cpp
```diff
@@ -49,14 +71,14 @@ int peerTest(int argc, char** argv)
...
-		ph.connect(remoteHost, remotePort);
+		ph.addNode(NodeId(), remoteHost, remotePort, remotePort);
 
-	for (int i = 0; ; ++i)
-	{
-		this_thread::sleep_for(chrono::milliseconds(100));
-		if (!(i % 10))
-			ph.pingAll();
-	}
+//	for (int i = 0; ; ++i)
```
chriseth: Add a note why this is commented out and not removed?

libp2p/NodeTable.h
```diff
@@ -32,26 +33,112 @@ namespace p2p
 {
 
 /**
- * NodeTable using S/Kademlia system for node discovery and preference.
+ * @brief IPv4,UDP/TCP endpoints.
+ */
+struct NodeIPEndpoint
+{
+	NodeIPEndpoint(): udp(bi::udp::endpoint()), tcp(bi::tcp::endpoint()) {}
+	NodeIPEndpoint(bi::udp::endpoint _udp): udp(_udp) {}
```
chriseth: non-explicit on purpose?

libp2p/NodeTable.h
```diff
@@ -32,26 +33,112 @@ namespace p2p
...
+	NodeIPEndpoint(bi::tcp::endpoint _tcp): tcp(_tcp) {}
+	NodeIPEndpoint(bi::udp::endpoint _udp, bi::tcp::endpoint _tcp): udp(_udp), tcp(_tcp) {}
+
+	bi::udp::endpoint udp;
+	bi::tcp::endpoint tcp;
+	
+	operator bool() const { return udp.address().is_unspecified() && tcp.address().is_unspecified(); }
```
chriseth: Usual semantics of bool conversion is "valid". This seems to be "invalid".

libp2p/NodeTable.h
```diff
@@ -32,26 +33,112 @@ namespace p2p
...
+/**
+ * NodeEntry
+ * @brief Entry in Node Table
+ */
+struct NodeEntry: public Node
+{
+	NodeEntry(Node _src, Public _pubk, NodeIPEndpoint _gw);
+	NodeEntry(Node _src, Public _pubk, bi::udp::endpoint _udp);
+
+	const unsigned distance;	///< Node's distance (xor of _src as integer).
```
chriseth: `unsigned const`

libp2p/NodeTable.h
```diff
@@ -183,21 +250,24 @@ class NodeTable: UDPSocketEvents, public std::enable_shared_from_this<NodeTable>
 	/// Sends FindNeighbor packet. See doFindNode.
 	void requestNeighbours(NodeEntry const& _node, NodeId _target) const;
 
+	std::unique_ptr<NodeTableEventHandler> m_nodeEvents;		///< Event handler for node events.
```
chriseth: From the name it looks like a list of events, but it actually is an event handler.

libp2p/Host.cpp
```diff
@@ -37,6 +38,13 @@ using namespace std;
 using namespace dev;
 using namespace dev::p2p;
 
+HostNodeTableHandler::HostNodeTableHandler(Host& _host): m_host(_host) {}
+
+void HostNodeTableHandler::processEvent(NodeId _n, NodeTableEventType _e)
```
chriseth: Pass by const reference

libp2p/Host.cpp
```diff
@@ -137,27 +145,24 @@ void Host::doneWorking()
...
-		return;
-	}
-
-	{
-		RecursiveGuard l(x_peers);
-		m_peers[_s->m_node->id] = _s;
+		RecursiveGuard l(x_sessions);
+		if (!m_peers.count(_s->m_peer->id))
+			m_peers[_s->m_peer->id] = _s->m_peer;
+		m_sessions[_s->m_peer->id] = _s;
```
chriseth: Why store the session in the case where we do not store the peer?

libp2p/Host.cpp
```diff
@@ -328,21 +289,17 @@ void Host::runAcceptor()
 	{
 		clog(NetConnect) << "Listening on local port " << m_listenPort << " (public: " << m_tcpPublic << ")";
 		m_accepting = true;
-		m_socket.reset(new bi::tcp::socket(m_ioService));
-		m_tcp4Acceptor.async_accept(*m_socket, [=](boost::system::error_code ec)
+		
+		bi::tcp::socket* s = new bi::tcp::socket(m_ioService);
+		m_tcp4Acceptor.async_accept(*s, [=](boost::system::error_code ec)
```
chriseth: Is the lambda guaranteed to be called? Otherwise we leak a socket. Perhaps use a unique_ptr? In this case, it gets deallocated even if the lambda is not called (and only destroyed).

libp2p/Host.cpp
```diff
@@ -355,20 +312,42 @@ void Host::runAcceptor()
...
 			{
 				boost::system::error_code ec;
-				m_socket->shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec);
-				m_socket->close();
+				s->shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec);
+				s->close();
 			}
 
 			m_accepting = false;
+			delete s;
```
chriseth: If I see that correctly, `*s` is `std::move`d inside `doHandshake`, so calling a constructor on it can crash.

libp2p/Host.cpp
```diff
@@ -355,20 +312,42 @@ void Host::runAcceptor()
 				}
 			}
 			
-			if (!success && m_socket->is_open())
+			if (!success && s->is_open())
```
chriseth: This code path could result from an exception being caught above. I'm not sure if it is still safe to access `s`.

libp2p/Host.cpp
```diff
@@ -376,65 +355,53 @@ string Host::pocHost()
...
+			bi::tcp::endpoint tcp = *_epIt;
+			addNode(Node(_node, NodeIPEndpoint(bi::udp::endpoint(tcp.address(), _udpNodePort), tcp)));
+		});
 	}
+	else
+		addNode(Node(_node, NodeIPEndpoint(bi::udp::endpoint(addr, _udpNodePort), bi::tcp::endpoint(addr, _tcpPeerPort))));
 }
 
-void Host::connect(bi::tcp::endpoint const& _ep)
+void Host::connect(std::shared_ptr<PeerInfo> const& _n)
```
chriseth: Why is `_n` the canonical name for a `shared_ptr` to `PeerInfo`?

libp2p/Host.cpp
```diff
@@ -445,50 +412,32 @@ void Host::connect(std::shared_ptr<Node> const& _n)
...
-	return !!m_peers.count(_id);
+		{
+			clog(NetConnect) << "Connected to" << _n->id.abridged() << "@" << _n->address;
+			
+			_n->lastConnected = std::chrono::system_clock::now();
+			auto ps = make_shared<Session>(this, std::move(*s), _n);
+			ps->start();
+			
+		}
+		delete s;
```
chriseth: `delete s` after `std::move(*s)`.

libp2p/Host.cpp
```diff
@@ -688,12 +555,18 @@ void Host::doWork()
...
 {
-	RecursiveGuard l(x_peers);
-	for (auto& i: m_peers)
-		if (auto j = i.second.lock())
-			j->ping();
+	RecursiveGuard l(x_sessions);
+	for (auto p: m_sessions)
+		if (auto pp = p.second.lock())
+		{
+			if (chrono::steady_clock::now() - pp->m_lastReceived >= chrono::seconds(60))
```
chriseth: The ping interval (30s) and this number (60s) should be defined close to each other.

libp2p/Host.cpp
```diff
@@ -730,16 +603,15 @@ bytes Host::saveNodes() const
...
 	RLP r(_b);
 	if (r.itemCount() > 0 && r[0].isInt())
 		switch (r[0].toInt<int>())
 		{
 		case 0:
 		{
-			auto oldId = id();
 			m_key = KeyPair(r[1].toHash<Secret>());
-			noteNode(id(), m_tcpPublic, Origin::Perfect, false, oldId);
+//			noteNode(id(), m_tcpPublic);
```
chriseth: Why is this commented out?

libp2p/Host.cpp
```diff
@@ -702,19 +575,19 @@ bytes Host::saveNodes() const
...
-			if (!n.dead && chrono::system_clock::now() - n.lastConnected < chrono::seconds(3600 * 48) && n.address.port() > 0 && n.address.port() < /*49152*/32768 && n.id != id() && !isPrivateAddress(n.address.address()))
+			if (chrono::system_clock::now() - n.lastConnected < chrono::seconds(3600 * 48) && n.address.port() > 0 && n.address.port() < /*49152*/32768 && n.id != id() && !isPrivateAddress(n.address.address()))
 			{
 				nodes.appendList(10);
 				if (n.address.address().is_v4())
 					nodes << n.address.address().to_v4().to_bytes();
 				else
 					nodes << n.address.address().to_v6().to_bytes();
-				nodes << n.address.port() << n.id << (int)n.idOrigin
+				nodes << n.address.port() << n.id /* << (int)n.idOrigin */ << 0
```
chriseth: Why is this commented out?

libp2p/Host.cpp
```diff
@@ -749,16 +621,16 @@ void Host::restoreNodes(bytesConstRef _b)
...
 				{
-					auto o = (Origin)i[3].toInt<int>();
-					auto n = noteNode(id, ep, o, true);
-					n->lastConnected = chrono::system_clock::time_point(chrono::seconds(i[4].toInt<unsigned>()));
-					n->lastAttempted = chrono::system_clock::time_point(chrono::seconds(i[5].toInt<unsigned>()));
-					n->failedAttempts = i[6].toInt<unsigned>();
-					n->lastDisconnect = (DisconnectReason)i[7].toInt<unsigned>();
-					n->score = (int)i[8].toInt<unsigned>();
-					n->rating = (int)i[9].toInt<unsigned>();
+////					auto o = (Origin)i[3].toInt<int>();
```
chriseth: Why is this commented out?

libp2p/Host.cpp
```diff
@@ -768,14 +640,38 @@ void Host::restoreNodes(bytesConstRef _b)
...
-			if (!m_nodes.count(id))
+			if (!m_peers.count(id))
 			{
 				bi::tcp::endpoint ep;
 				if (i[0].itemCount() == 4)
 					ep = bi::tcp::endpoint(bi::address_v4(i[0].toArray<byte, 4>()), i[1].toInt<short>());
 				else
 					ep = bi::tcp::endpoint(bi::address_v6(i[0].toArray<byte, 16>()), i[1].toInt<short>());
-				auto n = noteNode(id, ep, Origin::Self, true);
+//				auto n = noteNode(id, ep);
```
chriseth: Why is this commented out?

libp2p/Host.h
```diff
@@ -129,27 +148,20 @@ class Host: public Worker
...
-	static std::string pocHost();
-	void connect(std::string const& _addr, unsigned short _port = 30303) noexcept;
-	void connect(bi::tcp::endpoint const& _ep);
-	void connect(std::shared_ptr<Node> const& _n);
-
-	/// @returns true iff we have a peer of the given id.
-	bool havePeer(NodeId _id) const;
-
+	
+	bool havePeerSession(NodeId _id) { RecursiveGuard l(x_sessions); if (m_sessions.count(_id)) return !!m_sessions[_id].lock(); else return false; }
```
chriseth: Use ternary operator

libp2p/Host.h
```diff
@@ -178,20 +191,28 @@ class Host: public Worker
 
 	void registerPeer(std::shared_ptr<Session> _s, CapDescs const& _caps);
 
-	std::shared_ptr<Node> node(NodeId _id) const { if (m_nodes.count(_id)) return m_nodes.at(_id); return std::shared_ptr<Node>(); }
+//	std::shared_ptr<PeerInfo> node(NodeId _id) const { if (m_nodes.count(_id)) return m_nodes.at(_id); return std::shared_ptr<PeerInfo>(); }
```
chriseth: Why is this commented out?

libp2p/Host.h
```diff
@@ -178,20 +191,28 @@ class Host: public Worker
...
 	
+	void connect(std::shared_ptr<PeerInfo> const& _n);
+	
+	/// Ping the peers to update the latency information and disconnect peers which have timed out.
+	void keepAlivePeers();
+	
 	/// Called only from startedWorking().
 	void runAcceptor();
 	
+	/// Handler for verifying handshake siganture before creating session. _nodeId is passed for outbound connections.
```
chriseth: If `_socket` stays a plain pointer, please put ownership information into the documentation.

libp2p/NodeTable.cpp
```diff
@@ -74,10 +116,18 @@ list<NodeTable::NodeEntry> NodeTable::state() const
 	return move(ret);
 }
 
-NodeTable::NodeEntry NodeTable::operator[](NodeId _id)
+Node NodeTable::operator[](NodeId _id)
 {
 	Guard l(x_nodes);
-	return *m_nodes[_id];
+	auto n = m_nodes[_id];
+	return !!n ? *n : Node();
```
chriseth: Return by reference with a static Node() instance in this function?

libp2p/NodeTable.cpp
```diff
@@ -348,7 +388,19 @@ void NodeTable::onReceived(UDPSocketFace*, bi::udp::endpoint const& _from, bytes
...
+				for (auto it = m_evictions.begin(); it != m_evictions.end(); it++)
+					if (it->first.first == nodeid && it->first.second > std::chrono::steady_clock::now())
+					{
+						if (auto n = getNodeEntry(it->second))
+							dropNode(n);
+						
+						if (auto n = (*this)[it->first.first])
+							addNode(n);
+						
+						m_evictions.erase(it);
```
chriseth: erase invalidates the iterator!

libp2p/NodeTable.h
```diff
@@ -32,26 +33,112 @@ namespace p2p
...
+		for (auto const& e: events)
+			processEvent(e.first, e.second);
+	}
+	
+	/// Called by NodeTable to append event.
+	virtual void appendEvent(NodeId _n, NodeTableEventType _e) { Guard l(x_events); m_nodeEvents.push_back(_n); m_events[_n] = _e; }
+	
+	Mutex x_events;
+	std::list<NodeId> m_nodeEvents;
+	std::map<NodeId,NodeTableEventType> m_events;
```
chriseth: Space after comma.

libp2p/NodeTable.h
```diff
@@ -133,16 +180,36 @@ class NodeTable: UDPSocketEvents, public std::enable_shared_from_this<NodeTable>
...
+
 	void join();
 	
 	NodeEntry root() const { return NodeEntry(m_node, m_node.publicKey(), m_node.endpoint.udp); }
 	std::list<NodeId> nodes() const;
 	std::list<NodeEntry> state() const;
 	
-	NodeEntry operator[](NodeId _id);
-	
+	bool haveNode(NodeId _id) { Guard l(x_nodes); return !!m_nodes[_id]; }
```
chriseth: Do you really want to create the map entry?

libp2p/Session.cpp
```diff
@@ -202,57 +174,52 @@ bool Session::interpret(RLP const& _r)
...
 			return true;
 		}
-		m_info = PeerInfo({id, clientVersion, m_socket.remote_endpoint().address().to_string(), listenPort, std::chrono::steady_clock::duration(), _r[3].toSet<CapDesc>(), (unsigned)m_socket.native_handle(), map<string, string>() });
+		
+		m_info.clientVersion = clientVersion;
+		m_info.host = m_socket.remote_endpoint().address().to_string();
+		m_info.port = listenPort;
+		m_info.lastPing = std::chrono::steady_clock::duration();
+		m_info.caps = _r[3].toSet<CapDesc>();
+		m_info.socket = (unsigned)m_socket.native_handle();
```
chriseth: Is this cast fine on all platforms?

libp2p/Session.cpp
```diff
@@ -307,62 +282,37 @@ bool Session::interpret(RLP const& _r)
...
-			if (isPrivateAddress(peerAddress) && !m_server->m_netPrefs.localNetworking)
+			
+			clogS(NetAllDetail) << "Checking: " << ep << "(" << id.abridged() << ")";
+//			clogS(NetAllDetail) << "Checking: " << ep << "(" << id.abridged() << ")" << isPrivateAddress(peerAddress) << this->id().abridged() << isPrivateAddress(endpoint().address()) << m_server->m_peers.count(id) << (m_server->m_peers.count(id) ? isPrivateAddress(m_server->m_peers.at(id)->address.address()) : -1);
+
+			// ignore if dist(us,item) - dist(us,them) > 1
+			
+			// TODO: isPrivate
+			if (!m_server->m_netPrefs.localNetworking && isPrivateAddress(peerAddress))
 				goto CONTINUE;	// Private address. Ignore.
```
chriseth: If this function really needs 200 lines, then please at least don't use gotos...

libp2p/Session.cpp
```diff
@@ -307,62 +282,37 @@ bool Session::interpret(RLP const& _r)
...
-			// SECURITY: Not a valid assumption in general. Should compare ID origins and pick the best or note uncertainty and weight each equally.
-			for (auto const& i: m_server->m_nodes)
-				if (i.second->address == ep)
-					goto CONTINUE;		// Same address but a different node.
+				goto LAMEPEER;	// Private port according to IANA.
 
 			// OK passed all our checks. Assume it's good.
 			addRating(1000);
-			m_server->noteNode(id, ep, m_node->idOrigin == Origin::Perfect ? Origin::PerfectThird : Origin::SelfThird, true);
+			m_server->addNode(Node(id, NodeIPEndpoint(bi::udp::endpoint(ep.address(), 30303), ep)));
```
chriseth: That port number could also deserve a name.

test/peer.cpp
```diff
@@ -27,6 +28,27 @@ using namespace std;
...
+	Host host1("Test", host1prefs);
+	host1.start();
+		
+	Host host2("Test", host2prefs);
+	auto node2 = host2.id();
+	host2.start();
+	
+	host1.addNode(node2, "127.0.0.1", host2prefs.listenPort, host2prefs.listenPort);
+	
+	this_thread::sleep_for(chrono::seconds(1));
```
chriseth: What does this test test?

alethzero/MainWin.cpp
subtly: Both. This flag maybe removed with some other piece of information (likely: if this peer is 'required').

libp2p/Common.h
subtly: Afaik, this code won't be reviewed by auditors. This struct contains session info about the peer, as is described by the properties of the struct. Notes aren't touched by this PR, so I'm not sure. Docs, of course, are always helpful.

libp2p/Host.h
subtly: Re: This comment and others. This is an interim PR for a work-in-progress. Small steps -- this won't be merged into develop. Hopefully tomorrows code will be merged into develop, though :)

libp2p/Session.h
subtly: This is not a very helpful comment :) Please stick to changes and not old and unchanged code. Perhaps we can establish a time for reviewing legacy code however doing this in a PR is distracting and takes time away from both of us from what is important, which is code that's changed. If I took the time to change, DRY, and refactor unrelated code then the PR will have a lot of noise and it'll be difficult to determine what actually changed.

libp2p/Session.h
subtly: What does this have to do with Host::protocolVersion()?
Also, this code isn't within scope of the PR.

libp2p/Host.cpp
subtly: Did you mean destructor? I need to look into this. I think that 'moved' pointers can still be deleted. There may also be a proper way to do this with unique_ptr.

libp2p/Host.h
subtly: Nope. See implementation of pocHost.

libp2p/Host.cpp
subtly: TL;DR -- Temporary; will be fixed w/tomorrow's PR. This is the middle-point of integration; post-integration, a session will not be created unless the peer exists. 
In this PR, m_peers will act as a storage backend. Pruning peers will come later and require a mutex in connect and disconnect events; for now we just do this when Host is deallocated. Even better, when m_peers is a first-class storage backend, it'll take care of itself.

libp2p/Host.cpp
subtly: Yes. As I understand it, the lambda is guaranteed to be called; if it is not, the client is accepting.

chriseth: I do not understand you response. If the client is accepting, the lambda is not called? In that case, s is not deallocated. PLEASE use modern memory management concepts, especially if there is only a tiny grain of uncertainty about the lifetime of objects or the call structure of the code.

libp2p/Host.cpp
chriseth: Yes, I meant destructor, and you are right, the documentation says the socket should be in a valid state.
Still, this code could benefit a lot from a bit more RAII. Why do you have to manually call shutdown and close in case of failure, shouldn't a proper destructor do that?

libp2p/Host.cpp
subtly: Typo. The last word of prev comment should read 'exiting' and not accepting.
The acceptor is added to asio and when the acceptor is closed, asio will call the lamda with an error indicating acceptor was closed.

I'll see what I can do but for now it may remain allocated and deleted instead of a managed ptr. The reason for this is scope; it's been working for awhile as-is and the proper solution is abstracting boost's acceptor/socket/connect lifecycle into a "Connection". This has been implemented in another branch but present objective is node discovery. The reason I DID change this is because m_socket should not be a property of Host; it needs to be deallocated with respect to the acceptor (this lambda) and not Host.

libp2p/Host.cpp
subtly: Yep. ASIO is pre-RAII, doesn't shutdown the socket. Will abstract this to it's own class later when UDP sockets can also be considered, such that there's a cohesive API.

libp2p/Session.cpp
subtly: Hopefully it is. We've been using it for awhile. m_info.socket has only been used for logging and will be removed w/a refactor.

libp2p/Host.cpp
subtly: For now this is placeholder for framing; for now, the function is a canonical definition.

libp2p/Host.cpp
subtly: This corresponds w/todos for implementing import/export of peers; to be implemented after all other code-review passes. (see follow-up comment regarding remaining tasks)

subtly:
This is ready for code-review again; as follow-up to previous gotissues. There is one remaining task: import/export. Then interop and/or merge :)

libp2p/Host.cpp
chriseth: Perfect, then make it a shared_ptr that is bound by value to the lambda and thus the pointer is deallocated together with the lambad, unless it is passed on.

libp2p/Host.cpp
```diff
@@ -442,155 +423,39 @@ void Host::connect(std::shared_ptr<Node> const& _n)
...
-			i = m_peers.erase(i);
+		{
+			clog(NetConnect) << "Connected to" << _p->id.abridged() << "@" << _p->peerEndpoint();
+			
+			_p->lastConnected = std::chrono::system_clock::now();
+			auto ps = make_shared<Session>(this, std::move(*s), _p);
+			ps->start();
+			
+		}
+		delete s;
```
chriseth: Is it fine to delete a socket from one of its callbacks?

libp2p/NodeTable.cpp
```diff
@@ -74,10 +116,18 @@ list<NodeTable::NodeEntry> NodeTable::state() const
 	return move(ret);
 }
 
-NodeTable::NodeEntry NodeTable::operator[](NodeId _id)
+Node NodeTable::operator[](NodeId _id)
+{
+	Guard l(x_nodes);
+	auto n = m_nodes[_id];
```
chriseth: Would be better to use `find` so that we do not create a new entry at _id if the key is not present.

libp2p/Host.cpp
```diff
@@ -688,33 +543,53 @@ void Host::doWork()
...
-	for (auto& i: m_peers)
-		if (auto j = i.second.lock())
-			j->ping();
+	if (chrono::steady_clock::now() - c_keepAliveInterval < m_lastPing)
+		return;
+	
+	RecursiveGuard l(x_sessions);
+	for (auto p: m_sessions)
+		if (auto pp = p.second.lock())
+				pp->ping();
```
chriseth: Indentation.

libp2p/Session.cpp
chriseth: Static assert does not hurt, even for code that is planned to change later (which often never happens).

PR #662, opened 2014-12-18
boost InvalidNonce exception, issue #660

subtly:


gavofyork:
note to self: change static to thread-local static.

libethereum/Executive.cpp
```diff
@@ -56,7 +56,7 @@ bool Executive::setup(bytesConstRef _rlp)
 	if (m_t.nonce() != nonceReq)
 	{
 		clog(StateDetail) << "Invalid Nonce: Require" << nonceReq << " Got" << m_t.nonce();
-		BOOST_THROW_EXCEPTION(InvalidNonce(nonceReq, m_t.nonce()));
+		throw InvalidNonce(nonceReq, m_t.nonce());
```
CJentzsch: Can you explain why this is necessary?

gavofyork: It's not - was just there for debugging purposes.

PR #667, opened 2014-12-18
Possibility for binary operators to yield types different from their operands'.

chriseth:


libsolidity/Types.h
```diff
@@ -389,6 +408,12 @@ class TypeType: public Type
 	virtual bool canLiveOutsideStorage() const override { return false; }
 	virtual std::string toString() const override { return "type(" + m_actualType->toString() + ")"; }
 
+protected:
+	virtual TypePointer binaryOperatorResultImpl(Token::Value _operator, TypePointer const& _this, TypePointer const& _other) const override
+	{
+		return TypePointer();
```
subtly: Are these methods WIP?

chriseth: No, this simply means that you cannot apply any binary operator to a TypeType.

Issue #714, opened 2015-01-02
eth.defaultBlock not being used for stateAt and call


caktux:
This results in every call and stateAt to include pending transactions, making it impossible to differentiate between pending and mined state.

PR #715, opened 2015-01-02
use defaultBlock for codeAt, balanceAt and stateAt

caktux:
Fixes part of #714

### DIFF
--- a/libweb3jsonrpc/WebThreeStubServer.cpp
+++ b/libweb3jsonrpc/WebThreeStubServer.cpp
@@ -266,7 +266,7 @@ std::string WebThreeStubServer::shh_addToGroup(std::string const& _group, std::s
 
 std::string WebThreeStubServer::eth_balanceAt(string const& _address)
 {
-	int block = 0;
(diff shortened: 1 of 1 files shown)


libweb3jsonrpc/WebThreeStubServer.cpp
```diff
@@ -266,7 +266,7 @@ std::string WebThreeStubServer::shh_addToGroup(std::string const& _group, std::s
 
 std::string WebThreeStubServer::eth_balanceAt(string const& _address)
 {
-	int block = 0;
+	int block = client() ? client()->getDefault() : 0;
 	return toJS(client()->balanceAt(jsToAddress(_address), block));
```
debris: if client() does not exists, client()->balanceAt() does not make any sense.

debris:
@gavofyork can we assume, that client() always exists? if it does not, API is useless anyway and it should throw some exception. it will simplify this "if" statemants.

gavofyork:
A quick look at the code suggests that unguarded use of client() is safe.

debris:
fixed in pr #723

PR #716, opened 2015-01-03
fix compile on OSX

caktux:


debris:
issue is fixed in #716

PR #717, opened 2015-01-05
different way of fixing #716

debris:
header include order is broken. that should fix an issue on osx.


gavofyork:
rebase?

debris:
it is the same as 969ea19a44ff258b54c082d7920ffb8c8886d4a6. i'm closing.

PR #726, opened 2015-01-05
bugfix in trie test

CJentzsch:


gavofyork:
`BOOST_REQUIRE(t.check(true));` is an invariant - it should be assertable regardless of the nature of the change to the trie.

CJentzsch:
If I place it after the remove, it returns false. Unless I call t.check(false).

PR #733, opened 2015-01-06
eth --version display protocol version

CJentzsch:
closes https://github.com/ethereum/cpp-ethereum/issues/732


sveneh:
thanks, that was fast :grinning:

PR #738, opened 2015-01-07
Canonical Function signature creation in solidity

LefterisJP:
Creating the canonical function signature in solidity as requested by https://github.com/ethereum/wiki/wiki/Ethereum-Contract-ABI

### DIFF
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -122,6 +122,7 @@ std::string FunctionDefinition::getCanonicalSignature()
 		if (i != parameters.size()) {
 			ret += ",";
 		}
+		i++;
 	}
 	ret += ")";
 	return ret;
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -110,6 +110,23 @@ void FunctionDefinition::checkTypeRequirements()
 	m_body->checkTypeRequirements();
 }
 
+std::string FunctionDefinition::getCanonicalSignature()
+{
+	auto parameters = getParameters();
+	std::string ret = getName() + "(";
+	unsigned int i = 1;
+
+	for (ASTPointer<VariableDeclaration> const& member: parameters)
+	{
+		ret += member->getType()->toString();
+		if (i != parameters.size()) {
+			ret += ",";
+		}
+	}
+	ret += ")";
+	return ret;
+}
+
 void Block::checkTypeRequirements()
 {
 	for (shared_ptr<Statement> const& statement: m_statements)
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -110,14 +110,9 @@ void FunctionDefinition::checkTypeRequirements()
 	m_body->checkTypeRequirements();
 }
 
-std::string FunctionDefinition::getCanonicalSignature()
+string FunctionDefinition::getCanonicalSignature() const
 {
-	auto parameters = getParameters();
-	std::string ret = getName() + "(";
-
-	for (auto it = parameters.cbegin(); it != parameters.cend(); ++it)
-		ret += (*it)->getType()->toString() + (it + 1 == parameters.end() ? "" : ",");
-	return ret + ")";
+	return getName() + FunctionType(*this).getCanonicalSignature();
 }
 
 void Block::checkTypeRequirements()
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -114,18 +114,10 @@ std::string FunctionDefinition::getCanonicalSignature()
 {
 	auto parameters = getParameters();
 	std::string ret = getName() + "(";
-	unsigned int i = 1;
 
-	for (ASTPointer<VariableDeclaration> const& member: parameters)
-	{
-		ret += member->getType()->toString();
-		if (i != parameters.size()) {
-			ret += ",";
-		}
-		i++;
-	}
-	ret += ")";
-	return ret;
+	for (auto it = parameters.cbegin(); it != parameters.cend(); ++it)
+		ret += (*it)->getType()->toString() + (it + 1 == parameters.end() ? "" : ",");
+	return ret + ")";
 }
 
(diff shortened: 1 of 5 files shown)


libsolidity/AST.cpp
```diff
@@ -110,6 +110,16 @@ void FunctionDefinition::checkTypeRequirements()
 	m_body->checkTypeRequirements();
 }
 
+std::string FunctionDefinition::getCanonicalSignature()
```
chriseth: Please don't use `std::` here.

libsolidity/AST.h
```diff
@@ -277,6 +277,12 @@ class FunctionDefinition: public Declaration
 	/// Checks that all parameters have allowed types and calls checkTypeRequirements on the body.
 	void checkTypeRequirements();
 
+	/// Returns the canonical signature of the function
+	/// That consists of the name of the function followed by the
+	/// types of the arguments separated by commas all enclosed in parentheses
+	/// without any spaces
+	std::string getCanonicalSignature();
```
chriseth: const

libsolidity/AST.h
```diff
@@ -277,6 +277,12 @@ class FunctionDefinition: public Declaration
 	/// Checks that all parameters have allowed types and calls checkTypeRequirements on the body.
 	void checkTypeRequirements();
 
+	/// Returns the canonical signature of the function
```
chriseth: `@returns` and please use `.` to end a sentence :-)

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -321,6 +323,23 @@ BOOST_AUTO_TEST_CASE(comparison_bitop_precedence)
 	BOOST_CHECK_NO_THROW(parseTextAndResolveNames(text));
 }
 
+BOOST_AUTO_TEST_CASE(function_canonical_signature)
```
chriseth: Please don't forget to add a test with "uint", "hash" (which should be uint256 and hash256 in the signature) and "address" (which should stay as it is).

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -321,6 +323,23 @@ BOOST_AUTO_TEST_CASE(comparison_bitop_precedence)
...
+					   "  function foo(uint256 arg1, uint64 arg2, bool arg3) returns (uint256 ret) {\n"
+					   "    ret = arg1 + arg2;\n"
+					   "  }\n"
+					   "}\n";
+	BOOST_CHECK_NO_THROW(sourceUnit = parseTextAndResolveNames(text));
+	for (ASTPointer<ASTNode> const& node: sourceUnit->getNodes())
+		if (ContractDefinition* contract = dynamic_cast<ContractDefinition*>(node.get()))
+		{
+			auto functions = contract->getDefinedFunctions();
+			BOOST_ASSERT("foo(uint256,uint64,bool)" == functions[0]->getCanonicalSignature());
```
chriseth: `BOOST_ASSERT_EQ` will generate better error messages here.

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -321,6 +323,23 @@ BOOST_AUTO_TEST_CASE(comparison_bitop_precedence)
 	BOOST_CHECK_NO_THROW(parseTextAndResolveNames(text));
 }
```
chriseth: I think "NameAndTypeResolution" is not the right place for this test. What about ABIJSON?

libsolidity/Types.cpp
```diff
@@ -452,6 +452,17 @@ unsigned FunctionType::getSizeOnStack() const
 	}
 }
 
+std::string FunctionType::getCanonicalSignature() const
+{
+	auto parameters = getParameterTypes();
+	std::string ret = "NAME("; //TODO: how to get function name from FunctionType
+
+	for (auto it = parameters.cbegin(); it != parameters.cend(); ++it)
+		ret += (*it)->toString() + (it + 1 == m_parameterTypes.end() ? "" : ",");
```
chriseth: `cend`

libsolidity/Types.cpp
```diff
@@ -452,6 +452,17 @@ unsigned FunctionType::getSizeOnStack() const
 	}
 }
 
+std::string FunctionType::getCanonicalSignature() const
+{
+	auto parameters = getParameterTypes();
+	std::string ret = "NAME("; //TODO: how to get function name from FunctionType
```
chriseth: As discussed: The function type should not know about the name of its function, so it has to be prepended from a different place.

libsolidity/AST.cpp
```diff
@@ -110,6 +110,16 @@ void FunctionDefinition::checkTypeRequirements()
...
+{
+	auto parameters = getParameters();
+	std::string ret = getName() + "(";
```
chriseth: What about `return getName() + FunctionType(*this).getCanonicalSignature();`?

libsolidity/AST.h
LefterisJP: Okay but doxygen add '.' on its own if you don't add them yourself.

test/SolidityNameAndTypeResolution.cpp
LefterisJP: There is no real good fit anywhere. The ABIJson as the name implies only tests the generated JSON. Maybe the compiler tests would also make sense? But the reason I used name and type resolution is that this has to happen only after NameAndTypeResolution and as such thought it a good enough fit.

libsolidity/AST.cpp
LefterisJP: yeah good! I actually was just thinking about that. I wrote this function first and then noticed the existence of the FunctionType :)

test/SolidityNameAndTypeResolution.cpp
LefterisJP: There is no such function/macro. I assume you meant `BOOST_CHECK_EQUAL`

PR #750, opened 2015-01-08
latest changes from ethereumjs, sha3 method signature

debris:
upgraded ethereum.js version to `0.0.6`
renamed `main.js` -> `web3.js`
sha3 method signature
fixed error in contract example


debris:
update*
added sha3 method signature support

libjsqrc/ethereumjs/lib/contract.js
```diff
@@ -46,8 +46,10 @@ var contract = function (address, desc) {
                 call: function (extra) {
                     extra = extra || {};
                     extra.to = address;
-                    extra.data = parsed;
-                    return web3.eth.call(extra).then(onSuccess);
+                    return abi.methodSignature(desc, method.name).then(function (signature) {
+                        extra.data = signature.slice(0, 10) + parsed;
```
subtly: What is 10 derived from? Can this be declared somewhere?

debris: it's described here in [wiki](https://github.com/ethereum/wiki/wiki/Ethereum-Contract-ABI#examples)

`the Method ID. This is derived as the first 4 bytes of the Keccak hash of the ASCII form of the signature baz(uint32,bool).`

yes, it can

PR #754, opened 2015-01-08
Calculation of ABI Function Identifier Hash

LefterisJP:
Plus ... adjustments all around the code and its tests.


libsolidity/AST.cpp
```diff
@@ -27,6 +27,8 @@
 #include <libsolidity/Exceptions.h>
 #include <libsolidity/AST_accept.h>
 
+#include <libdevcrypto/SHA3.h>
```
chriseth: Move between algorithm and Utils.h (without blank lines)?

libsolidity/AST.cpp
```diff
@@ -50,18 +52,16 @@ void ContractDefinition::checkTypeRequirements()
...
 	for (ASTPointer<FunctionDefinition> const& f: m_definedFunctions)
 		if (f->isPublic() && f->getName() != getName())
-			exportedFunctions.push_back(f.get());
-	auto compareNames = [](FunctionDefinition const* _a, FunctionDefinition const* _b)
-	{
-		return _a->getName().compare(_b->getName()) < 0;
-	};
+		{
+			FixedHash<4> hash(dev::sha3(f->getCanonicalSignature()));
+			exportedFunctions[hash] = f.get();
```
chriseth: solAssert for hash collisions?

libsolidity/Compiler.cpp
```diff
@@ -100,7 +100,7 @@ void Compiler::appendConstructorCall(FunctionDefinition const& _constructor)
 	{
 		m_context << u256(argumentSize);
 		m_context.appendProgramSize();
-		m_context << u256(1); // copy it to byte one as expected for ABI calls
+		m_context << u256(g_functionIdentifierOffset); // copy it to byte four as expected for ABI calls
```
chriseth: Isn't it rather functionIdentifierSize?

libsolidity/Compiler.cpp
```diff
@@ -118,34 +118,30 @@ set<FunctionDefinition const*> Compiler::getFunctionsNeededByConstructor(Functio
 
 void Compiler::appendFunctionSelector(ContractDefinition const& _contract)
 {
-	vector<FunctionDefinition const*> interfaceFunctions = _contract.getInterfaceFunctions();
+	map<FixedHash<4>, FunctionDefinition const*> interfaceFunctions = _contract.getInterfaceFunctions();
 	vector<eth::AssemblyItem> callDataUnpackerEntryPoints;
 
-	if (interfaceFunctions.size() > 255)
-		BOOST_THROW_EXCEPTION(CompilerError() << errinfo_comment("More than 255 public functions for contract."));
+	if (interfaceFunctions.size() > 4294967295) // 2 ** 32
```
chriseth: We probably already have a serious problem before that number, because the probability of hash collisions is almost 1 even for 2**31. Please check for hash collisions somewhere else (already at parse time, perhaps during type checking) and remove this check here.

libsolidity/Compiler.cpp
```diff
@@ -118,34 +118,30 @@ set<FunctionDefinition const*> Compiler::getFunctionsNeededByConstructor(Functio
...
-	// retrieve the first byte of the call data, which determines the called function
-	// @todo This code had a jump table in a previous version which was more efficient but also
-	// error prone (due to the optimizer and variable length tag addresses)
-	m_context << u256(1) << u256(0) // some constants
-			  << eth::dupInstruction(1) << eth::Instruction::CALLDATALOAD
-			  << eth::dupInstruction(2) << eth::Instruction::BYTE
-			  << eth::dupInstruction(2);
+	// retrieve the function signature hash from the calldata
+	m_context << u256(1) << u256(0) << (u256(1) << 224) // some constants
+			  << eth::dupInstruction(2) << eth::Instruction::CALLDATALOAD
```
chriseth: Please use CompilerUtils::loadFromMemory

libsolidity/Compiler.cpp
```diff
@@ -118,34 +118,30 @@ set<FunctionDefinition const*> Compiler::getFunctionsNeededByConstructor(Functio
...
+			  << eth::Instruction::DIV;
 
-	// stack here: 1 0 <funid> 0, stack top will be counted up until it matches funid
-	for (unsigned funid = 0; funid < interfaceFunctions.size(); ++funid)
+	// stack now is: 1 0 <funhash>
+	for (auto it = interfaceFunctions.cbegin(); it != interfaceFunctions.cend(); ++it)
```
chriseth: foreach loop?

libsolidity/Compiler.cpp
```diff
@@ -118,34 +118,30 @@ set<FunctionDefinition const*> Compiler::getFunctionsNeededByConstructor(Functio
...
 	}
 	m_context << eth::Instruction::STOP; // function not found
 
-	for (unsigned funid = 0; funid < interfaceFunctions.size(); ++funid)
+	unsigned funid = 0;
+	for (auto it = interfaceFunctions.cbegin(); it != interfaceFunctions.cend(); ++it, ++funid)
 	{
-		FunctionDefinition const& function = *interfaceFunctions[funid];
+		FunctionDefinition const& function = *it->second;
 		m_context << callDataUnpackerEntryPoints[funid];
```
chriseth: If callDataUnpackerEntryPoints is a hash table, I think the meaning of the lines that use it are much clearer and you can use foreach loops in both parts.

libsolidity/CompilerUtils.h
```diff
@@ -30,6 +30,9 @@ namespace solidity {
 
 class Type; // forward
 
+/// The size in bytes of the function (hash) identifier
+static const unsigned int g_functionIdentifierOffset = 4;
```
chriseth: Please move it inside CompilerUtils.

libsolidity/InterfaceHandler.cpp
```diff
@@ -35,8 +35,9 @@ std::unique_ptr<std::string> InterfaceHandler::getDocumentation(ContractDefiniti
 std::unique_ptr<std::string> InterfaceHandler::getABIInterface(ContractDefinition const& _contractDef)
 {
 	Json::Value methods(Json::arrayValue);
+	auto interfaceFunctions = _contractDef.getInterfaceFunctions();
 
-	for (FunctionDefinition const* f: _contractDef.getInterfaceFunctions())
+	for (auto it = interfaceFunctions.cbegin(); it != interfaceFunctions.cend(); ++it)
```
chriseth: Why not foreach?

libsolidity/InterfaceHandler.cpp
```diff
@@ -68,11 +69,12 @@ std::unique_ptr<std::string> InterfaceHandler::getUserDocumentation(ContractDefi
 {
 	Json::Value doc;
 	Json::Value methods(Json::objectValue);
+	auto interfaceFunctions = _contractDef.getInterfaceFunctions();
 
-	for (FunctionDefinition const* f: _contractDef.getInterfaceFunctions())
+	for (auto it = interfaceFunctions.cbegin(); it != interfaceFunctions.cend(); ++it)
```
chriseth: foreach?

libsolidity/InterfaceHandler.cpp
```diff
@@ -110,10 +112,11 @@ std::unique_ptr<std::string> InterfaceHandler::getDevDocumentation(ContractDefin
 			doc["title"] = m_title;
 	}
 
-	for (FunctionDefinition const* f: _contractDef.getInterfaceFunctions())
+	auto interfaceFunctions = _contractDef.getInterfaceFunctions();
+	for (auto it = interfaceFunctions.cbegin(); it != interfaceFunctions.cend(); ++it)
```
chriseth: foreach?

libsolidity/Types.cpp
```diff
@@ -333,9 +333,10 @@ MemberList const& ContractType::getMembers() const
 	// We need to lazy-initialize it because of recursive references.
 	if (!m_members)
 	{
+		auto interfaceFunctions = m_contract.getInterfaceFunctions();
 		map<string, shared_ptr<Type const>> members;
-		for (FunctionDefinition const* function: m_contract.getInterfaceFunctions())
-			members[function->getName()] = make_shared<FunctionType>(*function, false);
+		for (auto it = interfaceFunctions.cbegin(); it != interfaceFunctions.cend(); ++it)
```
chriseth: foreach? :-)

libsolidity/AST.cpp
LefterisJP: I wanted to separate it as a dependency of a different project (directory)

libsolidity/AST.cpp
LefterisJP: good point!

libsolidity/Compiler.cpp
LefterisJP: True ... then perhaps rename it to `g_dataStartOffset` since, this is when the data starts. And if we make any other additions the semantics should not change.

libsolidity/CompilerUtils.h
LefterisJP: Ok but I think it's quite readable like this too. Plus it's kind of annoying to have static const int in a class if I remember well.

chriseth: It's not about readability, but about unnecessarily cluttering namespaces. As far as I know, static const POD is perfectly fine in C++11.

mix/QContractDefinition.cpp
```diff
@@ -33,11 +33,9 @@ using namespace dev::mix;
...
-	std::vector<FunctionDefinition const*> functions = _contract->getInterfaceFunctions();
-	for (unsigned i = 0; i < functions.size(); i++)
-	{
-		FunctionDefinition const* func = functions.at(i);
-		m_functions.append(new QFunctionDefinition(func, i));
-	}
+	auto interfaceFunctions = _contract->getInterfaceFunctions();
+	unsigned i = 0;
+	for (auto it = interfaceFunctions.cbegin(); it != interfaceFunctions.cend(); ++it, ++i)
+		m_functions.append(new QFunctionDefinition(it->second, i));
```
chriseth: Does this really work or do you want to leave the rest of the implementation to the mix team?

test/SolidityEndToEndTest.cpp
```diff
@@ -52,7 +52,7 @@ BOOST_AUTO_TEST_CASE(empty_contract)
 	char const* sourceCode = "contract test {\n"
 							 "}\n";
 	compileAndRun(sourceCode);
-	BOOST_CHECK(callContractFunction(0, bytes()).empty());
+	BOOST_CHECK(callContractFunction("iam_not_there()", bytes()).empty());
```
chriseth: Should be `i_am_not_there` :-)

LefterisJP: haahhah true

libsolidity/CompilerUtils.h
LefterisJP: you will see what I mean with my next commit. If there is a better way to do it inside the class tell me.

mix/QContractDefinition.cpp
LefterisJP: At the moment I made this change, I wanted it to compile. Did not really think of the Mix part of the thing. Will address it a subsequent PR, or leave it to the MIX team. Have to see how my time unfolds.

libsolidity/Types.cpp
LefterisJP: Hehe .. you really don't like how iterators look? I can change it in the places that looks readable. But I don't think it's bad as is.

PR #757, opened 2015-01-08
Ide m25 uidesign

yann300:
redesign of the debug panel. #648


mix/ConstantCompilationControl.cpp
```diff
@@ -34,15 +34,15 @@
 
 using namespace dev::mix;
 
-ConstantCompilationControl::ConstantCompilationControl(AppContext* _context): Extension(_context, ExtensionDisplayBehavior::Tab)
+ConstantCompilationControl::ConstantCompilationControl(AppContext* _context): Extension(_context, ExtensionDisplayBehavior::HeaderView)
 {
 	connect(_context->codeModel(), &CodeModel::compilationComplete, this, &ConstantCompilationControl::update);
-	connect(_context->codeModel(), &CodeModel::compilationComplete, this, &ConstantCompilationControl::update);
+	_context->appEngine()->rootContext()->setContextProperty("constantCompilation", this);
```
arkpar: I think now is a good time to rename it to something like StatusPane as it is going to display all kinds of information, not just compilation results.

mix/AssemblyDebuggerControl.cpp
```diff
@@ -198,11 +198,15 @@ void AssemblyDebuggerControl::executeSequence(std::vector<TransactionSettings> c
...
 {
+	Q_UNUSED(_reason);
 	m_appEngine->rootContext()->setContextProperty("debugStates", QVariant::fromValue(_wStates));
 	m_appEngine->rootContext()->setContextProperty("humanReadableExecutionCode", QVariant::fromValue(std::get<0>(_code)));
 	m_appEngine->rootContext()->setContextProperty("bytesCodeMapping", QVariant::fromValue(std::get<1>(_code)));
 	m_appEngine->rootContext()->setContextProperty("contractCallReturnParameters", QVariant::fromValue(new QVariableDefinitionList(_returnParam)));
-	this->addContentOn(this);
+	QVariant returnValue;
+	QObject* debugPanel = m_view->findChild<QObject*>("debugPanel", Qt::FindChildrenRecursively);
+	QMetaObject::invokeMethod(debugPanel, "init", Q_RETURN_ARG(QVariant, returnValue));
```
arkpar: Why not use Component.onCompleted in QML?

mix/ConstantCompilationControl.cpp
```diff
@@ -54,32 +54,21 @@ void ConstantCompilationControl::start() const
...
-		status->setProperty("color", "green");
-		content->setProperty("text", result->assemblyCode());
-	}
-	else
-	{
-		status->setProperty("text", "failure");
-		status->setProperty("color", "red");
-		content->setProperty("text", result->compilerMessage());
-	}
+void ConstantCompilationControl::update()
```
arkpar: Is this method still needed?

mix/ConstantCompilationControl.cpp
```diff
@@ -54,32 +54,21 @@ void ConstantCompilationControl::start() const
...
+{
+	QObject* ctrl = m_view->findChild<QObject*>("constantCompilationStatus", Qt::FindChildrenRecursively);
+	QMetaObject::invokeMethod(ctrl, "update");
 }
 
 void ConstantCompilationControl::resetOutPut()
```
arkpar: This can be probably removed too

mix/qml/DebugBasicInfo.qml
```diff
@@ -0,0 +1,109 @@
...
+import QtQuick.Layouts 1.0
+import QtQuick.Controls.Styles 1.1
+
+ColumnLayout {
+	property string currentStep
+	property string mem
+	property string stepCost
+	property string gasSpent
+	spacing: 0
+	RowLayout {
```
arkpar: Repeating elemets should be defined as components to avoid copy-pasting blocks of similair code

mix/qml/Debugger.qml
```diff
@@ -4,266 +4,329 @@ import QtQuick.Controls 1.1
...
+					spacing: 15
+					width: 250
+					height: parent.height
+
+					StepActionImage
+					{
+						id: jumpoutbackaction;
+						source: "qrc:/qml/img/jumpoutback.png"
+						disableStateImg: "qrc:/qml/img/jumpoutbackdisabled.png"
+						onClicked: Debugger.stepOutBack()
```
arkpar: Better change StepActionImage to be a Button with image and assign Action objects instead of onClicked. This way you'll also get keyboard shortcuts

mix/qml/DebugBasicInfo.qml
yann300: ​I have still some changes to make​, This pull request is still in "In
Progress". I think I will change to "Please Review" as soon as Alex will
approve the design implementation (And also as soon I have cleaned the
code).

2015-01-08 11:53 GMT-08:00 Arkadiy Paronyan <notifications@github.com>:

mix/qml/ImageButton.qml
```diff
@@ -0,0 +1,38 @@
+import QtQuick 2.2
+import QtQuick.Controls 1.1
+import QtQuick.Layouts 1.0
+import QtQuick.Controls.Styles 1.1
+
+Item {
```
arkpar: Is this used anywhere?

mix/AssemblyDebuggerControl.cpp
yann300: onCompleted is fired when the control is loaded on the view.
At this point, the control is already loaded. I will change the name of this function to "update"

mix/ConstantCompilationControl.cpp
yann300: this one is used

mix/ConstantCompilationControl.cpp
yann300: this one not anymore

libdevcore/CommonIO.h
```diff
@@ -48,6 +48,8 @@ void writeFile(std::string const& _file, bytes const& _data);
 
 /// Nicely renders the given bytes to a string, optionally as HTML.
 std::string memDump(bytes const& _b, unsigned _w = 8, bool _html = false);
+/// Nicely renders the given bytes to a string, store the content in an array.
```
chriseth: Please explain `_b` and `_w` and perhaps change their names.

libdevcore/CommonIO.cpp
```diff
@@ -26,6 +26,43 @@
...
+			ret << setfill('0') << setw(2) << hex << (unsigned)_b[j] << " ";
+		dumpLine.push_back(ret.str());
+		dump.push_back(dumpLine);
+	}
+	return dump;
+}
+
 string dev::memDump(bytes const& _b, unsigned _w, bool _html)
 {
 	stringstream ret;
```
chriseth: Is it possible to reuse the code above?

arkpar: Is it even needed anymore?

libdevcore/CommonIO.cpp
```diff
@@ -26,6 +26,43 @@
...
+		std::list<std::string> dumpLine;
+		ret << hex << setw(4) << setfill('0') << i << " ";
+		dumpLine.push_back(ret.str());
+		ret.str(std::string());
+		ret.clear();
+
+		for (unsigned j = i; j < i + _w; ++j)
+			if (j < _b.size())
+				if (_b[j] >= 32 && _b[j] < 127)
+					if ((char)_b[j] == '<')
```
arkpar: Better remove html escaping and set text control mode to plain

libdevcore/CommonIO.cpp
```diff
@@ -26,6 +26,43 @@
 using namespace std;
 using namespace dev;
 
+std::list<std::list<std::string>> dev::memDumpToList(bytes const& _b, unsigned _w)
```
arkpar: Using linked list for small objects is usually much more expensive than vector due to lots of allocations. Especially if you only push_back(). Also why not use QString here to avoid extra conversion later

mix/CodeEditorExtensionManager.h
```diff
@@ -54,17 +54,18 @@ class CodeEditorExtensionManager: public QObject
...
 private slots:
 	void applyCodeHighlight();
 
 private:
 	QVector<std::shared_ptr<Extension>> m_features;
-	QQuickItem* m_tabView;
-	QQuickItem* m_rightTabView;
+	QQuickItem* m_headerView;
+	QQuickItem* m_rightView;
+	QTextDocument* m_doc;
```
arkpar: m_doc is not needed anymore

mix/DebuggingStateWrapper.cpp
```diff
@@ -66,55 +66,102 @@ std::tuple<QList<QObject*>, QQMLMap*> DebuggingStateWrapper::getHumanReadableCod
...
+	filled.append(" ");
+	filled.append(" ");
+	return fillList(qVariantDump(dump), QVariant(filled));
+}
 
-	return QString::fromStdString(s.str());
+QVariantList DebuggingStateWrapper::qVariantDump(std::list<std::list<std::string>> const& _dump)
+{
+	QVariantList ret;
+	for (std::list<std::string> line: _dump)
```
arkpar: std::list<std::string> const& or auto const&, otherwise a list is copied

mix/DebuggingStateWrapper.cpp
```diff
@@ -66,55 +66,102 @@ std::tuple<QList<QObject*>, QQMLMap*> DebuggingStateWrapper::getHumanReadableCod
...
+	{
+		QStringList qLine;
+		for (std::string cell: line)
```
arkpar: std::string const&

mix/DebuggingStateWrapper.cpp
```diff
@@ -66,55 +66,102 @@ std::tuple<QList<QObject*>, QQMLMap*> DebuggingStateWrapper::getHumanReadableCod
...
+	}
+	return ret;
 }
 
-QString DebuggingStateWrapper::debugMemory()
+QStringList DebuggingStateWrapper::fillList(QStringList& _list, QString const& _emptyValue)
 {
-	return QString::fromStdString(memDump(m_state.memory, 16, false));
+	if (_list.size() < 20)
+	{
```
arkpar: Redundant brace

mix/DebuggingStateWrapper.cpp
```diff
@@ -136,6 +183,13 @@ QString DebuggingStateWrapper::headerInfo()
 	return QString::fromStdString(ss.str());
 }
 
+QString DebuggingStateWrapper::instruction()
+{
+	std::ostringstream ss;
+	ss << dev::eth::instructionInfo(m_state.inst).name;
```
arkpar: Looks like you can just use the name here, no need for a string stream

mix/qml/Debugger.qml
```diff
@@ -4,266 +4,505 @@ import QtQuick.Controls 1.1
...
+					height: parent.height
+					color: "transparent"
+					width: debugScrollArea.firstColumnWidth
+					RowLayout {
+						anchors.horizontalCenter: parent.horizontalCenter
+						id: jumpButtons
+						spacing: 3
+						StepActionImage
+						{
+							id: jumpoutbackaction;
```
arkpar: camelCase for ids

mix/qml.qrc
```diff
@@ -10,6 +10,28 @@
         <file>qml/ProjectList.qml</file>
         <file>qml/StateDialog.qml</file>
         <file>qml/StateList.qml</file>
+        <file>qml/js/main.js</file>
```
arkpar: I suggest using consistent style for file names. We already use PascalCase for cpp/h, qml and js files. alllowercase for images is ok I think, although looks ugly

mix/qml.qrc
```diff
@@ -12,5 +12,24 @@
         <file>qml/AlertMessageDialog.qml</file>
         <file>qml/StateDialog.qml</file>
         <file>qml/StateList.qml</file>
+        <file>qml/CompilationStatus.qml</file>
+        <file>qml/js/main.js</file>
```
arkpar: Is this needed?

libdevcore/CommonIO.cpp
chriseth: This is CommonIO.cpp, please don't use QString.

libdevcore/CommonIO.cpp
yann300: memDump is still used

libdevcore/CommonIO.cpp
arkpar: Oh, didn't realise that. In that case I'd suggest moving it to mix codebase as it is really too specific

libdevcore/CommonIO.cpp
yann300: I will move memDumpToList into mix project

gavofyork:
fails two builds?

PR #760, opened 2015-01-09
add blockhash test as vm tests

CJentzsch:
previously only done as state tests.

### DIFF
--- a/test/state.cpp
+++ b/test/state.cpp
@@ -39,16 +39,6 @@ using namespace dev::eth;
 
 namespace dev {  namespace test {
 
-LastHashes lastHashes(u256 _currentBlockNumber)
-{
-	LastHashes ret;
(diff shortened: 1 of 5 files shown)


test/state.cpp
```diff
@@ -39,16 +39,6 @@ using namespace dev::eth;
 
 namespace dev {  namespace test {
 
-LastHashes lastHashes(u256 _currentBlockNumber)
-{
-	LastHashes ret;
-	for (u256 i = 1; i <= 256 && i <= _currentBlockNumber; ++i)
-		ret.push_back(sha3(toString(_currentBlockNumber - i)));
```
CJentzsch: No. Since the data of the blockchain is not given, the opcode BLOCKHASH could not return the hashes of the corresponding blocks. Therefore we define the hash of block number n to be SHA256("n"). (see https://github.com/ethereum/tests/wiki/State-tests)

PR #762, opened 2015-01-09
fixed web3.sha3 method

debris:
asBytes should be used instead of jsToBytes
i will make appropriate issue, to fix naming of these functions.

### DIFF
--- a/libweb3jsonrpc/WebThreeStubServer.cpp
+++ b/libweb3jsonrpc/WebThreeStubServer.cpp
@@ -246,7 +246,7 @@ std::shared_ptr<dev::shh::Interface> WebThreeStubServer::face() const
 
 std::string WebThreeStubServer::web3_sha3(std::string const& _param1)
 {
-	return toJS(sha3(jsToBytes(_param1)));
+	return toJS(sha3(asBytes(_param1)));
 }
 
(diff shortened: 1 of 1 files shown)


subtly:
This needs a test :)

subtly:
Bug + crypto == mandatory test to prevent future regression. The test should reproduce original problem by reproducing original code and asserting failure, then in same unit, a test which asserts proper functionality.

subtly:
If this is a critical patch to be merged, then, we can create a github issue. (Not sure if we have a framework in place for testing JS; if we don't, that's superbad.)

debris:
i've just started writing js tests yesterday [here](https://github.com/ethereum/ethereum.js/tree/tests). it's using [mocha](https://github.com/mochajs/mocha) test framework. but it will take a while, until i cover most of the code.

and i think that this pr should go in earlier, cause without it, contract calls/transact from dapps are not working (same with #750)

gavofyork:
excellent example of the important of tests.

there is a fundamental misunderstanding about this interface. jsToBytes is correct. i seem to remember fixing this myself a while ago.

to be clear: asBytes should *never* be used. if the user wants a sha3 of some ascii, they should use web3.fromAscii first.

PR #766, opened 2015-01-09
web3.sha3 test

debris:
web3.sha3 test (not passing)

### DIFF
--- a/test/webthreestubclient.h
+++ b/test/webthreestubclient.h
@@ -12,6 +12,16 @@ class WebThreeStubClient : public jsonrpc::Client
     public:
         WebThreeStubClient(jsonrpc::IClientConnector &conn) : jsonrpc::Client(conn) {}
 
+        std::string web3_sha3(const std::string& param1) throw (jsonrpc::JsonRpcException)
+        {
+            Json::Value p;
+            p.append(param1);
+            Json::Value result = this->CallMethod("web3_sha3",p);
+            if (result.isString())
+                return result.asString();
+            else
+                throw jsonrpc::JsonRpcException(jsonrpc::Errors::ERROR_CLIENT_INVALID_RESPONSE, result.toStyledString());
+        }
         std::string eth_coinbase() throw (jsonrpc::JsonRpcException)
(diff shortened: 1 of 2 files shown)


test/webthreestubclient.h
```diff
@@ -12,6 +12,16 @@ class WebThreeStubClient : public jsonrpc::Client
     public:
         WebThreeStubClient(jsonrpc::IClientConnector &conn) : jsonrpc::Client(conn) {}
 
+        std::string web3_sha3(const std::string& param1) throw (jsonrpc::JsonRpcException)
```
CJentzsch: throw as exception specification is deprecated. See here for more infos: http://stackoverflow.com/questions/13841559/deprecated-throw-list-in-c11
I would just add a comment.

test/webthreestubclient.h
```diff
@@ -12,6 +12,16 @@ class WebThreeStubClient : public jsonrpc::Client
...
+        {
+            Json::Value p;
+            p.append(param1);
+            Json::Value result = this->CallMethod("web3_sha3",p);
+            if (result.isString())
+                return result.asString();
+            else
+                throw jsonrpc::JsonRpcException(jsonrpc::Errors::ERROR_CLIENT_INVALID_RESPONSE, result.toStyledString());
```
CJentzsch: Usually we use boost exceptions throughout the codebase. Make JsonRpcException derive from boost::exception and throw like that: BOOST_THROW_EXCEPTION(jsonrpc::JsonRpcException)

debris: that's autogenerated code :)

test/webthreestubclient.h
debris: that's autogenerated code :)

debris:
test is passing now, the problem was with the wrong padding in fromAscii i fixed that also in [47687cf085196e2e4fa7ec42ff426f241dae10c5](https://github.com/ethereum/ethereum.js/commit/47687cf085196e2e4fa7ec42ff426f241dae10c5)

PR #767, opened 2015-01-09
Padding for ABI types.

chriseth:


libsolidity/Compiler.cpp
```diff
@@ -181,10 +181,11 @@ void Compiler::appendReturnValuePacker(FunctionDefinition const& _function)
 								  << errinfo_sourceLocation(parameters[i]->getLocation())
 								  << errinfo_comment("Type " + paramType.toString() + " not yet supported."));
 		CompilerUtils(m_context).copyToStackTop(stackDepth, paramType);
+		ExpressionCompiler::appendTypeConversion(m_context, paramType, paramType, true);
 		bool const leftAligned = paramType.getCategory() == Type::Category::STRING;
-		CompilerUtils(m_context).storeInMemory(dataOffset, numBytes, leftAligned);
+		bool const padToWords = true;
```
gavofyork: `const` & `static const` variables should be prefixed c_ or (if part of an API) Capitalised.

PR #775, opened 2015-01-11
compiling ethereum.js is part of cmake build

debris:
if it is required and npm && node are found, compile ethereum.js, 

tested on ubuntu, osx and windows

first compilation may take some time, cause gulp and mocha needs to be downloaded


caktux:
Why is ethereum.js needed in cpp-ethereum and needs to be built?

debris:
ethereum.js is used by alethzero browser. and whenever we change something in js code, we want to see the effect of these changes. the better way of doing this would be running gulp watch, but i do not know if it is possible to integrate it with cmake.

PR #795, opened 2015-01-13
adding cpp-jit to random test simulation

CJentzsch:
I created a new executable "checkRandomTest" which takes the test as a commandline argument e.g. checkRandomTest '{..json formatted vm test content...}'. Unfortunately I had to copy, paste and modify some existing code from TestHelper.cpp. This was necessary because I can not (and will not) use boost_test macros here. Mainly because of performance (unnecessary overhead, parsing of output to detect success or failure, ...).


test/checkRandomTest.cpp
```diff
@@ -0,0 +1,279 @@
...
+        bytes output;
+        u256 gas;
+        bool vmExceptionOccured = false;
+        try
+        {
+            auto vm = eth::VMFactory::create(fev.gas);
+            output = vm->go(fev, fev.simpleTrace()).toBytes();
+            gas = vm->gas();
+        }
+        catch (eth::VMException const& _e)
```
chfast: Unsued variable `_e`. Remove the name.

PR #796, opened 2015-01-13
Random test efficiency

CJentzsch:
Fill up the stack at the beginning to reduce the stackunderflows and have more useful tests and include one invlaid opcode (0x0c).


test/createRandomTest.cpp
```diff
@@ -54,18 +54,26 @@ int main(int argc, char *argv[])
 	gen.seed(static_cast<unsigned int>(timeSinceEpoch));
 	boost::random::uniform_int_distribution<> lengthOfCodeDist(2, 16);
 	boost::random::uniform_int_distribution<> opcodeDist(0, 255);
+	boost::random::uniform_int_distribution<> BlockInfoOpcodeDist(0x40, 0x45);
```
chfast: Standard library has std::random already.

test/createRandomTest.cpp
```diff
@@ -106,7 +114,7 @@ int main(int argc, char *argv[])
 	read_string(s, v);
 
 	// insert new random code
-	v.get_obj().find("randomVMtest")->second.get_obj().find("pre")->second.get_obj().begin()->second.get_obj()["code"] = "0x" + randomCode;
+	v.get_obj().find("randomVMtest")->second.get_obj().find("pre")->second.get_obj().begin()->second.get_obj()["code"] = "0x" + randomCode + "55";
```
chfast: Does it add SSTORE at the end?

Maybe it's better to create 2 codes, one without the store, one with store. It's not very random if all the test have SSTORE at the end.

test/createRandomTest.cpp
```diff
@@ -54,18 +54,26 @@ int main(int argc, char *argv[])
...
+		{
+			randomCode += toHex(toCompactBigEndian((uint8_t)randGenBlockInfoOpcode()));
+			continue;
+		}
 
-		// disregard all invalid commands, except of one (0x10)
-		if (dev::eth::isValidInstruction(dev::eth::Instruction(opcode)) || opcode == 0x10)
+		uint8_t opcode = randGen();
+		// disregard all invalid commands, except of one (0x0c)
+		if ((dev::eth::isValidInstruction(dev::eth::Instruction(opcode)) || (opcode == 0x0c)))
```
chfast: Instead of fixing the value of invalid instruction it can be randomized also.

test/createRandomTest.cpp
CJentzsch: I added it, because most likely are differences on the stack, and we almost never check them unless we store something. But I can add both variants.

test/createRandomTest.cpp
CJentzsch: I can do that

test/createRandomTest.cpp
CJentzsch: they are not the same

PR #798, opened 2015-01-13
Natspec Popup Authentication on transaction

LefterisJP:
This is still a work in progress. I am opening a PR in order to start receiving feedback/comments.

As of the writting of this comment it's possible to open up Alethzero, create a contract with AZ and have its natspec be saved locally in a levelDB database. It's located under `~/.ethereum/natspec`

Whenever a transaction occurs via the JS API authentication happens by retrieving the Natspec with the transaction's function hash and querying the user for Allowing or Rejecting the transaction using the contents of the `@notice` tag as prompt. For an easy example look at: `ethereum.js/examples/natspec_contract.html`.

Remaining to be done:
  - Cleaning it all up
  - Allowing for evaluation of arguments and other stuff inside backtick quotes


alethzero/MainWin.cpp
```diff
@@ -1635,6 +1635,13 @@ static shh::Topic topicFromText(QString _s)
 	return ret;
 }
 
+
+bool Main::sourceIsSolidity(std::string const& _source)
+{
+	// TODO: Improve this heuristic
+	return (_source.substr(0, 8) == "contract" || _source.substr(0, 2) == "/*");
```
gavofyork: see latest way of checking this. hint: file begins with "//sol"

LefterisJP: Ops ... I actually deleted a place where this check was happening. Will add it back in this function then.

alethzero/MainWin.cpp
```diff
@@ -1872,7 +1880,31 @@ void Main::on_send_clicked()
...
+						for (std::string& s: compiler.getContractNames())
+						{
+							h256 contractHash = compiler.getContractCodeHash(s);
+							m_natspecDB.add(contractHash,
+											compiler.getMetadata(s, dev::solidity::DocumentationType::NATSPEC_USER));
+						}
+					}
+					catch (...)
+					{
+						statusBar()->showMessage("Couldn't make transaction: no single account contains at least the required amount.");
```
chriseth: This seems to be the wrong message.

alethzero/MainWin.cpp
```diff
@@ -1872,7 +1880,31 @@ void Main::on_send_clicked()
...
 				ethereum()->transact(s, value(), m_data, ui->gas->value(), gasPrice());
+				string src = ui->data->toPlainText().toStdString();
+				if (sourceIsSolidity(src))
+				{
+
+					dev::solidity::CompilerStack compiler;
+					try
+					{
+						m_data = compiler.compile(src, m_enableOptimizer);
+						for (std::string& s: compiler.getContractNames())
```
chriseth: You can omit "std::".

alethzero/MainWin.cpp
```diff
@@ -1872,7 +1880,31 @@ void Main::on_send_clicked()
 			debugFinished();
 			Secret s = i.secret();
 			if (isCreation())
+			{
+				// If execution is a contract creation, add Natspec to
+				// a local Natspec LEVELDB
 				ethereum()->transact(s, value(), m_data, ui->gas->value(), gasPrice());
+				string src = ui->data->toPlainText().toStdString();
```
chriseth: I think this code block should get its own function.

alethzero/MainWin.cpp
```diff
@@ -2257,6 +2289,16 @@ void Main::on_post_clicked()
 	whisper()->inject(m.seal(from, topicFromText(ui->shhTopic->toPlainText()), ui->shhTtl->value(), ui->shhWork->value()));
 }
 
+std::string Main::lookupNatSpec(dev::h256 const& _contractHash) const
```
chriseth: std:: can be omitted, also below

alethzero/NatspecHandler.cpp
```diff
@@ -0,0 +1,88 @@
...
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+/** @file NatspecHandler.h
```
chriseth: .cpp

alethzero/MainWin.cpp
LefterisJP: To be honest, this whole file should be Split and properly abstracted and refactored but well from the way I understand it, any time spent on AZ is gonna be a waste.

alethzero/MainWin.cpp
LefterisJP: I started doing that but then I saw that the whole file uses the `std::` namespace prefix

alethzero/NatspecHandler.cpp
```diff
@@ -0,0 +1,88 @@
...
+	boost::filesystem::create_directories(path);
+	ldb::Options o;
+	o.create_if_missing = true;
+	ldb::DB::Open(o, path + "/natspec", &m_db);
+}
+
+void NatspecHandler::add(dev::h256 const& _contractHash, std::string const& _doc)
+{
+	bytes k = _contractHash.asBytes();
+	std::string v = _doc;
```
chriseth: Why the local copy?

alethzero/NatspecHandler.cpp
```diff
@@ -0,0 +1,88 @@
...
+{
+	bytes k = _contractHash.asBytes();
+	std::string ret;
+	m_db->Get(m_readOptions, ldb::Slice((char const*)k.data(), k.size()), &ret);
+	return ret;
+}
+
+std::string NatspecHandler::getUserNotice(std::string const& json, const dev::bytes& _transactionData)
+{
+	Json::Value natspec, userNotice;
```
chriseth: Not allowed by coding style.

alethzero/NatspecHandler.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+std::string NatspecHandler::retrieve(dev::h256 const& _contractHash) const
+{
+	bytes k = _contractHash.asBytes();
+	std::string ret;
+	m_db->Get(m_readOptions, ldb::Slice((char const*)k.data(), k.size()), &ret);
+	return ret;
+}
+
+std::string NatspecHandler::getUserNotice(std::string const& json, const dev::bytes& _transactionData)
```
chriseth: Move "const"

alethzero/NatspecHandler.h
```diff
@@ -0,0 +1,57 @@
...
+	/// user notice for that transaction.
+	/// @returns The user notice or an empty string if no natspec for the contract exists
+	///          or if the existing natspec does not document the @c _methodName
+	std::string getUserNotice(dev::h256 const& _contractHash, dev::bytes const& _transactionDacta);
+	
+  private:
+	ldb::ReadOptions m_readOptions;
+	ldb::WriteOptions m_writeOptions;
+	ldb::DB* m_db;
+	Json::Reader m_reader;
```
chriseth: Is it necessary to have this as member?

alethzero/NatspecHandler.cpp
```diff
@@ -0,0 +1,88 @@
...
+{
+	Json::Value natspec, userNotice;
+	std::string retStr;
+	m_reader.parse(json, natspec);
+	bytes transactionFunctionPart(_transactionData.begin(), _transactionData.begin() + 4);
+	FixedHash<4> transactionFunctionHash(transactionFunctionPart);
+
+	Json::Value methods = natspec["methods"];
+	for (Json::ValueIterator it= methods.begin(); it != methods.end(); ++it)
```
chriseth: foreach?

alethzero/NatspecHandler.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+std::string NatspecHandler::getUserNotice(std::string const& json, const dev::bytes& _transactionData)
+{
+	Json::Value natspec, userNotice;
+	std::string retStr;
+	m_reader.parse(json, natspec);
+	bytes transactionFunctionPart(_transactionData.begin(), _transactionData.begin() + 4);
+	FixedHash<4> transactionFunctionHash(transactionFunctionPart);
+
+	Json::Value methods = natspec["methods"];
```
chriseth: What happens if the key is not present? What happens if the key in line 71 is not a string?

alethzero/NatspecHandler.cpp
```diff
@@ -0,0 +1,88 @@
...
+	for (Json::ValueIterator it= methods.begin(); it != methods.end(); ++it)
+	{
+		std::string functionSig = it.key().asString();
+		FixedHash<4> functionHash(dev::sha3(functionSig));
+
+		if (functionHash == transactionFunctionHash) {
```
chriseth: Curlies.

alethzero/NatspecHandler.h
```diff
@@ -0,0 +1,57 @@
...
+	/// Given a contract code hash and the transaction's data retrieve the natspec documention's
+	/// user notice for that transaction.
+	/// @returns The user notice or an empty string if no natspec for the contract exists
+	///          or if the existing natspec does not document the @c _methodName
+	std::string getUserNotice(dev::h256 const& _contractHash, dev::bytes const& _transactionDacta);
+	
+  private:
+	ldb::ReadOptions m_readOptions;
+	ldb::WriteOptions m_writeOptions;
+	ldb::DB* m_db;
```
chriseth: Do we need a destructor for this?

alethzero/OurWebThreeStubServer.cpp
```diff
@@ -39,36 +42,32 @@ std::string OurWebThreeStubServer::shh_newIdentity()
...
-bool OurWebThreeStubServer::authenticate(dev::TransactionSkeleton const& _t) const
+bool OurWebThreeStubServer::showAuthenticationPopup(std::string const& _title, std::string const& _text) const
 {
-	return true;
-
-	// To get the balance of the sender
-	cnote << "Sender has ETH: " << m_web3->ethereum()->postState().balance(_t.from);
-
-	Main* main; // don't know this yet, should be a member and set at construction time by Main, who will construct us.
+	// otherwise it's a transaction to a contract for which we have the natspec
```
chriseth: Why "otherwise"?

libethereum/Defaults.h
```diff
@@ -23,6 +23,7 @@
 
 #include <libdevcore/Common.h>
 
+
```
chriseth: No need to touch this file :-)

libsolidity/CompilerStack.cpp
```diff
@@ -134,6 +137,16 @@ bytes const& CompilerStack::getBytecode(string const& _contractName) const
 	return getContract(_contractName).bytecode;
 }
 
+bytes const& CompilerStack::getRuntimeBytecode(std::string const& _contractName) const
```
chriseth: std:: not needed (also below)

libsolidity/CompilerStack.h
```diff
@@ -75,7 +76,13 @@ class CompilerStack: boost::noncopyable
 	/// @returns the compiled bytecode
 	bytes const& compile(std::string const& _sourceCode, bool _optimize = false);
 
+	/// Gets the assembled bytecode for a contract
```
chriseth: @returns?

libsolidity/CompilerStack.h
```diff
@@ -75,7 +76,13 @@ class CompilerStack: boost::noncopyable
...
 	bytes const& getBytecode(std::string const& _contractName = "") const;
+	/// Get the runtime context's bytecode for a contract
```
chriseth: What about "@returns the runtime bytecode for the contract, i.e. the code that is returned by the constructor."?

libsolidity/CompilerStack.cpp
```diff
@@ -179,6 +192,24 @@ string const& CompilerStack::getMetadata(string const& _contractName, Documentat
 	return *(*doc);
 }
 
+std::string const CompilerStack::getFunctionHashes(std::string const& _contractName)
```
chriseth: This looks like pretty-printing for the user interface to me and I think it should not belong into this class.

alethzero/NatspecHandler.cpp
LefterisJP: Same reason they have it in WebThreeStubServer. For readability of the following calls where you take the `data()` and `size()` of it. Any decent compiler will optimize this away.

alethzero/NatspecHandler.cpp
LefterisJP: And for a good reason! Thanks for spotting it.

alethzero/NatspecHandler.h
LefterisJP: Not at the moment. But I thought that perhaps the Natspec handler may need to use Json parsing a bit more since we may want to retrieve more values from the natspec apart from the `@notice`.

alethzero/NatspecHandler.cpp
LefterisJP: I had originally tried a for_each but has problems getting the `key()` of the Json entry through the library's Json::Value iterator.

alethzero/NatspecHandler.cpp
LefterisJP: Since we are writting the json natspec formnat we can be sure that it is valid and will not contain anything than a string. But you are right that checks should be in place anyway. Especially since we may/will(?) take the json from elsewhere in the future.

alethzero/NatspecHandler.h
LefterisJP: yes we do! I did not notice that since the class I tried to learn what `levelDB` is from actually has it inside a shared pointer.

libsolidity/CompilerStack.h
LefterisJP: Yeah it's a good clarification

libsolidity/CompilerStack.cpp
LefterisJP: Agreed. It was the easiest place to put it in the beginning for testing but this will go somewhere else.

alethzero/NatspecHandler.h
chriseth: But what is the benefit of having it as member? Does it keep so much state between different parsing runs?

alethzero/NatspecHandler.cpp
chriseth: OK, then at least add the missing space :-)

alethzero/NatspecHandler.cpp
chriseth: Yes and also someone may take this code to somewhere else :-)

alethzero/MainWin.cpp
gavofyork: The whole codebase doesn't. This file shouldn't be any different. New code should be in line with Coding Standards. Old code should be fixed in a separate PR.

LefterisJP: I can do a new PR that fixes the whole file to not use the prefix. If I only correct these 2 additions the file will look very strange having incompatible styling until the rest is fixed.

alethzero/MainWin.cpp
```diff
@@ -1872,7 +1896,31 @@ void Main::on_send_clicked()
...
 			Secret s = i.secret();
 			if (isCreation())
+			{
+				// If execution is a contract creation, add Natspec to
+				// a local Natspec LEVELDB
 				ethereum()->transact(s, value(), m_data, ui->gas->value(), gasPrice());
+				string src = ui->data->toPlainText().toStdString();
+				if (sourceIsSolidity(src))
+				{
+
```
gavofyork: blank line

alethzero/MainWin.cpp
```diff
@@ -1872,7 +1896,31 @@ void Main::on_send_clicked()
...
+					dev::solidity::CompilerStack compiler;
```
gavofyork: move into next block and kill enclosing braces.

alethzero/NatspecHandler.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <boost/filesystem.hpp>
+#include <string>
+
+#include <libethereum/Defaults.h>
+#include <libdevcore/Common.h>
+#include <libdevcore/CommonData.h>
+#include <libdevcore/Exceptions.h>
+#include <libdevcore/Log.h>
+#include <libdevcrypto/SHA3.h>
+
```
gavofyork: blank line

alethzero/NatspecHandler.cpp
```diff
@@ -0,0 +1,104 @@
...
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+/** @file NatspecHandler.cpp
+ * @author Lefteris Karapetsas <lefteris@ethdev.com>
+ * @date 2015
+ */
+#include "NatspecHandler.h"
+#include <boost/filesystem.hpp>
+#include <string>
```
gavofyork: stl header before boost

alethzero/NatspecHandler.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+#include <libethereum/Defaults.h>
```
gavofyork: libethereum is higher level than others, it goes below

alethzero/NatspecHandler.cpp
```diff
@@ -0,0 +1,104 @@
...
+		{
+			Json::Value val = (*it)["notice"];
+			if (!val.isString())
+				BOOST_THROW_EXCEPTION(Exception() << errinfo_comment("Illegal Natspec JSON detected"));
+			return val.asString();
+		}
+	}
+
+	// not found
+	return "";
```
gavofyork: `return string()` is better practice.

PR #799, opened 2015-01-13
Specify gas and value for function calls and contract creation calls.

chriseth:


libsolidity/Types.cpp
```diff
@@ -561,6 +561,21 @@ FunctionType::FunctionType(FunctionDefinition const& _function, bool _isInternal
 	m_location = _isInternal ? Location::INTERNAL : Location::EXTERNAL;
 }
 
+FunctionType::FunctionType(TypePointers const& _parameterTypes, TypePointers const& _returnParameterTypes,
+						   FunctionType::Location _location, bool _gasSet, bool _valueSet):
+	m_parameterTypes(_parameterTypes), m_returnParameterTypes(_returnParameterTypes),
+	m_location(_location), m_gasSet(_gasSet), m_valueSet(_valueSet)
+{
+	if (m_location == Location::EXTERNAL)
```
arkpar: reuse getSizeOnStack() here?

chriseth: This is weird, seems I stopped in the middle of refactoring. Thanks for spotting!

libsolidity/ExpressionCompiler.cpp
```diff
@@ -232,27 +232,71 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+			m_context << u256(length) << u256(0);
+			if (function.valueSet())
+				m_context << eth::dupInstruction(3);
+			else
+				m_context << u256(0);
+			m_context << eth::Instruction::CREATE;
+			if (function.valueSet())
+				m_context << eth::swapInstruction(1) << eth::Instruction::POP;
+			return false;
 			break;
```
CJentzsch: This line can never be reached, you could as well omit it. (100% code coverage not possible :-) )

libsolidity/ExpressionCompiler.cpp
```diff
@@ -646,46 +660,52 @@ void ExpressionCompiler::appendHighBitsCleanup(IntegerType const& _typeOnStack)
...
 	// CALL arguments: outSize, outOff, inSize, inOff, value, addr, gas (stack top)
 	m_context << u256(retSize) << u256(0) << u256(dataOffset) << u256(0);
-	if (_options.obtainValue)
-		_options.obtainValue();
+	if (_functionType.valueSet())
+		m_context << eth::dupInstruction(5);
 	else
 		m_context << u256(0);
-	m_context << eth::dupInstruction(6); //copy contract address
+	m_context << eth::dupInstruction(6 + gasValueSize + (bare ? 0 : 1)); //copy contract address
```
CJentzsch: Can you explain why you have those hradcoded numbers here (6, or 5 a couple of lines above)?

libsolidity/Types.cpp
```diff
@@ -628,6 +674,13 @@ TypePointers FunctionType::parseElementaryTypeVector(strings const& _types)
 	return pointers;
 }
 
+TypePointer FunctionType::copyAndSetGasOrValue(bool _setGas, bool _setValue) const
+{
+	return make_shared<FunctionType>(m_parameterTypes, m_returnParameterTypes, m_location,
+									 m_gasSet || _setGas, m_valueSet || _setValue);
+}
+
+
```
CJentzsch: You usually have only one empty line between function declarations.

libsolidity/ExpressionCompiler.cpp
chriseth: Right, this is not nice. The basically count the number of stack slots that we added in the meantime. There are tools to avoid this, I will try.

Issue #809, opened 2015-01-15
Crash after starting mix


debris:
In latest develop 7583637c395ea2976493516f53e738fb1b29def0 mix is crashing just after it starts.
Tested on mac.

bt:
```
(lldb) bt
* thread #7: tid = 0x3f86d, 0x00000001096ed62c QtQml`___lldb_unnamed_function1606$$QtQml + 268, name = 'QQmlThread', stop reason = EXC_BAD_ACCESS (code=1, address=0x50)
  * frame #0: 0x00000001096ed62c QtQml`___lldb_unnamed_function1606$$QtQml + 268
    frame #1: 0x00000001096d49d5 QtQml`___lldb_unnamed_function1408$$QtQml + 725
    frame #2: 0x00000001096d378e QtQml`___lldb_unnamed_function1405$$QtQml + 366
    frame #3: 0x00000001096d0c86 QtQml`___lldb_unnamed_function1390$$QtQml + 326
    frame #4: 0x00000001096bec55 QtQml`QV4::IR::Optimizer::run(QQmlEnginePrivate*) + 26229
    frame #5: 0x0000000109725bb9 QtQml`QV4::JIT::InstructionSelection::run(int) + 153
    frame #6: 0x00000001096af31b QtQml`QV4::EvalInstructionSelection::compile(bool) + 43
    frame #7: 0x00000001096f0641 QtQml`___lldb_unnamed_function1632$$QtQml + 4417
    frame #8: 0x00000001098389b7 QtQml`___lldb_unnamed_function4320$$QtQml + 279
    frame #9: 0x0000000109838855 QtQml`___lldb_unnamed_function4319$$QtQml + 3029
    frame #10: 0x000000010982f4fe QtQml`QQmlDataBlob::tryDone() + 94
    frame #11: 0x000000010983188d QtQml`___lldb_unnamed_function4262$$QtQml + 157
    frame #12: 0x00000001098308fb QtQml`___lldb_unnamed_function4245$$QtQml + 1419
    frame #13: 0x00000001098310ff QtQml`___lldb_unnamed_function4255$$QtQml + 63
    frame #14: 0x0000000109835777 QtQml`___lldb_unnamed_function4289$$QtQml + 503
    frame #15: 0x000000010983a92d QtQml`___lldb_unnamed_function4327$$QtQml + 2637
    frame #16: 0x0000000109839e4f QtQml`___lldb_unnamed_function4326$$QtQml + 943
[...]

PR #814, opened 2015-01-15
mix: Changed ProjectModel to be a context property instead of a singleton to work around #809

arkpar:


mix/AppContext.cpp
```diff
@@ -48,7 +49,12 @@ AppContext::AppContext(QQmlApplicationEngine* _engine)
 	m_fileIo.reset(new FileIo());
 	m_applicationEngine->rootContext()->setContextProperty("appContext", this);
 	qmlRegisterType<FileIo>("org.ethereum.qml", 1, 0, "FileIo");
-	qmlRegisterSingletonType(QUrl("qrc:/qml/ProjectModel.qml"), "org.ethereum.qml.ProjectModel", 1, 0, "ProjectModel");
+	QQmlComponent projectModelComponent(m_applicationEngine, QUrl("qrc:/qml/ProjectModel.qml"));
+	QObject* projectModel = projectModelComponent.create();
+	if (projectModelComponent.isError())
+		throw std::runtime_error("Error loading ProjectModel: " + projectModelComponent.errorString().toStdString());
```
subtly: Can boost exception be used?

mix/AppContext.cpp
```diff
@@ -22,16 +22,17 @@
  * - KeyEventManager
  */
 
+#include <stdexcept>
```
subtly: Boost exceptions are used throughout other projects.

Issue #612, opened 2014-12-14
mix: Ethereum QML types


arkpar:
QML components for displaying and editing common ethereum types, such as u256, address

arkpar:
Moved to PT

PR #819, opened 2015-01-18
Ethereum QML types: #612

yann300:
- Display an Ether value (including units and formatting).
- Manage big integer in QML side.

### DIFF
--- a/mix/ClientModel.h
+++ b/mix/ClientModel.h
@@ -26,7 +26,6 @@
 #include <atomic>
 #include "DebuggingStateWrapper.h"
 #include "MixClient.h"
-#include "QEther.h"
 
 using AssemblyDebuggerData = std::tuple<QList<QObject*>, dev::mix::QQMLMap*>;
 
--- a/mix/ClientModel.h
+++ b/mix/ClientModel.h
@@ -1,18 +1,18 @@
 /*
-    This file is part of cpp-ethereum.
+	This file is part of cpp-ethereum.
 
-    cpp-ethereum is free software: you can redistribute it and/or modify
-    it under the terms of the GNU General Public License as published by
-    the Free Software Foundation, either version 3 of the License, or
-    (at your option) any later version.
+	cpp-ethereum is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
 
-    cpp-ethereum is distributed in the hope that it will be useful,
-    but WITHOUT ANY WARRANTY; without even the implied warranty of
-    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-    GNU General Public License for more details.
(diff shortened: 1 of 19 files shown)


mix/ClientModel.h
```diff
@@ -26,6 +26,7 @@
 #include <atomic>
 #include "DebuggingStateWrapper.h"
 #include "MixClient.h"
+#include "QEther.h"
```
arkpar: Is it needed here? Maybe a forward declaration will do?

mix/DebuggingStateWrapper.cpp
```diff
@@ -66,25 +67,19 @@ std::tuple<QList<QObject*>, QQMLMap*> DebuggingStateWrapper::getHumanReadableCod
 	return std::make_tuple(codeStr, QPointer<QQMLMap>(new QQMLMap(codeMapping)));
 }
 
-QString DebuggingStateWrapper::gasCost()
+QBigInt* DebuggingStateWrapper::gasCost()
 {
-	std::ostringstream ss;
-	ss << std::dec << m_state.gasCost;
-	return QString::fromStdString(ss.str());
+	return new QBigInt(m_state.gasCost, this);
```
arkpar: You should set ownership to javascript to avoid memory leaks

mix/QBigInt.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/** @file QBigInt.cpp
+ * @author Yann yann@ethdev.com
+ * @date 2015
+ */
+
+#include "boost/variant/multivisitors.hpp"
```
arkpar: angle brackets?

mix/ClientModel.h
yann300: OK I moved this include in the right header file.

mix/QBigInt.h
```diff
@@ -0,0 +1,97 @@
...
+	Q_INVOKABLE QBigInt* subtract(QBigInt* const& _value) const;
+	/// Add @a _value to the current big integer. Invokable from QML.
+	Q_INVOKABLE QBigInt* add(QBigInt* const& _value) const;
+	/// Multiply by @a _value. Invokable from QML.
+	Q_INVOKABLE QBigInt* multiply(QBigInt* const& _value) const;
+	/// divide by @a _value. Invokable from QML.
+	Q_INVOKABLE QBigInt* divide(QBigInt* const& _value) const;
+
+protected:
+	 BigIntVariant m_internalValue;
```
arkpar: Using variant here makes code quite complex. Maybe use template base class instead?

yann300: Yes I tried but we cannot use Template with a QObject class. And with
Template, we have to know the type to use (bigint or u256) at compile time.
With boost::variant. We just have to set a pool of type to manage and the
rest is done during the runtime.

2015-01-19 16:48 GMT+01:00 Arkadiy Paronyan <notifications@github.com>:

PR #824, opened 2015-01-19
Sol blockhash tests

LefterisJP:
A small PR with test for the blockhash command in Solidity.

Also some minor improvements to `SolidityExecutionFramework.h`

### DIFF
--- a/test/solidityExecutionFramework.h
+++ b/test/solidityExecutionFramework.h
@@ -29,7 +29,6 @@
 #include <libethereum/State.h>
 #include <libethereum/Executive.h>
 #include <libsolidity/CompilerStack.h>
-#include "TestHelper.h"
 
 namespace dev
 {
@@ -129,12 +128,6 @@ public:
 		return bytes();
 	}
 
-	eth::LastHashes setCurrentBlockNumber(u256 _currentBlockNumber)
-	{
-		m_lastHashes = dev::test::lastHashes(_currentBlockNumber);
-		return m_lastHashes;
-	}
-
 private:
 	template <class CppFunction, class... Args>
 	auto callCppAndEncodeResult(CppFunction const& _cppFunction, Args const&... _arguments)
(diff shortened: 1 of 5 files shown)


test/solidityExecutionFramework.h
```diff
@@ -46,7 +46,20 @@ class ExecutionFramework
...
-		compiler.compile(_sourceCode, m_optimize);
+		try
+		{
+			compiler.compile(_sourceCode, m_optimize);
+		}
+		catch (const std::exception& e)
+		{
+			std::string const* extra = boost::get_error_info<errinfo_comment>(e);
+			std::string msg = std::string("Parsing contract failed with: ") +
+				e.what() + std::string("\n");
```
CJentzsch: you could just do "msg =std::string("Parsing contract failed with: ") + diagnostic_information(_e);" Then you get all possible information.

test/solidityExecutionFramework.h
```diff
@@ -46,7 +46,20 @@ class ExecutionFramework
 	bytes const& compileAndRun(std::string const& _sourceCode, u256 const& _value = 0, std::string const& _contractName = "")
 	{
 		dev::solidity::CompilerStack compiler;
-		compiler.compile(_sourceCode, m_optimize);
+		try
+		{
+			compiler.compile(_sourceCode, m_optimize);
+		}
+		catch (const std::exception& e)
```
CJentzsch: _e

test/solidityExecutionFramework.h
LefterisJP: very good point. Will adjust here and in 2 other places in tests that I can remember.

test/solidityExecutionFramework.h
LefterisJP: did not see that in the coding standard. But by checking around the code we seem to always put underscore in the exception in catch so I will comply. Perhaps we should add a line in the coding standard? A quick look did not show anything.

LefterisJP:
Ok will do, but this is working too since boost::exceptions itself also derives from std::exception. But yes it's better to let those exceptions propagate upwards.

PR #825, opened 2015-01-19
Contract inheritance

chriseth:


libsolidity/Compiler.cpp
```diff
@@ -34,48 +35,84 @@ using namespace std;
...
+	neededFunctions = getFunctionsCalled(nodesUsedInConstructors);
 
 	for (FunctionDefinition const* fun: neededFunctions)
 		m_context.addFunction(*fun);
 
-	if (constructor)
-		appendConstructorCall(*constructor);
+	// Call constructors in base-to-derived order.
+	// The Constructor for the most derived contract is called later.
+	for (unsigned i = 1; i < bases.size(); i++)
```
LefterisJP: maybe `++i`?

libsolidity/Compiler.h
```diff
@@ -32,23 +32,24 @@ class Compiler: private ASTConstVisitor
...
 	bytes getRuntimeBytecode() { return m_runtimeContext.getAssembledBytecode(m_optimize);}
 	void streamAssembly(std::ostream& _stream) const { m_context.streamAssembly(_stream); }
 
 private:
-	/// Registers the global objects and the non-function objects inside the contract with the context.
-	void initializeContext(ContractDefinition const& _contract, std::vector<MagicVariableDeclaration const*> const& _magicGlobals,
+	/// Registers the non-function objects inside the contract with the context.
+	void initializeContext(ContractDefinition const& _contract,
 						   std::map<ContractDefinition const*, bytes const*> const& _contracts);
 	/// Adds the code that is run at creation time. Should be run after exchanging the run-time context
```
LefterisJP: Should be ran

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -369,6 +368,128 @@ BOOST_AUTO_TEST_CASE(hash_collision_in_interface)
...
+}
+
+BOOST_AUTO_TEST_CASE(constructor_visibility)
+{
+	// The constructor of a base class should not be visible in the derived class
+	char const* text = R"(
+		contract A { function A() { } }
+		contract B is A { function f() { A x = A(0); } }
+	)";
+	BOOST_CHECK_NO_THROW(parseTextAndResolveNames(text));
```
LefterisJP: If a constructor of a base class should not be visible in the derived one, then should this test not be using
`BOOST_CHECK_THROW(...)` ?

libsolidity/Compiler.cpp
```diff
@@ -34,48 +35,84 @@ using namespace std;
...
+			if (baseArguments.count(baseContract) == 0)
+			{
+				baseArguments[baseContract] = &base->getArguments();
+				for (ASTPointer<Expression> const& arg: base->getArguments())
+					nodesUsedInConstructors.insert(arg.get());
+			}
+		}
+	}
+
+	//@TODO add virtual functions
```
LefterisJP: From the EndToEndTests there seem to be virtual functions. Is this TODO still relevant?

test/SolidityNameAndTypeResolution.cpp
chriseth: The constructor should not be visible, but since its name is identical to the name of its contract, the name should signify the contract (type), and that is what happens here.

libsolidity/Compiler.cpp
chriseth: Yes, I will create a new PR for this.

PR #830, opened 2015-01-20
Include virtual function overrides in constructor context.

chriseth:


libsolidity/CallGraph.cpp
```diff
@@ -55,7 +55,12 @@ bool CallGraph::visit(Identifier const& _identifier)
 {
 	FunctionDefinition const* fun = dynamic_cast<FunctionDefinition const*>(_identifier.getReferencedDeclaration());
 	if (fun)
+	{
+		if (m_overrideResolver)
+			fun = (*m_overrideResolver)(fun->getName());
+		solAssert(fun, "");
```
LefterisJP: Shouldn't we have a message in solAssert here?

libsolidity/Parser.cpp
```diff
@@ -210,7 +210,8 @@ ASTPointer<FunctionDefinition> Parser::parseFunctionDefinition(bool _isPublic)
 	}
 	ASTPointer<Block> block = parseBlock();
 	nodeFactory.setEndPositionFromNode(block);
-	return nodeFactory.createNode<FunctionDefinition>(name, _isPublic, docstring,
+	bool const c_isConstructor = (_contractName && *name == *_contractName);
```
LefterisJP: why constant bool here?

libsolidity/AST.h
```diff
@@ -280,13 +280,13 @@ class FunctionDefinition: public Declaration
 {
 public:
 	FunctionDefinition(Location const& _location, ASTPointer<ASTString> const& _name,
-					bool _isPublic,
+					bool _isPublic, bool _isConstructor,
```
LefterisJP: Just to be consistent with the rest of the function, you could put `bool _isConstructor` to its own line to have all arguments on their own line

libsolidity/Parser.cpp
chriseth: Why not?

LefterisJP: I think it's an overkill

PR #833, opened 2015-01-21
design changes on debugging feature.

yann300:
- Right side pane resizing (list fill empty space).
 - Capitalisation.
 - Storage/MemoryDump/CallData add splitter.
 - Small changes on step icons.
 - App resizing: dock the right view on the main window.
 - Assembly code selection change (selection change move faster).
 - Reverse stack list.
 - Misc changes on debug info list.
 - Misc changes on debug button behavior + F10/F9 bug fixes.

### DIFF
--- a/mix/DebuggingStateWrapper.cpp
+++ b/mix/DebuggingStateWrapper.cpp
@@ -86,10 +86,8 @@ QBigInt* DebuggingStateWrapper::newMemSize()
 QStringList DebuggingStateWrapper::debugStack()
 {
 	QStringList stack;
-	for (auto i: m_state.stack)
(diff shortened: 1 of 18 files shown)


mix/DebuggingStateWrapper.cpp
```diff
@@ -89,6 +89,7 @@ QStringList DebuggingStateWrapper::debugStack()
 	for (auto i: m_state.stack)
 		stack.append(QString::fromStdString(prettyU256(i)));
 
+	std::reverse(stack.begin(), stack.end());
```
arkpar: It is be more effecient to just walk the stack backwards using rcbegin/rcend in the for loop above

PR #834, opened 2015-01-21
mix: Web preview and ethereum.js integration

arkpar:
Web preview support will be available only if qt web engine is detected at compile time


mix/CodeModel.cpp
```diff
@@ -156,14 +161,19 @@ void CodeModel::runCompilationJob(int _jobId, QString const& _code)
 	emit compilationCompleteInternal(result.release());
 }
 
-void CodeModel::onCompilationComplete(CompilationResult*_newResult)
+void CodeModel::onCompilationComplete(CompilationResult* _newResult)
```
yann300: Is CompilationResult* is instanciated using a QObject parent? if not maybe use a std:unique_ptr ...

arkpar: unique_ptr can't be used as an argument for signal/slot because of a missing copy constructor. In this case I release a pointer from a unique_ptr before emitting it as a signal and immediately capture to unique_ptr upon entering a slot in a new thread. There will be no leak unless there is a serious internal bug in QT signals/slots processing

mix/CodeModel.cpp
```diff
@@ -55,9 +56,12 @@ CompilationResult::CompilationResult(const dev::solidity::CompilerStack& _compil
 {
 	if (!_compiler.getContractNames().empty())
 	{
-		m_contract.reset(new QContractDefinition(&_compiler.getContractDefinition(std::string())));
+		auto const& contractDefinition = _compiler.getContractDefinition(std::string());
+		m_contract.reset(new QContractDefinition(&contractDefinition));
 		m_bytes = _compiler.getBytecode();
 		m_assemblyCode = QString::fromStdString(dev::eth::disassemble(m_bytes));
+		dev::solidity::InterfaceHandler interfaceHandler;
+		m_contractDefinition = QString::fromStdString(*interfaceHandler.getABIInterface(contractDefinition));
```
chriseth: Why not call it `contractInterface`? In my interpretation, "definition" also includes e.g. private functions.

Issue #820, opened 2015-01-19
mix: manage parameters in contract constructor


yann300:

arkpar:
Moved to PT

PR #837, opened 2015-01-22
Manage parameters in constructor #820

yann300:
Here is the behavior:

If the constructor does not contains parameters:
- Press F5 and it start debugging directly (using default values).
- Create a new state and the contract creation transaction is automatically created.

If the constructor contains parameters:
 - Press F5, a transaction modal dialog appear. This is a light version of the transaction modal dialog, it only contains the list of parameters (balance, value, gas, gasPrice are hidden => use of default values).
 - Create a new state and the contract creation transaction is automatically created (users can change gas, gasPrice, etc...).

Comments are welcome!


mix/qml/TransactionDialog.qml
```diff
@@ -67,112 +87,159 @@ Window {
...
 		}
+		else
+		{
+			item = TransactionHelper.defaultTransaction();
+			item.functionId = transactionDialog.functionId;
+			item.executeConstructor = isConstructorTransaction;
+		}
+
+		if (isConstructorTransaction)
+			item.functionId = qsTr("Constructor");
```
arkpar: This should include some special symbol, to avoid potential conflict with function name

arkpar: Ok, I see it is not used to id the constructor

mix/ClientModel.cpp
```diff
@@ -215,14 +217,20 @@ void ClientModel::showDebugError(QString const& _error)
 	m_context->displayMessageDialog(tr("Debugger"), _error);
 }
 
-ExecutionResult ClientModel::deployContract(bytes const& _code)
+ExecutionResult ClientModel::deployContract(bytes const& _code, TransactionSettings const& ctrTransaction)
```
gavofyork: `_ctrTransaction` (note underscore)

mix/ClientModel.h
```diff
@@ -116,8 +121,8 @@ private slots:
 
 private:
 	QString contractAddress() const;
-	void executeSequence(std::vector<TransactionSettings> const& _sequence, u256 _balance);
-	ExecutionResult deployContract(bytes const& _code);
+	void executeSequence(std::vector<TransactionSettings> const& _sequence, u256 _balance, TransactionSettings const& ctrTransaction = TransactionSettings());
```
gavofyork: don't kill the underscore!

yann300:
When the user press F5 and if a contract constructor contains parameters, the user is asked to fill the parameters. When a new state is created, the transaction deploying the contract is added in the list and the user can fill in the parameters.

PR #840, opened 2015-01-23
LLVM lshr bug workaround

chfast:


CJentzsch:
I checked it, and the tests tun through now.

Issue #609, opened 2014-12-14
mix: Code editor


arkpar:
Line numbers
Syntax highlighting
Bracket matching
Auto-completion
Error and warning annotations

arkpar:
Auto-completion and Error and warning annotations filed as separate issues

PR #845, opened 2015-01-23
mix: CodeMirror code editor #609

arkpar:


mix/qml/WebPreview.qml
```diff
@@ -27,7 +27,10 @@ Item {
 	}
 
 	function updateContract() {
-		webView.runJavaScript("updateContract(\"" + clientModel.contractAddress + "\", " + codeModel.code.contractInterface + ")");
+		var contractInterface = codeModel.code.contractInterface;
```
yann300: contractInterface cannot directly return  an empty array?

PR #855, opened 2015-01-23
fix gasPrice*gasLimit + value overflow

CJentzsch:
extremely important! Without the fix, the testcase "OverflowGasMakeMoney" would result in a post state with the sender account having a balance of 2**256-500 wei. So basically it opens up the opportunity to create as much ether as you want.


libethereum/Executive.cpp
```diff
@@ -82,7 +82,7 @@ bool Executive::setup(bytesConstRef _rlp)
 	}
```
chriseth: No cast to bigint necessary anymore in the line above :-)

PR #856, opened 2015-01-23
Function modifiers.

chriseth:


chriseth:
Tiny thing still missing: Modifiers for constructors are allowed by the parser (and should), but are not executed.

chriseth:
Correction: Modifiers on constructors already work, added a test.

PR #861, opened 2015-01-24
add evmjit build status

caktux:


chfast:
On build server, run tests as `testeth --jit`

PR #858, opened 2015-01-23
Solidity storage variable automatic accessors

LefterisJP:
This is a work in progress but comments are welcome.

Missing things are:
1) a little bit of cleanup and more tests
2) Acessors that accept parameters, like let's say ... accessor function for a mapping state variable. But this could possibly end up in a second Pull Request.

PR #877, opened 2015-01-27
Super keyword.

chriseth:
This allows access to overridden functions even in a diamond-shaped inheritance graph.


libsolidity/GlobalContext.cpp
```diff
@@ -83,5 +83,14 @@ MagicVariableDeclaration const* GlobalContext::getCurrentThis() const
...
 }
 
+MagicVariableDeclaration const* GlobalContext::getCurrentSuper() const
+{
+	if (!m_superPointer[m_currentContract])
+		m_superPointer[m_currentContract] = make_shared<MagicVariableDeclaration>(
+													"super", make_shared<ContractType>(*m_currentContract, true));
+	return m_superPointer[m_currentContract].get();
+
+}
```
LianaHus: is this space before "super..." line mistyping?

libsolidity/ExpressionCompiler.cpp
```diff
@@ -360,14 +360,22 @@ void ExpressionCompiler::endVisit(MemberAccess const& _memberAccess)
...
+		else
+		{
+			u256 identifier = type.getFunctionIdentifier(member);
+			if (identifier != Invalid256)
+			{
+				appendTypeConversion(type, IntegerType(0, IntegerType::Modifier::ADDRESS), true);
+				m_context << identifier;
+				break;
+			}
+			// fall-through to "integer" otherwise (address)
```
LianaHus: I would move this comment before if statement

libsolidity/GlobalContext.cpp
LefterisJP: no it's the unfortunate way that github chooses to show diffs with our code style. Has to do with number of spaces assigned to each tab.

LianaHus: ah. but in code(->View) was the same

libsolidity/GlobalContext.cpp
```diff
@@ -83,5 +83,14 @@ MagicVariableDeclaration const* GlobalContext::getCurrentThis() const
 
 }
 
+MagicVariableDeclaration const* GlobalContext::getCurrentSuper() const
+{
+	if (!m_superPointer[m_currentContract])
+		m_superPointer[m_currentContract] = make_shared<MagicVariableDeclaration>(
+													"super", make_shared<ContractType>(*m_currentContract, true));
+	return m_superPointer[m_currentContract].get();
+
```
gavofyork: blank line?

libsolidity/GlobalContext.cpp
chriseth: Sometimes you can add "?ts=4" to the URL in github to get the tab->space conversion we use.

chriseth:
Should be merged after #858 .

chriseth:
Removed a test since it was fragile with respect to the order of functions in the generated code and in the end it did not really test anything anyway.

PR #878, opened 2015-01-27
simplify solc's option output

guanqun:
The original one is too verbose.


LefterisJP:
Thank you that is a nice simplification. Didn't know that `boost::program_options::option_description` had such a functionality. Good to know

PR #886, opened 2015-01-28
mix: Transaction recording, standard contracts

arkpar:
UI Changes:
- Transaction log, double click on an entry to load into debugger
- State dialog now has a "default" checkbox. F5 launches default state
- New project is created with default state which includes 2 base contracts: Config and Namereg


mix/ClientModel.cpp
```diff
@@ -149,60 +169,66 @@ void ClientModel::executeSequence(std::vector<TransactionSettings> const& _seque
...
-				debuggingContent = callContract(address, transactonData.at(i), _sequence.at(i));
-
-			QList<QVariableDefinition*> returnParameters;
+					//encode data
+					f = nullptr;
+					if (t.functionId.isEmpty())
+						f = contractDef->constructor();
+					else
+					{
+						for (int tf = 0; tf < contractDef->functionsList().size(); tf++)
```
chriseth: Too many braces and perhaps use a range-based loop?

mix/ClientModel.cpp
```diff
@@ -149,60 +169,66 @@ void ClientModel::executeSequence(std::vector<TransactionSettings> const& _seque
 		try
 		{
 			bytes contractCode = compilerRes->bytes();
-			std::vector<dev::bytes> transactonData;
-			QFunctionDefinition* f = nullptr;
-			ContractCallDataEncoder c;
-			//encode data for all transactions
+			m_client->resetState(_balance);
+			onStateReset();
 			for (auto const& t: _sequence)
```
chriseth: If I read `t.functionId` in line 217 below and want to know what is actually happening, I have to go to this line to discover that the type of `t` might be derived from the type of `_sequence` which is caught by value in this lambda and defined as a `vector` of `TransactionSetting` in line 155.
I know that I might be one of the few developers that like explicit types, but please choose at least a descriptive name for `t`. :-)

mix/ClientModel.cpp
```diff
@@ -215,51 +241,131 @@ void ClientModel::executeSequence(std::vector<TransactionSettings> const& _seque
...
+	{
+		//contract creation
+		auto const stdContractName = m_stdContractNames.find(tr.contractAddress);
+		if (stdContractName != m_stdContractNames.end())
+		{
+			function = stdContractName->second;
+			contract = function;
+		}
+		else
+		{
```
chriseth: Braces

mix/ClientModel.cpp
```diff
@@ -215,51 +241,131 @@ void ClientModel::executeSequence(std::vector<TransactionSettings> const& _seque
...
+	}
 	else
 	{
-		u256 gasPrice = 10000000000000;
-		u256 gas = 125000;
-		u256 amount = 100;
-		newAddress = m_client->transact(m_client->userAccount().secret(), amount, _code, gas, gasPrice);
+		//call
+		if (tr.transactionData.size() >= 4)
+		{
```
chriseth: Braces

mix/CodeModel.h
```diff
@@ -131,6 +132,8 @@ class CodeModel: public QObject
 	bool hasContract() const;
 	/// Apply text document formatting. @todo Move this to editor module
 	void updateFormatting(QTextDocument* _document);
+	/// Get contract code by url. Contract is compiled on first access and cached
+	dev::bytes const& getStdContractCode(QString const& _url);
```
chriseth: What is the meaning of the "Std" in the name? What I can see from the implementation, this just compiles any contract in any file.

mix/CodeModel.cpp
```diff
@@ -188,3 +189,20 @@ void CodeModel::updateFormatting(QTextDocument* _document)
...
+{
+	auto cached = m_compiledContracts.find(_url);
+	if (cached != m_compiledContracts.end())
+		return cached->second;
+
+	FileIo fileIo;
+	std::string source = fileIo.readFile(_url).toStdString();
+	solidity::CompilerStack cs;
+	cs.setSource(source);
+	cs.compile(false);
```
chriseth: No exception handling?

mix/MixClient.cpp
```diff
@@ -86,25 +94,74 @@ void MixClient::executeTransaction(bytesConstRef _rlp, State& _state)
...
 
 void MixClient::validateBlock(int _block) const
 {
-	//TODO: throw exception here if _block != 0
-	(void)_block;
+	if (_block != -1 && _block != 0 && (unsigned)_block >= m_blocks.size() - 1)
+	{
+		InvalidBlockException exception;
+		exception << BlockIndex(_block);
+		BOOST_THROW_EXCEPTION(exception);
```
chriseth: All in one line?

mix/MixClient.cpp
```diff
@@ -140,92 +196,164 @@ bytes MixClient::call(Secret _secret, u256 _value, Address _dest, bytes const& _
...
-	return LocalisedLogEntries();
+	Guard l(m_filterLock);
+	h256 h = m_watches.at(_watchId).id;
+	auto filterIter = m_filters.find(h);
+	if (filterIter != m_filters.end())
+		return logs(filterIter->second.filter);
+	return eth::LocalisedLogEntries();
 }
 
-eth::LocalisedLogEntries MixClient::logs(eth::LogFilter const& _filter) const
```
chriseth: The pyramid below is visually quite appealing, so I do not dare to destroy it by a comment, but I am not sure if it is easily understandable. Please expand the names of some of the variables (it took me quite some minutes to check where `s` is actually used and that might be the number of matching logs that are to be skipped....).
It perhaps also might be useful to not introduce `m` at all. I also like to compare running indices against sizes using `<` or `<=` instead of `!=` to indicate which one is the running index and which is the size, but that is probably a matter of taste.

mix/CodeModel.h
arkpar: It also adds the contract to a cache of standard contracts. So, not that generic

mix/CodeModel.cpp
arkpar: Not required here. Handled down the stack

mix/MixClient.cpp
arkpar: Copy-pasted from libethereum. Will fix

mix/QContractDefinition.cpp
```diff
@@ -44,3 +44,14 @@ QContractDefinition::QContractDefinition(dev::solidity::ContractDefinition const
 		m_functions.append(new QFunctionDefinition(it->second, i));
 }
 
+
+QFunctionDefinition* QContractDefinition::getFunction(dev::FixedHash<4> _hash)
+{
+	if (m_constructor->hash() == _hash)
```
chriseth: Is that really used? Note that it is possible by inheritance that the final contract has a function that has the same hash as the constructor (tough, to be honest, I'm not sure whether that really works) and can be called after creation:
```
contract base { function f() { } }
contract f is base { function f() {} }
```

mix/stdc/config.sol
```diff
@@ -0,0 +1,45 @@
+//sol Config
+// Simple global configuration registrar.
+// @authors:
+//   Gav Wood <g@ethdev.com>
+#require mortal
```
chriseth: This will not compile anymore. Please use `import "qrc:///stdc/mortal.sol";` - or whatever you use as source unit identifier when you call the compiler.

mix/QContractDefinition.cpp
arkpar: not used, will remove

mix/CodeModel.cpp
```diff
@@ -188,3 +189,20 @@ void CodeModel::updateFormatting(QTextDocument* _document)
...
+dev::bytes const& CodeModel::getStdContractCode(const QString &_url)
+{
+	auto cached = m_compiledContracts.find(_url);
+	if (cached != m_compiledContracts.end())
+		return cached->second;
+
+	FileIo fileIo;
+	std::string source = fileIo.readFile(_url).toStdString();
+	solidity::CompilerStack cs;
+	cs.setSource(source);
```
chriseth: Note that if multiple contracts depend on each other, you have to `addSource(<identifier>, <content>)` them. Dependencies are resolved using import directives. But that is probably not yet in scope.

mix/stdc/config.sol
arkpar: Fixed in std.sol. This file is left as is to be reworked in future

mix/MixClient.cpp
arkpar: Done

arkpar:
OS X tests and Windows build failing due to unrelated issues

mix/MixClient.cpp
chriseth: Some braces can still go.

mix/MixClient.cpp
```diff
@@ -140,92 +198,155 @@ bytes MixClient::call(Secret _secret, u256 _value, Address _dest, bytes const& _
...
-unsigned MixClient::installWatch(eth::LogFilter const& _filter)
+unsigned MixClient::installWatch(eth::LogFilter const& _f)
 {
-	(void)_filter;
-	BOOST_THROW_EXCEPTION(InterfaceNotSupported("dev::eth::Interface::installWatch"));
+	h256 h = _f.sha3();
+	{
+		Guard l(m_filterLock);
+		if (!m_filters.count(h))
+			m_filters.insert(std::make_pair(h, _f));
```
chriseth: insert does not insert if the key already exists, so the check above is redundant.

gavofyork:
rebase?

arkpar:
Rebase done

PR #895, opened 2015-01-29
Manage solidity types

yann300:
Manage solidity types:
	- uint<N>, uint
	- int<N>, int
	- hash<N>, hash, address
	- bool
	- string<N>


libdevcore/CommonJS.cpp
```diff
@@ -45,6 +45,13 @@ bytes padded(bytes _b, unsigned _l)
 	return asBytes(asString(_b).substr(_b.size() - std::max(_l, _l)));
 }
 
+bytes paddedRight(bytes _b, unsigned _l)
+{
+	while (_b.size() < _l)
```
arkpar: Why not just _b.resize(_l); return _b; ?
Is it supposed to append zero bytes or '0' - zero character? It is not clear from the documentation comment.

mix/QVariableDefinition.cpp
```diff
@@ -53,3 +55,95 @@ QVariableDefinition* QVariableDefinitionList::val(int _idx)
...
+dev::bytes QStringType::encodeValue()
+{
+	qDebug() << QString::fromStdString(toJS(paddedRight(asBytes(value().toStdString()), 32)));
+	return paddedRight(asBytes(value().toStdString()), 32);
+}
+
+void QStringType::decodeValue(std::string const& _rawValue)
+{
+	std::string rawParam = _rawValue.substr(0, 32 * 2);
+	rawParam = unpadRight(rawParam);
```
arkpar: Why would you remove trailing zero characters?

libdevcore/CommonJS.cpp
```diff
@@ -52,6 +59,18 @@ bytes unpadded(bytes _b)
 	return _b;
 }
 
+std::string unpadRight(std::string _b)
```
arkpar: Can you give a case for this function to be required?

mix/QVariableDefinition.cpp
yann300: From the Wiki, we need to right aligned the string: 
string<N>: binary type of N bytes, N > 0. Assumed to be UTF-8 encoded, right-aligned, zero-passed data.

mix/ClientModel.cpp
```diff
@@ -168,15 +171,8 @@ void ClientModel::executeSequence(std::vector<TransactionSettings> const& _seque
...
-				{
-					QVariableDeclaration* var = (QVariableDeclaration*)f->parametersList().at(p);
-					u256 value = 0;
-					auto v = t.parameterValues.find(var->name());
-					if (v != t.parameterValues.cend())
-						value = v->second;
-					c.encode(var, value);
-				}
+				for (int p = 0; p < t.parameterValues.size(); p++)
+					c.push(t.parameterValues.at(p)->encodeValue());
```
arkpar: There needs to be a check if stored parameter type matches actual parameter type for the contract. The contract could have been edited after the state was configured

mix/ContractCallDataEncoder.cpp
```diff
@@ -74,59 +57,21 @@ QList<QVariableDefinition*> ContractCallDataEncoder::decode(QList<QVariableDecla
...
+		if (dec->type().contains("int"))
+			def = new QIntType(dec, QString());
+		else if (dec->type().contains("real"))
+			def = new QRealType(dec, QString());
+		else if (dec->type().contains("bool"))
+			def = new QBoolType(dec, QString());
+		else if (dec->type().contains("string") || dec->type().contains("text"))
+			def = new QStringType(dec, QString());
+		else if (dec->type().contains("hash") || dec->type().contains("address"))
+			def = new QHashType(dec, QString());
```
arkpar: else throw exception probably, otherwise it is a segfault

libdevcore/CommonJS.cpp
yann300: This is supposed to be zero bytes. so we can use _b.resize(_l)

mix/QVariableDeclaration.h
```diff
@@ -38,7 +40,8 @@ class QVariableDeclaration: public QBasicNodeDefinition
 	QVariableDeclaration() {}
 	QVariableDeclaration(solidity::VariableDeclaration const* _v): QBasicNodeDefinition(_v), m_type(QString::fromStdString(_v->getType()->toString())) {}
 	QVariableDeclaration(solidity::ParamDescription const& _v): QBasicNodeDefinition(_v.getName()), m_type(QString::fromStdString(_v.getType())) {}
-	QString type() const { return m_type; }
+	Q_INVOKABLE QString type() const { return m_type; }
```
arkpar: Better make if a property. Q_PROPERTY(QString type MEMBER m_type CONSTANT)

mix/ContractCallDataEncoder.cpp
```diff
@@ -44,29 +44,12 @@ void ContractCallDataEncoder::encode(QFunctionDefinition const* _function)
...
-	bytes data = padded(jsToBytes(s.str()), padding);
-	m_encodedData.insert(m_encodedData.end(), data.begin(), data.end());
-	encodedData();
-}
-
-QList<QVariableDefinition*> ContractCallDataEncoder::decode(QList<QVariableDeclaration*> _returnParameters, bytes _value)
+QList<QVariableDefinition*> ContractCallDataEncoder::decode(QList<QVariableDeclaration*> const& _returnParameters, bytes const& _value)
 {
 	QList<QVariableDefinition*> r;
 	std::string returnValue = toJS(_value);
```
arkpar: I think this conversion to string is not needed anymore

mix/QVariableDefinition.cpp
```diff
@@ -53,3 +55,95 @@ QVariableDefinition* QVariableDefinitionList::val(int _idx)
...
+{
+	dev::bigint i(value().toStdString());
+	if (i < 0)
+		i = i + dev::bigint("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff") + 1;
+	std::ostringstream s;
+	s << std::hex << "0x" << i;
+	return padded(jsToBytes(s.str()), 32);
+}
+
+void QIntType::decodeValue(std::string const& _rawValue)
```
arkpar: Why encode to bytes directly but decode to string and then from string to value?

mix/ClientModel.cpp
yann300: ok so I guess that check should be earlier in the use case.
Like : 
When the user click on "Run State" (before entering in the debugState function).
When the user edit a state (we can display transaction that are not in scope in red).

libdevcore/CommonJS.cpp
yann300: This has to be used only for real string. But the input parameter is a string not an int ...
Probably should add comment to warn user to not use with a non parse int.

mix/QVariableDefinition.cpp
```diff
@@ -53,3 +55,95 @@ QVariableDefinition* QVariableDefinitionList::val(int _idx)
 		return nullptr;
 	return m_def.at(_idx);
 }
+
+/*
+ * QIntType
+ */
+dev::bytes QIntType::encodeValue()
+{
+	dev::bigint i(value().toStdString());
```
arkpar: I think you can just use dev::toBigEndian here

mix/QVariableDefinition.cpp
arkpar: Consider using dev::fromBigEndian

yann300: so we need to encode to byte in order to call m_client.transact(). And we need to decode to string ... to use the return parameter. I don't get your point here.

mix/QVariableDefinition.cpp
```diff
@@ -53,3 +55,95 @@ QVariableDefinition* QVariableDefinitionList::val(int _idx)
...
+/*
+ * QStringType
+ */
+dev::bytes QStringType::encodeValue()
+{
+	qDebug() << QString::fromStdString(toJS(paddedRight(asBytes(value().toStdString()), 32)));
+	return paddedRight(asBytes(value().toStdString()), 32);
+}
+
+void QStringType::decodeValue(std::string const& _rawValue)
```
arkpar: This should not require converting a string to hex string and back

mix/ContractCallDataEncoder.cpp
yann300: At some point we need to convert to string to extract the return parameters ...

mix/QVariableDefinition.cpp
yann300: for a string, the return value is hex encoded, so we need to convert to ASCII char.

mix/ClientModel.cpp
arkpar: Checking on run will be enough for now I suppose, This should be done in this method

libdevcore/CommonJS.cpp
```diff
@@ -52,12 +58,21 @@ bytes unpadded(bytes _b)
...
 {
-	auto p = _b.find_first_not_of('0');
-	if (p == std::string::npos)
-		return "0";
-	return _b.substr(p, _b.length() - 1);
+	int i = 0;
+	if (_b.size() == 0)
+		return _b;
+	while (true)
+	{
```
gavofyork: no need for {s

libdevcore/CommonJS.h
```diff
@@ -48,12 +48,14 @@ inline std::string toJS(dev::bytes const& _n)
...
 bytes jsToBytes(std::string const& _s);
-/// Add '0' on the head of _b until _l.
+/// Add '0' on the head of @a _b until @a _l.
 bytes padded(bytes _b, unsigned _l);
+/// Add '0' on the queue of @a _b until @a _l.
+bytes paddedRight(bytes _b, unsigned _l);
 /// Removing all trailing '0'. Returns empty array if input contains only '0' char.
 bytes unpadded(bytes _s);
-/// Remove all '0' on the head of _s. Returns 0 if _s contains only '0'.
-std::string unpadLeft(std::string _s);
```
gavofyork: this alteration of a base api function doesn't break stuff in the rest of the project?

yann300: This function was only used in mix and it is not used anymore.

libdevcore/CommonJS.cpp
```diff
@@ -52,12 +58,21 @@ bytes unpadded(bytes _b)
...
 {
-	auto p = _b.find_first_not_of('0');
-	if (p == std::string::npos)
-		return "0";
-	return _b.substr(p, _b.length() - 1);
+	int i = 0;
+	if (_b.size() == 0)
+		return _b;
+
+	while (true)
```
arkpar: Will cause exception is if there is no zero byte in the vector.
You can use 
while(i < _b.size() && _b[i] == 0)
   i++;

mix/ClientModel.cpp
```diff
@@ -204,16 +208,14 @@ void ClientModel::executeSequence(std::vector<TransactionSettings> const& _seque
...
-						auto v = transaction.parameterValues.find(var->name());
-						if (v != transaction.parameterValues.cend())
-							value = v->second;
-						encoder.encode(var, value);
+						if (f->parametersList().at(p)->type() != transaction.parameterValues.at(p)->declaration()->type())
+							BOOST_THROW_EXCEPTION(ParameterChangedException() << FunctionName(f->parametersList().at(p)->type().toStdString()));
+						encoder.push(transaction.parameterValues.at(p)->encodeValue());
 					}
 
+
```
arkpar: extra linebreak

mix/ClientModel.cpp
```diff
@@ -224,7 +226,9 @@ void ClientModel::executeSequence(std::vector<TransactionSettings> const& _seque
 						}
 					}
 					else
+					{
```
arkpar: extra braces

mix/QVariableDefinition.cpp
```diff
@@ -53,3 +55,92 @@ QVariableDefinition* QVariableDefinitionList::val(int _idx)
...
+{
+	m_bigIntvalue = _value;
+	std::stringstream str;
+	str << std::dec << m_bigIntvalue;
+	m_value = QString::fromStdString(str.str());
+}
+
+dev::bytes QIntType::encodeValue()
+{
+	dev::bigint i(value().toStdString());
```
arkpar: can't m_bigIntvalue be used here?

mix/QVariableDefinition.cpp
```diff
@@ -53,3 +55,92 @@ QVariableDefinition* QVariableDefinitionList::val(int _idx)
...
+		bigint = bigint - dev::bigint("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff") - 1;
+	setValue(bigint);
+}
+
+/*
+ * QHashType
+ */
+dev::bytes QHashType::encodeValue()
+{
+	QByteArray b = value().toUtf8();
```
arkpar: I don't think you need toUtf8 here for hash

mix/QVariableDefinition.cpp
```diff
@@ -53,3 +55,92 @@ QVariableDefinition* QVariableDefinitionList::val(int _idx)
...
+{
+	setValue(QString::fromUtf8((char*)_rawValue.data()));
+}
+
+/*
+ * QBoolType
+ */
+dev::bytes QBoolType::encodeValue()
+{
+	return padded(jsToBytes(value().toStdString()), 32);
```
arkpar: can't m_boolValue be used here?

mix/QVariableDefinition.cpp
yann300: At this point, QIntType comes from QML, so we only have the string value. 
Maybe we can add a function to resolve the bigint value in QML side for thoses specific types (bool, int, uint).

yann300: but does it worth it?

libdevcore/CommonJS.cpp
arkpar: My mistake, it is ok

mix/QVariableDefinition.cpp
yann300: Same as for m_bigIntValue.

yann300:
When adding a function containing parameters, String<N> types should be managed by mix (as well as uint<N>, int<N>, hash<N>, bool). Return parameters are shown in the transaction logs table view (column returned).

PR #896, opened 2015-01-29
-added ability to pass parameters without name to the function, or retur...

LianaHus:
added ability to pass parameters without name to the function, or return them

### DIFF
--- a/libsolidity/DeclarationContainer.cpp
+++ b/libsolidity/DeclarationContainer.cpp
@@ -30,6 +30,9 @@ namespace solidity
 
 bool DeclarationContainer::registerDeclaration(Declaration const& _declaration, bool _update)
 {
+	if (_declaration.getName().empty())
+		return true;
+
 	if (!_update && m_declarations.find(_declaration.getName()) != m_declarations.end())
 		return false;
 	m_declarations[_declaration.getName()] = &_declaration;
@@ -38,6 +41,7 @@ bool DeclarationContainer::registerDeclaration(Declaration const& _declaration, 
 
 Declaration const* DeclarationContainer::resolveName(ASTString const& _name, bool _recursive) const
 {
+	solAssert(!_name.empty(), "Attempt to resolve empty name.");
 	auto result = m_declarations.find(_name);
 	if (result != m_declarations.end())
 		return result->second;
--- a/libsolidity/DeclarationContainer.cpp
+++ b/libsolidity/DeclarationContainer.cpp
@@ -31,7 +31,9 @@ namespace solidity
 bool DeclarationContainer::registerDeclaration(Declaration const& _declaration, bool _update)
 {
 	if (_declaration.getName().empty())
-		return true;
+		if (_declaration.getType() != nullptr)
+			if (_declaration.getType()->getCategory() != Type::Category::FUNCTION)
+				return true;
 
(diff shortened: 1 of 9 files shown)


libsolidity/DeclarationContainer.cpp
```diff
@@ -38,6 +41,7 @@ bool DeclarationContainer::registerDeclaration(Declaration const& _declaration,
 
 Declaration const* DeclarationContainer::resolveName(ASTString const& _name, bool _recursive) const
 {
+	solAssert(!_name.empty(), "It's not poosible resolve the name if the declaration is not registered.");
```
chriseth: Suggestion for better message: "Attempt to resolve empty name."

libsolidity/DeclarationContainer.h
```diff
@@ -42,8 +42,8 @@ class DeclarationContainer
 	explicit DeclarationContainer(Declaration const* _enclosingDeclaration = nullptr,
 								  DeclarationContainer const* _enclosingContainer = nullptr):
 		m_enclosingDeclaration(_enclosingDeclaration), m_enclosingContainer(_enclosingContainer) {}
-	/// Registers the declaration in the scope unless its name is already declared.
-	/// @returns true iff it was not yet declared.
+	/// Registers the declaration in the scope unless its name is already declared or the name is empty.
+	/// @returns true if the declaration was registered.
```
chriseth: If the name is empty, nothing is registered, so it should rather read "returns false if the name was already declared.".

test/SolidityEndToEndTest.cpp
```diff
@@ -1955,6 +1955,21 @@ BOOST_AUTO_TEST_CASE(super_in_constructor)
...
+				uint g = 8;
+				ret_k = k;
+				ret_g = g;
+			}
+		})";
+	compileAndRun(sourceCode);
+	BOOST_CHECK(callContractFunction("f(uint256,uint256)", 5, 9) != encodeArgs(5, 8));
+	BOOST_CHECK(callContractFunction("f(uint256,uint256)", 5, 9) == encodeArgs(9, 8));
+}
+
```
chriseth: The most important use-case is probably "function f(uint x) returns (uint) { ... return a; }" and I think it should have a test.

chriseth:
Needs rebase

libsolidity/DeclarationContainer.cpp
```diff
@@ -30,6 +30,11 @@ namespace solidity
 
 bool DeclarationContainer::registerDeclaration(Declaration const& _declaration, bool _update)
 {
+	if (_declaration.getName().empty())
+		if (_declaration.getType() != nullptr)
```
chriseth: I think it would be a cleaner solution to not distinguish between functions and other declarations here.
We should rather have a check inside `ContractDefinition::checkTypeRequirements` along these lines:
```
FunctionDefinition const* fallbackFunction = nullptr;
for (ASTPointer<FunctionDefinition> const& function: getDefinedFunctions())
  if (function->getName().empty())
    if (fallbackFunction)
      BOOST_THROW_EXCEPTION(...);
    else
      fallbackFunction = function.get();
```
And then continue with the check that is already there, but do not again test for the scope.

arkpar:
Please add tests for json and solidity ABI generators. Natspec generator is probably affected as well

LefterisJP:
The new changes seem fine but you need a rebase.

PR #906, opened 2015-01-30
implement named arguments for function call

guanqun:
This commit tries to add the support for named arguments as seen in other languages.

It would support the following syntax:

    func({a:1, b:2})

### DIFF
--- a/libsolidity/AST.h
+++ b/libsolidity/AST.h
@@ -983,14 +983,15 @@ class FunctionCall: public Expression
 {
 public:
 	FunctionCall(Location const& _location, ASTPointer<Expression> const& _expression,
-				 std::vector<ASTPointer<Expression>> const& _arguments):
-		Expression(_location), m_expression(_expression), m_arguments(_arguments) {}
+				 std::vector<ASTPointer<Expression>> const& _arguments, std::vector<std::string> const& _names):
+		Expression(_location), m_expression(_expression), m_arguments(_arguments), m_names(_names) {}
 	virtual void accept(ASTVisitor& _visitor) override;
 	virtual void accept(ASTConstVisitor& _visitor) const override;
 	virtual void checkTypeRequirements() override;
 
 	Expression const& getExpression() const { return *m_expression; }
 	std::vector<ASTPointer<Expression const>> getArguments() const { return {m_arguments.begin(), m_arguments.end()}; }
+	std::vector<std::string> const& getNames() const { return m_names; }
 
 	/// Returns true if this is not an actual function call, but an explicit type conversion
 	/// or constructor call.
@@ -999,6 +1000,7 @@ public:
 private:
 	ASTPointer<Expression> m_expression;
 	std::vector<ASTPointer<Expression>> m_arguments;
+	std::vector<std::string> m_names;
 };
 
 /**
--- a/libsolidity/ExpressionCompiler.cpp
+++ b/libsolidity/ExpressionCompiler.cpp
@@ -193,6 +193,7 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
 	{
 		//@todo struct construction
 		solAssert(_functionCall.getArguments().size() == 1, "");
+		solAssert(_functionCall.getNames().empty(), "");
 		Expression const& firstArgument = *_functionCall.getArguments().front();
 		firstArgument.accept(*this);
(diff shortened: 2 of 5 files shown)


chriseth:
This is great! We planned this feature with exactly the syntax you give :-)
Could you add some tests for parsing, type checking and execution (end-to-end)?

libsolidity/AST.h
```diff
@@ -999,6 +1000,7 @@ class FunctionCall: public Expression
 private:
 	ASTPointer<Expression> m_expression;
 	std::vector<ASTPointer<Expression>> m_arguments;
+	std::vector<std::string> m_names;
```
chriseth: Please change this to `std::vector<ASTPointer<ASTString>> m_names;`. Yes, it is ugly, but it should be consistent and allows for some planned optimizations to be made more fluently.

libsolidity/ExpressionCompiler.cpp
```diff
@@ -200,8 +201,35 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+		vector<string> const& parameterNames = function.getParameterNames();
+		vector<ASTPointer<Expression const>> const& callArguments = _functionCall.getArguments();
+		vector<string> const& callArgumentNames = _functionCall.getNames();
+		solAssert(callArguments.size() == parameterTypes.size(), "");
+
+		vector<ASTPointer<Expression const>> arguments;
+		if (callArgumentNames.empty())
+		{
+			// normal arguments
+			arguments = {callArguments.begin(), callArguments.end()};
```
chriseth: Coding style does not allow braces or single statements.
Furthermore, I think the initializer list confuses visual c++ due to a bug in that compiler.
Perhaps it would be better to change the type of `arguments` to a pointer in order to avoid the copy here in the (probably more common) positional call case.

libsolidity/ExpressionCompiler.cpp
```diff
@@ -200,8 +201,35 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
+		}
+		else
+		{
+			// named arguments
+			for (size_t i = 0; i < parameterNames.size(); i++) {
```
chriseth: The local variable `parameterNames` could be avoided by using a range-based loop here.

libsolidity/Parser.cpp
```diff
@@ -692,6 +694,39 @@ vector<ASTPointer<Expression>> Parser::parseFunctionCallArguments()
 	return arguments;
 }
 
+void Parser::parseFunctionCallArguments(vector<ASTPointer<Expression>>& _arguments, vector<string>& _names)
```
chriseth: Please return a `pair` or a `tuple` and use `std::tie` at the call site.

libsolidity/Parser.cpp
```diff
@@ -692,6 +694,39 @@ vector<ASTPointer<Expression>> Parser::parseFunctionCallArguments()
...
+{
+	Token::Value token = m_scanner->getCurrentToken();
+	if (token == Token::LBRACE)
+	{
+		// call({arg1 : 1, arg2 : 2 })
+		expectToken(Token::LBRACE);
+		while (m_scanner->getCurrentToken() != Token::RBRACE)
+		{
+			string identifier = *expectIdentifierToken();
```
chriseth: Please keep the ASTPointer. And perhaps do not introduce local variables `identifier` and `expression`?

libsolidity/Parser.cpp
```diff
@@ -692,6 +694,39 @@ vector<ASTPointer<Expression>> Parser::parseFunctionCallArguments()
...
+			expectToken(Token::COLON);
+			ASTPointer<Expression> expression = parseExpression();
+
+			_arguments.push_back(expression);
+			_names.push_back(identifier);
+
+			if (m_scanner->getCurrentToken() == Token::COMMA)
+			{
```
chriseth: This and the following two pairs of braces should go according to the coding style.

chriseth:
Adding tag "gotissues" according to our workflow. If you think you addressed all comments, please remove that tag and add "pleasereview" to signal that.

guanqun:
Thanks @chriseth for giving these comments, I was out for travel and no wifi for checking this... Thanks @gavofyork for fixing these comments.

So what I'm left is to add some tests: here it is: [pull request 948](https://github.com/ethereum/cpp-ethereum/pull/948)

PR #916, opened 2015-01-31
transaction address length test fix

winsvega:
address which is less or more 20 size should be invalid
 however we could produce a valid Address from the string of wrong size but it would be diffrent address
 thus leads us to erros with transaction tests. cause if we mark this case as invalid,
 we wont include transaction section in test result json (see wiki tests)
 but when test runner would try to read such rlp it would consider it valid.
 Right now creating Address from string of invalid length results in diffrent valid address creation
 however such test could not be impemented


test/TestHelper.cpp
```diff
@@ -108,6 +108,12 @@ void ImportTest::importState(json_spirit::mObject& _o, State& _state)
 		BOOST_REQUIRE(o.count("storage") > 0);
 		BOOST_REQUIRE(o.count("code") > 0);
 
+		bigint biValue256 = bigint("115792089237316195423570985008687907853269984665640564039457584007913129639936");
+		if (bigint(o["balance"].get_str()) >= biValue256)
+			BOOST_THROW_EXCEPTION(Exception() << errinfo_comment("State 'balance' is equal or grater than 2**256") );
```
CJentzsch: greater (not grater)

winsvega:
Is that all?

test/transaction.cpp
```diff
@@ -152,6 +151,9 @@ void doTransactionTests(json_spirit::mValue& _v, bool _fillin)
 				if (!txFromFields.signature().isValid())
 					BOOST_THROW_EXCEPTION(Exception() << errinfo_comment("transaction from RLP signature is invalid") );
 
+				//cause Address is length20 array, when trying to create address from sting of another length, field "to" would be diffrent from RLP encoded Address
+				BOOST_CHECK_MESSAGE(Address(tObj["to"].get_str()) == txFromFields.receiveAddress(), "seems that transaction 'to' address has wrong format");
```
CJentzsch: If I put in a sting which is longer than 20, but looks like 0xffff...fff . You still get a valid address.
But also you do not need this line at all. When creating the Transaction from an RLP this line gets executed:
m_receiveAddress = rlp[field = 3].toHash<Address>();
In toHash an exception is thrown if the size of the input data does not match the size of Address. Therefore you could never reach that line in that case.

winsvega: Oh, I will check it (the exception) 
cause I didn't got anything, just wrong address created from string with incorrect length

winsvega: I see. It is just that I cant create this situation with fillers. Cause when filler runs it will convert string to Address: rlpStream << Address(tObj["to"].get_str());

CJentzsch: I still dont see how this error can ever be reached. Since the constructor of txfromFields will already throw if this is the case.

test/transaction.cpp
```diff
@@ -84,7 +100,7 @@ void doTransactionTests(json_spirit::mValue& _v, bool _fillin)
 			BOOST_REQUIRE(o.count("transaction") > 0);
 
 			mObject tObj = o["transaction"].get_obj();
-			Transaction txFromFields = createTransactionFromFields(tObj);
+			Transaction txFromFields(createRLPStreamFromTransactionFields(tObj).out(),CheckSignature::Sender);
```
CJentzsch: space after ","

test/TestHelper.cpp
```diff
@@ -140,6 +146,17 @@ void ImportTest::importTransaction(json_spirit::mObject& _o)
...
+	if (bigint(_o["nonce"].get_str()) >= biValue256)
+		BOOST_THROW_EXCEPTION(Exception() << errinfo_comment("Transaction 'nonce' is equal or greater than 2**256") );
+	if (bigint(_o["gasPrice"].get_str()) >= biValue256)
+		BOOST_THROW_EXCEPTION(Exception() << errinfo_comment("Transaction 'gasPrice' is equal or greater than 2**256") );
+	if (bigint(_o["gasLimit"].get_str()) >= biValue256)
+		BOOST_THROW_EXCEPTION(Exception() << errinfo_comment("Transaction 'gasLimit' is equal or greater than 2**256") );
+	if (bigint(_o["value"].get_str()) >= biValue256)
+		BOOST_THROW_EXCEPTION(Exception() << errinfo_comment("Transaction 'value' is equal or greater than 2**256") );
+
+
```
CJentzsch: only one empty line

test/TestHelper.cpp
```diff
@@ -108,6 +108,12 @@ void ImportTest::importState(json_spirit::mObject& _o, State& _state)
 		BOOST_REQUIRE(o.count("storage") > 0);
 		BOOST_REQUIRE(o.count("code") > 0);
 
+		bigint biValue256 = bigint("115792089237316195423570985008687907853269984665640564039457584007913129639936");
+		if (bigint(o["balance"].get_str()) >= biValue256)
+			BOOST_THROW_EXCEPTION(Exception() << errinfo_comment("State 'balance' is equal or greater than 2**256") );
```
CJentzsch: Add a specific exception type such as "ValueTooLarge" or "Stackoverflow" to Excpetions.h, and use that instead of a general Exception. Same for exceptions below.

test/ttTransactionTestFiller.json
```diff
@@ -242,14 +241,14 @@
         }
     },
 
-   "AddressMore20" : {
+    "WrongAddress" : {
         "transaction" :
 	{ 	    
  	    "data" : "",
             "gasLimit" : "",
```
CJentzsch: a lot of lines in the fillers are not aligned. For the json file you can use Ctrl + A and Ctrl + I in QtCreator to auto align all lines in the file.

test/TestHelper.cpp
```diff
@@ -108,6 +108,12 @@ void ImportTest::importState(json_spirit::mObject& _o, State& _state)
 		BOOST_REQUIRE(o.count("storage") > 0);
 		BOOST_REQUIRE(o.count("code") > 0);
 
+		bigint biValue256 = bigint("115792089237316195423570985008687907853269984665640564039457584007913129639936");
```
gavofyork: is this `2 ^ 256`? if so just express as `bigint(1) << 256`.

test/TestHelper.cpp
```diff
@@ -140,6 +146,16 @@ void ImportTest::importTransaction(json_spirit::mObject& _o)
 	BOOST_REQUIRE(_o.count("secretKey") > 0);
 	BOOST_REQUIRE(_o.count("data") > 0);
 
+	bigint biValue256 = bigint("115792089237316195423570985008687907853269984665640564039457584007913129639936");
```
gavofyork: as above

libdevcore/Exceptions.h
```diff
@@ -33,6 +33,7 @@ namespace dev
 // base class for all exceptions
 struct Exception: virtual std::exception, virtual boost::exception { mutable std::string m_message; };
 
+struct ValueTooLarge: virtual Exception {};
```
gavofyork: what's the difference between this and `BadCast` as used on line 262 of RLP.h?

gavofyork: move exceptions specific to tests into `test` dir.

test/TestHelper.cpp
```diff
@@ -108,6 +108,12 @@ void ImportTest::importState(json_spirit::mObject& _o, State& _state)
 		BOOST_REQUIRE(o.count("storage") > 0);
 		BOOST_REQUIRE(o.count("code") > 0);
 
+		bigint biValue256 = bigint(1) << 256;
```
gavofyork: `bigint biValue256 = bigint(1) << 256;`

becomes:

`bigint const c_max256plus1 = bigint(1) << 256;`

(move descriptive variable name, note it is `const`).

test/TestHelper.cpp
```diff
@@ -140,6 +146,16 @@ void ImportTest::importTransaction(json_spirit::mObject& _o)
 	BOOST_REQUIRE(_o.count("secretKey") > 0);
 	BOOST_REQUIRE(_o.count("data") > 0);
 
+	bigint biValue256 = bigint(1) << 256;
```
gavofyork: as above.

PR #922, opened 2015-02-01
Sol state variable accessors for mapping

LefterisJP:
Is a work in progress

### DIFF
--- a/libsolidity/ExpressionCompiler.cpp
+++ b/libsolidity/ExpressionCompiler.cpp
@@ -823,26 +823,85 @@ unsigned ExpressionCompiler::appendArgumentCopyToMemory(TypePointers const& _typ
 	return length;
 }
 
-unsigned ExpressionCompiler::appendExpressionCopyToMemory(Type const& _expectedType,
-														  Expression const& _expression, unsigned _memoryOffset)
+unsigned ExpressionCompiler::appendTypeCopyToMemory(Type const& _expectedType, TypePointer const& _type,
+													Location const& _location, unsigned _memoryOffset)
 {
-	_expression.accept(*this);
-	appendTypeConversion(*_expression.getType(), _expectedType, true);
+	appendTypeConversion(*_type, _expectedType, true);
 	unsigned const c_numBytes = CompilerUtils::getPaddedSize(_expectedType.getCalldataEncodedSize());
 	if (c_numBytes == 0 || c_numBytes > 32)
 		BOOST_THROW_EXCEPTION(CompilerError()
-							  << errinfo_sourceLocation(_expression.getLocation())
+							  << errinfo_sourceLocation(_location)
 							  << errinfo_comment("Type " + _expectedType.toString() + " not yet supported."));
 	bool const c_leftAligned = _expectedType.getCategory() == Type::Category::STRING;
 	bool const c_padToWords = true;
 	return CompilerUtils(m_context).storeInMemory(_memoryOffset, c_numBytes, c_leftAligned, c_padToWords);
 }
 
+unsigned ExpressionCompiler::appendExpressionCopyToMemory(Type const& _expectedType,
+														  Expression const& _expression,
+														  unsigned _memoryOffset)
+{
+	_expression.accept(*this);
+	return appendTypeCopyToMemory(_expectedType, _expression.getType(), _expression.getLocation(), _memoryOffset);
+}
+
 void ExpressionCompiler::appendStateVariableAccessor(VariableDeclaration const& _varDecl)
 {
-	m_currentLValue.fromStateVariable(_varDecl, _varDecl.getType());
-	solAssert(m_currentLValue.isInStorage(), "");
-	m_currentLValue.retrieveValue(_varDecl.getType(), Location(), true);
+	auto mappingType = dynamic_cast<const MappingType*>(_varDecl.getType().get());
+	unsigned sizeOnStack;
+	if (mappingType != nullptr)
+	{
+		// this copies from Compiler::visit(FunctionDefinition..) for argument reading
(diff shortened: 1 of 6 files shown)


libsolidity/ExpressionCompiler.cpp
```diff
@@ -823,26 +823,85 @@ unsigned ExpressionCompiler::appendArgumentCopyToMemory(TypePointers const& _typ
 	return length;
 }
 
-unsigned ExpressionCompiler::appendExpressionCopyToMemory(Type const& _expectedType,
-														  Expression const& _expression, unsigned _memoryOffset)
+unsigned ExpressionCompiler::appendTypeCopyToMemory(Type const& _expectedType, TypePointer const& _type,
```
chriseth: Perhaps "appendTypeConversionAndCopyToMemory"?

test/SolidityEndToEndTest.cpp
```diff
@@ -919,6 +919,24 @@ BOOST_AUTO_TEST_CASE(multiple_elementary_accessors)
...
+							 "    to_bool_map[42] = false;\n"
+							 "    to_uint_map[42] = 12;\n"
+							 "  }\n"
+							 "}\n";
+	compileAndRun(sourceCode);
+	BOOST_CHECK(callContractFunction("to_string_map(uint256)", 42) == encodeArgs("24"));
+	BOOST_CHECK(callContractFunction("to_bool_map(uint256)", 42) == encodeArgs(false));
+	BOOST_CHECK(callContractFunction("to_uint_map(uint256)", 42) == encodeArgs(12));
+}
+
```
chriseth: I'm not sure if we are already there, but at some point, we should also have tests for things like `mapping(uint => mapping(uint => "some_struct"))`.

libsolidity/ExpressionCompiler.cpp
```diff
@@ -823,26 +823,85 @@ unsigned ExpressionCompiler::appendArgumentCopyToMemory(TypePointers const& _typ
...
+	unsigned sizeOnStack;
+	if (mappingType != nullptr)
+	{
+		// this copies from Compiler::visit(FunctionDefinition..) for argument reading
+		unsigned parameterSize = mappingType->getKeyType()->getSizeOnStack();
+		m_context.adjustStackOffset(parameterSize);
+		m_context.addVariable(_varDecl, parameterSize);
+		// this copies from ExpressionCompiler::visit(IndexAccess .. ) for mapping access
+		TypePointer const& keyType = mappingType->getKeyType();
+		unsigned length = appendTypeCopyToMemory(*keyType, mappingType->getValueType(), Location());
```
chriseth: Conversion from key to value type?

libsolidity/ExpressionCompiler.cpp
```diff
@@ -823,26 +823,85 @@ unsigned ExpressionCompiler::appendArgumentCopyToMemory(TypePointers const& _typ
...
-	m_currentLValue.fromStateVariable(_varDecl, _varDecl.getType());
-	solAssert(m_currentLValue.isInStorage(), "");
-	m_currentLValue.retrieveValue(_varDecl.getType(), Location(), true);
+	auto mappingType = dynamic_cast<const MappingType*>(_varDecl.getType().get());
+	unsigned sizeOnStack;
+	if (mappingType != nullptr)
+	{
+		// this copies from Compiler::visit(FunctionDefinition..) for argument reading
+		unsigned parameterSize = mappingType->getKeyType()->getSizeOnStack();
+		m_context.adjustStackOffset(parameterSize);
```
chriseth: I think this should not be here.

libsolidity/ExpressionCompiler.cpp
```diff
@@ -823,26 +823,85 @@ unsigned ExpressionCompiler::appendArgumentCopyToMemory(TypePointers const& _typ
...
+		m_context.addVariable(_varDecl, parameterSize);
+		// this copies from ExpressionCompiler::visit(IndexAccess .. ) for mapping access
+		TypePointer const& keyType = mappingType->getKeyType();
+		unsigned length = appendTypeCopyToMemory(*keyType, mappingType->getValueType(), Location());
+		solAssert(length == 32, "Mapping key has to take 32 bytes in memory (for now).");
+		// @todo move this once we actually use memory
+		length += CompilerUtils(m_context).storeInMemory(length);
```
chriseth: You need to retrieve the position of the mapping `getStorageLocationOfVariable` to the stack, too. In `visit(IndexAccess` this is done by `_indexAccess.getBaseExpression().accept(*this);`.

libsolidity/ExpressionCompiler.cpp
```diff
@@ -823,26 +823,85 @@ unsigned ExpressionCompiler::appendArgumentCopyToMemory(TypePointers const& _typ
...
-	solAssert(m_currentLValue.isInStorage(), "");
-	m_currentLValue.retrieveValue(_varDecl.getType(), Location(), true);
+	auto mappingType = dynamic_cast<const MappingType*>(_varDecl.getType().get());
+	unsigned sizeOnStack;
+	if (mappingType != nullptr)
+	{
+		// this copies from Compiler::visit(FunctionDefinition..) for argument reading
+		unsigned parameterSize = mappingType->getKeyType()->getSizeOnStack();
+		m_context.adjustStackOffset(parameterSize);
+		m_context.addVariable(_varDecl, parameterSize);
```
chriseth: This adds a local variable, so this is not needde.

libsolidity/ExpressionCompiler.cpp
LefterisJP: this is also how FunctionDeclaration's parameters are read at Compiler.cpp. If this is not there I get assertions in the subsequent `m_context.addVariable()` for `m_asm.deposit()` being zero

PR #927, opened 2015-02-02
mix on osx

debris:
workaround for macdeployqt 5.4 bug


subtly:
This has problems after running install. Having to reclean everything between compiles and there are several errors when building.

debris:
what are the errors? we have checked, and it's working on two macs.

subtly:
After running install target the app doesn't run. To reproduce:
set target to Mix, Build, Run, set target to install, build. copy mix.app somewhere. try to run mix.app.

debris:
after install (which also produced a few errors) I cannot run mix. I'm getting the following error:
```bash
Dyld Error Message:
  Library not loaded: @executable_path/../../Contents/MacOS/QtWebEngineCore
  Referenced from: /Users/marekkotewicz/ethereum/cpp-ethereum/build/mix/Debug/mix.app/Contents/Frameworks/QtWebEngine.framework/Versions/5/QtWebEngine
  Reason: image not found
```

When i looked into QtWebEngine, i've noticed that path to QtWebEngineCore is broken. 
`otool -L QtWebEngine`
```bash
QtWebEngine:
	@executable_path/../Frameworks/QtWebEngine.framework/Versions/5/QtWebEngine (compatibility version 5.4.0, current version 5.4.0)
	@executable_path/../Frameworks/QtQuick.framework/Versions/5/QtQuick (compatibility version 5.4.0, current version 5.4.0)
	@executable_path/../Frameworks/QtGui.framework/Versions/5/QtGui (compatibility version 5.4.0, current version 5.4.0)
	@executable_path/../Frameworks/QtCore.framework/QtCore (compatibility version 5.4.0, current version 5.4.0)
	/System/Library/Frameworks/DiskArbitration.framework/Versions/A/DiskArbitration (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
	@executable_path/../Frameworks/QtQml.framework/Versions/5/QtQml (compatibility version 5.4.0, current version 5.4.0)
	@executable_path/../Frameworks/QtNetwork.framework/Versions/5/QtNetwork (compatibility version 5.4.0, current version 5.4.0)
	@executable_path/../../Contents/MacOS/QtWebEngineCore (compatibility version 5.4.0, current version 5.4.0)
	/System/Library/Frameworks/OpenGL.framework/Versions/A/OpenGL (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/AGL.framework/Versions/A/AGL (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1213.0.0)
```

path to QtWebEngineCore points to dylib directory instead of frameworks directory.
`@executable_path/../../Contents/MacOS/QtWebEngineCore (compatibility version 5.4.0, current version 5.4.0)`

i will try to investigate further this issue and find a solution. but it seems like another serious macdeployqt bug.

debris:
I investigated the problem today, and I can now list all the issues that we have and propose a solution for it:

#### issues:
##### 1. 
macdeployqt in qt 5.4 is broken. It has problems with Qt plugins. This script: [macdeployfix.sh](https://github.com/debris/cpp-ethereum/blob/mixonosx/macdeployfix.sh) is resolving part of the macdeployqt issues, and makes mix buildable.

##### 2. 
fixup_bundle is not working properly with qt 5.4 plugins (QtWebEngine exactly, I don't why yet). I described it [here](https://github.com/ethereum/cpp-ethereum/pull/927#issuecomment-72616540). I think, that we cannot use it anymore.

##### 3. 
our favorite! jsonrpccpp! 
so here is proper output of running `otool -L` on jsoncpp.dylib
```bash
--- local/lib ‹master› » otool -L libjsoncpp.dylib
libjsoncpp.dylib:
	/usr/local/lib/libjsoncpp.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
```

and this is the output of running `otool -L` on libjsonrpccpp
```bash
--- local/lib ‹master› » otool -L libjsonrpccpp-server.dylib
libjsonrpccpp-server.dylib:
	libjsonrpccpp-server.0.dylib (compatibility version 0.0.0, current version 0.3.2)
	libjsonrpccpp-common.0.dylib (compatibility version 0.0.0, current version 0.3.2)
	/usr/local/lib/libjsoncpp.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
```

so the first line of output -L should be the path to the library itself. We can see, that libjsoncpp is ok. libjsonrpccpp is not, cause the path is relative. If we are running just build, we cannot spot the difference, cause we are using libsjsonrpccpp from `/usr/local/lib`. But if we want to make self contained bundle we have a problem. During install phase, we can see errors that libjsonrpccpp cannot be found.

`cannot resolve item '@executable_path/../Frameworks/libjsonrpccpp-server.0.dylib`

that's because it was not copied to bundle at all!

##### 4. 
*not an issue, but also can be improved*
After build, QtWebEngineProcess (mix.app/Contents/Frameworks/QtWebEngineCore.framework/Helpers/QtWebEngineProcess.app/Contents/MacOS) have broken paths to QtWebEngineCore && QtCore. I'm saying broken, cause they are pointing to symbolic links, that are not copied proparly while standard 'cp' command is being used. However 'cp -R' or drag'n'drop on mix.app will work properly.

#### as a solution I think that we should:
##### 1. 
use macdeplofix.sh after macdeployqt
##### 2. 
stop using fixup_bundle
##### 3. 
alter our ethereum homebrew tap to use install_name_tool after installation as jsoncpp does (https://github.com/Homebrew/homebrew/blob/master/Library/Formula/jsoncpp.rb#L28).
##### 4. 
(optional) write a script using install_name_tool to fix paths for QtWebEngineProcess

debris:
issue 1. fixed.
issue 2. fixed. we don't need to run fixup_bundle any more, so we [cant remove duplicated libs](8bdf589117461f8ef21ca510960ccdfb65ec3bc7). we will only run verify_app to check, if everything is ok.
issue 3. closed. occurs only if you install jsonrpc manually. installation from current version of ethereum homebrew-tap works fine.
issue 4. not relevant.

I am able to run mix bundle on clean machine.

PR #932, opened 2015-02-02
Visibility specifiers.

chriseth:
Note that currently, only public and non-public is distinguished, i.e. private and protected behave in the same way.

### DIFF
--- a/libsolidity/AST.h
+++ b/libsolidity/AST.h
@@ -133,12 +133,17 @@ class Declaration: public ASTNode
 {
 public:
 	enum class LValueType { NONE, LOCAL, STORAGE };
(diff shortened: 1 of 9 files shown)


libsolidity/AST.h
```diff
@@ -330,16 +339,15 @@ class FunctionDefinition: public Declaration, public VariableScope, public Docum
 {
 public:
 	FunctionDefinition(Location const& _location, ASTPointer<ASTString> const& _name,
-					bool _isPublic,
-					bool _isConstructor,
+					Declaration::Visibility _visibility, bool _isConstructor,
```
CJentzsch: That is the only line where you have two values on one line.

PR #937, opened 2015-02-03
Solidity multiple arg sha3

LefterisJP:
It currently works for multiple arguments of the same type that can be converted to hash.
And here comes the question. Do we have, or if we don't have do we want to introduce, a FunctionType attribute that states it can take any arbitrary number of arguments of any type? Because that would be a much cleaner solution.

### DIFF
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -493,19 +493,9 @@ void FunctionCall::checkTypeRequirements()
 		if (m_names.empty())
 		{
 			for (size_t i = 0; i < m_arguments.size(); ++i)
-			{
-				if (functionType->getLocation() == FunctionType::Location::SHA3)
-				{
-#if 0 // are we sure we want that? Literal constant nums can't live outside storage and so sha3(42) will fail
-					if (!m_arguments[i]->getType()->canLiveOutsideStorage())
-						BOOST_THROW_EXCEPTION(createTypeError("SHA3 called with argument that can't live outside storage"));
-#endif
-					if (!m_arguments[i]->getType()->isImplicitlyConvertibleTo(*parameterTypes[0]))
-						BOOST_THROW_EXCEPTION(m_arguments[i]->createTypeError("SHA3 argument can't be converted to hash"));
-
-				} else if (!m_arguments[i]->getType()->isImplicitlyConvertibleTo(*parameterTypes[i]))
+				if (functionType->getLocation() != FunctionType::Location::SHA3 &&
+					!m_arguments[i]->getType()->isImplicitlyConvertibleTo(*parameterTypes[i]))
 					BOOST_THROW_EXCEPTION(createTypeError("Invalid type for argument in function call."));
-			}
 		}
 		else if (functionType->getLocation() == FunctionType::Location::SHA3)
 			BOOST_THROW_EXCEPTION(createTypeError("Named arguments can't be used for SHA3."));
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -487,7 +487,7 @@ void FunctionCall::checkTypeRequirements()
 		// and then ask if that is implicitly convertible to the struct represented by the
 		// function parameters
 		TypePointers const& parameterTypes = functionType->getParameterTypes();
-		if (functionType->getLocation() !=FunctionType::Location::SHA3 && parameterTypes.size() != m_arguments.size())
+		if (functionType->getLocation() != FunctionType::Location::SHA3 && parameterTypes.size() != m_arguments.size())
 			BOOST_THROW_EXCEPTION(createTypeError("Wrong argument count for function call."));
 
 		if (m_names.empty())  // LTODO: Totally ignoring sha3 case for named arguments for now just for the rebase to work
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -490,7 +490,7 @@ void FunctionCall::checkTypeRequirements()
 		if (functionType->getLocation() != FunctionType::Location::SHA3 && parameterTypes.size() != m_arguments.size())
 			BOOST_THROW_EXCEPTION(createTypeError("Wrong argument count for function call."));
 
-		if (m_names.empty())  // LTODO: Totally ignoring sha3 case for named arguments for now just for the rebase to work
+		if (m_names.empty())
 		{
 			for (size_t i = 0; i < m_arguments.size(); ++i)
 			{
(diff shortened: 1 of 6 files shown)


gavofyork:
Still needs a review?

libsolidity/AST.cpp
```diff
@@ -485,11 +485,24 @@ void FunctionCall::checkTypeRequirements()
...
 		TypePointers const& parameterTypes = functionType->getParameterTypes();
-		if (parameterTypes.size() != m_arguments.size())
+		if (functionType->getLocation() !=FunctionType::Location::SHA3 && parameterTypes.size() != m_arguments.size())
 			BOOST_THROW_EXCEPTION(createTypeError("Wrong argument count for function call."));
 		for (size_t i = 0; i < m_arguments.size(); ++i)
-			if (!m_arguments[i]->getType()->isImplicitlyConvertibleTo(*parameterTypes[i]))
+		{
+			if (functionType->getLocation() == FunctionType::Location::SHA3)
+			{
+#if 0 // are we sure we want that? Literal constant nums can't live outside storage and so sha3(42) will fail
```
chriseth: I would rather not convert everything to hash256. For example I think it is expected that `sha3("")` is the sha3 of the empty string and not `sha3(0)`. Everything should be copied to memory and take exactly "calldataEncodedSize" bytes. Note that there is a special case for IntegerConstants.

libsolidity/ExpressionCompiler.cpp
```diff
@@ -813,9 +816,9 @@ void ExpressionCompiler::appendExternalFunctionCall(FunctionType const& _functio
 	}
 }
 
-unsigned ExpressionCompiler::appendArgumentCopyToMemory(TypePointers const& _types,
-														vector<ASTPointer<Expression const>> const& _arguments,
-														unsigned _memoryOffset)
+unsigned ExpressionCompiler::appendArgumentsCopyToMemory(TypePointers const& _types,
```
chriseth: What about making the types optional and do not convert if they are not present?

gavofyork:
```
sha3("") == sha3(0x)
sha3(uint256(0)) == sha3(0x0000000000000000000000000000000000000000000000000000000000000000)
sha3(0) // ??? make literals default to uint256?
```

LefterisJP:
Rebased, had quite a few conflicts with the named args that are not properly fixed.

test/SolidityEndToEndTest.cpp
```diff
@@ -2095,6 +2095,25 @@ BOOST_AUTO_TEST_CASE(event_lots_of_data)
...
+		contract c {
+			// function foo(uint a) returns (hash d)
+			function foo(uint a, uint b, uint c) returns (hash d)
+			{
+				d = sha3(a, b, c);
+			}
+		})";
+	compileAndRun(sourceCode);
+
+	BOOST_CHECK(callContractFunction("foo(uint256,uint256,uint256)", 10 , 12, 13) == encodeArgs(
```
gavofyork: `10 ,` -> `10,`

gavofyork:
another few tests? include single args and args of multiple types.

LefterisJP:
yeah I would like to split it into 2 PRs actually. The args of multiple types could be its own PR. To make them work properly quite a bit more work is needed.

libsolidity/AST.cpp
```diff
@@ -487,14 +487,27 @@ void FunctionCall::checkTypeRequirements()
 		// and then ask if that is implicitly convertible to the struct represented by the
 		// function parameters
 		TypePointers const& parameterTypes = functionType->getParameterTypes();
-		if (parameterTypes.size() != m_arguments.size())
+		if (functionType->getLocation() != FunctionType::Location::SHA3 && parameterTypes.size() != m_arguments.size())
 			BOOST_THROW_EXCEPTION(createTypeError("Wrong argument count for function call."));
 
-		if (m_names.empty())
+		if (m_names.empty())  // LTODO: Totally ignoring sha3 case for named arguments for now just for the rebase to work
```
chriseth: Throw an exception if named arguments are used for sha3, it does not make sense.

libsolidity/AST.cpp
```diff
@@ -487,14 +487,27 @@ void FunctionCall::checkTypeRequirements()
...
-				if (!m_arguments[i]->getType()->isImplicitlyConvertibleTo(*parameterTypes[i]))
+			{
+				if (functionType->getLocation() == FunctionType::Location::SHA3)
+				{
+#if 0 // are we sure we want that? Literal constant nums can't live outside storage and so sha3(42) will fail
+					if (!m_arguments[i]->getType()->canLiveOutsideStorage())
+						BOOST_THROW_EXCEPTION(createTypeError("SHA3 called with argument that can't live outside storage"));
+#endif
+					if (!m_arguments[i]->getType()->isImplicitlyConvertibleTo(*parameterTypes[0]))
+						BOOST_THROW_EXCEPTION(createTypeError(std::string("SHA3 argument ") +
```
chriseth: Use `m_arguments[i]->createTypeError(...)`, then you don't have to count.

libsolidity/Types.h
```diff
@@ -130,6 +130,8 @@ class Type: private boost::noncopyable, public std::enable_shared_from_this<Type
 	/// i.e. it behaves differently in lvalue context and in value context.
 	virtual bool isValueType() const { return false; }
 	virtual unsigned getSizeOnStack() const { return 1; }
+	/// @returns the real type of some types, like e.g: IntegerConstant
+	virtual TypePointer getRealType() const { return TypePointer(); }
```
chriseth: Shouldn't this return `shared_from_this()`?

libsolidity/Types.cpp
```diff
@@ -361,6 +361,14 @@ u256 IntegerConstantType::literalValue(Literal const* _literal) const
 	return value;
 }
 
+TypePointer IntegerConstantType::getRealType() const
+{
+	auto intType = getIntegerType();
+	if (!intType)
```
chriseth: Any particular reason not to use solAssert?

libsolidity/Types.h
```diff
@@ -175,6 +177,7 @@ class IntegerType: public Type
 	virtual MemberList const& getMembers() const { return isAddress() ? AddressMemberList : EmptyMemberList; }
 
 	virtual std::string toString() const override;
+	virtual TypePointer getRealType() const { return std::make_shared<IntegerType>(m_bits, m_modifier); }
```
chriseth: No need to override anymore (also in other classes below).

chriseth:
Removing the overrides can be done in the next PR.

PR #940, opened 2015-02-03
Fixed calculatiion of new block hash after mine

arkpar:


arkpar:
Test failing due to other issues

PR #941, opened 2015-02-04
mix: Mining and transactions on a real blockchain

arkpar:


gavofyork:
`complex_accessors` test still failing.

PR #956, opened 2015-02-05
Optional parameter names.

LianaHus:
Now the names of input/return parameters of function do not have to be specified.

- added appropriate tests

### DIFF
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -60,7 +60,6 @@ void ContractDefinition::checkTypeRequirements()
 
 	FunctionDefinition const* fallbackFunction = nullptr;
 	for (ASTPointer<FunctionDefinition> const& function: getDefinedFunctions())
-	{
 		if (function->getName().empty())
 		{
 			if (fallbackFunction)
@@ -72,8 +71,6 @@ void ContractDefinition::checkTypeRequirements()
 					BOOST_THROW_EXCEPTION(fallbackFunction->getParameterList().createTypeError("Fallback function cannot take parameters."));
 			}
 		}
-	}
-
 	for (ASTPointer<ModifierDefinition> const& modifier: getFunctionModifiers())
 		modifier->checkTypeRequirements();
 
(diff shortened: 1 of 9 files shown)


libsolidity/AST.cpp
```diff
@@ -58,10 +58,21 @@ void ContractDefinition::checkTypeRequirements()
 		BOOST_THROW_EXCEPTION(constructor->getReturnParameterList()->createTypeError(
 									"Non-empty \"returns\" directive for constructor."));
 
-	FunctionDefinition const* fallbackFunction = getFallbackFunction();
-	if (fallbackFunction && fallbackFunction->getScope() == this && !fallbackFunction->getParameters().empty())
-		BOOST_THROW_EXCEPTION(fallbackFunction->getParameterList().createTypeError(
-									"Fallback function cannot take parameters."));
+	FunctionDefinition const* fallbackFunction = nullptr;
+	for (ASTPointer<FunctionDefinition> const& function: getDefinedFunctions())
+	{
```
gavofyork: remove unneeded `{`s.

gavofyork:
looks good apart from minor style alteration.

libsolidity/Parser.cpp
```diff
@@ -268,17 +268,28 @@ ASTPointer<VariableDeclaration> Parser::parseVariableDeclaration(VarDeclParserOp
...
 		isIndexed = true;
 		m_scanner->next();
 	}
-	Declaration::Visibility visibility(Declaration::Visibility::DEFAULT);
-	if (_options.isStateVariable && Token::isVisibilitySpecifier(token))
-		visibility = parseVisibilitySpecifier(token);
+	if (_options.allowEmptyName && m_scanner->getCurrentToken() != Token::IDENTIFIER)
+	{
+		identifier = make_shared<ASTString>("");
+		nodeFactory.setEndPositionFromNode(type);
```
chriseth: Unfortunately, this does not set the end position correctly... Yes, it is a really crappy mechanism.
I think we have to call it multiple times: Once `setEndPositionFromNode(type)` right after parsing the type and then `markEndPosition` at the beginning of the two if bodies (for visibility and indexed).

gavofyork:
All parameters should be named when it's non-private; otherwise the JS side of the API will get confused.

gavofyork:
also no compile on os x....

chriseth:
@gavofyork Is that a temporary restriction, i.e. is it planned to add this feature to the JS side or is it impossible?
Also, is it ok to have anonymous return parameters or should parameters and return parameters be named?

chriseth:
Seems to need a rebase again...

PR #953, opened 2015-02-05
Solidity ether subdenominations

LefterisJP:

PR #962, opened 2015-02-06
fix libsolidity's warning on Mac

guanqun:
/Users/q/Repos/cpp-ethereum/libsolidity/Parser.cpp:188:11: warning: variable 'visibility' is used uninitialized whenever 'if'
          condition is false [-Wsometimes-uninitialized]
            else if (_token == Token::PRIVATE)
                     ^~~~~~~~~~~~~~~~~~~~~~~~
    /Users/q/Repos/cpp-ethereum/libsolidity/Parser.cpp:193:9: note: uninitialized use occurs here
            return visibility;
                   ^~~~~~~~~~
    /Users/q/Repos/cpp-ethereum/libsolidity/Parser.cpp:188:7: note: remove the 'if' if its condition is always true
            else if (_token == Token::PRIVATE)
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    /Users/q/Repos/cpp-ethereum/libsolidity/Parser.cpp:183:2: note: variable 'visibility' is declared here
            Declaration::Visibility visibility;
            ^
    1 warning generated.

### DIFF
--- a/libsolidity/Parser.cpp
+++ b/libsolidity/Parser.cpp
@@ -180,7 +180,7 @@ ASTPointer<InheritanceSpecifier> Parser::parseInheritanceSpecifier()
 
 Declaration::Visibility Parser::parseVisibilitySpecifier(Token::Value _token)
 {
-	Declaration::Visibility visibility;
+	Declaration::Visibility visibility = Declaration::Visibility::DEFAULT;
 	if (_token == Token::PUBLIC)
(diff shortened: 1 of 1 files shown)


LefterisJP:
Thanks for spotting, but it should already be fixed and merged by this PR: https://github.com/ethereum/cpp-ethereum/pull/953

PR #965, opened 2015-02-06
natspec separated to it's own module. tests

debris:


libnatspec/NatspecExpressionEvaluator.cpp
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+#include <libdevcore/Log.h>
+#include <libdevcore/Exceptions.h>
+#include "NatspecExpressionEvaluator.h"
+
+using namespace std;
+using namespace dev;
+
+static QString contentsOfQResource(string const& res)
```
LefterisJP: `_res`

libnatspec/NatspecExpressionEvaluator.h
```diff
@@ -0,0 +1,35 @@
...
+ */
+/** @file NatspecExpressionEvaluator.h
+ * @author Marek Kotewicz <marek@ethdev.com>
+ * @date 2015
+ */
+
+#include <QtCore/QObject>
+#include <QtCore/QtCore>
+#include <QtQml/QJSEngine>
+
```
LefterisJP: MAybe a small docstring in the class and especially in its constructor explaining the arguments? I know what they are but for someone new looking at the code that would look like black magic :)

test/natspec.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <libdevcore/Log.h>
+#include <libnatspec/NatspecExpressionEvaluator.h>
+
+using namespace std;
+
+BOOST_AUTO_TEST_SUITE(natspec)
+
+BOOST_AUTO_TEST_CASE(natspec_eval_function_exists)
+{
+	cnote << "testing existance of evaluateExpression function";
```
LefterisJP: Maybe it would be better here (and in the other places below) if you did not output these messages. We should not clutter the test output.

test/natspec.cpp
```diff
@@ -0,0 +1,137 @@
...
+
```
LefterisJP: there is no reason to have spaces in between all those statements. Here and in all the test cases below please remove the extra spaces.

libnatspec/NatspecExpressionEvaluator.h
debris: sure

debris: done

PR #978, opened 2015-02-08
add exponent operator

guanqun:
This finishes this story:
https://www.pivotaltracker.com/n/projects/1189488/stories/83746404

### DIFF
--- a/libsolidity/Types.cpp
+++ b/libsolidity/Types.cpp
@@ -320,6 +320,16 @@ TypePointer IntegerConstantType::binaryOperatorResult(Token::Value _operator, Ty
 				return TypePointer();
 			value = m_value % other.m_value;
 			break;
+		case Token::Exp:
+			if (other.m_value < 0)
+				BOOST_THROW_EXCEPTION(InternalCompilerError() << errinfo_comment("exponent can't be negative"));
+			else
+			{
+				value = boost::multiprecision::powm(m_value, other.m_value, bigint(2) << 256);
+				if (value >= (bigint(1) << 256))
+					BOOST_THROW_EXCEPTION(InternalCompilerError() << errinfo_comment("exp result overflowed"));
+			}
+			break;
 		default:
 			return TypePointer();
 		}
--- a/libsolidity/Types.cpp
+++ b/libsolidity/Types.cpp
@@ -26,6 +26,8 @@
 #include <libsolidity/Types.h>
 #include <libsolidity/AST.h>
 
+#include <limits>
+
 using namespace std;
 
 namespace dev
@@ -322,13 +324,11 @@ TypePointer IntegerConstantType::binaryOperatorResult(Token::Value _operator, Ty
 			break;
 		case Token::Exp:
 			if (other.m_value < 0)
-				BOOST_THROW_EXCEPTION(InternalCompilerError() << errinfo_comment("exponent can't be negative"));
+				return TypePointer();
+			else if (other.m_value > std::numeric_limits<unsigned int>::max())
+				return TypePointer();
 			else
-			{
-				value = boost::multiprecision::powm(m_value, other.m_value, bigint(2) << 256);
-				if (value >= (bigint(1) << 256))
-					BOOST_THROW_EXCEPTION(InternalCompilerError() << errinfo_comment("exp result overflowed"));
-			}
+				value = boost::multiprecision::pow(m_value, other.m_value.convert_to<unsigned int>());
 			break;
 		default:
 			return TypePointer();
--- a/test/SolidityEndToEndTest.cpp
+++ b/test/SolidityEndToEndTest.cpp
@@ -80,7 +80,7 @@ BOOST_AUTO_TEST_CASE(exp_operator_const_signed)
 {
 	char const* sourceCode = R"(
 		contract test {
-			function f() returns(int d) { return -2 ** 3; }
+			function f() returns(int d) { return (-2) ** 3; }
 		})";
 	compileAndRun(sourceCode);
 	BOOST_CHECK(callContractFunction("f()", bytes()) == toBigEndian(u256(-8)));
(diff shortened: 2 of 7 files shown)


libsolidity/Types.cpp
```diff
@@ -323,6 +323,9 @@ TypePointer IntegerConstantType::binaryOperatorResult(Token::Value _operator, Ty
 				return TypePointer();
 			value = m_value % other.m_value;
 			break;
+		case Token::EXP:
+			value = boost::multiprecision::powm(m_value, other.m_value, bigint(2) << 256);
```
chriseth: I have seen that in other places here, but shouldn't it be `bigint(1) << 256`?

chriseth:
Nice and concise change :-)
Could you add some tests for signed types?
I think that we should disallow negative exponents, they should behave differently than expected.

libsolidity/Types.cpp
guanqun: I use the one in libevm, let me double check the number.

guanqun:
OK. I'll add the check for negative exponents.

libsolidity/Types.cpp
```diff
@@ -323,6 +323,16 @@ TypePointer IntegerConstantType::binaryOperatorResult(Token::Value _operator, Ty
 				return TypePointer();
 			value = m_value % other.m_value;
 			break;
+		case Token::EXP:
+			if (other.m_value < 0)
+				BOOST_THROW_EXCEPTION(InternalCompilerError() << errinfo_comment("exponent can't be negative"));
+			else
+			{
+				value = boost::multiprecision::powm(m_value, other.m_value, bigint(2) << 256);
```
guanqun: @chriseth How about this change? I explicitly check the overflow condition and the negative exponent.

guanqun:
Hi @chriseth, the negative exponent is only checked for constant integers, if the exponent is a variable, then it's not checked at all. I don't think we need to check that at this moment, as if we want to achieve that, more code has to be generated.

guanqun:
How do I add a "please review" label?

libsolidity/Types.cpp
chriseth: It is actually legal for integer constants to overflow during evaluation.
So in fact, we should not use `powm` at all, but rather ordinary pow.

test/SolidityEndToEndTest.cpp
```diff
@@ -56,6 +56,36 @@ BOOST_AUTO_TEST_CASE(empty_contract)
...
+		})";
+	compileAndRun(sourceCode);
+	BOOST_CHECK(callContractFunction("f()", bytes()) == toBigEndian(u256(8)));
+}
+
+BOOST_AUTO_TEST_CASE(exp_operator_const_signed)
+{
+	char const* sourceCode = R"(
+		contract test {
+			function f() returns(int d) { return -2 ** 3; }
```
chriseth: I am not 100% certain that this is parsed as `(-2) ** 3` here. Also to shield against such errors in the parser, please add parens.

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -974,6 +974,24 @@ BOOST_AUTO_TEST_CASE(overflow_caused_by_ether_units)
 	BOOST_CHECK_THROW(parseTextAndResolveNames(sourceCode), TypeError);
 }
 
+BOOST_AUTO_TEST_CASE(exp_operator_negative_exponent)
+{
+	char const* sourceCode = R"(
+		contract test {
+			function f() returns(uint d) { return 2 ** -3; }
+		})";
+	BOOST_CHECK_THROW(parseTextAndResolveNames(sourceCode), InternalCompilerError);
```
chriseth: `InternalCompilerError` always signify bugs in the compiler implementation, so please use TypeError here and below.

libsolidity/Types.cpp
```diff
@@ -323,6 +323,16 @@ TypePointer IntegerConstantType::binaryOperatorResult(Token::Value _operator, Ty
 				return TypePointer();
 			value = m_value % other.m_value;
 			break;
+		case Token::EXP:
+			if (other.m_value < 0)
+				BOOST_THROW_EXCEPTION(InternalCompilerError() << errinfo_comment("exponent can't be negative"));
```
chriseth: Please return a `TypeError()` here, as we do in the case of division by zero.

test/SolidityEndToEndTest.cpp
guanqun: Ah, you're right, I missed the parenthesis.

libsolidity/Types.cpp
guanqun: Sure, will fix that.

libsolidity/Types.cpp
guanqun: OK. then why during the execution of evm, it uses `powm` and it didn't check any overflow

chriseth: This is a special feature of integer literals in Solidity: https://github.com/ethereum/wiki/wiki/Solidity-Tutorial#integer-literals
According to the Solidity specification, you can do something like `unit8 x = 2**(2**(2**(2**10)))) * 0` without a type error (though the compiler will probably not be able to generate the code).

libsolidity/Types.cpp
```diff
@@ -320,6 +322,14 @@ TypePointer IntegerConstantType::binaryOperatorResult(Token::Value _operator, Ty
 				return TypePointer();
 			value = m_value % other.m_value;
 			break;
+		case Token::Exp:
+			if (other.m_value < 0)
+				return TypePointer();
+			else if (other.m_value > std::numeric_limits<unsigned int>::max())
+				return TypePointer();
+			else
+				value = boost::multiprecision::pow(m_value, other.m_value.convert_to<unsigned int>());
```
guanqun: Hi @chriseth, there's no `pow` function for both `big_int` and `big_int` in `boost::multiprecision`. So I resort to this approach, which is to check whether exponent can be fit into an unsigned int, if it can, `boost` has a function for it right away.

PR #979, opened 2015-02-08
windows fixes

debris:
##### fixed winows build
* `ambiguous std::array operator << ` unsigned replaced with [std::size_t](http://en.cppreference.com/w/cpp/types/size_t)
* serpent library being always included to `libweb3jsonrpc`
* missing `Boost_CHRONO` && `Boost_DATE_TIME` libraries


solc/CMakeLists.txt
```diff
@@ -13,6 +13,8 @@ add_executable(${EXECUTABLE} ${SRC_LIST} ${HEADERS})
 
 target_link_libraries(${EXECUTABLE} ${Boost_FILESYSTEM_LIBRARIES})
 target_link_libraries(${EXECUTABLE} ${Boost_PROGRAM_OPTIONS_LIBRARIES})
+target_link_libraries(${EXECUTABLE} ${Boost_DATE_TIME_LIBRARIES})
```
chriseth: Ugh. solc needs datetime and chrono?

chriseth:
cmake allows for transitive dependencies, does it? Wouldn't it be better to state the link dependency against boost chrono and date time at the place where it is needed?

solc/CMakeLists.txt
debris: with msvc yes :/

debris:
yes, cmake allows it. I will try to make it cleaner

PR #990, opened 2015-02-09
simplify ether unit with templates

guanqun:
I think it's easier to count the zeroes.

### DIFF
--- a/libethcore/CommonEth.cpp
+++ b/libethcore/CommonEth.cpp
@@ -35,27 +35,30 @@ namespace eth
 const unsigned c_protocolVersion = 53;
 const unsigned c_databaseVersion = 5;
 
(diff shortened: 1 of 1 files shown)


libethcore/CommonEth.cpp
```diff
@@ -35,27 +35,30 @@ namespace eth
 const unsigned c_protocolVersion = 53;
 const unsigned c_databaseVersion = 5;
 
+template <size_t n> constexpr u256 eth_unit() { return eth_unit<n-1>() * u256(1000); }
```
gavofyork: `eth_unit<n - 1>` (note the spaces :) )

guanqun: Eagle eye!

PR #996, opened 2015-02-10
Arbitrary parameters for call() and all hash functions.

chriseth:
This allows manual calls to other contracts. It contains something that might be considered a hack: If the first argument to `call` is a type of exactly four bytes, it is not padded to 32 bytes, allowing to use
`call(string4(string32("f(uint256,bool)")), 2, 3)` to call a specific ABI function.

### DIFF
--- a/libsolidity/ExpressionCompiler.cpp
+++ b/libsolidity/ExpressionCompiler.cpp
@@ -206,7 +206,7 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
 		TypePointers const& parameterTypes = function.getParameterTypes();
 		vector<ASTPointer<Expression const>> const& callArguments = _functionCall.getArguments();
 		vector<ASTPointer<ASTString>> const& callArgumentNames = _functionCall.getNames();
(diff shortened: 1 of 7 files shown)


libsolidity/ExpressionCompiler.cpp
```diff
@@ -713,8 +713,20 @@ void ExpressionCompiler::appendTypeConversion(Type const& _typeOnStack, Type con
...
+			// clear lower-order bytes for conversion to shorter strings - we always clean
 			// nothing to do, strings are high-order-bit-aligned
-			//@todo clear lower-order bytes if we allow explicit conversion to shorter strings
+			StaticStringType const& targetType = dynamic_cast<StaticStringType const&>(_targetType);
+			if (targetType.getNumBytes() < typeOnStack.getNumBytes())
+			{
+				if (targetType.getNumBytes() == 0)
+					m_context << eth::Instruction::DUP1 << eth::Instruction::XOR;
+				else
+				{
```
chriseth: Will remove braces.

PR #997, opened 2015-02-10
Arbitrary parameters for call() and all hash functions.

chriseth:
This allows manual calls to other contracts. It contains something that might be considered a hack: If the first argument to `call` is a type of exactly four bytes, it is not padded to 32 bytes, allowing to use
`call(string4(string32("f(uint256,bool)")), 2, 3)` to call a specific ABI function.

### DIFF
--- a/libsolidity/Types.h
+++ b/libsolidity/Types.h
@@ -367,14 +367,15 @@ public:
 	explicit FunctionType(VariableDeclaration const& _varDecl);
 	explicit FunctionType(EventDefinition const& _event);
 	FunctionType(strings const& _parameterTypes, strings const& _returnParameterTypes,
-				 Location _location = Location::Internal):
+				 Location _location = Location::Internal, bool _arbitraryParameters = false):
 		FunctionType(parseElementaryTypeVector(_parameterTypes), parseElementaryTypeVector(_returnParameterTypes),
-					 _location) {}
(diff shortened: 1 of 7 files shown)


libsolidity/Types.h
```diff
@@ -407,6 +408,9 @@ class FunctionType: public Type
 	/// Can contain a nullptr in which case indicates absence of documentation
 	ASTPointer<ASTString> getDocumentation() const;
 
+	/// true iff arguments are to be padded to multiples of 32 bytes for external calls
+	bool padArguments() const { return !(m_location == Location::SHA3 || m_location == Location::SHA256 || m_location == Location::RIPEMD160); }
```
LefterisJP: Small suggestion for predicate functions like this whose name could be considered ambiguous. Maybe change to `shouldPadArguments()` or `canPadArguments()`. If it was lisp we could just have an identifier with a question mark like `padArguments?()`

chriseth: I think it is a bit more subtle than this. The method will be called as `f.shouldPadArguments()` or `f.canPadArguments()`. In both variants, it looks like `f` would pad something, but actually it is the entity that wants to call a function of this type that has to pad the arguments.
Someone who knows the codebase also knows that type objects are immutable, so it is quite clear that it is a constant function.

PR #1002, opened 2015-02-10
MSVC constexpr fix.

chriseth:


libethcore/CommonEth.cpp
```diff
@@ -35,40 +35,23 @@ namespace eth
...
-template <> constexpr u256 exp10<0>()
-{
-	return u256(1);
-}
-
 vector<pair<u256, string>> const& units()
 {
-	static const vector<pair<u256, string>> s_units =
+	static vector<pair<u256, string>> s_units;
+	if (s_units.empty())
```
arkpar: Does not seem to be thread safe

chriseth: Uff, you are right.

PR #1007, opened 2015-02-11
Set genesis block difficulty to 2^22

Gustav-Simonsson:


CJentzsch:
I think we shoud wait until the genesis release with that. The genesis block will have to change anyways (pre sale...) and for the test net this difficulty would be just very high.

gavofyork:
YP changed - we'll probably set genesis according to outcome of poc-9 mining anyway.

Issue #1006, opened 2015-02-10
builds broke on macos


subtly:
Might be a duplicate! Is /usr/bin/cmake hard-coded? Which cmake file can I find these commands in?

/bin/sh: cmake: command not found
make: *** [/Users/USER/code/cpp-ethereum-build/CMakeFiles/BuildInfo.h] Error 127
Command /bin/sh failed with exit code 2

cd /Users/USER/code/cpp-ethereum-origin/libweb3jsonrpc && cmake -E rename webthreestubclient.h ../test/webthreestubclient.h
/bin/sh: cmake: command not found
make: *** [jsonrpcstub_buildpart_0] Error 127
Command /bin/sh failed with exit code 2

$ which cmake
/usr/local/bin/cmake

caktux:
Similar here, not sure why the builds passed though
```
Scanning dependencies of target BuildInfo.h
[  1%] [  1%] Unrecognized option or missing parameter, use --help.
make[2]: *** [jsonrpcstub] Error 255
make[1]: *** [libweb3jsonrpc/CMakeFiles/jsonrpcstub.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[  2%] Automatic moc for target sc
Automatic moc for target checkRandomTest
Built target secp256k1
[  2%] [  2%] Built target sc_automoc
Built target checkRandomTest_automoc
[  2%] Built target BuildInfo.h
[  7%] Built target devcore
make: *** [all] Error 2
```

caktux:
For some reason I had a lingering `jsonrpc3`, had to do `brew remove jsonrpc3` and `brew install jsonrpc` to clean things up... Note we'll be switching to the official Homebrew`libjson-rpc-cpp` soon.

PR #1008, opened 2015-02-11
cmake fixes

### LINKED ISSUE #1004: check spec.json diff before running jsonrpcstub

debris:
##### changes:
* fixed #1004 && #1006 
* removed calling bash `diff`, `mv`, `rm` from cmake


CJentzsch:
Fixed my issue with recompiling jsonrpcstub again and again.

Issue #994, opened 2015-02-10
Mining fails due to InvalidTimestamp


psalami:
After starting mining (with force mining enabled), a few blocks are mined correctly at first. After the first few blocks are mined, the following error occurs for any new blocks that my client mines:

```
*** [ 16:34:24 | miner-2 ] Completed 439b359a… 0d32ee86… 131582 1
--- [ 16:34:24 | miner-2 ] Completing mine!
*** [ 16:34:24 | miner-2 ] Mined 519c60e0… (parent: 4a4bd152… )
!!! [ 16:34:24 | eth ] Unexpected exception! Could not import block! /Users/psalami/git/cpp-ethereum-upstream/libethcore/BlockInfo.cpp(212): Throw in function void dev::eth::BlockInfo::verifyParent(const dev::eth::BlockInfo &) const
Dynamic exception type: boost::exception_detail::clone_impl<dev::eth::InvalidTimestamp>
std::exception::what: std::exception
[dev::tag_comment*] = Malformed block 
```

no more blocks can be mined after this starts happening. I'm assuming this is a side effect of 7b573c499b501ce20abcfd4829636ba72b52b619 (Move to PV53 along with timestamp being strictly > last time.)

psalami:
This may be a symptom of the difficulty being too low; I believe this issue primarily occurs when multiple mining threads are running and two local miners find successive blocks (perhaps a race condition). If another node finds a block after a local miner finds a block, the issue is avoided because the timestamp is reset. I was able to avoid the issue by reducing the number of local miners to 1 (see #1009).

gavofyork:
please check this on latest develop - it should be fixed.

psalami:
I wasn't able to reproduce the issue on the latest develop. Thanks!

PR #1009, opened 2015-02-11
CLI flag to set number of mining threads

psalami:
Added optional flag -t / --miners to eth and neth, which sets the number of CPU threads to use for mining (the default is the previous behavior of using all available CPU cores). By allowing the user to set only a single mining thread, the InvalidTimestamp exception can be avoided (see #994). It is also helpful to control the number of mining threads to manage CPU and energy usage and to aid in debugging.

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -59,7 +59,7 @@ void VersionChecker::setOk()
 	}
 }
 
-Client::Client(p2p::Host* _extNet, std::string const& _dbPath, bool _forceClean, u256 _networkId):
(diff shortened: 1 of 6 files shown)


libethereum/Client.cpp
```diff
@@ -69,7 +69,10 @@ Client::Client(p2p::Host* _extNet, std::string const& _dbPath, bool _forceClean,
 {
 	m_host = _extNet->registerCapability(new EthereumHost(m_bc, m_tq, m_bq, _networkId));
 
-	setMiningThreads();
+	if(miners > -1)
```
gavofyork: spacing `if (`

gavofyork:
minor style issue.

PR #1016, opened 2015-02-11
Conversion safety

CJentzsch:
The main problem is that <pre><code> bytes dev::fromHex(std::string const& _s) </pre></code> is just replacing bad hex characters with 0. So if someone has an invalid receiver address with a wrong hex character we will still execute the transaction, replacing this character with 0. FromHex is used quite often, I just changed one example. Should we throw and try/catch every instance of fromHex (I would say yes), or leave it as is? @gavofyork


alethzero/MainWin.cpp
```diff
@@ -1327,7 +1327,23 @@ void Main::ourAccountsRowsMoved()
 void Main::on_inject_triggered()
 {
 	QString s = QInputDialog::getText(this, "Inject Transaction", "Enter transaction dump in hex");
-	bytes b = fromHex(s.toStdString());
+	bytes b;
+	try
+	{
+		b = fromHex(s.toStdString());
+	}
+	catch(BadHexCharacter& _e)
```
gavofyork: space between `catch` and `(`

alethzero/MainWin.cpp
```diff
@@ -1327,7 +1327,23 @@ void Main::ourAccountsRowsMoved()
...
+	{
+		cnote << "invalid hex character, transaction rejected";
+		cnote << boost::diagnostic_information(_e);
+		return;
+	}
+	catch(...)
```
gavofyork: space again

gavofyork:
if you make `fromHex` throw, you need to check all other uses of it and guarantee no user initiated action can ever lead to an unhandled exception.

CJentzsch:
There are over 70 uses cases. And mostly it is just used as a small helper function. I can of course put a try/catch block around all of those, but here is another suggestion (see last commit). We add a bool parameter "_throw" to the function which is per dafault false, but can be switched to true for sensitive use cases (I will add more then the example above). @gavofyork

subtly:
We can't throw to javascript or C and this is a really basic method (like FixedHash). Would returning 0-length bytes work?

gavofyork:
Prefer dual-value enums over bool parameters.

CJentzsch:
I replaced fromHex with fromHex(..., throw) only for the most important use cases. Other cases such as informative output, er even secretkey input  (transaction just stays unsigned) is not touched.

PR #1028, opened 2015-02-12
Mix: deploy to network

arkpar:
1. Javascript code should now use "deploy" object instead of parent for contract, address, etc
2. Deploy to network command is added, which does the following:
 * Deploys a contract via json rpc connection to localhost:8080
 * Creates a subdirectory inside a project directory and copies frontend files there. Html code is modified so that 'deploy' object is configured for network contract address


arkpar:
Linux build failing due to unrelated issue

chriseth:
Rebuild triggered.

PR #1032, opened 2015-02-12
Copying structs.

chriseth:


libsolidity/ExpressionCompiler.cpp
```diff
@@ -1064,36 +1089,60 @@ void ExpressionCompiler::LValue::storeValue(Expression const& _expression, Type
...
-							dynamic_cast<ByteArrayType const&>(*_expression.getType()),
+							dynamic_cast<ByteArrayType const&>(*m_dataType),
 							dynamic_cast<ByteArrayType const&>(_sourceType));
-			else if (_expression.getType()->getCategory() == Type::Category::Struct)
+			else if (m_dataType->getCategory() == Type::Category::Struct)
 			{
-				//@todo
-				solAssert(false, "Struct copy not yet implemented.");
+				// stack layout: source_ref target_ref
+				auto const& structType = dynamic_cast<StructType const&>(*m_dataType);
```
CJentzsch: When you do a dynamic cast to StructType you do not need auto, because you know exactly which type it will be.
Can you explain why you need that dynamic cast?

CJentzsch:
Since the above remark is not a real issue, I mark it looksgood :)

libsolidity/ExpressionCompiler.cpp
chriseth: The auto is used to shorten the code -- as you say, we know exactly which type the auto will resolve to.
Admittedly, the benefit is not too large here.
Concerning the dynamic cast: A cast is needed because we call `getStorageOffsetOfMember` below and dynamic_cast is used so that a type problem is detected before `structType` is used incorrectly.

PR #1039, opened 2015-02-13
Performance test

CJentzsch:
This test takes about 58s without JIT, and 4 ms with JIT (impressive!). Thanks @chfast for providing the test.

### DIFF
--- a/test/vm.cpp
+++ b/test/vm.cpp
@@ -32,8 +32,6 @@ using namespace dev;
 using namespace dev::eth;
 using namespace dev::test;
 
-using Millisecs = chrono::duration<int,milli>;
-
 FakeExtVM::FakeExtVM(eth::BlockInfo const& _previousBlock, eth::BlockInfo const& _currentBlock, unsigned _depth):			/// TODO: XXX: remove the default argument & fix.
 	ExtVMFace(Address(), Address(), Address(), 0, 1, bytesConstRef(), bytes(), _previousBlock, _currentBlock, test::lastHashes(_currentBlock.number), _depth) {}
 
@@ -522,12 +520,12 @@ BOOST_AUTO_TEST_CASE(vmPerformanceTest)
 		string arg = boost::unit_test::framework::master_test_suite().argv[i];
 		if (arg == "--performance")
 		{
-			chrono::steady_clock::time_point start = chrono::steady_clock::now();
+			auto start = chrono::steady_clock::now();
 
 			dev::test::executeTests("vmPerformanceTest", "/VMTests", dev::test::doVMTests);
 
-			chrono::steady_clock::time_point end = chrono::steady_clock::now();
-			Millisecs duration(chrono::duration_cast<Millisecs>(end - start));
+			auto end = chrono::steady_clock::now();
+			chrono::milliseconds duration(chrono::duration_cast<chrono::milliseconds>(end - start));
 			cnote << "test duration: " << duration.count() << " milliseconds.\n";
 		}
 	}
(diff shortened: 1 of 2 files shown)


test/vm.cpp
```diff
@@ -32,6 +32,8 @@ using namespace dev;
 using namespace dev::eth;
 using namespace dev::test;
 
+using Millisecs = chrono::duration<int,milli>;
```
chfast: It is already predefined in `<chrono>` as `std::chrono::milliseconds`.

test/vm.cpp
```diff
@@ -513,13 +515,39 @@ BOOST_AUTO_TEST_CASE(vmSystemOperationsTest)
...
 }
 
+BOOST_AUTO_TEST_CASE(vmPerformanceTest)
+{
+	for (int i = 1; i < boost::unit_test::framework::master_test_suite().argc; ++i)
+	{
+		string arg = boost::unit_test::framework::master_test_suite().argv[i];
+		if (arg == "--performance")
+		{
+			chrono::steady_clock::time_point start = chrono::steady_clock::now();
```
chfast: I'm using high resolution clock for measuring timings.

test/vmPerformanceTestFiller.json
```diff
@@ -0,0 +1,38 @@
...
+            "currentGasLimit" : "100000000000",
+            "currentDifficulty" : "256",
+            "currentTimestamp" : "1",
+            "currentCoinbase" : "2adc25665018aa1fe0e6bc666dac8fc2697ff9ba"
+        },
+        "pre" : {
+            "0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6" : {
+                "balance" : "1000000000000000000",
+                "nonce" : "0",
+                "//" : "contract PerformanceTester {",
```
chfast: I think it is good idea to add the PerformanceTester.sol file to the repo. I want to add more functions there. Here we can reference just the function name.

CJentzsch: You can do that when you add your .sol file to the repo.

test/vm.cpp
```diff
@@ -513,13 +513,39 @@ BOOST_AUTO_TEST_CASE(vmSystemOperationsTest)
...
+	{
+		string arg = boost::unit_test::framework::master_test_suite().argv[i];
+		if (arg == "--performance")
+		{
+			auto start = chrono::steady_clock::now();
+
+			dev::test::executeTests("vmPerformanceTest", "/VMTests", dev::test::doVMTests);
+
+			auto end = chrono::steady_clock::now();
+			chrono::milliseconds duration(chrono::duration_cast<chrono::milliseconds>(end - start));
```
chfast: How about `auto duration = chrono::duration_cast<chrono::milliseconds>(end - start);`

CJentzsch: Why? It is casted to chrono::milliseconds, because that is the precision I wanted. So the type is clear, and its not so long that I would need a shortcut.

chfast: Why wouldn't you use the same style as line before?
Why want you to repeat the same type twice?
There are 2  forms you can express the same thing and you have chosen the more unclear one.

CJentzsch: I see your point, so you think it is more readable, and you can switch precision at on place, not two. (I just dont like auto too much, it feels so python :-)  ). For me its more readable when I see the type directly. But I will change it to auto in order to have only one place to touch when I would change precision.

gavofyork: CodingStandards.txt:6i (newly added) `auto` is preferable when type is clear.

PR #1042, opened 2015-02-13
mix - bug fix: storage view in debug pane broken for long values

yann300:
TODO: get the type of the stored variable from solidity in order to display the value properly.


alethzero/NatspecHandler.h
```diff
@@ -26,7 +26,8 @@
 #pragma warning(disable: 4100 4267)
 #include <leveldb/db.h>
 #pragma warning(pop)
-#include <json/json.h>
+//#include <json/json.h>
+#include <jsoncpp/json/json.h>
```
debris: please don't do that, it may break the build for others. #1038 should fix all issues with it. Please revert "#include <jsoncpp/json/json.h>"  and rebase to latest develop. Cleaning build directory may also be required.

yann300: Yes this was just a fix to get the build working, I will revert.

debris: ;)

mix/qml/Debugger.qml
```diff
@@ -598,9 +598,11 @@ Rectangle {
 										Layout.minimumWidth: parent.width / 2
 										Layout.maximumWidth: parent.width / 2
 										Text {
+											maximumLineCount: 1
+											clip: true
 											anchors.leftMargin: 5
 											width: parent.width - 5
-											wrapMode: Text.NoWrap
+											wrapMode: Text.WrapAnywhere
```
arkpar: why is it set to wrap?

PR #1043, opened 2015-02-13
Enums in Solidity

LefterisJP:


libsolidity/Parser.cpp
```diff
@@ -263,6 +266,39 @@ ASTPointer<StructDefinition> Parser::parseStructDefinition()
...
+	expectToken(Token::LBrace);
+
+	while (m_scanner->getCurrentToken() != Token::RBrace)
+	{
+		members.push_back(parseEnumDeclaration());
+		if (m_scanner->getCurrentToken() == Token::RBrace)
+			break;
+		expectToken(Token::Comma);
+		if (m_scanner->getCurrentToken() != Token::Identifier) {
+			BOOST_THROW_EXCEPTION(createParserError("Expected Identifier after ,"));
```
LianaHus: isn't "," better? or ',' ?

libsolidity/Token.h
```diff
@@ -67,27 +67,27 @@ namespace solidity
...
+																		\
+	/* Punctuators (ECMA-262, section 7.7, page 15). */				\
+	T(LParen, "(", 0)													\
+	T(RParen, ")", 0)													\
+	T(LBrack, "[", 0)													\
+	T(RBrack, "]", 0)													\
+	T(LBrace, "{", 0)													\
+	T(RBrace, "}", 0)													\
+	T(Colon, ":", 0)													\
+	T(Semicolon, ";", 0)												\
```
LianaHus: hmm, what's goin gon with spaces before "\" here?

LefterisJP: I was in the middle of changing them to tabs. What you see here is TABS. IF it shows nicely in github it's not TABS :P

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -992,6 +992,87 @@ BOOST_AUTO_TEST_CASE(exp_operator_exponent_too_big)
...
+
+BOOST_AUTO_TEST_CASE(enum_duplicate_values)
+{
+	char const* text = R"(
+			contract test {
+				enum ActionChoices { GoLeft, GoRight, GoLeft, Sit };
+				function test()
+				{
+                   a = 1;
+                   b = 2;
```
LianaHus: indentation

libsolidity/Parser.cpp
LefterisJP: yeah it would be

libsolidity/Types.cpp
```diff
@@ -924,6 +949,13 @@ MemberList const& TypeType::getMembers() const
 					if (!f->isConstructor() && !f->getName().empty())
 						members[f->getName()] = make_shared<FunctionType>(*f);
 		}
+		else if (m_actualType->getCategory() == Category::Enum)
+		{
+			EnumDefinition const& enumDef = dynamic_cast<EnumType const&>(*m_actualType).getEnumDefinition();
+			for (ASTPointer<EnumDeclaration> const& enumValue: enumDef.getMembers())
+			members.insert(make_pair(enumValue->getName(), make_shared<EnumType>(enumDef)));
+			m_members.reset(new MemberList(members));
```
LianaHus: do you forget braces  or Tabs?

libsolidity/AST.h
```diff
@@ -166,6 +166,21 @@ class Declaration: public ASTNode
 };
 
 /**
+ * Declaration of an Enum Value
+ */
+class EnumDeclaration : public Declaration
```
chriseth: space

libsolidity/AST.h
```diff
@@ -166,6 +166,21 @@ class Declaration: public ASTNode
...
+{
+  public:
+	EnumDeclaration(Location const& _location,
+					ASTPointer<ASTString> const& _name):
+	Declaration(_location, _name) {}
```
chriseth: indent

libsolidity/AST.h
```diff
@@ -166,6 +166,21 @@ class Declaration: public ASTNode
 };
 
 /**
+ * Declaration of an Enum Value
```
chriseth: Could you move that right above or below EnumDefinition? The classes around here are quite general beasts.

chriseth:
Please also update `grammar.txt`.

libsolidity/AST.cpp
```diff
@@ -206,6 +206,12 @@ vector<pair<FixedHash<4>, FunctionTypePointer>> const& ContractDefinition::getIn
 	return *m_interfaceFunctionList;
 }
 
+TypePointer EnumValue::getType(ContractDefinition const*) const
+{
+	EnumDefinition const* parentDef = dynamic_cast<EnumDefinition const*>(getScope());
```
chriseth: Please either use `dynamic_cast<EnumDefinition const&>` or `solAssert(parentDef, "");`.

libsolidity/AST.cpp
```diff
@@ -255,6 +261,37 @@ void StructDefinition::checkRecursion() const
...
+{
+	vector<ASTPointer<EnumValue>> members(getMembers());
+	auto compareDecls = [](ASTPointer<EnumValue> a, ASTPointer<EnumValue> b)
+	{
+		return a->getName() < b->getName();
+	};
+	sort(begin(members), end(members), compareDecls);
+	for (size_t i = 0; i < members.size() - 1; ++i)
+		if (members[i]->getName() == members[i + 1]->getName())
+			BOOST_THROW_EXCEPTION(createTypeError("Duplicate member detected in Enum"));
```
chriseth: Please create the type error on one of the members (so that location is set correctly).

chriseth: Hm, isn't that already detected in `DeclarationRegistrationHelper::registerDeclaration`?

libsolidity/AST.h
```diff
@@ -315,6 +334,31 @@ class StructDefinition: public Declaration
...
+		Declaration(_location, _name), m_members(_members) {}
+	virtual void accept(ASTVisitor& _visitor) override;
+	virtual void accept(ASTConstVisitor& _visitor) const override;
+
+	std::vector<ASTPointer<EnumValue>> const& getMembers() const { return m_members; }
+
+	virtual TypePointer getType(ContractDefinition const*) const override;
+
+	/// @returns the value that the string has in the Enum
+	unsigned int getMemberValue(ASTString const& _member) const;
```
chriseth: This is non-syntactic information, can you move that to the type?
Perhaps such a function should even be available in the class `MemberList` - as we noticed that the list of members should really be a list, it makes sense to assign indices to the members.

libsolidity/ExpressionCompiler.cpp
```diff
@@ -489,19 +489,36 @@ void ExpressionCompiler::endVisit(MemberAccess const& _memberAccess)
...
+		EnumDefinition const& enumDef = type.getEnumDefinition();
+		m_context << enumDef.getMemberValue(_memberAccess.getMemberName());
+		break;
+	}
 	case Type::Category::TypeType:
 	{
 		TypeType const& type = dynamic_cast<TypeType const&>(*_memberAccess.getExpression().getType());
+		ContractType const* contractType;
+		EnumType const* enumType;
 		if (type.getMembers().getMemberType(member))
```
chriseth: Perhaps negate the condition and throw here to avoid long indentation?

libsolidity/ExpressionCompiler.cpp
```diff
@@ -489,19 +489,36 @@ void ExpressionCompiler::endVisit(MemberAccess const& _memberAccess)
...
 		{
-			ContractDefinition const& contract = dynamic_cast<ContractType const&>(*type.getActualType())
-													.getContractDefinition();
-			for (ASTPointer<FunctionDefinition> const& function: contract.getDefinedFunctions())
-				if (function->getName() == member)
-				{
-					m_context << m_context.getFunctionEntryLabel(*function).pushTag();
-					return;
-				}
+			if ((contractType = dynamic_cast<ContractType const*>(type.getActualType().get())))
```
chriseth: Move the declaration of `contractType` here?

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -992,6 +992,87 @@ BOOST_AUTO_TEST_CASE(exp_operator_exponent_too_big)
...
+					a = ActionChoices.GoStraight;
+					b = ActionChoices.Sit;
+				}
+				uint256 a;
+				uint64 b;
+			}
+	)";
+	BOOST_CHECK_THROW(parseTextAndResolveNames(text), TypeError);
+}
+
```
chriseth: Explicit conversion to enum should also work.

libsolidity/Token.h
```diff
@@ -67,31 +67,31 @@ namespace solidity
 
 #define IGNORE_TOKEN(name, string, precedence)
 
-#define TOKEN_LIST(T, K)                                               \
-	/* End of source indicator. */                                     \
-	T(EOS, "EOS", 0)                                                   \
+#define TOKEN_LIST(T, K)												\
```
chriseth: Were all these changes on purpose?

libsolidity/Parser.cpp
```diff
@@ -263,6 +266,39 @@ ASTPointer<StructDefinition> Parser::parseStructDefinition()
 	return nodeFactory.createNode<StructDefinition>(name, members);
 }
 
+ASTPointer<EnumValue> Parser::parseEnumDeclaration()
```
chriseth: Your auto-renamer did not rename this :-)

libsolidity/Parser.cpp
```diff
@@ -263,6 +266,39 @@ ASTPointer<StructDefinition> Parser::parseStructDefinition()
...
+{
+	ASTNodeFactory nodeFactory(*this);
+	ASTPointer<ASTString> name = expectIdentifierToken();
+	nodeFactory.markEndPosition();
```
chriseth: It should be
```
ASTNodeFactory nodeFactory(*this);
nodeFactory.markEndPosition();
return nodeFactory.createNode<EnumValue>(expectIdentifierToken());
```
I am *really* sorry for creating such a crappy mechanism.

libsolidity/Parser.cpp
```diff
@@ -263,6 +266,39 @@ ASTPointer<StructDefinition> Parser::parseStructDefinition()
...
+	vector<ASTPointer<EnumValue>> members;
+	expectToken(Token::LBrace);
+
+	while (m_scanner->getCurrentToken() != Token::RBrace)
+	{
+		members.push_back(parseEnumDeclaration());
+		if (m_scanner->getCurrentToken() == Token::RBrace)
+			break;
+		expectToken(Token::Comma);
+		if (m_scanner->getCurrentToken() != Token::Identifier) {
```
chriseth: nobraces

libsolidity/Parser.cpp
```diff
@@ -263,6 +266,39 @@ ASTPointer<StructDefinition> Parser::parseStructDefinition()
...
+			BOOST_THROW_EXCEPTION(createParserError("Expected Identifier after ','"));
+		}
+	}
+
+	nodeFactory.markEndPosition();
+	expectToken(Token::RBrace);
+	expectToken(Token::Semicolon);
```
chriseth: Why the semicolon? It is not required by any other element that requires an explicit `}`.

libsolidity/Parser.cpp
```diff
@@ -823,6 +859,13 @@ ASTPointer<ASTString> Parser::expectIdentifierToken()
 	return getLiteralAndAdvance();
 }
 
+ASTPointer<ASTString> Parser::peekIdentifierToken()
```
chriseth: Seems to be unused.

libsolidity/Types.h
```diff
@@ -369,6 +368,30 @@ class StructType: public Type
...
+ * The type of an enum instance, there is one distinct type per enum definition.
+ */
+class EnumType: public Type
+{
+public:
+	virtual Category getCategory() const override { return Category::Enum; }
+	explicit EnumType(EnumDefinition const& _enum): m_enum(_enum) {}
+	virtual TypePointer unaryOperatorResult(Token::Value _operator) const override;
+	virtual bool operator==(Type const& _other) const override;
+	virtual unsigned getSizeOnStack() const override { return 1; /*@todo*/ }
```
chriseth: `1` is always correct. We *could* explicitly force the number of members to be less than 2^256, though...

libsolidity/Types.h
```diff
@@ -369,6 +368,30 @@ class StructType: public Type
...
+	virtual std::string toString() const override;
+	virtual bool isValueType() const override { return true; }
+
+	virtual bool isExplicitlyConvertibleTo(Type const& _convertTo) const override;
+
+	EnumDefinition const& getEnumDefinition() const { return m_enum; }
+
+private:
+	EnumDefinition const& m_enum;
+	/// List of member types, will be lazy-initialized because of recursive references.
```
chriseth: I think this is not true for enums, so you might go without a unique_ptr.

libsolidity/Types.cpp
```diff
@@ -924,6 +949,13 @@ MemberList const& TypeType::getMembers() const
 					if (!f->isConstructor() && !f->getName().empty())
 						members[f->getName()] = make_shared<FunctionType>(*f);
 		}
+		else if (m_actualType->getCategory() == Category::Enum)
+		{
+			EnumDefinition const& enumDef = dynamic_cast<EnumType const&>(*m_actualType).getEnumDefinition();
+			for (ASTPointer<EnumValue> const& enumValue: enumDef.getMembers())
+				members.insert(make_pair(enumValue->getName(), make_shared<EnumType>(enumDef)));
```
chriseth: You could reuse the shared ptr.

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -992,6 +992,87 @@ BOOST_AUTO_TEST_CASE(exp_operator_exponent_too_big)
...
+BOOST_AUTO_TEST_CASE(enum_duplicate_values)
+{
+	char const* text = R"(
+			contract test {
+				enum ActionChoices { GoLeft, GoRight, GoLeft, Sit };
+				function test()
```
chriseth: The test code can end here.

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -992,6 +992,87 @@ BOOST_AUTO_TEST_CASE(exp_operator_exponent_too_big)
...
+
+BOOST_AUTO_TEST_CASE(enum_implicit_conversion_is_not_okay)
+{
+	char const* text = R"(
+			contract test {
+				enum ActionChoices { GoLeft, GoRight, GoStraight, Sit };
+				function test()
+				{
+					a = ActionChoices.GoStraight;
+					b = ActionChoices.Sit;
```
chriseth: Please only test conversion to uint256 as it is least strict.

test/SolidityParser.cpp
```diff
@@ -703,6 +703,48 @@ BOOST_AUTO_TEST_CASE(literal_constants_with_ether_subdenominations_in_expression
...
 }
 
+BOOST_AUTO_TEST_CASE(enum_declaration)
+{
+	char const* text = R"(
+		contract c {
+			enum foo { WARNING, NOTICE, ERROR, CRITICAL };
+			function c ()
+			{
+				a = foo.CRITICAL;
```
chriseth: Could suggest that this is valid code...

test/SolidityParser.cpp
```diff
@@ -703,6 +703,48 @@ BOOST_AUTO_TEST_CASE(literal_constants_with_ether_subdenominations_in_expression
...
+		})";
+	BOOST_CHECK_NO_THROW(parseTextExplainError(text));
+}
+
+BOOST_AUTO_TEST_CASE(empty_enum_declaration)
+{
+	char const* text = R"(
+		contract c {
+			enum foo { };
+			function c ()
```
chriseth: Also here (and below): The test should test exactly one thing (unless this makes it really hard to write).

libsolidity/Parser.cpp
LefterisJP: Oh come on! Let's be fair, it's not the same symbol :). On it now.

libsolidity/Parser.cpp
LefterisJP: Don't worry I have my fair share of crappy mechanisms under my belt too. We can try to see if we can make it a bit more intuitive in a small PR.

libsolidity/Parser.cpp
LefterisJP: https://www.pivotaltracker.com/story/show/86670106

libsolidity/Parser.cpp
LefterisJP: oh yes it is! Thought I had deleted it.

libsolidity/Types.h
LefterisJP: yeah, I know it is always correct. Just forgot to remove the @todo. Copied the signature of the Type class.

test/SolidityNameAndTypeResolution.cpp
LefterisJP: But we should be able to convert to any integer type. So why not have a test for at least one more type other than uint256?

test/SolidityParser.cpp
LefterisJP: hehe, yeah. Will make the enum Values a bit more innocent.

libsolidity/AST.h
LefterisJP: Yeah sure. Did not know we had that distinction.

libsolidity/ExpressionCompiler.cpp
LefterisJP: I see what you are trying to save by suggesting this but it does not seem to compile. A pity. Would have been nice to avoid having to declare type of contractType (and later enumType) above.

libsolidity/AST.h
chriseth: Actually I am not happy about all the type checking being part of the AST - at some point I would like to move it to its own class. And also the type system has too much "implementation logic", it would be better if it only had abstract type information.

test/SolidityNameAndTypeResolution.cpp
LefterisJP: Hmm .. okay. I am adding that too. But there is a question here. So if I have an Enum with 5 values and I try to convert an int literal with a value of 64. Then we can detect that something is gonna be wrong and throw a compile error. 

But what if it's an integer type whose value is only known at compile time? What do we do then?

chriseth: I think we ignore both problems and trust in the author that she knows what she does.

libsolidity/Parser.cpp
chriseth: Sure, but consistency is more important I would say.

test/SolidityNameAndTypeResolution.cpp
LefterisJP: I could add a small PR with a warning for the literal at least. Should be easy to do.

PR #1044, opened 2015-02-13
Mix: reordered import statements to work around macdeployqt bug

arkpar:


debris:
required cleaning build directory, but it's working :+1:

PR #1046, opened 2015-02-13
bug fix: storage view in debug pane broken for long values.

yann300:
TODO: 
Get types from solidity and display right value according to type.

### DIFF
--- a/libethcore/CommonJS.cpp
+++ b/libethcore/CommonJS.cpp
@@ -117,9 +117,9 @@ std::string prettyU256(u256 _n, bool _abridged)
 			n = toHex(a.ref());
 
 		if (n.empty())
-			s << "0x" << a;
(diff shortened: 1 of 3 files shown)


libethcore/CommonJS.cpp
```diff
@@ -110,11 +110,16 @@ std::string prettyU256(u256 _n)
...
+		if (_abridged)
+			n =  a.abridged();
+		else
+			n = toHex(a.ref());
+
 		if (n.empty())
 			s << "0x" << a;
 		else
-			s << n << "(0x" << a.abridged() << ")";
+			s << n << "(0x" << n << ")";
```
arkpar: Does not make much sense to display the same value twice. I'd suggest displaying 0x0 for zero address, hex value otherwise

PR #1050, opened 2015-02-13
update windows precompiled qt version 5.3 -> 5.4

debris:
Downloading qt-5.4 from [here](http://build.ethdev.com/builds/windows-precompiled/). Mix working on windows.


gavofyork:
If necessary, could we get the windows building wiki updated accordingly?

debris:
It's not necessary.

PR #1056, opened 2015-02-14
removed libqwebthree

debris:
**changes:**
* removed libqwebthree
* AlethZero is using jsonrpc::HttpServer instead of QWebThreeConnector
* HttpServer is on port 8080


gavofyork:
awesome :) didn't particularly want to do this myself. is there any special requirement for dapps or will it work transparently with existing?

debris:
It will work :)

PR #1068, opened 2015-02-16
ManyFunctions performance test

chfast:
~200 functions in a contract, half of them are called randomly (seed is an argument)


test/ManyFunctions.sol
```diff
@@ -0,0 +1,1485 @@
+
+    function right1(uint seed) returns (uint) {
+        var r = nextRand(seed);
+        if (r >= 2^79)
```
chriseth: Note that `**` is exponentiation, `^` is exclusive or.

chfast: Oops :)

PR #1071, opened 2015-02-16
Test Correction

winsvega:
etherpad check

### DIFF
--- a/test/ttTransactionTestFiller.json
+++ b/test/ttTransactionTestFiller.json
@@ -241,21 +241,6 @@
         }
     },
 
-    "WrongAddress" : {
-        "transaction" :
-        {
-            "data" : "",
-            "gasLimit" : "",
-            "gasPrice" : "",
-            "nonce" : "",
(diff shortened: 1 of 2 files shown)


test/ttTransactionTestFiller.json
```diff
@@ -241,21 +241,6 @@
         }
     },
 
-    "WrongAddress" : {
```
CJentzsch: Why do you remove this testFiller?

winsvega: because when we try to read this wrong address from json filler, an exception is thrown not producing a test rlp and it makes a failure
instead I created wrong transaction rlps manually in ttWrongRLPTransaction.json




Четверг, 19 февраля 2015, 4:27 -08:00 от Christoph Jentzsch <notifications@github.com>:
>In  test/ttTransactionTestFiller.json :
>> @@ -241,21 +241,6 @@
>          }
>      },
>  
> -    "WrongAddress" : {
>Why do you remove this testFiller?
>—
>Reply to this email directly or  view it on GitHub .


--------------
Wins Vega Wins

PR #1074, opened 2015-02-16
Mix: Allow more than one contract

arkpar:
TODO: Contract dependencies (deployment in correct order)


mix/ClientModel.cpp
```diff
@@ -181,7 +188,7 @@ void ClientModel::setupState(QVariantMap _state)
 				transactionSettings.parameterValues.push_back(param);
 			}
 
-			if (transaction.value("executeConstructor").toBool())
+			if (contractId == functionId || functionId == "Constructor")
```
yann300: Maybe it's possible to continue using transaction.value("executeConstructor").toBool()
rather than : functionId == "Constructor" as this label "Constructor" will certainly be changed.

mix/ClientModel.cpp
```diff
@@ -172,8 +177,10 @@ void ClientModel::setupState(QVariantMap _state)
...
 		else
 		{
+			if (contractId.isEmpty() && m_context->codeModel()->hasContract()) //TODO: This is to support old project files, remove later
+				contractId = m_context->codeModel()->contracts().keys()[0];
 			QVariantList qParams = transaction.value("qType").toList();
-			TransactionSettings transactionSettings(functionId, value, gas, gasPrice);
+			TransactionSettings transactionSettings(contractId, functionId, value, gas, gasPrice);
 
 			for (QVariant const& variant: qParams)
 			{
```
yann300: brace?

mix/ClientModel.cpp
arkpar: The "Constructor" label is no more. This code is to support old project files, will be removed once everyone upgrades

mix/ClientModel.cpp
arkpar: There are two lines in the block, github hides the second

PR #1082, opened 2015-02-17
latest changes from ethereum.js 

debris:
* version `0.0.13` -> `0.0.15`
* added transactionCount && uncleCount methods
* updated Readme
* more testable code
* tests
* coveralls.io integration (not run by default)
* gulpfile.js refactored


libjsqrc/ethereumjs/lib/watches.js
```diff
@@ -0,0 +1,49 @@
...
+    { name: 'getMessages', call: 'eth_filterLogs' }
+    ];
+};
+
+/// @returns an array of objects describing web3.shh.watch api methods
+var shh = function () {
+    return [
+    { name: 'newFilter', call: 'shh_newFilter' },
+    { name: 'uninstallFilter', call: 'shh_uninstallFilter' },
+    { name: 'getMessages', call: 'shh_getMessages' }
```
gavofyork: should be `messages`, consistent with the naming of other watchables, `logs` and `events`.

gavofyork:
rename required as discussed.

libjsqrc/ethereumjs/lib/watches.js
debris: these methods are used only internally by filter.js and are not publicly visible.

Methods that are exposed can be found [here](https://github.com/debris/cpp-ethereum/blob/jsonrpc/libjsqrc/ethereumjs/lib/filter.js#L102-L109).
```javascript
  return {
        changed: changed,
        arrived: changed,
        happened: changed,
        messages: messages,
        logs: messages,
        uninstall: uninstall
    };
```

PR #1086, opened 2015-02-18
p2p: rlpx transport

subtly:
Request to review what the code does.
Known: PeerHandshake needs to be classy moved into an implementation file.


libdevcrypto/Common.cpp
```diff
@@ -82,6 +82,22 @@ bool dev::decrypt(Secret const& _k, bytesConstRef _cipher, bytes& o_plaintext)
 	return true;
 }
 
+void dev::encryptECIES(Public const& _k, bytesConstRef _plain, bytes& o_cipher)
```
chriseth: Why not return `bytes`?

subtly: Because decrypt does not return bytes.

libdevcrypto/Common.h
```diff
@@ -96,6 +96,21 @@ void encryptSym(Secret const& _k, bytesConstRef _plain, bytes& o_cipher);
...
+void encryptECIES(Public const& _k, bytesConstRef _plain, bytes& o_cipher);
+	
+/// Decrypt payload using ECIES standard with AES-CTR. TODO: move into class.
+bool decryptECIES(Secret const& _k, bytesConstRef _cipher, bytes& o_plaintext);
+	
+/// Encrypts payload with random IV using AES-CTR. TODO: prefix IV.
+h128 encryptSymNoAuth(Secret const& _k, bytesConstRef _plain, bytes& o_cipher);
+
+/// Encrypts payload with specified IV using AES-CTR TODO: prefix IV.
+h128 encryptSymNoAuth(Secret const& _k, bytesConstRef _plain, bytes& o_cipher, h128 const& _iv);
```
chriseth: Provide default argument instead of having two implementations?

libdevcrypto/Common.cpp
```diff
@@ -94,6 +110,54 @@ bool dev::decryptSym(Secret const& _k, bytesConstRef _cipher, bytes& o_plain)
...
+	const int c_aesKeyLen = 16;
+	SecByteBlock key(_k.data(), c_aesKeyLen);
+	try
+	{
+		CTR_Mode<AES>::Encryption e;
+		e.SetKeyWithIV(key, key.size(), _iv.data());
+		e.ProcessData(o_cipher.data(), _plain.data(), _plain.size());
+		return _iv;
+	}
+	catch(CryptoPP::Exception& e)
```
chriseth: space.

libdevcrypto/Common.cpp
```diff
@@ -94,6 +110,54 @@ bool dev::decryptSym(Secret const& _k, bytesConstRef _cipher, bytes& o_plain)
...
 
+h128 dev::encryptSymNoAuth(Secret const& _k, bytesConstRef _plain, bytes& o_cipher)
+{
+	h128 iv(Nonce::get());
+	return encryptSymNoAuth(_k, _plain, o_cipher, iv);
+}
+
+h128 dev::encryptSymNoAuth(Secret const& _k, bytesConstRef _plain, bytes& o_cipher, h128 const& _iv)
+{
+	o_cipher.resize(_plain.size());
```
chriseth: Size cannot change due to encryption?

libdevcrypto/Common.cpp
```diff
@@ -94,6 +110,54 @@ bool dev::decryptSym(Secret const& _k, bytesConstRef _cipher, bytes& o_plain)
...
+	const size_t c_aesKeyLen = 16;
+	SecByteBlock key(_k.data(), c_aesKeyLen);
+	try
+	{
+		CTR_Mode<AES>::Decryption d;
+		d.SetKeyWithIV(key, key.size(), _iv.data());
+		d.ProcessData(o_plaintext.data(), _cipher.data(), _cipher.size());
+		return true;
+	}
+	catch(CryptoPP::Exception& e)
```
chriseth: same as above

libdevcrypto/CryptoPP.cpp
```diff
@@ -31,6 +32,124 @@ static_assert(dev::Secret::size == 32, "Secret key must be 32 bytes.");
 static_assert(dev::Public::size == 64, "Public key must be 64 bytes.");
 static_assert(dev::Signature::size == 65, "Signature must be 65 bytes.");
 
+bytes Secp256k1::eciesKDF(Secret _z, bytes _s1, unsigned kdByteLen)
+{
+	// interop w/go ecies implementation
+	
+	// for sha3, blocksize is 136 bytes
+	// for sha256, blocksize is 64 bytes
+	auto reps = ((kdByteLen + 7) * 8) / (64 * 8);
```
chriseth: Formula seems strange. What does it do exactly?

libdevcrypto/CryptoPP.cpp
```diff
@@ -31,6 +32,124 @@ static_assert(dev::Secret::size == 32, "Secret key must be 32 bytes.");
...
+	bytes ctr({0, 0, 0, 1});
+	bytes k;
+	CryptoPP::SHA256 ctx;
+	for (unsigned i = 0; i <= reps; i++)
```
chriseth: Sure about `<=`?

libdevcrypto/CryptoPP.cpp
```diff
@@ -31,6 +32,124 @@ static_assert(dev::Secret::size == 32, "Secret key must be 32 bytes.");
...
+		ctx.Update(_s1.data(), _s1.size());
+		// append hash to k
+		bytes digest(32);
+		ctx.Final(digest.data());
+		ctx.Restart();
+		
+		k.reserve(k.size() + h256::size);
+		move(digest.begin(), digest.end(), back_inserter(k));
+		
+		if (ctr[3]++ && ctr[3] != 0) {
```
chriseth: Braces.

chriseth: Is that a manual increment of a four-byte integer?
If yes, I would consider `ctr` being a real integer combined with `eth::toBigEndian()` where it is needed as a byte array safer and more readable.

libdevcrypto/CryptoPP.cpp
```diff
@@ -31,6 +32,124 @@ static_assert(dev::Secret::size == 32, "Secret key must be 32 bytes.");
...
+	ecdh::agree(r.sec(), _k, z);
+	auto key = eciesKDF(z, bytes(), 512);
+	bytesConstRef eKey = bytesConstRef(&key).cropped(0, 32);
+	bytesRef mKey = bytesRef(&key).cropped(32, 32);
+	CryptoPP::SHA256 ctx;
+	ctx.Update(mKey.data(), mKey.size());
+	ctx.Final(mKey.data());
+	
+	bytes cipherText;
+	encryptSymNoAuth(*(Secret*)eKey.data(), bytesConstRef(&io_cipher), cipherText, h128());
```
chriseth: There is a safe constructor for `Secret` you could use. Note that it involves the incredible copy of 32 bytes (unless the compiler optimizes that away).

libdevcrypto/CryptoPP.cpp
```diff
@@ -31,6 +32,124 @@ static_assert(dev::Secret::size == 32, "Secret key must be 32 bytes.");
...
+	if (!cipherText.size())
```
chriseth: `.empty()`?

libdevcrypto/CryptoPP.cpp
```diff
@@ -31,6 +32,124 @@ static_assert(dev::Secret::size == 32, "Secret key must be 32 bytes.");
...
+	
+	if (io_text.size() < (1 + Public::size + h128::size + 1 + h256::size))
+		// invalid message: length
+		return false;
+
+	h256 z;
+	ecdh::agree(_k, *(Public*)(io_text.data()+1), z);
+	auto key = eciesKDF(z, bytes(), 64);
+	bytesConstRef eKey = bytesConstRef(&key).cropped(0, 32);
+	bytesRef mKeyMaterial = bytesRef(&key).cropped(16, 16);
```
chriseth: `eKey` and `mKeyMaterial` overlap?

libdevcrypto/Common.h
subtly: In this case a default argument is somewhat dangerous and a ref is needed rather than a copy.

libdevcrypto/Common.cpp
subtly: Not for AES. And it's also possible that _plain.data() == o_cipher.data().
The IV is external because it's only communicated for the first block. This is why both methods return the IV -- in CTR the IV is the counter and is incremented for every block. With ECIES we include the IV in the stream but for the framing protocol we don't because the IV is assumed on both sides to be zero.

libdevcrypto/CryptoPP.cpp
subtly: Pretty sure this is a silly way of rounding. Fortunately there are unit tests with vectors from go :)

libdevcrypto/CryptoPP.cpp
subtly: Yes. We need it to run at least once... reps could be 0. This is based on go interop which is supposed to be according to standard. That is still to-be-determined.

libdevcrypto/CryptoPP.cpp
subtly: Same as others -- from go implementation. I'm up for refactoring after we both have test vectors for each step of these methods. (And we've got OpenSSL which might implement this algorithm for us)

libdevcrypto/CryptoPP.cpp
subtly: Petition for FixedHashConstRef, h256ConstRef, and h128ConstRef, h256Ref, h128Ref, etc?
Style: Just not going to copy key material. Also: encryptSymNoAuth will likely take h128 and h256 instead of Secret. Still will use *(h128*) so as to not copy key material.

libdevcrypto/CryptoPP.cpp
chriseth: Then at least remove the braces :-)

libp2p/Host.cpp
```diff
@@ -337,7 +338,8 @@ void Host::runAcceptor()
 				{
 					// doHandshake takes ownersihp of *s via std::move
 					// incoming connection; we don't yet know nodeid
-					doHandshake(s, NodeId());
+					auto handshake = make_shared<PeerHandshake>(this, s);
```
chriseth: Why not just
```
PeerHandshake(this, s).start();
```

libp2p/Host.cpp
```diff
@@ -351,15 +353,18 @@ void Host::runAcceptor()
...
-				boost::system::error_code ec;
-				s->shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec);
-				s->close();
+				if (s->is_open())
+				{
+					boost::system::error_code ec;
+					s->shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec);
+					s->close();
+				}
+				delete s;
```
chriseth: Closing my eyes here again.

libp2p/Host.h
```diff
@@ -236,5 +231,71 @@ class Host: public Worker
...
+	Secret encryptK;
+	Secret macK;
+	h256 egressMac;
+	h256 ingressMac;
+	
+	bytes magicCipherAndMac;
+	bytes recvdMagicCipherAndMac;
+};
+
+struct PeerHandshake: public std::enable_shared_from_this<PeerHandshake>
```
chriseth: Serious violation of coding standards. Please change this to a class and prefix its members with `m_`.

libp2p/Host.h
```diff
@@ -62,24 +63,21 @@ class HostNodeTableHandler: public NodeTableEventHandler
 
 	Host& m_host;
 };
-
+
```
chriseth: whitespace

libp2p/Host.cpp
```diff
@@ -367,22 +372,233 @@ void Host::runAcceptor()
...
+	{
+		nextState = AckAuth;
+		
+		clog(NetConnect) << "Authenticating connection for " << socket->remote_endpoint();
+		
+		if (originated)
+		{
+			clog(NetConnect) << "devp2p.connect.egress sending auth";
+			// egress: tx auth
+			asserts((bool)remote);
```
chriseth: If I understand this correctly, `remote` is a `h512` then you don't really want a c-style cast to `bool` here. If you must, try double negation.

libp2p/Host.cpp
```diff
@@ -367,22 +372,233 @@ void Host::runAcceptor()
...
+			{
+				transition(ec);
+			});
+		}
+		else
+		{
+			clog(NetConnect) << "devp2p.connect.ingress recving auth";
+			// ingress: rx auth
+			authCipher.resize(321);
+			ba::async_read(*socket, ba::buffer(authCipher, 321), [this, self](boost::system::error_code ec, std::size_t)
```
chriseth: Do I understand it correctly that `safe` is a Münchhausen-variable, i.e. it keeps `this` alive as long as the async callback is present?

libp2p/Host.cpp
```diff
@@ -367,22 +372,233 @@ void Host::runAcceptor()
...
+			});
+		}
+	}
+	else if (nextState == Authenticating)
+	{
+		if (originated)
+			clog(NetConnect) << "devp2p.connect.egress sending magic sequence";
+		else
+			clog(NetConnect) << "devp2p.connect.ingress sending magic sequence";
+		PeerSecrets* k = new PeerSecrets;
```
chriseth: Leaks memory on exception.

libp2p/Host.cpp
```diff
@@ -367,22 +372,233 @@ void Host::runAcceptor()
...
+		bytes keyMaterialBytes(512);
+		bytesRef keyMaterial(&keyMaterialBytes);
+
+		ecdhe.agree(remoteEphemeral, ess);
+		ess.ref().copyTo(keyMaterial.cropped(0, h256::size));
+		ss.ref().copyTo(keyMaterial.cropped(h256::size, h256::size));
+		//		auto token = sha3(ssA);
```
chriseth: remove?

libp2p/Host.h
```diff
@@ -62,24 +63,21 @@ class HostNodeTableHandler: public NodeTableEventHandler
...
  * @todo determinePublic: ipv6, udp
  * @todo handle conflict if addNode/requireNode called and Node already exists w/conflicting tcp or udp port
- * @todo write host identifier to disk w/nodes
  * @todo per-session keepalive/ping instead of broadcast; set ping-timeout via median-latency
- * @todo configuration-management (NetworkPrefs+Keys+Topology)
  */
 class Host: public Worker
 {
 	friend class HostNodeTableHandler;
+	friend struct PeerHandshake;
```
chriseth: `Host` is a really social class.

libp2p/Host.h
```diff
@@ -236,5 +231,71 @@ class Host: public Worker
...
+	enum State
+	{
+		Error = -1,
+		New,				// New->AckAuth				[egress: tx auth, ingress: rx auth]
+		AckAuth,			// AckAuth->Authenticating	[egress: rx ack, ingress: tx ack]
+		Authenticating,	// Authenticating			[tx caps, rx caps, authenticate]
+	};
+
+	/// Handshake for ingress connection. Takes ownership of socket.
+	PeerHandshake(Host* _host, bi::tcp::socket* _socket): host(_host), socket(std::move(_socket)), originated(false) { crypto::Nonce::get().ref().copyTo(nonce.ref()); }
```
chriseth: Do you really mean to `move` a pointer?

libp2p/Peer.h
```diff
@@ -29,7 +29,7 @@ namespace dev
 
 namespace p2p
 {
-
+
```
chriseth: whitespace

libp2p/Host.cpp
subtly: Promising again, that I'll move this socket into it's own safe and secure class which follows RAII.

libp2p/Host.h
subtly: Also moving this into implementation files.

libp2p/Host.cpp
subtly: safe or self?  Yes to the latter.
Can variables have a psychological disorder?

libdevcrypto/CryptoPP.cpp
```diff
@@ -19,8 +19,9 @@
  * @date 2014
  */
 
-#include "CryptoPP.h"
 #include <libdevcore/Guards.h>
+#include "ECDHE.h"
+#include "CryptoPP.h"
```
chriseth: Own header should be first.

libdevcrypto/CryptoPP.cpp
```diff
@@ -31,6 +32,125 @@ static_assert(dev::Secret::size == 32, "Secret key must be 32 bytes.");
...
+		ctx.Update(_s1.data(), _s1.size());
+		// append hash to k
+		bytes digest(32);
+		ctx.Final(digest.data());
+		ctx.Restart();
+		
+		k.reserve(k.size() + h256::size);
+		move(digest.begin(), digest.end(), back_inserter(k));
+		
+		if (ctr[3]++ && ctr[3] != 0)
```
chriseth: These two checks seem really wrong. It could be something like
`ctr[3]++ || ctr[2]++ || ctr[1] ++ || ctr[0]++;`

libp2p/Host.cpp
```diff
@@ -469,8 +465,8 @@ void Host::connect(std::shared_ptr<Peer> const& _p)
 	}
 
 	clog(NetConnect) << "Attempting connection to node" << _p->id.abridged() << "@" << _p->peerEndpoint() << "from" << id().abridged();
-	bi::tcp::socket* s = new bi::tcp::socket(m_ioService);
-	s->async_connect(_p->peerEndpoint(), [=](boost::system::error_code const& ec)
+	auto socket = make_shared<RLPXSocket>(new bi::tcp::socket(m_ioService));
```
chriseth: Why do you create a `bi::tcp::socket` on the heap just to move it into `RLPXSocket` in the next instant?

libp2p/Host.cpp
```diff
@@ -538,26 +533,30 @@ void Host::run(boost::system::error_code const&)
...
+		m_connecting.remove_if([](std::weak_ptr<RLPXHandshake> h){ return h.lock(); });
+	}
+	
 	for (auto p: m_sessions)
 		if (auto pp = p.second.lock())
 			pp->serviceNodesRequest();
 
 	keepAlivePeers();
-	disconnectLatePeers();
+//	disconnectLatePeers();
```
chriseth: Commented out

libp2p/Host.h
```diff
@@ -48,7 +50,7 @@ namespace dev
 
 namespace p2p
 {
-
+
```
chriseth: Space

libp2p/Session.cpp
```diff
@@ -356,52 +278,39 @@ void Session::ping()
...
 
-	if (!checkPacket(bytesConstRef(&_msg)))
+	bytesConstRef msg(&_msg);
+	if (!checkPacket(msg))
 		clogS(NetWarn) << "INVALID PACKET CONSTRUCTED!";
 
 	if (!m_socket.is_open())
 		return;
-
+
```
chriseth: space

libp2p/Session.cpp
```diff
@@ -356,52 +278,39 @@ void Session::ping()
...
-
 void Session::send(bytes&& _msg)
 {
-	clogS(NetLeft) << RLP(bytesConstRef(&_msg).cropped(8));
+	clogS(NetLeft) << RLP(bytesConstRef(&_msg).cropped(1));
 
-	if (!checkPacket(bytesConstRef(&_msg)))
+	bytesConstRef msg(&_msg);
+	if (!checkPacket(msg))
 		clogS(NetWarn) << "INVALID PACKET CONSTRUCTED!";
```
chriseth: Is it still sent?

libp2p/Session.cpp
```diff
@@ -484,95 +394,88 @@ void Session::disconnect(DisconnectReason _reason)
...
+			if (!m_io->authAndDecryptHeader(header))
+			{
+				clog(NetWarn) << "header decrypt failed";
+				drop(BadProtocol); // todo: better error
+				return;
+			}
+
+			/// check frame size
+			uint32_t frameSize = (m_data[0] * 256 + m_data[1]) * 256 + m_data[2];
+			if (frameSize > 16777216)
```
chriseth: `1 << 24`?

libdevcrypto/CryptoPP.cpp
subtly: Not fixing this until Gav confirms the change in coding standards. This is the same header format for every header created < 2015.

libp2p/Host.cpp
subtly: I don't wish to optimize this at this time (would add even more to commit). I've been asked to make changes small.

libdevcrypto/CryptoPP.cpp
subtly: If it is wrong, it is wrong in Go and correcting would break interoperability. Therefore a separate issue must be created for this particular problem. (Hopefully they'll catch it in audit of go code)

libp2p/Session.cpp
subtly: Yes. The logic is unchanged -- log to NetWarn and let packet through.

libdevcrypto/CryptoPP.cpp
gavofyork: Confirmed. It was notionally altered around May 2014.

PR #1091, opened 2015-02-19
Account proxy queues

chriseth:


libweb3jsonrpc/WebThreeStubServerBase.cpp
```diff
@@ -72,6 +73,18 @@ static Json::Value toJson(dev::eth::Transaction const& _t)
 	return res;
 }
 
+static Json::Value toJson(dev::eth::TransactionSkeleton const& _t)
+{
+	Json::Value res;
+	res["to"] = toJS(_t.to);
+	res["from"] = toJS(_t.from);
+	res["gas"] = (int)_t.gas;
```
chriseth: I copied this cast from above. Is it really fine?

gavofyork: it really is.

libweb3jsonrpc/AccountHolder.cpp
```diff
@@ -0,0 +1,104 @@
...
+	Address const* bestMatch = &m_accounts.front();
+	for (auto const& account: m_accounts)
+		if (m_client()->balanceAt(account) > m_client()->balanceAt(*bestMatch))
+			bestMatch = &account;
+	return *bestMatch;
+}
+
+int AccountHolder::addProxyAccount(const Address& _account)
+{
+	static std::mt19937 s_randomNumberGenerator(time(0));
```
gavofyork: time(0) is no good if two proxy accounts are created within the same second. probably need a static int here, too.

gavofyork:
rebase this so the checks work?

PR #1099, opened 2015-02-21
Sol source libraries

LefterisJP:
Turned out to be smaller than expected.

Some source units can be added as Libraries. If that is so, they they are compiled iff they are requested by an import statement.

Removed the temporary `#require Contract` preprocessor hack. Those contracts are now be added as libraries.

There already are some tests in the `EndToEndTests` which check for usage of the std library so this should actually already be tested.

### DIFF
--- a/libsolidity/CompilerStack.h
+++ b/libsolidity/CompilerStack.h
@@ -64,8 +64,8 @@ public:
 
 	/// Adds a source object (e.g. file) to the parser. After this, parse has to be called again.
 	/// @returns true if a source object by the name already existed and was replaced.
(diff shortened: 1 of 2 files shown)


libsolidity/CompilerStack.h
```diff
@@ -125,7 +125,8 @@ class CompilerStack: boost::noncopyable
 		std::shared_ptr<Scanner> scanner;
 		std::shared_ptr<SourceUnit> ast;
 		std::string interface;
-		void reset() { scanner.reset(); ast.reset(); interface.clear(); }
+		bool isLibrary;
```
chriseth: Default value?

LefterisJP: I thought that `reset()` function addition was enough.
But aaaahh my C background did not let me think of a default argument in a struct :)

Issue #1101, opened 2015-02-21
libmicrohttpd-dll.dll is missing on Windows


nicksavers:
The latest [build](https://build.ethdev.com/builders/Windows%20C%2B%2B%20master%20branch) of the master branch for poc-8 (commit https://github.com/ethereum/cpp-ethereum/commit/d1ed000f70218064b3b8e92291f7330ac73c3016)
doesn't work on Windows 7 64-bit.

![The program can't start because libmicrohttpd-dll.dll is missing from your computer. 
Try reinstalling the program to fix this problem](http://i.imgur.com/jOtcIpf.png)

debris:
@arkpar can you check if it is related with #1081 ?

nicksavers:
Fix confirmed.

arkpar:
Needs to be merged to master or poc-8 I guess

PR #1104, opened 2015-02-22
Fixed microhttpd deployment

arkpar:
#1101
We should either build it as static lib like the rest of the non-qt libs, or at least designate a common 
location for dlls in extdeps


nicksavers:
I tested running it on the build-server (Windows Server 2012 R2) and the startup issue is now fixed for AlethZero. But eth.exe (which is not distributed in the zip file) isn't fixed yet. Possibly https://github.com/ethereum/cpp-ethereum/blob/develop/eth/CMakeLists.txt also needs an update

nicksavers:
Now eth.exe also works

PR #1109, opened 2015-02-23
Transaction bug fix

CJentzsch:
fixes https://github.com/ethereum/tests/issues/61.
The question is, is it allowed to have addresses with less than 20 bytes?
According to my interpretation of the YP it is not. YP: "to: The 160-bit address of the message call’s recipient".
But one could also argue that an address with less than 20 bytes (but still not empty) belongs to the set B_20 and could be considered valid. Then we need to ask should it be encoded in the RLP as 20 bytes, filling up the missing bytes with 0's, or is it allowed to have, e.g. 10 bytes, adresses RLP encoded? @gavofyork

### DIFF
--- a/libdevcore/RLP.h
+++ b/libdevcore/RLP.h
@@ -242,7 +242,9 @@ public:
 		AllowNonCanon = 1,
 		ThrowOnFail = 4,
 		FailIfTooBig = 8,
+		FailIfTooSmall = 16,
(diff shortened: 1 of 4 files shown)


gavofyork:
B_20 only includes byte sequences that are exactly 20 items large, so, no, transactions must have *exactly* 20 bytes or *exactly* no bytes. anything else is invalid.

PR #1120, opened 2015-02-23
Fixed contract creation transaction return address and other issues

arkpar:


arkpar:
In a few places hash was implicitly converted to bool, resulting in value being lost

CJentzsch:
looks good

PR #1134, opened 2015-02-24
DApp loading in Alethzero

arkpar:


arkpar:
Fails to compile because buildbot does not have qt 5.4 yet

PR #1140, opened 2015-02-25
LValue refactoring.

chriseth:
Create a new virtual class for LValues.

### DIFF
--- a/libsolidity/ExpressionCompiler.cpp
+++ b/libsolidity/ExpressionCompiler.cpp
@@ -29,6 +29,7 @@
 #include <libsolidity/ExpressionCompiler.h>
 #include <libsolidity/CompilerContext.h>
 #include <libsolidity/CompilerUtils.h>
+#include <libsolidity/LValue.h>
 
 using namespace std;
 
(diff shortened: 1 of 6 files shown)


libsolidity/ExpressionCompiler.cpp
```diff
@@ -37,41 +38,167 @@ namespace dev
...
+					m_context << (u256(1) << (256 - targetType.getNumBytes() * 8))
+							  << eth::Instruction::DUP1 << eth::Instruction::SWAP2
+							  << eth::Instruction::DIV << eth::Instruction::MUL;
+			}
+		}
+	}
+	else if (stackTypeCategory == Type::Category::Enum)
+		solAssert(targetTypeCategory == Type::Category::Integer ||
+			targetTypeCategory == Type::Category::Enum, "");
+	else if (stackTypeCategory == Type::Category::Integer || stackTypeCategory == Type::Category::Contract ||
```
LefterisJP: since you already got a line break here why not put each condition in its own line? I think it would be more readable that way.

libsolidity/ExpressionCompiler.cpp
```diff
@@ -998,319 +1029,32 @@ void ExpressionCompiler::appendExpressionCopyToMemory(Type const& _expectedType,
...
-	m_context(&_compilerContext), m_type(_type), m_dataType(_dataType),
-	m_baseStackOffset(_baseStackOffset)
+void ExpressionCompiler::setLValueFromDeclaration(Declaration const& _declaration, Expression const& _expression)
 {
-	//@todo change the type cast for arrays
-	solAssert(m_dataType->getStorageSize() <= numeric_limits<unsigned>::max(),
-			  "The storage size of " + m_dataType->toString() + " should fit in unsigned");
-	if (m_type == LValueType::Storage)
-		m_size = unsigned(m_dataType->getStorageSize());
+	solAssert(!m_currentLValue, "Current LValue not reset when trying to set to new one.");
```
LefterisJP: When trying to set the new one maybe?

libsolidity/ExpressionCompiler.cpp
```diff
@@ -998,319 +1029,32 @@ void ExpressionCompiler::appendExpressionCopyToMemory(Type const& _expectedType,
...
-void ExpressionCompiler::LValue::retrieveValueFromStorage(bool _remove) const
+void ExpressionCompiler::setLValueToStorageItem(Expression const& _expression)
 {
-	if (!m_dataType->isValueType())
-		return; // no distinction between value and reference for non-value types
-	if (!_remove)
-		*m_context << eth::Instruction::DUP1;
-	if (m_size == 1)
-		*m_context << eth::Instruction::SLOAD;
+	solAssert(!m_currentLValue, "Current LValue not reset when trying to set to new one.");
```
LefterisJP: same as previous comment here

libsolidity/ExpressionCompiler.cpp
chriseth: I wish we had `else if (stackTypeCategory in {Type::Category::Integer || ...})`. Or is there something like that?

libsolidity/LValue.h
```diff
@@ -0,0 +1,112 @@
...
+	/// @returns true if this lvalue reference type occupies a slot on the stack.
+	virtual bool storesReferenceOnStack() const = 0;
+	/// Copies the value of the current lvalue to the top of the stack and, if @a _remove is true,
+	/// also removes the reference from the stack.
+	/// @a _location source location of the current expression, used for error reporting.
+	virtual void retrieveValue(SourceLocation const& _location, bool _remove = false) const = 0;
+	/// Moves a value from the stack to the lvalue. Removes the value if @a _move is true.
+	/// @a _location is the source location of the expression that caused this operation.
+	/// Stack pre: value [lvalue_ref]
+	/// Stack post if !_move: value_of(lvalue_ref)
```
LefterisJP: Stack post:

libsolidity/LValue.h
```diff
@@ -0,0 +1,112 @@
...
+	virtual void storeValue(Type const& _sourceType,
+		SourceLocation const& _location = SourceLocation(), bool _move = false) const = 0;
```
LefterisJP: I think since you broke one argument to a new line maybe you should break them all?

PR #1146, opened 2015-02-26
Sol fix accessing public variable of base class

LefterisJP:
The following should now work
```
contract Parent {
    uint256 public m_aMember;
}
contract Child is Parent{
    function foo() returns (uint256) { return Parent.m_aMember; }
}
```

### DIFF
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -215,26 +215,23 @@ vector<ASTPointer<Declaration>> const& ContractDefinition::getInheritableMembers
 	{
 		set<string> memberSeen;
 		m_inheritableMembers.reset(new vector<ASTPointer<Declaration>>());
-		for (ContractDefinition const* contract: getLinearizedBaseContracts())
+		auto addInheritableMember = [&](ASTPointer<Declaration> const& _decl)
 		{
-			auto addInheritableMember = [&](ASTPointer<Declaration> const& _decl)
+			if (memberSeen.count(_decl->getName()) == 0 && _decl->isVisibleInDerivedContracts())
 			{
-				if (memberSeen.count(_decl->getName()) == 0 && _decl->isVisibleInDerivedContracts())
-				{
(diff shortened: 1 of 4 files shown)


libsolidity/AST.cpp
```diff
@@ -209,6 +209,40 @@ vector<pair<FixedHash<4>, FunctionTypePointer>> const& ContractDefinition::getIn
...
+			for (ASTPointer<FunctionDefinition> const& f: contract->getDefinedFunctions())
+				if (f->isPublic() && !f->isConstructor() && !f->getName().empty()
+					&& memberSeen.count(f->getName()) == 0 && f->isVisibleInDerivedContracts())
+				{
+					memberSeen.insert(f->getName());
+					m_inheritableMembers->push_back(f);
+				}
+
+			for (ASTPointer<VariableDeclaration> const& v: contract->getStateVariables())
+				if (v->isPublic() && memberSeen.count(v->getName()) == 0)
```
chriseth: "Internal" would also suffice. Can you check whether we can make "isVisibleInDerivedContracts" virtual so that it behaves exactly as these conditions here? Then we can treat all the elements of those vectors equally after a cast to a Declaration.

chriseth:
Please also add a test to access an "internal" member in the same way. The test should succeed but I think currently it will fail.

libsolidity/AST.h
```diff
@@ -405,6 +409,11 @@ class FunctionDefinition: public Declaration, public VariableScope, public Docum
 	ASTPointer<ParameterList> const& getReturnParameterList() const { return m_returnParameters; }
 	Block const& getBody() const { return *m_body; }
 
+	virtual bool isVisibleInDerivedContracts() const override
+	{
+		return !isConstructor() && !getName().empty() && isVisibleInContract() &&
```
chriseth: Call to `Declaration::isVisibleInDerivedContracts`?

PR #1161, opened 2015-02-27
 mix - Add validation for deployment values

yann300:


gavofyork:
Fails cpp-test:

process killed by signal 9
program finished with exit code -1
elapsedTime=246.365399

yann300:
I don't think this is related with this PR, but I'm checking

PR #1162, opened 2015-02-27
implement overloaded funciton

guanqun:
@chriseth This is my almost done patch, I'd like to have some comments from you instead of heading down on the wrong path :) It can pass all solidity test cases except "inheritance_diamond_basic", working on that.

Here are my design choices, at first place, I created another ASTNode called FunctionIdentifier, but being more familiar with the code base, I want to combine FunctionIdentifier with simply Identifier. I want your advice on this.

Another point is whether I should treat this illegal or legal? Base contract has a function called fun(), but the derived one has a different signature, in the original implementation, we don't allow this, how about now?

### DIFF
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -356,7 +356,12 @@ void VariableDeclaration::checkTypeRequirements()
 	else
 	{
 		// no type declared and no previous assignment, infer the type
-		m_value->checkTypeRequirements();
+		Identifier* identifier = dynamic_cast<Identifier*>(m_value.get());
+		if (identifier)
+			identifier->checkTypeRequirementsFromVariableDeclaration();
+		else
+			m_value->checkTypeRequirements();
+
 		TypePointer type = m_value->getType();
 		if (type->getCategory() == Type::Category::IntegerConstant)
 		{
@@ -484,39 +489,7 @@ void Return::checkTypeRequirements()
 
 void VariableDeclarationStatement::checkTypeRequirements()
 {
-<<<<<<< HEAD
 	m_variable->checkTypeRequirements();
-=======
-	// Variables can be declared without type (with "var"), in which case the first assignment
-	// sets the type.
-	// Note that assignments before the first declaration are legal because of the special scoping
-	// rules inherited from JavaScript.
-	if (m_variable->getValue())
-	{
-		if (m_variable->getType())
-			m_variable->getValue()->expectType(*m_variable->getType());
-		else
-		{
-			// no type declared and no previous assignment, infer the type
-			Identifier* identifier = dynamic_cast<Identifier*>(m_variable->getValue().get());
-			if (identifier)
-				identifier->checkTypeRequirementsFromVariableDeclaration();
-			else
-				m_variable->getValue()->checkTypeRequirements();
-			TypePointer type = m_variable->getValue()->getType();
-			if (type->getCategory() == Type::Category::IntegerConstant)
-			{
-				auto intType = dynamic_pointer_cast<IntegerConstantType const>(type)->getIntegerType();
-				if (!intType)
-					BOOST_THROW_EXCEPTION(m_variable->getValue()->createTypeError("Invalid integer constant " + type->toString()));
-				type = intType;
-			}
-			else if (type->getCategory() == Type::Category::Void)
-				BOOST_THROW_EXCEPTION(m_variable->createTypeError("var cannot be void type"));
-			m_variable->setType(type);
-		}
-	}
->>>>>>> implement overload resolution
 }
 
 void Assignment::checkTypeRequirements()
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -674,15 +674,23 @@ void FunctionCall::checkTypeRequirements()
 	else if (OverloadedFunctionType const* overloadedTypes = dynamic_cast<OverloadedFunctionType const*>(expressionType))
 	{
 		// this only applies to "x(3)" where x is assigned by "var x = f;" where f is an overloaded functions.
-		overloadedTypes->m_identifier->overloadResolution(*this);
-		FunctionType const* functionType = dynamic_cast<FunctionType const*>(overloadedTypes->m_identifier->getType().get());
+		auto identifier = dynamic_cast<Identifier*>(m_expression.get());
+		solAssert(identifier, "only applies to 'var x = f;'");
 
-		// @todo actually the return type should be an anonymous struct,
-		// but we change it to the type of the first return value until we have structs
+		Declaration const* function = overloadedTypes->getIdentifier()->overloadResolution(*this);
+		if (!function)
+			BOOST_THROW_EXCEPTION(createTypeError("Can't resolve declarations"));
+
+		identifier->setReferencedDeclaration(*function);
+		identifier->checkTypeRequirements();
+
+		TypePointer type = identifier->getType();
+		FunctionType const* functionType = dynamic_cast<FunctionType const*>(type.get());
+
 		if (functionType->getReturnParameterTypes().empty())
 			m_type = make_shared<VoidType>();
 		else
-			m_type = functionType->getReturnParameterTypes().front();		
+			m_type = functionType->getReturnParameterTypes().front();
 	}
 	else
 		BOOST_THROW_EXCEPTION(createTypeError("Type is not callable."));
@@ -781,7 +789,7 @@ void Identifier::checkTypeRequirementsWithFunctionCall(FunctionCall const& _func
 	solAssert(m_referencedDeclaration || !m_overloadedDeclarations.empty(), "Identifier not resolved.");
 
 	if (!m_referencedDeclaration)
-		overloadResolution(_functionCall);
+		setReferencedDeclaration(*overloadResolution(_functionCall));
 
 	checkTypeRequirements();
 }
@@ -791,7 +799,7 @@ void Identifier::checkTypeRequirementsFromVariableDeclaration()
 	solAssert(m_referencedDeclaration || !m_overloadedDeclarations.empty(), "Identifier not resolved.");
 
 	if (!m_referencedDeclaration)
-		m_type = make_shared<OverloadedFunctionType>(m_overloadedDeclarations, this);
+		m_type = make_shared<OverloadedFunctionType>(this);
 	else
 		checkTypeRequirements();
 
@@ -808,12 +816,10 @@ void Identifier::checkTypeRequirements()
 		BOOST_THROW_EXCEPTION(createTypeError("Declaration referenced before type could be determined."));
 }
 
-void Identifier::overloadResolution(FunctionCall const& _functionCall)
+Declaration const* Identifier::overloadResolution(FunctionCall const& _functionCall)
 {
 	solAssert(m_overloadedDeclarations.size() > 1, "FunctionIdentifier not resolved.");
 	solAssert(!m_referencedDeclaration, "Referenced declaration should be null before overload resolution.");
-
-	bool resolved = false;
 
 	std::vector<ASTPointer<Expression const>> arguments = _functionCall.getArguments();
 	std::vector<ASTPointer<ASTString>> const& argumentNames = _functionCall.getNames();
@@ -844,7 +850,7 @@ void Identifier::overloadResolution(FunctionCall const& _functionCall)
 			{
 				return declaration->getScope() == possibles.front()->getScope();
 			}))
-				setReferencedDeclaration(*possibles.front());
+				return possibles.front();
 		else
 			BOOST_THROW_EXCEPTION(createTypeError("Can't resolve identifier"));
 	}
@@ -852,6 +858,7 @@ void Identifier::overloadResolution(FunctionCall const& _functionCall)
 		// named arguments
 		// TODO: don't support right now
 		BOOST_THROW_EXCEPTION(createTypeError("Named arguments with overloaded functions are not supported yet."));
+	return nullptr;
 }
 
 void ElementaryTypeNameExpression::checkTypeRequirements()
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -82,7 +82,8 @@ void ContractDefinition::checkTypeRequirements()
 	{
 		string signature = function->getCanonicalSignature();
 		if (functions.count(signature))
-			BOOST_THROW_EXCEPTION(DeclarationError() << errinfo_comment("Duplicate functions are not allowed."));
+			BOOST_THROW_EXCEPTION(DeclarationError() << errinfo_sourceLocation(function->getLocation())
+														<< errinfo_comment("Duplicate functions are not allowed."));
 		functions.insert(signature);
 	}
 
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -493,14 +493,10 @@ void VariableDeclarationStatement::checkTypeRequirements()
 	if (m_variable->getValue())
 	{
 		if (m_variable->getType())
-		{
-			std::cout << "getType() ok" << std::endl;
 			m_variable->getValue()->expectType(*m_variable->getType());
-		}
 		else
 		{
 			// no type declared and no previous assignment, infer the type
-			std::cout << "here's where called...." << std::endl;
 			Identifier* identifier = dynamic_cast<Identifier*>(m_variable->getValue().get());
 			if (identifier)
 				identifier->checkTypeRequirementsFromVariableDeclaration();
@@ -804,14 +800,9 @@ void Identifier::checkTypeRequirementsFromVariableDeclaration()
 
 void Identifier::checkTypeRequirements()
 {
-	// var x = f; TODO!
 	solAssert(m_referencedDeclaration, "Identifier not resolved.");
 
 	m_isLValue = m_referencedDeclaration->isLValue();
-	if (m_isLValue)
-		std::cout << "Identifier: " << string(getName()) << " -> true" << std::endl;
-	else
-		std::cout << "Identifier: " << string(getName()) << " -> true" << std::endl;
 	m_type = m_referencedDeclaration->getType(m_currentContract);
 	if (!m_type)
 		BOOST_THROW_EXCEPTION(createTypeError("Declaration referenced before type could be determined."));
@@ -846,7 +837,6 @@ void Identifier::overloadResolution(FunctionCall const& _functionCall)
 							})))
 				possibles.push_back(declaration);			
 		}
-		std::cout << "possibles: " << possibles.size() << std::endl;
 		if (possibles.empty())
 			BOOST_THROW_EXCEPTION(createTypeError("Can't resolve identifier"));
 		else if (std::none_of(possibles.cbegin() + 1, possibles.cend(),
@@ -859,11 +849,9 @@ void Identifier::overloadResolution(FunctionCall const& _functionCall)
 			BOOST_THROW_EXCEPTION(createTypeError("Can't resolve identifier"));
 	}
 	else
-	{
 		// named arguments
 		// TODO: don't support right now
-		// BOOST_THROW_EXCEPTION(createTypeError("Named arguments with overloaded functions are not supported yet."));
-	}
+		BOOST_THROW_EXCEPTION(createTypeError("Named arguments with overloaded functions are not supported yet."));
 }
 
 void ElementaryTypeNameExpression::checkTypeRequirements()
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -76,6 +76,15 @@ void ContractDefinition::checkTypeRequirements()
 
 	for (ASTPointer<FunctionDefinition> const& function: getDefinedFunctions())
 		function->checkTypeRequirements();
+	// check for duplicate declaration
+	set<string> functions;
+	for (ASTPointer<FunctionDefinition> const& function: getDefinedFunctions())
+	{
+		string signature = function->getCanonicalSignature();
+		if (functions.count(signature))
+			BOOST_THROW_EXCEPTION(DeclarationError() << errinfo_comment("Duplicate functions are not allowed."));
+		functions.insert(signature);
+	}
 
 	for (ASTPointer<VariableDeclaration> const& variable: m_stateVariables)
 		variable->checkTypeRequirements();
@@ -129,6 +138,7 @@ void ContractDefinition::checkIllegalOverrides() const
 	// TODO unify this at a later point. for this we need to put the constness and the access specifier
 	// into the types
 	map<string, FunctionDefinition const*> functions;
+	set<string> functionNames;
 	map<string, ModifierDefinition const*> modifiers;
 
 	// We search from derived to base, so the stored item causes the error.
@@ -141,7 +151,8 @@ void ContractDefinition::checkIllegalOverrides() const
 			string const& name = function->getName();
 			if (modifiers.count(name))
 				BOOST_THROW_EXCEPTION(modifiers[name]->createTypeError("Override changes function to modifier."));
-			FunctionDefinition const*& override = functions[name];
+			FunctionDefinition const*& override = functions[function->getCanonicalSignature()];
+			functionNames.insert(name);
 			if (!override)
 				override = function.get();
 			else if (override->getVisibility() != function->getVisibility() ||
@@ -152,13 +163,13 @@ void ContractDefinition::checkIllegalOverrides() const
 		for (ASTPointer<ModifierDefinition> const& modifier: contract->getFunctionModifiers())
 		{
 			string const& name = modifier->getName();
-			if (functions.count(name))
-				BOOST_THROW_EXCEPTION(functions[name]->createTypeError("Override changes modifier to function."));
 			ModifierDefinition const*& override = modifiers[name];
 			if (!override)
 				override = modifier.get();
 			else if (ModifierType(*override) != ModifierType(*modifier))
 				BOOST_THROW_EXCEPTION(override->createTypeError("Override changes modifier signature."));
+			if (functionNames.count(name))
+				BOOST_THROW_EXCEPTION(override->createTypeError("Override changes modifier to function."));
 		}
 	}
 }
@@ -185,16 +196,21 @@ vector<pair<FixedHash<4>, FunctionTypePointer>> const& ContractDefinition::getIn
 	if (!m_interfaceFunctionList)
 	{
 		set<string> functionsSeen;
+		set<string> signaturesSeen;
 		m_interfaceFunctionList.reset(new vector<pair<FixedHash<4>, FunctionTypePointer>>());
 		for (ContractDefinition const* contract: getLinearizedBaseContracts())
 		{
 			for (ASTPointer<FunctionDefinition> const& f: contract->getDefinedFunctions())
-				if (f->isPublic() && !f->isConstructor() && !f->getName().empty() && functionsSeen.count(f->getName()) == 0)
+			{
+				string functionSignature = f->getCanonicalSignature();
+				if (f->isPublic() && !f->isConstructor() && !f->getName().empty() && signaturesSeen.count(functionSignature) == 0)
 				{
 					functionsSeen.insert(f->getName());
-					FixedHash<4> hash(dev::sha3(f->getCanonicalSignature()));
+					signaturesSeen.insert(functionSignature);
+					FixedHash<4> hash(dev::sha3(functionSignature));
 					m_interfaceFunctionList->push_back(make_pair(hash, make_shared<FunctionType>(*f, false)));
 				}
+			}
 
 			for (ASTPointer<VariableDeclaration> const& v: contract->getStateVariables())
 				if (v->isPublic() && functionsSeen.count(v->getName()) == 0)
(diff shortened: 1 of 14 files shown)


libsolidity/AST.h
```diff
@@ -405,7 +405,7 @@ class FunctionDefinition: public Declaration, public VariableScope, public Docum
 	ASTPointer<ParameterList> const& getReturnParameterList() const { return m_returnParameters; }
 	Block const& getBody() const { return *m_body; }
 
-	virtual TypePointer getType(ContractDefinition const*) const override;
+	virtual TypePointer getType(ContractDefinition const* _contract = nullptr) const override;
```
guanqun: This line is added by accident, will remove this...

libsolidity/DeclarationContainer.cpp
```diff
@@ -1,18 +1,18 @@
...
-    GNU General Public License for more details.
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
 
-    You should have received a copy of the GNU General Public License
-    along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
```
guanqun: oops, will change this format...

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -448,23 +448,25 @@ BOOST_AUTO_TEST_CASE(cyclic_inheritance)
...
-BOOST_AUTO_TEST_CASE(illegal_override_direct)
+// TODO: add an end to end test
+BOOST_AUTO_TEST_CASE(legal_override_direct)
 {
 	char const* text = R"(
 		contract B { function f() {} }
 		contract C is B { function f(uint i) {} }
 	)";
-	BOOST_CHECK_THROW(parseTextAndResolveNames(text), TypeError);
+	BOOST_CHECK_NO_THROW(parseTextAndResolveNames(text));
```
guanqun: Here's what I'm asking for advice. I'm making it NO_THROW right now.

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -448,23 +448,25 @@ BOOST_AUTO_TEST_CASE(cyclic_inheritance)
...
+// TODO: add an end to end test
+BOOST_AUTO_TEST_CASE(legal_override_indirect)
 {
 	char const* text = R"(
 		contract A { function f(uint a) {} }
 		contract B { function f() {} }
 		contract C is A, B { }
 	)";
-	BOOST_CHECK_THROW(parseTextAndResolveNames(text), TypeError);
+	BOOST_CHECK_NO_THROW(parseTextAndResolveNames(text));
```
guanqun: making it NO_THROW right now.

test/SolidityParser.cpp
```diff
@@ -142,6 +142,31 @@ BOOST_AUTO_TEST_CASE(missing_argument_in_named_args)
...
+	char const* text = R"(
+		contract test {
+			function fun(uint a) returns(uint r) { return a; }
+			function fun(uint a) returns(uint r) { return a; }
+		}
+	)";
+	// with support of overloaded functions, during parsing,
+	// we can't determine whether they match exactly, however
+	// it will throw TypeError in following stage.
+	BOOST_CHECK_NO_THROW(parseText(text));
```
guanqun: I believe this is valid as during parsing we can't determine the types.

guanqun:
The main change is to use a set of Declarations in DeclarationContainer.

test/SolidityNameAndTypeResolution.cpp
chriseth: I would say we allow it, i.e. less strict than C++ inheritance.

libsolidity/AST.cpp
```diff
@@ -673,6 +676,106 @@ void Identifier::checkTypeRequirements()
...
+	}
+
+	bool resolved = false;
+	auto const& arguments = getFunctionCall().getArguments();
+	auto const& argumentNames = getFunctionCall().getNames();
+
+	if (argumentNames.empty())
+	{
+		// positional arguments
+		for (auto&& declaration: m_overloadedDeclarations)
```
chriseth: Double `&&` on purpose?

libsolidity/AST.cpp
```diff
@@ -673,6 +676,106 @@ void Identifier::checkTypeRequirements()
...
+		{
+			auto type = declaration->getType();
+			FunctionType const& functionType = dynamic_cast<FunctionType const&>(*type);
```
chriseth: I would actually prefer the reversed behaviour:
Since this is a dynamic cast, we actually know its type, so `auto` is good idea to shorten the code.
On the other hand, using things like `auto type` above hides the information whether `type` is a `TypePointer` or a `Type` and thus can lead to serious errors when e.g. comparing types by equality.

libsolidity/AST.cpp
```diff
@@ -673,6 +676,106 @@ void Identifier::checkTypeRequirements()
...
+						}
+				}
+
+
+			}
+		}
+	}
+
+	if (!resolved)
+		BOOST_THROW_EXCEPTION(createTypeError("Can't resolve identifier"));
```
chriseth: "Correct function overload not found?" - oh, no JavaScript programmer will get the meaning of those error messages...

libsolidity/AST.cpp
```diff
@@ -673,6 +676,106 @@ void Identifier::checkTypeRequirements()
...
+			auto type = declaration->getType();
+			FunctionType const& functionType = dynamic_cast<FunctionType const&>(*type);
+
+			auto const& parameterTypes = functionType.getParameterTypes();
+
+			if (arguments.size() == parameterTypes.size())
+			{
+				bool ok = true;
+				for (size_t i = 0; ok && i < arguments.size(); i++)
+					if (!functionType.takesArbitraryParameters() &&
```
chriseth: The first condition is independent of the loop.

libsolidity/AST.cpp
```diff
@@ -673,6 +676,106 @@ void Identifier::checkTypeRequirements()
...
+		for (auto&& declaration: m_overloadedDeclarations)
+		{
+			auto type = declaration->getType();
+			FunctionType const& functionType = dynamic_cast<FunctionType const&>(*type);
+
+			auto const& parameterTypes = functionType.getParameterTypes();
+
+			if (arguments.size() == parameterTypes.size())
+			{
+				bool ok = true;
```
chriseth: `invalidTypeFound`?

libsolidity/AST.cpp
chriseth: `const` for safety?

libsolidity/AST.cpp
```diff
@@ -673,6 +676,106 @@ void Identifier::checkTypeRequirements()
...
+				{
+					bool found = false;
+					for (size_t i = 0; !found && i < argumentNames.size(); i++)
+						if ((found = (parameterName == *argumentNames[i])))
+						{
+							// we found the actual parameter position
+							// positionedArguments.push_back(arguments[i]);								
+						}
+				}
+
```
chriseth: Blank lines.

libsolidity/AST.cpp
```diff
@@ -673,6 +676,106 @@ void Identifier::checkTypeRequirements()
...
+			if (parameterTypes.size() == parameterNames.size() && parameterTypes.size() == arguments.size())
+			{
+				bool ok = true;
+				for (auto const& parameterName: parameterNames)
+				{
+					bool found = false;
+					for (size_t i = 0; !found && i < argumentNames.size(); i++)
+						if ((found = (parameterName == *argumentNames[i])))
+						{
+							// we found the actual parameter position
```
chriseth: This is still todo?

libsolidity/AST.cpp
```diff
@@ -673,6 +676,106 @@ void Identifier::checkTypeRequirements()
...
+}
+
+void FunctionIdentifier::checkTypeRequirements()
+{
+	std::vector<Declaration const*> declarations{ m_overloadedDeclarations.cbegin(), m_overloadedDeclarations.cend() };
+
+	if (declarations.size() > 1)
+		for (auto&& declaration: declarations)
+			if (dynamic_cast<FunctionDefinition const*>(declaration) == nullptr)
+				BOOST_THROW_EXCEPTION(createTypeError("Only functions allow multiple declarations."));
```
chriseth: That should actually be an error of `declaration` and should also be checked somewhere else (for example, it is still an error, even if the declarations are not used).

libsolidity/AST.cpp
```diff
@@ -673,6 +676,106 @@ void Identifier::checkTypeRequirements()
...
+
+	for (size_t i = 0; i < declarations.size(); i++)
+		for (size_t j = i + 1; j < declarations.size(); j++)
+			if (dynamic_cast<FunctionType const*>(declarations[i]->getType().get())->getCanonicalSignature() ==
+				dynamic_cast<FunctionType const*>(declarations[j]->getType().get())->getCanonicalSignature())
+				BOOST_THROW_EXCEPTION(createTypeError("Two exact functions are not allowed."));
```
chriseth: Again, this should be checked at the contract level.

libsolidity/CompilerContext.cpp
```diff
@@ -102,8 +102,12 @@ eth::AssemblyItem CompilerContext::getVirtualFunctionEntryLabel(FunctionDefiniti
 	solAssert(!m_inheritanceHierarchy.empty(), "No inheritance hierarchy set.");
 	for (ContractDefinition const* contract: m_inheritanceHierarchy)
 		for (ASTPointer<FunctionDefinition> const& function: contract->getDefinedFunctions())
-			if (!function->isConstructor() && function->getName() == _function.getName())
+		{
+			if (!function->isConstructor() &&
+				dynamic_cast<FunctionType const*>(function->getType().get())->getCanonicalSignature() ==
```
chriseth: `dynamic_cast<FunctionType const&>`

libsolidity/DeclarationContainer.h
```diff
@@ -48,14 +48,15 @@ class DeclarationContainer
 	/// @param _update if true, replaces a potential declaration that is already present
 	/// @returns false if the name was already declared.
 	bool registerDeclaration(Declaration const& _declaration, bool _invisible = false, bool _update = false);
-	Declaration const* resolveName(ASTString const& _name, bool _recursive = false) const;
+    bool checkDuplicateFunctionDeclaration();
```
chriseth: Indentation.

libsolidity/NameAndTypeResolver.cpp
```diff
@@ -82,6 +82,7 @@ void NameAndTypeResolver::checkTypeRequirements(ContractDefinition& _contract)
 	for (ASTPointer<StructDefinition> const& structDef: _contract.getDefinedStructs())
 		structDef->checkValidityOfMembers();
 	_contract.checkTypeRequirements();
+	checkDuplicateFunctionDeclaration(_contract); // TODO:
```
chriseth: I think this should rather be done inside `_contract.checkTypeRequirements`.
We have kind of a two-stage process: References to declarations are resolved using the `DeclarationContainer`, but it is quite "dumb" and actually thrown away after that stage.

libsolidity/NameAndTypeResolver.cpp
```diff
@@ -361,24 +366,42 @@ bool ReferencesResolver::visit(Mapping&)
...
 bool ReferencesResolver::visit(UserDefinedTypeName& _typeName)
 {
-	Declaration const* declaration = m_resolver.getNameFromCurrentScope(_typeName.getName());
-	if (!declaration)
+	auto declarations = m_resolver.getNameFromCurrentScope(_typeName.getName());
+	if (declarations.empty())
 		BOOST_THROW_EXCEPTION(DeclarationError() << errinfo_sourceLocation(_typeName.getLocation())
 												 << errinfo_comment("Undeclared identifier."));
-	_typeName.setReferencedDeclaration(*declaration);
+	else if (declarations.size() > 1)
```
chriseth: Wouldn't it be better to add a flag to `getNameFromCurrentScope` to check for these conditions?

libsolidity/Parser.cpp
```diff
@@ -792,7 +792,15 @@ ASTPointer<Expression> Parser::parseLeftHandSideExpression(
 			std::tie(arguments, names) = parseFunctionCallArguments();
 			nodeFactory.markEndPosition();
 			expectToken(Token::RParen);
-			expression = nodeFactory.createNode<FunctionCall>(expression, arguments, names);
+			if (Identifier* identifier = dynamic_cast<Identifier*>(expression.get()))
+			{
+				ASTPointer<FunctionIdentifier> functionIdentifier = make_shared<FunctionIdentifier>(identifier->getLocation(), make_shared<ASTString>(identifier->getName()));
+				auto functionCall = nodeFactory.createNode<FunctionCall>(functionIdentifier, arguments, names);
+				functionIdentifier->setFunctionCall(functionCall);
```
chriseth: This is the only place in the parser where a node is modified right after it was created (and not only during the "resole names and types" or "checkTypeRequirements" stage). Can this be done differently?

test/SolidityExpressionCompiler.cpp
```diff
@@ -79,7 +80,9 @@ Declaration const& resolveDeclaration(vector<string> const& _namespacedName,
 	// bracers are required, cause msvc couldnt handle this macro in for statement
 	for (string const& namePart: _namespacedName)
 	{
-		BOOST_REQUIRE(declaration = _resolver.resolveName(namePart, declaration));
+		auto declarations = _resolver.resolveName(namePart, declaration);
+		BOOST_REQUIRE(!declarations.empty());
+		BOOST_REQUIRE(declaration = *declarations.begin());
```
chriseth: `.front()`?

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -1185,6 +1187,18 @@ BOOST_AUTO_TEST_CASE(array_with_nonconstant_length)
 	BOOST_CHECK_THROW(parseTextAndResolveNames(text), TypeError);
 }
 
+BOOST_AUTO_TEST_CASE(overloaded_function_cannot_resolve)
```
chriseth: What about testing an ambiguous call?

chriseth:
This is really great!
I'm not really sure about is what happens if you first assign a function identifier to a local variable and then call it, as it is done here: https://github.com/ethereum/wiki/wiki/Solidity-Tutorial#esoteric-features
I would say it is fine if that is not possible anymore, but we should guard against such uses.

test/SolidityExpressionCompiler.cpp
guanqun: I'm using a `std::set` here, so no `.front()` method. It's a bit awkward to use `.begin()` and the dereference it.

test/SolidityNameAndTypeResolution.cpp
guanqun: I have one but forgot to cherry-pick into this branch, will add it.

guanqun:
Assigning a function identifier to a local variable and then call it with overloaded function doesn't work in this pull request. However, if this function is not an overloaded function, it would work, because it will default to use the first one, and there's only one.

This behaviour is ideal, I'll make it all work or disable them all.

libsolidity/Parser.cpp
guanqun: Good catch! This piece of code doesn't smell well...

Another approach I'm thinking about is I don't create a new `FunctionIdentifier` node, just use this `Identifier` node and then store declarations into this node. Let me check which is better.

libsolidity/NameAndTypeResolver.cpp
guanqun: Thank you. I'll move it into `checkTypeRequirements()`. The reason of another round check is: if the overloaded function is not used, there's nowhere to check for duplication, so it has to be called somewhere after the type requirements are checked.

libsolidity/AST.cpp
guanqun: OK.

libsolidity/AST.cpp
guanqun: Right... The named arguments with overloaded function is still missing.

libsolidity/AST.cpp
guanqun: Yep! Will fix it.

libsolidity/AST.cpp
guanqun: Nice catch! Rather reasonable!

libsolidity/AST.cpp
guanqun: Naming is hard. ;D

guanqun:
@chriseth please review again. thanks. this can support `var x = f;` where `f` is an overloaded function.

libsolidity/AST.cpp
```diff
@@ -76,6 +76,15 @@ void ContractDefinition::checkTypeRequirements()
 
 	for (ASTPointer<FunctionDefinition> const& function: getDefinedFunctions())
 		function->checkTypeRequirements();
+	// check for duplicate declaration
+	set<string> functions;
+	for (ASTPointer<FunctionDefinition> const& function: getDefinedFunctions())
+	{
+		string signature = function->getCanonicalSignature();
+		if (functions.count(signature))
+			BOOST_THROW_EXCEPTION(DeclarationError() << errinfo_comment("Duplicate functions are not allowed."));
```
chriseth: Please add the source location (of one of the functions) to the error.

libsolidity/AST.cpp
```diff
@@ -185,16 +196,21 @@ vector<pair<FixedHash<4>, FunctionTypePointer>> const& ContractDefinition::getIn
...
 	{
 		set<string> functionsSeen;
+		set<string> signaturesSeen;
 		m_interfaceFunctionList.reset(new vector<pair<FixedHash<4>, FunctionTypePointer>>());
 		for (ContractDefinition const* contract: getLinearizedBaseContracts())
 		{
 			for (ASTPointer<FunctionDefinition> const& f: contract->getDefinedFunctions())
-				if (f->isPublic() && !f->isConstructor() && !f->getName().empty() && functionsSeen.count(f->getName()) == 0)
+			{
+				string functionSignature = f->getCanonicalSignature();
```
chriseth: `const&`

guanqun: Do we really need to make this `functionSignature` a `const&`? the return type of `getCanonicalSignature()` is just a `std::string`.

libsolidity/AST.cpp
guanqun: OK.

guanqun:
rebased onto the latest develop branch.

libsolidity/AST.cpp
chriseth: Oh sorry, then it is of course fine.

libsolidity/AST.cpp
```diff
@@ -544,10 +566,16 @@ void BinaryOperation::checkTypeRequirements()
...
 void FunctionCall::checkTypeRequirements()
 {
-	m_expression->checkTypeRequirements();
+	// we need to check arguments' type first as their info will be used by m_express(Identifier).
 	for (ASTPointer<Expression> const& argument: m_arguments)
 		argument->checkTypeRequirements();
 
+	auto identifier = dynamic_cast<Identifier*>(m_expression.get());
+	if (identifier)
+		identifier->checkTypeRequirementsWithFunctionCall(*this);
```
chriseth: Another way to solve this would be to store from here inside Expression, which of the three cases we currently have, as it is done for lvalues, and then differentiate inside Identifier::checkTypeRequirements on that.
What do you think about that?

libsolidity/AST.cpp
```diff
@@ -617,6 +645,27 @@ void FunctionCall::checkTypeRequirements()
...
+
+		Declaration const* function = overloadedTypes->getIdentifier()->overloadResolution(*this);
+		if (!function)
+			BOOST_THROW_EXCEPTION(createTypeError("Can't resolve declarations"));
+
+		identifier->setReferencedDeclaration(*function);
+		identifier->checkTypeRequirements();
+
+		TypePointer type = identifier->getType();
+		FunctionType const* functionType = dynamic_cast<FunctionType const*>(type.get());
```
chriseth: Please either assert or do a `dynamic_cast<FunctionType const&>`.

libsolidity/AST.cpp
```diff
@@ -719,6 +790,51 @@ void Identifier::checkTypeRequirements()
...
+				 std::equal(arguments.cbegin(), arguments.cend(), parameterTypes.cbegin(),
+							[](ASTPointer<Expression const> const& argument, TypePointer const& parameterType)
+							{
+								return argument->getType()->isImplicitlyConvertibleTo(*parameterType);
+							})))
+				possibles.push_back(declaration);			
+		}
+		if (possibles.empty())
+			BOOST_THROW_EXCEPTION(createTypeError("Can't resolve identifier"));
+		else if (std::none_of(possibles.cbegin() + 1, possibles.cend(),
```
chriseth: Could you add a comment to exlpain this rule? I am not sure if we really want it.

libsolidity/AST.h
```diff
@@ -1151,7 +1151,15 @@ class Identifier: public PrimaryExpression
 	Declaration const* getReferencedDeclaration() const { return m_referencedDeclaration; }
 	ContractDefinition const* getCurrentContract() const { return m_currentContract; }
 
+	void setOverloadedDeclarations(std::set<Declaration const*> const& _declarations) { m_overloadedDeclarations = _declarations; }
+	std::set<Declaration const*> getOverloadedDeclarations() const { return m_overloadedDeclarations; }
+
+	void checkTypeRequirementsWithFunctionCall(FunctionCall const& _functionCall);
+	void checkTypeRequirementsFromVariableDeclaration();
+
+	Declaration const* overloadResolution(FunctionCall const& _functionCall);
```
chriseth: Please add some comments.

libsolidity/AST.h
```diff
@@ -1151,7 +1151,15 @@ class Identifier: public PrimaryExpression
 	Declaration const* getReferencedDeclaration() const { return m_referencedDeclaration; }
 	ContractDefinition const* getCurrentContract() const { return m_currentContract; }
 
+	void setOverloadedDeclarations(std::set<Declaration const*> const& _declarations) { m_overloadedDeclarations = _declarations; }
+	std::set<Declaration const*> getOverloadedDeclarations() const { return m_overloadedDeclarations; }
```
chriseth: Return const ref?

libsolidity/CompilerContext.cpp
```diff
@@ -106,8 +106,12 @@ eth::AssemblyItem CompilerContext::getVirtualFunctionEntryLabel(FunctionDefiniti
 	solAssert(!m_inheritanceHierarchy.empty(), "No inheritance hierarchy set.");
 	for (ContractDefinition const* contract: m_inheritanceHierarchy)
 		for (ASTPointer<FunctionDefinition> const& function: contract->getDefinedFunctions())
-			if (!function->isConstructor() && function->getName() == _function.getName())
+		{
+			if (!function->isConstructor() &&
+				dynamic_cast<FunctionType const&>(*function->getType(contract)).getCanonicalSignature() ==
```
chriseth: `contract` in `getType(contract)` is the contract the function is called from, so I think this should be removed here.

libsolidity/CompilerContext.cpp
```diff
@@ -117,7 +121,7 @@ eth::AssemblyItem CompilerContext::getSuperFunctionEntryLabel(string const& _nam
 	auto it = getSuperContract(_base);
 	for (; it != m_inheritanceHierarchy.end(); ++it)
 		for (ASTPointer<FunctionDefinition> const& function: (*it)->getDefinedFunctions())
-			if (!function->isConstructor() && function->getName() == _name)
+			if (!function->isConstructor() && function->getName() == _name)     // TODO: add a test case for this!
```
chriseth: Searching by name should not be correct here, but we can leave it for now.

libsolidity/DeclarationContainer.cpp
```diff
@@ -34,17 +35,35 @@ bool DeclarationContainer::registerDeclaration(Declaration const& _declaration,
 	if (name.empty())
 		return true;
 
-	if (!_update && (m_declarations.count(name) || m_invisibleDeclarations.count(name)))
-		return false;
+	if (!_update)
+	{
+		if (dynamic_cast<FunctionDefinition const*>(&_declaration))
```
chriseth: You no longer search inside `m_invisibleDeclarations`.

libsolidity/ExpressionCompiler.cpp
```diff
@@ -845,6 +845,13 @@ void ExpressionCompiler::endVisit(Identifier const& _identifier)
 	{
 		// no-op
 	}
+	else if (declaration == nullptr && _identifier.getOverloadedDeclarations().size() > 1)
```
chriseth: Perhaps put this case first so that we see that it actually can be null?

libsolidity/ExpressionCompiler.cpp
```diff
@@ -845,6 +845,13 @@ void ExpressionCompiler::endVisit(Identifier const& _identifier)
...
+	{
+		// var x = f;
+		declaration = *_identifier.getOverloadedDeclarations().begin();
+		FunctionDefinition const* functionDef = dynamic_cast<FunctionDefinition const*>(declaration);
+		m_context << m_context.getVirtualFunctionEntryLabel(*functionDef).pushTag();
```
chriseth: Assert functionDef?

libsolidity/NameAndTypeResolver.cpp
```diff
@@ -90,15 +90,15 @@ void NameAndTypeResolver::updateDeclaration(Declaration const& _declaration)
 	solAssert(_declaration.getScope() == nullptr, "Updated declaration outside global scope.");
 }
 
-Declaration const* NameAndTypeResolver::resolveName(ASTString const& _name, Declaration const* _scope) const
+std::set<Declaration const*> NameAndTypeResolver::resolveName(ASTString const& _name, Declaration const* _scope) const
 {
 	auto iterator = m_scopes.find(_scope);
 	if (iterator == end(m_scopes))
-		return nullptr;
+		return std::set<Declaration const*>({});
```
chriseth: What is the benefit of having `{}`?

libsolidity/NameAndTypeResolver.cpp
```diff
@@ -108,13 +108,11 @@ void NameAndTypeResolver::importInheritedScope(ContractDefinition const& _base)
...
-	{
-		Declaration const* declaration = nameAndDeclaration.second;
-		// Import if it was declared in the base, is not the constructor and is visible in derived classes
-		if (declaration->getScope() == &_base && declaration->getName() != _base.getName() &&
-				declaration->isVisibleInDerivedContracts())
-			m_currentScope->registerDeclaration(*declaration);
-	}
+		for (auto const& declaration: nameAndDeclaration.second)
+			// Import if it was declared in the base, is not the constructor and is visible in derived classes
+			if (declaration->getScope() == &_base && declaration->getName() != _base.getName() &&
```
chriseth: The name check can now be removed (it is already done in `isVisibleInDeriedContracts`).

libsolidity/NameAndTypeResolver.cpp
chriseth: Any thoughts on this?

test/SolidityEndToEndTest.cpp
```diff
@@ -3170,6 +3170,98 @@ BOOST_AUTO_TEST_CASE(pass_dynamic_arguments_to_the_base_base_with_gap)
...
+}
+
+BOOST_AUTO_TEST_CASE(overloaded_function_with_var)
+{
+	char const* sourceCode = R"(
+		contract test {
+			function f(uint k) returns(uint d) { return k; }
+			function f(uint a, uint b) returns(uint d) { return a + b; }
+			function g(bool flag) returns(uint d) {
+				var x = f;
```
chriseth: It is great that you made that work, but I doubt that the following works:
```
var x = f;
if (flag) x = g;
x(3);
```
(with changed signature of g of course). So I think the cleaner solution would be to disallow this completely (but that can be done in another PR), i.e. when a function is not directly called, its name has to be unique.

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -1287,6 +1287,31 @@ BOOST_AUTO_TEST_CASE(storage_variable_initialization_with_incorrect_type_string)
...
+			function f(uint a) returns(uint) { return a; }
+			function g() returns(uint) { return f(3, 5); }
+		}
+	)";
+	BOOST_CHECK_THROW(parseTextAndResolveNames(sourceCode), TypeError);
+}
+
+BOOST_AUTO_TEST_CASE(ambiguous_overloaded_function)
+{
+	// literal 1 can be both converted to uint8 and uint8, so it's ambiguous.
```
chriseth: Change on `uint8` to `uint` ;-)

chriseth:
One feature that might still be missing (at least there is no test for it):
```
contract c { function f(uint) {} function f(uint,uint) {}}
contract d is c { function f(uint) {} function test() { c.f(7); } }
```
should call `c`'s `f(uint)`.
But that can wait for another PR.

chriseth:
Continued as #1634.

PR #1169, opened 2015-03-01
Eth c++ client structured logging

LefterisJP:
Implementing the [Structured Logging Spec](https://github.com/ethereum/system-testing/wiki/Log-Events) for the C++ client.

* Some fields are missing since I am not 100% familiar with that part of the codebase but I am positive with some quick feedback they will filled in fast.

* At the moment starting the client with ``` ./eth/eth --structured-logging``` will enable structured logging. 
A nice bonus is that with the ```--structured-logging-format <%Y-%m-%d>``` we can give a [strftime()](http://en.cppreference.com/w/cpp/chrono/c/strftime) type of format string to influence the way timestamps are printed.

* I took the liberty of adding a `Stop` log event when the client shuts down. I think it would be needed.

* Finally I apologize for "infecting" the PR with style fixes but I tried to follow the style guide as I was touching the code here and there.

### DIFF
--- a/libdevcore/StructuredLogger.cpp
+++ b/libdevcore/StructuredLogger.cpp
@@ -54,7 +54,7 @@ void StructuredLogger::outputJson(Json::Value const& _value, std::string const& 
 	cout << event << endl << flush;
 }
 
-void StructuredLogger::logStarting(string const& _clientImpl, const char* _ethVersion) const
+void StructuredLogger::starting(string const& _clientImpl, const char* _ethVersion) const
 {
 	if (m_enabled)
 	{
@@ -67,7 +67,7 @@ void StructuredLogger::logStarting(string const& _clientImpl, const char* _ethVe
 	}
 }
 
-void StructuredLogger::logStopping(string const& _clientImpl, const char* _ethVersion) const
+void StructuredLogger::stopping(string const& _clientImpl, const char* _ethVersion) const
 {
 	if (m_enabled)
 	{
@@ -80,7 +80,7 @@ void StructuredLogger::logStopping(string const& _clientImpl, const char* _ethVe
 	}
 }
 
-void StructuredLogger::logP2PConnected(string const& _id, bi::tcp::endpoint const& _addr,
+void StructuredLogger::p2pConnected(string const& _id, bi::tcp::endpoint const& _addr,
 	chrono::system_clock::time_point const& _ts, string const& _remoteVersion, unsigned int _numConnections) const
 {
 	if (m_enabled)
@@ -98,7 +98,7 @@ void StructuredLogger::logP2PConnected(string const& _id, bi::tcp::endpoint cons
 	}
 }
 
-void StructuredLogger::logP2PDisconnected(string const& _id, bi::tcp::endpoint const& _addr, unsigned int _numConnections) const
+void StructuredLogger::p2pDisconnected(string const& _id, bi::tcp::endpoint const& _addr, unsigned int _numConnections) const
 {
 	if (m_enabled)
 	{
@@ -114,7 +114,7 @@ void StructuredLogger::logP2PDisconnected(string const& _id, bi::tcp::endpoint c
 	}
 }
 
-void StructuredLogger::logMinedNewBlock(string const& _hash, string const& _blockNumber,
+void StructuredLogger::minedNewBlock(string const& _hash, string const& _blockNumber,
 	string const& _chainHeadHash, string const& _prevHash) const
 {
 	if (m_enabled)
@@ -130,7 +130,7 @@ void StructuredLogger::logMinedNewBlock(string const& _hash, string const& _bloc
 	}
 }
 
-void StructuredLogger::logChainReceivedNewBlock(string const& _hash, string const& _blockNumber,
+void StructuredLogger::chainReceivedNewBlock(string const& _hash, string const& _blockNumber,
 	string const& _chainHeadHash, string const& _remoteID, string const& _prevHash) const
 {
 	if (m_enabled)
@@ -147,7 +147,7 @@ void StructuredLogger::logChainReceivedNewBlock(string const& _hash, string cons
 	}
 }
 
-void StructuredLogger::logChainNewHead(string const& _hash, string const& _blockNumber,
+void StructuredLogger::chainNewHead(string const& _hash, string const& _blockNumber,
 	string const& _chainHeadHash, string const& _prevHash) const
 {
 	if (m_enabled)
@@ -163,7 +163,7 @@ void StructuredLogger::logChainNewHead(string const& _hash, string const& _block
 	}
 }
 
-void StructuredLogger::logTransactionReceived(string const& _hash, string const& _remoteId) const
+void StructuredLogger::transactionReceived(string const& _hash, string const& _remoteId) const
 {
 	if (m_enabled)
 	{
--- a/libdevcore/StructuredLogger.cpp
+++ b/libdevcore/StructuredLogger.cpp
@@ -66,6 +66,20 @@ void StructuredLogger::logStarting(string const& _clientImpl, const char* _ethVe
 	}
 }
 
+void StructuredLogger::logStopping(string const& _clientImpl, const char* _ethVersion) const
+{
+	if (m_enabled)
+	{
+		Json::Value event;
+		event["comment"] = "last event before termination of the client";
+		event["client_impl"] = _clientImpl;
+		event["eth_version"] = std::string(_ethVersion);
+		event["ts"] = timePointToString(std::chrono::system_clock::now());
+
+		outputJson(event, "stopping");
+	}
+}
+
 void StructuredLogger::logP2PConnected(string const& _id, bi::tcp::endpoint const& _addr,
 	chrono::system_clock::time_point const& _ts, string const& _remoteVersion, unsigned int _numConnections) const
 {
--- a/libdevcore/StructuredLogger.cpp
+++ b/libdevcore/StructuredLogger.cpp
@@ -80,7 +80,8 @@ void StructuredLogger::stopping(string const& _clientImpl, const char* _ethVersi
 	}
 }
 
-void StructuredLogger::p2pConnected(string const& _id,
+void StructuredLogger::p2pConnected(
+	string const& _id,
 	bi::tcp::endpoint const& _addr,
 	chrono::system_clock::time_point const& _ts,
 	string const& _remoteVersion,
@@ -117,7 +118,8 @@ void StructuredLogger::p2pDisconnected(string const& _id, bi::tcp::endpoint cons
 	}
 }
 
-void StructuredLogger::minedNewBlock(string const& _hash,
+void StructuredLogger::minedNewBlock(
+	string const& _hash,
 	string const& _blockNumber,
 	string const& _chainHeadHash,
 	string const& _prevHash)
@@ -135,7 +137,8 @@ void StructuredLogger::minedNewBlock(string const& _hash,
 	}
 }
 
-void StructuredLogger::chainReceivedNewBlock(string const& _hash,
+void StructuredLogger::chainReceivedNewBlock(
+	string const& _hash,
 	string const& _blockNumber,
 	string const& _chainHeadHash,
 	string const& _remoteID,
@@ -155,7 +158,8 @@ void StructuredLogger::chainReceivedNewBlock(string const& _hash,
 	}
 }
 
-void StructuredLogger::chainNewHead(string const& _hash,
+void StructuredLogger::chainNewHead(
+	string const& _hash,
 	string const& _blockNumber,
 	string const& _chainHeadHash,
 	string const& _prevHash)
--- a/libdevcore/StructuredLogger.cpp
+++ b/libdevcore/StructuredLogger.cpp
@@ -80,8 +80,11 @@ void StructuredLogger::stopping(string const& _clientImpl, const char* _ethVersi
 	}
 }
 
-void StructuredLogger::p2pConnected(string const& _id, bi::tcp::endpoint const& _addr,
-	chrono::system_clock::time_point const& _ts, string const& _remoteVersion, unsigned int _numConnections) const
+void StructuredLogger::p2pConnected(string const& _id,
+	bi::tcp::endpoint const& _addr,
+	chrono::system_clock::time_point const& _ts,
+	string const& _remoteVersion,
+	unsigned int _numConnections) const
 {
 	if (m_enabled)
 	{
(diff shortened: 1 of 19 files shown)


libdevcore/StructuredLogger.cpp
```diff
@@ -0,0 +1,185 @@
...
+	event[_name] = _value;
+	cout << event << endl;
+}
+
+void StructuredLogger::logStarting(string const& _clientImpl, const char* _ethVersion) const
+{
+	if (m_enabled)
+	{
+		Json::Value event;
+		event["comment"] = "one of the first log events, before any operation is started";
```
subtly: Afaik this isn't implemented in the client and can be removed.

libdevcore/StructuredLogger.cpp
```diff
@@ -0,0 +1,185 @@
...
+	time_t time = chrono::system_clock::to_time_t(_ts);
+	tm* ptm = localtime(&time);
+	if (strftime(buffer, sizeof(buffer), m_timeFormat.c_str(), ptm))
+		return string(buffer);
+	return "";
+}
+
+void StructuredLogger::outputJson(Json::Value const& _value, std::string const& _name) const
+{
+	Json::Value event;
```
subtly: Can this be done before creating the detailed event via a helper method, rather than after?
This would eliminate this method. Since we're already instantiating a class for this whole thing, it's just an idea but Json::Value could be wrapped in a class whose destructor is cout << self << endl;. Also note: Look at how other logging is implemented. cout is not thread safe. Said destructor (or the output method) should have a mutex.

subtly:
This CLI parameters are okay. Instead of creating a new class for this please look at the existing logging implementation and build on those. Would recommend creating a StructuredLogOutputStream similar to LogOutputStream; remove operator<<, add a log(string&,unsigned) and log(string&,string&), and replace m_sstr with Json::Value m_event. Set fwd declration for Json::Value in Log.h and implement log() in Log.cpp so we don't have to change includes in every file.

libdevcore/StructuredLogger.cpp
LefterisJP: There is no reason to have this as a class member. Since its contects would need to change and be reinitialized every time the logging methods would be called. I had the same question when I used `Json::Value`s in the Solidity Interface handler.

I will look at the thread safety now.

LefterisJP:
I can understand the thread safety argument, but I really strongly believe that it should stay a separate class as it is right now.

I really don't like the idea of making another LogOutputStream. These 2 different kinds of logging systems serve totally different purposes. I will add a mutex to the current class to protect the output as the one in LogOutputStream does.

LefterisJP:
rebased

LefterisJP:
The 2 fields missing from the logging for now are:

* chain head has from `eth.miner.new_block` event
* remote id from `eth.chain.received.new_block` event

libdevcore/StructuredLogger.h
```diff
@@ -0,0 +1,89 @@
...
+
+	static StructuredLogger& get()
+	{
+		static StructuredLogger instance;
+		return instance;
+	}
+
+	void starting(std::string const& _clientImpl, const char* _ethVersion) const;
+	void stopping(std::string const& _clientImpl, const char* _ethVersion) const;
+	void p2pConnected(std::string const& _id, bi::tcp::endpoint const& _addr,
```
gavofyork: style is incorrect. see latest CodingStandards.txt, 0.a and the golden rule.

libdevcore/StructuredLogger.h
```diff
@@ -0,0 +1,95 @@
...
+private:
+	// Singleton class, no copying
+	StructuredLogger() {}
+	StructuredLogger(StructuredLogger const&) = delete;
+	void operator=(StructuredLogger const&) = delete;
+	/// @returns a string representation of a timepoint
+	static std::string timePointToString(std::chrono::system_clock::time_point const& _ts);
+	void outputJson(Json::Value const& _value, std::string const& _name) const;
+
+	bool m_enabled;
```
chfast: Uninitialized variable

libdevcore/StructuredLogger.h
```diff
@@ -0,0 +1,95 @@
...
+		std::string const& _remoteID,
+		std::string const& _prevHash);
+	static void chainNewHead(std::string const& _hash,
+		std::string const& _blockNumber,
+		std::string const& _chainHeadHash,
+		std::string const& _prevHash);
+	static void transactionReceived(std::string const& _hash, std::string const& _remoteId);
+private:
+	// Singleton class, no copying
+	StructuredLogger() {}
```
chfast: That means `= default`

libdevcore/StructuredLogger.h
LefterisJP: This is a singleton class and is only supposed to be used after calling `initialize()` and then just using the static methods.

But yeah it does not hurt to have a default value here.

libdevcore/StructuredLogger.h
LefterisJP: I know

LefterisJP:
Can anyone try to build this locally? I have no idea why buildbot fails to find json/json.h. It makes no sense. We use it in other files too.

onepremise:
I've seen this happen before with builds which reference jsoncpp in nonstandard path. I had to explicitly include jsoncpp before /usr/include because it was getting confused with distros that have the standard jsonc library installed, "-I/include/jsoncpp -I/include". You could try changing '#include json/json.h" to "#include jsoncpp/json/json.h". That is if the build machine was recently updated and jsoncpp is now in the path it's usually installed at on most distributions.

LefterisJP:
I see. Thanks for the info @onepremise. I'd rather not have to do black magic to appease the buildbot though. I guess the buildbot server itself needs some kind of updating? Will investigate further.

onepremise:
```
 I'd rather not have to do black magic to appease the buildbot 
```
LefterisJP, wasn't trying to point this out as a 'black magic' fix. It really is a distribution issue. Most linux distros will have installs which deploy to /usr/include/json and /usr/include/jsoncpp. If you include in your source files, '#include json/json.h', there's no guarantee that others will be able to build your project without encountering this same issue. I know i did when i started building cpp-ethereum. They will have to explicitly know to include jsonccp first in their include path before this succeeds, that is if they have both jsonc and jsoncpp installed. Many open source projects have deferred to referencing jsoncpp in their source files to avoid this issue. Others, using cmake, have updated their FindJsoncpp.cmake file to search for features.h instead of json.h:

http://stackoverflow.com/questions/18005880/how-to-writing-a-cmake-module-for-jsoncpp

Luckily your buildbot environment isolates the issue.

LefterisJP:
This is why we have cmake. The fix was to adjust the cmakefile for libdevcore. All seems to be sorted.

I did not want any discrepancies between source files. All of our jsoncpp includes should be `json/json.h` and let cmake sort out the rest.

libwebthree/WebThree.h
```diff
@@ -103,7 +103,10 @@ class WebThreeDirect : public WebThreeNetworkFace
 public:
 	/// Constructor for private instance. If there is already another process on the machine using @a _dbPath, then this will throw an exception.
 	/// ethereum() may be safely static_cast()ed to a eth::Client*.
-	WebThreeDirect(std::string const& _clientVersion, std::string const& _dbPath, bool _forceClean = false, std::set<std::string> const& _interfaces = {"eth", "shh"}, p2p::NetworkPreferences const& _n = p2p::NetworkPreferences(), bytesConstRef _network = bytesConstRef(), int miners = -1);
+	WebThreeDirect(std::string const& _clientVersion, std::string const& _dbPath, bool _forceClean = false,
+				   std::set<std::string> const& _interfaces = {"eth", "shh"},
```
gavofyork: style (one indent only)

libwebthree/WebThree.cpp
```diff
@@ -35,15 +35,16 @@ using namespace dev::p2p;
 using namespace dev::eth;
 using namespace dev::shh;
 
-WebThreeDirect::WebThreeDirect(std::string const& _clientVersion, std::string const& _dbPath, bool _forceClean, std::set<std::string> const& _interfaces, NetworkPreferences const& _n, bytesConstRef _network, int miners):
-	m_clientVersion(_clientVersion),
-	m_net(_clientVersion, _n, _network)
+WebThreeDirect::WebThreeDirect(std::string const& _clientVersion, std::string const& _dbPath, bool _forceClean,
+	std::set<std::string> const& _interfaces, NetworkPreferences const& _n,
+	bytesConstRef _network, int _miners):
+		m_clientVersion(_clientVersion),
```
gavofyork: style (one indent only)

libp2p/Host.h
```diff
@@ -82,23 +82,24 @@ class Host: public Worker
 	friend class HostNodeTableHandler;
 	friend class Session;
 	friend class HostCapabilityFace;
-	
+
 public:
 	/// Start server, listening for connections on the given port.
-	Host(std::string const& _clientVersion, NetworkPreferences const& _n = NetworkPreferences(), bytesConstRef _restoreNetwork = bytesConstRef());
+	Host(std::string const& _clientVersion, NetworkPreferences const& _n = NetworkPreferences(),
```
gavofyork: style (one indent for all args)

libethereum/Client.h
```diff
@@ -166,7 +167,8 @@ class Client: public MinerHost, public Interface, Worker
 
 public:
 	/// New-style Constructor.
-	explicit Client(p2p::Host* _host, std::string const& _dbPath = std::string(), bool _forceClean = false, u256 _networkId = 0, int miners = -1);
+	explicit Client(p2p::Host* _host, std::string const& _dbPath = std::string(), bool _forceClean = false,
```
gavofyork: style (one indent for all args)

libdevcore/StructuredLogger.h
```diff
@@ -0,0 +1,96 @@
...
+
+	static StructuredLogger& get()
+	{
+		static StructuredLogger instance;
+		return instance;
+	}
+
+	static void starting(std::string const& _clientImpl, const char* _ethVersion);
+	static void stopping(std::string const& _clientImpl, const char* _ethVersion);
+	static void p2pConnected(std::string const& _id,
```
gavofyork: style (first arg not on own line)

gavofyork:
style fixes needed. simple rule: if an expression goes on multiple lines, it's always:
```
PRE
    PART1 SEP
    PART2 SEP
    PART3 SEP
    ...
POST
```

note, this is not:
```
PRE PART1 SEP
    PART2 SEP
    PART3 SEP
    ...
POST
```

nor

```
PRE PART1 SEP PART2 SEP
    PART3 SEP ...
POST
```

`SEP` is `,`, `;` &c.

`PART`s are arguments, initialiser lines &c.

libwebthree/WebThree.h
```diff
@@ -103,7 +103,14 @@ class WebThreeDirect : public WebThreeNetworkFace
 public:
 	/// Constructor for private instance. If there is already another process on the machine using @a _dbPath, then this will throw an exception.
 	/// ethereum() may be safely static_cast()ed to a eth::Client*.
-	WebThreeDirect(std::string const& _clientVersion, std::string const& _dbPath, bool _forceClean = false, std::set<std::string> const& _interfaces = {"eth", "shh"}, p2p::NetworkPreferences const& _n = p2p::NetworkPreferences(), bytesConstRef _network = bytesConstRef(), int miners = -1);
+	WebThreeDirect(
+		std::string const& _clientVersion, std::string const& _dbPath,
```
gavofyork: two items on one line

LefterisJP: arrggh sorry

libdevcore/StructuredLogger.cpp
```diff
@@ -0,0 +1,190 @@
...
+		event["block_number"] = _blockNumber;
+		event["chain_head_hash"] = _chainHeadHash;
+		event["ts"] = timePointToString(std::chrono::system_clock::now());
+		event["block_prev_hash"] = _prevHash;
+
+		get().outputJson(event, "eth.miner.new_block");
+	}
+}
+
+void StructuredLogger::chainReceivedNewBlock(string const& _hash,
```
gavofyork: still wrong.

libdevcore/StructuredLogger.cpp
```diff
@@ -0,0 +1,190 @@
...
+		event["remote_addr"] = addrStream.str();
+		event["remote_id"] = _id;
+		event["num_connections"] = Json::Value(_numConnections);
+		event["ts"] = timePointToString(chrono::system_clock::now());
+
+		get().outputJson(event, "p2p.disconnected");
+	}
+}
+
+void StructuredLogger::minedNewBlock(string const& _hash,
```
gavofyork: still wrong.

libdevcore/StructuredLogger.cpp
```diff
@@ -0,0 +1,190 @@
...
+		event["remote_addr"] = addrStream.str();
+		event["remote_id"] = _id;
+		event["num_connections"] = Json::Value(_numConnections);
+		event["ts"] = timePointToString(_ts);
+
+		get().outputJson(event, "p2p.connected");
+	}
+}
+
+void StructuredLogger::p2pDisconnected(string const& _id, bi::tcp::endpoint const& _addr, unsigned int _numConnections)
```
gavofyork: line not too long?

libdevcore/StructuredLogger.cpp
```diff
@@ -0,0 +1,190 @@
...
+		Json::Value event;
+		event["client_impl"] = _clientImpl;
+		event["eth_version"] = std::string(_ethVersion);
+		event["ts"] = timePointToString(std::chrono::system_clock::now());
+
+		get().outputJson(event, "stopping");
+	}
+}
+
+void StructuredLogger::p2pConnected(string const& _id,
```
gavofyork: still wrong.

libp2p/Host.h
```diff
@@ -82,23 +82,26 @@ class Host: public Worker
 	friend class HostNodeTableHandler;
 	friend class Session;
 	friend class HostCapabilityFace;
-	
+
 public:
 	/// Start server, listening for connections on the given port.
-	Host(std::string const& _clientVersion, NetworkPreferences const& _n = NetworkPreferences(), bytesConstRef _restoreNetwork = bytesConstRef());
+	Host(std::string const& _clientVersion,
```
gavofyork: still wrong.

libwebthree/WebThree.h
```diff
@@ -103,7 +103,15 @@ class WebThreeDirect : public WebThreeNetworkFace
 public:
 	/// Constructor for private instance. If there is already another process on the machine using @a _dbPath, then this will throw an exception.
 	/// ethereum() may be safely static_cast()ed to a eth::Client*.
-	WebThreeDirect(std::string const& _clientVersion, std::string const& _dbPath, bool _forceClean = false, std::set<std::string> const& _interfaces = {"eth", "shh"}, p2p::NetworkPreferences const& _n = p2p::NetworkPreferences(), bytesConstRef _network = bytesConstRef(), int miners = -1);
+	WebThreeDirect(
+		std::string const& _clientVersion,
+        std::string const& _dbPath,
```
gavofyork: indent missing.

libdevcore/StructuredLogger.cpp
LefterisJP: It's barely okay. In my other projects I try to follow the 80 columns rule. In our codebase here 120 would be some form of limit where it just becomes unreadable. This is 118

PR #1174, opened 2015-03-02
let the error output take tab into consideration

guanqun:
if the source code in solidity has leading tabs, the output of `solc` is incorrect, this PR fixes this.


chriseth:
Thanks, that was a big source of annoyance :-)

PR #1175, opened 2015-03-02
transaction hash in event logs, second approach

debris:
this commit will add transaction hash to event logs (as proposed [here](https://github.com/ethereum/wiki/wiki/JSON-RPC-Improvement-Proposal#eth_logs-eth_filterlogs-eth_logs-and-eth_changed))


libethereum/Client.cpp
```diff
@@ -782,11 +786,15 @@ LocalisedLogEntries Client::logs(LogFilter const& _f) const
...
-			for (TransactionReceipt receipt: m_bc.receipts(h).receipts)
+		auto info = m_bc.info(h);
+		auto receipts = m_bc.receipts(h).receipts;
+		if (_f.matches(info.logBloom))
+			for (size_t i = 0; i < receipts.size(); i++)
 			{
+				TransactionReceipt receipt = receipts[i];
 				if (_f.matches(receipt.bloom()))
 				{
+					auto sha3 = transaction(info.hash, i).sha3();
```
gavofyork: better to use e.g. `h` rather than `sha3`, since the latter is a well-used symbol for the SHA3 function in `dev::`.

libevm/ExtVMFace.h
```diff
@@ -63,9 +63,10 @@ using LogEntries = std::vector<LogEntry>;
 struct LocalisedLogEntry: public LogEntry
 {
 	LocalisedLogEntry() {}
-	LocalisedLogEntry(LogEntry const& _le, unsigned _number): LogEntry(_le), number(_number) {}
+	LocalisedLogEntry(LogEntry const& _le, unsigned _number, h256 _sha3 = h256(u256(0))): LogEntry(_le), number(_number), sha3(_sha3) {}
```
gavofyork: `h256(u256(0))` == `h256()` == `{}`

PR #1179, opened 2015-03-02
add secure trie tests

CJentzsch:


CJentzsch:
SecureGenericTrieDB does not exist any more, and https://github.com/ethereum/cpp-ethereum/commit/e9e310822b65e82e8ad66389493c52832478092e fixes it.

PR #1187, opened 2015-03-02
Accounts management

yann300:


mix/ClientModel.h
```diff
@@ -27,6 +27,7 @@
 #include <map>
 #include <QString>
 #include <QVariantMap>
+#include "QBigInt.h"
```
arkpar: Does not seem to be needed here

mix/MixClient.cpp
```diff
@@ -62,16 +62,18 @@ class MixBlockChain: public dev::eth::BlockChain
 };
 
 MixClient::MixClient(std::string const& _dbPath):
-	m_userAccount(c_userAccountSecret), m_dbPath(_dbPath), m_minigThreads(0)
+	m_dbPath(_dbPath), m_minigThreads(0)
 {
-	resetState(10000000 * ether);
+	std::map<Secret, u256> account;
```
arkpar: accounts or accountMap

mix/qml/StateDialog.qml
```diff
@@ -90,15 +100,112 @@ Window {
...
+								width: parent.width
+								Button
+								{
+									iconSource: "qrc:/qml/img/delete_sign.png"
+									action: deleteAccountAction
+								}
+
+								Action {
+									id: deleteAccountAction
+									tooltip: qsTr("delete Account")
```
arkpar: capitalize?

PR #1199, opened 2015-03-03
Block chain tests + bugfix BlockGasLimit check

CJentzsch:
The existing block tests are replaced/restructered by/to blockchain tests which can handle not just one block, but a chain of blocks.


CJentzsch:
needs to be adapted to latest PoC9 changes... (in progress)

PR #1200, opened 2015-03-04
ethereum.js 0.0.16

debris:
* constants should be before array contents && (dynamic) string contents
* separated signatures functions from `abi.js`
* event hash


gavofyork:
fails on evmjit?

caktux:
Random npm failure again, triggered a rebuild.

PR #1202, opened 2015-03-04
Test HEX fields standart fix

winsvega:
work in progress to prefix all HEX fields with "0x..."

### DIFF
--- a/libdevcore/CommonData.cpp
+++ b/libdevcore/CommonData.cpp
@@ -80,7 +80,11 @@ int dev::fromHex(char _i)
 
 bytes dev::fromHex(std::string const& _s, ThrowType _throw)
 {
-	unsigned s = (_s[0] == '0' && _s[1] == 'x') ? 2 : 0;
+	bool isPrefix = (_s[0] == '0' && _s[1] == 'x');
+	if (!isPrefix)
+		cwarn << "WARNING: Reading HEX string without 0x prefix! (deprecated)";
+
+	unsigned s = isPrefix ? 2 : 0;
 	std::vector<uint8_t> ret;
 	ret.reserve((_s.size() - s + 1) / 2);
 
--- a/test/TestHelper.cpp
+++ b/test/TestHelper.cpp
@@ -67,6 +67,33 @@ namespace test
 struct ValueTooLarge: virtual Exception {};
 bigint const c_max256plus1 = bigint(1) << 256;
 
+void fixHexDataField(json_spirit::mObject& _o, string const& _sField)
+{
+	BOOST_REQUIRE(_o.count(_sField) > 0);
+	string str = _o[_sField].get_str();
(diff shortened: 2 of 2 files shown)


libdevcore/CommonData.cpp
```diff
@@ -80,7 +80,11 @@ int dev::fromHex(char _i)
 
 bytes dev::fromHex(std::string const& _s, ThrowType _throw)
 {
-	unsigned s = (_s[0] == '0' && _s[1] == 'x') ? 2 : 0;
+	bool isPrefix = (_s[0] == '0' && _s[1] == 'x');
```
CJentzsch: you dont need to change that file

test/TestHelper.cpp
```diff
@@ -255,7 +282,11 @@ byte toByte(json_spirit::mValue const& _v)
 
 bytes importByteArray(std::string const& _str)
 {
-	return fromHex(_str.substr(0, 2) == "0x" ? _str.substr(2) : _str, ThrowType::Throw);
```
CJentzsch: import should not change, only output.

test/TestHelper.cpp
```diff
@@ -276,7 +307,7 @@ bytes importCode(json_spirit::mObject& _o)
 		if (_o["code"].get_str().find_first_of("0x") != 0)
 			code = compileLLL(_o["code"].get_str(), false);
 		else
-			code = fromHex(_o["code"].get_str().substr(2));
```
CJentzsch: import should not change, only output.

test/TestHelper.cpp
```diff
@@ -335,8 +366,6 @@ void checkOutput(bytes const& _output, json_spirit::mObject& _o)
 			BOOST_CHECK_MESSAGE(_output[j] == toInt(d), "Output byte [" << j << "] different!");
 			++j;
 		}
-	else if (_o["out"].get_str().find("0x") == 0)
```
CJentzsch: import should not change, only output.

test/TestHelper.cpp
```diff
@@ -491,6 +520,31 @@ void executeTests(const string& _name, const string& _testPathAppendix, std::fun
 			}
 			break;
 		}
+
+		if (arg == "--fixhexstring")
```
CJentzsch: we do not need a commandline argument for that.

libdevcore/CommonData.cpp
winsvega: this is temporary changes. 
the idea is to make fillers HEX fields starting with "0x.."
then it would be easier to make every test in Test repository to satisfy this condition.

1) each "data" starting with 0x
2) each "to", address field starting with 0x
3) each hash field prefixed 0x 

as I understand other clients are having issues when reading .json test files hex string lines

PR #1203, opened 2015-03-04
Transaction Test Fillers POC9

winsvega:
gas fees increased, most tests are incorrect now
need to review (wip)


CJentzsch:
incorrect is the wrong word, some tests dont test what they should anymore. But they are still valid.
Please add the remaining tests (VMTests and StateTests)

PR #1205, opened 2015-03-04
Added source lines to output

LianaHus:
todo: should figure out better output format. maybe should align them to the right or add positions.

### DIFF
--- a/libevmcore/Assembly.cpp
+++ b/libevmcore/Assembly.cpp
@@ -171,8 +171,9 @@ ostream& dev::eth::operator<<(ostream& _out, AssemblyItemsConstRef _i)
 	return _out;
 }
 
-ostream& Assembly::streamRLP(ostream& _out, string const& _prefix) const
+ostream& Assembly::streamRLP(ostream& _out, string const& _prefix, StringMap _sourceCodes) const
 {
+	(void)_sourceCodes;
 	_out << _prefix << ".code:" << endl;
 	for (AssemblyItem const& i: m_items)
 		switch (i.m_type)
--- a/libevmcore/Assembly.cpp
+++ b/libevmcore/Assembly.cpp
@@ -182,16 +182,16 @@ string Assembly::getLocationFromSources(StringMap const& _sourceCodes, SourceLoc
 	if (it == _sourceCodes.end())
 		return "";
 
-	string toReturn = it->second;
-	if (_location.start >= (int)toReturn.size())
+	string const& source = it->second;
+	if (size_t(_location.start) >= source.size())
 		return "";
 
-	toReturn = toReturn.substr(_location.start, _location.end - _location.start);
-	auto newLinePos = toReturn.find_first_of("\n");
-	if (newLinePos != string::npos && newLinePos != toReturn.size() - 1)
-		toReturn = toReturn.substr(0, newLinePos) + "...";
+	string cut = source.substr(_location.start, _location.end - _location.start);
+	auto newLinePos = cut.find_first_of("\n");
+	if (newLinePos != string::npos)
+		cut = cut.substr(0, newLinePos) + "...";
 
-	return move(toReturn);
+	return move(cut);
 }
 
 ostream& Assembly::streamRLP(ostream& _out, string const& _prefix, StringMap const& _sourceCodes) const
@@ -239,7 +239,7 @@ ostream& Assembly::streamRLP(ostream& _out, string const& _prefix, StringMap con
 		default:
 			BOOST_THROW_EXCEPTION(InvalidOpcode());
 		}
-		_out << string(20, ' ') << sourceLine << endl;
+		_out << string("\t\t") << sourceLine << endl;
 	}
 
 	if (!m_data.empty() || !m_subs.empty())
--- a/libevmcore/Assembly.cpp
+++ b/libevmcore/Assembly.cpp
@@ -21,6 +21,8 @@
 
 #include "Assembly.h"
 
+#include <fstream>
+
 #include <libdevcore/Log.h>
 
 using namespace std;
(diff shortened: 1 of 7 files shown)


gavofyork:
fails on windows?

libevmcore/Assembly.cpp
```diff
@@ -171,61 +173,96 @@ ostream& dev::eth::operator<<(ostream& _out, AssemblyItemsConstRef _i)
...
+		toReturn = toReturn.substr(0, newLinePos) + "...";
+
+	toReturn = "        " + toReturn;
+	return move(toReturn);
+}
+
+ostream& Assembly::streamRLP(ostream& _out, string const& _prefix, StringMap const& _sourceCodes) const
 {
-	_out << _prefix << ".code:" << endl;
+	auto appendLineWithPrefix = [&](string const& __sData, string const& _sourceCodeLine = string()) {
```
chriseth: Better not use a lambda here at all, this will also fix the compilation problems on Windows.

libevmcore/Assembly.cpp
```diff
@@ -171,61 +173,96 @@ ostream& dev::eth::operator<<(ostream& _out, AssemblyItemsConstRef _i)
...
-			_out << _prefix << "  PUSH #[$" << h256(i.m_data).abridged() << "]" << endl;
+			lineToPrint << "  PUSH #[$" << h256(i.m_data).abridged() << "]";
 			break;
 		case PushProgramSize:
-			_out << _prefix << "  PUSHSIZE" << endl;
+			lineToPrint << "  PUSHSIZE";
 			break;
 		case Tag:
-			_out << _prefix << "tag" << i.m_data << ": " << endl << _prefix << "  JUMPDEST" << endl;
+			lineToPrint << "tag" << i.m_data <<  ": " << endl << _prefix << "  JUMPDEST";
```
chriseth: Added space :-)

libevmcore/Assembly.cpp
```diff
@@ -171,61 +173,96 @@ ostream& dev::eth::operator<<(ostream& _out, AssemblyItemsConstRef _i)
 	return _out;
 }
 
-ostream& Assembly::streamRLP(ostream& _out, string const& _prefix) const
+string Assembly::getLocationFromSources(StringMap const& _sourceCodes, SourceLocation const& _location) const
+{
+	if (_location.isEmpty() || _sourceCodes.empty())
+		return "";
+
+	string toReturn = _sourceCodes.find(*_location.sourceName)->second;
```
chriseth: Please first check whether the key is there.

libevmcore/Assembly.cpp
```diff
@@ -171,61 +173,96 @@ ostream& dev::eth::operator<<(ostream& _out, AssemblyItemsConstRef _i)
...
+
+	if (_location.end < (int)toReturn.size() - 1)
+			return "";
+
+	toReturn = toReturn.substr(_location.start, _location.end);
```
chriseth: I think it should be `substr(_location.start, _location.end - _location.start);`.

libevmcore/Assembly.cpp
```diff
@@ -171,61 +173,96 @@ ostream& dev::eth::operator<<(ostream& _out, AssemblyItemsConstRef _i)
...
 
-ostream& Assembly::streamRLP(ostream& _out, string const& _prefix) const
+string Assembly::getLocationFromSources(StringMap const& _sourceCodes, SourceLocation const& _location) const
+{
+	if (_location.isEmpty() || _sourceCodes.empty())
+		return "";
+
+	string toReturn = _sourceCodes.find(*_location.sourceName)->second;
+
+	if (_location.end < (int)toReturn.size() - 1)
```
chriseth: I think this should rather check for a valid `_start`, as `substr` throws for an invalid start. Furthermore, you could also check for a valid location (start nonnegative and smaller than end) here.

libevmcore/Assembly.cpp
```diff
@@ -171,61 +173,96 @@ ostream& dev::eth::operator<<(ostream& _out, AssemblyItemsConstRef _i)
...
+			return "";
+
+	toReturn = toReturn.substr(_location.start, _location.end);
+	auto newLinePos = toReturn.find_first_of("\n");
+	if (newLinePos != string::npos && newLinePos != toReturn.size() - 1)
```
chriseth: Please remove the second check, the string should not contain a newline even if it is the last character.

libevmcore/Assembly.cpp
```diff
@@ -171,48 +173,74 @@ ostream& dev::eth::operator<<(ostream& _out, AssemblyItemsConstRef _i)
...
+string Assembly::getLocationFromSources(StringMap const& _sourceCodes, SourceLocation const& _location) const
+{
+	if (_location.isEmpty() || _sourceCodes.empty() || _location.start >= _location.end || _location.start < 0)
+		return "";
+
+	auto it = _sourceCodes.find(*_location.sourceName);
+	if (it == _sourceCodes.end())
+		return "";
+
+	string toReturn = it->second;
```
chriseth: This copies the whole source file. Can you use a `const&` here?

libevmcore/Assembly.cpp
```diff
@@ -171,48 +173,74 @@ ostream& dev::eth::operator<<(ostream& _out, AssemblyItemsConstRef _i)
...
+	if (_location.start >= (int)toReturn.size())
```
chriseth: Please don't use c-style casts and perhaps it would be better to cast the left side to size_t.

libevmcore/Assembly.cpp
```diff
@@ -171,48 +173,74 @@ ostream& dev::eth::operator<<(ostream& _out, AssemblyItemsConstRef _i)
...
+		return "";
+
+	toReturn = toReturn.substr(_location.start, _location.end - _location.start);
+	auto newLinePos = toReturn.find_first_of("\n");
+	if (newLinePos != string::npos && newLinePos != toReturn.size() - 1)
```
chriseth: I think this condition is still wrong. If the last character is a newline, it will be included in `toReturn`.

libevmcore/Assembly.cpp
```diff
@@ -171,48 +173,74 @@ ostream& dev::eth::operator<<(ostream& _out, AssemblyItemsConstRef _i)
...
+			_out << "DoNotOptimze{{";
 			break;
 		case NoOptimizeEnd:
-			_out << _prefix << "DoNotOptimze}}" << endl;
+			_out << "DoNotOptimze}}";
 			break;
 		default:
 			BOOST_THROW_EXCEPTION(InvalidOpcode());
 		}
+		_out << string(20, ' ') << sourceLine << endl;
```
chriseth: Perhaps two tabs? Then it at least aligns most of the time.

libevmcore/Assembly.h
```diff
@@ -116,18 +114,17 @@ class Assembly
...
 	void setDeposit(int _deposit) { m_deposit = _deposit; if (asserts(m_deposit >= 0)) BOOST_THROW_EXCEPTION(InvalidDeposit()); }
 
 	bytes assemble() const;
 	Assembly& optimise(bool _enable);
-	std::ostream& streamRLP(std::ostream& _out, std::string const& _prefix = "") const;
+	std::ostream& streamRLP(std::ostream& _out, std::string const& _prefix = "", const StringMap &_sourceCodes = StringMap()) const;
 
-private:
+protected:
+	std::string getLocationFromSources(const StringMap &_sourceCodes, const SourceLocation &_location) const;
```
chriseth: move `const` to the right

PR #1208, opened 2015-03-04
Mix: solidity debugger inspectors

arkpar:


mix/CodeModel.h
```diff
@@ -64,6 +64,29 @@ public slots:
 	CodeModel* m_model;
 };
 
+using LocationPair = QPair<int, int>;
+
+struct SolidityType
```
chriseth: Do you really want to create new structs there that are needed to be kept in sync with libsolidity/Types.h?

arkpar: Solidity types hierarchy is too complex and does not directly map to what we handle in Mix. Mix only cares about how to present a value of a type to the user, so this is more like type formatting info. It is not that much work to keep it in sync since the code to support new types would need to be added anyway.

mix/CodeModel.cpp
```diff
@@ -43,6 +46,91 @@ using namespace dev::mix;
...
+		QHash<unsigned, SolidityDeclaration>* m_storage;
+		bool m_functionScope;
+		uint m_storageSlot;
+
+		LocationPair nodeLocation(ASTNode const& _node)
+		{
+			return LocationPair(_node.getLocation().start, _node.getLocation().end);
+		}
+
+		SolidityType nodeType(Type const* type)
```
yann300: missing _

mix/DebuggingStateWrapper.cpp
```diff
@@ -69,7 +69,7 @@ namespace
 	}
 }
 
-QCode* QMachineState::getHumanReadableCode(QObject* _owner, const Address& _address, const bytes& _code)
+QCode* QMachineState::getHumanReadableCode(QObject* _owner, const Address& _address, const bytes& _code, QHash<int, int>& o_codeMap)
```
yann300: o_codeMap ?

arkpar: Yes, It's a coding standard for output parameters

PR #1211, opened 2015-03-05
new jsonrpc

debris:
**changes:**
* updated json-rpc method names and input params according to [wiki](https://github.com/ethereum/wiki/wiki/JSON-RPC)
* all integers are now passed as hex
* `eth_getTransactionByHash` - we can lookup for transaction just by passing it's hash.
* new names for eth filter options: `fromBlock`, `toBlock`, `limit`, `offset`, `address` && `topics`
* if we want to get state for latest block we pass string `latest` instead of -1
* if there is a problem with decoding input params (eg. hex), we send standard jsonrpc error response:

  ```json
{
  "error": {
    "code": -32602,
    "message": "INVALID_PARAMS: Invalid method parameters (invalid name and/or type) recognised"
  },
  "id": 0,
  "jsonrpc": "2.0"
}
  ```

* `eth_getBlockByHash` && `eth_getBlockByNumber` may also return array of Transactions 

**other improvements worth considering:**
* return an error instead of empty result when polling a removed filter (suggested by @janx)
* same for proxy accounts
* and any other case when input params are incorrect
* not reuse filter id, to avoid confusion (suggested by @janx)
* `eth_changed` return a transaction two times, on pending and mined, can we add a status field in result so requester can distinguish them (suggested by @janx)


janx:
If I understand correctly - all integers in json must be hex after this change? why is this better than decimals?

caktux:
I have to object to further breaking changes to JSON-RPC unless absolutely necessary, and I also simply don't see the point in requiring hex numbers, cpp can very well handle int strings, and the convenience is quite important.

gavofyork:
It's for the sake of consistency; sometimes we really need bigints. Implementations should accept strings of "0x"-prefixed hex or string representations of decimals.

For returning, we'll just return hex.

caktux:
Was confused by your answer on first read, but I think we're saying the same thing. Return values as hex, inputs can be either as it is now.

libethereum/Client.cpp
```diff
@@ -828,6 +833,31 @@ unsigned Client::uncleCount(h256 _blockHash) const
...
+
+TransactionHashes Client::transactionHashes(h256 _blockHash) const
+{
+	return m_bc.transactionHashes(_blockHash);
+}
+
+LocalisedLogEntries Client::logs(unsigned _watchId) const
+{
+	try {
+		Guard l(m_filterLock);
```
subtly: LogFilter f;
{
    Guard l(m_filterLock);
    f = m_filters.at(m_watches.at(_watchId).id).filter);
}
return logs(f);

^ Recommend releasing mutex as quickly as possible.

libjsqrc/ethereumjs/.travis.yml
```diff
@@ -1,13 +1,17 @@
 language: node_js
 node_js:
+  - "0.12"
   - "0.11"
   - "0.10"
 before_script:
   - npm install
   - npm install jshint
+  - export DISPLAY=:99.0 
+  - sh -e /etc/init.d/xvfb start
```
subtly: Why?
Can we move these assets, deployment scripts, CI configs, etc. into another repos which pulls in ethereum as a submodule?  We're pulling in 5-15k+ LOC of JS every time small changes are made to those scripts and they're re-"compiled". Instead, a separate project would need to pull in far fewer files from cpp-ethereum, and changes are made, it doesn't require a 5k+ LOC PR.

libweb3jsonrpc/WebThreeStubServerBase.cpp
```diff
@@ -124,35 +145,30 @@ static dev::eth::LogFilter toLogFilter(Json::Value const& _json)	// commented to
...
+		for (auto t: _json["topics"])
 		{
-			if (t.isArray())
-				for (auto tt: t)
-					filter.topic(i, jsToFixed<32>(tt.asString()));
-			else if (t.isString())
-				filter.topic(i, jsToFixed<32>(t.asString()));
+			for (auto tt: t)
+				filter.topic(i, jsToFixed<32>(tt.asString()));
 			i++;
```
subtly: The result of this is never used. This line, along with the outer-for and outer-if curlies, and 'unsigned i = 0', can be removed.

libweb3jsonrpc/WebThreeStubServerBase.cpp
```diff
@@ -561,91 +756,325 @@ std::string WebThreeStubServerBase::eth_solidity(std::string const& _code)
...
+	catch (...)
+	{
+		BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS));
+	}
+
+	// TODO: throw an error on no account with given id
+	return m_accounts->removeProxyAccount(id);
+}
+
+Json::Value WebThreeStubServerBase::eth_queuedTransactions(string const& _accountId)
```
subtly: All of the other methods have "get" in front of them now. Why not this one?

subtly: fetchQueuedTransactions maybe a good name (with the consideration that a side effect of this method is clearing of the queue)

libjsqrc/ethereumjs/.travis.yml
debris: this is moved into another repo. But we are using **git subtrees** which is pulling whole subrepo into master repository.

libweb3jsonrpc/WebThreeStubServerBase.cpp
gavofyork: or `moveQueuedTransactions`

debris: renamed to `fetchQueuedTransactions`

libdevcore/CommonJS.h
```diff
@@ -38,12 +38,21 @@ template <unsigned S> std::string toJS(FixedHash<S> const& _h)
...
-inline std::string toJS(dev::bytes const& _n)
+inline std::string toJS(bytes const& _n, std::size_t _padding = 0)
 {
-	return "0x" + dev::toHex(_n);
+	bytes n = _n;
+	n.resize(std::max<unsigned>(n.size(), _padding));
+	return "0x" + toHex(n);
+}
+
+template< typename T >std::string toJS( T const& i )
```
gavofyork: what whitespace styling is this!? :)

PR #1212, opened 2015-03-05
Index access for calldata and byte arrays.

chriseth:


libsolidity/CompilerUtils.cpp
```diff
@@ -41,18 +41,22 @@ unsigned CompilerUtils::loadFromMemory(unsigned _offset, Type const& _type,
 	return loadFromMemoryHelper(_type, _fromCalldata, _padToWordBoundaries);
 }
 
-void CompilerUtils::loadFromMemoryDynamic(Type const& _type, bool _fromCalldata, bool _padToWordBoundaries)
+void CompilerUtils::loadFromMemoryDynamic(
+	Type const& _type, bool _fromCalldata, bool _padToWordBoundaries, bool _keepUpdatedMemoryOffset)
```
LefterisJP: In my latest PR had a lot of back and forth with Gavin regarding how breaking a function to multiple lines should be. Here is his comment that shed a light on it. https://github.com/ethereum/cpp-ethereum/pull/1169#issuecomment-76964724

So I suppose if you really want to break it you should add each argument to its own line.

libsolidity/CompilerUtils.h
```diff
@@ -46,7 +46,8 @@ class CompilerUtils
 	/// Dynamic version of @see loadFromMemory, expects the memory offset on the stack.
 	/// Stack pre: memory_offset
 	/// Stack post: value... (memory_offset+length)
-	void loadFromMemoryDynamic(Type const& _type, bool _fromCalldata = false, bool _padToWordBoundaries = true);
+	void loadFromMemoryDynamic(Type const& _type, bool _fromCalldata = false,
```
LefterisJP: Same thing here. It should be:

```
void loadFromMemoryDynamic(
	Type const& _type,
	 bool _fromCalldata = false,
	bool _padToWordBoundaries = true,
	bool _keepUpdatedMemoryOffset = true
);
```

libsolidity/LValue.cpp
```diff
@@ -232,6 +232,64 @@ void StorageItem::setToZero(SourceLocation const&, bool _removeReference) const
...
+	else
+		m_context << eth::Instruction::DUP2 << eth::Instruction::SLOAD
+			<< eth::Instruction::DUP2 << eth::Instruction::BYTE;
+}
+
+void StorageByteArrayElement::storeValue(Type const&, SourceLocation const&, bool _move) const
+{
+	//@todo optimize this
+
+	// stack: value ref bytenr
```
LefterisJP: Maybe add some `static const int` or `#define` here for all those constants? Of course if you plan to change/optimize soon, then it won't be needed.

libsolidity/LValue.cpp
```diff
@@ -232,6 +232,64 @@ void StorageItem::setToZero(SourceLocation const&, bool _removeReference) const
...
+	m_context << u256(31) << eth::Instruction::SUB << u256(0x100) << eth::Instruction::EXP;
+	// stack: value ref (1<<(8*(31-bytenr)))
+	m_context << eth::Instruction::DUP2 << eth::Instruction::SLOAD;
+	// stack: value ref (1<<(8*(31-bytenr))) old_full_value
```
LefterisJP: in all those comments, is `bytenr` number of bytes? Wouldn't it be more understandable as `bytenum` or `byte_num`?

libsolidity/LValue.h
```diff
@@ -114,6 +113,23 @@ class StorageItem: public LValue
...
+ * Stack: <storage_ref> <byte_number>
+ */
+class StorageByteArrayElement: public LValue
+{
+public:
+	/// Constructs the LValue and assumes that the storage reference is already on the stack.
+	StorageByteArrayElement(CompilerContext& _compilerContext);
+	virtual unsigned sizeOnStack() const override { return 2; }
+	virtual void retrieveValue(SourceLocation const& _location, bool _remove = false) const override;
+	virtual void storeValue(Type const& _sourceType,
```
LefterisJP: Same thing I mentioned above about breaking a function into multiple arguments each on its own line.

libsolidity/LValue.h
```diff
@@ -114,6 +113,23 @@ class StorageItem: public LValue
...
+		SourceLocation const& _location = SourceLocation(), bool _move = false) const override;
+	virtual void setToZero(
```
LefterisJP: here too

libsolidity/LValue.cpp
chriseth: I think naming them does not help in understanding this.

PR #1224, opened 2015-03-06
check for minGasLimit and minDifficulty when constructing block

CJentzsch:
this is critical for consensus between the clients. Otherwise cpp could accept a block which is not valid.


gavofyork:
> this is critical for consensus between the clients. Otherwise cpp could accept a block which is not valid.

Wrong.

All blocks go through verifyParent before being accepted which validates difficulty and gasUsed.

PR #1225, opened 2015-03-06
move all solidity related test cases into testsol

guanqun:
The test case for solidity is very self contained, and during development, it would be great to just test the solidity test cases.

So the solidity test cases is split into `testsol` binary, and it's added to `make test` by default.


LefterisJP:
This is not necessary.
You can simply provide an argument with/without a wildmark to `testeth` like so:

```
./build/test/testeth --run_test=Solidity*
./build/test/testeth --run_test=SolidityNameAndTypeResolution   
```

chriseth:
Are there substantially less link dependencies? If yes, I would say this pr adds some value.

guanqun:
@LefterisJP ah, forgot the boost test case parameters. Thank you, this should be enough for me.

PR #1228, opened 2015-03-06
Test macros for printing Exception information

LefterisJP:
`BOOST_XXX()` macros don't show any information about the exceptions that are thrown. With this PR I am adding some additional macros, the `ETH_TEST_XXX()` family of macros that should act exactly as the BOOST macros but instead print information about the thrown exception.

This is really necessary when debugging code we are not familiar with and need to know why and where a test fails. Only use them in SolidityTests for now but feel free to use them in your own tests too.

### DIFF
--- a/test/SolidityExpressionCompiler.cpp
+++ b/test/SolidityExpressionCompiler.cpp
@@ -44,6 +44,23 @@ namespace test
 namespace
 {
 
+// LTODO: Move to some more generic location. We really need it
+/// Make sure that no Exception is thrown during testing. If one is thrown show its info.
+/// @param _expression    The expression for which to make sure no exceptions are thrown
+/// @param _message       A message to act as a prefix to the expression's error information
+#define ETH_TEST_REQUIRE_NO_THROW(_expression, _message)				\
+	do {																\
+		try															\
+		{																\
+			_expression;												\
+		}																\
+		catch (boost::exception const& _e)								\
+		{																\
+			auto msg = std::string(_message) + boost::diagnostic_information(_e); \
+			BOOST_FAIL(msg);											\
+		}																\
+	}while (0)
+
 /// Helper class that extracts the first expression in an AST.
 class FirstExpressionExtractor: private ASTVisitor
 {
@@ -72,8 +89,8 @@ private:
 	Expression* m_expression;
 };
 
-Declaration const& resolveDeclaration(vector<string> const& _namespacedName,
-											  NameAndTypeResolver const& _resolver)
+Declaration const& resolveDeclaration(
+	vector<string> const& _namespacedName, NameAndTypeResolver const& _resolver)
 {
 	Declaration const* declaration = nullptr;
 	// bracers are required, cause msvc couldnt handle this macro in for statement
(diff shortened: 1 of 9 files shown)


CJentzsch:
Actually BOOST_XXX gives you a lot of information. When an exception is thrown with BOOST_THROW_EXCEPTION , you can catch it and print boost::diagnostic_information(_excetpionObject) and get more than you need :-)

LefterisJP:
Yes I know and this is what the macros I introduce here do.

The `BOOST_XXX` test macros don't do that. Relevant SO post:
http://stackoverflow.com/questions/15133259/boost-check-no-throw-how-to-get-exception-message-printed

test/TestHelper.h
```diff
@@ -44,6 +44,41 @@ void connectClients(Client& c1, Client& c2);
...
+		try															\
+		{																\
+			_expression;												\
+		}																\
+		catch (boost::exception const& _e)								\
+		{																\
+			auto msg = std::string(_message) + boost::diagnostic_information(_e); \
+			BOOST_FAIL(msg);											\
+		}																\
+	}while (0)
```
chriseth: space.
Why would a normal block not work?

test/TestHelper.h
```diff
@@ -44,6 +44,41 @@ void connectClients(Client& c1, Client& c2);
 namespace test
 {
 
+/// Make sure that no Exception is thrown during testing. If one is thrown show its info and fail the test.
+/// Our version of BOOST_REQUIRE_NO_THROW()
+/// @param _expression    The expression for which to make sure no exceptions are thrown
+/// @param _message       A message to act as a prefix to the expression's error information
+#define ETH_TEST_REQUIRE_NO_THROW(_expression, _message)				\
+	do {																\
```
CJentzsch: { on next line

test/TestHelper.h
```diff
@@ -44,6 +44,41 @@ void connectClients(Client& c1, Client& c2);
...
+			BOOST_FAIL(msg);											\
+		}																\
+	}while (0)
+
+/// Check if an Exception is thrown during testing. If one is thrown show its info and continue the test
+/// Our version of BOOST_CHECK_NO_THROW()
+/// @param _expression    The expression for which to make sure no exceptions are thrown
+/// @param _message       A message to act as a prefix to the expression's error information
+#define ETH_TEST_CHECK_NO_THROW(_expression, _message)					\
+	do {																\
```
CJentzsch: same

test/TestHelper.h
```diff
@@ -44,6 +44,41 @@ void connectClients(Client& c1, Client& c2);
...
+		try															\
+		{																\
+			_expression;												\
+		}																\
+		catch (boost::exception const& _e)								\
+		{																\
+			auto msg = std::string(_message) + boost::diagnostic_information(_e); \
+			BOOST_MESSAGE(msg);										\
+		}																\
+	}while (0)
```
CJentzsch: space between } and while

test/TestHelper.h
CJentzsch: space between } and while

test/SolidityExpressionCompiler.cpp
```diff
@@ -72,8 +72,8 @@ class FirstExpressionExtractor: private ASTVisitor
 	Expression* m_expression;
 };
 
-Declaration const& resolveDeclaration(vector<string> const& _namespacedName,
-											  NameAndTypeResolver const& _resolver)
```
CJentzsch: If if interpreste this correctly in github, then it looks like a lot of white spaces

test/TestHelper.h
LefterisJP: This is the standard way to create *good* macros that can end in a semicolon and can act as an expression.

test/SolidityExpressionCompiler.cpp
LefterisJP: Yeah but they are red. So that means they are removed

LefterisJP:
rebased

PR #1230, opened 2015-03-06
fixed compiling on osx

debris:


LefterisJP:
looks good but if someone with mac tested it also that would be awesome

caktux:
It does fix compilation but strict tests are failling.

PR #1232, opened 2015-03-06
Output source lines for sub-assemblies.

chriseth:


LianaHus:
your builds are failing but it seems develop is broken.

PR #1237, opened 2015-03-07
Fixed: Some instructions did not have source locations.

chriseth:


libsolidity/CompilerContext.h
```diff
@@ -130,12 +130,15 @@ class CompilerContext
 	class LocationSetter: public ScopeGuard
 	{
 	public:
-		LocationSetter(CompilerContext& _compilerContext, ASTNode const* _node):
-			ScopeGuard(std::bind(&CompilerContext::popVisitedNodes, _compilerContext)) { _compilerContext.pushVisitedNodes(_node); }
+		LocationSetter(CompilerContext& _compilerContext, ASTNode const& _node):
+			ScopeGuard([&]{ _compilerContext.popVisitedNodes(); })
+		{ _compilerContext.pushVisitedNodes(&_node); }
```
LianaHus: shouldn't the implementation be at the same line as declaration?

libsolidity/CompilerContext.cpp
```diff
@@ -182,34 +182,7 @@ void CompilerContext::resetVisitedNodes(ASTNode const* _node)
...
-CompilerContext& CompilerContext::operator<<(u256 const& _value)
-{
-	solAssert(!m_visitedNodes.empty(), "No node on the visited stack");
-	m_asm.append(_value, m_visitedNodes.top()->getLocation());
-	return *this;
-}
-
-CompilerContext& CompilerContext::operator<<(bytes const& _data)
-{
-	solAssert(!m_visitedNodes.empty(), "No node on the visited stack");
```
subtly: This assert was removed in the new header definitions of << and isn't in append(). Is this change is covered elsewhere?

chriseth: I think it is fine to append code without any location and there are enough checks in Assembly.cpp:264

PR #1243, opened 2015-03-09
Handle out of gas for contract code storage

arkpar:
Was just silently ignored with contract not being created.

There is a code that handles out-of-gas in the following way (Executive.cpp:213):
```
		catch (VMException const& _e)
		{
			clog(StateSafeExceptions) << "Safe VM Exception. " << diagnostic_information(_e);
			m_endGas = 0;
			m_excepted = true;
			m_ext->revert();
		}
```
This was not called for the case when running out of gas for contract code storage.
@CJentzsch A test for that might be usefull

### DIFF
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -199,13 +199,13 @@ bool Executive::go(OnOpFunc const& _onOp)
 
 			if (m_isCreation)
 			{
-				bigint storageCost = (bigint)m_out.size() * c_createDataGas;
+				u512 storageCost = (u512)m_out.size() * c_createDataGas;
 				if (storageCost <= m_endGas)
 					m_endGas -= (u256)storageCost;
 				else
 				{
 					clog(StateDetail) << "Not enough gas to pay for the contract code storage: Require >" << storageCost << " Got" << m_endGas;
-					BOOST_THROW_EXCEPTION(OutOfGas() << RequirementError(storageCost, (bigint)m_endGas));
+					BOOST_THROW_EXCEPTION(OutOfGas() << RequirementError((bigint)storageCost, (bigint)m_endGas));
 				}
 				m_s.m_cache[m_newAddress].setCode(m_out.toBytes());
 			}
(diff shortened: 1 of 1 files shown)


libethereum/Executive.cpp
```diff
@@ -199,10 +199,14 @@ bool Executive::go(OnOpFunc const& _onOp)
 
 			if (m_isCreation)
 			{
-				if (m_out.size() * c_createDataGas <= m_endGas)
-					m_endGas -= m_out.size() * c_createDataGas;
+				bigint storageCost = (bigint)m_out.size() * c_createDataGas;
```
subtly: Is bigint unbound? Is storage type bound?

CJentzsch:
This is a real protocol change. I am in favour of throwing an OOG exception as we always do when an unexpected result would be produced (creating account with no code in this case).
Also the casting to bigint is extremly important, otherwise one could create an extreme large contracte by overflowing the storage cost. Because when m_out.size * c_createDataGas > u256.max , one could pass the ...<=m_endGas check, and place the large code in the blockchain.

libethereum/Executive.cpp
CJentzsch: bigint is unbound. m_out.size() is of type size_t , which is long unsigned int. As long as c_createDataGas stays at this order of magnitude (=200 right now) then you can not overflow by doing the multiplication (200 * size_t.max can never be larger than u256.max), but the you can overflow the size() function by returning m_out with a real size larger than size_t.max. I will write a test for that.

arkpar:
[17:31:41] gavofyork: "No code is deposited in the state if the gas does not
cover the additional per-byte contract deposit fee."
[17:31:56] gavofyork: the YP specifically doesn't refer to this as an exceptional condition
[17:32:15] gavofyork: simply a silent failure when applying the code
[17:32:47] gavofyork: i.e. no reversion!
[17:33:12] gavofyork: this PR would surely result in quite a catastrophic consensus failure
[17:33:37] gavofyork: i.e. when a creation tx made a CALL in its init
[17:34:26] gavofyork: the relevant eq in the YP is 77
[17:35:34] gavofyork: note the two different resultant states (bold sigma) dependent on whether the execution was exceptional (OOG) or whether there isn't enough gas for storage

PR #1258, opened 2015-03-10
Milliseconds granularity and UTC in Structured Logging timestamp

LefterisJP:


libdevcore/CommonTime.cpp
```diff
@@ -0,0 +1,15 @@
+#include "CommonTime.h"
```
subtly: License text missing

libdevcore/CommonTime.h
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+#pragma once
+
+#include <time.h>
+
+namespace dev
+{
+
+/// Cross platform wrapper for thread safe gmtime()
```
subtly: This says thread-safe but I don't see a mutex. Are the gmtime_* methods thread-safe?

libdevcore/CommonTime.cpp
LefterisJP: ah damn yeah .. sorry

libdevcore/CommonTime.h
LefterisJP: It's thread safe in the way that we provide the result and the static storage of the implementation of the function is not used. It's the difference between `gmtime()` and `gmtime_r()`.

For more info regarding this you can check [this](http://stackoverflow.com/a/25618891/110395) SO post

libdevcore/CommonTime.h
```diff
@@ -0,0 +1,34 @@
...
+tm *timeToUTC(const time_t *_timeInput, struct tm *_result);
```
subtly: If it's not used elsewhere this can be added to structured logger rather than another translation unit. Windows #ifdef statements are already popular in Common.h/cpp and future structured logger events will require other definitions from Common.

libdevcore/CommonTime.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+namespace dev
+{
+
+tm *timeToUTC(const time_t *_timeInput, struct tm *_result)
+{
+#ifdef __WIN32
```
LefterisJP: needs one underscore only. Will fix.

libdevcore/CommonTime.h
LefterisJP: The idea is to have some common Time related stuff here. Mainly workarounds aroudn what C++11 chrono does not provide. It's much more structured and modular to have a common header for that stuff.

subtly: I meant to write, "added to structured logger or Common.h/cpp".

LefterisJP:
rebased

gavofyork:
Prefer to have a function accepting `chrono::time_point` as input, not the outdated C struct `time_t`.

arkpar:
Msvc build fails: CommonTime.cpp(35): error C2440: 'return' : cannot convert from 'errno_t' to 'tm *'

LefterisJP:
acknowledged, will try to fix.

gavofyork:
Ok so I'd prefer not to be exposing any time_t stuff in headers. I consider it a deprecated API. I'd much prefer to just bite the bullet and import [https://howardhinnant.github.io/duration_io/chrono_io.html](Chrono I/O) into the codebase and just use the templated `dev::toString`.

If we can't make that work, then `operator<<` on templated `chrono::time_point<T>` would do, still using the existing templated `dev::toString`.

LefterisJP:
I don't really agree with the C-era thing. The C standard library and Posix API is used everywhere today and is good and trusted. But yes if you want to go full C++11 you need that guy's library. It's not a bad idea.

So what would you like me to do with this PR?

gavofyork:
Now I put chrono_io in, so toString(chrono::time_point<T>) should work ok; you might want a specialised toString that gives a particular format out. the chrono_io has some docs on how to alter format with stream inserters.

LefterisJP:
As long as it is possible to get the timestamp in UTC time with the UTC format `2015-03-09T12:34:37.401335Z ` then we will be golden!

Notice the subsecond granulatiry of the timestamp.

PR #1263, opened 2015-03-10
Remove StructuredLogger dependency on libp2p

LefterisJP:


subtly:
Boost doesn't need to be used here at all. It would be better for this method to accept a string and an integer for IP and port.

LefterisJP:
I prefer it this way, you pass it an endpoint and it takes care of printing it. The log method is thus typechecked. A string can be anything.

PR #1266, opened 2015-03-10
Test's Hex String Fields

winsvega:
Prefixed with 0x.. on output


CJentzsch:
issues as discussed privatly

winsvega:
so it would be smth like this then

test/transaction.cpp
```diff
@@ -95,6 +95,11 @@ void doTransactionTests(json_spirit::mValue& _v, bool _fillin)
 				cnote << "Transaction Exception: " << diagnostic_information(_e);
 				o.erase(o.find("transaction"));
 			}
+
+			fixHexDataField(tObj, "data");
+			fixHexDataField(tObj, "to");
```
CJentzsch: addresses do net get prefixed with 0x

test/TestHelper.cpp
```diff
@@ -174,6 +186,24 @@ void ImportTest::importTransaction(json_spirit::mObject& _o)
...
+		fixHexDataField(tObj, "currentCoinbase");
+		fixHexDataField(tObj, "previousHash");
+	}
+
+	//fix transaction fields
+	if (m_TestObject.count("transaction") > 0)
+	{
+		json_spirit::mObject& tObj = m_TestObject["transaction"].get_obj();
+		fixHexDataField(tObj, "data");
+		fixHexDataField(tObj, "to");
```
CJentzsch: addresses do not get prefixed with 0x

test/vm.cpp
```diff
@@ -98,10 +98,10 @@ void FakeExtVM::push(mArray& a, u256 _v)
 mObject FakeExtVM::exportEnv()
 {
 	mObject ret;
-	ret["previousHash"] = toString(previousBlock.hash);
+	ret["previousHash"] = "0x"+toString(previousBlock.hash);
```
CJentzsch: addresses and hashes do not need to get prefixed by 0x. Only numbers which are given in hex format. Therefore all changes in this file can be reverted.

PR #1277, opened 2015-03-11
fixed #1273, npm is not used by default

### LINKED ISSUE #1273: npm install in cmake should be optional
This is causing random build failures, and running npm as part of the build process should simply not be the default and done using a special cmake flag only.

debris:


LefterisJP:
Nice. We needed this. Let's merge when all buildbots give the green light.

PR #1279, opened 2015-03-11
Mix - JavaScript debugging Console

yann300:


mix/qml/WebPreview.qml
```diff
@@ -254,6 +303,96 @@ Item {
...
+			Column {
+				id: expressionPanel
+				width: 350
+				Layout.preferredWidth: 350
+				Layout.fillHeight: true
+				spacing: 0
+				visible: false
+				function addExpression()
+				{
+					if (expressionInput.text === "" || expressionInput.text === qsTr("Expression"))
```
arkpar: "Expression" should be a valid expression

mix/qml/WebPreview.qml
```diff
@@ -254,6 +303,96 @@ Item {
...
+						return;
+					webView.runJavaScript("executeJavaScript(\"" + expressionInput.text.replace(/"/g, '\\"') + "\")", function(result) {
```
arkpar: Does work with strings? What if user enters double quoted string as input?

mix/qml/html/WebContainer.html
```diff
@@ -36,6 +36,12 @@
 	web3.setProvider(new web3.providers.HttpSyncProvider(url));
 };
 
+executeJavaScript = function(script) {
+	var preview = document.getElementById('preview');
+	var obj = preview.contentWindow.eval(script);
+	return JSON.stringify(obj);
```
arkpar: Can we use json pretty print format here? i.e.  JSON.stringify(obj, null, 2);

PR #1286, opened 2015-03-11
Initialize jump type.

chriseth:


LianaHus:
I put looksgood but probably we should wait for build results

chriseth:
Failures not caused by this change.

Issue #1254, opened 2015-03-09
Ethash Spam Vulnerabilities


xcthulhu:
There are two potential spam attacks:

(1) An attacker can send blocks from the future, causing caches to be created and saved to disk as they are checked.  An attacker can potentially flood the victims hard-drive with 100GB of caches.

(2) `mixdigests` are not verified.  This is a problem, because a spammer can solve sha3 challenges more rapidly than a client can verify at very low cost, causing the cpp-client to be slowed down computing much more expensive ethash commands.

PR #1292, opened 2015-03-12
Introducing bounds checking, DOS quick verification

### LINKED ISSUE #1255: Ethash Bounds Check
Ethash only has 2048 `cache_size`s and `dag_size`s, so there needs to be a bounds check on these table lookups (there's an assert that will fail if stuff is looked up out-of-bounds)

xcthulhu:
Fixes #1255 and #1254

### DIFF
--- a/libethash/ethash.h
+++ b/libethash/ethash.h
@@ -82,11 +82,11 @@ static inline int ethash_check_difficulty(
 	return 1;
 }
 
-int ethash_quick_check_difficulty(
+void ethash_quick_hash(
+        uint8_t return_hash[32],
         const uint8_t header_hash[32],
         const uint64_t nonce,
-        const uint8_t mix_hash[32],
-        const uint8_t difficulty[32]);
+        const uint8_t mix_hash[32]);
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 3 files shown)


libethcore/Ethasher.cpp
```diff
@@ -43,6 +43,12 @@ Ethasher* dev::eth::Ethasher::s_this = nullptr;
 bytes const& Ethasher::cache(BlockInfo const& _header)
 {
 	RecursiveGuard l(x_this);
+	if (_header.number > EPOCH_LENGTH*2048) {
```
gavofyork: Doesn't comply with coding standards:
- Preprocessor define should contain "namespace" prefix (e.g. `ETHASH_EPOCH_LENGTH`)
- Space on either side of `*`.
- `{` goes on own line.

Whitespace for much of this PR is bad.

See cpp-ethereum/CodingStandards.txt

gavofyork:
This also fails to build on *all* platforms. Did you try building?!

libethash/ethash.h
```diff
@@ -82,11 +82,11 @@ static inline int ethash_check_difficulty(
...
 }
 
-int ethash_quick_check_difficulty(
+void ethash_quick_hash(
+        uint8_t return_hash[32],
         const uint8_t header_hash[32],
         const uint64_t nonce,
-        const uint8_t mix_hash[32],
-        const uint8_t difficulty[32]);
+        const uint8_t mix_hash[32]);
```
chfast: This `32` is misleading because the argument will be passed by pointer. We should think about safer interface.

PR #1294, opened 2015-03-12
testeth: optional VM tracing (--vmtrace)

chfast:
Also: command line options refactoring.


CJentzsch:
Really like what you did here!

PR #1309, opened 2015-03-13
Fix calldataload

CJentzsch:
The problem is that when the input value is larger than unsigned.max, it will be downcasted to a position where there could be some non zero data which will be pushed to the stack, although the correct behavior according to YP would be to push zeroes. The test given below, without the fix, would result in having 0x4200...00 in the storage instead of 0.


chriseth:
Looks good apart from c-style casts.

PR #1315, opened 2015-03-13
Mix - Track files status.

yann300:
- Warn user to reload file if file changed on disk.
 - Ctrl-s: save the current doc only.
 - Ctrl-s: if contract: rerun, if document: reload preview only.
 - Ctrl-s: if compilation error do not rerun.
 - File change: add * and sync with codeeditor.js


mix/FileIo.h
```diff
@@ -39,8 +40,11 @@ class FileIo: public QObject
 signals:
 	/// Signalled in case of IO error
 	void error(QString const& _errorText);
+	/// Signnalled when a file is changed.
+	void fileChanged(QString const& _filePath);
 
 public:
+	FileIo(): m_watcher(new QFileSystemWatcher(this)) { connect(m_watcher, &QFileSystemWatcher::fileChanged, this, &FileIo::fileChanged); }
```
arkpar: Better move this to cpp and use forward declaration for QFileSystemWatcher here

mix/qml/CodeEditorView.qml
```diff
@@ -98,17 +111,64 @@ Item {
...
+
+		onContractSaved: {
+			resetEditStatus(documentId);
+		}
+
+		onDocumentSaving: {
+			for (var i = 0; i < editorListModel.count; i++)
+			{
+				var doc = editorListModel.get(i);
+				if (doc.path === document)
```
arkpar: Please use documentId for document id )

mix/qml/js/ProjectModel.js
```diff
@@ -246,6 +263,7 @@ function removeDocument(documentId) {
 	var document = projectListModel.get(i);
 	if (!document.isContract) {
 		projectListModel.remove(i);
+		fileIo.stopWatching(projectPath+ documentId);
```
arkpar: space

mix/qml/main.qml
```diff
@@ -279,12 +302,20 @@ ApplicationWindow {
 	Action {
 		id: saveAllFilesAction
 		text: qsTr("Save All")
-		shortcut: "Ctrl+S"
+		shortcut: "Ctrl+A"
```
arkpar: Ctrl+A is usually used to select all text. I Suggest using Ctrl+Shift+S instead.

PR #1325, opened 2015-03-16
Adding msg.sig Solidity Magic type

LefterisJP:
msg.sig will return a bytes4 with the function signature located in CALLDATALOAD

### DIFF
--- a/libsolidity/ExpressionCompiler.cpp
+++ b/libsolidity/ExpressionCompiler.cpp
@@ -646,6 +646,8 @@ void ExpressionCompiler::endVisit(MemberAccess const& _memberAccess)
 			m_context << eth::Instruction::GASPRICE;
 		else if (member == "data")
 			m_context << u256(0) << eth::Instruction::CALLDATASIZE;
+		else if (member == "sig")
+			m_context << u256(0) << eth::Instruction::CALLDATALOAD << (u256(0xffffffff) << (256 - 32))<< eth::Instruction::AND;
 		else
 			BOOST_THROW_EXCEPTION(InternalCompilerError() << errinfo_comment("Unknown magic member."));
 		break;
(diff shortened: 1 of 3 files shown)


test/SolidityEndToEndTest.cpp
```diff
@@ -1031,6 +1031,19 @@ BOOST_AUTO_TEST_CASE(blockchain)
 	BOOST_CHECK(callContractFunctionWithValue("someInfo()", 28) == encodeArgs(28, 0, 1));
 }
 
+BOOST_AUTO_TEST_CASE(msg_sig)
+{
+	char const* sourceCode = R"(
+		contract test {
+			function foo(uint256 a) returns (bytes4 value) {
+                return msg.sig;
```
LefterisJP: ops, will fix

libsolidity/ExpressionCompiler.cpp
```diff
@@ -646,6 +646,8 @@ void ExpressionCompiler::endVisit(MemberAccess const& _memberAccess)
 			m_context << eth::Instruction::GASPRICE;
 		else if (member == "data")
 			m_context << u256(0) << eth::Instruction::CALLDATASIZE;
+		else if (member == "sig")
+			m_context << u256(0) << eth::Instruction::CALLDATALOAD << (u256(0xffffffff) << (256 - 32))<< eth::Instruction::AND;
```
chriseth: Missing space

test/SolidityEndToEndTest.cpp
```diff
@@ -1031,6 +1031,19 @@ BOOST_AUTO_TEST_CASE(blockchain)
...
+		contract test {
+			function foo(uint256 a) returns (bytes4 value) {
+				return msg.sig;
+			}
+		}
+	)";
+	compileAndRun(sourceCode);
+	BOOST_CHECK(callContractFunctionWithValue("foo(uint256)", 13) == encodeArgs(asString(FixedHash<4>(dev::sha3("foo(uint256)")).asBytes())));
+}
+
```
chriseth: Perhaps another test that checks that the signature does not change after an internal call?

chriseth: Actually I meant "inside an internal call":
```
function f() returns (bytes4) { return msg.call; }
function g() returns (bytes4) { return f(); }
```

PR #1326, opened 2015-03-16
Let Solidity pack its storage

chriseth:
Everything is packed as tightly as possible and it starts in a new storage slot, if it does not fit.
Structs always start a new storage slot.
Mapping (heads) always start a new storage slot.
Array data always starts a new storage slot.

The reason behind this is that we do not want run-time computation of whether a type has to be moved to the next slot or not because it does not fit, or put otherwise: The code to access a struct item does not depend on the alignment of the struct itself.

Still to be implemented: Pack array elements.


chriseth:
Rebased

arkpar:
WIndows build fails

PR #1327, opened 2015-03-16
Mix: Support for solidity arrays and structs

arkpar:


mix/ClientModel.cpp
```diff
@@ -256,11 +245,11 @@ void ClientModel::executeSequence(std::vector<TransactionSettings> const& _seque
 						BOOST_THROW_EXCEPTION(FunctionNotFoundException() << FunctionName(transaction.functionId.toStdString()));
 					if (!transaction.functionId.isEmpty())
 						encoder.encode(f);
-					for (int p = 0; p < transaction.parameterValues.size(); p++)
+					for (QVariableDeclaration const* p: f->parametersList())
 					{
-						if (f->parametersList().size() <= p || f->parametersList().at(p)->type() != transaction.parameterValues.at(p)->declaration()->type())
-							BOOST_THROW_EXCEPTION(ParameterChangedException() << FunctionName(transaction.functionId.toStdString()));
```
yann300: There is no check anymore to check if parameters has changed (or it this has been moved elsewhere)?

arkpar: If the type of the parameter has changed it will still try to interpret it from string to a new type later

PR #1338, opened 2015-03-17
Solidity: Convert contract types to addresses

LianaHus:
https://www.pivotaltracker.com/n/projects/1189488/stories/88772706

### DIFF
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -431,7 +431,7 @@ void EventDefinition::checkTypeRequirements()
 		if (!var->getType()->canLiveOutsideStorage())
 			BOOST_THROW_EXCEPTION(var->createTypeError("Type is required to live outside storage."));
 		if (!var->getType()->externalType())
-			BOOST_THROW_EXCEPTION(var->createTypeError("Internal type is not allowed for Events"));
+			BOOST_THROW_EXCEPTION(var->createTypeError("Internal type is not allowed as event parameter type."));
 	}
 	if (numIndexed > 3)
 		BOOST_THROW_EXCEPTION(createTypeError("More than 3 indexed arguments for event."));
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -88,7 +88,7 @@ void ContractDefinition::checkTypeRequirements()
 		if (hashes.count(hash))
 			BOOST_THROW_EXCEPTION(createTypeError(
 									  std::string("Function signature hash collision for ") +
-									  it.second->externalTypes()));
+									  it.second->externalSignature()));
 		hashes.insert(hash);
 	}
 }
@@ -192,7 +192,7 @@ vector<pair<FixedHash<4>, FunctionTypePointer>> const& ContractDefinition::getIn
 				if (functionsSeen.count(f->getName()) == 0 && f->isPartOfExternalInterface())
 				{
 					functionsSeen.insert(f->getName());
-					FixedHash<4> hash(dev::sha3(f->externalTypes()));
+					FixedHash<4> hash(dev::sha3(f->externalSignature()));
 					m_interfaceFunctionList->push_back(make_pair(hash, make_shared<FunctionType>(*f, false)));
 				}
 
@@ -202,7 +202,7 @@ vector<pair<FixedHash<4>, FunctionTypePointer>> const& ContractDefinition::getIn
 					FunctionType ftype(*v);
 					solAssert(v->getType().get(), "");
 					functionsSeen.insert(v->getName());
-					FixedHash<4> hash(dev::sha3(ftype.externalTypes(v->getName())));
+					FixedHash<4> hash(dev::sha3(ftype.externalSignature(v->getName())));
 					m_interfaceFunctionList->push_back(make_pair(hash, make_shared<FunctionType>(*v)));
 				}
 		}
@@ -320,9 +320,9 @@ void FunctionDefinition::checkTypeRequirements()
 	m_body->checkTypeRequirements();
 }
 
-string FunctionDefinition::externalTypes() const
+string FunctionDefinition::externalSignature() const
 {
-	return FunctionType(*this).externalTypes(getName());
+	return FunctionType(*this).externalSignature(getName());
 }
 
 bool VariableDeclaration::isLValue() const
@@ -430,7 +430,7 @@ void EventDefinition::checkTypeRequirements()
 			numIndexed++;
 		if (!var->getType()->canLiveOutsideStorage())
 			BOOST_THROW_EXCEPTION(var->createTypeError("Type is required to live outside storage."));
-		if (!var->getType()->externalType() && getVisibility() >= Visibility::Public)
+		if (!var->getType()->externalType())
 			BOOST_THROW_EXCEPTION(var->createTypeError("Internal type is not allowed for Events"));
 	}
 	if (numIndexed > 3)
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -346,8 +346,11 @@ void VariableDeclaration::checkTypeRequirements()
 	if (!m_value)
 		return;
 	if (m_type)
+	{
 		m_value->expectType(*m_type);
-	else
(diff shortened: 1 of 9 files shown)


libsolidity/AST.cpp
```diff
@@ -653,6 +657,10 @@ void MemberAccess::checkTypeRequirements()
 	if (!m_type)
 		BOOST_THROW_EXCEPTION(createTypeError("Member \"" + *m_memberName + "\" not found or not "
 											  "visible in " + type.toString()));
+	//todo check for visibility
```
chriseth: This should be removed.

libsolidity/Types.cpp
```diff
@@ -738,6 +738,25 @@ string ArrayType::toString() const
 	return ret + "]";
 }
 
+TypePointer ArrayType::externalType() const
+{
+	if (m_location != Location::CallData)
+		return TypePointer();
+	if (m_isByteArray)
+		return shared_from_this();
+	if (!(m_baseType->externalType()))
```
chriseth: No parantheses needed and no braces needed.

libsolidity/Types.cpp
```diff
@@ -738,6 +738,25 @@ string ArrayType::toString() const
...
+	{
+		return TypePointer();
+	}
+	if (dynamic_cast<ArrayType const*>(m_baseType.get()) && m_baseType->isDynamicallySized())
```
chriseth: You can also call `getCategory` instead of the dynamic cast.

libsolidity/Types.cpp
```diff
@@ -738,6 +738,25 @@ string ArrayType::toString() const
...
+		return TypePointer();
+
+	if (m_baseType->isDynamicallySized())
```
chriseth: Should be `if (isDynamicallySized())`

libsolidity/Types.h
```diff
@@ -187,6 +187,9 @@ class Type: private boost::noncopyable, public std::enable_shared_from_this<Type
 																		 "for type without literals."));
 	}
 
+	/// Returns address of type for ABI interface
```
chriseth:
```
/// @returns a type suitable for outside of Solidity, i.e. for contract types it returns address.
/// If there is no such type, returns an empty shared pointer.
```

LianaHus:
rebased

libsolidity/AST.h
```diff
@@ -424,7 +424,7 @@ class FunctionDefinition: public Declaration, public VariableScope, public Docum
 	/// @returns the canonical signature of the function
```
chriseth: Please update the docstring and I think we should keep the word "signature" as part of this function name.

libsolidity/AST.cpp
```diff
@@ -422,6 +430,8 @@ void EventDefinition::checkTypeRequirements()
 			numIndexed++;
 		if (!var->getType()->canLiveOutsideStorage())
 			BOOST_THROW_EXCEPTION(var->createTypeError("Type is required to live outside storage."));
+		if (!var->getType()->externalType() && getVisibility() >= Visibility::Public)
```
chriseth: Visibility does not matter for events.

LianaHus:
rebased

libsolidity/AST.cpp
```diff
@@ -342,8 +347,11 @@ void VariableDeclaration::checkTypeRequirements()
 	if (!m_value)
 		return;
 	if (m_type)
+	{
 		m_value->expectType(*m_type);
-	else
+		if (m_isStateVariable && !m_type->externalType() && getVisibility() >= Visibility::Public)
```
chriseth: It is actually more complicated here: You have to create the function type of the accessor (by using the FunctionType constructor which takes a variable declaration as argument) and check that its type is compatible with external visibility.

libsolidity/AST.cpp
```diff
@@ -422,6 +430,8 @@ void EventDefinition::checkTypeRequirements()
 			numIndexed++;
 		if (!var->getType()->canLiveOutsideStorage())
 			BOOST_THROW_EXCEPTION(var->createTypeError("Type is required to live outside storage."));
+		if (!var->getType()->externalType())
+			BOOST_THROW_EXCEPTION(var->createTypeError("Internal type is not allowed for Events"));
```
chriseth: Perhaps "as event parameter type"?

libsolidity/AST.h
```diff
@@ -421,10 +421,10 @@ class FunctionDefinition: public Declaration, public VariableScope, public Docum
 	/// Checks that all parameters have allowed types and calls checkTypeRequirements on the body.
 	void checkTypeRequirements();
 
-	/// @returns the canonical signature of the function
-	/// That consists of the name of the function followed by the types of the
+	/// @returns the external signature of the function
+	/// That consists of the name of the function followed by the types(external types if isExternalCall) of the
```
chriseth: There is a space missing.

libsolidity/AST.h
```diff
@@ -421,10 +421,10 @@ class FunctionDefinition: public Declaration, public VariableScope, public Docum
...
 	/// arguments separated by commas all enclosed in parentheses without any spaces.
-	std::string getCanonicalSignature() const;
+	std::string externalSignature(bool isExternalCall = false) const;
```
chriseth: I think the signature as a string should always be used for "external calls".

libsolidity/Types.h
```diff
@@ -539,10 +550,11 @@ class FunctionType: public Type
...
 
 	Location const& getLocation() const { return m_location; }
-	/// @returns the canonical signature of this function type given the function name
+	/// @returns the external signature of this function type given the function name
 	/// If @a _name is not provided (empty string) then the @c m_declaration member of the
 	/// function type is used
-	std::string getCanonicalSignature(std::string const& _name = "") const;
+	/// @a isExternalCall shows if it is external function call
+	std::string externalSignature(bool isExternalCall = false, std::string const& _name = "") const;
 	Declaration const& getDeclaration() const
```
chriseth: I think functionType should also implement externalType, but in a kind of different way:
It creates a new function type where parameters and return parameters are replaced by their external types.
This function, in turn, can be used inside `externalSignature`.

test/SolidityNameAndTypeResolution.cpp
```diff
@@ -376,8 +389,63 @@ BOOST_AUTO_TEST_CASE(function_canonical_signature_type_aliases)
...
+			uint a;
+		}
+		contract Test {
+			address a;
+			function g (C c) {}
+			function internalCall() {
+				g(a);
+			}
+	})";
+	BOOST_CHECK_THROW(parseTextAndResolveNames(text), TypeError);
```
chriseth: Can you also add a test for the "positive" case here, i.e. calling `g` with an argument of type `C`?

libsolidity/Types.h
LianaHus: nice idea, will change it

libsolidity/AST.cpp
```diff
@@ -305,8 +306,12 @@ TypePointer FunctionDefinition::getType(ContractDefinition const*) const
 void FunctionDefinition::checkTypeRequirements()
 {
 	for (ASTPointer<VariableDeclaration> const& var: getParameters() + getReturnParameters())
+	{
 		if (!var->getType()->canLiveOutsideStorage())
 			BOOST_THROW_EXCEPTION(var->createTypeError("Type is required to live outside storage."));
+		if (!(var->getType()->externalType()) && getVisibility() >= Visibility::Public)
```
chriseth: You might reverse the order here so that we do not need to create the external type if it is not necessary.

libsolidity/Types.cpp
```diff
@@ -1081,6 +1098,19 @@ unsigned FunctionType::getSizeOnStack() const
 	return size;
 }
 
+TypePointer FunctionType::externalType() const
+{
+	TypePointers paramTypes;
+	TypePointers retParamTypes;
+
+	for (auto it = m_parameterTypes.cbegin(); it != m_parameterTypes.cend(); ++it)
```
chriseth: You could use a range-based for loop here and perhaps assert that the external types exist.

PR #1340, opened 2015-03-17
Structured Logger output is now one-liner

LefterisJP:


libdevcore/StructuredLogger.cpp
```diff
@@ -38,9 +38,10 @@ void StructuredLogger::outputJson(Json::Value const& _value, std::string const&
 {
 	Json::Value event;
 	static Mutex s_lock;
+	Json::FastWriter fastWriter;
 	Guard l(s_lock);
 	event[_name] = _value;
-	cout << event << endl << flush;
+	cout << fastWriter.write(event) << endl << flush;
```
chfast: `endl` does `flush`.

LefterisJP: yes you are absolutely right. I did not add the flush here, was from before but I will remove it.

PR #1344, opened 2015-03-18
Explicit conversion between bytes and int of any size

LefterisJP:
- Allowing aforementioned conversion and adding tests for it


test/SolidityEndToEndTest.cpp
```diff
@@ -1127,45 +1127,43 @@ BOOST_AUTO_TEST_CASE(type_conversions_cleanup)
...
 }
 
-BOOST_AUTO_TEST_CASE(convert_uint_to_fixed_bytes_same_size)
+BOOST_AUTO_TEST_CASE(convert_fixed_bytes_to_fixed_bytes_greater_size)
 {
 	char const* sourceCode = R"(
 		contract Test {
-			function uintToBytes(uint256 h) returns (bytes32 s) {
-				return bytes32(h);
+			function bytesToUint(bytes2 input) returns (bytes4 ret) {
```
chriseth: Function should be called differently.

PR #1347, opened 2015-03-18
p2p: Bug fixes and critical discovery improvements

subtly:
track pings where pubk is unknown so pongs are properly handled
correct mutex usage
notify Host of new node only when it is inserted into table
temporarily double discovery timeout until concatenated packets are supported
remove duplicate eviction-timeout entry
update node's udp endpoint when node is noted as active (fixes bug which duplicates NodeEntry)
don't note pending nodes as active, don't add pending nodes to table
change NodeEntryRemoved to NodeEntryDroppped (Remove will be permanent)
note active node after packet is processed instead of before
do not respond to unsolicited packets
store node pubk and tcp endpoint when Ping is received
fix bug in Host causing empty Peer shared-ptr to be created

### DIFF
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -176,11 +176,14 @@ unsigned Host::protocolVersion() const
 
 void Host::startPeerSession(Public const& _id, RLP const& _rlp, RLPXFrameIO* _io, bi::tcp::endpoint _endpoint)
 {
-	shared_ptr<Peer> p(new Peer());
-	if (m_peers.count(_id))
-		p = m_peers[_id];
-	else
+	shared_ptr<Peer> p;
+	if (!m_peers.count(_id))
+	{
+		p.reset(new Peer());
 		p->id = _id;
+	}
+	else
+		p = m_peers[_id];
 	p->m_lastDisconnect = NoDisconnect;
 	if (p->isOffline())
 		p->m_lastConnected = std::chrono::system_clock::now();
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -176,14 +176,11 @@ unsigned Host::protocolVersion() const
 
 void Host::startPeerSession(Public const& _id, RLP const& _rlp, RLPXFrameIO* _io, bi::tcp::endpoint _endpoint)
 {
-	/// Get or create Peer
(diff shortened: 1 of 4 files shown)


libp2p/Host.cpp
```diff
@@ -176,14 +176,11 @@ unsigned Host::protocolVersion() const
 
 void Host::startPeerSession(Public const& _id, RLP const& _rlp, RLPXFrameIO* _io, bi::tcp::endpoint _endpoint)
 {
-	/// Get or create Peer
-	shared_ptr<Peer> p;
-	p = m_peers[_id];
-	if (!p)
-	{
-		p.reset(new Peer()); // this maybe redundant
+	shared_ptr<Peer> p(new Peer());
```
chriseth: `p = make_shared<Peer>();`

chriseth:
PR description sounds like this could easily be 5 PRs.

libp2p/NodeTable.cpp
```diff
@@ -307,23 +294,26 @@ void NodeTable::noteActiveNode(Public const& _pubk, bi::udp::endpoint const& _en
...
+		node->endpoint.udp.address(_endpoint.address());
+		node->endpoint.udp.port(_endpoint.port());
+		
 		shared_ptr<NodeEntry> contested;
 		{
 			Guard l(x_state);
 			NodeBucket& s = bucket_UNSAFE(node.get());
-			s.nodes.remove_if([&node](weak_ptr<NodeEntry> n)
+			bool removed = false;
+			s.nodes.remove_if([&node, &removed](weak_ptr<NodeEntry> n)
```
chriseth: const ref to n?

libp2p/NodeTable.cpp
```diff
@@ -411,37 +402,50 @@ void NodeTable::onReceived(UDPSocketFace*, bi::udp::endpoint const& _from, bytes
...
+								n->pending = false;
 						
 						it = m_evictions.erase(it);
 					}
+				
+				// if not, check if it's known/pending or a pubk discovery ping
+				if (!evictionEntry)
+				{
+					if (auto n = nodeEntry(nodeid))
+						if (n->pending)
```
gavofyork: `n->pending = false;` would suffice regardless

PR #1348, opened 2015-03-18
Mix: UI improvements

arkpar:
Various small issues fixed

### DIFF
--- a/mix/qml/js/ProjectModel.js
+++ b/mix/qml/js/ProjectModel.js
@@ -61,6 +61,7 @@ function closeProject(callBack) {
 
 function saveProject() {
 	if (!isEmpty) {
+		projectSaving();
 		var projectData = saveProjectFile();
 		if (projectData !== null)
 		{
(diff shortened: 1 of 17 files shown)


mix/qml/js/ProjectModel.js
```diff
@@ -86,6 +85,7 @@ function saveProjectFile()
 		for (var i = 0; i < projectListModel.count; i++)
 			projectData.files.push(projectListModel.get(i).fileName);
 
+		projectSaving(projectData);
```
yann300: SaveProjectFile is used to save the .mix file only, so I think projectSaving should stay in SaveProject.
But the name SaveProjectFile is confusing, maybe better use SaveProjectDesc

arkpar: This signal is used to notify all subscribers that the project is about to get saved, so that they could add custom information to saved data. Saving states is currently broken because of this

PR #1330, opened 2015-03-16
refactored libethereum/Client, parallel tests

debris:
**note**
This PR should not contain any breaking changes, yet it alters one of the most important classes. Please review carefully.

**changes**
- new class in libethereum `InterfaceStub`. [InterfaceStub](https://github.com/debris/cpp-ethereum/blob/jsonrpc_tests/libethereum/InterfaceStub.h) is base implementation of the `Interface`
- `dev::eth::Client` and `dev::mix::MixClient` are using `InterfaceStub`
- `FixedStateInterface` (used in jsonrpc tests) is using `InterfaceStub`
- moved few utility methods from `test/blockchain.cpp` to `TestHelper.cpp`
- parallel tests for `InterfaceStub`
- parallel build of new suites using [cmake's ctest](http://www.cmake.org/Wiki/CMake/Testing_With_CTest)
- `test/TestUtils` file containing few utility functions && fixtures

**issues**
- boost::test [is not thread-safe](https://codecrafter.wordpress.com/2012/11/01/c-unit-test-framework-adapter-part-3/) :-1: 

**jsonrpccpp**
- Since this pull request already contains a lot of changes, I will make separate pull request with jsonrpccpp tests

PR #1351, opened 2015-03-18
refactored libethereum/Client, parallel tests for interface, jsonrpc test executable

debris:
based on #1330 

**note**
This PR should not contain any breaking changes, yet it alters one of the most important classes. Please review carefully.

**changes**
- new class in libethereum `InterfaceStub`. [InterfaceStub](https://github.com/debris/cpp-ethereum/blob/jsonrpc_tests/libethereum/InterfaceStub.h) is base implementation of the `Interface`
- `dev::eth::Client` and `dev::mix::MixClient` are using `InterfaceStub`
- `FixedStateInterface` (used in jsonrpc tests) is using `InterfaceStub`
- moved few utility methods from `test/blockchain.cpp` to `TestHelper.cpp`
- parallel tests for `InterfaceStub`
- parallel build of new suites using [cmake's ctest](http://www.cmake.org/Wiki/CMake/Testing_With_CTest)
- `test/TestUtils` file containing few utility functions && fixtures
- new `libtestutils` library with few helper test classes && methods
- new `ethrpctest` executable which can load blockchain from json file && expose jsonrpc api. (required for testing [rpc against go implementation](https://github.com/ethereum/rpc-tests))
- fixed most of the issues which appeared during tests with `rpc-tests`


debris:
I need to check this pull request once more, to make sure that I merged everything correctly. I will reopen it tomorrow morning.

CJentzsch:
This is really a very big PR, you should split it up in different ones with different topics. The changes to dev::eth::Client are not related to the changes on the test folder, ...

cmake/EthUtils.cmake
```diff
@@ -22,3 +22,43 @@ macro(replace_if_different SOURCE DST)
 	endif()
 endmacro()
 
+macro(eth_add_test NAME) 
+
+	# parse arguments here
+	set(commands)
+	set(current_command "")
+	foreach(arg ${ARGN})
```
CJentzsch: space between foreach and (

ethrpctest/CommandLineInterface.cpp
```diff
@@ -0,0 +1,131 @@
...
+	auto server = new jsonrpc::HttpServer(8080, "", "", 2);
+	jsonrpcServer.reset(new FixedWebThreeServer(*server, {}, &client));
+	jsonrpcServer->StartListening();
+
+	signal(SIGABRT, &sighandler);
+	signal(SIGTERM, &sighandler);
+	signal(SIGINT, &sighandler);
+
+	while (!g_exit)
+	{
```
CJentzsch: you dont need braces here

libtestutils/BlockChainLoader.cpp
```diff
@@ -0,0 +1,101 @@
...
+{
+namespace test
+{
+dev::eth::BlockInfo toBlockInfo(Json::Value const& _json);
+bytes toGenesisBlock(Json::Value const& _json);
+
+}
+}
+
+dev::eth::BlockInfo dev::test::toBlockInfo(Json::Value const& _json)
```
CJentzsch: that is a duplication of  eth::BlockInfo constructBlock(json_spirit::mObject& _o) in TestHelper.cpp. The only difference is that I use json_spirit and you use Json lib. May be you should consider to switch to json_spirit so we only use one lib for json files in the client, and can reuse our helper functions.

libtestutils/BlockChainLoader.cpp
```diff
@@ -0,0 +1,101 @@
...
+	rlpStream << toBytes(_json["nonce"].asString());
+	
+	BlockInfo result;
+	RLP rlp(rlpStream.out());
+	result.populateFromHeader(rlp, IgnoreNonce);
+	return result;
+}
+
+bytes dev::test::toGenesisBlock(Json::Value const& _json)
+{
```
CJentzsch: you can use RLPStream createFullBlockFromHeader(const BlockInfo& _bi,const bytes& _txs, const bytes& _uncles ) in blockchain.cpp instead (move this function to TestHelper.cpp/.h)

libtestutils/BlockChainLoader.cpp
```diff
@@ -0,0 +1,101 @@
...
+	fullStream.appendRaw(RLPEmptyList);
+	bi.verifyInternals(&fullStream.out());
+	
+	return fullStream.out();
+}
+
+BlockChainLoader::BlockChainLoader(Json::Value const& _json)
+{
+	// load pre state
+	StateLoader sl(_json["pre"]);
```
CJentzsch: Instead of StateLoader, you can use the Class ImportTest in TestHelper.h. Again, only difference is the json lib.

ethrpctest/CommandLineInterface.cpp
debris: sure, will remove them

libtestutils/Common.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include <libdevcore/CommonData.h>
+#include <libdevcore/CommonIO.h>
+#include <libdevcrypto/FileSystem.h>
+#include "Common.h"
+
+using namespace std;
+using namespace dev;
+using namespace dev::test;
+
+bytes dev::test::toBytes(std::string const& _str)
```
CJentzsch: fromHex does exactly the same. It already checks for 0x prefix. See  https://github.com/ethereum/cpp-ethereum/blob/develop/libdevcore/CommonData.cpp#L83

libtestutils/Common.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+	return fromHex(_str.substr(0, 2) == "0x" ? _str.substr(2) : _str, ThrowType::Throw);
+}
+
+std::string dev::test::getTestPath()
```
CJentzsch: that is a duplication of https://github.com/ethereum/cpp-ethereum/blob/develop/test/TestHelper.cpp#L377.
Why dont you use this function directly?

libtestutils/ShortLivingDirectory.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "ShortLivingDirectory.h"
+#include "Common.h"
+
+using namespace std;
+using namespace dev;
+using namespace dev::test;
+
+ShortLivingDirectory::ShortLivingDirectory()
+{
+	m_path = getRandomPath();
```
CJentzsch: How random is this?
You should use something like /tmp, or create a folder in the default folder (/~.ethereum).

libtestutils/StateLoader.cpp
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+#include "StateLoader.h"
+
+using namespace std;
+using namespace dev;
+using namespace dev::eth;
+using namespace dev::test;
+
+StateLoader::StateLoader(Json::Value const& _json)
```
CJentzsch: This is a lot of code duplication from TestImporter, just use the existing class instead.

libtestutils/StateLoader.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <json/json.h>
+#include <libethereum/State.h>
+
+namespace dev
+{
+namespace test
+{
+
+class StateLoader
```
CJentzsch: As I said, this class is not needed. Use TestImporter instead.

mix/MixClient.cpp
```diff
@@ -44,27 +44,18 @@ namespace mix
 
 const Secret c_defaultUserAccountSecret = Secret("cb73d9408c4720e230387d956eb0f829d8a4dd2c1055f96257167e14e7169074");
 const u256 c_mixGenesisDifficulty = c_minimumDifficulty; //TODO: make it lower for Mix somehow
-
-class MixBlockChain: public dev::eth::BlockChain
+	
+bytes MixBlockChain::createGenesisBlock(h256 _stateRoot)
```
CJentzsch: Code duplication, use https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/CanonBlockChain.cpp#L75 instead

mix/MixClient.cpp
```diff
@@ -44,27 +44,18 @@ namespace mix
...
 {
-public:
-	MixBlockChain(std::string const& _path, h256 _stateRoot):
-		BlockChain(createGenesisBlock(_stateRoot), _path, true)
-	{
-	}
-
-	static bytes createGenesisBlock(h256 _stateRoot)
```
CJentzsch: Code duplication, use https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/CanonBlockChain.cpp#L75 instead

mix/MixClient.cpp
CJentzsch: Code duplication, use https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/CanonBlockChain.cpp#L75 instead

libtestutils/BlockChainLoader.cpp
debris: > that is a duplication of eth::BlockInfo constructBlock(json_spirit::mObject& _o) in TestHelper.cpp. 

yes, I know, but `constructBlock` is a part of `testeth` executable, and I needed this method in the library.

> The only difference is that I use json_spirit and you use Json lib. May be you should consider to switch to json_spirit so we only use one lib for json files in the client, and can reuse our helper functions.

Changing the code to use json_spirit instead Json lib is quite easy. I can do this, if this is needed. Although Json lib is also used in many other ethereum libraries (eg. web3jsonrpc or solidity)

cmake/EthUtils.cmake
debris: I don't not sure if CodingStandards to cmake files. But I agree, that it looks better with a space ;)

libtestutils/BlockChainLoader.cpp
debris: same as in previous pr

libtestutils/BlockChainLoader.cpp
debris: same as before.

libtestutils/Common.cpp
debris: cool, I will fix this!

libtestutils/ShortLivingDirectory.cpp
debris: I'm using `random_device`, so it's ok. The returned path is using `getDataDir` internally and looks like this : `/~.ethereum/EthereumTests/random_suffix`

libtestutils/StateLoader.cpp
debris: yes, same as before, I didn't want to move `TestImporter` to this library. Also TestImporter do have many other functions and I didn't want to bring all of them here.

debris: Also I did't like it's interface.

mix/MixClient.cpp
debris: Actually not, it was not a part of this pull request. It is actually result of my incorrect merge with develop branch. I haven't changed this. Thanks for careful review. I will fix it.

libtestutils/BlockChainLoader.cpp
debris: Also once, .json test files will have genesis block rlp (as we discussed yesterday :) ), I will not need this function any more.

mix/MixClient.cpp
debris: update
The merge was correct. 
Besides that, MixBlockChain was not a part of this pull request. The only change that I made, was moving Class definition to .h file. Also, it is not code duplication. MixBlockChain `createGenesisBlock` is completely different.

debris:
I talked with @CJentzsch and we agreed on few things related with this pr.

Right now, we could leave libtestutils as it is, but in future we can unify it with TestHelper. At first just by moving the functions from TestHelper to libtestutils. Json::Value && json_spirit can coexist, cause we can easily transform one to another (if this will be required) by doing fromString/toString operations. Later we can decide to which library we want to dedicate ourselves. I would opt for Json lib, mostly because it is used also in other ethereum libraries.


Also, windows build should be working now. I will reopen the pr now.

libethereum/Client.cpp
```diff
@@ -40,7 +40,7 @@ VersionChecker::VersionChecker(string const& _dbPath):
 {
 	auto protocolContents = contents(m_path + "/protocol");
 	auto databaseContents = contents(m_path + "/database");
-	m_ok = RLP(protocolContents).toInt<unsigned>(RLP::LaisezFaire) == eth::c_protocolVersion && RLP(databaseContents).toInt<unsigned>(RLP::LaisezFaire) == c_databaseVersion;
```
gavofyork: for good practice, should be explicitly `eth::c_protocolVersion` since we now have `p2p::c_protocolVersion`.

libethereum/Client.cpp
```diff
@@ -55,7 +55,7 @@ void VersionChecker::setOk()
 		{
 			cwarn << "Unhandled exception! Failed to create directory: " << m_path << "\n" << boost::current_exception_diagnostic_information();
 		}
-		writeFile(m_path + "/protocol", rlp(eth::c_protocolVersion));
```
gavofyork: for good practice, should be explicitly `eth::c_protocolVersion` since we now have `p2p::c_protocolVersion`.

debris:
this pr has been splited into:
#1405 #1406 #1407 #1408 #1409 #1410

PR #1360, opened 2015-03-20
work in progress

subtly:
This is an enhancement and fix which prevents a peer from keeping stalled handshake attempts and bumps the protocol version so that poc8 nodes are dropped from poc9 node tables. This is a problem posed by poc8 nodes which connect but never send enough data for the new handshake read buffer to be filled. Both poc8 and poc9 nodes keep waiting for more bytes and indefinitely stall the connection.


subtly:
See issue #1365

PR #1385, opened 2015-03-23
String is now a reserved keyword

### LINKED ISSUE #1384: Solidity "string" keyword should not be allowed to proceed to typechecking
The `string` keyword is reserved for future usage in solidity, but does nothing at the moment. We still allow it as an elementary type token which is not good since this can cause an Internal compiler error as @yann300 noticed.

Will fix now.

LefterisJP:
- The string keyword is reserved for future use but should not be a
  token in the code since it can trigger internal compiler
  assertions.

- fixes #1384

### DIFF
--- a/libsolidity/Token.h
+++ b/libsolidity/Token.h
@@ -310,6 +310,9 @@ namespace solidity
 	K(Throw, "throw", 0)                                               \
 	K(Try, "try", 0)                                                   \
 	K(Catch, "catch", 0)                                               \
+	K(Using, "using", 0)                                               \
+	K(Type, "type", 0)                                                 \
+	K(TypeOf, "typeof", 0)                                             \
 	/* Illegal token - not able to scan. */                            \
 	T(Illegal, "ILLEGAL", 0)                                           \
 	\
--- a/libsolidity/Token.h
+++ b/libsolidity/Token.h
@@ -143,7 +143,6 @@ namespace solidity
 	\
 	/* Keywords */                                                     \
 	K(Break, "break", 0)                                               \
-	K(Case, "case", 0)                                                 \
 	K(Const, "constant", 0)                                            \
 	K(Anonymous, "anonymous", 0)                                       \
 	K(Continue, "continue", 0)                                         \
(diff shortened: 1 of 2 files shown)


chriseth:
Does this change make it possible to use "string" as an identifier? If yes, it is not really a reserved keyword anymore.

LefterisJP:
Hmm ... you got a point. Should add a warning for that.

libsolidity/Token.h
```diff
@@ -306,6 +303,12 @@ namespace solidity
 	/* Identifiers (not keywords or future reserved words). */         \
 	T(Identifier, NULL, 0)                                             \
 	\
+	/* Keywords reserved for future. use*/                             \
```
chriseth: Please also add `try`.

LefterisJP:
done

libsolidity/Token.h
```diff
@@ -306,6 +303,13 @@ namespace solidity
...
+	T(String, "string", 0)                                             \
+	K(Case, "case", 0)                                                 \
+	K(Switch, "switch", 0)                                             \
+	K(Throw, "throw", 0)                                               \
+	K(Try, "try", 0)                                                   \
+	K(Catch, "catch", 0)                                               \
```
gavofyork: and maybe either `using` or `type` (not sure what is more javascripty), for when we get typedefs.

chriseth:
Probably good to add both. Perhaps also `typeof` (that also exists in javascript)?

LefterisJP:
Done. Sorry for the late response. Did not get any notifications from github for some reason.

LefterisJP:
Hmm no idea why Linux C++ pull request appears to fail. Error message sounds like an one time buildbot thing:

```
[ 81%] Building CXX object test/CMakeFiles/testeth.dir/SolidityCompiler.cpp.o

command timed out: 1200 seconds without output running make -j $(cat /proc/cpuinfo | grep processor | wc -l), attempting to kill
process killed by signal 9
program finished with exit code -1
```

PR #1387, opened 2015-03-23
Transaction Data Must Be Array

winsvega:
+libsecp256k1test


libethereum/Transaction.cpp
```diff
@@ -68,6 +68,10 @@ Transaction::Transaction(bytesConstRef _rlpData, CheckSignature _checkSig)
 		m_type = rlp[field = 3].isEmpty() ? ContractCreation : MessageCall;
 		m_receiveAddress = rlp[field = 3].isEmpty() ? Address() : rlp[field = 3].toHash<Address>(RLP::VeryStrict);
 		m_value = rlp[field = 4].toInt<u256>();
+
+		if (rlp[field = 5].isList())
```
gavofyork: `if (!rlp[field = 5].isData())` is more readable since it directly corresponds to what we're trying to do - namely interpret it as data.

PR #1391, opened 2015-03-24
Mix - solidity autocompletion ui design

yann300:


mix/qml/html/cm/show-hint.css
```diff
@@ -31,33 +31,53 @@
...
+  color: white !important;
+}
+
+.CodeMirror-hint-active .solToken,
+.CodeMirror-hint-active .solTokenType
+{
+  color: white !important;
 }
 
 .solcurrency {
```
arkpar: Can these values be moved to color scheme file?

yann300: Done

Issue #575, opened 2014-12-08
Split projects, re-licence libraries as more permissive


vonj:
https://github.com/ethereum/cpp-ethereum/issues/3

"license is and always has been GPL"

Yet, you Gavin said:

"Kord's point is taken and the licence will be changed to something more liberal in due course", back in January. When is due course?

But now the issue is closed. If you actively want to hinder the adoption of this code, keep the GPL license. Many companies will not take the risk of building their solutions on GPL code, or likely can't even find a business case for it all. If you want to protect the direction of the code, then at least use LGPL. (Though MIT is clearly the simplest license to understand.)

For us GPL is a no go - we would have to use the Go or Python or Java implementation, even though C++ would fit much better.

I understand your point that:
"It's worth remembering that this codebase is a proof-of-concept prototype, created mainly to test the whitepaper's premise and verify as reasonable the architecture it asserts."

However, realistically, probably no one is going to implement a clean room implementation from scratch in C++. If strict clean room principles are employed, the team doing it can't even look at your C++ code, unless they want to use GPL too.

I woke up in the wee hours - couldn't go back to sleep until writing this rant, because it has bugged me since January and now I decided to look back to see if the issue had been resolved.

vonj:
I saw a github project which had an interesting Contributor License Agreement:

https://github.com/flarum/core

Something like this would be very appropriate here as well.

vonj:
Any news on this?

vonj:
Just for completeness, can we have a comment on why it was closed?

bobsummerwill:
Hey @vonj,
I believe this was just a duplicate issue.

The project flipped from GPL to MIT several months back.

chriseth:
@vonj this issue was scheduled for poc-9 and the last substantial comment was over a year ago. Licensing is explained here https://github.com/ethereum/wiki/wiki/Licensing and we have other issues to track the changes that still need to be done.

vonj:
Thank you so much!

bobsummerwill:
Hey @vonj,

Further to this ...

- https://bobsummerwill.com/2016/07/12/c-re-licensing-plan/
- https://bobsummerwill.com/2016/07/12/ethereum-everywhere/

It seems that the earlier attempted switch to MIT was never completed.   I'm driving on an Apache 2.0 relicensing now, so that this long-running saga can finally be put to bed!

PR #1392, opened 2015-03-24
Suggestion for contributor license

vonj:
As a first step to be able to close this issue:
A first step to close issue 575 can be to require for contributors to license their changes under a permissive license.
Gavin can then choose to sublicense to GPL for the main client, and to LGPL or MIT license for the library.

https://github.com/ethereum/cpp-ethereum/issues/575


gavofyork:
Should be Eth Dev (UK) CIC named rather than me, and should be more in line with ethereum.org's Licences section.

i.e. binaries/tools are GPL, rest is MIT

subtly:
Hi @vonj,

As a contributor to this project and several other open source projects I'm interested in this conversation and that we Do The Right Thing :tm:.

GLP covers the subject matter of transfer of rights and in this regard, the changes proposed are redundant; as such, this is very risky without employing legal counsel. In cases where I've seen a contributor license like this, it is not enforceable unless it's "signed". Not recommended! Example: https://cla.developers.google.com/about/google-individual?csw=1

Let's also keep in mind, we depend on 3rd party libraries which we may have modified, included or linked. Just because we modify the readme doesn't mean that the entire codebase isn't already grandfathered into GPLv3 anyways. For example, Qt is GPL.

There have been discussions about commercial entities not using ethereum because of GPL. This simply isn't accurate -- enterprises use GPL licensed software. See: Linux. The question is: What parts of the codebase do core developers need upstream contributions? What parts of the codebase will others want to change without contributing upstream?

Answering the first question makes the second irrelevant. What we could do is license these libraries as GLP+linking...
* ethash
* ethcore (excluding CommonJS)
* devcrypto (Trie code)
* Binaries/Tools (ex: Mix)
* p2p (move to MIT post-audit and when devp2p "1.0" is out)
* ethereum (for v1; v2 this is "stable" and moves away from GPL)

...and the rest as MIT. This leaves the entire API (incl EVM) open for tweaking and limits copyleft licensing to kernel-like components and major developer tools.

I do think voluntary legal counsel (attorneys for GNU/EFF/CC/Wikipedia) would be able to help determine if a contributor agreement is necessary for relicensing to MIT. Until then, it maybe easier to do GLP+linking where we know there aren't upstream licenses (Qt/microhttpd/etc).

Does this approach work for how you envision using the ethereum blockchain?

vonj:
@gavofyork 
@subtly 

Excellent that I have your attention, thank you. Regarding the enforceability of developer license agreement - you might be right. When it comes to transfer of right. But as things are now, cpp-ethereum accepts change-request after change-request, all implicitly licensed with the GPL. This is a problem. All "core" patches should have a more liberal license, such as X11 or MIT.

"Let's also keep in mind, we depend on 3rd party libraries which we may have been modified, included or linked. Just because we modify the readme doesn't mean that the entire codebase isn't implicitly limited to GPL anyways. For example, Qt is GPL."

I am interested mostly in the core parts of cpp-ethereum - what would be an Ethereum library. But anyway, QT is LGPL, not GPL, so QT does *not* implicitly make the entire codebase GPL.


"There have been discussions about commercial entities not using ethereum because of GPL. This simply isn't accurate -- enterprises use GPL licensed software. See: Linux."

This argument is just not complete enough - "GPL is used in one commercial context - therefore it can be used in all commercial contexts". The iOS Apple Store does not allow GPL, not even LGPL. That is one problem. Another problem is that if someone, say I, creates a proprietary, closed source program, how can I link it to cpp-ethereum (core only!) and sell it on the App Store, or anywhere else?

"The question is: What parts of the codebase do core developers need upstream contributions? What parts of the codebase will others want to change without contributing upstream?"

This is a good question. LGPL for me strikes a sweet spot in theory - it can be linked to proprietary code, but it protects changes to cpp-ethereum core and requires to it to be made public. *However*, LGPL is not compatible with the App Store, because of the static linking used on iOS. LGPL + static linking exception would be OK, or an MIT or X11 license.

"Answering the first question makes the second irrelevant."

As I hope I have demonstrated, the second is not irrelevant.

"What we could do is license these libraries as GLP+linking..."

LGPL + static linking exception would be fine, and X11 or MIT license would be even simpler. I am only interested in library parts, what you would need to integrate Ethereum into your own application.


If you followed me through here, I will come to my most important point:

it's fine to talk about re-licensing at some point in the future. But how do you propose to do that in practice? A lot of contributors have by now contributed code, I presume under the GPL license.

Gavin can not step in and just declare that all their code is now not GPL. @gavofyork, you can declare that *your* core-library code is now suddenly not GPL but instead LGPL+static linking exception, or MIT, or whatever. But you cannot declare that other peoples code is now licensed under some other license.

As we speak, this problem grows with every commit. This is called "letting time making the decision". Actually it's pretty similar to the blockchain itself... ;-)  The further you wait, the harder it is to go back and "undo" a decision as more and more blocks (or commits) are added. A decision was made to use GPL. At some point, it will become impossible from a practical standpoint to change license. I hope this is still possible.

PR #1401, opened 2015-03-24
Expected state

winsvega:
check Expected results when filling state tests


libethereum/Transaction.cpp
```diff
@@ -68,6 +68,10 @@ Transaction::Transaction(bytesConstRef _rlpData, CheckSignature _checkSig)
 		m_type = rlp[field = 3].isEmpty() ? ContractCreation : MessageCall;
 		m_receiveAddress = rlp[field = 3].isEmpty() ? Address() : rlp[field = 3].toHash<Address>(RLP::VeryStrict);
 		m_value = rlp[field = 4].toInt<u256>();
+
+		if (!rlp[field = 5].isData())
+			BOOST_THROW_EXCEPTION(BadRLP() << errinfo_comment("transaction data RLP must be an array"));
```
CJentzsch: this is already in another PR from you. You need to take this out here. Otherwise it will give merge conflicts for Gav.

test/TestHelper.cpp
```diff
@@ -172,6 +173,61 @@ void ImportTest::importTransaction(json_spirit::mObject& _o)
 	}
 }
 
+bool ImportTest::compareStates(State const& _stateExpect, State const& _statePost)
+{
+	//TestNet Addresses
+	static Addresses testNetAddressList = boost::assign::list_of
```
CJentzsch: you don't need that. You can use this constructor to get an empty state:
myState(Address(), OverlayDB(), eth::BaseState::Empty)

test/TestHelper.cpp
```diff
@@ -172,6 +173,61 @@ void ImportTest::importTransaction(json_spirit::mObject& _o)
 	}
 }
 
+bool ImportTest::compareStates(State const& _stateExpect, State const& _statePost)
```
CJentzsch: Everything you do in this function, I do here: https://github.com/ethereum/cpp-ethereum/blob/develop/test/state.cpp#L99-115
Please use the same helper function and unify this code. So we have no code duplication and can have one simple function for both cases.

test/TestHelper.cpp
```diff
@@ -180,8 +236,16 @@ void ImportTest::exportTest(bytes const& _output, State const& _statePost)
 	// export logs
 	m_TestObject["logs"] = exportLog(_statePost.pending().size() ? _statePost.log(0) : LogEntries());
 
-	// export post state
+	// compare expected state with post state
+	if (m_TestObject.count("expect") > 0)
```
CJentzsch: Please make commandline flag for that, or only give a wanring if this fails. Because we may have protocol changes, which will give a different post state (most certainly). So it should be possible to check onkly single value in the "expect" onject. For example only that the value at key 0 is 1  (for one address). But dont compare the complete post state.

winsvega: that is how it is right now. if something is set in expected section, than it would be checked.
so you could check only sender account, or only coinbase account for certain state in the post state.

test/stInitCodeTestFiller.json
```diff
@@ -8,6 +8,22 @@
...
+                }
+            },
+	    "a94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
+                "balance" : "67401",
+                "code" : "0x",
+                "nonce" : "1",
+                "storage" : {
+                }
+            }
+	},
```
CJentzsch: I want to do something like 
"a94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
                "nonce" : "1",
            }
Because when ther is a protocol change, I will only check one value, like the nonce here. Or a single value in storage. Not the complete account, that is not that useful. Because we already have that in the post after creating the test. (or in the "old" tests)

test/stInitCodeTestFiller.json
```diff
@@ -501,6 +839,29 @@
...
+                }
+            },
+            "a94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
+                "balance" : "99941972",
+                "code" : "0x",
+                "nonce" : "1",
+                "storage" : {
+                }
+            }
+        },
```
CJentzsch: all those expects are the same like in the post state. I can do a git diff in the new tests to see that. As mentioned, please only  check the most critical thing, like single storage value, a return value, or a single balance. Not the complete state.

test/stInitCodeTestFiller.json
winsvega: done

test/TestHelper.h
```diff
@@ -97,6 +97,17 @@ namespace test
 	}																	\
 	while (0)
 
+struct ImportStateOptions
+{
+	ImportStateOptions():m_bHasBalance(false), m_bHasNonce(false), m_bHasCode(false), m_bHasStorage(false)	{}
```
CJentzsch: You do not need this constructor. Just use ImportStateOptions(bool _bSetAll = false) in the other constructor. This way _bSetAll will default to false if not set explicitly.

test/TestHelper.cpp
```diff
@@ -172,6 +201,74 @@ void ImportTest::importTransaction(json_spirit::mObject& _o)
 	}
 }
 
+void ImportTest::checkExpectedState(State const& _stateExpect, State const& _statePost, WhenError _throw)
```
CJentzsch: you do not need this function when you use default values for ImportTest::checkExpectedState in line 210

winsvega: there is stateOptionsMap const& in second constructor. which requires an instance of stateOptionsMap. 

it could be changed to stateOptionsMap const = stateOptionsMap(), but thus it would not be transfered via &

test/TestHelper.cpp
```diff
@@ -172,6 +201,68 @@ void ImportTest::importTransaction(json_spirit::mObject& _o)
...
+
+				//Check for unexpected storage values
+				stateStorage = _stateExpect.storage(a.first);
+				for (auto const& s: _statePost.storage(a.first))
+					CHECK(stateStorage[s.first] == s.second,
+							"Check State: " << a.first <<  ": incorrect storage [" << s.first << "] = " << toHex(s.second) << ", expected [" << s.first << "] = " << toHex(stateStorage[s.first]));
+			}
+
+			if (addressOptions.m_bHasCode)
+			{
```
gavofyork: unneeded parens (and many more all over this PR)

test/TestHelper.cpp
```diff
@@ -102,44 +102,73 @@ void ImportTest::importEnv(json_spirit::mObject& _o)
...
-			_state.setStorage(address, toInt(j.first), toInt(j.second));
+		if (o.count("storage") > 0)
+		{
+			stateOptions.m_bHasStorage = true;
+			for (auto const& j: o["storage"].get_obj())
+				_state.setStorage(address, toInt(j.first), toInt(j.second));
+		}
 
-		for(int i=0; i<toInt(o["nonce"]); ++i)
+		for (int i=0; i<iNonce; ++i)
```
CJentzsch: space between i < iNonce

test/TestHelper.cpp
```diff
@@ -172,6 +201,68 @@ void ImportTest::importTransaction(json_spirit::mObject& _o)
...
+				}
+				catch(std::out_of_range)
+				{
+					BOOST_ERROR("expectedStateOptions map does not match expectedState in checkExpectedState!");
+					break;
+				}
+			}
+
+			if (addressOptions.m_bHasBalance)
+			{
```
CJentzsch: dont need parens for this one liner

test/TestHelper.cpp
```diff
@@ -172,6 +201,68 @@ void ImportTest::importTransaction(json_spirit::mObject& _o)
...
+				CHECK(_stateExpect.balance(a.first) == _statePost.balance(a.first),
+						"Check State: " << a.first <<  ": incorrect balance " << _statePost.balance(a.first) << ", expected " << _stateExpect.balance(a.first));
+			}
+
+			if (addressOptions.m_bHasNonce)
+			{
```
CJentzsch: dont need parens for this one liner

test/TestHelper.cpp
```diff
@@ -102,44 +102,73 @@ void ImportTest::importEnv(json_spirit::mObject& _o)
...
-			_state.setStorage(address, toInt(j.first), toInt(j.second));
+		if (o.count("storage") > 0)
+		{
+			stateOptions.m_bHasStorage = true;
+			for (auto const& j: o["storage"].get_obj())
+				_state.setStorage(address, toInt(j.first), toInt(j.second));
+		}
 
-		for(int i=0; i<toInt(o["nonce"]); ++i)
+		for (int i=0; i < iNonce; ++i)
```
CJentzsch: i = 0 instead i=0

test/TestHelper.cpp
```diff
@@ -106,44 +106,73 @@ void ImportTest::importEnv(json_spirit::mObject& _o)
...
 	for (auto& i: _o)
 	{
 		json_spirit::mObject o = i.second.get_obj();
 
-		assert(o.count("balance") > 0);
-		assert(o.count("nonce") > 0);
-		assert(o.count("storage") > 0);
-		assert(o.count("code") > 0);
+		ImportStateOptions stateOptions;
+		u256 balance = 0, nonce = 0;
```
CJentzsch: only one declaration  per line: https://github.com/ethereum/cpp-ethereum/blob/develop/CodingStandards.txt#L110

test/TestHelper.cpp
```diff
@@ -184,8 +271,17 @@ void ImportTest::exportTest(bytes const& _output, State const& _statePost)
...
 	m_TestObject["logs"] = exportLog(_statePost.pending().size() ? _statePost.log(0) : LogEntries());
 
-	// export post state
+	// compare expected state with post state
+	if (m_TestObject.count("expect") > 0)
+	{
+		stateOptionsMap stateMap;
+		State expectState(OverlayDB(), eth::BaseState::Empty);
+		importState(m_TestObject["expect"].get_obj(), expectState, stateMap);
+		checkExpectedState(expectState, _statePost, stateMap, Options::get().checkState ? WhenError::Throw : WhenError::DontThrow);
```
CJentzsch: instead of stateMap, you can just do stateOptionsMap(), since you do not use stateMap otherwise, So you can delet line 277.

test/TestHelper.cpp
```diff
@@ -106,44 +106,73 @@ void ImportTest::importEnv(json_spirit::mObject& _o)
...
 
 		_state.ensureCached(address, false, false);
+		_stateOptionsMap[address] = stateOptions;
 	}
 }
 
+void ImportTest::importState(json_spirit::mObject& _o, State& _state)
+{
+	stateOptionsMap importedMap;
+	importState(_o, _state, importedMap);
```
CJentzsch: you can just do stateOptionsMap() instead of importedMap.

gavofyork:
builds all failed - could we get them bumped to check it passes?

winsvega:
test-cpp-strict strict test failed on other PRs too

gavofyork:
http://ethereum-buildboard.meteor.com/ shows all well...

gavofyork:
windows fails

PR #1402, opened 2015-03-24
fixed numeber of http server threads, fixed #1147

### LINKED ISSUE #1147: Fix HttpServer threads number
currently, default 50

https://github.com/cinemast/libjson-rpc-cpp/blob/master/src/jsonrpccpp/server/connectors/httpserver.h#L50

debris:
changed number of servers from 50 -> 4
in debug drom 50 -> 1


gavofyork:
Better if these are constants, but i'll do it myself to expedite matters.

PR #1406, opened 2015-03-25
split #1351, fixed dependency on windows.h include order when using std::min && std::max

debris:
[stackoverflow issue](http://stackoverflow.com/questions/4913922/possible-problems-with-nominmax-on-visual-c)


subtly:
Any idea if we can do something like this for the boost asio thing? Is this related? (boost asio is included everywhere so that windows.h isn't included first)

PR #1408, opened 2015-03-25
split #1351, libtestutils

debris:
Includes ClientBase . Please review #1407 first.

PR #1407, opened 2015-03-25
split #1351, ClientBase

debris:

PR #1410, opened 2015-03-25
split #1351, tests

debris:
Parallel tests for new ClientBase class

includes:

- #1407 ClientBase
- #1408 libtestutils
- #1405 eth_add_test macro

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -30,6 +30,7 @@ target_link_libraries(testeth ethereum)
 target_link_libraries(testeth ethcore)
 target_link_libraries(testeth secp256k1)
 target_link_libraries(testeth solidity)
+target_link_libraries(testeth testutils)
 if (NOT HEADLESS AND NOT JUSTTESTS)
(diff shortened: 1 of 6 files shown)


libtestutils/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+
+file(GLOB HEADERS "*.h")
+
+if (ETH_STATIC)
+	add_library(${EXECUTABLE} STATIC ${SRC_LIST} ${HEADERS})
+else()
+	add_library(${EXECUTABLE} SHARED ${SRC_LIST} ${HEADERS})
+endif()
+
+target_link_libraries(${EXECUTABLE} ${LEVELDB_LIBRARIES})
```
chriseth: Link dependencies seem wrong. Should be something like
libwebthree, libdevcore, libethereum, libdevcrypto, some json stuff and boost filesystem.

It should not depend on: leveldb, mhd, secp256k1, solidity, serpent.

Issue #1397, opened 2015-03-24
Run Transactions never finishes in Mix tutorial.


Smithgift:
In the course of hunting down another issue, I discovered the current Mix tutorial is broken. After a setRating transaction is added to the state, that state will never finish.

Specifically, it will display "Running transactions..." at the top forever, and in the terminal, puke:

/usr/include/boost/multiprecision/cpp_int.hpp(1396): Throw in function void       boost::multiprecision::backends::cpp_int_backend<MinBits, MaxBits, SignType, Checked, Allocator>::do_assign_string(const char*, const false_&) [with unsigned int MinBits = 0u; unsigned int MaxBits = 0u; boost::multiprecision::cpp_integer_type SignType = (boost::multiprecision::cpp_integer_type)1u; boost::multiprecision::cpp_int_check_type Checked = (boost::multiprecision::cpp_int_check_type)0u; Allocator = std::allocator<long long unsigned int>; mpl_::false_ = mpl_::bool_<false>]
Dynamic exception type: boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<std::runtime_error> >
std::exception::what: Unexpected character encountered in input.
qrc:/qml/StatusPane.qml:82: TypeError: Cannot read property 'length' of null

I have attempted to duplicate the tutorial's project as close as possible, several times, and it has always done this. I can get one of these projects up on pastebin or somewhere if necessary.

EDIT: This is the absolute latest Ubuntu binary from ethereum-dev.

yann300:
I have made a pr for those two bugs. Please could you close these 2 issues when it's merged and working on your side?

Smithgift:
The latest update fixed it, yes. Thank you.

Issue #1386, opened 2015-03-23
Mix - check if function still exists in contract before executing a state


yann300:
and display user friendly error message.
Now displayed message is:
mix::TransactionSettings>&, const std::map<dev::FixedHash<32u>, boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<256u, 256u, (boost::multiprecision::cpp_integer_type)0u, (boost::multiprecision::cpp_int_check_type)0u, void> >

PR #1413, opened 2015-03-25
Mix - Bug fixes

yann300:
- Bugfix: #1397. 
- Debugger panel: UI bug fixes.

### DIFF
--- a/mix/ClientModel.cpp
+++ b/mix/ClientModel.cpp
@@ -242,7 +242,12 @@ void ClientModel::executeSequence(std::vector<TransactionSettings> const& _seque
 								break;
 							}
 					if (!f)
-						BOOST_THROW_EXCEPTION(FunctionNotFoundException() << FunctionName(transaction.functionId.toStdString()));
+					{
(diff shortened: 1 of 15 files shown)


gavofyork:
rebase.

mix/ClientModel.cpp
```diff
@@ -350,6 +350,10 @@ void ClientModel::showDebuggerForTransaction(ExecutionResult const& _t)
 			CompiledContract const* contract = contracts[s.codeIndex];
 			AssemblyItem const& instruction = codeItems[s.codeIndex][instructionIndex];
 
+			std::stringstream str;
+			str << instruction.getLocation().sourceName;
+			qDebug() <<  QString::fromStdString(str.str());
```
arkpar: Do we really need that in debug output?

yann300: I will delete that later, I have to fix one other bug before.

2015-03-26 11:06 GMT+01:00 Arkadiy Paronyan <notifications@github.com>:

arkpar: Does not compile. MixClient::bc should return eth::BlockChain

yann300:
I've pushed a fix for #1386.

PR #1426, opened 2015-03-26
fix #1415 for neth

### LINKED ISSUE #1415: Building without FATDB causes InterfaceNotSupported exception in neth
using `eth -i`:
```
> listcontracts
libc++abi.dylib: terminating with uncaught exception of type dev::InterfaceNotSupported: Interface State::addresses() not supported.
```
Same error when starting neth.

Introduced in 36e2c9d

caktux:


chriseth:
This code is horrible in various ways, but the change kind of looks good...

PR #1429, opened 2015-03-26
p2p: forge Ping ipAddress from UDP packet headers when not public

subtly:
This is due to previous changes to ignore unsolicited packets and is an interim solution until discovery is improved for NAT. Specifically, if host A pings host B and host A advertises '0.0.0.0' address, host B will ignore host A. Now we properly handle unspecified addresses by adopting the UDP address iff it's public. This will be reversed when discovery is updated and NAT hosts are able to determine their public address from remote hosts via ping/pong/etc packets.

Also fixed a unit test and added mutex.


gavofyork:
fails on windows.

PR #1434, opened 2015-03-26
encoding hex && numbers, fixed #1433

### LINKED ISSUE #1433: RPC return value for 0
return values of 0x00 should be 0x

debris:
rpc will return `0x`

### DIFF
--- a/libdevcore/CommonJS.h
+++ b/libdevcore/CommonJS.h
@@ -38,7 +38,10 @@ template <unsigned S> std::string toJS(FixedHash<S> const& _h)
 
 template <unsigned N> std::string toJS(boost::multiprecision::number<boost::multiprecision::cpp_int_backend<N, N, boost::multiprecision::unsigned_magnitude, boost::multiprecision::unchecked, void>> const& _n)
 {
-	return "0x" + toHex(toCompactBigEndian(_n, 1));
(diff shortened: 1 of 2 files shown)


gavofyork:
this breaks latest json rpc spec

debris:
I fixed pr, now it behaves exactly as described [here](https://github.com/ethereum/cpp-ethereum/issues/1433#issuecomment-87050619)

libdevcore/CommonJS.h
```diff
@@ -38,7 +38,10 @@ template <unsigned S> std::string toJS(FixedHash<S> const& _h)
 
 template <unsigned N> std::string toJS(boost::multiprecision::number<boost::multiprecision::cpp_int_backend<N, N, boost::multiprecision::unsigned_magnitude, boost::multiprecision::unchecked, void>> const& _n)
 {
-	return "0x" + toHex(toCompactBigEndian(_n, 1));
+	std::string h = toHex(toCompactBigEndian(_n, 1));
+	// remove first 0, if it is necessary;
+	std::string res = h[0] != '0' ? h : h.substr(1);
```
gavofyork: how does this render `0`? it looks very much like it'll come out with `"0x"`, which is against the spec.

PR #1440, opened 2015-03-27
removed [null] log on filter init

debris:


gavofyork:
no good - for 'pending'/'latest' filters, this transactionHash will still be null, yet we need to return something.

Issue #661, opened 2014-12-18
macos compiler flags for 1stgen i5 and core2duo


subtly:

debris:
@subtly , can you check if #1454 fixes the issue?

subtly:
Update/TL;DR
I dug into my notes and resolving this in cmake would mean compiling cryptopp from source. The easiest solution would be to modify the homebrew recipe for cryptopp on the macos build-server. It's also possible that by now clang has been updated to this is no longer an issue; or, if the build-server is running in a VM or older CPU, this may not be an issue anyways.

debris:
@subtly can we close this? We will reopen if someone reports that it's not working on possibly affected machine.

PR #1454, opened 2015-03-30
-msse3 clang flag

debris:
#661

### DIFF
--- a/cmake/EthCompilerSettings.cmake
+++ b/cmake/EthCompilerSettings.cmake
@@ -25,8 +25,10 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
 	set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O2 -g -DETH_DEBUG")
 	set(ETH_SHARED 1)
 
+	# For macos we need to set CLANG_X86_VECTOR_INSTRUCTIONS to ssse3
+	# This should result in binaries which will execute on Apple hardware which are < Mavericks
 	if ("${CMAKE_SYSTEM_NAME}" MATCHES "Linux")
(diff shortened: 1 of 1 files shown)


cmake/EthCompilerSettings.cmake
```diff
@@ -25,8 +25,10 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
 	set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O2 -g -DETH_DEBUG")
 	set(ETH_SHARED 1)
 
+	# For macos we need to set CLANG_X86_VECTOR_INSTRUCTIONS to ssse3
+	# This should result in binaries which will execute on Apple hardware which are < Mavericks
 	if ("${CMAKE_SYSTEM_NAME}" MATCHES "Linux")
-		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libstdc++ -fcolor-diagnostics -Qunused-arguments -DBOOST_ASIO_HAS_CLANG_LIBCXX")
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libstdc++ -fcolor-diagnostics -Qunused-arguments -DBOOST_ASIO_HAS_CLANG_LIBCXX -msse3")
```
gavofyork: this is the linux condition - shouldn't this be under an apple/darwin condition?

subtly: This doesn't fix this issue. Also worth noting, I think CMAKE_SYSTEM_NAME is Linux even when it's macos.

subtly:
@debris This won't fix #661. I know for sure that setting CLANG_X86_VECTOR_INSTRUCTIONS to ssse3 works; not sure what -msse3 does. As noted; this should only apply for macos builds.

debris:
After discussing issue with @subtly, I'm closing the pr. #661 needs to be updated.

Issue #1448, opened 2015-03-27
Some mappings seem to be broken in mix.


Smithgift:
On the latest Ubuntu dev binary, mappings seem to be returning erroneous data when accessed through its accessor or a function. For example, in the mix tutorial's Rating contract, setting a movie's rating to 1234, and then accessing the rating later returns some number over five trillion. I have experimented with other mappings, such as address to uint which seems to work and bytes32 to uint and uint to uint which seem to not work.

Smithgift:
Fixed by pr, thanks.

Issue #1447, opened 2015-03-27
Mix's Edit State does not allow giving parameters to constructors.


Smithgift:
In the latest Ubuntu dev binary, when the Edit State's Edit Transaction dialog is creating a contract, it has a blank panel where the parameters for the constructor would be, even if the constructor accepts arguments..

Smithgift:
Fixed by pr, thanks.

PR #1457, opened 2015-03-30
Mix - Bug fixes

yann300:
- #1448
 - #1447

### DIFF
--- a/mix/ContractCallDataEncoder.cpp
+++ b/mix/ContractCallDataEncoder.cpp
@@ -159,7 +159,11 @@ dev::bytes ContractCallDataEncoder::decodeBytes(dev::bytes const& _rawValue)
 
 QString ContractCallDataEncoder::toString(dev::bytes const& _b)
 {
-	return QString::fromStdString(dev::toJS(_b));
+	QString str;
+	if (isString(_b, str))
+		return "- " + str +  " - " + QString::fromStdString(dev::toJS(_b));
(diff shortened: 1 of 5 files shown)


mix/ContractCallDataEncoder.cpp
```diff
@@ -196,3 +201,18 @@ QStringList ContractCallDataEncoder::decode(QList<QVariableDeclaration*> const&
...
 	return r;
 }
+
+
+bool ContractCallDataEncoder::isString(dev::bytes const& _b, QString& _str)
+{
+	dev::bytes bunPad = unpadded(_b);
+	for (unsigned i = 0; i < bunPad.size(); i++)
+	{
+		u256 value(bunPad.at(i));
```
arkpar: should by just byte here instead of u256

mix/ContractCallDataEncoder.cpp
```diff
@@ -196,3 +201,18 @@ QStringList ContractCallDataEncoder::decode(QList<QVariableDeclaration*> const&
 	}
 	return r;
 }
+
+
+bool ContractCallDataEncoder::isString(dev::bytes const& _b, QString& _str)
```
arkpar: name is a bit misleading as this function also does string conversion. asString perhaps?

yann300: Done

Issue #1459, opened 2015-03-30
RPC: eth_getCode returns wrong state at block 0


frozeman:
Sending:

```
{"jsonrpc":"2.0","method":"eth_getCode","params":["0x6295ee1b4f6dd65047762f924ecd367c17eabf8f","0x0"],"id":5}
```

Should return '0x60003560e060020a...' according to the loaded blockstate at https://github.com/ethereum/tests/blob/develop/BlockTests/bcJS_API_Test.json#L448

Issue #1458, opened 2015-03-30
RPC: eth_getStorageAt returns wrong state at block 0


frozeman:
Sending:

```
{"jsonrpc":"2.0","method":"eth_getStorageAt","params":["0x6295ee1b4f6dd65047762f924ecd367c17eabf8f","0x0","0x0"],"id":3}
```

Should return `0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b` according to the loaded teststate file at https://github.com/ethereum/tests/blob/develop/BlockTests/bcJS_API_Test.json#L451

PR #1461, opened 2015-03-30
failing tests, incomplete genesis state object

debris:
related with #1458, #1459, #1460.

Client method [`asOf`](https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/Client.cpp#L559) uses [this State constructor](https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/State.cpp#L122) to get state with given block hash. Calling method asOf with hash of block 0 returns incomplete state object. I expect it to contain data about pre state.

Reproducable with `test.ClientBase` target, or by running:

```
testeth --run_test=ClientBase --eth_testfile=BlockTests/bcJS_API_Test --eth_threads=1
```

### DIFF
--- a/test/ClientBase.cpp
+++ b/test/ClientBase.cpp
@@ -34,34 +34,45 @@ BOOST_AUTO_TEST_CASE(blocks)
 {
 	enumerateClients([](Json::Value const& _json, dev::eth::ClientBase& _client) -> void
 	{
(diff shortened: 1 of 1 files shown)


gavofyork:
fixed in latest develop. tests would still be good though.

debris:
great! the tests are now passing. Can we merge this pr?

Issue #1366, opened 2015-03-20
Network HostCap accessing dereferenced Cap


subtly:

subtly:
Fixed in #1462

PR #1462, opened 2015-03-30
Fix for #1366.

subtly:
Explicit instantiation to prevent reference and force copy of shared-ptr.

### DIFF
--- a/libp2p/Host.h
+++ b/libp2p/Host.h
@@ -96,7 +96,7 @@ public:
 	static std::string pocHost();
 
 	/// Register a peer-capability; all new peer connections will have this capability.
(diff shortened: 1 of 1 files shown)


subtly:
Confirmed via unit test akin to:
	Host host1("Test", NetworkPreferences("127.0.0.1", 30303, false));
	{
		host1.registerCapability(new dev::shh::WhisperHost());
	}
	host1.start();
	sleep(1);

Without this change, placing a log in dealloc of WhisperHost will show WhisperHost deallocating even though there should remain a shared-ptr in m_capabilities map. It maybe worth looking into this more, but based on the test above, WhisperHost shouldn't remain allocated b/c web3 has a weak_ptr and not shared_ptr. Maybe there is some undefined behavior here due to the combination of shared_ptrs, inheritance, and auto.

PR #1464, opened 2015-03-30
Detect if non-existant parameter is documented with natspec

LefterisJP:


libsolidity/InterfaceHandler.cpp
```diff
@@ -175,8 +175,17 @@ std::unique_ptr<std::string> InterfaceHandler::getDevDocumentation(ContractDefin
 				method["author"] = m_author;
 
 			Json::Value params(Json::objectValue);
+			std::vector<std::string> paramNames = it.second->getParameterNames();
```
LianaHus: you can remove std::

LianaHus: you can remove "std::"s

LefterisJP: I'll do it in a separate PR.

The whole file needs it's `std::`s removed.

Issue #1400, opened 2015-03-24
Just built the latest mix and I cant paste into the contract editor or the javascript editor.


KenKappler:
I can still paste into the html editor and the css window - so not sure what the problem is. OSX

PR #1478, opened 2015-03-31
Mix: Fixed editor clipboard for os x

arkpar:


mix/qml/html/codeeditor.js
```diff
@@ -199,3 +189,5 @@ compilationComplete = function()
 	}
 	compilationCompleteBool = true;
 }
+
+editor.setOption("extraKeys", extraKeys);
```
yann300: Im not sure that extraKeys at this point contains key used by ternserver.

arkpar: This works because CM keeps reference to extraKeys object, so it can be modified at any time

arkpar:
#1400

PR #1482, opened 2015-03-31
jsonrpc eth_version method

debris:
added [eth_version](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_version)


libweb3jsonrpc/WebThreeStubServerBase.cpp
```diff
@@ -284,6 +284,11 @@ bool WebThreeStubServerBase::net_listening()
 	return network()->isNetworkStarted();
 }
 
+string WebThreeStubServerBase::eth_version()
```
gavofyork: rename to `eth_protocolVersion` (version is a bit too vague, given the plethora of version ids we have)

debris:
renamed

PR #1483, opened 2015-03-31
mix - Logs window redesign

yann300:


mix/qml/LogsPane.qml
```diff
@@ -7,81 +7,259 @@ import "."
...
+	}
+
+	onVisibleChanged:
+	{
+		if (visible && (logsModel.count === 0 || (logsModel.get(0).date !== currentStatus.date && logsModel.get(0).content !== currentStatus.content)))
+			logsModel.insert(0, { "type": currentStatus.type, "date": currentStatus.date, "content": currentStatus.content, "level": currentStatus.level });
+		else if (!visible)
+		{
+			for (var k = 0; k < logsModel.count; k++)
+			{
```
LianaHus: I think you don't need braces here

PR #1484, opened 2015-03-31
Mix: Debug highlighting filter

arkpar:


mix/CodeModel.cpp
```diff
@@ -93,8 +93,24 @@ class CollectDeclarationsVisitor: public ASTConstVisitor
...
+
+dev::eth::AssemblyItems filterLocations(dev::eth::AssemblyItems const& _locations, dev::solidity::ContractDefinition const& _contract, QHash<LocationPair, QString> _functions)
+{
+	dev::eth::AssemblyItems result;
+	result.reserve(_locations.size());
+	std::string sourceName = *_contract.getLocation().sourceName;
+	for (dev::eth::AssemblyItem item : _locations)
+	{
+		dev::SourceLocation const& l = item.getLocation();
+		if (sourceName != *l.sourceName || _contract.getLocation() == l || _functions.contains(LocationPair(l.start, l.end)))
```
yann300: We are also filtering using the sourceName? because at some point it could be interesting to highlight the code in the targeted sourceName (if available).

PR #1489, opened 2015-04-01
libabi init + tests [in progress]

debris:
**changes:**

- moved code from abi executable to library (called abil, needs to be renamed)
- created dev::abi namespace (I'm not sure about it)
- added CommanLineInterface class
- created tests for dev::abi::CommandLineInterface - `test/Abi.cpp`
- added `TransientFile` class in libtestutils

**abi executable issues:**
 - when encoding methods with number of input params > 0, we need to add additional command line param, which is unused
 - formatting negative ints is not correct (see encode_negative_int test_case)
 - usage of --format-prefix is not clear or do not work
 - missing encoding for events

**decoding was not yet tested**


libabi/Abi.cpp
```diff
@@ -0,0 +1,521 @@
...
+
+bytes ABIMethod::encode(vector<pair<bytes, Format>> const& _params) const
+{
+	bytes ret = name.empty() ? bytes() : id().asBytes();
+	bytes suffix;
+	
+	// int int[] int
+	// example: 42 [ 1 2 3 ] 69
+	// int[2][][3]
+	// example: [ [ [ 1 2 3 ] [ 4 5 6 ] ] [ ] ]
```
chriseth: Not concerning this PR, but that change should be made once this PR is merged:
`int[2][][3]` should be a 3-array of an array of a 2-array of ints, i.e. the example should be
`[ [ [ 1 2 ] [ 3 4 ] ] [ ] [ ] ]`.
Furthermore, there are plans to change the encoding of nested dynamic arrays (those are not really covered in the specification).

PR #1491, opened 2015-04-01
Solidity refactoring. Assembly and AssemblyItem are no longer friend classes

LianaHus:
added set functions for type and data members of AssemblyItem


chriseth:
"fiend" :-)

caktux:
Integration builder failed twice waiting for `getTransactionCount` to increase, but it could be caused by something in develop as it started to happen randomly in the past day or two.

PR #1501, opened 2015-04-02
Abstract contract contructors

LefterisJP:
Implementing last bullet point of the [Abstract contracts story](https://www.pivotaltracker.com/story/show/88344782) converning constructors of base contracts and proper checks for providing all arguments. 

If not all arguments to base constructors are provided the contract is marked as abstract.

### DIFF
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -166,7 +166,7 @@ void ContractDefinition::checkAbstractConstructors()
 		if (constructor)
 		{
 			if (!constructor->getParameters().empty())
-				argumentsProvided.insert(constructor);
+				argumentsNeeded.insert(constructor);
 			for (auto const& modifier: constructor->getModifiers())
 			{
 				auto baseContract = dynamic_cast<ContractDefinition const*>(
@@ -174,8 +174,8 @@ void ContractDefinition::checkAbstractConstructors()
 				if (baseContract)
 				{
 					FunctionDefinition const* baseConstructor = baseContract->getConstructor();
-					if (argumentsProvided.count(baseConstructor) == 1)
-						argumentsNeeded.insert(baseConstructor);
+					if (argumentsNeeded.count(baseConstructor) == 1)
+						argumentsProvided.insert(baseConstructor);
 				}
 			}
 		}
@@ -186,10 +186,13 @@ void ContractDefinition::checkAbstractConstructors()
 				base->getName()->getReferencedDeclaration());
 			solAssert(baseContract, "");
 			FunctionDefinition const* baseConstructor = baseContract->getConstructor();
-			if (argumentsProvided.count(baseConstructor) == 1)
-				argumentsNeeded.insert(baseConstructor);
+			if (argumentsNeeded.count(baseConstructor) == 1)
+				argumentsProvided.insert(baseConstructor);
 		}
 	}
+	// add this contract's constructor to the provided too
+	if (getConstructor() && !getConstructor()->getParameters().empty())
+		argumentsProvided.insert(getConstructor());
 	if (argumentsProvided != argumentsNeeded)
 		setFullyImplemented(false);
 }
--- a/libsolidity/AST.cpp
+++ b/libsolidity/AST.cpp
@@ -728,7 +728,7 @@ void NewExpression::checkTypeRequirements()
 	if (!m_contract)
 		BOOST_THROW_EXCEPTION(createTypeError("Identifier is not a contract."));
 	if (!m_contract->isFullyImplemented())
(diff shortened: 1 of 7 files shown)


libsolidity/AST.cpp
```diff
@@ -281,7 +323,7 @@ void InheritanceSpecifier::checkTypeRequirements()
 	ContractDefinition const* base = dynamic_cast<ContractDefinition const*>(m_baseName->getReferencedDeclaration());
 	solAssert(base, "Base contract not available.");
 	TypePointers parameterTypes = ContractType(*base).getConstructorType()->getParameterTypes();
-	if (parameterTypes.size() != m_arguments.size())
+	if (m_arguments.size() != 0 && parameterTypes.size() != m_arguments.size())
```
LefterisJP: will change to `empty()`

mix/CodeModel.cpp
```diff
@@ -135,8 +135,8 @@ CompiledContract::CompiledContract(const dev::solidity::CompilerStack& _compiler
 
 	CollectDeclarationsVisitor visitor(&m_functions, &m_locals, &m_storage);
 	contractDefinition.accept(visitor);
-	m_assemblyItems = filterLocations(_compiler.getRuntimeAssemblyItems(name), contractDefinition, m_functions);
-	m_constructorAssemblyItems = filterLocations(_compiler.getAssemblyItems(name), contractDefinition, m_functions);
+	m_assemblyItems = filterLocations(*_compiler.getRuntimeAssemblyItems(name), contractDefinition, m_functions);
```
LefterisJP: @arkpar There is a minor, tiny touch to Mix's codebase here so I think it's prudent to also notify you.

libsolidity/AST.h
```diff
@@ -376,7 +377,7 @@ class EnumValue: public Declaration
 
 	virtual void accept(ASTVisitor& _visitor) override;
 	virtual void accept(ASTConstVisitor& _visitor) const override;
-	TypePointer getType(ContractDefinition const* = nullptr) const;
+	TypePointer getType(ContractDefinition const* = nullptr) const override;
```
LianaHus: "virtual"?

LefterisJP:
Note: Some end to end test are failing. Looking into it.

libsolidity/AST.h
LefterisJP: thanks! Will do

LefterisJP:
rebased

libsolidity/AST.cpp
```diff
@@ -152,6 +153,50 @@ void ContractDefinition::checkAbstractFunctions()
...
+		{
+			ContractDefinition const* baseContract = dynamic_cast<ContractDefinition const*>(
+				base->getName()->getReferencedDeclaration());
+			solAssert(baseContract, "");
+			FunctionDefinition const* baseConstructor = baseContract->getConstructor();
+			if (argumentsNeeded.count(baseConstructor) == 1)
+				argumentsProvided.insert(baseConstructor);
+		}
+	}
+	// add this contract's constructor to the provided too
```
chriseth: Wouldn't it be better to remove them from "needed"?

libsolidity/AST.cpp
```diff
@@ -152,6 +153,50 @@ void ContractDefinition::checkAbstractFunctions()
...
+			if (!constructor->getParameters().empty())
+				argumentsNeeded.insert(constructor);
+			for (auto const& modifier: constructor->getModifiers())
+			{
+				auto baseContract = dynamic_cast<ContractDefinition const*>(
+					modifier->getName()->getReferencedDeclaration());
+				if (baseContract)
+				{
+					FunctionDefinition const* baseConstructor = baseContract->getConstructor();
+					if (argumentsNeeded.count(baseConstructor) == 1)
```
chriseth: baseConstructor might be nullptr - we should not insert that into argumentsProvided. Please add a test for that case.

chriseth: Why this check here? There should be an assertion that we do not provide arguments unless they are needed. Also note that `baseConstructor` might be `nullptr` here.

libsolidity/AST.cpp
```diff
@@ -152,6 +153,50 @@ void ContractDefinition::checkAbstractFunctions()
...
+			}
+		}
+
+		for (ASTPointer<InheritanceSpecifier> const& base: contract->getBaseContracts())
+		{
+			ContractDefinition const* baseContract = dynamic_cast<ContractDefinition const*>(
+				base->getName()->getReferencedDeclaration());
+			solAssert(baseContract, "");
+			FunctionDefinition const* baseConstructor = baseContract->getConstructor();
+			if (argumentsNeeded.count(baseConstructor) == 1)
```
chriseth: Same applies here.

chriseth:
Please also add an end to end test that uses this feature.

PR #1503, opened 2015-04-02
Control flow analysis.

chriseth:
Remove unused tags/blocks and rearrange blocks to remove unconditional jumps.

This makes it possible that the following contract
```
contract c {
function g(uint x) internal returns (uint) { return x**2; }
function f(uint a, uint b) returns (uint) { return g(b); }
}
```
is compiled to assembly code that has only a single jump (from the function selector) and only loads `b` from calldata, i.e. it inlines `g` and removes the unnecessary loading of `a`.

### DIFF
--- a/libevmcore/ControlFlowGraph.cpp
+++ b/libevmcore/ControlFlowGraph.cpp
@@ -1,0 +1,260 @@
+/*
+	This file is part of cpp-ethereum.
+
+	cpp-ethereum is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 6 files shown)


gavofyork:
fails tests.

LianaHus:
I'm not completly sure, but in void ControlFlowGraph::removeUnusedBlocks() try to replace erase(it++) with it = erase(it). with vectors this definitly couse problems but I'm not sure about maps.

LianaHus:
erase(it) returns next valid iterator. but should check what is garanteed by standarts for map.(for all platforms) in general modifying container in loops always could be dangarous

libevmcore/ControlFlowGraph.cpp
```diff
@@ -0,0 +1,260 @@
...
+		{
+			neededBlocks.insert(block.next);
+			blocksToProcess.push_back(block.next);
+		}
+	}
+	for (auto it = m_blocks.begin(); it != m_blocks.end();)
+		if (neededBlocks.count(it->first))
+			++it;
+		else
+			m_blocks.erase(it++);
```
LianaHus: Are you sure you don't want to change this?

chriseth: Yes, map::erase does not return an iterator and the specification says that all iterators except the one being deleted stay valid.

PR #1514, opened 2015-04-04
less tolerant peer fallback timer

subtly:


libp2p/Session.cpp
```diff
@@ -358,16 +359,11 @@ void Session::drop(DisconnectReason _reason)
 		}
 		catch (...) {}
 
-	if (m_peer)
```
gavofyork: removing this could be bad, no? why is it now guaranteed to be non-null?

subtly: It's required by the constructor. If the session exists, the peer exists.

PR #1519, opened 2015-04-05
update and enable whisper tests

subtly:


gavofyork:
you call SIGABORT & SIGSEGV passing? :-P

```
!N! [ 21:19:28 | p2p ] Couldn't start accepting connections on host. Failed to accept socket.
 Throw location unknown (consider using BOOST_THROW_EXCEPTION)
Dynamic exception type: boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >
std::exception::what: bind: Address already in use

terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >'
  what():  bind: Address already in use
unknown location(0): fatal error in "topic": signal: SIGABRT (application abort requested)
/slave/build-cpp-ethereum-pr/build/test/SolidityScanner.cpp(283): last checkpoint
unknown location(0): fatal error in "forwarding": signal: SIGSEGV, si_code: 0 (memory access violation at address: 0x00000080)
```

subtly:
This is by design and due to another eth process running on the build server and using the same port used by a test (verified this was possible w/caktux). Just as runs between dropping a database should expect that another database doesn't magically appear in filesystem, tests which don't use random ports must be able to run in a clean environment.
The test *explicitly* sets a listen port, which if used by another process, will fail. The SIGABRT is likely ASIO killing off it's worker threads. A user which explicitly sets a port expects the software to use that port, not a random port. There are several use-cases which require this and it's important from a perspective of security; specifically, given an explicit configuration, the software cannot be coerced into a different configuration. Admittedly, the error handling could see some improvements :)

subtly:
Green :)

gavofyork:
where is it green?

subtly:
Closed?

PR #1523, opened 2015-04-06
WebPreview has libjsqrc dependencies

kobigurk:


mix/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ set(EXECUTABLE mix)
 
 if ("${Qt5WebEngine_VERSION_STRING}" VERSION_GREATER "5.3.0")
 	set (ETH_HAVE_WEBENGINE TRUE)
-	qt5_add_resources(UI_RESOURCES web.qrc)
+	qt5_add_resources(UI_RESOURCES web.qrc ../libjsqrc/js.qrc)
```
arkpar: The resource is already added with jsqrc static lib below. It was not initialized properly in the code, this is already fixed in latest develop

kobigurk:
Thanks, I see the fix!

PR #1528, opened 2015-04-06
Mix: fixed debugging for array parameters/storage

arkpar:
Also added tests for array debugging and tutorial DApp


mix/CodeModel.cpp
```diff
@@ -238,6 +238,13 @@ CompiledContract const& CodeModel::contract(QString _name) const
 	return *res;
 }
 
+CompiledContract const* CodeModel::tryGetContract(QString _name) const
```
yann300: const& _name ?

arkpar: QString is reference counted, passing it by reference does not optimize much

PR #1531, opened 2015-04-06
Make currect some of the testing framework

gavofyork:
Doesn't actually fix it (broken in libsol anyway, it seems), but does make more correct certain parts (though seems to break a lot of tests :-/)

- Ensure enough gas is in the sender account and that the sender is actually that acocunt, not just some randomly generated one
- Ensure the gas given to the call is <= block limit

### DIFF
--- a/test/solidityExecutionFramework.h
+++ b/test/solidityExecutionFramework.h
@@ -40,7 +40,7 @@ namespace test
 class ExecutionFramework
 {
 public:
-	ExecutionFramework() { g_logVerbosity = 0; }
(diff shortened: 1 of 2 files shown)


test/solidityExecutionFramework.h
```diff
@@ -40,7 +40,7 @@ namespace test
 class ExecutionFramework
 {
 public:
-	ExecutionFramework() { g_logVerbosity = 0; }
+	ExecutionFramework() { g_logVerbosity = 14; }
```
chriseth: I would rather only change this during a debugging session.

PR #1533, opened 2015-04-06
Add create costs to contract creation tx

CJentzsch:


gavofyork:
rebase needed.

CJentzsch:
The question is, do we want this PR in now? Because it would break consensus with the other implementations. When you have decided I will rebase.

gavofyork:
true - don't have to keep it in sync in the meantime i guess..

PR #1537, opened 2015-04-07
ported fancy colorful console from geth to mix

debris:
- fixed displaying functions in mix console
- nicer colorful console output in mix

was:
screen shot 2015-04-07 at 03 55 50

is:
screen shot 2015-04-07 at 03 55 00


arkpar:
Looks cool, thanks!

PR #1540, opened 2015-04-07
Mix - Various fixes

yann300:
iconSource in JavaScript console


mix/qml/StatusPane.qml
```diff
@@ -281,24 +284,15 @@ Rectangle {
...
+					iconSource: "qrc:/qml/img/bugiconactive.png"
 					action: debugRunActionIcon
-					states: [
-						State{
-							name: "active"
-							PropertyChanges { target: debugImg; iconSource: "qrc:/qml/img/bugiconactive.png"}
-						}
-					]
 				}
 				Action {
```
arkpar: I think this Action should be removed completely. There is already an action to toggle right pane in Application.qml

Issue #1525, opened 2015-04-06
mix: go to next problem shortcut


gavofyork:
need a shortcut to let us jump directly to the exact site of the (next) compilation issue.

yann300:
Is that needed for POC9? For now this is not available in the sol library.

arkpar:
We can do "jump to current error" for now

yann300:
Just need some ui redesign by Alex

arkpar:
implemented

PR #1549, opened 2015-04-07
MACOS bug fixes

yann300:
- MACOS bugs fixes.
 - go to next problem shortcut #1525


mix/qml/Application.qml
```diff
@@ -398,4 +398,14 @@ ApplicationWindow {
 		enabled: !projectModel.isEmpty && codeModel.hasContract
 		onTriggered: projectModel.deployProject();
 	}
+
+	Action {
+		id: goToCompilationError
+		text: qsTr("Go to compilation error")
+		shortcut: "Ctrl+Shift+E"
```
arkpar: This should be something more accessible I think. How about F4?

yann300:
TODO: 
- Test cases.
- Go to compilation btn redesign.

Issue #1526, opened 2015-04-06
mix: Automate gas/balance requirement.


gavofyork:
Mix always seems to throw on deploy due to lack of gas or cash. AZ's Transact dialog now has code which predeploys the contract in an environment faked with maximum gas to check how much gas is actually required. Check it and use what you can so this error never shows up unless you're not using the automatic/default amount of gas.

PR #1550, opened 2015-04-07
Mix: Auto gas estimation

arkpar:
Old projects will have auto gas checkbox off. It is on for new projects by default.


mix/ClientModel.cpp
```diff
@@ -173,6 +181,7 @@ void ClientModel::setupState(QVariantMap _state)
 		QString contractId = transaction.value("contractId").toString();
 		QString functionId = transaction.value("functionId").toString();
 		u256 gas = boost::get<u256>(qvariant_cast<QBigInt*>(transaction.value("gas"))->internalValue());
+		bool gasAuto = transaction.value("gas").toBool();
```
yann300: "gas" mean two different things? should it be transaction.value("gasAuto") instead of transaction.value("gas") ?

arkpar:
#1526

PR #1562, opened 2015-04-08
Fixing the build

LefterisJP:
I guess that line is redundant


debris:
looks good, and tests are passing. merging

Issue #1296, opened 2015-03-12
pysol: Compilation issues.


simondlr:
Hi. I've been trying to get solidity to work with pyethereum's tester. I first need to install the python wrapper.

In pysol, when I run setup.py, there are 2 errors. It doesn't find CommonEth & Exceptions.cpp. This is just because it is old & not yet been fixed up. I will submit a pull request once I figure out the rest of the issues.

I encounter a problem where it tries to find 'json/json.h' in ASTJsonConverter.h. 

I assume this is because I need to cmake the CMakeList in libsolidity (which seems to create json.h?). I'm not entirely sure if this is the correct assumption. When I do run cmake, it works. But when I run make, there's several errors.

The first of which is warnings stating things C++11 extensions. I'm posting the logs here. Unsure what I'm doing wrong? Doing cmake wrong? Other errors? Please help.

```
Simon-2:libsolidity simon$ make
Scanning dependencies of target solidity
[  5%] Building CXX object CMakeFiles/solidity.dir/ArrayUtils.o
In file included from /Users/simon/code/ethereum/cpp-ethereum/libsolidity/ArrayUtils.cpp:24:
In file included from /Users/simon/code/ethereum/cpp-ethereum/libsolidity/../libevmcore/Instruction.h:24:
In file included from /Users/simon/code/ethereum/cpp-ethereum/libsolidity/../libdevcore/Common.h:46:
/Users/simon/code/ethereum/cpp-ethereum/libsolidity/../libdevcore/vector_ref.h:16:21: warning: alias declarations are a C++11 extension [-Wc++11-extensions]
        using value_type = _T;
                           ^
/Users/simon/code/ethereum/cpp-ethereum/libsolidity/../libdevcore/vector_ref.h:17:23: warning: alias declarations are a C++11 extension [-Wc++11-extensions]
        using element_type = _T;
                             ^
/Users/simon/code/ethereum/cpp-ethereum/libsolidity/../libdevcore/vector_ref.h:18:29: warning: alias declarations are a C++11 extension [-Wc++11-extensions]
        using mutable_value_type = typename std::conditional<std::is_const<_T>::value, typename std::remove_const<_T>::type, _T>::type;
                                   ^
/Users/simon/code/ethereum/cpp-ethereum/libsolidity/../libdevcore/vector_ref.h:28:2: warning: explicit conversion functions are a C++11 extension [-Wc++11-extensions]
        explicit operator bool() const { return m_data && m_count; }
[...]

PR #1565, opened 2015-04-09
pysol work in progress

LefterisJP:
WARNING: Work in progress don't merge.

Just some attempts to address #1296. Will work on it more when I find more time.
Not really solved anything yet. It's a work in progress.

I am not familiar with the python `setup-tools` module enough but what I think is the proper solution for this is:

- Do not include the source files into the final package. Solidity itself is actually a library.
- Use the cpp-ethereum build files to build solidity
- The final package created by setup-tools should somehow link to the built libraries. This is the part I have not figured out yet. How can you package the libsolidity.so into a way that python will find it later? Is there any standard way in python? If I set the `LD_PATH` manually or if I copy it into the path this works.

### DIFF
--- a/pysol/MANIFEST.in
+++ b/pysol/MANIFEST.in
@@ -1,13 +1,5 @@
 include pysol/*.cpp
 include *.py
-include libdevcore/*cpp
-include libdevcore/*h
-include libdevcrypto/*cpp
-include libdevcrypto/*h
-include libethcore/*cpp
-include libethcore/*h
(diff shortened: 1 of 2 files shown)


heikoheiko:
We reverted to call `solc` as a subprocess, which works good enough. 
https://github.com/ethereum/pyethereum/commit/f89042b74fb4b053414b38b93e91f70836f0c500

LefterisJP:
That's great then, closing this.

Issue #1560, opened 2015-04-08
JS api inconsistent on mix (too functional.)


Smithgift:
Bizarrely, mix works TOO well. The code from #1545, for example, will show web3 is functional even before the page load on mix, whereas AZ will neither acknowledge web3 before the page is loaded or during the body tag's onload. This gives new developers (such as say, myself) the false impression that their initialization code will be run on a real system, when, in fact, it will not.

Aside: Is body onload really supposed to be not-guaranteed to have web3 when it triggers? This would require a kludge for any initialization code.

gavofyork:
perhaps @arkpar and @debris can shed some light onto this?

debris:
I have checked and that's true. Mix code editor is an iframe with [web3 object injected from parent](https://github.com/ethereum/cpp-ethereum/blob/develop/mix/qml/WebPreview.qml#L187) **before** the page is loaded. Whereas in alethzero we are [adding web3 object](https://github.com/ethereum/cpp-ethereum/blob/develop/alethzero/MainWin.cpp#L186-L192) **after** the page is loaded. We should unify this in AZ, mix and mist.

I opt for delivering web3 object as fast as we can.

Smithgift:
I can confirm web3 is now existent before page load. I think the issue can be closed now.

PR #1566, opened 2015-04-09
Load html urls through a proxy in AZ

arkpar:
Regular pages are now served through built-in Http server same way as DApp packages to allow js injection.
#1560


alethzero/DappLoader.cpp
```diff
@@ -96,13 +97,31 @@ DappLocation DappLoader::resolveAppUri(QString const& _uri)
...
 void DappLoader::downloadComplete(QNetworkReply* _reply)
 {
+	QUrl requestUrl = _reply->request().url();
+	if (m_pageUrls.count(requestUrl) != 0)
+	{
+		//inject web3 js
+		QByteArray content = "<script>\n";
+		content.append(web3Content());
+		content.append("</script>\n");
+		content.append(_reply->readAll());
```
yann300: In that case, the script tag containing the web3 obj will be added outside of the HTML tag.
Can this cause side effect as this is not standard?

arkpar: It works in chromium and we don't have to care about compatibility with anything else here

PR #1570, opened 2015-04-09
BlockTests with Expected State

winsvega:


test/blockchain.cpp
```diff
@@ -287,12 +287,24 @@ void doBlockchainTests(json_spirit::mValue& _v, bool _fillin)
 					blObj.erase(blObj.find("blockHeader"));
 					blObj.erase(blObj.find("uncleHeaders"));
 					blObj.erase(blObj.find("transactions"));
+					state = State(OverlayDB(), BaseState::Empty, biGenesisBlock.coinbaseAddress);
+					importer.importState(o["pre"].get_obj(), state);
```
CJentzsch: This is wrong. The state should be according to the stateRoot in the last valid block, not the genesis (pre) state.

test/TestHelper.cpp
```diff
@@ -671,7 +671,6 @@ Options::Options()
 			memory = true;
 			inputLimits = true;
 			bigData = true;
-			checkState = true;
```
CJentzsch: why?

PR #1572, opened 2015-04-09
UI

alexvandesande:
Minor stylistic changes:
* executing code has an underline instead of a light background
* removed shadow from text on error
* added support for high res icons on some elements
* Changed the “go to error” icon


alexvandesande:
I don't see where to add tags.. so "please review".

PR #1581, opened 2015-04-10
update whisper tests

subtly:


gluk256:
It should run smoothly after #2088 is merged.

PR #1593, opened 2015-04-11
fixed #1590, msvc build with cmake 3.2.1

### LINKED ISSUE #1590: msvc build with cmake 3.2.1 fails

debris:


CMakeLists.txt
```diff
@@ -1,10 +1,22 @@
...
+
 # let cmake autolink dependencies on windows
 # it's specified globally, cause qt libraries requires that on windows and they are also found globally
 cmake_policy(SET CMP0020 NEW)
 
-project(ethereum)
+# 3.1 and above
+if ((${CMAKE_MAJOR_VERSION} GREATER 2) AND (${CMAKE_MINOR_VERSION} GREATER 0))
+	# implicitly dereference variables (deprecated in 3.1)
+	cmake_policy(SET CMP0054 OLD)
```
arkpar: Is this the problem with our cmake files? Is it possible to fix the cmake files to confirm to a new policy instead?

debris: I'm just a little bit afraid about new policy. From cmake description

>CMake 3.1 and above no longer implicitly dereference variables or
interpret keywords in an ``if()`` command argument when
it is a :ref:`Quoted Argument` or a :ref:`Bracket Argument`.

>The ``OLD`` behavior for this policy is to dereference variables and
interpret keywords even if they are quoted or bracketed.
The ``NEW`` behavior is to not dereference variables or interpret keywords
that have been quoted or bracketed.

>Given the following partial example:
```cmake
 set(MONKEY 1)
 set(ANIMAL MONKEY)
 if("${ANIMAL}" STREQUAL "MONKEY")
```
After explicit expansion of variables this gives:
```cmake
 if("MONKEY" STREQUAL "MONKEY")
```
With the policy set to ``OLD`` implicit expansion reduces this semantically to:
```cmake
 if("1" STREQUAL "1")
```
With the policy set to ``NEW`` the quoted arguments will not be
further dereferenced:
```cmake
 if("MONKEY" STREQUAL "MONKEY")
```

arkpar: The new policy looks reasonable. If you want ANIMAL to be alias, the proper way to define it is
```cmake
set(ANIMAL ${MONKEY})
```
This should work for both old and new version

debris:
I've changed the policy CMP0054 to new. I've also merged #1693 into this branch to make it compilable on windows.

libdevcore/Terminal.h
```diff
@@ -79,6 +79,14 @@ namespace con
 #define EthCyanOnI ""    // Cyan
 #define EthWhiteOnI ""   // White
 
+#define EthNavy ""
```
gavofyork: not necessary any more.

gavofyork:
you can remove the windows stuff now (and fwiw since it was a build issue, you could have merged that fix, *separately* immediately on review)

debris:
windows stuff removed

CMakeLists.txt
```diff
@@ -1,10 +1,22 @@
 # cmake global
 cmake_minimum_required(VERSION 2.8.12)
+
+project(ethereum)
+
+set(CMAKE_AUTOMOC ON)
+
+# link_directories interprate relative paths with respect to CMAKE_CURRENT_SOURCE_DIR
```
chriseth: "interpr_e_te"

PR #1607, opened 2015-04-14
catch cerr in neth

caktux:
catch cerr in neth
~~do not save first transaction signer as default~~

### DIFF
--- a/neth/main.cpp
+++ b/neth/main.cpp
@@ -626,6 +626,7 @@ int main(int argc, char** argv)
 
 	logwin = newwin(height * 2 / 5 - 2, width * 2 / 3, qheight, 0);
 	nc::nc_window_streambuf outbuf(logwin, std::cout);
+	nc::nc_window_streambuf eoutbuf(logwin, std::cerr);
 
 	consolewin   = newwin(qheight, width / 4, 0, 0);
 	nc::nc_window_streambuf coutbuf(consolewin, ccout);


libdevcore/Log.cpp
```diff
@@ -43,7 +43,7 @@ void dev::simpleDebugOut(std::string const& _s, char const*)
 	static Mutex s_lock;
 	Guard l(s_lock);
 
-	cerr << _s << endl << flush;
+	cout << _s << endl << flush;
```
gavofyork: logs should be going to std err by default. if you want them going elsewhere for neth, do as AZ does and register a different function in `g_logPost`.

caktux: I don't think logs should be going to std err by default, only program errors, no matter what log/debug level is being used. The most obvious effect of that is on the waterfall page, where it previously only got an `eth.err` logfile if there were actually errors, now we get both `eth.log` and `eth.err` no matter what, so it's impossible to tell whether there really was an error or not.

gavofyork: in the case of eth, there is "expected, domain-specific output" (stdout) and "out of band output" (stderr); logs are out of band and include debug info, notes, warnings and critical errors. 

it's a convention that i employ and it isn't going to change. if you want to employ different meanings for neth, please do.

caktux: You never employed that convention since this project started, and it made a lot more sense that way. I'll remove that so you can merge the fix to neth.

caktux:
mind merging?

PR #1618, opened 2015-04-14
Mix - Test Cases for Miner Selection.

yann300:


mix/test/qml/TestMain.qml
```diff
@@ -75,6 +76,11 @@ TestCase
 	function test_dbg_transactionWithParameter() { TestDebugger.test_transactionWithParameter(); }
 	function test_dbg_constructorParameters() { TestDebugger.test_constructorParameters(); }
 	function test_dbg_arrayParametersAndStorage() { TestDebugger.test_arrayParametersAndStorage(); }
+<<<<<<< HEAD
```
arkpar: Looks like the file is not merged properly

PR #1633, opened 2015-04-15
Attempt a fix of dagger tests to fix the build

LefterisJP:


chriseth:
Unfortunately, there is no tag "looksmuchbetter" :-).

PR #1640, opened 2015-04-15
fix for ipv4 private address spaces

debris:
it should be:
172.16.0.0 - 172.31.255.255

according to wikipedia:
http://en.wikipedia.org/wiki/Private_network

### DIFF
--- a/libp2p/Common.cpp
+++ b/libp2p/Common.cpp
@@ -48,7 +48,9 @@ bool p2p::isPrivateAddress(bi::address const& _addressToCheck)
 		bi::address_v4::bytes_type bytesToCheck = v4Address.to_bytes();
 		if (bytesToCheck[0] == 10 || bytesToCheck[0] == 127)
 			return true;
-		if (bytesToCheck[0] == 172 && (bytesToCheck[1] >= 16 && bytesToCheck[1] <= 31))
+		if (bytesToCheck[0] == 172 && bytesToCheck[1] >= 16)
+			return true;
+		if (bytesToCheck[0] == 173 && bytesToCheck[1] <= 31)
 			return true;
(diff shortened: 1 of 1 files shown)


libp2p/Common.cpp
```diff
@@ -48,7 +48,9 @@ bool p2p::isPrivateAddress(bi::address const& _addressToCheck)
 		bi::address_v4::bytes_type bytesToCheck = v4Address.to_bytes();
 		if (bytesToCheck[0] == 10 || bytesToCheck[0] == 127)
 			return true;
-		if (bytesToCheck[0] == 172 && (bytesToCheck[1] >= 16 && bytesToCheck[1] <= 31))
+		if (bytesToCheck[0] == 172 && bytesToCheck[1] >= 16)
+			return true;
+		if (bytesToCheck[0] == 173 && bytesToCheck[1] <= 31)
```
subtly: 173.31.0.0 isn't a private address

libp2p/Common.cpp
```diff
@@ -48,7 +48,9 @@ bool p2p::isPrivateAddress(bi::address const& _addressToCheck)
 		bi::address_v4::bytes_type bytesToCheck = v4Address.to_bytes();
 		if (bytesToCheck[0] == 10 || bytesToCheck[0] == 127)
 			return true;
-		if (bytesToCheck[0] == 172 && (bytesToCheck[1] >= 16 && bytesToCheck[1] <= 31))
+		if (bytesToCheck[0] == 172 && bytesToCheck[1] >= 16)
```
subtly: This will return true for 172.32.0.1, which is a public IP.

libp2p/Common.cpp
debris: nvm

PR #1642, opened 2015-04-15
Don't use unsigned for negative value.

subtly:
Prevents client from trying to connect to every peer when peercount > idealpeercount.


subtly:
@chriseth Agreed. Might also exclude auto when the container is a temporary (return from a method).

PR #1645, opened 2015-04-16
Removed dependencies on external tools from getstuff.bat

arkpar:


arkpar:
WiP. Getting random download fails. Same issues as with curl probably

arkpar:
Resolved

Issue #1571, opened 2015-04-09
solc contract constructor json-abi


debris:
Can we add contract constractor abi to solc output?

for this contract:

```solidity
contract hello {
    function hello() {
        Hello();
    }
}
```

abi looks like this:

```json
[]
```

but I would like it to be like this:

```json
[
   {
      "constant" : false,
      "inputs" : [],
      "name" : "hello",
      "outputs" : [],
      "type" : "constructor"
   }
]
```

chriseth:
This is fixed now, though the output is a bit different:
```
[
  {
    "constant" : false,
    "inputs" : [],
    "type" : "constructor" 
  }
]
```
I.e. it neither has a "name" nor a "outputs" attribute. The reason is that the constructor is not called like a function, so the name is not needed to compute the signature hash and it might even be misleading since in other languages, the constructor might be called differently (note that the ABI is not only for solidity).

PR #1646, opened 2015-04-16
Solidity: add constructor for ABI json

LianaHus:
https://github.com/ethereum/cpp-ethereum/issues/1571

### DIFF
--- a/libsolidity/AST.h
+++ b/libsolidity/AST.h
@@ -281,9 +281,6 @@ public:
 	/// Returns the fallback function or nullptr if no fallback function was specified.
 	FunctionDefinition const* getFallbackFunction() const;
 
-	///@returns hash of the constructor
-	FixedHash<4> getConstructorsInterface() const;
-
 private:
 	/// Checks that two functions defined in this contract with the same name have different
 	/// arguments and that there is at most one constructor.
--- a/libsolidity/AST.h
+++ b/libsolidity/AST.h
@@ -281,6 +281,9 @@ public:
 	/// Returns the fallback function or nullptr if no fallback function was specified.
 	FunctionDefinition const* getFallbackFunction() const;
 
+	///@returns hash of the constructor
+	FixedHash<4> getConstructorsInterface() const;
+
 private:
 	/// Checks that two functions defined in this contract with the same name have different
 	/// arguments and that there is at most one constructor.
--- a/libsolidity/InterfaceHandler.cpp
+++ b/libsolidity/InterfaceHandler.cpp
@@ -47,7 +47,6 @@ std::unique_ptr<std::string> InterfaceHandler::getABIInterface(ContractDefinitio
 		allFunctions.insert(make_pair(_contractDef.getConstructorsInterface(), functionTypePointer));
 	}
 
-	//allFunctions.insert(_contractDef.getConstructor());
 	for (auto it: allFunctions)
 	{
 		auto populateParameters = [](vector<string> const& _paramNames, vector<string> const& _paramTypes)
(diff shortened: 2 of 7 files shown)


libsolidity/AST.h
```diff
@@ -281,6 +281,9 @@ class ContractDefinition: public Declaration, public Documented, public Implemen
 	/// Returns the fallback function or nullptr if no fallback function was specified.
 	FunctionDefinition const* getFallbackFunction() const;
 
+	///@returns hash of the constructor
+	FixedHash<4> getConstructorsInterface() const;
```
chriseth: This should not be needed, the constructor is not called in such a way.

libsolidity/InterfaceHandler.cpp
```diff
@@ -38,7 +38,16 @@ std::unique_ptr<std::string> InterfaceHandler::getDocumentation(ContractDefiniti
 std::unique_ptr<std::string> InterfaceHandler::getABIInterface(ContractDefinition const& _contractDef)
 {
 	Json::Value abi(Json::arrayValue);
-	for (auto const& it: _contractDef.getInterfaceFunctions())
+	auto allFunctions = _contractDef.getInterfaceFunctions();
```
chriseth: Please do not use auto here. We want this to be a copy.

libsolidity/InterfaceHandler.cpp
```diff
@@ -38,7 +38,16 @@ std::unique_ptr<std::string> InterfaceHandler::getDocumentation(ContractDefiniti
...
+
+	FunctionTypePointer functionTypePointer = nullptr;
```
chriseth: `constructorTypePointer`?

test/libsolidity/SolidityABIJSON.cpp
```diff
@@ -495,6 +495,23 @@ BOOST_AUTO_TEST_CASE(empty_name_return_parameter)
...
+{
+	char const* sourceCode = R"(
+		contract test {
+			function test() {
+		}
+	})";
+
+	char const* interface = R"([
+	{
+	   "inputs" : [],
```
chriseth: Could you change the test so that we also get types here? Perhaps even a conversion from a contract type to address?

libsolidity/Types.cpp
```diff
@@ -1244,9 +1264,15 @@ TypePointers FunctionType::parseElementaryTypeVector(strings const& _types)
 
 TypePointer FunctionType::copyAndSetGasOrValue(bool _setGas, bool _setValue) const
 {
-	return make_shared<FunctionType>(m_parameterTypes, m_returnParameterTypes, m_location,
-									 m_arbitraryParameters,
-									 m_gasSet || _setGas, m_valueSet || _setValue);
+	return make_shared<FunctionType>(
+		m_parameterTypes,
+		m_returnParameterTypes,
+		strings(),
```
chriseth: Please copy the names from the current object.

libsolidity/Types.cpp
```diff
@@ -1143,7 +1143,7 @@ FunctionTypePointer FunctionType::externalFunctionType() const
 			return FunctionTypePointer();
 		retParamTypes.push_back(type->externalType());
 	}
-	return make_shared<FunctionType>(paramTypes, retParamTypes, m_location, m_arbitraryParameters);
+	return make_shared<FunctionType>(paramTypes, retParamTypes, m_parameterNames, m_location, m_arbitraryParameters);
```
chriseth: You have to construct the names alongside the parameters, some of them are missing!
And you should also retain the names of the return parameters.

debris:
:+1:

Issue #1632, opened 2015-04-15
Difficulty set to 0


yann300:
*** [ 06:22:14 | main ] Opened blockchain DB. Latest: 85dffe5257e0a2a127dd34be75
2a0ae88451342ee7bdd4b77b606657ee8a113d
*** [ 06:22:14 | main ] Closing blockchain DB
QQmlComponent: Component destroyed while completion pending
..\..\libethcore\BlockInfo.cpp(163): Throw in function void __thiscall dev::eth:
:BlockInfo::populateFromHeader(const class dev::RLP &,enum dev::eth::Strictness,
const class dev::FixedHash<32> &)
Dynamic exception type: class boost::exception_detail::clone_impl<struct dev::et
h::InvalidDifficulty>
std::exception::what: Unknown exception
[struct dev::tag_got *] = 0
[struct dev::tag_required *] = 131072

PR #1647, opened 2015-04-16
Mix: windows issues

arkpar:


mix/MixClient.cpp
```diff
@@ -40,7 +40,7 @@ namespace mix
 {
 
 Secret const c_defaultUserAccountSecret = Secret("cb73d9408c4720e230387d956eb0f829d8a4dd2c1055f96257167e14e7169074");
-u256 const c_mixGenesisDifficulty = c_minimumDifficulty; //TODO: make it lower for Mix somehow
```
yann300: Maybe this constant is used in AZ or eth. So the value will be 0 in the windows build ...

yann300: issue #1632

Issue #1639, opened 2015-04-15
tests: Test RLP 0x81xx is invalid when xx < 128


winsvega:
https://github.com/ethereum/yellowpaper/issues/116

it is defined that if one byte is < 128 then it's RLP encoding should be the byte itself.
however if we put it as byte array length of 1 it is still a valid encoding. 

that is incorrect with YP st.160

gavofyork:
should be fixed now. needs tests.

winsvega:
tests
https://github.com/ethereum/tests/commit/ebe1f586678c1a631bafef23e3a96e3cf5180b60

PR #1648, opened 2015-04-16
RLP length with first zeros

winsvega:
>If a string is more than 55 bytes long, the RLP encoding consists of a single byte with value 0xb7 plus the length of the length of the string in binary form, followed by the length of the string, followed by the string.

leading zeros in  length of the string are not covered. 

https://github.com/ethereum/yellowpaper/issues/116

### DIFF
--- a/libdevcore/RLP.cpp
+++ b/libdevcore/RLP.cpp
@@ -149,7 +149,7 @@ unsigned RLP::length() const
 		return n - c_rlpDataImmLenStart;
 	else if (n < c_rlpListStart)
 	{
-		if ((int)m_data.size() <= n - c_rlpDataIndLenZero)
+		if ((int)m_data.size() <= n - c_rlpDataIndLenZero || m_data[1] == 0)
 			BOOST_THROW_EXCEPTION(BadRLP());
 		for (int i = 0; i < n - c_rlpDataIndLenZero; ++i)
 			ret = (ret << 8) | m_data[i + 1];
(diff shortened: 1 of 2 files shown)


gavofyork:
christoph - please review.

CJentzsch:
@winsvega do you have tests for that?

CJentzsch:
also please lways a little description/or link to the PR such as: https://github.com/ethereum/cpp-ethereum/issues/1639

winsvega:
Yes, I add to the test repo:

transaction RLP's
RLPListLengthWithFirstZeros
RLPArrayLengthWithFirstZeros

block RLPS's:
RLPLengthOfLengthWithFirstZeros

libdevcore/RLP.cpp
```diff
@@ -149,7 +149,7 @@ unsigned RLP::length() const
 		return n - c_rlpDataImmLenStart;
 	else if (n < c_rlpListStart)
 	{
-		if ((int)m_data.size() <= n - c_rlpDataIndLenZero)
+		if ((int)m_data.size() <= n - c_rlpDataIndLenZero || m_data[1] == 0)
```
CJentzsch: When I send you the wrong "RLP" : 0xb8, you will arrive here, and just end with a segfault, because m_data[1] does not exist. This means, when this PR is merged, I can send a malicous transaction with 0xb8 and crash the client.

winsvega: I also was worried about such issue. 
Should not it be that a compiler would first execute this expression
"(int)m_data.size() <= n - c_rlpDataIndLenZero"  and then stops as it already true?

CJentzsch: 1. we should not rely on this
2. In my case, m_data.size() = 1, n - c_rlpDataIndLenZero = 184 - 183 = 1. so the first statement is true. So it will evaluate m_data[1] and fail

winsvega:
tests: 
https://github.com/ethereum/tests/commit/ebe1f586678c1a631bafef23e3a96e3cf5180b60

PR #1653, opened 2015-04-16
various opencl related changes and a Blockchain.cpp bug fix

Genoil:
and some other stuff i thought i'd already gotten rid off. admittedly I'm terrible with git.

### DIFF
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -134,6 +134,7 @@ void help()
 		<< "    -f,--force-mining  Mine even when there are no transactions to mine (default: off)" << endl
 		<< "    -C,--cpu  When mining, use the CPU." << endl
 		<< "    -G,--opencl  When mining use the GPU via OpenCL." << endl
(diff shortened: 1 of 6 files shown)


chriseth:
Just a note: This pull request contains several merges. Github completely ignores them when showing changes and does not have a way to review how possible conflicts in these merges were resolved.
Could you perhaps open new pull request(s) that do not contain merges? It would also be great if you could split the pull request into opencl and blockqueue related.

Genoil:
Yes I'm aware of the mess. I'll try to clean this up.

libdevcore/CommonIO.h
```diff
@@ -48,6 +48,10 @@ std::string contentsString(std::string const& _file);
 /// Retrieve and returns the allocated contents of the given file. If the file doesn't exist or isn't readable, returns nullptr. Don't forget to delete [] when finished.
 bytesRef contentsNew(std::string const& _file, bytesRef _dest = bytesRef());
```
gavofyork: this does the same as `contentsToBuffer` when passed a second param, except that it's size-safe.

PR #1656, opened 2015-04-16
decent platforms do not need dlls

debris:


LefterisJP:
:+1: ... was really wondering why my make was not finding DLLs ..

LefterisJP:
Tested locally and solves my issue

PR #1666, opened 2015-04-17
0x -> 0x0 for numbers

CJentzsch:


test/TestHelper.cpp
```diff
@@ -140,7 +140,7 @@ json_spirit::mObject& ImportTest::makeAllFieldsHex(json_spirit::mObject& _o)
 			str = value.get_str();
 		else continue;
 
-		_o[key] = (str.substr(0, 2) == "0x") ? str : "0x" + toHex(toCompactBigEndian(toInt(str)));
+		_o[key] = (str.substr(0, 2) == "0x") ? str : "0x" + (toHex(toCompactBigEndian(toInt(str))).empty() ? "0" : toHex(toCompactBigEndian(toInt(str))));
```
chriseth: Setting the second parameter of `toCompactBigEndian` to 1 would result in `0x00`, right?

CJentzsch: Thanks, overlooked the option to toCompactBigEndian

PR #1679, opened 2015-04-20
non-const enqueueWriteBuffer ptr

subtly:
osx didn't like const


libethash-cl/ethash_cl_miner.cpp
```diff
@@ -277,13 +277,13 @@ void ethash_cl_miner::search(uint8_t const* header, uint64_t target, search_hook
 	};
 	std::queue<pending_batch> pending;
 
-	static uint32_t const c_zero = 0;
```
chriseth: Indentation

subtly:
Fixed.

subtly:
Moving to ethash.

PR #1678, opened 2015-04-20
Windows strftime compat

subtly:
%F and %T aren't supported on windows (Win7 x64, VS2013 Community).

PR #1690, opened 2015-04-20
Windows doesn't like %F %T datetime format

Genoil:
%F and %T aren't supported by Windows so I changed this into something that does the same and works on Windows. 

btw the output that is produced doesn't really make sense:
```
*** [ 17:38:40 | eth ] Since 1970-02-11 20:12:43.702Z (15): 15ticks
```
but at least it doesn't crash


gavofyork:
Linux displays a time fine with that - guess it's not exactly a full fix. Perhaps Windows has its own way of doing things, as usual.

subtly:
Yep. Windows crashes with %F %T.

subtly:
dupe PR #1678

PR #1693, opened 2015-04-21
fixed missing windows colors defines

debris:


chriseth:
I think this file should not be in libdevcore. Otherwise it looks good :-)

chriseth:
Ah, it does not, `EthOrage` is missing.

chriseth:
And another comment: What about using a macro for the two choices a la
```
#define EthOrage EthColor("\magicescapeforLinux", "\magicescapeforWindows")

debris:
I don't know if using escapes on windows is so simple...
According to [this stackoverflow](http://stackoverflow.com/questions/16755142/how-to-make-win32-console-recognize-ansi-vt100-escape-sequences) answer it is not. But there is [C library](https://github.com/adoxa/ansicon), we might want to use.

gavofyork:
someone else was already on the job...

PR #1696, opened 2015-04-21
Removed (and added) some #ifs (but removed more than I added).

chriseth:


test/libevmcore/CMakeLists.txt
```diff
@@ -1,5 +0,0 @@
-cmake_policy(SET CMP0015 NEW)
-
-aux_source_directory(. SRCS)
-
-add_sources(${SRCS})
```
CJentzsch: Why do you remove this?
The test in Assembly.cpp will not be included in testeth after this. Also every directory should have a CMake file.

CJentzsch: Ok I see, you move the file in the libsolidity folder. But you could still leave CMakeLists.txt as it is. So one can add test files there without touching the CMake file.

chriseth: I removed the whole directory because there was no .cpp file left afterwards.
Also note that you moved some libevmcore tests to the libsolidity folder already. I know that I did not draw a clear line between those tests, I will clean that up when we have more dedicated libevmcore tests.

PR #1698, opened 2015-04-21
Mix - use contract type in ctor param

yann300:
- allows write a ctor with contract type param and select the contract to use as a param in the transaction dialog.


mix/ClientModel.cpp
```diff
@@ -280,6 +282,16 @@ void ClientModel::executeSequence(vector<TransactionSettings> const& _sequence,
 					{
 						QSolidityType const* type = p->type();
 						QVariant value = transaction.parameterValues.value(p->name());
+						if (type->type().type == SolidityType::Type::Address && value.toString().startsWith("<"))
+						{
```
chriseth: No braces.

Issue #1558, opened 2015-04-08
update ping and pong packets


subtly:
Ping: Add destination endpoint. Rename endpoint property to 'destination' (cosmetic update).
Pong: Add destination endpoint to payload.

halfalicious:
It looks like this was done at some point. 
* PingNode destination: https://github.com/ethereum/aleth/blob/b867f9eba00996765fed90c3b627d86305d53dda/libp2p/NodeTable.h#L347-L358
* Pong destination: https://github.com/ethereum/aleth/blob/b867f9eba00996765fed90c3b627d86305d53dda/libp2p/NodeTable.h#L383-L391

Issue #1557, opened 2015-04-08
add additional port to discovery endpoint encapsulation


subtly:
As was originally specified! This makes network less fragmented from hosts which are behind NAT.

PR #1706, opened 2015-04-21
Discovery protocol updates

subtly:
* storing/sharing of udp+tcp ports and add endpoint data to ping/pong packets (1557,1558)
* bump network version
* RLP encapsulation of NodeIPEndpoint
* update save/restore nodes
* remove redundant addNode

#1557, #1558


libp2p/NodeTable.h
```diff
@@ -299,30 +296,20 @@ struct InvalidRLP: public Exception {};
...
 	static const uint8_t type = 1;
 
 	unsigned version = 0;
-	std::string ipAddress;
-//	uint16_t udpPort;
-	uint16_t tcpPort;
-	unsigned ts;
+	NodeIPEndpoint source;
+	NodeIPEndpoint destination;
+	uint32_t ts;
```
chriseth: Not always initialized.

libp2p/NodeTable.h
```diff
@@ -330,22 +317,20 @@ struct PingNode: RLPXDatagram<PingNode>
...
-	Pong(bi::udp::endpoint _ep): RLPXDatagram<Pong>(_ep), ts(futureFromEpoch(std::chrono::seconds(60))) {}
+	Pong(bi::udp::endpoint const& _ep): RLPXDatagram<Pong>(_ep), destination(UnspecifiedNodeIPEndpoint) {}
+	Pong(NodeIPEndpoint const& _dest): RLPXDatagram<Pong>((bi::udp::endpoint)_dest), destination(_dest), ts(secondsSinceEpoch()) {}
 
 	static const uint8_t type = 2;
 
+	NodeIPEndpoint destination;
 	h256 echo;				///< MCD of PingNode
-	unsigned ts;
+	uint32_t ts;
```
chriseth: Not always initialized.

libp2p/NodeTable.h
```diff
@@ -363,57 +348,45 @@ struct Pong: RLPXDatagram<Pong>
...
 {
 	FindNode(bi::udp::endpoint _ep): RLPXDatagram<FindNode>(_ep) {}
-	FindNode(bi::udp::endpoint _ep, NodeId _target, std::chrono::seconds _ts = std::chrono::seconds(60)): RLPXDatagram<FindNode>(_ep), target(_target), ts(futureFromEpoch(_ts)) {}
+	FindNode(bi::udp::endpoint _ep, NodeId _target): RLPXDatagram<FindNode>(_ep), target(_target), ts(secondsSinceEpoch()) {}
 
 	static const uint8_t type = 3;
 
 	h512 target;
-	unsigned ts;
+	uint32_t ts;
```
chriseth: Not always initialized.

chriseth:
Are there tests for ipv6?

subtly:
@chriseth simple IPv6 goes in after this is working, where simple means that only IPv4 or IPv6 is supposed (not both at once). It's supported by encapsulation but because the listener doesn't support IPv6 such nodes are simply not added to the table. IPv6 was removed previously because some changes are needed:
* additional address verification and NAT is different
* most IPv6 systems can connect to IPv4 systems, but reverse isn't true
* must consider considering IPv6 as another network (node table, networkid)

libp2p/NodeTable.h
```diff
@@ -330,22 +317,20 @@ struct PingNode: RLPXDatagram<PingNode>
...
-	Pong(bi::udp::endpoint _ep): RLPXDatagram<Pong>(_ep), ts(futureFromEpoch(std::chrono::seconds(60))) {}
+	Pong(bi::udp::endpoint const& _ep): RLPXDatagram<Pong>(_ep), destination(UnspecifiedNodeIPEndpoint) {}
+	Pong(NodeIPEndpoint const& _dest): RLPXDatagram<Pong>((bi::udp::endpoint)_dest), destination(_dest), ts(secondsSinceEpoch()) {}
 
 	static const uint8_t type = 2;
 
+	NodeIPEndpoint destination;
 	h256 echo;				///< MCD of PingNode
-	unsigned ts;
+	uint32_t ts = 0;
```
fjl: `uint32_t` for unix timestamps? well, it'll work until 2038.

fjl:
This is not interoperable with go-ethereum because IP addresses are encoded as RLP lists.
I think lists are not a good choice for IP addresses. But we need to decide.

    I0427 02:33:57.213023   87752 udp.go:404] Bad packet from 127.0.0.1:30303: rlp: expected input string or byte for net.IP, decoding into (discover.ping).From.IP

libp2p/Common.h
```diff
@@ -177,6 +185,9 @@ struct NodeIPEndpoint
 	operator bool() const { return !address.is_unspecified() && udpPort > 0 && tcpPort > 0; }
 	
 	bool isAllowed() const { return NodeIPEndpoint::test_allowLocal ? !address.is_unspecified() : isPublicAddress(address); }
+	
+	void streamRLP(RLPStream& _s, RLPAppend _inline = StreamList) const { if (_inline == StreamList) _s.appendList(3); if (address.is_v4()) _s << address.to_v4().to_bytes(); else if (address.is_v6()) _s << address.to_v6().to_bytes(); else _s << ""; _s << udpPort << tcpPort; }
+	void interpretRLP(RLP const& _r) { if (_r[0].size() == 0) address = bi::address(); else if (_r[0].size() == 4) address = bi::address_v4(_r[0].toArray<byte, 4>()); else address = bi::address_v6(_r[0].toArray<byte, 16>()); udpPort = _r[1].toInt<uint16_t>(); tcpPort = _r[2].toInt<uint16_t>(); }
```
fjl: Stylistic issues aside (296 characters in a line, including a three-way braceless conditional, oh my), there
seems to be some confusion about the type of the input here. 

`size()` only works for RLP strings and returns 0 otherwise, yet `toArray()` is used, which only works for lists. 
If IP addresses are supposed to be represented as lists, maybe this should switch on `itemCount()` instead.

subtly: @fjl Byte string makes sense. Is that supported by Go?

libp2p/NodeTable.cpp
```diff
@@ -514,16 +508,17 @@ void NodeTable::onReceived(UDPSocketFace*, bi::udp::endpoint const& _from, bytes
 			{
 				PingNode in = PingNode::fromBytesConstRef(_from, rlpBytes);
 				if (in.version != dev::p2p::c_protocolVersion)
+					return;
+				if (RLPXDatagramFace::secondsSinceEpoch() - in.ts > 3)
```
fjl: A packet with an expiration > current time is valid, all other packets are invalid. At least that is how 
all other implementations work.

The timestamps are `uint32` and the one sent by the other end is usually bigger than the current time.
Subtracting the remote timestamp from the current time will overflow into a very big number, causing
the packet to be considered invalid.

fjl: Looking at the other places, you seem to have mixed in a change where the expiration stamp is now the time when the packet was sent.

fjl:
Looking also at the other places where the timestamps are generated/used, you seem to have mixed in a change where the expiration stamp is now the time when the packet was sent.

Could you please undo this change? You changed this [three days ago in the spec](https://github.com/ethereum/devp2p/commit/14dfa136a795428e2514838be0ac1b885d03a5ec)
without telling anyone, and without any justification. The expiration that was there before works
just as well and everyone has it implemented.

subtly:
The expiration/timestamp system was, until now, undefined. I don't understand how it's been implemented in any client -- afaik, c++ has never had interop with Go timestamps because the feature was thrown out until now.

arkpar:
Compilation fails with msvc

gavofyork:
while there was no text to strictly define the meaning of `expiration`, i would interpret it as a unix time (seconds from beginning of UTC 1970) for the expiration time of the packet (i.e. throw out packets with a time in the past, throw out packets with a time much > present).

unfortunately, this is a slightly awkward way of doing it - much better to note the actual timestamp and let the receiver determine what should be done.

all in all, it should always have been properly defined and the lack of proper definition should have been challenged by implementers.

to expedite release, stick with the majority opinion over what it should be, define it properly and @all should learn from this mistake.

gavofyork:
oh... and it should indeed be an RLP data ("string") item that is 4 bytes long for an IPv4 address.

libp2p/NodeTable.cpp
```diff
@@ -286,14 +271,14 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeId _target)
...
 	for (auto& nodes: found)
 		for (auto n: nodes.second)
-			if (n->endpoint.isAllowed())
+			if (ret.size() < s_bucketSize && !!n->endpoint && n->endpoint.isAllowed())
 				ret.push_back(n);
 	return move(ret);
 }
 
-void NodeTable::ping(bi::udp::endpoint _to) const
+void NodeTable::ping(NodeIPEndpoint _to) const
```
gavofyork: `const&` ?

gavofyork:
code looks reasonable - as long as this is 100% compatible with Go, it looksgood.

fjl:
RLP encoding of pong is missing the destination: 

```diff
@@ -330,8 +330,8 @@ struct Pong: RLPXDatagram<Pong>
 	h256 echo;				///< MCD of PingNode
 	uint32_t ts = 0;
-	void streamRLP(RLPStream& _s) const { _s.appendList(2); _s << echo << ts; }
-	void interpretRLP(bytesConstRef _bytes) { RLP r(_bytes); echo = (h256)r[0]; ts = r[1].toInt<uint32_t>(); }
+	void streamRLP(RLPStream& _s) const { _s.appendList(3); destination.streamRLP(_s); _s << echo << ts; }
+	void interpretRLP(bytesConstRef _bytes) { RLP r(_bytes); destination.interpretRLP(r); echo = (h256)r[0]; ts = r[1].toInt<uint32_t>(); }
 };

 /**
```

fjl:
With the pong encoding fixed, we are once again interoperable.

fjl:
Is the sha3 distance going to be part of this PR?

subtly:
@fjl Nice catch. How'd you create a comment like that?

fjl:
I pasted the diff and put \`\`\`diff ... ```` around it.

fjl:
I tested again, the ping/pong seems to work. eth crashed after ~10seconds. Trace here: https://gist.github.com/fjl/f08e36c5d4026f37085c

fjl:
I have tested this again. The crash is no longer happening and cpp responds to findnode if it has nodes in the table.

PR #1712, opened 2015-04-22
issues: min transaction s value is 1

winsvega:


libdevcrypto/Common.cpp
```diff
@@ -39,7 +39,8 @@ bool dev::SignatureStruct::isValid() const
 {
 	if (v > 1 ||
 		r >= h256("0xfffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141") ||
```
CJentzsch: did the value for the r value stay the same?

winsvega: yes
https://github.com/ethereum/yellowpaper/commit/cdc849968e51192ae44b8124e5800e61f4f4ff97

PR #1716, opened 2015-04-22
Gas pricer

CJentzsch:
throws exception when overflowing during setting of ether price or block fees

nice new gasPricer, old one did not work because:
* `r[1].size()` (Client.h L94) returns always 0, because of this: https://github.com/ethereum/cpp-ethereum/blob/develop/libdevcore/RLP.h#L120 - and `r[1]` is a list on not data
* ` t <= total * q / 8` (Client.h L111) is always true, because `t` was initialized as zero and never changes.

The new gasPricer uses a gaussian distribution to calculate the new pasPrice. Lowest and Highest priority is the lowest and highest value in the distribution, the other values are:
* Low - lower gasPrice then ~75% of all transaction
* Medium - mean gasPrice
* High - higher than ~75% of all transactions

closes https://github.com/ethereum/cpp-ethereum/issues/1453


eth/main.cpp
```diff
@@ -1137,13 +1137,33 @@ int main(int argc, char** argv)
 			else if (c && cmd == "setblockfees")
 			{
 				iss >> blockFees;
-				gasPricer->setRefBlockFees(u256(blockFees * 1000));
+				try
+				{
+					gasPricer->setRefBlockFees(u256(blockFees * 1000));
+				}
+				catch(Overflow const& _e)
```
chriseth: space

eth/main.cpp
```diff
@@ -1137,13 +1137,33 @@ int main(int argc, char** argv)
...
+				}
+
 				cout << "Block fees: " << blockFees << endl;
 			}
 			else if (c && cmd == "setetherprice")
 			{
 				iss >> etherPrice;
-				gasPricer->setRefPrice(u256(double(ether / 1000) / etherPrice));
+				if (etherPrice == 0)
+					cout << "ether price can not be set to zero" << endl;
```
chriseth: "cannot"?

eth/main.cpp
```diff
@@ -1137,13 +1137,33 @@ int main(int argc, char** argv)
...
+				else
+				{
+					try
+					{
+						gasPricer->setRefPrice(u256(double(ether / 1000) / etherPrice));
+					}
+					catch(Overflow const& _e)
```
chriseth: space

eth/main.cpp
```diff
@@ -1137,13 +1137,33 @@ int main(int argc, char** argv)
...
 			{
 				iss >> etherPrice;
-				gasPricer->setRefPrice(u256(double(ether / 1000) / etherPrice));
+				if (etherPrice == 0)
+					cout << "ether price can not be set to zero" << endl;
+				else
+				{
+					try
+					{
+						gasPricer->setRefPrice(u256(double(ether / 1000) / etherPrice));
```
chriseth: Direct conversion from double to u256?
And why not `(double(ether) / 1000) / etherPrice`?

libethereum/Client.cpp
```diff
@@ -91,28 +94,44 @@ void BasicGasPricer::update(BlockChain const& _bc)
...
 			RLP r(bb);
 			BlockReceipts brs(_bc.receipts(bi.hash()));
-			for (unsigned i = 0; i < r[1].size(); ++i)
+			size_t i = 0;
+			for (auto const& tr: r[1])
 			{
+				Transaction tx(tr.data(), CheckTransaction::None);
 				auto gu = brs.receipts[i].gasUsed();
-				dist[Transaction(r[1][i].data(), CheckTransaction::None).gasPrice()] += (unsigned)brs.receipts[i].gasUsed();
+				dist[tx.gasPrice()] += (unsigned)gu;
```
chriseth: Please, never ever c-style cast an auto variable to a POD! It might silently change into a pointer.
I would recommend not to use auto here at all.

libethereum/Client.cpp
```diff
@@ -91,28 +94,44 @@ void BasicGasPricer::update(BlockChain const& _bc)
...
+		u256 sdSquared = 0;
+		for (auto const& i: dist)
+			sdSquared += i.second * (i.first - mean) * (i.first - mean);
+		sdSquared /= total;
+		double sd = sqrt(sdSquared.convert_to<double>());
+		double normalizedSd = 4.0 * sd / mean.convert_to<double>();
+
+		// calc octiles normalized to gaussian distribution
+		boost::math::normal gauss(4, normalizedSd ? normalizedSd : 1);
+		for (int i=1; i < 8; i++)
```
chriseth: space

libethereum/Client.cpp
```diff
@@ -91,28 +94,44 @@ void BasicGasPricer::update(BlockChain const& _bc)
...
+			m_octiles[i] = mean / 4000 * int(boost::math::quantile(gauss, (double)i / 8) * 1000);
```
chriseth: Why not just `i / 8.0`?

libethereum/Client.h
```diff
@@ -77,8 +77,8 @@ class BasicGasPricer: public GasPricer
 public:
 	explicit BasicGasPricer(u256 _weiPerRef, u256 _refsPerBlock): m_weiPerRef(_weiPerRef), m_refsPerBlock(_refsPerBlock) {}
 
-	void setRefPrice(u256 _weiPerRef) { m_weiPerRef = _weiPerRef; }
-	void setRefBlockFees(u256 _refsPerBlock) { m_refsPerBlock = _refsPerBlock; }
+	void setRefPrice(u256 _weiPerRef) { if ((bigint)m_refsPerBlock * _weiPerRef > std::numeric_limits<u256>::max() ) BOOST_THROW_EXCEPTION(Overflow() << errinfo_comment("ether price * block fees is larger than 2**256-1, choose a smaller number.") ); else m_weiPerRef = _weiPerRef; }
+	void setRefBlockFees(u256 _refsPerBlock) { if ((bigint)m_weiPerRef * _refsPerBlock > std::numeric_limits<u256>::max() ) BOOST_THROW_EXCEPTION(Overflow() << errinfo_comment("ether price * block fees is larger than 2**256-1, choose a smaller number.") ); else m_refsPerBlock = _refsPerBlock; }
```
chriseth: space

neth/main.cpp
```diff
@@ -718,13 +718,33 @@ int main(int argc, char** argv)
 		else if (c && cmd == "setblockfees")
 		{
 			iss >> blockFees;
-			gasPricer->setRefBlockFees(u256(blockFees * 1000));
+			try
+			{
+				gasPricer->setRefBlockFees(u256(blockFees * 1000));
+			}
+			catch(Overflow const& _e)
```
chriseth: space (is probably present at some other places as well)

eth/main.cpp
CJentzsch: Thats actually Gavs code ;-), but nevertheless, I also prefer to make the cast as late as necessary for accuracy. So ether / 1000 is fine, but when divided by etherPrice it needs to be a double, therefore the cast. The cast from double to u256 is also necessary.

gavofyork:
win32 fails to compile, but i guess this is unrelated.

PR #1719, opened 2015-04-22
Windows build fixes

arkpar:


libdevcore/Log.h
```diff
@@ -26,7 +26,6 @@
 #include <ctime>
 #include <chrono>
 #include <boost/thread.hpp>
-#include <boost/asio.hpp>
```
arkpar: including all of boost::asio like this pulls an implicit link dependency on boost_regex

PR #1720, opened 2015-04-22
fixed webthreestubclient generation directory

debris:
after refactor of test directory webthreestubclient was not being created in correct directory.


arkpar:
I'd suggest all generated files should go to the build tree and not to the source tree

debris:
We can do this, but it also implies the requirement to have jsonrpcstub executable in the path.

PR #1726, opened 2015-04-23
bring back proper topics parsing

debris:
bring back proper topics parsing in WebThreeStubServerBase. I must have accidentally removed short after poc-8 was released.

[old eth topics parsing](https://github.com/ethereum/cpp-ethereum/blob/poc-8-tag/libweb3jsonrpc/WebThreeStubServerBase.cpp#L142-L154)

[old shh topics parsing](https://github.com/ethereum/cpp-ethereum/blob/poc-8-tag/libweb3jsonrpc/WebThreeStubServerBase.cpp#L180-L188)

### DIFF
--- a/libweb3jsonrpc/WebThreeStubServerBase.cpp
+++ b/libweb3jsonrpc/WebThreeStubServerBase.cpp
@@ -201,7 +201,7 @@ static dev::eth::LogFilter toLogFilter(Json::Value const& _json)	// commented to
 					if (!t.isNull())
 						filter.topic(i, jsToFixed<32>(t.asString()));
 			}
-			else if (!_json["topics"][i].isNull())
+			else if (!_json["topics"][i].isNull()) // if it is anything else then string, it should and will fail
 				filter.topic(i, jsToFixed<32>(_json["topics"][i].asString()));
 		}
 	return filter;
@@ -233,7 +233,16 @@ static shh::Envelope toSealed(Json::Value const& _json, shh::Message const& _m, 
 	
 	if (!_json["topics"].empty())
 		for (auto i: _json["topics"])
-			bt.shift(jsToBytes(i.asString()));
+		{
+			if (i.isArray())
+			{
+				for (auto j: i)
+					if (!j.isNull())
+						bt.shift(jsToBytes(j.asString()));
+			}
+			else if (!i.isNull()) // if it is anything else then string, it should and will fail
+				bt.shift(jsToBytes(i.asString()));
+		}
 	
 	return _m.seal(_from, bt, ttl, workToProve);
 }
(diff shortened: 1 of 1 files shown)


libweb3jsonrpc/WebThreeStubServerBase.cpp
```diff
@@ -194,7 +194,16 @@ static dev::eth::LogFilter toLogFilter(Json::Value const& _json)	// commented to
...
 		for (unsigned i = 0; i < _json["topics"].size(); i++)
-			filter.topic(i, jsToFixed<32>(_json["topics"][i].asString()));
+		{
+			if (_json["topics"][i].isArray())
+			{
+				for (auto t: _json["topics"][i])
+					if (!t.isNull())
+						filter.topic(i, jsToFixed<32>(t.asString()));
+			}
+			else if (!_json["topics"][i].isNull()) // if it is anything else then string, it should and will fail
```
chriseth: What does it mean to fail here?

libweb3jsonrpc/WebThreeStubServerBase.cpp
```diff
@@ -194,7 +194,16 @@ static dev::eth::LogFilter toLogFilter(Json::Value const& _json)	// commented to
 	}
 	if (!_json["topics"].empty())
 		for (unsigned i = 0; i < _json["topics"].size(); i++)
-			filter.topic(i, jsToFixed<32>(_json["topics"][i].asString()));
+		{
```
chriseth: No braces.

libweb3jsonrpc/WebThreeStubServerBase.cpp
```diff
@@ -224,7 +233,16 @@ static shh::Envelope toSealed(Json::Value const& _json, shh::Message const& _m,
 	
 	if (!_json["topics"].empty())
 		for (auto i: _json["topics"])
-			bt.shift(jsToBytes(i.asString()));
+		{
```
chriseth: No braces.

libweb3jsonrpc/WebThreeStubServerBase.cpp
debris: I added bracers to avoid dangling else. That's Xcode suggestion with which I agree. It's just more readable.

screen shot 2015-04-24 at 15 18 35

libweb3jsonrpc/WebThreeStubServerBase.cpp
debris: if one of the topics will be number or object, an exception will be thrown, and JSON-RPC will return 'INVALID_PARAMS'

libweb3jsonrpc/WebThreeStubServerBase.cpp
debris: same here, dangling else.

chriseth:
Ok, everything fine then.

Issue #1714, opened 2015-04-22
Mix invariably crashes after two new blocks.


Smithgift:
This is what it says in the terminal afterwards
```
  ℹ  08:40:48 AM|Thread (pooled)  Mined 398b9af8… (parent: 34b3c4ce… )
☍ ◌  08:40:48 AM|Thread (pooled)  Attempting import of #398b9af8… ...
  ✘  08:40:48 AM|Thread (pooled)  Unable to sync to 34b3c4ce… ; state root d0dc0072… not found in database.
  ✘  08:40:48 AM|Thread (pooled)  Database corrupt: contains block without stateRoot: 34b3c4ce8a3ed18f28250ae77d83e5288fd2cca55b690aad3a46f0a17af990a6 d1ca4df1caae11c2556bff868ee802a092de79fac5fe54d7d4b4488127044b84 1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347 38f388fadf4a6a35c61c3f88194ec5ae162c8944 d0dc0072fc2b95d669b0e63f8ac8941277386c6fbf3e6f95a2b6b0d276c55600 8ce82d51a26cea2a54f02b3d9165e9e344bc44908fb4e3f25b1a0289b6d18bdc 1fdebaaee68d11b1d87ec6589a952cfcdca5209c8d7009d0901d3a89dc402feb 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 131072 1 3141592 1068508 1429706444 0000000000000000000000000000000000000000000000000000000000000000 0000000000000000 (0000000000000000000000000000000000000000000000000000000000000000)
  ✘  08:40:48 AM|Thread (pooled)  Bailing.
Segmentation fault (core dumped)
```

PR #1727, opened 2015-04-23
Mix bugfixes

arkpar:
* Fixed issues having multiple contracts in the same source
* Fixed issues navigating between pages in web preview
* #1714
* Added tests

### DIFF
--- a/mix/CodeModel.cpp
+++ b/mix/CodeModel.cpp
@@ -279,9 +279,22 @@ void CodeModel::runCompilationJob(int _jobId)
 				QQmlEngine::setObjectOwnership(contract, QQmlEngine::CppOwnership);
 				result[name] = contract;
 				CompiledContract* prevContract = nullptr;
+				// find previous contract by name
 				for (ContractMap::const_iterator c = m_contractMap.cbegin(); c != m_contractMap.cend(); ++c)
-					if (c.value()->documentId() == contract->documentId())
+					if (c.value()->contract()->name() == contract->contract()->name())
 						prevContract = c.value();
+
+				// if not found, try by documentId
+				if (!prevContract)
+				{
+					for (ContractMap::const_iterator c = m_contractMap.cbegin(); c != m_contractMap.cend(); ++c)
+						if (c.value()->documentId() == contract->documentId())
+						{
+							//make sure there are no other contracts in the same source, otherwise it is not a rename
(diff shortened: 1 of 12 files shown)


mix/CodeModel.cpp
```diff
@@ -279,9 +279,22 @@ void CodeModel::runCompilationJob(int _jobId)
 				QQmlEngine::setObjectOwnership(contract, QQmlEngine::CppOwnership);
```
chriseth: Quite long function. Does it make sense to create a new function where the guards start their work?

mix/HttpServer.cpp
```diff
@@ -145,16 +145,23 @@ void HttpServer::readClient()
...
 			{
 				QUrl url(hdr.split(' ')[1]);
 				QString l;
 				do
+				{
 					l = socket->readLine();
+					//collect headers
+					int colon = l.indexOf(':');
+					if (colon > 0)
+						headers[l.left(colon)] = l.right(l.length() - colon - 1);
```
chriseth: Perhaps lowercase the key?

mix/qml/WebPreview.qml
```diff
@@ -184,8 +185,10 @@ Item {
 						content = fileIo.readFile(doc.path);
 				}
 
-				if (documentName === urlInput.text.replace(httpServer.url + "/", "")) {
-					//root page, inject deployment script
+				var accept = _request.headers["Accept"];
+				if (accept && accept.indexOf("text/html") >= 0 && !_request.headers["HTTP_X_REQUESTED_WITH"])
```
chriseth: That looks quite hackish. Which problem does it solve?

mix/HttpServer.cpp
arkpar: HTTP headers are case sensitive

mix/qml/WebPreview.qml
arkpar: This is do differentiate ajax requests from navigation requests. We only need to inject web3 object if the page request is coming from the browser, and not from javascript

mix/HttpServer.cpp
chriseth: To my knowledge, they are not: https://tools.ietf.org/html/rfc7230#section-3.2

arkpar: You're absolutely right

PR #1728, opened 2015-04-23
json output for the commandline compiler.

chriseth:


solc/CommandLineInterface.cpp
```diff
@@ -50,6 +51,7 @@ namespace dev
 namespace solidity
 {
 
+
```
LefterisJP: why the extra space here?

chriseth: You know, sometimes you really need a bit more space. ;-)

PR #1730, opened 2015-04-23
Direct more mundane failures to cerr as well.

ryepdx:
I was using the i/o redirection operators in bash to direct stdout from solc to files in a build directory, and discovered that if a filename is given incorrectly it just writes the "Skipping file" message to the built file. From my perspective, any message relating to the failure of the requested operation should be written to stderr. Only the results of the requested operation should be written to stdout.


chriseth:
Thanks!

PR #1731, opened 2015-04-23
issues: right r<1, s<1 comparation

winsvega:
bug fix


CJentzsch:
You can use this constructor: https://github.com/ethereum/cpp-ethereum/blob/develop/libdevcore/FixedHash.h#L65 and just write h256(1).
Try to see if that works, it should.

PR #1740, opened 2015-04-24
Try to keep source location during optimisation.

chriseth:


test/libsolidity/SolidityOptimizer.cpp
```diff
@@ -84,9 +85,20 @@ class OptimizerTestFramework: public ExecutionFramework
...
+			item.setLocation(SourceLocation(1, 3, make_shared<string>("")));
+
 		eth::CommonSubexpressionEliminator cse;
-		BOOST_REQUIRE(cse.feedItems(_input.begin(), _input.end()) == _input.end());
-		return cse.getOptimizedItems();
+		BOOST_REQUIRE(cse.feedItems(input.begin(), input.end()) == input.end());
+		AssemblyItems output = cse.getOptimizedItems();
+
+		for (AssemblyItem const& item: output)
+		{
```
LianaHus: do you need braces here?

chriseth: Yes, for compiling on windows, the macro seems to not expand to a single statement.

PR #1743, opened 2015-04-24
Move assembly related files to libevmasm and split parameters.

chriseth:


gavofyork:
Params.h/.cpp does not include merely EVM stuff; several of the parameters are not EVM related, and it doesn't make sense to place them in an "evmcore" library.

chriseth:
The problem is that in order to use the parameters, you have to pull in many unnecessary dependencies and even have to link against ethash, which is certainly not an option.

PR #1749, opened 2015-04-25
Fix missing one hash in reply as result of double parent reference

romanman:


gavofyork:
how clever, you're attempting to return the same hash that you sent in the first place.

romanman:
@gavofyork : actually you are getting the real hash next line 
inside the loop : `for (unsigned i = 0; i < c && p; ++i, p = host()->m_chain.details(p).parent)`

so if you will ask the parent like now
you will miss the first one.

PR #1751, opened 2015-04-26
added -Wconversion to compiler options, fixes #1023

### LINKED ISSUE #1023: Add -Wconversion warnings
What do you think about adding [-Wconversion](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) warnings to compiler options? It would improve compatibility with Visual Studio and (my opinion) make the code a bit cleaner.

debris:
@chfast I also think that it is a good idea to add this flag.

### DIFF
--- a/cmake/EthCompilerSettings.cmake
+++ b/cmake/EthCompilerSettings.cmake
@@ -3,7 +3,7 @@
 # C++11 check and activation
 if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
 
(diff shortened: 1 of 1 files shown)


debris:
forgot about Werror :p But still, I think that it would be good to introduce this.

chfast:
Looks like some work needs to be done on linux. In the same time OSX seems to have at least some part of it enabled by default.

chfast:
Is it this change accepted?

Issue #1750, opened 2015-04-26
new icap code -> core dumped


youlian:
on startup alethzero explodes merrily with the following:
terminate called after throwing an instance of 'dev::eth::InvalidICAP'
  what():  std::exception
Aborted (core dumped)

PR #1757, opened 2015-04-27
Safe ICAP decoding

arkpar:
#1750

### DIFF
--- a/libethcore/ICAP.cpp
+++ b/libethcore/ICAP.cpp
@@ -62,6 +62,18 @@ std::pair<string, string> ICAP::fromIBAN(std::string _iban)
 	return make_pair(c, d);
 }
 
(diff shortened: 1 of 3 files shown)


libethcore/ICAP.cpp
```diff
@@ -62,6 +62,18 @@ std::pair<string, string> ICAP::fromIBAN(std::string _iban)
 	return make_pair(c, d);
 }
 
+ICAP ICAP::safeDecoded(std::string const& _encoded)
+{
+	try
+	{
+		return decoded(_encoded);
+	}
+	catch (InvalidICAP)
```
chriseth: Catch by const reference?

gavofyork:
all fixed in latest.

PR #1762, opened 2015-04-28
Fix missing one hash in reply as result of double parent reference

romanman:
@gavofyork : actually you are getting the real hash next line 
inside the loop : `for (unsigned i = 0; i < c && p; ++i, p = host()->m_chain.details(p).parent)`

so if you will ask the parent like now
you will miss the first one cause of `parent.parent`.


gavofyork:
invalid, as i mentioned in the issue. we do not return the hash from which they asked. it's redundant and a waste of bandwidth.

PR #1767, opened 2015-04-28
State export/import

arkpar:
Export dialog in AZ:
export


gavofyork:
how long does it take to populate the state dropdown?

arkpar:
Only last 10 blocks are populated and then you can type to set any other block. Accounts and contracts are populated once a block is selected - less than a second total on current test network

PR #1777, opened 2015-05-01
fixed excessive RAM usage when GPU mining

Genoil:
By using the CL_MEM_USE_HOST_PTR flag on Buffer creation, only as much RAM as the dag is sized is required, instead of (1 + num GPUs) * dag_size, as it was before.

To verify the dag is really on the device, I wrote a simple kernel that outputs the first 256 bytes of the dag file. It seems to work fine.

### DIFF
--- a/libethash-cl/ethash_cl_miner.cpp
+++ b/libethash-cl/ethash_cl_miner.cpp
@@ -117,7 +117,7 @@ void ethash_cl_miner::finish()
 	}
 }
 
-bool ethash_cl_miner::init(ethash_params const& params, std::function<void(void*)> _fillDAG, unsigned workgroup_size, unsigned _platformId, unsigned _deviceId)
+bool ethash_cl_miner::init(ethash_params const& params, const void * dag_ptr, unsigned workgroup_size, unsigned _platformId, unsigned _deviceId)
 {
 	// store params
 	m_params = params;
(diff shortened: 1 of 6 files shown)


LefterisJP:
The PR is still not ready I suppose, right? You have a lot of commented out code in there. The extra kernel is just for visual verification by yourself and as such is not really intended to be commited, correct?

The main question is would the `CL_MEM_USE_HOST_PTR` option work fine with a memory mapped DAG? I take it that this is what gavin's comment in there originally meant.

Also try to use the C++ coding style as can be seen here: https://github.com/ethereum/cpp-ethereum/blob/develop/CodingStandards.txt.

Genoil:
Well I left all that stuff in because I did expect some discussion. I really don't see the point of memory mapping if all it does is eat up loads of RAM for multi-GPU systems.

That said, somebody else tried to run this fork and it borked out so it does need some more love.

gavofyork:
@programmerTim  comments?

gavofyork:
please check latest cpp develop - this should be fixed (i used enqueueWriteBuffer rather than hostptr stuff, but should have the same effect).

Genoil:
haven't got multi-GPU set up atm, but it works fine with one. host ptr would indeed only be useful if the dag would be modified by the GPU, but as this isn't the case, this looks more straightforward.

PR #1781, opened 2015-05-02
sha3 xor metric

subtly:


chriseth:
Wouldn't it be better to directly use the sha3 of the address as NodeId, or at least store it inside Nodeid?

subtly:
@chriseth Perhaps as an optimization. The output of the distance is already being stored.

libp2p/NodeTable.h
```diff
@@ -138,7 +138,7 @@ class NodeTable: UDPSocketEvents, public std::enable_shared_from_this<NodeTable>
 	~NodeTable();
 
 	/// Returns distance based on xor metric two node ids. Used by NodeEntry and NodeTable.
-	static unsigned distance(NodeId const& _a, NodeId const& _b) { u512 d = _a ^ _b; unsigned ret; for (ret = 0; d >>= 1; ++ret) {}; return ret; }
+	static unsigned distance(NodeId const& _a, NodeId const& _b) { u256 d = sha3(_a) ^ sha3(_b); unsigned ret; for (ret = 0; d >>= 1; ++ret) {}; return ret; }
```
gavofyork: is this called often? `sha3` is not an especially quick function to be calling...

fjl:
For cross-reference, the go-ethereum PR is here: ethereum/go-ethereum#791.
The Go PR has a protocol change which makes findnode target hash-sized instead of public key sized.
Will submit a devp2p PR for that.

fjl:
https://github.com/ethereum/devp2p/pull/25

libp2p/NodeTable.h
subtly: Not often. It's called by the constructor of NodeEntry and the result is stored in a property.

PR #1782, opened 2015-05-02
Limit DNS resolution to ipv4.

subtly:
On some systems w/ipv6 enabled the resolver returns ipv6 address even though the system doesn't have a routable ipv6 address.


chriseth:
What happens on ipv6-only systems?

subtly:
That's rare. I'm not sure -- it's a TODO. As of right now the code doesn't work with IPv6 and without this code IPv4 users with IPv6 ISPs won't be able to connect to the network.

PR #1785, opened 2015-05-02
adding --network-id option to eth cli

psalami:
This CLI option allows setting up a private testnet using remote nodes that have public IPs (i.e. when using a public cloud provider). It also allows you to set up a private testnet with both eth and geth nodes (this is a port of the --networkid CLI option in geth). 
Without this option, nothing stops your private testnet nodes from connecting to other networks (i.e. poc-9).

Usage:
Start eth with the ``--network-id`` option and enter a random number greater than 0. Use the same network id for all the nodes that you are starting in your testnet. Nodes started with this option will only connect to other nodes that have the same network id. You will be able to maintain your own private chain, even if your nodes have public IP addresses.

### DIFF
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -146,6 +146,7 @@ void help()
 		<< "    --listen <port>  Listen on the given port for incoming connections (default: 30303)." << endl
 		<< "    -r,--remote <host>(:<port>)  Connect to remote host (default: none)." << endl
 		<< "    --port <port>  Connect to remote port (default: 30303)." << endl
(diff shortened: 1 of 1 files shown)


subtly:
@psalami This functionality isn't yet supported or stable. It requires a bit more work so that the network isn't flooded with incompatible nodes. This is coming soon (before frontier).

gavofyork:
merged since this feature is already in use in alethzero. it is indeed an issue for the peer network, but a simple workaround is to require all nodes support all subprotocols; a different networkid leads to the eth subprotocol dropping, and could trivially be changed to just dropping the peer (see Capability.cpp:35).

PR #1786, opened 2015-05-02
add constuctor name to json-abi

kobigurk:
This ensures that those who rely on methods having a name (like pytester) will not fail.


chriseth:
I think that the constructor part of the ABI should not have a name property. It is not a function that is called by name (actually it is not even a function from the ABI perspective). Solidity inherited C++'s convention to use a member function of the class' name as constructor, but note that e.g. in Python, those functions are called "__init__", so this could confuse serpent users of the ABI.
Is this fixable in pytester?

kobigurk:
Well, good points  - I also thought it was the wrong way to fix it after I opened it. it's redundant information. I think it should indeed be fixed in python.

PR #1799, opened 2015-05-04
Feature for commandline compiler to output the function signature hashes.

chriseth:


solc/CommandLineInterface.cpp
```diff
@@ -55,6 +55,7 @@ namespace solidity
 // extensions and other attributes would be a better choice here?
 static string const g_argAbiStr         = "json-abi";
 static string const g_argSolAbiStr      = "sol-abi";
+static string const g_argSignatureHashes = "hashes";
```
LianaHus: I think we should either keep them aligned or remove spaces everywere

PR #1800, opened 2015-05-04
Blockchain download and import performance optimizations

arkpar:


libethcore/EthashAux.cpp
```diff
@@ -87,13 +87,10 @@ ethash_params EthashAux::params(h256 const& _seedHash)
 {
 	Guard l(get()->x_epochs);
 	unsigned epoch = 0;
-	try
+	auto epochIter = get()->m_epochs.find(_seedHash);
+	if (epochIter == get()->m_epochs.end())
 	{
-		epoch = get()->m_epochs.at(_seedHash);
```
arkpar: lots of exceptions thrown here

libethereum/BlockChain.cpp
```diff
@@ -609,7 +614,6 @@ ImportRoute BlockChain::import(bytes const& _block, OverlayDB const& _db, Import
 		}
 
 		clog(BlockChainNote) << "   Imported and best" << td << " (#" << bi.number << "). Has" << (details(bi.parentHash).children.size() - 1) << "siblings. Route:" << route;
-		noteCanonChanged();
```
arkpar: Please check if this can be removed safely

gavofyork: it's critical.

libethcore/EthashAux.cpp
gavofyork: odd - there should only be at most the number of epochs in the current chain i.e. < 10.

libethereum/BlockChain.h
```diff
@@ -40,6 +42,11 @@
 #include "BlockQueue.h"
 namespace ldb = leveldb;
 
+template <> struct std::hash<std::pair<dev::h256, unsigned>>
+{
+	size_t operator()(const pair<dev::h256, unsigned> &x ) const { return std::hash<dev::h256>()(x.first) ^ std::hash<unsigned>()(x.second); }
```
gavofyork: bad coding style.

gavofyork:
merged.

PR #1805, opened 2015-05-05
Wallet tests

CJentzsch:
100% code coverage for https://github.com/ethereum/dapp-bin/blob/master/wallet/wallet2.sol


test/libethereum/StateTestsFiller/stCallCreateCallCodeTestFiller.json
```diff
@@ -618,6 +618,96 @@
         }
     },
 
+    "createNameRegistratorPreStore1NotEnoughGas": {
+        "env" : {
+            "previousHash" : "5e20a0453cecd065ea59c37ac63e079ee08998b6045136a8ce6635c7912ec0b6",
+            "currentNumber" : "0",
```
chfast: What about ongoing hexification? Shouldn't all numbers be in hex base?

CJentzsch: this is only the filler, here we can use what we want :-), when creating the real tests, everything his hexified.

chfast:
Hard to check all the test details. We should give it a try.

PR #1806, opened 2015-05-05
Specifing concrete test when filling testsuite

winsvega:


test/TestHelper.cpp
```diff
@@ -741,6 +741,12 @@ Options::Options()
 			inputLimits = true;
 			bigData = true;
 		}
+		else if (arg.compare(0, 12, "--singletest") == 0)
```
CJentzsch: dont use compare here. just do arg == "singletest" && i + 1 < argc; singleTestName = argv[i + 1];

test/libethereum/state.cpp
```diff
@@ -43,6 +43,10 @@ void doStateTests(json_spirit::mValue& v, bool _fillin)
 {
 	for (auto& i: v.get_obj())
 	{
+		if (test::Options::get().singletest == true)
+			if (test::Options::get().singletestName != i.first)
```
CJentzsch: use && instead of two ifs

test/TestHelper.cpp
winsvega: I just took a line 718::      else if (arg.compare(0, 7, "--stats") == 0)
as an example

PR #1808, opened 2015-05-05
Smart VM

chfast:
The first step of SmartVM implementation. It is more like proof of concept than production ready feature. Currently SmartVM selects JIT when EVM code execution has be requested before (during current session).

### DIFF
--- a/evmjit/libevmjit/Common.h
+++ b/evmjit/libevmjit/Common.h
@@ -31,7 +31,7 @@ enum class ReturnCode
 
 	// Standard error codes
 	OutOfGas           = -1,
-	StackUnderflow      = -2,
+	StackUnderflow     = -2,
 	BadJumpDestination = -3,
 	BadInstruction     = -4,
 	Rejected           = -5, ///< Input data (code, gas, block info, etc.) does not meet JIT requirement and execution request has been rejected
@@ -45,16 +45,6 @@ enum class ReturnCode
 
 	LinkerWorkaround = -299,
 };
-
-/// Representation of 256-bit value binary compatible with LLVM i256
-struct i256
-{
-	uint64_t a = 0;
-	uint64_t b = 0;
-	uint64_t c = 0;
-	uint64_t d = 0;
-};
-static_assert(sizeof(i256) == 32, "Wrong i265 size");
 
 #define UNTESTED assert(false)
 
--- a/libevm/SmartVM.cpp
+++ b/libevm/SmartVM.cpp
@@ -1,0 +1,79 @@
+/*
+	This file is part of cpp-ethereum.
+
+	cpp-ethereum is free software: you can redistribute it and/or modify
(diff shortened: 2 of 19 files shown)


evmjit/libevmjit/Common.h
```diff
@@ -31,7 +31,7 @@ enum class ReturnCode
 
 	// Standard error codes
 	OutOfGas           = -1,
-	StackUnderflow      = -2,
+	StackUnderflow     = -2,
 	BadJumpDestination = -3,
 	BadInstruction     = -4,
 	Rejected           = -5, ///< Input data (code, gas, block info, etc.) does not meet JIT requirement and execution request has been rejected
```
CJentzsch: how do you handle stack depth and stack size overflow (both limited to 1024)?

libevm/VMFactory.h
```diff
@@ -23,10 +23,11 @@ namespace dev
 namespace eth
 {
 
-enum class VMKind: bool
+enum class VMKind
 {
 	Interpreter,
-	JIT
+	JIT,
+	Smart,
```
CJentzsch: remove last `,`

libevm/SmartVM.cpp
```diff
@@ -0,0 +1,79 @@
...
+		cnote << "Jitted";
+		vmKind = VMKind::JIT;
+	}
+	else
+	{
+		// Check EVM code hit count
+		static const uint64_t c_hitTreshold = 1;
+		auto& hits = getHitMap()[codeHash];
+		++hits;
+		if (hits > c_hitTreshold)
```
CJentzsch: can you explain to me how that works? So whats is the criteria here for selecting jit instead of interpreter?

CJentzsch:
Its more a got question than a got issues :-)

evmjit/libevmjit/Common.h
chfast: These return codes are a bit outdated. It is ongoing process but in the end JIT will not inform what kind of error happened (only OutOfGas generic abort).
So, answering your question, in case of stack size overflow and call depth you will get OutOfGas error.

libevm/VMFactory.h
chfast: If you have comma in the end and want to add another element later you will not have to modify existing line. Diff looks better. But if you like it more without a comma :+1:

libevm/SmartVM.cpp
chfast: In general: if you see an EVM code for the first time it goes to interpreter. If you see it for the second time it goes to JIT, and JIT will compile it to machine code and load that machine code into memory. If you see the EVM code for the third time it goes to JIT as well, but then the machine code is in the memory already and the execution should be super-fast.

So currently you can control only "hit count" param. And `JIT::isCodeReady` is redundant (it checks if machine code is already in memory). But that's only an example.

Issue #1764, opened 2015-04-28
windows installer


debris:
improve windows installer. right now it installs alethzero, third and mix. It should install:

- [x] AZ
- [x] mix
- [x] solc
- [x] eth
- [x] ethminer
- [x] rlp
- [x] abi

PR #1809, opened 2015-05-05
windows package fixes

debris:
part of the #1764 
copy alethzero && mix dlls while packaging

mix executable is working fine, but i have problems with alethzero

@arkpar can you look at this? Alethzero cannot QtWebEngineProcess (it works fine for mix).

screen shot 2015-05-06 at 01 22 10


arkpar:
QtWebEngineProcess  message is just a warning, should work fine. Did you build with FATDB=0 ? There is a segfault in this case, see #1812

arkpar:
Here is an issue: 64-bit binaries should install to "C:\Program Files" and not "C:\Program Files (x86)"
See http://www.cmake.org/pipermail/cmake/2012-January/048365.html
Also, it would be nice to capitalize "ethereum" in the installer UI and directory names

debris:
>QtWebEngineProcess message is just a warning, should work fine. Did you build with FATDB=0 ? There is a segfault in this case, see #1812

Yes, it fixed the issue

debris:
>Here is an issue: 64-bit binaries should install to "C:\Program Files" and not "C:\Program Files (x86)"
See http://www.cmake.org/pipermail/cmake/2012-January/048365.html
Also, it would be nice to capitalize "ethereum" in the installer UI and directory names

It should be fine now ;)

PR #1813, opened 2015-05-06
Static Analysis Engine.

chriseth:
This enhances and refactors the existing static analysis engine so that it can be used to retain knowledge about the state across jumps.


libevmasm/KnownState.h
```diff
@@ -0,0 +1,163 @@
...
+ * derived from the operation and the equivalence class of its arguments. DUPi, SWAPi and some
+ * arithmetic instructions are used to infer equivalences while these classes are determined.
+ */
+class KnownState
+{
+public:
+	using Id = ExpressionClasses::Id;
+	struct StoreOperation
+	{
+		enum Target { Invalid, Memory, Storage };
```
LianaHus: why this is not enum class?

libevmasm/KnownState.cpp
```diff
@@ -0,0 +1,324 @@
...
+				assertThrow(info.ret <= 1, InvalidDeposit, "");
+				if (info.ret == 1)
+					setStackElement(
+						m_stackHeight + _item.deposit(),
+						m_expressionClasses->find(_item, arguments, _copyItem)
+					);
+			}
+		}
+		for (int p = m_stackHeight; p > m_stackHeight + _item.deposit(); --p)
+			m_stackElements.erase(p);
```
LianaHus: you can use erase by range here

libevmasm/KnownState.h
chriseth: It only spoils the namespace of the struct, so I would say it is ok.

libevmasm/KnownState.cpp
chriseth: Oh yes, erase by range is a very good idea!

PR #1817, opened 2015-05-06
Add nodes directly to node table when creating network.

subtly:


libp2p/Host.cpp
```diff
@@ -578,7 +578,13 @@ void Host::run(boost::system::error_code const&)
 	// is always live and to ensure reputation and fallback timers are properly
 	// updated. // disconnectLatePeers();
 
-	int openSlots = m_idealPeerCount - peerCount();
+	// todo: update peerSlotsAvailable()
+	unsigned pendingCount = 0;
+	{
+		Guard l(x_pendingNodeConns);
```
gavofyork:
```
ETH_GUARDED(x_pendingNodeConns)
    pendingCount = m_pendingPeerConns.size();
```

and you don't need an additional block.

PR #1819, opened 2015-05-07
Fix macos builds following changes from #1809.

subtly:


chfast:
It might not be directly related, but I would advice to set a global `CMAKE_RUNTIME_OUTPUT_DIRECTORY` to "bin". CMake will move there all executables (including DLLs on Windows). It will also allow to build shared libraries in Windows properly. Such settings is common practice in big projects.

gavofyork:
@chfast patch welcome :)

debris:
instead of writing `if (APPLE) ..` multiple times, can we add additional option (eg. `BUNDLE`) to `eth_install_executable` [macro](https://github.com/ethereum/cpp-ethereum/blob/develop/cmake/EthExecutableHelper.cmake#L76)?

PR #1824, opened 2015-05-07
Interface function for SourceLocation/position convertion

LianaHus:
(should be used for logs printing in mix)


chriseth:
We should do the same with `location.end`.
And please add something like "line and columns are numbered starting from 1" to the documentation.

LianaHus:
@chriseth  should I return tuple?

chriseth:
tuple sounds good.

PR #1827, opened 2015-05-08
Fix mutex when evicting node from node table.

subtly:
The code where a deadlock may occur is just after Line 443 where nodeEntry() requires a lock on x_nodes.

```
	Guard le(x_evictions);
	bool evictionEntry = false;
	for (auto it = m_evictions.begin(); it != m_evictions.end(); it++)
		if (it->first.first == nodeid && it->first.second > std::chrono::steady_clock::now())
		{
			evictionEntry = true;
			if (auto n = nodeEntry(it->second))
				dropNode(n);
```


gavofyork:
Do you really need two mutexes? If so, do you really need to have them locked simultaneously?

PR #1828, opened 2015-05-08
Mix - basic transaction creation

yann300:
Allow to create basic tr and not only contract call.


mix/ClientModel.cpp
```diff
@@ -255,7 +255,10 @@ void ClientModel::setupState(QVariantMap _state)
 		u256 value = (qvariant_cast<QEther*>(transaction.value("value")))->toU256Wei();
 		u256 gasPrice = (qvariant_cast<QEther*>(transaction.value("gasPrice")))->toU256Wei();
 		QString sender = transaction.value("sender").toString();
-		bool isStdContract = (transaction.value("stdContract").toBool());
+		bool isStdContract = transaction.value("stdContract").toBool();
+		bool isContractCreation;
```
arkpar: Can be left unitialized

mix/ClientModel.cpp
```diff
@@ -307,6 +308,14 @@ void ClientModel::executeSequence(vector<TransactionSettings> const& _sequence,
 			m_gasCosts.clear();
 			for (TransactionSettings const& transaction: _sequence)
 			{
+				QString contractName = resolveContractName(transaction.contractId);
+				QString address = resolveToken(transaction.contractId, deployedContracts);
+				if (transaction.functionId == "(transfert)")
```
arkpar: Can we make it without a magic string constant? Using empty string or explicit flag

mix/qml/TransactionDialog.qml
```diff
@@ -96,9 +114,7 @@ Dialog {
 				functionsModel.append({ text: functions[f].name });
 			}
 		}
-		//append constructor
-		functionsModel.append({ text: contractId });
-
+		functionsModel.append({ text: "(transfert)" });
```
arkpar: "None" would be better

yann300: done

PR #1829, opened 2015-05-08
testeth: fix --singletest option

chfast:


winsvega:
no the intention was to define singletest name within the testsuite
so when filling or testing it was possible to run exact one test from the suite

the filename option is needed when running userDefinedFile case

chfast:
That, somehow, makes sense. But you have used the same option for different purposes. "userDifinedFile" case does not work because during execution tests are filtered again with condition: testName == userFile.json.

winsvega:
then we have to rename on of the options

chfast:
I added some hocus-pocus in the parser to support both cases for --singletest option.

winsvega:
"I added some hocus-pocus in the parser to support both cases for --singletest option."
Wizzard! )

Compilation error?
[ 76%] Building CXX object test/fuzzTesting/CMakeFiles/checkRandomStateTest.dir/__/TestHelper.cpp.o
command timed out: 1200 seconds without output running make -j $(cat /proc/cpuinfo | grep processor | wc -l), attempting to kill

PR #1839, opened 2015-05-09
Mix: fixed transaction signatures

arkpar:


mix/MixClient.h
```diff
@@ -87,9 +87,9 @@ class MixClient: public dev::eth::ClientBase
 	virtual void prepareForTransaction() override {}
 
 private:
-	void executeTransaction(dev::eth::Transaction const& _t, eth::State& _state, bool _call, bool _gasAuto);
+	void executeTransaction(dev::eth::Transaction const& _t, dev::Secret const& _secret, eth::State& _state, bool _call, bool _gasAuto);
```
subtly: If secret is optional, can it be the last argument (and optional)?

PR #1851, opened 2015-05-10
fix and add guards

subtly:


subtly:
There are two critical fixes here. Both issues below were reproducible and are resolved by this PR.

First, new code was added in the discovery update which resulted in concurrent mutation of m_nodes without a lock. This issue was previously present but didn't surface because the node-table typically functioned in a serial manner. 

The second issue is due to ICMP/UDP error messages which were causing the socket read loop to disconnect based on condition of the error -- unlike TCP though, UDP sockets can receive an error even though they are in a valid state. I suspect that new clients being added to the network have a network stack which respond differently and these packets/errors were previously rare.

PR #1854, opened 2015-05-11
optional wallet blockChain tests, because it takes a while

CJentzsch:


test/TestHelper.cpp
```diff
@@ -732,6 +732,8 @@ Options::Options()
 			bigData = true;
 		else if (arg == "--checkstate")
 			checkState = true;
+		else if (arg == "--wallet")
```
winsvega: --wallettest might be better?

CJentzsch: this would be out of line with the other names (performance not performancetest, quadratic not quadratictest, memory..., inpulimits .., bigdata ...., ...)
I think the the name is fine, since we are executing testeth,  it is clear it is about a test.

PR #1855, opened 2015-05-11
CMake: set default RUNTIME_OUTPUT_DIRECTORY property to "bin"

chfast:
This commit changes output directory for runtime components (executables and DLLs) to "bin" directory. That allows running executables on Windows without need of install step.


test/CMakeLists.txt
```diff
@@ -95,6 +95,12 @@ if (JSONRPC)
 	target_link_libraries(testeth ${JSON_RPC_CPP_CLIENT_LIBRARIES})
 endif()
 
+if (UNIX) # Create symlink to old testeth location to make bildbot happy
```
debris: :+1:

chfast:
Let wait for test builds...

debris:
sure :)

chfast:
As it was reverted... should we continue working on that?
I used a symlink test/testeth -> bin/testeth to keep buildbot testing working. We can try the same for other case. But I'm not sure it is worth our time.

Issue #1801, opened 2015-05-04
Mix: JS API cannot comprehend solidity enums in function returns


Smithgift:
When a solidity function returns an enum, the JS API doesn't work, probably because that enum does not exist on the JS side.

Here's a representative selection from a terminal, calling an enum-returning function right out of the JS console in mix:
```
qml: {"jsonrpc":"2.0","method":"eth_sendTransaction","params":[{"to":"0x5b864aadd9ae77a178eef9728a24d28937dd5b77","data":"0x2399131c"}],"id":2}
/build/buildd/cpp-ethereum-0.9.13+780SNAPSHOT20150430134852trusty/mix/ContractCallDataEncoder.cpp(215): Throw in function QVariant dev::mix::ContractCallDataEncoder::decode(const dev::mix::SolidityType&, const bytes&)
Dynamic exception type: boost::exception_detail::clone_impl<dev::Exception>
std::exception::what: std::exception
[dev::tag_comment*] = Parameter declaration not found
qml: 
qml: http://localhost:47334/WebContainer.html:0: Uncaught SyntaxError: Unexpected end of input
```

debris:
That's true, they are not supported in JS api. @chriseth can you guide me how enums should be encoded/decoded? It's not mentioned in the [docs](https://github.com/ethereum/wiki/wiki/Ethereum-Contract-ABI)

chriseth:
enums are not visible "to the outside". If you look at the json abi, you will see that they are automatically converted to suitable uint types.
This is the theory ;-)
Looks like a bug in the interface between mix and solidity. @arkpar, could you please take a look?

Smithgift:
@chriseth Looking at the json abi straight from solc, a return type of an enum foo is actually `"enum foo"`

chriseth:
This seems like a bug in `InterfaceHandler::getABIInterface`, it does not correctly convert the parameters to their external types.

chriseth:
@Smithgift: Should be fixed on the solidity side. Would you be so kind and check whether it is fixed for you now all the way through the layers?

Smithgift:
@chriseth: Mix is currently broken for me (c.f. #1874), but when it's working again I'll be sure to check again. Thank you all.

Smithgift:
Update: Now that mix is a little more functional, it turns out that mix's end **is still** bugged. Trying to get a enum in a function return gives:

```
[...]

PR #1857, opened 2015-05-11
bug in abi.  wrong external type for return parameters

LianaHus:
bug #1801


test/libsolidity/SolidityABIJSON.cpp
```diff
@@ -499,7 +499,8 @@ BOOST_AUTO_TEST_CASE(constructor_abi)
 {
 	char const* sourceCode = R"(
 		contract test {
-			function test(uint param1, test param2, bool param3) {}
+			enum ActionChoices { GoLeft, GoRight, GoStraight, Sit }
```
chfast: Was this unit test wrong? If not, I would recommend to keep it, and another one (e.g. "constructor_abi2").

PR #1859, opened 2015-05-11
Mix - Remove contract source

yann300:
- Remove contract source.
 - Bug fix when when removing a project file.


mix/qml/FilesSection.qml
```diff
@@ -268,6 +273,17 @@ Rectangle
 							}
 						}
 					}
+
+					Menu {
+						id: contextMenuContract
+						MenuItem {
+							text: qsTr("Delete")
+							onTriggered: {
+								projectModel.removeDocument(documentId);
```
arkpar: Pleas add a confirmation dialog

yann300: done

PR #1864, opened 2015-05-11
Transfer knowledge about the state across jumps.

chriseth:


libevmasm/ControlFlowGraph.cpp
```diff
@@ -281,6 +283,15 @@ void ControlFlowGraph::gatherKnowledge()
 		)
 			workQueue.push_back(make_pair(block.next, state->copy()));
 	}
+
+	// Remove all blocks we never visited here. This might happen because a tag is pushed but
+	// never used for a JUMP.
+	// Note that this invalidates some contents of pushedTags
+	for (auto it = m_blocks.begin(); it != m_blocks.end();)
+		if (!it->second.startState)
+			m_blocks.erase(it++);
```
LianaHus: it = m_blocks.erase(it); looks better for me anyway :). I've checked from c++11 there is already a return iterator for map::erase

gavofyork:
linux build fails.

chriseth:
Fixed.

PR #1870, opened 2015-05-12
Guard m_node.endpoint access and updates.

subtly:


libp2p/Common.h
```diff
@@ -172,6 +172,7 @@ struct NodeIPEndpoint
 	/// Setting true causes isAllowed to return true for all addresses. (Used by test fixtures)
 	static bool test_allowLocal;
 
+	NodeIPEndpoint(): address(bi::address()), udpPort(0), tcpPort(0) {}
 	NodeIPEndpoint(bi::address _addr, uint16_t _udp, uint16_t _tcp): address(_addr), udpPort(_udp), tcpPort(_tcp) {}
 	NodeIPEndpoint(RLP const& _r) { interpretRLP(_r); }
```
chfast: This constructor is implicit.Is that correct?

libp2p/Common.h
```diff
@@ -172,6 +172,7 @@ struct NodeIPEndpoint
 	/// Setting true causes isAllowed to return true for all addresses. (Used by test fixtures)
 	static bool test_allowLocal;
 
+	NodeIPEndpoint(): address(bi::address()), udpPort(0), tcpPort(0) {}
```
chfast: My advice would be to use non-static member initialization feature: `uint16_t udpPort = 0` instead of initializing in to 0 in many constructors. Same can be done for `address`.

libp2p/NodeTable.cpp
```diff
@@ -79,7 +79,7 @@ shared_ptr<NodeEntry> NodeTable::addNode(Node const& _node, NodeRelation _relati
 {
 	if (_relation == Known)
 	{
-		shared_ptr<NodeEntry> ret(new NodeEntry(m_node, _node.id, _node.endpoint));
+		shared_ptr<NodeEntry> ret(new NodeEntry(m_node.id, _node.id, _node.endpoint));
```
chfast: It is always better to use `std::make_shared` instead of `std::shared_ptr(new ...)`.

subtly: That can be changed later. This is a critical fix and I won't be changing anything that doesn't have to be changed.

Why is make_shared better?

libp2p/Common.h
subtly: Nope (on macos/clang).

libp2p/Common.h
subtly: It's not initialized to zero in multiple constructors.

libp2p/NodeTable.cpp
gavofyork: avoid duplication of type in quick succession.

PR #1816, opened 2015-05-06
Ethash reloaded

LefterisJP:
Integrating the new ethash API into cpp-ethereum.

Using `git subtree` with `split` and `merge` in order to pull the two different directories of ethash that we use in cpp-ethereum as subtrees and keep history and contributions intact. The method is described [here](http://stackoverflow.com/questions/22334382/git-subtree-only-one-file-or-directory/22335955)

This is a work in progress. DO NOT merge yet.

PR #1756, opened 2015-04-27
install ethash if ethash-cl not built

subtly:
Ideally ethash and ethash-cl would be joined into one library with CL files being conditionally added.

PR #1871, opened 2015-05-12
Recover lost code after ethash merge

LefterisJP:
- Recover installing ethash if ethash-cl is not built


libethash-cl/CMakeLists.txt
```diff
@@ -17,3 +17,6 @@ TARGET_LINK_LIBRARIES(${EXECUTABLE} ${OpenCL_LIBRARIES} ethash)
 install( TARGETS ${EXECUTABLE} RUNTIME DESTINATION bin ARCHIVE DESTINATION lib LIBRARY DESTINATION lib )
 install( FILES ${HEADERS} DESTINATION include/${EXECUTABLE} )
 
+if (NOT ETHASHCL)
+	install( TARGETS ${LIBRARY} RUNTIME DESTINATION bin ARCHIVE DESTINATION lib LIBRARY DESTINATION lib )
```
debris: this will never be executed, see [this](https://github.com/LefterisJP/cpp-ethereum/blob/install_ethash_merge_fix/CMakeLists.txt#L350)

LefterisJP: That's interesting ...

This was in cpp-ethereum until now and was committed by: #1756.
Due to it being made in cpp-ethereum and never mirrored in the actual ethash repository it got lost during the subtreee merging from #1816 

I just noticed it now and wanted to put it back. What would the proper way to do it be then? And are you sure that it's not needed? I remember this solving some build problems before this got committed.

debris: look carefully ;) it was added to `ethash` not `ethash-cl` ;)

LefterisJP: aaaargh .... damn ... you are right .. wrong CMake file

PR #1880, opened 2015-05-13
Improve getpass() usage

LefterisJP:
As can be seen from: http://unixhelp.ed.ac.uk/CGI/man-cgi?getpass+3
getpass() is a deprecated function and it's advised not to use
it.

What's more all signals are disabled so (SIGINT, SIGQUIT, SIGSTOP,
SIGTSTOP) will do nothing and as such if the user wants to quit the
program while typing the password in the password loop he is out of
luck.

With this commit we get a manual version of getpass(), inspired by this
SO answer: http://stackoverflow.com/a/1196696/110395

### DIFF
--- a/libdevcore/CommonIO.cpp
+++ b/libdevcore/CommonIO.cpp
@@ -24,6 +24,9 @@
 #include <cstdlib>
 #include <fstream>
 #include "Exceptions.h"
+#ifndef _WIN32
+#include <termios.h>
+#endif
 using namespace std;
 using namespace dev;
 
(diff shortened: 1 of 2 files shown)


libdevcore/CommonIO.cpp
```diff
@@ -126,6 +129,29 @@ std::string dev::getPassword(std::string const& _prompt)
 	std::getline(cin, ret);
 	return ret;
 #else
-	return getpass(_prompt.c_str());
+	struct termios oflags;
+	struct termios nflags;
+	char password[128];
```
subtly: Have a few friends who use >128chars for their ssh passwords. Can we bump this to 256?

LefterisJP: Yeah ofc. The reason for choosing 128 is that this is the internal buffer limit that `getpass()` actually used if you check out the documentation I linked above.

libdevcore/Exceptions.h
```diff
@@ -54,6 +54,7 @@ struct FileError: virtual Exception {};
 struct Overflow: virtual Exception {};
 struct InterfaceNotSupported: virtual Exception { public: InterfaceNotSupported(std::string _f): Exception("Interface " + _f + " not supported.") {} };
 struct FailedInvariant: virtual Exception {};
+struct ExternalFunctionFailure: virtual Exception { public: ExternalFunctionFailure(std::string _f): Exception("Function " + _f + "() failed.") {} };
```
subtly: super

libdevcore/CommonIO.cpp
```diff
@@ -126,6 +129,29 @@ std::string dev::getPassword(std::string const& _prompt)
...
+	if (tcsetattr(fileno(stdin), TCSANOW, &nflags) != 0)
+		BOOST_THROW_EXCEPTION(ExternalFunctionFailure("tcsetattr"));
+
+	printf("%s", _prompt.c_str());
+	if (!fgets(password, sizeof(password), stdin))
+		BOOST_THROW_EXCEPTION(ExternalFunctionFailure("fgets"));
+	password[strlen(password) - 1] = 0;
+
+	// restore terminal
+	if (tcsetattr(fileno(stdin), TCSANOW, &oflags) != 0) {
```
subtly: curly

subtly:
macos :+1:

PR #1886, opened 2015-05-13
Block Info Test

winsvega:
Should block.info be equal at state and the blockchain?


test/libethereum/blockchain.cpp
```diff
@@ -795,4 +795,45 @@ BOOST_AUTO_TEST_CASE(userDefinedFile)
...
+	bytes blockRLP = dev::test::importByteArray(blockRLPstr);
+	BlockInfo block(blockRLP);
+	trueBc.import(blockRLP, trueState.db());
+	trueState.sync(trueBc);
+
+	if (trueState.info() != trueBc.info())
```
subtly: Can BOOST_CHECK or BOOST_REQUIRE be used here?

subtly:
What does this test?

test/libethereum/blockchain.cpp
winsvega: No. Too much output with '<<'

winsvega:
https://github.com/ethereum/cpp-ethereum/issues/1940  
I suppose this (info()) values should be equal. I might be wrong. 
However it is not equal. I wrote this test to see that it is not equal for sure. So it might be an issue.

CJentzsch:
It should be equal, please dig a little deeper. Output the complete BlockInfo for both to see where's the difference.
Also do a try catch around the sync and the block import, could be that one of them failed.

winsvega:
I found. 
https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/State.cpp#L340

The state is resetting it's current block to be the next one. So it returns i+1 block after getting synced with the blockchain. May be this is how it supposed to be. Just confusing.

CJentzsch:
If so, than the test is wrong.

winsvega:
I still think state and blockchain interfaces should be refactored.

PR #1890, opened 2015-05-13
Use seedhash everywhere in EthashAux

LefterisJP:
We now use seedhash everywhere and generate a block number only where
it's needed. Namely ethash_light_new()

### DIFF
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -857,7 +857,7 @@ int main(int argc, char** argv)
 				auto boundary = bi.boundary();
 				m = boost::to_lower_copy(string(argv[++i]));
 				bi.nonce = h64(m);
-				auto r = EthashAux::eval((uint64_t)bi.number, powHash, bi.nonce);
+				auto r = EthashAux::eval(bi.seedHash(), powHash, bi.nonce);
 				bool valid = r.value < boundary;
 				cout << (valid ? "VALID :-)" : "INVALID :-(") << endl;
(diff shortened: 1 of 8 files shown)


eth/main.cpp
```diff
@@ -866,7 +866,7 @@ int main(int argc, char** argv)
 				cout << "  with seed as " << seedHash << endl;
 				if (valid)
 					cout << "(mixHash = " << r.mixHash << ")" << endl;
-				cout << "SHA3( light(seed) ) = " << sha3(EthashAux::light((uint64_t)bi.number)->data()) << endl;
+				cout << "SHA3( light(seed) ) = " << sha3(EthashAux::light(bi)->data()) << endl;
```
gavofyork: any reason to use `bi` and not `bi.seedHash()`?

LefterisJP: We already got the function overload for `BlockInfo` to `seedHash` for `light()` so I thought why not use it? Makes for smaller line length

gavofyork: line length isn't everything :)

by passing BlockInfo rather than seedhash, more data is being passed yet merely the same amount of information. this makes possible several errors (e.g. trying to use bi.number) that wouldn't otherwise be possible.

even though you know you're not going to make that mistake right now, it's generally better to organise the structure of code to make such things impossible.

LefterisJP: Fair enough. But that means that this function should be deleted in general since we want to loosen the coupling between EthashAux and other modules. On it right now!

PR #1905, opened 2015-05-14
call m_miner->finish after mining

CJentzsch:
this function is never called. I assume it is needed to free the memory, but I am not an OpenCL expert.


gavofyork:
included in destructor in latest.

PR #1908, opened 2015-05-14
Trie performance optimizations

arkpar:
* SHA3 was calculated twice for nodes in some cases.
* Replaced FATDB with secure Trie + simple hash to key mapping

Further optimizations won't gain much since 95% insertion time is SHA3 calculation. Might worth looking into faster SHA3 implementation.


gavofyork:
running it on the network a little while before merging

chfast:
https://github.com/coruus/keccak-tiny looks to be faster than Crypto++. 

We should also try a memory aligned implementation.

PR #1919, opened 2015-05-15
FreeBSD fix for ethash byte ordering

tymat:
byteswap.h does not exist on FreeBSD and related platforms (i.e. NetBSD, OpenBSD, and DragonFlyBSD) which breaks compilation of ethash.

Thankfully FreeBSD adds [byteorder related operations](https://www.freebsd.org/cgi/man.cgi?query=byteorder&sektion=9).


LefterisJP:
Thank you but can you open the same PR on the actual ethash repo? https://github.com/ethereum/ethash

As far as the non OpenCL stuff is related that is the master repo and cpp-ethereum just `git subtree pull` from it.

PR #1922, opened 2015-05-15
Update gas limit modulation.

subtly:


libethcore/BlockInfo.cpp
```diff
@@ -228,7 +228,10 @@ u256 BlockInfo::selectGasLimit(BlockInfo const& _parent) const
 		return c_genesisGasLimit;
 	else
 		// target minimum of 3141592
-		return max<u256>(max<u256>(c_minGasLimit, 3141592), _parent.gasLimit - _parent.gasLimit / c_gasLimitBoundDivisor + 1 + (_parent.gasUsed * 6 / 5) / c_gasLimitBoundDivisor);
+		if (_parent.gasLimit < c_genesisGasLimit)
+			return min<u256>(c_genesisGasLimit, _parent.gasLimit + _parent.gasLimit / c_gasLimitBoundDivisor - 1);
```
CJentzsch: why using min her. You as increase as much as allowed.

chriseth: Note that this tries to increase the limit maximally, independent of the parent gas use, so I think it is good not to overshoot the target here.

CJentzsch: Both ways are fine. I actually prefer the increase as much as possible when you go below the minimum target, but going to the target is fine too.

PR #1925, opened 2015-05-15
DAG creation should also be under Guard

LefterisJP:
If 2 threads get in the DAG creation and it does not exist both would try to create it. We also need to guard the DAG creation itself.


gavofyork:
TODO: `tryFull` as the only public API. DagGenerator should be the only thread that can use it.

Issue #1943, opened 2015-05-16
ethminer windows dag creation fails


subtly:
When creating dag with ethminer on windows, creation of the DAG fails. And, I don't think it creates the DAG. Log says it's loading from ethash, then done, but it only takes a few seconds and no file is saved. It's not generated the dag because that should take much much longer. ping @LefterisJP

subtly:
And here is the most likely culprit:

![filepath](http://i.imgur.com/oltlFzV.png)

LefterisJP:
Windows I/O ... will need to provide a fix and test it to see if it's indeed that.

Genoil:
See #1914

tcoulter:
@LefterisJP I should buy you a :beer:

As per the comments in [the stackoverlow article @LefterisJP linked to](http://stackoverflow.com/questions/9542611/how-to-get-the-current-users-home-directory-in-windows), people suggested using the function `SHGetFolderPath` instead of `SHGetFolderPathW`. When I compiled with `SHGetFolderPath`, ethminer was able to create my DAG successfully (giving percent output, even) and I was able to start mining. 

Note that Ethash's default directly is `AppData/Ethash`, which doesn't seem to be within the typical Windows conventions. As far as I can tell, `AppData/Roaming/Ethash` or `AppData/Local/Ethash` should be used (but I don't know which one). 

Still, `SHGetFolderPath` (and not `SHGetFolderPathW`) seems to be the way to go. My output when benchmarking is now this, instead of a crash: :tada: 

```
$ ethminer.exe -G -M --opencl-device 1
  D  12:15:29|main  #00004000ΓÇª
Benchmarking on platform: { "platform": "AMD Accelerated Parallel Processing", "device": "Tahiti", "version": "OpenCL 1.2 AMD-APP (1642.5)" }
Preparing DAG...
  i  12:15:44|main  Loading from libethash...
DAG  12:15:58|main  Generating DAG file. Progress: 0 %
DAG  12:16:10|main  Generating DAG file. Progress: 1 %
<snip>
DAG  12:34:52|main  Generating DAG file. Progress: 98 %
DAG  12:35:04|main  Generating DAG file. Progress: 99 %
DAG  12:35:15|main  Generating DAG file. Progress: 100 %
  i  12:35:15|main  Done loading.
  D  12:35:15|main  pause took 0
  i  12:35:15|gpuminer0  workLoop 0 #00000000ΓÇª #00000000ΓÇª
[...]

PR #1948, opened 2015-05-17
Fix DAG creation for GPU mining on windows.

subtly:
GetFolderPathW needs wchar_t and can otherwise fail (returning an invalid path) and use AppData path instead of users home directory. Use Local AppData instead of roaming to users on domain networks aren't uploading DAG files to servers.


LefterisJP:
Reminder: This is an ethash change. As we discussed once confirmed please open a PR in the ethash repo itself. We will subtree pull it in cpp ethereum.

subtly:
@LefterisJP Is it a subtree now or copy/paste?

LefterisJP:
Subtree as you can see from this PR: https://github.com/ethereum/cpp-ethereum/pull/1816
As for the exact method of subtree pulling you can see [here](http://stackoverflow.com/questions/22334382/git-subtree-only-one-file-or-directory/22335955).

This is why this should be a PR on the ethash repo itself and not on cpp-ethereum.

LefterisJP:
By the way it seems that we have confirmation of succesful mining by users who tinkered with the code.
https://github.com/ethereum/cpp-ethereum/issues/1943#issuecomment-102841521

So yes the path is the issue.

libethash/io_posix.c
```diff
@@ -86,9 +90,22 @@ bool ethash_file_size(FILE* f, size_t* ret_size)
 
 bool ethash_get_default_dirname(char* strbuf, size_t buffsize)
 {
+#if defined(__APPLE__)
+	static const char dir_suffix[] = "Ethash/";
+	strbuf[0] = '\0';
+	char* home_dir = getenv("HOME");
+	if (!home_dir || strlen(home_dir) == 0)
```
LefterisJP: Why are you making this change? If the user has not set a $HOME variable and is in Linux/Mac then it's his problem. The `getpwuid()` function is another way of doing the same thing as querying $HOME. I say we advice our users set $HOME and keep an identical implementation for Posix compliant Operating systems.

If you really want this in I suppose it does not hurt, but then there is absolutely no need for the #ifndef. This will work on all Linux systems too.

libethash/io_posix.c
```diff
@@ -86,9 +90,22 @@ bool ethash_file_size(FILE* f, size_t* ret_size)
 
 bool ethash_get_default_dirname(char* strbuf, size_t buffsize)
 {
+#if defined(__APPLE__)
+	static const char dir_suffix[] = "Ethash/";
```
LefterisJP: This is not following the [spec](https://github.com/ethereum/wiki/wiki/Ethash-DAG)

LefterisJP:
Putting "Got issues" for the reasons stated above, but mainly for being opened in the wrong repo at the moment. I guess though it's much easier for Alex to test his changes like this so it's understandable.

libethash/io_posix.c
subtly: Not using getpwuid() simply breaks on macs. Windows isn't POSIX anyways/either, so mac is getting an exception here too. Now that you mention it, I'm going to change this back to .ethash for now. For macos, similar to windows, we have to make a special call which returns the appropriate "Application Data" folder.

LefterisJP: If you need the `getpwuid()` call for Mac put it in but remove the ifndef. It will work on Linux too.

As for the "Application Data" Folder, Windows isn't Posix compliant but Macosx ... mostly ... is. Why change it to a specific application data folder for Mac? Doesn't Mac have the notion of a user's home directory? Can't an application create a subdirectory in there called `.ethash`? If yes then why bother having it in a different location?

Windows has and always will be an exception. Macosx should be only when it's absolutely necessary. Of course not being a MacOS user I may be missing something so please bear with my questions.

subtly: Side Note: I'll create final PR for this on ethash monday morningish. (this being one of the changes to be made)

Regarding macos. Using .* folders is pretty much deprecated in favour of things like ~/Library/Application Support and plist Preferences. We could just put everything in .* folders like in Linux but then AlethZero would be in a separate path -- one or the other had to be used and for a few months now the codebase was putting everything in ~/Library/Application Support. (Fwiw, the plists are sort of like windows registry)

Re: POSIX and such. Apple also has a sandboxing system which it can run apps in. When it does, the app can't access a home directory, it can only access application data. As said before, for macos, getpwuid actually needs to be replaced with the appropriate CF methods.

For now I'll be changing it back to .ethash because I for the 'Library/Application Support' prefix.

PR #1951, opened 2015-05-17
Little improved base64 implementation.

chfast:
Global variable type changed from dynamic constructed std::string to constant char array.
Truncating conversion eliminated (no MSVC warning).
Unit tests added.


libdevcore/Base64.cpp
```diff
@@ -40,6 +38,11 @@ static inline bool is_base64(byte c) {
 	return (isalnum(c) || (c == '+') || (c == '/'));
 }
 
+static inline byte find_base64_char_index(byte c) {
+	auto it = std::find(base64_chars, base64_chars + sizeof(base64_chars), c);
```
chriseth: What about
```
if ('A' <= c && c <= 'Z') return c - 'A';
else if ('a' <= c && c <= 'z') return c - 'a' + 1 + find_base64_char_index('Z');
else if (c == '+') return 1 + find_base64_char_index('z');
else if (c == '/') return 1 + find_base64_char_index('+');
else return 1 + find_base64_char_index('/');
```
That could result in about a 10-fold improvement.

test/libdevcrypto/hexPrefix.cpp
```diff
@@ -60,4 +61,30 @@ BOOST_AUTO_TEST_CASE(hexPrefix_test)
 	}
 }
 
+BOOST_AUTO_TEST_CASE(base64)
+{
+	static char const* const s_tests[][2] = {
```
chriseth: `{` should go to the next line, also in other places of this PR.

libdevcore/Base64.cpp
chfast: It's probably good idea. I will check whole implementation again.

gavofyork:
"I will check whole implementation again." please resubmit once done.

chfast:
I applied @chriseth's proposition for improvement. The compiler wasn't smart enough. That will probably rise more MSVC warnings than it was before...

PR #1952, opened 2015-05-17
More block import optimizations

arkpar:


libethereum/BlockQueue.cpp
```diff
@@ -38,7 +38,8 @@ const char* BlockQueueChannel::name() { return EthOrange "▣┅▶"; }
 
 BlockQueue::BlockQueue()
 {
-	for (unsigned i = 0; i < thread::hardware_concurrency(); ++i)
+	unsigned verifierThreads = std::max(thread::hardware_concurrency(), 3U) - 2U;
```
chfast: Explanation would be nice.

chfast:
Looks Good To Me™, but can you explain what is this optimization about?

libethcore/EthashAux.cpp
```diff
@@ -101,13 +101,13 @@ uint64_t EthashAux::number(h256 const& _seedHash)
...
 {
-	RecursiveGuard l(x_lights);
+	WriteGuard l(x_lights);
 	m_lights.erase(_s);
 }
 
 EthashAux::LightType EthashAux::light(h256 const& _seedHash)
 {
-	RecursiveGuard l(get()->x_lights);
+	ReadGuard l(get()->x_lights);
```
arkpar: Prevents contention on parallel ethash verification

libethereum/BlockChain.cpp
```diff
@@ -317,7 +317,7 @@ tuple<h256s, h256s, bool> BlockChain::sync(BlockQueue& _bq, OverlayDB const& _st
 	{
 		try
 		{
-			auto r = import(block.first, block.second, _stateDB);
+			auto r = import(block.first, block.second, _stateDB, ImportRequirements::Default & ~ImportRequirements::ValidNonce);
```
arkpar: Nonce is already checked by verifier thread at this point. No need to check it again

chfast: This comment should be in the code.

PR #1953, opened 2015-05-18
Add deque for windows to compile.

subtly:


tcoulter:
I've verified that in order to compile `develop` on Windows at the time of this writing I need these changes to BlockQueue.h. However, I can't vouch for the changes to WhisperPeer.cpp. Someone with more knowledge should take a look.

PR #1954, opened 2015-05-18
Fixed error introduced in 953be7886b5681f0fa8ecde94b1fdb8c25afbfd4

holiman:
A file wasn't changed according to the changes


chriseth:
Thanks for the fix!

PR #1955, opened 2015-05-18
Fixing windows build

arkpar:


chfast:
The same is done in #1953.

chriseth:
#1953 does something else apart from the mere build fix.

gavofyork:
merged #1953 instead.

Issue #1960, opened 2015-05-18
Constructor is not part of solidity ABI


chriseth:
The solidity compiler does not include the constructor as part of the interface generated with the switch "--sol-abi".

PR #1979, opened 2015-05-19
Added constructor as part of Solidity ABI

LianaHus:
issue #1960

### DIFF
--- a/libsolidity/InterfaceHandler.cpp
+++ b/libsolidity/InterfaceHandler.cpp
@@ -108,8 +108,7 @@ unique_ptr<string> InterfaceHandler::getABISolidityInterface(ContractDefinition 
 {
 	string ret = "contract " + _contractDef.getName() + "{";
 
-	auto populateParameters = [](vector<string> const& _paramNames,
-			vector<string> const& _paramTypes)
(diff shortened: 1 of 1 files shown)


libsolidity/InterfaceHandler.cpp
```diff
@@ -107,17 +107,28 @@ std::unique_ptr<std::string> InterfaceHandler::getABIInterface(ContractDefinitio
 unique_ptr<string> InterfaceHandler::getABISolidityInterface(ContractDefinition const& _contractDef)
 {
 	string ret = "contract " + _contractDef.getName() + "{";
+
+	auto populateParameters = [](vector<string> const& _paramNames,
+			vector<string> const& _paramTypes)
```
chriseth: Might this fit one line now? If not, please correct indentation.

chriseth:
Hm, just noticed that this code never generated the "returns (...)" part. We might add that too.

LianaHus:
yeah we never have "returns....".
do you want me to do it in this PR?

chriseth:
Does not need to be in this PR.

chriseth:
Failure in Linux C++ is probably due to out of memory on the build server.

Issue #1977, opened 2015-05-19
block 137555 causing Gas overflow in VM.cpp


LefterisJP:
#### Observed problem

When importing the downloaded blockchain I was constantly stuck at block 137555. It could never get out of there no matter how many times I deleted the entire blockchain or try different peers.

 After some debugging along with @arkpar we found out there is an overflow in VM.cpp in the `boost::u256` number. The overflow happens around [VM.cpp:625](https://github.com/ethereum/cpp-ethereum/blob/develop/libevm/VM.cpp#L625)

as you can see from the gdb output

```
EVM  02:47:17 PM|eth   < 0 : @79652f18… : #39627 : 3140 : SUB : 17 : -3 : 0x32 >
EVM  02:47:17 PM|eth  
    STACK
0000000000000000000000000000000000000000000000000000000045d96cd7
000000000000000000000000000000000000000000000000000000000001d9a0
000000000000000000000000000000000000000000000000000000000001d9dc
000000000000000000000000000000000000000000000000000000000001da40
0000000000000000000000000000000000000000000000000000000000000020
0000000000000000000000000000000000000000000000000000000000000a20
0000000000000000000000000000000000000000000000000000000000000064
000000000000000000000000000000000000000000000000000000000001d9dc
0000000000000000000000000000000000000000000000000000000000000000
00000000000000000000000079652f18f281c113fac20e8434e223228340e383
ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe4
    MEMORY
 mem size greater than 1000 bytes     STORAGE

EVM  02:47:17 PM|eth   < 0 : @79652f18… : #39628 : 3141 : CALL : 14 : -115792089237316195423570985008687907853269984665640564039457584007913129639948 : 0x32 >

[...]

PR #1981, opened 2015-05-19
Test for issue #1977

LefterisJP:
Test that reproduces issue #1977 and will fail on ArchLinux with
Boost version: 1.58.0-1
gcc version: 4.9.2-4

I am opening a PR with it just to enable other people who want to test if they can reproduce the issue on their system.


chfast:
It is not a VM test. There is a "BasicTests" suite in test/libdevcrypto.

chriseth:
If I see that correctly, libdevcrypto does not even depend on boost's bigints.
I would say either leave it where it is or create a new test suite...

chfast:
Yeah. There is mess there. We don't have basic unit tests. Let just keep
the mess out of the VM part of the testeth.

chriseth:
Why not put it in the libdevcore folder? libdevcore/Common.h defines u256, so I think a good fit would be a test suite called something like "externalDependencies" inside the libdevcore test folder.

chfast:
I agree with "externalDependencies". If you do not have time for that I can do it later.

test/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ endmacro()
 
 add_subdirectory(fuzzTesting)
 add_subdirectory(libdevcore)
+add_subdirectory(libdevcore/external_dependencies)
```
CJentzsch: rather do a `add_subdirectory(external_dependencies)` in `libdevcore/CMakeLists.txt`

chfast: I don't think we use unserscore in folder/file names anywhere.

test/libdevcore/external_dependencies/boost.cpp
```diff
@@ -0,0 +1,37 @@
...
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/** @file boost.cpp
+ * @author Lefteris Karapetsas <lefteris@ethdev.com>
+ * @date 205
```
chfast: Does not look like a date.

test/libdevcore/external_dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+cmake_policy(SET CMP0015 NEW)
+
+aux_source_directory(. SRCS)
```
chfast: This is not the best way of adding files to CMake. The best is to just list all files.

LefterisJP: This is the way that's used all over the tests

test/libdevcore/external_dependencies/boost.cpp
LefterisJP: indeed it does not :)

PR #1983, opened 2015-05-19
Gas estimation taking known state into account.

chriseth:
Gas estimation takes into account as much as it can possibly know about the current state of the VM and thus makes it much more exact, e.g. computing the gas cost of instructions that use memory is possible.

### DIFF
--- a/libevmasm/GasMeter.h
+++ b/libevmasm/GasMeter.h
@@ -22,6 +22,7 @@
 #pragma once
 
 #include <ostream>
+#include <libevmasm/ExpressionClasses.h>
 #include <libevmasm/AssemblyItem.h>
 
 namespace dev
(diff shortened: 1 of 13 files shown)


libevmasm/GasMeter.h
```diff
@@ -47,11 +53,28 @@ class GasMeter
...
+	/// Constructs a new gas meter given the current state.
+	GasMeter(std::shared_ptr<KnownState> const& _state): m_state(_state) {}
+
+	/// @returns an upper bound on the gas consumed by the given instruction and updates
+	/// the state.
 	GasConsumption estimateMax(AssemblyItem const& _item);
 
 private:
+	/// @returns _multiplier * (_value + 31) / 32, if _value is a known constant and infinite otherwise.
+	GasConsumption wordGas(u256 const& _multiplier, ExpressionClasses::Id _value);
```
LianaHus: could it be const? as well as memoryGas?

chriseth: No, they cannot. They query expression classes and any query to expression classes might create a new class.

LianaHus: I see

PR #1998, opened 2015-05-21
Bug fix: Net fixes x2

subtly:


libethereum/EthereumHost.cpp
```diff
@@ -218,25 +218,32 @@ void EthereumHost::maintainTransactions()
...
+tuple<vector<shared_ptr<EthereumPeer>>, vector<shared_ptr<EthereumPeer>>, list<shared_ptr<Session>>> EthereumHost::randomSelection(unsigned _percent, std::function<bool(EthereumPeer*)> const& _allow)
 {
-	pair<vector<shared_ptr<EthereumPeer>>, vector<shared_ptr<EthereumPeer>>> ret;
-	ret.second.reserve(peerSessions().size());
-	for (auto const& j: peerSessions())
+	vector<shared_ptr<EthereumPeer>> chosen;
+	vector<shared_ptr<EthereumPeer>> allowed;
+	list<shared_ptr<Session>> sessions;
+	
+	auto ps = peerSessions();
```
chfast: Avoid copy.

libethereum/EthereumHost.cpp
```diff
@@ -218,25 +218,32 @@ void EthereumHost::maintainTransactions()
...
 	{
-		unsigned n = rand() % ret.second.size();
-		ret.first.push_back(std::move(ret.second[n]));
-		ret.second.erase(ret.second.begin() + n);
+		unsigned n = rand() % allowed.size();
+		chosen.push_back(std::move(allowed[n]));
+		allowed.erase(allowed.begin() + n);
 	}
-	return ret;
+	return make_tuple(chosen, allowed, sessions);
```
chfast: std::move should be used here for all arguments.

libp2p/Session.cpp
```diff
@@ -319,10 +319,14 @@ void Session::send(bytes&& _msg)
 
 void Session::write()
 {
-	const bytes& bytes = m_writeQueue[0];
-	m_io->writeSingleFramePacket(&bytes, m_writeQueue[0]);
+	bytes *out;
```
chfast: const disappeared.

libethereum/EthereumHost.cpp
```diff
@@ -218,25 +218,32 @@ void EthereumHost::maintainTransactions()
...
 
-pair<vector<shared_ptr<EthereumPeer>>, vector<shared_ptr<EthereumPeer>>> EthereumHost::randomSelection(unsigned _percent, std::function<bool(EthereumPeer*)> const& _allow)
+tuple<vector<shared_ptr<EthereumPeer>>, vector<shared_ptr<EthereumPeer>>, list<shared_ptr<Session>>> EthereumHost::randomSelection(unsigned _percent, std::function<bool(EthereumPeer*)> const& _allow)
 {
-	pair<vector<shared_ptr<EthereumPeer>>, vector<shared_ptr<EthereumPeer>>> ret;
-	ret.second.reserve(peerSessions().size());
-	for (auto const& j: peerSessions())
+	vector<shared_ptr<EthereumPeer>> chosen;
+	vector<shared_ptr<EthereumPeer>> allowed;
+	list<shared_ptr<Session>> sessions;
```
chfast: Any reason list is used in this case?

libp2p/Session.cpp
subtly: This can't be const as it's assigned within the Guard.

chfast: Pointer itself cannot be (`bytes* const`) but can point a const bytes (`bytes const*`).

libethereum/EthereumHost.cpp
subtly: It doesn't need any particular data structure.

chfast: List is almost worse than vector.

libethereum/EthereumHost.cpp
subtly: My understanding is that make_tuple moves constructor arguments (&&).

libp2p/Session.cpp
subtly: Ah. Fixed.

libethereum/EthereumHost.cpp
chfast: That's somehow correct. make_tuple moves arguments but only if you pass a rvalue reference. Currently you're passing lvalue references as all arguments have a name. To make them a rvalue reference you need to cast. The cast instruction is just `std::move`.

libp2p/Session.cpp
gavofyork: whitespace wrong, too.

gavofyork:
all chfast's comments should be addressed.

subtly:
All fixed! buildbots pending.

Issue #1791, opened 2015-05-02
Suggestion for mix: Simulate constant mining option.


Smithgift:
This hasn't come up for me that much yet, but I'd like a way to have mix spawn new blocks on an average of every twelve seconds, to test a Dapp in a world where blocks are happening automatically.

Smithgift:
This works, thanks.

Issue #1790, opened 2015-05-02
Two minor mix cosmetic issues.


Smithgift:
I can't say either of these really deserves it's own issue.

* The `delete` keyword is not highlighted in mix's editor.
* In CTRL-F's little bar, the underscore is invisible. It's still there, you just can't see it.

PR #2004, opened 2015-05-21
Mix - Bug fixes

yann300:


yann300:
#1791 
#1790

mix/ClientModel.cpp
```diff
@@ -389,7 +389,7 @@ std::pair<QString, int> ClientModel::resolvePair(QString const& _contractId)
 {
 	 std::pair<QString, int> ret;
```
chfast: How about std::make_pair?

mix/CodeModel.cpp
```diff
@@ -358,7 +360,7 @@ void CodeModel::gasEstimation(solidity::CompilerStack const& _cs)
 {
 	if (m_gasCostsMaps)
 		m_gasCostsMaps->deleteLater();
-	m_gasCostsMaps = new GasMapWrapper(this);
+	m_gasCostsMaps = new GasMapWrapper();
```
chfast: 1. A raw `new` allocation? Does it mean manual memory management?
2. Avoid `()` after object creation expression. You can use `{}` or nothing. http://stackoverflow.com/questions/1424510/most-vexing-parse-why-doesnt-a-a-work

mix/CodeModel.h
```diff
@@ -137,7 +137,7 @@ class GasMapWrapper: public QObject
 	Q_PROPERTY(GasCostsMaps gasMaps MEMBER m_gasMaps CONSTANT)
 
 public:
-	GasMapWrapper(QObject* _parent): QObject(_parent){}
+	GasMapWrapper(QObject* _parent = 0): QObject(_parent){}
```
chfast: ` = nullptr`

chfast:
I haven't checked QML.
Minor issues in C++.

mix/CodeModel.cpp
yann300: 1. usually we have to set the parent. once the parent is deleted all the children are deleted (this is how works QOBject). This is not possible here because the parent is in an another thread (and it gives warnings).
now we call deleteLater() in the destructor of codemodel.cpp.

chfast:
C++ looks good to me. If you do not need QML review you can mark it as "looks good".

mix/CodeModel.cpp
```diff
@@ -197,6 +197,8 @@ CodeModel::~CodeModel()
 	stop();
 	disconnect(this);
 	releaseContracts();
+	if (m_gasCostsMaps)
+		m_gasCostsMaps->deleteLater();
```
arkpar: There will be no event handlers after this, should be deleted immediatelly i think

PR #2006, opened 2015-05-22
Functional gas estimator

chriseth:
This adds the feature to estimate the gas costs of a specific function of a contract on arbitrary input.

### DIFF
--- a/libevmasm/GasMeter.cpp
+++ b/libevmasm/GasMeter.cpp
@@ -148,7 +148,7 @@ GasMeter::GasConsumption GasMeter::estimateMax(AssemblyItem const& _item)
 			if (u256 const* value = classes.knownConstant(m_state->relativeStackElement(-1)))
 				gas += c_expByteGas * (32 - (h256(*value).firstBitSet() / 8));
 			else
-				gas = GasConsumption::infinite();
+				gas += c_expByteGas * 32;
 			break;
 		default:
 			break;
--- a/libevmasm/GasMeter.cpp
+++ b/libevmasm/GasMeter.cpp
@@ -29,12 +29,13 @@ using namespace dev::eth;
 
 GasMeter::GasConsumption& GasMeter::GasConsumption::operator+=(GasConsumption const& _other)
 {
-	isInfinite = isInfinite || _other.isInfinite;
+	if (_other.isInfinite && !isInfinite)
+		*this = infinite();
 	if (isInfinite)
 		return *this;
 	bigint v = bigint(value) + _other.value;
 	if (v > std::numeric_limits<u256>::max())
-		isInfinite = true;
+		*this = infinite();
 	else
 		value = u256(v);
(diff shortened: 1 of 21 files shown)


mix/CodeModel.cpp
```diff
@@ -348,8 +348,7 @@ void CodeModel::gasEstimation(solidity::CompilerStack const& _cs)
 			continue;
 		dev::solidity::SourceUnit const& sourceUnit = _cs.getAST(*contractDefinition.getLocation().sourceName);
 		AssemblyItems const* items = _cs.getRuntimeAssemblyItems(n);
-		StructuralGasEstimator estimator;
-		std::map<ASTNode const*, GasMeter::GasConsumption> gasCosts = estimator.breakToStatementLevel(estimator.performEstimation(*items, std::vector<ASTNode const*>({&sourceUnit})), {&sourceUnit});
+		std::map<ASTNode const*, GasMeter::GasConsumption> gasCosts = GasEstimator::breakToStatementLevel(GasEstimator::structuralEstimation(*items, std::vector<ASTNode const*>({&sourceUnit})), {&sourceUnit});
```
LianaHus: I think there is no need for "std::" here

libevmasm/GasMeter.cpp
```diff
@@ -29,12 +29,13 @@ using namespace dev::eth;
 
 GasMeter::GasConsumption& GasMeter::GasConsumption::operator+=(GasConsumption const& _other)
 {
-	isInfinite = isInfinite || _other.isInfinite;
+	if (_other.isInfinite && !isInfinite)
+		*this = infinite();
 	if (isInfinite)
 		return *this;
 	bigint v = bigint(value) + _other.value;
 	if (v > std::numeric_limits<u256>::max())
```
LianaHus: std::

test/libsolidity/GasMeter.cpp
```diff
@@ -47,30 +48,47 @@ class GasMeterTestFramework: public ExecutionFramework
...
+		GasMeter::GasConsumption gas = meter.estimateMax(0, state);
+		u256 bytecodeSize(m_compiler.getRuntimeBytecode().size());
 		gas += bytecodeSize * c_createDataGas;
 		BOOST_REQUIRE(!gas.isInfinite);
 		BOOST_CHECK(gas.value == m_gasUsed);
 	}
 
+	/// Compares the gas computed by PathGasMeter for the given signature (but unknown arguments)
+	/// against the actual gas usage computed by the VM on the given set of argument variants.
+	void testRunTimeGas(std::string const& _sig, vector<bytes> _argumentVariants)
```
LianaHus: std::

mix/CodeModel.cpp
chriseth: There is no "using std" probably because of the many name clashes between Qt and std.

PR #2007, opened 2015-05-22
Fixed segmentation fault, also added more user confirmation

holiman:
…before killing the selected account when a user selects "Kill Account" in alethzero. The user now has to explicitly write 'YES' in a box (regardless of whether the account holds funds or not - it may still own contracts). 

The segfault happened because `ui->accounts->currentItem` was used instead of `ui->ourAccounts->currentItem`

### DIFF
--- a/alethzero/MainWin.cpp
+++ b/alethzero/MainWin.cpp
@@ -1984,6 +1984,7 @@ void Main::on_newAccount_triggered()
 		m_keyManager.import(p.secret(), s.toStdString());
 	keysChanged();
 }
+
 void Main::on_killAccount_triggered()
 {
 	if (ui->ourAccounts->currentRow() >= 0)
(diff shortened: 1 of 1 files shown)


alethzero/MainWin.cpp
```diff
@@ -1984,22 +1984,26 @@ void Main::on_newAccount_triggered()
 		m_keyManager.import(p.secret(), s.toStdString());
 	keysChanged();
 }
-
```
chfast: Don't like this empty line?

alethzero/MainWin.cpp
```diff
@@ -1984,22 +1984,26 @@ void Main::on_newAccount_triggered()
...
+
 		auto k = m_keyManager.accountDetails()[h];
-		if (
-			ethereum()->balanceAt(h) != 0 &&
-			QMessageBox::critical(this, QString::fromStdString("Kill Account " + k.first + "?!"),
-				QString::fromStdString("Account " + k.first + " (" + render(h) + ") has " + formatBalance(ethereum()->balanceAt(h)) + " in it. It, and any contract that this account can access, will be lost forever if you continue. Do NOT continue unless you know what you are doing.\n"
-				"Are you sure you want to continue?"),
-				QMessageBox::Yes, QMessageBox::No) == QMessageBox::No)
+	
+		//ethereum()->balanceAt(h) != 0 &&
```
chfast: Remove this comment if this condition not needed.

holiman:
Fixed the comments

chfast:
Thanks.

PR #2008, opened 2015-05-22
evmJit warnings fix

winsvega:


chfast:
JIT is only for 64-bit systems. Currently and probably forever. Sorry.

PR #2016, opened 2015-05-26
fix output memory for CALL

CJentzsch:
In the YP in Appendix H in CALL , the output *o* is described as the values in the memory from µ[5] to µ[6]. Even for a normal STOP in the CALL , those values are zero and should overwrite the parent memory (see Equation 92). This fixes the current fork.


gavofyork:
Point me to the place in the YP where it states that the CALLer's memory be zero-filled.

Issue #570, opened 2014-12-06
Improve UPnP support


kapacorn:
Windows 7 and 8 issue for alethzero

*N* [ 14:13:32 | main ] Id: 8a569788…
*** [ 14:13:32 | main ] UPnP device not found.
*N* [ 14:13:32 | main ] Id: 8a569788…
!!! [ 14:13:58 | p2p ] Error reading: The semaphore timeout period has expired
*** [ 14:17:33 | main ] UPnP device not found.
*N* [ 14:17:33 | main ] Id: 8a569788…
*** [ 14:17:33 | main ] UPnP device not found.
*N* [ 14:17:33 | main ] Id: 8a569788…
*** [ 14:17:34 | main ] UPnP device not found.
*N* [ 14:17:34 | main ] Id: 8a569788…
*** [ 14:17:35 | main ] UPnP device not found.
*N* [ 14:17:35 | main ] Id: 8a569788…
*** [ 14:17:36 | main ] UPnP device not found.
*N* [ 14:17:36 | main ] Id: 8a569788…
*** [ 14:17:36 | main ] UPnP device not found.

loreV:
Hey guys.. I found this similar to the problem I have: I receive a message "UPnP device not found" followed by "Terminating discover after 0 rounds", after clicking "Go". Is there anybody who has experienced this issue?

debris:
should be fixed in #2022

PR #2021, opened 2015-05-26
Added UPnP support to Windows (#570)

lotrf3:
Adds a dependency on miniupnpc (already a dependency on Linux, just not on Windows).
Requires [precompiled lib](https://drive.google.com/file/d/0B8p3EHFIFv1dMDRUN2RUb3VxekE/view?usp=sharing) be uploaded to https://build.ethdev.com/builds/windows-precompiled
Lib can be reproduced by downloading [source](http://miniupnp.tuxfamily.org/files/) and then running [this .bat file](https://drive.google.com/file/d/0B8p3EHFIFv1ddS02WmhJemFvaWM/view?usp=sharing). Note that .bat file outputs to `/miniupnpc-1.9/miniupnpc-1.9.tar.gz`, NOT `/miniupnpc-1.9.tar.gz`.
Windows devs will need to rerun `/extdep/getstuff.bat/` to get up to date.


subtly:
Not sure what this does but I can't build/compile on windows now. Says there are missing symbols.

PR #2023, opened 2015-05-27
Mix - Functional gas estimation

yann300:
- Functional gas estimation
 - bug fix: misselected address for input parameter


mix/CodeModel.h
```diff
@@ -253,6 +257,7 @@ public slots:
 	std::map<QString, dev::bytes> m_compiledContracts; //by name
 	dev::Mutex x_pendingContracts;
 	std::map<QString, QString> m_pendingContracts; //name to source
+	bool m_optimizeCode;
```
arkpar: Default value?

mix/CodeModel.h
```diff
@@ -138,7 +138,7 @@ class GasMapWrapper: public QObject
 
 public:
 	GasMapWrapper(QObject* _parent = nullptr): QObject(_parent){}
-	void push(QString _source, int _start, int _end, QString _value, bool _isInfinite);
+	void push(QString _source, int _start, int _end, QString _value, bool _isInfinite, QString _payload);
```
arkpar: Better not use magic strings in C++ code. Store an enum and convert to string in property getter

gavofyork:
could we get the buildbot pinged to check the linux builds are indeed good?

PR #2025, opened 2015-05-28
Executive on a diet

chfast:
This set of patches changes the way Executive collects execution results. Executive's user must provide a storage for results (ExecutionResult) up front. If not provided Executive will handle only outputs needed  for execution itself.

All this allows better handling of outputs from calls when the output buffer is known before the execution has begun. VM implementations are not stateless (you can execute EVM code using a temporary instance of a VM).


libethereum/Precompiled.cpp
```diff
@@ -44,47 +47,48 @@ static bytes ecrecoverCode(bytesConstRef _in)
...
-	{ 2, { [](bytesConstRef i) -> bigint { return c_sha256Gas + (i.size() + 31) / 32 * c_sha256WordGas; }, sha256Code }},
-	{ 3, { [](bytesConstRef i) -> bigint { return c_ripemd160Gas + (i.size() + 31) / 32 * c_ripemd160WordGas; }, ripemd160Code }},
-	{ 4, { [](bytesConstRef i) -> bigint { return c_identityGas + (i.size() + 31) / 32 * c_identityWordGas; }, identityCode }}
-};
+}
 
 std::unordered_map<unsigned, PrecompiledAddress> const& dev::eth::precompiled()
 {
-	return c_precompiled;
+	static const std::unordered_map<unsigned, PrecompiledAddress> s_precompiled =
```
gavofyork: `static const` -> `c_precompiled` (constness is more important than staticness).

chfast: Ok. I thought `c_` is for compile-time constants only.

I moved it inside the function because C++11 guarantees thread-safety in this case.

libethereum/Executive.h
```diff
@@ -112,30 +112,25 @@ class Executive
...
+
 	/// @returns the new address for the created contract in the CREATE operation.
 	h160 newAddress() const { return m_newAddress; }
 	/// @returns true iff the operation ended with a VM exception.
 	bool excepted() const { return m_excepted != TransactionException::None; }
 
-	/// Get the above in an amalgamated fashion.
-	ExecutionResult executionResult() const;
+	/// Collect execution results in the result storage provided.
+	void collectResult(ExecutionResult& _res) { m_res = &_res; }
```
gavofyork: should be `setResultRecipient`.

libethereum/Precompiled.cpp
```diff
@@ -44,47 +47,48 @@ static bytes ecrecoverCode(bytesConstRef _in)
...
 	h256 ret;
-
-	if ((u256)in.v > 28)
-		return ret.asBytes();
-	SignatureStruct sig(in.r, in.s, (byte)((int)(u256)in.v - 27));
-	if (!sig.isValid())
-		return ret.asBytes();
-
-	try
+	if ((u256)in.v <= 28)
```
gavofyork: should probably be >= 27, too

gavofyork:
looks reasonable otherwise. i presume it passes all tests on all platforms?

libethereum/Precompiled.cpp
chfast: If I get it right, v can be 27 or 28. @CJentzsch can you add tests for that?

chfast:
Required changes applied. All tests pass.

PR #2031, opened 2015-05-29
Mix - Add validation for input parameters

### LINKED ISSUE #2012: --master stopped working
Same on the build server, now only getting:
```
terminate called after throwing an instance of 'dev::eth::UnknownPassword'
  what():  std::exception
```

yann300:
- Add validation for input parameters
- Fix #2012

### DIFF
--- a/mix/qml/TransactionDialog.qml
+++ b/mix/qml/TransactionDialog.qml
@@ -6,6 +6,7 @@ import QtQuick.Window 2.0
 import QtQuick.Controls.Styles 1.3
 import org.ethereum.qml.QEther 1.0
 import "js/TransactionHelper.js" as TransactionHelper
+import "js/InputValidator.js" as InputValidator
 import "."
 
 Dialog {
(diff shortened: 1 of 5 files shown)


yann300:
TODO:
Validation for Enum

mix/qml/TransactionDialog.qml
```diff
@@ -503,10 +504,22 @@ Dialog {
...
-						accepted();
+						var invalid = InputValidator.validate(paramsModel, paramValues);
+						if (invalid.length === 0)
+						{
+							close();
+							accepted();
+						}
+						else
+						{
+							errorDialog.text = qsTr("some parameters are invalid:\n");
```
arkpar: Capitalize please

mix/qml/js/InputValidator.js
```diff
@@ -0,0 +1,112 @@
...
+			if (type.indexOf("int") !== -1)
+				res = validateInt(type, values[model[k].name]);
+			else if (type.indexOf("bytes") !== -1)
+				res = validateBytes(type, values[model[k].name]);
+			else if (type.indexOf("bool") !== -1)
+				res = validateBool(type, values[model[k].name]);
+			else if (type.indexOf("address") !== -1)
+				res = validateAddress(type, values[model[k].name]);
+			else
+				res = validateAddress(type, values[model[k].name]); //we suppose that this is a ctr type.
```
arkpar: Better check for contract type explicitly so that this code would not fail if a new type is introduced. What is the type value in case of a contract? If it contains contract name this code would fail for contracts with names such as "MyboolContract"

mix/qml/js/InputValidator.js
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+	ret.valid = nbRegEx.test(_value);
+	if (!ret.valid)
+		ret.message = _value + " does not represent " + _type + " type.";
+	else
+	{
+		var t = _type.replace("uint", "").replace("int", "");
+		var max = parseInt(t) / 4;
+		if (_value.length > max)
```
arkpar: Is value hex or dec here? If it is decimal this does not seem right

PR #2032, opened 2015-05-29
Jumpdest memory usage

CJentzsch:
The current version is vulnerable to a code which executes 1024 recursive CALLs and has a lot of JUMPDESTS at the end of the code like : <valid code>...005b5b5b..5b (see test).
Current version needs 1.2 GB of RAM for this test and only costs about 80k gas.


libevm/VM.cpp
```diff
@@ -546,7 +546,7 @@ bytesConstRef VM::go(ExtVMFace& _ext, OnOpFunc const& _onOp, uint64_t _steps)
 			break;
 		case Instruction::JUMP:
 			nextPC = m_stack.back();
-			if (!m_jumpDests.count(nextPC))
+			if (find(m_jumpDests.begin(), m_jumpDests.end(), (uint64_t)nextPC) == m_jumpDests.end() || nextPC > numeric_limits<uint64_t>::max() )
```
chriseth: Shouldn't the two conditions be swapped?

libevm/VM.h
```diff
@@ -74,7 +74,7 @@ class VM: public VMFace
 	u256 m_curPC = 0;
 	bytes m_temp;
 	u256s m_stack;
-	std::set<u256> m_jumpDests;
+	std::vector<uint64_t> m_jumpDests;
```
chriseth: `size_t` instead of `uint64_t`?

chriseth:
Change looks good, the comments are just suggestions.

PR #2037, opened 2015-05-29
Workaround for incorrect mixed type big int comparison

chfast:
Fixes ethereum/cpp-ethereum#1977.


gavofyork:
windows build problems.

chfast:
I will handle Windows tomorrow. Windows is not affected because we build boost from source so we can choose the correct version easily.

PR #2039, opened 2015-05-29
Increase win32 executable stack size to 32MB

LefterisJP:
@gluk256 Please confirm that windows stack tests pass with a clean build and this commit in. So make sure that the `testeth` executable is no longer affected by our `editbin` changes.


gluk256:
Works on my machine.

Issue #1982, opened 2015-05-19
Error GPU mining: clEnqueueMapBuffer ( -4 )


gillhofer:
a `./eth -M -G --opencl-device 0` results in 
```
[user@pc eth]$ ./eth -M -G --opencl-device 0
Benchmarking on platform: { "platform": "NVIDIA CUDA", "device": "GeForce GTX 560", "version": "OpenCL 1.1 CUDA" }
Preparing DAG...
  ℹ  21:57:41|eth  Loading from libethash...
  ℹ  21:57:41|eth  Done loading.
Warming up...
  ℹ  21:57:41|gpuminer0  workLoop 0 #00000000… #00000000…
Using platform: NVIDIA CUDA
Using device: GeForce GTX 560(OpenCL 1.1 CUDA)
  ✘  21:57:42|gpuminer0  Error GPU mining: clEnqueueMapBuffer ( -4 )
Trial 1... 0
Trial 2... 0
Trial 3... 0
Trial 4... 0
Trial 5... 0
min/mean/max: 0/0/0 H/s
inner mean: 0 H/s
Phoning home to find world ranking...
Error phoning home. ET is sad.
```

awrelll:
I'm having a similar issue 

```
 ℹ  10:27:49|ethminer  Getting work package...
  ℹ  10:28:04|ethminer  Loading from libethash...
  ℹ  10:28:04|ethminer  Done loading.
  ℹ  10:28:04|ethminer  Getting work package...
  ℹ  10:28:14|ethminer  Loading from libethash...
  ℹ  10:28:14|ethminer  Done loading.
  ℹ  10:28:14|ethminer  Got work package:
  ℹ  10:28:14|ethminer    Header-hash: c569cedde3bf8a6133744f153da4c89dea5f6c6e7776cc9be77d6e74689fc132
  ℹ  10:28:14|ethminer    Seedhash: 9b1ca7a86d339a3f175e48937848ac1cc012678e046b8c433a226d06f0f9bf9b
  ℹ  10:28:14|ethminer    Target: 0000000072c7dddfbb5390fd64d16dbd48050b82e9dbdd702c41cfa889460b34
  ℹ  10:28:14|gpuminer0  workLoop 0 #00000000… #9b1ca7a8…
Using platform: AMD Accelerated Parallel Processing
Using device: Tahiti(OpenCL 1.2 AMD-APP (1445.5))
[...]

PR #2040, opened 2015-05-29
Chunk upload of DAG for older AMD cards (OpenCL < 2.0)

### LINKED ISSUE #2010: CL_INVALID_BUFFER_SIZE and CL_INVALID_MEM_OBJECT for OpenCL 1.2 cards
I'm getting mining errors:

```
gpuminer0  Error GPU mining: clCreateBuffer ( -61 ) // CL_INVALID_BUFFER_SIZE
gpuminer0  workLoop .....
gpuminer0  Error GPU mining: clEnqueueWriteBuffer ( -38 ) // CL_INVALID_MEM_OBJECT
```

This has been confirmed on both Radeon Tahiti and Cayman platforms. At least in the Cayman case, this is a dual-core card. But I know of other dual-core cards that work, for example 290x2, so it is maybe the OpenCL 1.2 that is the issue - or the combination of the two.

Does anyone know of things to try? I'm happy playing around with the code.

See also #1982

kejace:
This fixes #2010 and potentially #1982. All credit to @sontol for writing the updated code. I just made an ugly merge.

How to use: There's a `#define chunks` at the top. Put it to true and re-compile. Then copy the *.cl file to your binary folder.

`chunks` should probably be merged into the `CL_VERSION_1_2` flag.

The updated .cl file has code duplicated as `...hash_chunks()` etc. This can be refactored, but I kept it for more convenient compilation.


subtly:
Is the 512MB a parameter which can be queried from opencl?

gavofyork:
@LefterisJP please take a look and refactor as required

LefterisJP:
Will do, but can't "properly" test since I have no such AMD card.

Gustav-Simonsson:
@subtly The max allocation size for a device can be queried with https://www.khronos.org/registry/cl/sdk/1.2/docs/man/xhtml/clGetDeviceInfo.html  and the  CL_DEVICE_MAX_MEM_ALLOC_SIZE 
@LefterisJP suggested we can use this to calculate optimal chunk number.

PR #2048, opened 2015-06-01
PV61 hash downloading fixes

arkpar:


libethereum/EthereumHost.cpp
```diff
@@ -91,7 +91,10 @@ void EthereumHost::doWork()
 	bool netChange = ensureInitialised();
 	auto h = m_chain.currentHash();
 	// If we've finished our initial sync (including getting all the blocks into the chain so as to reduce invalid transactions), start trading transactions & blocks
-	if (!isSyncing() && m_chain.isKnown(m_latestBlockSent))
+	bool syncing = false;
```
gavofyork: isSyncing() is guarded by x_sync anyway. why twice?

libethereum/EthereumHost.cpp
```diff
@@ -470,8 +444,7 @@ void EthereumHost::onPeerBlocks(EthereumPeer* _peer, RLP const& _r)
 
 void EthereumHost::onPeerNewHashes(EthereumPeer* _peer, h256s const& _hashes)
 {
-	Guard l(x_sync);
-	if (_peer->m_asking != Asking::Nothing)
+	if (isSyncing())
```
gavofyork: no guard of `x_sync` here?

libethereum/EthereumHost.cpp
```diff
@@ -91,7 +91,10 @@ void EthereumHost::doWork()
 	bool netChange = ensureInitialised();
 	auto h = m_chain.currentHash();
 	// If we've finished our initial sync (including getting all the blocks into the chain so as to reduce invalid transactions), start trading transactions & blocks
-	if (!isSyncing() && m_chain.isKnown(m_latestBlockSent))
+	bool syncing = false;
+	DEV_GUARDED(x_sync)
```
gavofyork: why move x_sync guard out of isSyncing?

gavofyork:
need to explain this patch a little more - i don't see the need to remove x_sync from isSyncing.

libethereum/EthereumHost.h
```diff
@@ -94,6 +92,7 @@ class EthereumHost: public p2p::HostCapability<EthereumPeer>, Worker
 	std::pair<std::vector<std::shared_ptr<EthereumPeer>>, std::vector<std::shared_ptr<EthereumPeer>>> randomSelection(unsigned _percent = 25, std::function<bool(EthereumPeer*)> const& _allow = [](EthereumPeer const*){ return true; });
 	void forEachPeerPtr(std::function<void(std::shared_ptr<EthereumPeer>)> const& _f) const;
 	void forEachPeer(std::function<void(EthereumPeer*)> const& _f) const;
+	bool isSyncingInternal() const;
```
gavofyork: best renamed to `isSyncing_UNSAFE` as is used in other classes (TransactionQueue/BlockQueue, iirc)

arkpar:
Added some comments to isSyncing_UNSAFE

libethereum/EthereumPeer.cpp
```diff
@@ -88,8 +86,7 @@ void EthereumPeer::setIdle()
 
 void EthereumPeer::requestStatus()
 {
-	if (m_asking != Asking::Nothing)
-	clog(NetWarn) << "Bad state: requesting state should be the first action";
+	assert(m_asking == Asking::Nothing);
```
gavofyork: this is /very/ harsh; why bother crashing the app in the case it's wrong rather than attempting to recover?

arkpar: This is programming invariant, not a runtime error we can recover from. requestStatus is private and called from constructor only. If assert fails here this is bad programming, and better Fail Fast

gavofyork: OK, yes, that's true here.

libethereum/EthereumPeer.cpp
```diff
@@ -106,22 +103,22 @@ void EthereumPeer::requestStatus()
 
 void EthereumPeer::requestHashes()
 {
-	if (m_asking == Asking::Blocks)
-		return;
+	assert(m_asking == Asking::Nothing);
```
gavofyork: this is /very/ harsh; why bother crashing the app in the case, rather than attempting to recover?

gavofyork:
will merge anyway, though i'd prefer all *dynamic* state assumptions/requirements to be recoverable.

PR #2049, opened 2015-06-01
exception in destructor

gluk256:
Tests crash in my Windows environment because of exception in TransientDirectory destructor. Apparently, sleeping for approximately 10 ms solves the problem, but exceptions in destructor should be avoided.

### DIFF
--- a/libdevcore/TransientDirectory.cpp
+++ b/libdevcore/TransientDirectory.cpp
@@ -52,7 +52,7 @@ TransientDirectory::~TransientDirectory()
 		}
 		catch (...)
 		{
-			std::this_thread::sleep_for(chrono::milliseconds(10));
+			this_thread::sleep_for(chrono::milliseconds(10));
 		}
 	}
 }
--- a/libdevcore/TransientDirectory.cpp
+++ b/libdevcore/TransientDirectory.cpp
@@ -19,6 +19,7 @@
  * @date 2015
  */
 
+#include <thread>
 #include <boost/filesystem.hpp>
 #include "Exceptions.h"
 #include "TransientDirectory.h"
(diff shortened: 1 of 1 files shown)


gluk256:
My previous PR #2047 should fix the Windows build.

libdevcore/TransientDirectory.cpp
```diff
@@ -42,5 +43,16 @@ TransientDirectory::TransientDirectory(std::string const& _path):
...
+	for (int i = 0; i < 3; ++i)
+	{
+		try
+		{
+			boost::filesystem::remove_all(m_path);
+			break;
+		}
+		catch (...)
+		{
+			std::this_thread::sleep_for(chrono::milliseconds(10));
```
LianaHus: you could remove "std::" here

chfast:
I don't like it. It just masks the problem.

Why is remove_all throwing in the fist place?
Can we leave the garbage if all 3 tries failed?

LefterisJP:
We were debugging the issue with Vlad. It's quite tricky. It's already in the destruction path due to having compiled without -DFATDB.

The whole thing runs in a single thread, yet still `remove_all()` throws with an exception that does not contain any information and it succeeds only after a little time passes. The solution Vlad proposes is definitely a hack and won't really solve anything, but we have not yet found the reason why this throws. The first assumption one makes is that something else has a windows handle to the directory and it releases after a bit of time. The problem is that as mentioned above, the whole test is single threaded.

Must be a quite obscure Windows specific thing. Further investigation is needed before any merging.

chfast:
Please at least include a comment about that in the code.

LefterisJP:
Yes a comment will be included, for sure. Funny thing is that it's probably the Mcaffee anti-virus scanner, trying to scan the new directory, and the tests trying to delete it at the same time. Fun times ...

chfast:
And you also have a non-throwing version `remove_all(const path& p, system::error_code& ec)`.

LefterisJP:
yeah I know, just depends on the way Vlad wants to do the error checking. He will get onto this PR soon.

gluk256:
Never mind the Windows build -- it was fixed in the previous PR #2047, which already was merged.

PR #2050, opened 2015-06-01
OpenCL Amd chunk dag upload

LefterisJP:


LefterisJP:
Continuation from #2040 so that we can quickly add any required changes

subtly:
I don't think this is the right solution and we must be super-careful not to introduce branches or increase the complexity of this code because it's already full of issues.

OpenCL can be queried for this parameter. If it can't be queried, then user should supply the size of the chunk and not a binary enable/disable. It'll be different for different cards -- as will the upper limit, which we should also be checking for and properly informing the user.

LefterisJP:
I also feel uneasy with this code due to lack of OpenCL knowledge and inability to actually test anything. I was simply asked to review and refactor.

The chunk size could be given by the user, but this is simply a refactoring of the already opened PR by @kejace where it's either 1 or 4. For other sizes, the kernels should also be modified to accept some form of array of buffers and not a given number as it is now. Plus some other functions introduced in the PR. If that is done we would not have a different `hash` and a different `search` kernel for when we have a chunk or not.

But to be honest I really can't say if what you suggest is even possible. We would need to delve deeper into OpenCL documentation to see if there is some magic parameter that would tell us the optimal chunk size for each different GPU. The function the author of the PR uses [`enqueueUnmapMemObject()`](https://www10.informatik.uni-erlangen.de/Teaching/Courses/SS2011/HESP/exercises/doc-opencl-cpp-1.0/html/classcl_1_1CommandQueue.html#9e972e773a4771bd0a4a65c8585d7245) does not seem to suggest something like that.

But as I said above, if we do indeed go that way the cl kernel code will also need rewritting.

So what do we do with this PR at the moment? Investigate further and act only when we are more certain?

Can spend more time looking at the whole

gavofyork:
If it's not the most perfect code ever, but it's strictly more functional and no worse than the existing code then it can be merged.

subtly:
"We would need to delve deeper into OpenCL documentation to see if there is some magic parameter that would tell us the optimal chunk size for each different GPU."

https://www.khronos.org/registry/cl/sdk/1.0/docs/man/xhtml/clGetDeviceInfo.html

There are also some nice parameters here that'll allow us to bail before we even attempt to allocate the DAG slab. Thus allowing us to bisect some of the incoming GPU issues. We can return an error ahead of time knowing that the GPU doesn't have enough memory or can't allocate enough memory. (And/or we don't have a kernel generated for CL_DEVICE_MAX_MEM_ALLOC_SIZE)

LefterisJP:
Yes I will look further into it. PErhaps it's better to spend some days researching, looking at professional OpenCL code and then write somethign that works for "all platforms".

gavofyork:
I'm happy for you to do that, however, given the imminent release, expediency is valuable. Getting this in, I think, will help a fair few people. If it's no worse (read: crashy) than before, then it should go in sooner rather than later, regardless of whether it's the best architected code. Just make sure there are notes wherever it's flakey or dangerous.

LefterisJP:
This code is just hastily made and untested. Refactoring someone else's code which is supposed to solve an issue for a setup I don't have, can't replicate and can't test is a recipe for disaster.

I can't properly test it because I don't have the kind of setup for which this PR would alleviate issues. If you would like to merge it now and then we can refactor on top of it.

But I am afraid that this will introduce more problems than it would solve.

libethash-cl/ethash_cl_miner.cpp
```diff
@@ -126,103 +127,166 @@ void ethash_cl_miner::finish()
 {
 	if (m_queue())
 		m_queue.finish();
+
+	if (m_dagChunks)
+		delete [] m_dagChunks;
```
arkpar: If  finish can be called twice m_dagChunks should be set to nullptr here. Right now finsh is public, so I guess this is possible

arkpar: Why not just use vector?

arkpar:
Looksgood  to me

gavofyork:
rebase?

LefterisJP:
Rebase would have proved trickier due to number of conflicts. I simply merged develop into the branch and addressed conflicts only once.

Gustav-Simonsson:
Got a 280x this weekend and just tried this branch. Running benchmark gives clCreateKernel error -46 (CL_INVALID_KERNEL_NAME) so there is probably some merge funkiness or some such going on.

LefterisJP:
Working with Gustav on it. Looking into it.

Gustav-Simonsson:
Got it running with chunks now on my 280x. It runs, but benchmarking gives only 59% of the hash rate of develop branch.

Moreover, in a more scientific observation, I noticed that benchmark trial in develop branch brings my GPU up to around 75 C but with chunking enabled it only goes up to about 65 C.

Clearly not hot enough. We need to work more on this.

LefterisJP:
The breaking issues are solved. I can also confirm that my GPU with using the 4 chunks gives around 70% of the hashrate I get when not using them.

If this PR is useful for someone for enabling them to GPU mine (because their card needs chunks) then this should be merged.

But next PR should be on making chunk number dynamic, finding more optimal chunk number, and make it GPU dependant.

Gustav-Simonsson:
Btw one guy tested with Radeon HD 6770M and it has max alloc of 134MB.  So to support such cards we'd need to either make the chunking even more granular or dynamic based on the device's max alloc value.

LefterisJP:
I would go with dynamic based on the queried values.

PR #2053, opened 2015-06-01
Reduced stack usage for nested calls

arkpar:


chfast:
Are you sure there are ever more than one Executive on the stack? Every Executive creates new instance of ExtVM and therefore a new stack for Executives.

libethereum/ExtVM.cpp
```diff
@@ -45,13 +55,23 @@ h160 ExtVM::create(u256 _endowment, u256& io_gas, bytesConstRef _code, OnOpFunc
...
+		}
+		io_gas = e.endGas();
+		m_callStack.pop();
+		return e.newAddress();
+	}
+	catch(...)
 	{
-		e.go(_onOp);
-		e.accrueSubState(sub);
+		m_callStack.pop();
```
chriseth: Why not use a scope guard here? If `e.newAddress` throws, the stack is popped twice.

arkpar:
Closing this since it does not completely help. OSX stil runs out of stack in some cases. Call flattening is a way to go

PR #2055, opened 2015-06-01
BlockWeight: uncle POW issue

winsvega:


test/libethereum/blockchain.cpp
```diff
@@ -214,9 +214,12 @@ void doBlockchainTests(json_spirit::mValue& _v, bool _fillin)
...
 				}
 
-				// update unclehash in case of invalid uncles
-				current_BlockHeader.sha3Uncles = sha3(uncleStream.out());
-				updatePoW(current_BlockHeader);
+				if (vBiUncles.size())
+				{
+					// update unclehash in case of invalid uncles
+					current_BlockHeader.sha3Uncles = sha3(uncleStream.out());
+					updatePoW(current_BlockHeader);
```
subtly: updatePoW was always called in previous version. Does it no longer need to be called?

CJentzsch: Its just an optimization. In the case we dont have uncles (vBiUncles.size() == 0), the value on `current_BlockHeader.sha3Uncles` is already an empty list and does not to be overwritten again with the same value. And since the header does not change, the PoW does not need to be done again. In the case of uncles, we nee to do both. Before this change we always did update `current_BlockHeader.sha3Uncles` and PoW, which was waste of time.

PR #2065, opened 2015-06-02
Boost mutiprecision number comparison bug workaround 2

chfast:
I add fixed version of boost/multiprecision/detail/number_compare.hpp to sources. It is injected with #included if buggy version of boost detected.


LefterisJP:
Tested it with 1.58.0. It works like a charm. So we are basically introducing the same change that will get into the next release of boost if the user has the "infected" boost version.

Only one question. We are sure that the bug exists only in 1.58.0, right?

chfast:
I checked boost 1.57 and it was ok.

PR #2066, opened 2015-06-02
Remove dead assignments in eth/main.cpp

LefterisJP:
Clang static analyzer spotted those. Removing.

### DIFF
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -774,11 +774,9 @@ int main(int argc, char** argv)
 		string l;
 		while (!g_exit)
 		{
-			silence = false;
 			cout << logbuf << "Press Enter" << flush;
 			std::getline(cin, l);
 			logbuf.clear();
-			silence = true;
 
 #if ETH_READLINE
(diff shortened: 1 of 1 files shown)


chriseth:
Could you explain why this change is OK?

LefterisJP:
Well, after deeper inspection it seems that you may be right. This is ... quite ugly. It did not seem like a global variable since it was just a part of main and it is not used further downn in main.

But ... there is a lambda function that seems to capture it by reference
```c++
g_logPost = [&](std::string const& a, char const*){
		if (silence)
			logbuf += a + "\n";
		else
			cout << "\r           \r" << a << endl << additional << flush;
	};
```

I guess then this PR is invalid.

Still ... this is ugly. At least a comment to explain why silence is assigned there would have been nice. Another comment or renaming of the variable to indicate it's global would also be nice.

LefterisJP:
In order to not confuse future readers of this code: https://github.com/ethereum/cpp-ethereum/pull/2068

Issue #2060, opened 2015-06-02
[mix] uint256 parameter in function call


debris:
I was unable to make a function call with `uint256` value larger than 287.

![screen shot 2015-06-02 at 09 57 44](https://cloud.githubusercontent.com/assets/2613714/7931095/26b855fa-090e-11e5-9974-101db05d9871.png)

PR #2067, opened 2015-06-02
Mix - bug fix: #2060

yann300:


mix/QBigInt.cpp
```diff
@@ -65,7 +65,9 @@ QVariantMap QBigInt::checkAgainst(QString const& _type) const
 	QString capacity = type.replace("uint", "").replace("int", "");
 	if (capacity.isEmpty())
 		capacity = "256";
-	bigint range = 256^(capacity.toInt() / 8);
```
arkpar: This should also work: bigint range = bigint(256)^(capacity.toInt() / 8);

yann300: not working...

PR #2068, opened 2015-06-02
Clarify usage of silence in eth/main

LefterisJP:
A comment is needed to specify to the reader that silence is actually a
global variable. The code with g_logPost() in general could use refactoring but that's left for later

### DIFF
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -66,6 +66,8 @@ using namespace dev::eth;
 using namespace boost::algorithm;
 using dev::eth::Instruction;
 
+static bool g_silence = false;
+
 void interactiveHelp()
 {
 	cout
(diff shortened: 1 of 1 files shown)


gavofyork:
`silence` is local, not global.

LefterisJP:
But it's actually used globally. It's captured by reference by `g_logPost()` which itself is used globally. Anyone reading the code (I know I did this mistake) will think that silence being assigned to false and then to true at [line 777](https://github.com/ethereum/cpp-ethereum/blob/develop/eth/main.cpp#L777) is some kind of mistake exactly because they would think that silence is local.

I was lead to this by the clang static analyzer thinking the same thing. It tagged it as a dead assignment, and I actually tried to remove it [here](https://github.com/ethereum/cpp-ethereum/pull/2066). We owe it to ourselves and to future maintainers of the code to clarify its usage.

gavofyork:
A well placed comment at the point of assignment should suffice. If you're absolutely desperate to change the name, then actually move the declaration global.

Just don't name a local variable `g_`...

LefterisJP:
Made it global

PR #2072, opened 2015-06-03
PoC: CALL execution stack offloading

chfast:
The strategy behind is simple: instead of using system stack for all calls and creates a new heap powered stack space is allocated for every call and create.

This uses Boost.Context from version 1.55. I know the API has changes in later version. Moreover, the implementation is heavy architecture dependent (different for x86, x64, Windows, Unix).

I'm not sure it is the best possible long term solution but let's test is for some time. Currently stacks has 16KB of size. That's what I need on linux/debug version.

Argument passing between contexts, calls, executives and others is a mess. It need to be refactored later on.

### DIFF
--- a/cmake/EthDependencies.cmake
+++ b/cmake/EthDependencies.cmake
@@ -207,7 +207,7 @@ elseif (UNIX)
 
 endif()
 
-find_package(Boost 1.54.0 REQUIRED COMPONENTS thread date_time system regex chrono filesystem unit_test_framework program_options)
+find_package(Boost 1.54.0 REQUIRED COMPONENTS thread date_time system regex chrono filesystem unit_test_framework program_options context)
 
 message(" - boost header: ${Boost_INCLUDE_DIRS}")
 message(" - boost lib   : ${Boost_LIBRARIES}")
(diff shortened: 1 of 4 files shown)


LefterisJP:
The windows builder seems [unable](https://build.ethdev.com/builders/Windows%20C%2B%2B%20pull%20requests/builds/2017/steps/configure/logs/stdio) to find that boost library. We would need to add it to the dependencies somehow.

You also seem to have triggered some warning for the [OSX evmjit build](https://build.ethdev.com/builders/OSX%20C%2B%2B%20evmjit%20pull%20requests/builds/1423/steps/compile/logs/warnings%20%288%29) due to unused variables.

chriseth:
I'm really not sure boost::context is a good idea. Can't we just use a heap-allocated vector to keep track of our stack variables?

chriseth:
Perhaps it would already help to create Executive via a shared_ptr on the heap?

gavofyork:
If the implementation and interface for boost::context is indeed so flakey (and still dependent on hardcoding stack size), I'm tempted to go with a heavier refactoring to get a stack of Executives on the heap as per the original design I proposed.

chfast:
I'm not a big fun of architecture dependent context switching from Boost.Context. On the other hand that approach makes that OS stack memory usage does not depend on CALL level.

My investigation showed that each execution level needs ~4000B of stack memory. Executive itself is only ~600B and creating it on heap do not give us as much as we need. The biggest stack memory consumer is the VM.

@chriseth, by suggesting shared_ptr do you mean an instance of Executive should be reused between calls?

@gavofyork, what do you mean by "original design I proposed"? The case where a VM is suspended by throwing an exception before doing nested call and resumed after the nested call is finished? That model will not suite JIT very well. Throwing exception from or through jitted code is currently impossible. I would need to use some other tricks similarly to what Boost.Context does.

gavofyork:
rebase and fix broken platforms?

chfast:
I think it will not be needed. Classic refactoring I did should keep stack usage below 4MB.

PR #2073, opened 2015-06-03
Check if GPU has over 2GB of memory

LefterisJP:
If the user's machine contains no OpenCL device that has enough GPU
memory then don't allow creation of a GPU miner. Instead default to CPU.

Keep in mind that this is the first in hopefully a series of many PRs where we will improve the GPU miner.

### DIFF
--- a/libethash-cl/ethash_cl_miner.cpp
+++ b/libethash-cl/ethash_cl_miner.cpp
@@ -35,6 +35,7 @@
 #include "ethash_cl_miner_kernel.h"
 
 #define ETHASH_BYTES 32
+#define ETHASH_CL_MINIMUM_MEMORY 2000000000
 
 // workaround lame platforms
 #if !CL_VERSION_1_2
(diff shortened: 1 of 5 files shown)


LefterisJP:
Hmm ... seems the buildbots get build errors with this. Looking into it.

libethash-cl/ethash_cl_miner.cpp
```diff
@@ -35,6 +35,7 @@
 #include "ethash_cl_miner_kernel.h"
 
 #define ETHASH_BYTES 32
+#define ETHASH_CL_MINIMUM_MEMORY 2000000000
```
chriseth: So it is 2 gigabytes and not 2 gibibytes?

libethash-cl/ethash_cl_miner.cpp
```diff
@@ -47,6 +48,9 @@
 
 using namespace std;
 
+// TODO: If at any point we can use libdevcore in here then we should switch to using a LogChannel
+#define ETHCL_LOG(_contents) cout << "[OPENCL]:" << _contents << endl
```
chriseth: Brackets? `<<` has quite low precedence.

PR #2084, opened 2015-06-03
Compute constants

chriseth:


libevmasm/ConstantOptimiser.cpp
```diff
@@ -0,0 +1,225 @@
...
+		eth::Instruction::MLOAD, // back up memory
+		u256(32),
+		AssemblyItem(PushData, u256(1) << 16), // has to be replaced
+		eth::Instruction::DUP4,
+		eth::Instruction::CODECOPY,
+		eth::Instruction::DUP2,
+		eth::Instruction::MLOAD,
+		eth::Instruction::SWAP2,
+		eth::Instruction::MSTORE
+	}
```
arkpar: Does not compile with msvc

chriseth:
Fixed (hopefully).

chriseth:
Failures seem to be test infrastructure issues.

libevmasm/ConstantOptimiser.cpp
```diff
@@ -0,0 +1,225 @@
...
+	m_copyRoutine[4] = _assembly.newData(data);
+	replaceConstants(_items, m_copyRoutine);
+}
+
+AssemblyItems ComputeMethod::findRepresentation(u256 const& _value)
+{
+	if (_value < 0x10000)
+		// Very small value, not worth computing
+		return AssemblyItems{_value};
+	else if (dev::bytesRequired(~_value) < dev::bytesRequired(_value))
```
LianaHus: you could remove dev:: here

chriseth: Actually not, because there is a method called `bytesRequired`. Removed some other namespace prefixes, though.

PR #2088, opened 2015-06-04
Net fixes: rlp exception in handshake, peer disconnect during read

subtly:


subtly:
Malformed hello packet causes segfault.

PR #2089, opened 2015-06-04
Issues: VMTest filler fix

winsvega:


test/libevm/vm.h
```diff
@@ -80,7 +80,7 @@ class FakeExtVM: public eth::ExtVMFace
 	eth::Transactions callcreates;
 	bytes thisTxData;
 	bytes thisTxCode;
-	u256 gas;
+	u256 gas, execGas;
```
CJentzsch: only one declaration per line (https://github.com/ethereum/cpp-ethereum/blob/develop/CodingStandards.txt#L110)

PR #2091, opened 2015-06-04
fixed peer test

gluk256:
with a bit of refactoring

### DIFF
--- a/libdevcore/Log.h
+++ b/libdevcore/Log.h
@@ -59,6 +59,14 @@ extern std::function<void(std::string const&, char const*)> g_logPost;
 /// or equal to the currently output verbosity (g_logVerbosity).
 extern std::map<std::type_info const*, bool> g_logOverride;
 
+/// Temporary changes system's verbosity for specific function. Restores the old verbosity when function returns.
+struct VerbosityHolder
+{
+	VerbosityHolder(int _temporaryValue) : oldLogVerbosity(g_logVerbosity) { g_logVerbosity = _temporaryValue; }
+	~VerbosityHolder() { g_logVerbosity = oldLogVerbosity; }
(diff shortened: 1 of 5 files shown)


libdevcore/Log.h
```diff
@@ -59,6 +59,14 @@ extern std::function<void(std::string const&, char const*)> g_logPost;
 /// or equal to the currently output verbosity (g_logVerbosity).
 extern std::map<std::type_info const*, bool> g_logOverride;
 
+/// Temporary changes system's verbosity for specific function. Restores the old verbosity when function returns.
+struct VerbosityHolder
+{
+	VerbosityHolder(int _temporaryValue) : oldLogVerbosity(g_logVerbosity) { g_logVerbosity = _temporaryValue; }
```
subtly: space before colon

gluk256:
fixed

gavofyork:
fails on linux.

libdevcore/Log.h
```diff
@@ -59,6 +59,14 @@ extern std::function<void(std::string const&, char const*)> g_logPost;
 /// or equal to the currently output verbosity (g_logVerbosity).
 extern std::map<std::type_info const*, bool> g_logOverride;
 
+/// Temporary changes system's verbosity for specific function. Restores the old verbosity when function returns.
+struct VerbosityHolder
```
arkpar: This is not thread safe for general use. Please add a comment at least

gluk256:
The warnings have been fixed.
Concerning the VerbosityHolder -- the corresponding comment is added. However, I think this comment was redundant, since I have only refactored the existing code, and so the VerbosityHolder is going to be used exactly the same as g_logVerbosity variable was used before.

PR #2096, opened 2015-06-05
Fallback takes constant amount of gas, and send no gas with send.

chriseth:
Several interleaving changes:
 - send now returns bool (true if sending worked) and never throws
 - send provides zero gas
 - function selector in receiver is cheap on failure (works with gas stipend alone)


libsolidity/ExpressionCompiler.cpp
```diff
@@ -1111,19 +1117,27 @@ void ExpressionCompiler::appendExternalFunctionCall(
 	else
 		m_context << eth::Instruction::CALL;
 
-	//Propagate error condition (if CALL pushes 0 on stack).
-	m_context << eth::Instruction::ISZERO;
-	m_context.appendConditionalJumpTo(m_context.errorTag());
+	unsigned remainsSize = 1 + // contract address
+			_functionType.valueSet() +
```
LianaHus: intend

PR #2112, opened 2015-06-08
Pessimising moves and missing overrides

chfast:
Non functional change. Removes explicit and pessimising moves and adds missing override specifiers.

### DIFF
--- a/json_spirit/json_spirit_writer_template.h
+++ b/json_spirit/json_spirit_writer_template.h
@@ -25,13 +25,9 @@ namespace json_spirit
         return 'A' - 10 + ch;
     }
 
-#pragma GCC diagnostic push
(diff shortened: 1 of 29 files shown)


json_spirit/json_spirit_writer_template.h
```diff
@@ -25,13 +25,9 @@ namespace json_spirit
         return 'A' - 10 + ch;
     }
 
-#pragma GCC diagnostic push
```
chriseth: Was that change intended? This is an external library.

chriseth:
Ah, nice change! :-)

json_spirit/json_spirit_writer_template.h
chfast: I know. But we added macros to suppress warning about unused typedef. The same must be done for clang. I think it is better to remove that unused typedef.

Issue #1651, opened 2015-04-16
Problem with "bytes0" in storage.


chriseth:
Using a bytes0 type in storage (and writing to it) creates internal compiler errors.

gavofyork:
`bytes0` is not a valid type.

PR #2114, opened 2015-06-08
Sol removed bytes0

LianaHus:
for #1651

### DIFF
--- a/test/libsolidity/SolidityEndToEndTest.cpp
+++ b/test/libsolidity/SolidityEndToEndTest.cpp
@@ -567,7 +567,7 @@ BOOST_AUTO_TEST_CASE(strings)
 BOOST_AUTO_TEST_CASE(empty_string_on_stack)
 {
 	char const* sourceCode = "contract test {\n"
-							 "  function run(bytes0 empty, uint8 inp) returns(uint16 a, bytes0 b, bytes4 c) {\n"
+							 "  function run(string empty, uint8 inp) external returns(uint16 a, string b, bytes4 c) {\n"
 							 "    var x = \"abc\";\n"
 							 "    var y = \"\";\n"
 							 "    var z = inp;\n"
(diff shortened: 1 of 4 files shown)


test/libsolidity/SolidityEndToEndTest.cpp
```diff
@@ -564,20 +564,6 @@ BOOST_AUTO_TEST_CASE(strings)
 	BOOST_CHECK(callContractFunction("pipeThrough(bytes2,bool)", string("\0\x02", 2), true) == encodeArgs(string("\0\x2", 2), true));
 }
 
-BOOST_AUTO_TEST_CASE(empty_string_on_stack)
```
chriseth: This test should still work.

test/libsolidity/SolidityEndToEndTest.cpp
```diff
@@ -3786,25 +3772,25 @@ BOOST_AUTO_TEST_CASE(packed_storage_structs_delete)
 	BOOST_CHECK(m_state.storage(m_contractAddress).empty());
 }
 
-BOOST_AUTO_TEST_CASE(packed_storage_structs_with_bytes0)
+BOOST_AUTO_TEST_CASE(packed_storage_structs_with_empty_string)
 {
 	char const* sourceCode = R"(
-		contract C {
```
chriseth: This test does not fulfill its purpose anymore, I guess it can be removed.

test/libsolidity/SolidityNameAndTypeResolution.cpp
```diff
@@ -1673,7 +1672,7 @@ BOOST_AUTO_TEST_CASE(bytes0_array)
 			bytes0[] illegalArray;
```
chriseth: This test should also be removed.

test/libsolidity/SolidityEndToEndTest.cpp
LianaHus: yeah I know. Should this be string instead of bytes0?

PR #2123, opened 2015-06-09
VM refactor: more OS stack usage reductions

chfast:
- VM instruction pointer is now uint64_t.
- CALL params are placed on stack instead of heap. That reduces OS stack usage by ~30%.

### DIFF
--- a/libevm/VM.cpp
+++ b/libevm/VM.cpp
@@ -194,6 +194,14 @@ bytesConstRef VM::execImpl(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
 			i += _ext.code[i] - (size_t)Instruction::PUSH1 + 1;
 	}
 
+	auto verifyJumpDest = [](u256 const& _dest, std::vector<uint64_t> const& _validDests)
+	{
+		auto nextPC = static_cast<uint64_t>(_dest);
+		if (!std::binary_search(_validDests.begin(), _validDests.end(), nextPC) || _dest > std::numeric_limits<uint64_t>::max())
+			BOOST_THROW_EXCEPTION(BadJumpDestination());
+		return nextPC;
+	};
+
 	m_steps = 0;
 	for (auto nextPC = m_curPC + 1; true; m_curPC = nextPC, nextPC = m_curPC + 1, ++m_steps)
 	{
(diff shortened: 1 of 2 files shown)


libevm/VM.cpp
```diff
@@ -535,16 +531,16 @@ bytesConstRef VM::execImpl(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
 			m_stack.pop_back();
 			break;
 		case Instruction::JUMP:
-			nextPC = m_stack.back();
-			if (find(m_jumpDests.begin(), m_jumpDests.end(), (uint64_t)nextPC) == m_jumpDests.end() || nextPC > numeric_limits<uint64_t>::max() )
+			nextPC = static_cast<uint64_t>(m_stack.back());
+			if (find(m_jumpDests.begin(), m_jumpDests.end(), nextPC) == m_jumpDests.end() || m_stack.back() > numeric_limits<uint64_t>::max())
```
chriseth: I think you could use std::binary_search here.

libevm/VM.cpp
```diff
@@ -598,7 +594,7 @@ bytesConstRef VM::execImpl(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
 			break;
 		case Instruction::CREATE:
 		{
-			u256 endowment = m_stack.back();
+			auto& endowment = m_stack.back();
```
chriseth: const ref?

libevm/VM.cpp
```diff
@@ -614,16 +610,14 @@ bytesConstRef VM::execImpl(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
 		case Instruction::CALL:
 		case Instruction::CALLCODE:
 		{
-			if (!callParams)
-				callParams.reset(new CallParameters);
-
-			callParams->gas = m_stack.back();
+			CallParameters callParams;
```
chriseth: Interesting, I would have said this change takes more space on the stack...

chfast: Me too. But it is exactly the opposite.

libevm/VM.cpp
```diff
@@ -45,165 +45,161 @@ static array<InstructionMetric, 256> metrics()
...
 }
 
-bytesConstRef VM::execImpl(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
+void VM::checkRequirements(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp, Instruction _inst)
 {
-	// Reset leftovers from possible previous run
-	m_curPC = 0;
-	m_jumpDests.clear();
+	static const auto c_metrics = metrics();
+	auto& metric = c_metrics[static_cast<size_t>(_inst)];
```
chriseth: Why not cast to `uint8_t`?

chfast:
All issues fixed.

Issue #2110, opened 2015-06-08
Packet that can kill a cpp node


gavofyork:
We have found a packet that kills any cpp peer that we see
on the dashboard , it's RLP looks like this :

http://pastebin.com/raw.php?i=XJFgjacf

and btw, the guy who found it asking if his help deserves
any kind of bonus... you know - not a consensus break but
also a good help.

CJentzsch:
I tried to decode the RLP and it is throwing an exception at: https://github.com/ethereum/cpp-ethereum/blob/develop/libdevcore/RLP.cpp#L35. So this RLP should behave excactly as all other invalid RLP, nothing special with it. So looks like somewhere in the network code we are missing to catch invalid RLP exception in general.

romanman:
I guess one try catch block will solve it,
so good to know you are aware of the
problem

On Mon, Jun 8, 2015 at 9:42 AM, Christoph Jentzsch <notifications@github.com

chriseth:
I guess this was fixed by #2129. If not, please reopen.

PR #2129, opened 2015-06-09
Catch exception for maformed rlp received by node table.

subtly:


CJentzsch:
is this related to this bug: https://github.com/ethereum/cpp-ethereum/issues/2110 ?

subtly:
@CJentzsch yes

PR #2132, opened 2015-06-09
Refactoring around OpenCL DAG chunks usage

LefterisJP:
* Basically removing --use-chunks, and figuring out whether or not to use them by queruing the max allocateable memory of the device.

* Refactoring of some stuff in cl_miner that can be abstracted in a better way

* Style fixes


LefterisJP:
Another chunk related PR will follow. In the next one we will hopefully tackle having a chunk amount calculated and used dynamically.

PR #2134, opened 2015-06-09
Executive refactor

chfast:


gavofyork:
why switch shared_ptr to unique_ptr?

chfast:
Because ExtVM instance is never shared. Executive is the only owner.

gavofyork:
sure, and indeed given *that*, they're essentially equivalent here, the only difference being that shared_ptr doesn't require the ExtVM to be fully declared prior to a usage of the safe pointer.

chfast:
Right. Another option is to move Executive destructor to .cpp. Both solutions are a bit misleading for me. shared_ptr suggests something is shared but is not. Destructor in .cpp suggests there is something specific in there, but is not.

Tell me what to do.

gavofyork:
shared_ptr with a short comment about why.

chfast:
Done.

PR #2144, opened 2015-06-10
Cleanup of SecretStore.

chriseth:
Added documentation, more failure checking and some general cleanup.

### DIFF
--- a/libdevcore/CommonIO.cpp
+++ b/libdevcore/CommonIO.cpp
@@ -95,9 +95,11 @@ string dev::contentsString(string const& _file)
 	return contentsGeneric<string>(_file);
 }
 
-void dev::writeFile(std::string const& _file, bytesConstRef _data)
+bool dev::writeFile(std::string const& _file, bytesConstRef _data)
 {
-	ofstream(_file, ios::trunc|ios::binary).write((char const*)_data.data(), _data.size());
+	ofstream s(_file, ios::trunc | ios::binary);
+	s.write(reinterpret_cast<char const*>(_data.data()), _data.size());
+	return !!s;
 }
 
 std::string dev::getPassword(std::string const& _prompt)
--- a/libdevcore/CommonIO.cpp
+++ b/libdevcore/CommonIO.cpp
@@ -95,11 +95,12 @@ string dev::contentsString(string const& _file)
 	return contentsGeneric<string>(_file);
 }
 
-bool dev::writeFile(std::string const& _file, bytesConstRef _data)
+void dev::writeFile(std::string const& _file, bytesConstRef _data)
 {
 	ofstream s(_file, ios::trunc | ios::binary);
 	s.write(reinterpret_cast<char const*>(_data.data()), _data.size());
-	return !!s;
+	if (!s)
+		BOOST_THROW_EXCEPTION(FileError());
 }
 
 std::string dev::getPassword(std::string const& _prompt)
--- a/libdevcore/CommonIO.cpp
+++ b/libdevcore/CommonIO.cpp
@@ -23,13 +23,14 @@
 #include <iostream>
 #include <cstdlib>
 #include <fstream>
-#include "Exceptions.h"
 #include <stdio.h>
 #ifdef _WIN32
 #include <windows.h>
 #else
 #include <termios.h>
 #endif
+#include <boost/filesystem.hpp>
+#include "Exceptions.h"
 using namespace std;
 using namespace dev;
 
@@ -95,12 +96,23 @@ string dev::contentsString(string const& _file)
 	return contentsGeneric<string>(_file);
 }
 
-void dev::writeFile(std::string const& _file, bytesConstRef _data)
+void dev::writeFile(std::string const& _file, bytesConstRef _data, bool _writeDeleteRename)
 {
-	ofstream s(_file, ios::trunc | ios::binary);
-	s.write(reinterpret_cast<char const*>(_data.data()), _data.size());
-	if (!s)
-		BOOST_THROW_EXCEPTION(FileError());
+	if (_writeDeleteRename)
+	{
+		namespace fs = boost::filesystem;
+		fs::path tempPath = fs::unique_path(_file + "-%%%%%%");
+		writeFile(tempPath.string(), _data, false);
+		// will delete _file if it exists
+		fs::rename(tempPath, _file);
+	}
+	else
+	{
+		ofstream s(_file, ios::trunc | ios::binary);
+		s.write(reinterpret_cast<char const*>(_data.data()), _data.size());
+		if (!s)
+			BOOST_THROW_EXCEPTION(FileError());
+	}
 }
 
 std::string dev::getPassword(std::string const& _prompt)
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -178,15 +178,35 @@ bool dev::verify(Public const& _p, Signature const& _s, h256 const& _hash)
 bytes dev::pbkdf2(string const& _pass, bytes const& _salt, unsigned _iterations, unsigned _dkLen)
 {
 	bytes ret(_dkLen);
-	PKCS5_PBKDF2_HMAC<SHA256> pbkdf;
-	pbkdf.DeriveKey(ret.data(), ret.size(), 0, (byte*)_pass.data(), _pass.size(), _salt.data(), _salt.size(), _iterations);
+	if (PKCS5_PBKDF2_HMAC<SHA256>().DeriveKey(
+		ret.data(),
+		ret.size(),
+		0,
+		reinterpret_cast<byte const*>(_pass.data()),
+		_pass.size(),
+		_salt.data(),
+		_salt.size(),
+		_iterations
+	) != _iterations)
+		return bytes();
 	return ret;
 }
 
 bytes dev::scrypt(std::string const& _pass, bytes const& _salt, uint64_t _n, uint32_t _r, uint32_t _p, unsigned _dkLen)
 {
 	bytes ret(_dkLen);
-	libscrypt_scrypt((uint8_t const*)_pass.data(), _pass.size(), _salt.data(), _salt.size(), _n, _r, _p, ret.data(), ret.size());
+	if (libscrypt_scrypt(
+		reinterpret_cast<uint8_t const*>(_pass.data()),
+		_pass.size(),
+		_salt.data(),
+		_salt.size(),
+		_n,
+		_r,
+		_p,
+		ret.data(),
+		ret.size()
+	) != 0)
+		return bytes();
 	return ret;
 }
 
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -31,6 +31,7 @@
 #include <libdevcore/FileSystem.h>
 #include "AES.h"
 #include "CryptoPP.h"
+#include "Exceptions.h"
 using namespace std;
 using namespace dev;
 using namespace dev::crypto;
@@ -188,7 +189,7 @@ bytes dev::pbkdf2(string const& _pass, bytes const& _salt, unsigned _iterations,
 		_salt.size(),
 		_iterations
 	) != _iterations)
-		return bytes();
+		BOOST_THROW_EXCEPTION(CryptoException());
 	return ret;
 }
 
@@ -206,7 +207,7 @@ bytes dev::scrypt(std::string const& _pass, bytes const& _salt, uint64_t _n, uin
 		ret.data(),
 		ret.size()
 	) != 0)
-		return bytes();
+		BOOST_THROW_EXCEPTION(CryptoException());
 	return ret;
 }
 
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -254,42 +254,71 @@ h256 crypto::kdf(Secret const& _priv, h256 const& _hash)
 	return s;
 }
 
-h256 Nonce::get(bool _commit)
+mutex Nonce::s_x;
+
+h256 Nonce::get()
 {
 	// todo: atomic efface bit, periodic save, kdf, rr, rng
 	// todo: encrypt
-	static h256 s_seed;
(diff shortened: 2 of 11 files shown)


LianaHus:
compile errors

chriseth:
Compile errors should be fixed now.

libdevcore/CommonIO.cpp
```diff
@@ -119,9 +119,11 @@ string dev::contentsString(std::string const& _file)
 	return ret;
 }
 
-void dev::writeFile(std::string const& _file, bytesConstRef _data)
+bool dev::writeFile(std::string const& _file, bytesConstRef _data)
```
arkpar: why not throw exception?

chriseth:
I will change this to use exceptions.

chriseth:
Updated.

libdevcrypto/SecretStore.cpp
```diff
@@ -84,36 +85,38 @@ static js::mValue upgraded(std::string const& _s)
...
 }
 
-h128 SecretStore::importSecret(bytes const& _s, std::string const& _pass)
+h128 SecretStore::importSecret(bytes const& _s, string const& _pass)
 {
-	h128 r = h128::random();
-	m_cached[r] = _s;
-	m_keys[r] = make_pair(encrypt(_s, _pass), std::string());
-	save();
+	h128 r;
```
gavofyork: not better to throw in case difficulty to encrypt rather than return a zero uuid?

gavofyork:
also, 4 platforms failing?

chriseth:
I just wanted to see if I can get it to succeed the test here. We should probably have a second thought about how and when the nonce is written to the file and what to do in the failure case. Also note that it is written in the destructor, and we should catch the exception there.

chriseth:
First wanted to see if the last commit succeeds the tests. We should probably have a second thought about the nonce file. Also note that the None destructor may throw...

libdevcrypto/Exceptions.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include <libdevcore/Exceptions.h>
+
+namespace dev
+{
+namespace crypto
+{
+
+/// Rare malfunction of cryptographic functions.
+DEV_SIMPLE_EXCEPTION_RLP(CryptoException);
```
gavofyork: should be `DEV_SIMPLE_EXCEPTION`, no?

libdevcrypto/Common.cpp
```diff
@@ -259,12 +280,12 @@ h256 Nonce::get(bool _commit)
 			BOOST_THROW_EXCEPTION(InvalidState());
 		
 		// prevent seed reuse if process terminates abnormally
-		writeFile(s_seedFile, bytes());
+		try { writeFile(s_seedFile, bytes()); } catch (FileError const&) {}
```
subtly: As per comment on previous line, this needs to throw. If it crashes: perfect! There's no sense or use in hiding something like this.

libdevcrypto/Common.cpp
```diff
@@ -259,12 +280,12 @@ h256 Nonce::get(bool _commit)
...
 	}
 	h256 prev(s_seed);
 	sha3(prev.ref(), s_seed.ref());
 	if (_commit)
-		writeFile(s_seedFile, s_seed.asBytes());
+		try { writeFile(s_seedFile, s_seed.asBytes()); } catch (FileError const&) {}
```
subtly: Same as above. If this is throwing the user *must* know about it. At the very least we get a crash report.

libdevcrypto/SecretStore.cpp
```diff
@@ -183,62 +190,66 @@ h128 SecretStore::readKeyContent(std::string const& _content, std::string const&
...
 }
 
-std::string SecretStore::encrypt(bytes const& _v, std::string const& _pass, KDF _kdf)
+static bytes deriveNewKey(string const& _pass, KDF _kdf, js::mObject& o_ret)
 {
-	js::mObject ret;
-
-	// KDF info
 	unsigned dklen = 32;
+	unsigned iterations = 1 << 19;
```
subtly: 1 << 18

chriseth: Thanks!

libdevcore/CommonIO.cpp
```diff
@@ -121,7 +121,10 @@ string dev::contentsString(std::string const& _file)
 
 void dev::writeFile(std::string const& _file, bytesConstRef _data)
 {
-	ofstream(_file, ios::trunc|ios::binary).write((char const*)_data.data(), _data.size());
+	ofstream s(_file, ios::trunc | ios::binary);
+	s.write(reinterpret_cast<char const*>(_data.data()), _data.size());
+	if (!s)
```
CJentzsch: shouldn't that be checked before trying to write to the file?
Additionally using a try/catch block?

libdevcrypto/SecretStore.cpp
```diff
@@ -29,6 +29,7 @@
 #include <libdevcore/SHA3.h>
 #include <libdevcore/FileSystem.h>
 #include <test/JsonSpiritHeaders.h>
```
CJentzsch: > [Mittwoch, 17. Juni 2015 11:56:01 gavofyork] Yeah we should move over to libjsoncpp completely at some point. 
Not required for this PR, only nice to have.

libdevcrypto/SecretStore.cpp
```diff
@@ -84,35 +86,37 @@ static js::mValue upgraded(std::string const& _s)
 	return js::mValue();
 }
 
-SecretStore::SecretStore(std::string const& _path): m_path(_path)
+SecretStore::SecretStore(string const& _path): m_path(_path)
 {
-	load();
+	if (!m_path.empty())
```
CJentzsch: Shouldn't we throw otherwise, since load is necessary for other operations?

libdevcrypto/SecretStore.cpp
```diff
@@ -132,50 +136,53 @@ void SecretStore::clearCache() const
...
+			fs::remove(filename);
 	}
 }
 
-void SecretStore::load(std::string const& _keysPath)
+void SecretStore::load(string const& _keysPath)
 {
 	fs::path p(_keysPath);
-	boost::filesystem::create_directories(p);
+	fs::create_directories(p);
```
CJentzsch: needs a check whether it is successful or not.

libdevcrypto/SecretStore.h
```diff
@@ -35,41 +35,81 @@ enum class KDF {
...
+ * UUID of the key.
+ * @note that most of the functions here affect the filesystem and throw exceptions on failure,
+ * and they also throw exceptions upon rare malfunction in the cryptographic functions.
+ */
 class SecretStore
 {
 public:
+	/// Construct a new SecretStore and read all keys in the given directory.
 	SecretStore(std::string const& _path = defaultPath());
-	~SecretStore();
```
CJentzsch: why did you remove the destructor?

chriseth: I don't think it is necessary, or did I miss something?

libdevcore/CommonIO.cpp
chriseth: It is checked internally in `ofstream::write`. Furthermore, `ofstream` functions should not usually throw exceptions, they only set flags which can be checked by using e.g. `!s`.

libdevcrypto/SecretStore.cpp
chriseth: It is not strictly necessary. If no path is provided (which is not the case in real use of this class), it just does not store the keys permanently.

libdevcrypto/SecretStore.cpp
chriseth: It throws on failure.

libdevcrypto/SecretStore.cpp
subtly: Please remove this condition and let create_directories throw. I don't see how the condition is necessary; if it is, consider a ensureFileStore() method which calls create_directories and is called by constructor and save(). This check is inevitable so I can't see a reason to defer the call. If load() works here, then save() *should* work later. This class is useless if it can't ensure read/write to m_path; thus, a good idea to check/ensure in the constructor.

libdevcore/CommonIO.cpp
chfast: Not bad, but `if (s.fail())` would be more informative.

libdevcrypto/SecretStore.cpp
```diff
@@ -132,50 +136,53 @@ void SecretStore::clearCache() const
...
 		v["crypto"] = crypto;
 		v["id"] = uuid;
 		v["version"] = c_keyFileVersion;
 		writeFile(filename, js::write_string(js::mValue(v), true));
-		if (!k.second.second.empty() && k.second.second != filename)
-			boost::filesystem::remove(k.second.second);
-		k.second.second = filename;
+		swap(k.second.filename, filename);
+		if (!filename.empty() && !fs::equivalent(filename, k.second.filename))
+			fs::remove(filename);
```
subtly: Suggestion for future: boost vs write && rename on success. Some file systems act differently when fs errors occur (disk space, permissions, etc.). And unfortunately, due to default partitioning schemes, dual-boot scenarios, etc., running out of disk space is something which happens a *LOT* (among thousands of users). Doing something similar to vim would save someone in both of these situations: write-out _file.json, delete file.json, then rename _file.json to file.json. If something bad happens in the middle it's recoverable.

libdevcrypto/SecretStore.h
```diff
@@ -35,41 +35,81 @@ enum class KDF {
...
+	/// Store all keys in the managed directory.
 	void save() { save(m_path); }
 
+	/// @returns the default path for the managed directory.
 	static std::string defaultPath() { return getDataDir("web3") + "/keys"; }
 
 private:
+	struct EncryptedKey
+	{
+		std::string encryptedKey;
```
subtly: Curious. Why string and not bytes?

libdevcrypto/SecretStore.cpp
```diff
@@ -84,35 +86,37 @@ static js::mValue upgraded(std::string const& _s)
...
 	return key;
 }
 
-h128 SecretStore::importSecret(bytes const& _s, std::string const& _pass)
+h128 SecretStore::importSecret(bytes const& _s, string const& _pass)
 {
-	h128 r = h128::random();
+	h128 r;
+	EncryptedKey key{encrypt(_s, _pass), string()};
+	if (!key.encryptedKey.empty())
```
subtly: !empty() ??

libdevcrypto/SecretStore.cpp
```diff
@@ -84,35 +86,37 @@ static js::mValue upgraded(std::string const& _s)
...
+		BOOST_THROW_EXCEPTION(crypto::CryptoException());
```
subtly: Why isn't this thrown inside encrypt?

libdevcrypto/SecretStore.cpp
```diff
@@ -183,62 +190,66 @@ h128 SecretStore::readKeyContent(std::string const& _content, std::string const&
...
-bool SecretStore::recode(h128 const& _uuid, string const& _newPass, std::function<std::string()> const& _pass, KDF _kdf)
+bool SecretStore::recode(h128 const& _uuid, string const& _newPass, function<string()> const& _pass, KDF _kdf)
 {
-//	cdebug << "recode:" << toUUID(_uuid);
 	bytes s = secret(_uuid, _pass, true);
 	if (s.empty())
 		return false;
-	m_keys[_uuid].first = encrypt(s, _newPass, _kdf);
+	m_cached.erase(_uuid);
+	m_keys[_uuid].encryptedKey = encrypt(s, _newPass, _kdf);
```
subtly: If exception is not moved into encrypt then an exception is necessary here. If encrypt fails the key is lost.

libdevcrypto/SecretStore.h
chriseth: I don't think I changed this here.

libdevcrypto/SecretStore.cpp
chriseth: I guess I will also add some further tests for this class...

libdevcrypto/SecretStore.cpp
chriseth: Added that feature to `writeFile`.

chriseth:
Updated and added tests.

libdevcrypto/SecretStore.cpp
CJentzsch: You added an error handling in the `srypt` function in the form of an throwing exception. I think this is a good place to catch and handle it.

test/libdevcrypto/crypto.cpp
```diff
@@ -45,6 +49,7 @@ static CryptoPP::OID s_curveOID(CryptoPP::ASN1::secp256k1());
 static CryptoPP::DL_GroupParameters_EC<CryptoPP::ECP> s_params(s_curveOID);
 static CryptoPP::DL_GroupParameters_EC<CryptoPP::ECP>::EllipticCurve s_curve(s_params.GetCurve());
 
+
```
CJentzsch: uneeded free line

test/libdevcrypto/SecretStore.cpp
```diff
@@ -29,11 +29,16 @@
...
+#include <test/TestHelper.h>
+#include <test/TestUtils.h>
 using namespace std;
 using namespace dev;
+using namespace dev::test;
 
 namespace js = json_spirit;
+namespace fs = boost::filesystem;
+
+BOOST_GLOBAL_FIXTURE( MoveNonceToTempDir );
```
CJentzsch: extra `;` -> build failure

libdevcrypto/SecretStore.cpp
chriseth: The exceptions that are thrown there are actually kind of failed assertions, they should not really happen. What is more likely to happen are io errors and they are also not caught here.
So basically, this function returns false if the password was wrong and throws an exception in exceptional circumstances.

Issue #2141, opened 2015-06-10
AlethZero crashes importing blocks on Mac OS


Benoit9:
This is version is AlethZero/v0.9.24/Release/Darwin/clang/int
4e17c58938ae116cb642750401ee330623f236d3
CLEAN

It is a new install, downloading the chain from scratch.

The last imported block is 374395.

App downloads new blocks then crashes, never goes above 374395.

Here is the report: https://gist.github.com/Benoit9/6ebbe5054be7161373f2

Benoit9:
It seems to crash in a malloc. I launched again with all other apps closed, it still crashes. This is a 16GB machine.

arkpar:
Ran out of stack

Benoit9:
Related to this then: https://github.com/ethereum/cpp-ethereum/issues/2139 ?

Benoit9:
Looks similar: https://gist.github.com/Benoit9/3c22a09464f2db64c1a0
Is it a recursive call?

If it just a stack size limit, can I do anything to make it run?

chfast:
It is a duplicate of #1995. But you might be right that correct stack limit for OSX must be set.

Benoit9:
Thanks! Is there a way to clear the hurdle and to run AlethZero at all?

chfast:
I'm trying to set the correct stack limit on OSX, but it need to go through buildbot as I don't use OSX.

Benoit9:
I am not set up to build from source on OSX, but let me know if I can help with testing.

Benoit9:
I tried the buildbot AlethZero-OSX-20150611181529-0.9.25-61-3f94350.dmg version. No luck. Same crash.

chfast:
There is a new version of AlethZero available: AlethZero-OSX-20150613082031-0.9.25-61-41397c3.dmg
Can you test it and tell me if it works for you?

Benoit9:
Yes, working now. Thank you!

chfast:
Fixed by #2167.

PR #2152, opened 2015-06-11
Set stack memory limit on OSX

chfast:


Benoit9:
Does it apply to all threads? I still get a crash at the same call depth: https://github.com/ethereum/cpp-ethereum/issues/2141

chfast:
No. Only for main thread unfortunately.

Benoit9:
Thread #5 crashing.

PR #2165, opened 2015-06-12
KeyManager refactoring to increase readability.

chriseth:


libethcore/EthashAux.cpp
```diff
@@ -253,6 +253,6 @@ Ethash::Result EthashAux::eval(h256 const& _seedHash, h256 const& _headerHash, N
 			return dag->compute(_headerHash, _nonce);
 	DEV_IF_THROWS(return EthashAux::get()->light(_seedHash)->compute(_headerHash, _nonce))
 	{
-		return Ethash::Result{ ~h256(), h256() };
+		return Ethash::Result{ h256(), h256() };
```
gavofyork: explanation?

libethcore/KeyManager.cpp
```diff
@@ -131,19 +137,19 @@ std::string KeyManager::getPassword(h128 const& _uuid, function<std::string()> c
...
 		return it->second;
-	for (unsigned i = 0; i< 10; ++i)
+	for (unsigned i = 0; i < 10; ++i)
 	{
-		std::string p = _pass();
+		string p = _pass();
 		if (p.empty())
 			break;
-		if (hashPassword(p) == _passHash || _passHash == UnknownPassword)
+		if (hashPassword(p) == _passHash || !_passHash)
```
gavofyork: don't you think the former is more readable?

gavofyork: also, order of the two components should probably be swapped for efficiency.

libethcore/KeyManager.cpp
```diff
@@ -209,67 +215,92 @@ void KeyManager::kill(Address const& _a)
...
-	for (auto const& i: m_addrLookup)
-		if (m_keyInfo.count(i.second) > 0)
-			ret[i.first] = make_pair(m_keyInfo.count(i.second) ? m_keyInfo.at(i.second).info : "", m_keyInfo.count(i.second) && m_passwordInfo.count(m_keyInfo.at(i.second).passHash) ? m_passwordInfo.at(m_keyInfo.at(i.second).passHash) : "");
-	return ret;
+	return m_addrLookup.count(_address) && m_keyInfo.count(m_addrLookup.at(_address));
+}
+
+string const& KeyManager::accountName(Address const& _address) const
+{
+	try
```
gavofyork: use `DEV_IF_THROWS` instead?

libethcore/KeyManager.h
```diff
@@ -35,13 +35,13 @@ class PasswordUnknown: public Exception {};
 struct KeyInfo
 {
 	KeyInfo() = default;
-	KeyInfo(h256 const& _passHash, std::string const& _info): passHash(_passHash), info(_info) {}
+	KeyInfo(h256 const& _passHash, std::string const& _accountName): passHash(_passHash), accountName(_accountName) {}
 
-	h256 passHash;		///< Hash of the password or h256() if unknown.
```
gavofyork: style-wise, tabs before or after the line's code is fine. they're just not allowed within.

libweb3jsonrpc/AccountHolder.cpp
```diff
@@ -103,7 +103,8 @@ void AccountHolder::clearQueue(int _id)
 
 AddressHash SimpleAccountHolder::realAccounts() const
 {
-	return m_keyManager.accounts();
+	auto accounts = m_keyManager.accounts();
```
gavofyork: This snippet appears twice at least - better to use a utility function? also note you can do `AddressHash() + accounts`

libethcore/KeyManager.cpp
chriseth: `DEV_IF_THROWS` is currently only defined in `EthashAux.cpp`.
Apart from that, I think that a custom macro that assumes a trailing expression is not a good way to save three lines of code, given that anyone reading that code has to look up the macro at least once or - even worse - make false assumptions about its contents.

gavofyork:
style still wrong (tabs should be used for comments).

chriseth:
Updated.

PR #2170, opened 2015-06-12
 bloom filter

gluk256:
initial version

### DIFF
--- a/libwhisper/BloomFilter.cpp
+++ b/libwhisper/BloomFilter.cpp
@@ -25,48 +25,45 @@ using namespace std;
 using namespace dev;
 using namespace dev::shh;
 
-bool BloomFilter::matches(AbridgedTopic const& _t) const
-{
-	static unsigned const c_match = ~unsigned(0);
-	unsigned topic = AbridgedTopic::Arith(_t).convert_to<unsigned>();
-	unsigned matchingBits = m_filter & topic;
-	matchingBits |= ~topic;
-	return (c_match == matchingBits);
-}
+static unsigned const c_mask[] = { 1, 2, 4, 8, 16, 32, 64, 128 };
 
-void SharedBloomFilter::add(AbridgedTopic const& _t)
+void TopicBloomFilter::add(AbridgedTopic const& _h)
 {
-	unsigned const topic = AbridgedTopic::Arith(_t).convert_to<unsigned>();
-	m_filter |= topic;
-
-	unsigned nextBit = 1;
-	for (unsigned i = 0; i < ArrSize; ++i, nextBit <<= 1)
-		if (topic & nextBit)
-			if (m_refCounter[i] != numeric_limits<unsigned short>::max())
+	*this |= _h;
+	unsigned count = 0;
+	for (unsigned i = 0; i < CounterSize; ++i)
+		if (isBitSet(_h, i))
+		{
+			if (m_refCounter[i] != numeric_limits<uint16_t>::max())
 				m_refCounter[i]++;
 			//else: overflow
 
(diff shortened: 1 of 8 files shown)


gluk256:
It seems, the errors are not related to my changes.

libwhisper/BloomFilter.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+void TopicBloomFilter::addRaw(AbridgedTopic const& _h)
+{
+	*this |= _h;
+	for (unsigned i = 0; i < CounterSize; ++i)
+		if (isBitSet(_h, i))
+		{
+			if (m_refCounter[i] != numeric_limits<uint16_t>::max())
+				m_refCounter[i]++;
+			//else: overflow
```
subtly: Could an exception be thrown?

gluk256:
I don't think it is a severe error. First, it is very unlikely, that somebody will watch more than 65536 topics. It will only happen if somebody only adds topics to the filter, but does not remove them. It does not make any sense, though -- several hundreds of them will already fill the entire bloom filter with very high probability. 
But even in case of overflow, it will cause no immediate problems.The error will only arise after at least 65536 of them will be removed later. Since overflow only can happen if topics are never removed, even if overflow happens, it will probably not cause any problem in the future. That's why I decided to leave it like this.

chriseth:
Is this an attack vector a malicious DApp could use?

gluk256:
At the moment it should not be a problem.

gavofyork:
the only thing we can sign up to protecting users from regarding malicious dapps is the use of secrets. everything else is fair game for now.

libwhisper/BloomFilter.cpp
```diff
@@ -0,0 +1,67 @@
...
+void TopicBloomFilter::removeRaw(AbridgedTopic const& _h)
+{
+	for (unsigned i = 0; i < CounterSize; ++i)
+		if (isBitSet(_h, i))
+		{
+			if (m_refCounter[i])
+				m_refCounter[i]--;
+
+			if (!m_refCounter[i])
+				(*this)[i/8] &= ~c_mask[i%8];
```
gavofyork: whitespace. (`i / 8`, `i % 8`).

libwhisper/Common.cpp
```diff
@@ -84,6 +85,26 @@ bool TopicFilter::matches(Envelope const& _e) const
 	return false;
 }
 
+TopicFilter::TopicFilter(RLP const& _r)
+{
+	for (RLP i: _r)
```
gavofyork: `RLP const&`

libwhisper/Common.cpp
```diff
@@ -84,6 +85,26 @@ bool TopicFilter::matches(Envelope const& _e) const
...
+	{
+		m_topicMasks.push_back(TopicMask());
+		for (RLP j: i)
```
gavofyork: `RLP const&`

libwhisper/Common.cpp
```diff
@@ -84,6 +85,26 @@ bool TopicFilter::matches(Envelope const& _e) const
...
+			m_topicMasks.back().push_back(j.toPair<FixedHash<4>, FixedHash<4>>());
+	}
+}
+
+AbridgedTopic TopicFilter::exportBloom() const
+{
+	AbridgedTopic ret;
+	for (TopicMask const& t: m_topicMasks)
+		for (auto i: t)
```
gavofyork: `auto const&`

gavofyork:
whitespace and minor improvements to existing code.

PR #2176, opened 2015-06-13
ImportRoute class instead of pair

debris:


libethereum/Client.cpp
```diff
@@ -613,10 +613,11 @@ bool Client::submitWork(ProofOfWork::Solution const& _solution)
 
 void Client::syncBlockQueue()
 {
-	ImportRoute ir;
 	cwork << "BQ ==> CHAIN ==> STATE";
-	tie(ir.first, ir.second, m_syncBlockQueue) = m_bc.sync(m_bq, m_stateDB, rand() % 10 + 5);
-	if (ir.first.empty())
+	pair <h256s, h256s> blocks;
```
gavofyork: additional space.

libethereum/Client.cpp
```diff
@@ -613,10 +613,11 @@ bool Client::submitWork(ProofOfWork::Solution const& _solution)
 
 void Client::syncBlockQueue()
 {
-	ImportRoute ir;
 	cwork << "BQ ==> CHAIN ==> STATE";
-	tie(ir.first, ir.second, m_syncBlockQueue) = m_bc.sync(m_bq, m_stateDB, rand() % 10 + 5);
```
gavofyork: make `m_bc.sync` return a `pair<ImportRoute, bool>` to avoid this additional code

libethereum/BlockChain.h
```diff
@@ -80,7 +80,19 @@ ldb::Slice toSlice(h256 const& _h, unsigned _sub = 0);
 using BlocksHash = std::unordered_map<h256, bytes>;
 using TransactionHashes = h256s;
 using UncleHashes = h256s;
-using ImportRoute = std::pair<h256s, h256s>;
+
+class ImportRoute
```
gavofyork: planning more logic for this class? otherwise a struct would suffice.

chriseth:
Does not compile on linux.

libethereum/BlockChain.cpp
```diff
@@ -364,21 +364,21 @@ pair<ImportResult, ImportRoute> BlockChain::attemptImport(bytes const& _block, O
 	}
 	catch (UnknownParent&)
 	{
-		return make_pair(ImportResult::UnknownParent, make_pair(h256s(), h256s()));
+		return make_pair(ImportResult::UnknownParent, ImportRoute());
```
chfast: Preferred syntax is `ImportRoute{}` instead of `ImportRoute()`.

gavofyork:
fails on 4 platforms?

libethereum/BlockChain.cpp
gavofyork: why?

chfast: Because `ImportRoute()` syntax can be ambiguous in some cases. Like `ImportRoute x();` :) Not exactly the same but `()` are wrong here.

I think we discussed it in the past and decided to prefer `{}` over `()`. If not should we choose one option in coding standards?

chfast:
@debris please bump the buildbots. They failed because of neth.

libethereum/BlockChain.cpp
gavofyork: Indeed it's not the same. When creating temporaries, unless the C++ actually prefers one over the other (i.e. it's considering deprecating one or compiler support for the other is patchy), I consider both to be acceptable.

PR #2179, opened 2015-06-14
Prep, tests: RLPX Framing (mux, demux, socket prototype)

subtly:


gavofyork:
fails compile on all platform?

subtly:
Depends on #2246.

gavofyork:
fails on windows?

subtly:
macos failure not mine:
/Users/administrator/slave/build-cpp-ethereum-osx-pr/build/test/TestHelper.cpp:664: fatal error in "stPreCompiledContractsTransaction": Contents of /Users/administrator/slave/build-cpp-ethereum-osx-pr/tests/StateTests/stPreCompiledContractsTransaction.json is empty. Have you cloned the 'tests' repo branch develop and set ETHEREUM_TEST_PATH to its path?

CJentzsch:
just rebase, this test has been removed here: https://github.com/ethereum/cpp-ethereum/pull/2397

subtly:
This PR is ready for review and contains preparation changes as well as a test suite for new frame reader/write. This PR adds test code and does not make functional changes. FrameReader, FrameWriter, Packet are used in tests and SocketIO is proof-of-concept.

libp2p/RLPXFrameCoder.cpp
```diff
@@ -88,54 +93,81 @@ RLPXFrameCoder::RLPXFrameCoder(RLPXHandshake const& _init)
...
 	m_ingressMac.Update(keyMaterial.data(), keyMaterial.size());
 }
 
-void RLPXFrameCoder::writeSingleFramePacket(bytesConstRef _packet, bytes& o_bytes)
+void RLPXFrameCoder::writeFrame(uint16_t _protocolType, bytesConstRef _payload, bytes& o_bytes)
 {
-	// _packet = type || rlpList()
+	RLPStream header;
+	uint32_t len = (uint32_t)_payload.size();
+	header.appendRaw(bytes({byte((len >> 16) & 0xff), byte((len >> 8) & 0xff), byte(len & 0xff)}));
```
arkpar: You could probably use std::array here instead of vector

libp2p/RLPXFrameCoder.h
```diff
@@ -61,12 +71,27 @@ class RLPXFrameCoder
...
 	RLPXFrameCoder(RLPXHandshake const& _init);
+	
+	/// Construct with external key material.
+	RLPXFrameCoder(bool _originated, h512 _remoteEphemeral, h256 _remoteNonce, crypto::ECDHE const& _ephemeral, h256 _nonce, bytesConstRef _ackCipher, bytesConstRef _authCipher);
+	
 	~RLPXFrameCoder() {}
 	
-	/// Encrypt _packet as RLPx frame.
+	/// Establish shared secrets and setup AES and MAC states.
+	void setup(bool _originated, h512 _remoteEphemeral, h256 _remoteNonce, crypto::ECDHE const& _ephemeral, h256 _nonce, bytesConstRef _ackCipher, bytesConstRef _authCipher);
```
arkpar: const& hash args

libp2p/RLPXFrameCoder.h
```diff
@@ -61,12 +71,27 @@ class RLPXFrameCoder
 	friend class RLPXFrameIOMux;
 	friend class Session;
 public:
-	/// Constructor.
-	/// Requires instance of RLPXHandshake which has completed first two phases of handshake.
+	/// Construct; requires instance of RLPXHandshake which has encrypted ECDH key exchange (first two phases of handshake).
 	RLPXFrameCoder(RLPXHandshake const& _init);
+	
+	/// Construct with external key material.
+	RLPXFrameCoder(bool _originated, h512 _remoteEphemeral, h256 _remoteNonce, crypto::ECDHE const& _ephemeral, h256 _nonce, bytesConstRef _ackCipher, bytesConstRef _authCipher);
```
arkpar: const& hash args

libp2p/RLPXFrameWriter.h
```diff
@@ -0,0 +1,88 @@
...
+	 */
+	struct WriterState
+	{
+		std::deque<RLPXPacket> q;
+		mutable Mutex x;
+		
+		RLPXPacket* writing = nullptr;
+		size_t remaining = 0;
+		bool multiFrame = false;
+		uint16_t sequence;
```
arkpar: ok to be left uninitialized?

libp2p/RLPXPacket.h
```diff
@@ -0,0 +1,69 @@
...
+static bytesConstRef nextRLP(bytesConstRef _b) { try { RLP r(_b, RLP::AllowNonCanon); return _b.cropped(0, std::min((size_t)r.actualSize(), _b.size())); } catch(...) {} return bytesConstRef(); }
+
+/**
+ * RLPX Packet
+ */
+class RLPXPacket
+{
+public:
+	/// Construct packet. RLPStream data is invalidated.
+	RLPXPacket(uint8_t _capId, RLPStream& _type, RLPStream& _data): m_cap(_capId), m_type(std::move(_type.out())), m_data(std::move(_data.out())) {}
```
arkpar: _type and _data should be made && to state that they are moved

libp2p/RLPXFrameCoder.h
```diff
@@ -33,18 +33,24 @@ namespace dev
 namespace p2p
 {
 
+struct RLPXFrameDecrytFailed: virtual dev::Exception {};
```
gluk256: A typo: 
RLPXFrameDecrytFailed vs.
RLPXFrameDecryptFailed.

libp2p/RLPXFrameReader.cpp
```diff
@@ -0,0 +1,90 @@
...
+		
+		uint32_t remaining = _info.totalLength - type.size() - packet.size();
+		if (p.isValid())
+			ret.push_back(std::move(p));
+		else if (_info.multiFrame && remaining)
+			m_incomplete.insert(std::make_pair(_info.sequenceId, std::make_pair(std::move(p), remaining)));
+		else
+			BOOST_THROW_EXCEPTION(RLPXInvalidPacket());
+	}
+	return ret;
```
gluk256: would be better: return std::move(ret) ?
or it does not matter in this case?

libp2p/RLPXFrameWriter.cpp
```diff
@@ -0,0 +1,161 @@
...
+		DEV_GUARDED(m_q.first.x)
+			highPending = !!m_q.first.q.size();
+		DEV_GUARDED(m_q.second.x)
+			lowPending = !!m_q.second.q.size();
+
+		if (!highPending && !lowPending)
+			return ret;
+		
+		// first run when !swapQueues, high > low, otherwise low > high
+		bool high = highPending && !swapQueues ? true : lowPending ? false : true;
```
gluk256: would be much better to write (!lowPending) instead of (lowPending ? false : true)

libp2p/RLPXFrameWriter.cpp
```diff
@@ -0,0 +1,161 @@
...
+		WriterState &qs = high ? m_q.first : m_q.second;
+		size_t frameAllot = (!swapQueues && highPending && lowPending ? frameLen / 2 - (c_overhead + c_blockSize) > 0 ? frameLen / 2 : frameLen : frameLen) - c_overhead;
```
gluk256: it's difficult to read. i suggest braces around the embedded ternary operator.

libp2p/RLPXFrameWriter.cpp
```diff
@@ -0,0 +1,161 @@
...
+			// !qs.writing && payload.empty()	initial entry
+			// !qs.writing && !payload.empty()	continuation (multiple packets per frame)
+			// qs.writing && payload.empty()	initial entry, continuation (multiple frames per packet)
+			// qs.writing && !payload.empty()	INVALID
+			
+			// write packet-type
+			if (qs.writing == nullptr)
+			{
+				{
+					Guard l(qs.x);
```
gluk256: maybe DEV_GUARDED would be better?

libp2p/RLPXFrameReader.cpp
subtly: My understanding is that move is implicitly performed here.

libp2p/RLPXFrameWriter.h
subtly: Yes; will add a comment. Sequences are meant to roll over so an initial value is somewhat useless (it can be random). If I had a favorite number I would use that.

libp2p/RLPXFrameCoder.h
subtly: Will change.

libp2p/RLPXFrameWriter.cpp
subtly: DEV_GUARDED is a for loop so that causes break to be ineffective :(

subtly:
Only failure is the unrelated test. Will make semantic changes in follow-up PR.

libp2p/RLPXFrameCoder.h
```diff
@@ -33,18 +33,24 @@ namespace dev
...
+	uint8_t padding = 0;			///< Length of padding which follows @length.
 	
-	uint16_t protocolId = 0;
-	bool hasSequence = false;
-	uint16_t sequenceId = 0;
-	uint32_t totalLength = 0;
+	uint16_t protocolId = 0;		///< Protocol ID as negotiated by handshake.
+	bool multiFrame = false;		///< If this frame is part of a sequence
+	uint16_t sequenceId = 0;		///< Sequence ID of frame
+	uint32_t totalLength = 0;		///< Set to total length of packet in first frame of multiframe packet
```
chfast: Can you make some of the fields (possibly all of them) const? That would make clear that they are set when FrameInfo is created and never modified.

gavofyork:
waiting on fixes for issues pawel raised

subtly:
@chfast FrameInfo is const now. OSX tests are passing locally. Waiting on linux develop check.

chfast:
Thanks. I don't have time to review it more now. As it "looked good" before, I think it's ready to merge. If you think it needs more review, I can check it next week after the merge.

libp2p/RLPXFrameCoder.cpp
```diff
@@ -23,42 +23,46 @@
 
 #include <libdevcore/Assertions.h>
 #include "RLPxHandshake.h"
+#include "RLPXPacket.h"
```
chfast: Looks like there is some inconsistency in the namings of the files. Prefixes RLPx... and RLPX...

libp2p/RLPXFrameCoder.cpp
```diff
@@ -88,54 +92,81 @@ RLPXFrameCoder::RLPXFrameCoder(RLPXHandshake const& _init)
 	// Recipient egress-mac: sha3(mac-secret^initiator-nonce || auth-sent-ack)
 	//           ingress-mac: sha3(mac-secret^recipient-nonce || auth-recvd-init)
  
-	(*(h256*)outRef.data() ^ _init.m_remoteNonce).ref().copyTo(keyMaterial);
-	bytes const& egressCipher = _init.m_originated ? _init.m_authCipher : _init.m_ackCipher;
+	(*(h256*)outRef.data() ^ _remoteNonce).ref().copyTo(keyMaterial);
```
chfast: I really don't like this `*(h256*)` appearing in your code. You bypass the C++ type checking with that. Any future changes to the type h256 cannot be applied safely as the compiler will not be able to analyze the places like that one.

libp2p/RLPXFrameReader.cpp
chfast: Yes. Return statement implies the argument can be bind to rvalue reference. However explicit cast with std::move is pessimization.

libp2p/RLPXFrameWriter.cpp
chfast: Agreed

libp2p/RLPXFrameWriter.cpp
chfast: Agreed.

libp2p/RLPXPacket.h
```diff
@@ -0,0 +1,69 @@
...
+
+	/// Construct packet from single bytestream. RLPStream data is invalidated.
+	RLPXPacket(unsigned _capId, bytesConstRef _in): m_cap(_capId), m_type(nextRLP(_in).toBytes()) { if (_in.size() > m_type.size()) { m_data.resize(_in.size() - m_type.size()); _in.cropped(m_type.size()).copyTo(&m_data); } }
+	
+	RLPXPacket(RLPXPacket const& _p): m_cap(_p.m_cap), m_type(_p.m_type), m_data(_p.m_data) {}
```
chfast: It does not anything more that the default version generated by the compiler. In the same time you do not define copy assignment operator.

libp2p/RLPXPacket.h
chfast: std::move is just overused here.

libp2p/RLPXFrameWriter.h
chfast: If that uninitialized value is every used it must be set.

subtly: The uninitialized value is never used. The struct is private to RLPXFrameWriter; nothing should call a method on RLPXPacket* writing without a valid value, nothing uses sequence without setting it first -- these values are essentially memos. This is prep and only used in tests -- the next PR (waiting on this one) cleans up the queue state objects (like WriterState).

libp2p/RLPXFrameCoder.cpp
subtly: I'll fix in separate PR -- a lot of things need renaming.

libp2p/RLPXFrameCoder.cpp
subtly: This LOC is renaming variable... and yes, the bypass is intentional. Not going to fix this PR (or after the audit). Please explain in further detail in a github issue -- if this must be changed then we need to change it throughout the entire codebase.

libp2p/RLPXFrameReader.cpp
subtly: I've been told to remove move() (@chriseth) and now to add it... so not changing.

libp2p/RLPXPacket.h
subtly: Copy assignment operator isn't used and I'd rather it not be used. Packets are lightweight and need to be moved through the network not copied.

chfast: 1. Not defining the copy assignment operator is not enough to block it. The compiler with define the default version for you. To block it you must mark it as deleted: `RLPXPacket& operator=(RLPXPacket const&) = delete;`
2. Your implementation of the copy constructor is exactly the same as the default implementation of the copy constructor. In other words, if you remove it your class will behave _exactly_ the same.
3. If you wanted to make this class movable but not copyable it is done wrong. If you need help with that I can guide you.

subtly: Acknowledged. I noticed this afterwards. It's been removed and deleted in next PR. Is this a problem that needs to block the PR?

Issue #2172, opened 2015-06-12
ethminer version 0.9.25 giving segmentation error after chunking change


Ulysseus:
After updating to latest version of the cpp client I get the following warning and segmentation error, it used to mine perfectly before this

[OPENCL]:Found suitable OpenCL device [AMD Radeon HD - FirePro D300 Compute Engine] with 2147483648 bytes of GPU memory
Benchmarking on platform: { "platform": "Apple", "device": "AMD Radeon HD - FirePro D300 Compute Engine", "version": "OpenCL 1.2 " }
Preparing DAG...
Warming up...
  ℹ  17:57:09|gpuminer0  workLoop 0 #00000000… #00000000…
  ℹ  17:57:09|gpuminer0  Initialising miner...
[OPENCL]:Using platform: Apple
[OPENCL]:Using device: AMD Radeon HD - FirePro D300 Compute Engine(OpenCL 1.2 )
[OPENCL]:Using 4 chunks. Max OpenCL allocateable memory is536870912
[OPENCL]:Printing program log
[OPENCL]:<program source>:45:6: warning: no previous prototype for function 'keccak_f1600_round'
void keccak_f1600_round(uint2* a, uint r, uint out_size)
     ^
<program source>:161:6: warning: no previous prototype for function 'keccak_f1600_no_absorb'
void keccak_f1600_no_absorb(ulong* a, uint in_size, uint out_size, uint isolate)
     ^
<program source>:203:6: warning: no previous prototype for function 'fnv'
uint fnv(uint x, uint y)
     ^
<program source>:208:7: warning: no previous prototype for function 'fnv4'
uint4 fnv4(uint4 x, uint4 y)
      ^
<program source>:213:6: warning: no previous prototype for function 'fnv_reduce'
uint fnv_reduce(uint4 v)
     ^
<program source>:236:10: warning: no previous prototype for function 'init_hash'
hash64_t init_hash(__constant hash32_t const* header, ulong nonce, uint isolate)
         ^
<program source>:252:6: warning: no previous prototype for function 'inner_loop_chunks'
uint inner_loop_chunks(uint4 init, uint thread_id, __local uint* share, __global hash128_t const* g_dag, __global hash128_t const* g_dag1, __global hash128_t const* g_dag2, __global hash128_t const* g_dag3, uint isolate)
     ^
<program source>:286:6: warning: no previous prototype for function 'inner_loop'
[...]

PR #2187, opened 2015-06-15
Fix macosx opencl warnings

LefterisJP:
It seems that OpenCL macosx implementation needs a static on the
function implementations if there is no corresponding declaration as can
be seen by this report: https://github.com/ethereum/cpp-ethereum/issues/2172


LefterisJP:
Maybe wait with merging, until we test that this breaks nothing in an actual mining capable machine.

LefterisJP:
No problem detected.

PR #2190, opened 2015-06-15
test for tx in block 504980

CJentzsch:
* fix for expected section in ecrecover tests
* replicate of transaction in (forking) block 504980 (memory overlapp for identity contract)


test/libethereum/StateTestsFiller/stPreCompiledContractsFiller.json
```diff
@@ -140,7 +140,7 @@
         "expect" : {
             "095e7baea6a6c7c4c2dfeb977efac326af552d87" : {
                 "storage" : {
-                    "0x" : "0x3f17f1962b36e491b30a40b2405849e597ba5fb5",
```
chfast: Why?

CJentzsch: Because the YP states:“Importantly in the case of an invalid signature (ECDSARECOVER(h, v, r, s) = ∅), then we have no output.”
We did that wrong, here is the fix: https://github.com/ethereum/cpp-ethereum/pull/2128
Bit I forgot to update the "expect" section for those tests.

PR #2193, opened 2015-06-15
Solidity: contract documentation is now parsing during compilation and not by request

LianaHus:


LianaHus:
fails

libsolidity/CompilerStack.h
```diff
@@ -148,6 +148,9 @@ class CompilerStack: boost::noncopyable
 	/// start line, start column, end line, end column
 	std::tuple<int, int, int, int> positionFromSourceLocation(SourceLocation const& _sourceLocation) const;
 
+	/// Parses Natspec documentations. Throws exceptions in case of wrong documented contract
+	void parseNatspecDocumentation(dev::solidity::ContractDefinition& _contract);
```
chriseth: You can omit `dev::solidity` here.

chriseth:
I think I would prefer if the `InterfaceHandler` did not use a mutable reference to the contract. Furthermore, the unique pointers are not needed anymore, they were introduced to enable lazy initialisation.
What about the following: Add methods to ContractDefinition that return mutable and constant references to the doc strings, let InterfaceHandler just return the string (bare) and assign the string inside CompilerStack using the function that returns the mutable reference.

LianaHus:
sure I can do that

LianaHus:
rebased

libsolidity/CompilerStack.cpp
```diff
@@ -124,10 +124,19 @@ void CompilerStack::parse()
...
 			}
 	m_parseSuccessful = true;
 }
 
+void CompilerStack::parseNatspecDocumentation(ContractDefinition const& _contract)
+{
+	InterfaceHandler interfaceHandler;
+	string devDoc =_contract.devDocumentation();
+	devDoc = interfaceHandler.devDocumentation(_contract);
+	//interfaceHandler.generateUserDocumentation(_contract);
```
chriseth: Can that be removed?

libsolidity/InterfaceHandler.cpp
```diff
@@ -16,7 +16,7 @@ InterfaceHandler::InterfaceHandler()
 	m_lastTag = DocTagType::None;
 }
 
-unique_ptr<string> InterfaceHandler::getDocumentation(
+string InterfaceHandler::getDocumentation(
```
chriseth: Return a const reference to prevent copy?
In the error case at the end, you can return e.g. the user documentation.

libsolidity/InterfaceHandler.h
```diff
@@ -75,18 +74,18 @@ class InterfaceHandler
 	/// @param _contractDef The contract definition
 	/// @return             A unique pointer contained string with the json
```
chriseth: Please also update this and the other doc strings.

PR #2196, opened 2015-06-15
Check block header/pow before fixating on peer.

subtly:


gavofyork:
half of the point of the blockqueue is to parellelise the pow check.

subtly:
Correct. I've added code closer to the source and renamed PR to "Check block header/pow before fixating on peer".

The BlockQueue is useful for asynchronously noting blocks, however with the current implementation when EthereumHost is accepting *New* Blocks, the BlockQueue is likely to be empty. I'm not saying the BlockQueue shouldn't be used, only that an immediate improvement can be made.

If a New block with an unknown parent is received (onPeerNewBlock) then the action which follows is immediate, and the blockqueue isn't a consideration. This yields a denial of service attack vector which is worse than spending CPU cycles on the rare occasion (~10 seconds) that a peer has sent us a higher-difficulty block with an unknown parent *and* the host isn't sync'ing.

If a peer sends us a block with an unknown parent which doesn't verify, we ought to ignore the peer rather than fixate on sync'ing to it.

cc @arkpar

libethereum/EthereumHost.cpp
```diff
@@ -576,8 +576,20 @@ void EthereumHost::onPeerNewBlock(EthereumPeer* _peer, RLP const& _r)
 				u256 difficulty = _r[1].toInt<u256>();
 				if (m_syncingTotalDifficulty < difficulty)
 				{
+					BlockInfo bi;
+					RLP root(_r[0].data());
+					try
+					{
+						bi.populateFromHeader(root[0], CheckEverything, h);
```
arkpar: Generating a valid block that passes this is not much harder

subtly: How?

arkpar: An attacker can pregenerate any number of blocks with valid nonce and minimum difficulty and feed them here. BlockQueue now has a limit of unknown blocks it can handle that is supposed to protect against this.

subtly: If an attacker did that it wouldn't get to this line. The difficulty of the block is checked on line 577 before bothering with header/nonce check.
It doesn't matter what the BlockQueue does. Lines 591+ irrevocably modify state prior to *any* verification of the block or difficulty. The block could be filled with 0s except for the difficulty and without checking the header/nonce it will immediately begin syncing to the peer. Would it not?

I understand there maybe a PR in the future that will tag the block with the peer? I think until then, this PR will prevent certain malformed blocks from fooling the client into sync'ing to the peer.

arkpar: The difficulty on line 577 comes from the packet and this check can be bypassed with ease. +1 for tracking bad blocks origin and banning peers that sent them

libethereum/EthereumHost.cpp
```diff
@@ -576,8 +576,20 @@ void EthereumHost::onPeerNewBlock(EthereumPeer* _peer, RLP const& _r)
...
+					}
+					catch (Exception const& _e)
+					{
+						_peer->disable("Malformed block received.");
+						return;
+					}
+					
 					_peer->m_latestHash = h;
-					_peer->m_totalDifficulty = difficulty;
+					_peer->m_totalDifficulty = bi.difficulty;
```
arkpar: _peer->m_totalDifficulty is supposed to be a sum of difficulties for all blocks in the peer's blockchain.

subtly:
@arkpar I've come to understand an issue with my approach here. It's possible that the block number is out of bounds for the current epoch which would require 64x more SHA3 calls which would otherwise not be required. And of course, the blockq is being used now.

PR #2197, opened 2015-06-15
bloom update

gluk256:
a minor refactoring of FixedHash::bloom()


gluk256:
Should not affect anything, because FixedHash::bloom() has not been used so far.

chfast:
Just function rename?

gluk256:
Actually, the main reason for refactoring was the static_assert (line 171), because the function will not properly if M is not power of 2.

PR #2198, opened 2015-06-15
Fixed sync stalling issues

arkpar:
A few fixes on the current sync code while I'm working on refactoring.


LefterisJP:
This PR looks good to me. I have been experiencing the problems mentioned at https://github.com/ethereum/cpp-ethereum/issues/2209 while GPU mining and struggling to keep in sync with the network. Using this PR for the last hour or so, my node has not fallen out of sync at all.

As opposed to latest develop where it would slowly start to fall out of sync until it gets to a point where it would get stuck at an arbitrary block.

PR #2202, opened 2015-06-16
Pow tests

CJentzsch:
* invalid PoW in blockheader tests
* invalid PoW in uncle header tests
* diff is zero test


test/libethereum/blockchain.cpp
```diff
@@ -581,6 +581,12 @@ mArray importUncles(mObject const& _blObj, vector<BlockInfo>& _vBiUncles, vector
 		}
 
 		updatePoW(uncleBlockFromFields);
+		if (overwrite == "nonce" || overwrite == "mixHash")
```
chfast: I don't like this block. Split these independent cases into 2 ifs.

test/libethereum/blockchain.cpp
```diff
@@ -581,6 +581,10 @@ mArray importUncles(mObject const& _blObj, vector<BlockInfo>& _vBiUncles, vector
 		}
 
 		updatePoW(uncleBlockFromFields);
+
+		uncleBlockFromFields.nonce = overwrite == "nonce" ? Nonce(uncleHeaderObj["nonce"].get_str()) : uncleBlockFromFields.nonce;
```
chfast: Still don't like it. Use if please.

PR #2203, opened 2015-06-16
Add cmake command to turn CL Kernel code into a byte array

LefterisJP:
The OpenCL kernel gets parsed and copied into a byte array accessible
by a specific header during the cmake configuration step.

We are now adding a special command "make clbin2h" which would generate
this header byte array on demand


LefterisJP:
will add 1 more commit on top of this since it does not yet always work as we would need it

PR #2204, opened 2015-06-16
check nonce

CJentzsch:
The cpp equivalent of https://github.com/ethereum/go-ethereum/pull/1198
We actually did not check the nonce when `ImportRequirements::ValidNonce` was set.

### DIFF
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -1072,7 +1072,7 @@ VerifiedBlockRef BlockChain::verifyBlock(bytes const& _block, function<void(Exce
 	try
 	{
 		Strictness strictness = Strictness::CheckEverything;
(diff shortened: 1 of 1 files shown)


libethereum/BlockChain.cpp
```diff
@@ -1072,7 +1072,7 @@ VerifiedBlockRef BlockChain::verifyBlock(bytes const& _block, function<void(Exce
 	try
 	{
 		Strictness strictness = Strictness::CheckEverything;
-		if (_ir & ~ImportRequirements::ValidNonce)
+		if (~_ir & ImportRequirements::ValidNonce)
```
chfast: Why not `if (!(_ir & ImportRequirements::ValidNonce))`?

Issue #2121, opened 2015-06-08
Mix: can't pass an array to a function when testing


joeykrug:
When passing an array to something on Mix I get

```"Some parameters are invalid:
uint256[] should be between 0 and 0"?```

and ```Some parameters are invalid:
[1,1,1] does not represent uint256[] type```

PR #2206, opened 2015-06-16
Mix - bug fixes

yann300:
- array validation in QML ( fix for #2121 )
- compliant with multi dim/dynamic array


mix/ContractCallDataEncoder.cpp
```diff
@@ -39,13 +42,18 @@ bytes ContractCallDataEncoder::encodedData()
 	bytes r(m_encodedData);
 	size_t headerSize = m_encodedData.size() & ~0x1fUL; //ignore any prefix that is not 32-byte aligned
 	//apply offsets
-	for (auto const& p: m_offsetMap)
+	for (auto const& p: m_dynamicOffsetMap)
+	{
+		vector_ref<byte> offsetRef(m_dynamicData.data() + p.first, 32);
+		toBigEndian<size_t, vector_ref<byte>>(p.second + headerSize, offsetRef); //add header size minus signature hash
```
chfast: Try not to specify template arguments for _functions_. Is it necessary?

mix/ContractCallDataEncoder.cpp
```diff
@@ -55,54 +63,92 @@ void ContractCallDataEncoder::encode(QFunctionDefinition const* _function)
...
+		}
+		k++;
+	}
+}
+
 void ContractCallDataEncoder::encode(QVariant const& _data, SolidityType const& _type)
 {
-	u256 count = 1;
 	QStringList strList;
+	vector<int> dim;
```
chfast: It's not used, is it?

mix/ContractCallDataEncoder.cpp
```diff
@@ -55,54 +63,92 @@ void ContractCallDataEncoder::encode(QFunctionDefinition const* _function)
...
 	if (_type.array)
 	{
-		if (_data.type() == QVariant::String)
-			strList = _data.toString().split(",", QString::SkipEmptyParts);  //TODO: proper parsing
-		else
-			strList = _data.toStringList();
-		count = strList.count();
+		QList<int> dim = extractDimension(_type. name);
```
chfast: Shadows `vector<int> dim`. I don't know much about QList but vector is probably better.

mix/ContractCallDataEncoder.cpp
```diff
@@ -55,54 +63,92 @@ void ContractCallDataEncoder::encode(QFunctionDefinition const* _function)
 	m_encodedData.insert(m_encodedData.end(), hash.begin(), hash.end());
 }
 
+void ContractCallDataEncoder::encodeArray(QJsonArray const& _array, QList<int> _dim, SolidityType const& _type, bytes& _content)
```
chfast: You really want to copy the `_dim` list?

mix/ContractCallDataEncoder.cpp
yann300: just because QList already provides .pop_front(), but I can also use std::list, right

mix/ContractCallDataEncoder.cpp
yann300: is the way how it's used, yes it's needed...

mix/ContractCallDataEncoder.cpp
yann300: yes this is used by encodeArray()

chfast: No, you pass `QList<int> dim` to encodeArray() because `vector<int> dim` has been shadowed.

yann300: ah yes my mistake, i did not see that.

yann300: done

mix/ContractCallDataEncoder.h
```diff
@@ -71,11 +71,15 @@ class ContractCallDataEncoder
 	QString toString(bool _b);
 	QString toString(dev::bytes const& _b);
 	bool asString(dev::bytes const& _b, QString& _str);
+	QList<int> extractDimension(QString const& _type);
```
chfast: If this function is used internally (looks like) do not place is as a member function. Hide it in cpp file.

mix/qml/js/InputValidator.js
```diff
@@ -1,27 +1,20 @@
 Qt.include("QEtherHelper.js")
 
 var nbRegEx = new RegExp('^[0-9]+$');
+var arrayRegEx = new RegExp('\\[[^\\]]*\\]', "g");
```
chfast: The same is done in init() function.

mix/qml/js/InputValidator.js
```diff
@@ -29,6 +22,99 @@ function validate(model, values)
...
+	}
+	return res;
+}
+
+function isArray(_type)
+{
+	if (!arrayRegEx.test(_type))
+		return false
+	else
+		return (_type.indexOf("int") !== -1 || _type.indexOf("bytes") !== -1 || _type.indexOf("bool") !== -1 || _type.indexOf("adress") !== -1)
```
chfast: How about int256? or uint?

mix/ContractCallDataEncoder.cpp
```diff
@@ -55,54 +63,91 @@ void ContractCallDataEncoder::encode(QFunctionDefinition const* _function)
...
 		toBigEndian(count, sizeRef);
-		m_offsetMap.push_back(std::make_pair(m_encodedData.size(), sizePos));
+		m_staticOffsetMap.push_back(std::make_pair(m_encodedData.size(), sizePos));
 		m_encodedData += empty; //reserve space for offset
 	}
 	else
+		encodeSingleItem(_data.toString(), _type, m_encodedData);
+}
+
+QList<int> ContractCallDataEncoder::extractDimension(QString const& _type)
```
arkpar: This function should analyze SolidityType and not a string

mix/ContractCallDataEncoder.cpp
```diff
@@ -134,6 +179,11 @@ unsigned ContractCallDataEncoder::encodeSingleItem(QString const& _data, Solidit
 		catch (std::exception const&)
 		{
 			// manage input as a string.
+			QRegExp strExtract("\"(.*)\""); //check if contains both string and hex value, keep the string.
```
arkpar: There should be no need to handle such inputs. User should only enter one representation.

mix/qml/js/InputValidator.js
yann300: _type.indexOf("int") will work for uint and int256

mix/ContractCallDataEncoder.cpp
yann300: it was easier for now to parse string. will change it when the dimension will be available in SolidityType or in the ABI.

mix/qml/js/InputValidator.js
chriseth: Is that check needed at all? What about `string`?

mix/qml/js/InputValidator.js
yann300: Those regex are used by a few functions here... but I will change this to only instantiate them in init().

PR #2219, opened 2015-06-17
Corrected typo about max peer count.

chriseth:


LefterisJP:
not really an issue, but perhaps it would be better to make a variable/define of this max peer count and have both the condition and the log message use it?

subtly:
That could would be removed before July. This logic will be revisited and the value will become dynamic with a default and tunable min/max values.

PR #2225, opened 2015-06-17
Blockchain sync refactoring 

arkpar:
Refactored old PV60 sync code with a few improvements:
* Resync triggered on new unknown future block
* Transactions are ignored from a peer we are syncing to
* Hashchain is preserved if syncing peer disconnects
* BlockQueue difficulty is taken into account when deciding if sync required

TODO:

PV61
PV62
Store node ids in block queue
Code cleanup and more documentation


libethcore/Common.h
```diff
@@ -100,7 +100,8 @@ enum class ImportResult
 {
 	Success = 0,
 	UnknownParent,
-	FutureTime,
+	FutureTimeKnown,
+	FutureTimeUnkwnown,
```
chfast: Typo!

subtly:
It's difficult to quickly determine what's changed. Would it be too much to separate out the PR which renames and moves all the PV60 methods?

"Resync triggered on new unknown future block"
Haven't quite understood how this can work. Previously this meant that, if a particular peer sends us a new block, all other peer packets are ignored until it has finished sync'ing to the purported "better" peer. Is this still the case?

arkpar:
It's not just rename and move. Since the sync logic moved from peer to a dedicated class that manages all peers the code structure had to change a little.  Also given current network condition some improvements were required, otherwise the sync would be too slow.

"Resync triggered on new unknown future block". Before if a peer sends us an unknown block we begin syncing to this peer. It is not like all the packets are ignored, we just won't be downloading hashes/blocks from other peers. This behavior is preserved. The difference is that unknown **future** blocks now also trigger this.

libethereum/BlockChainSync.cpp
```diff
@@ -0,0 +1,781 @@
...
+	else
+	{
+		unsigned estimatedHashes = estimateHashes();
+		_peer->m_expectedHashes = estimatedHashes;
+		onNewPeer(_peer);
+	}
+	DEV_INVARIANT_CHECK;
+}
+
+unsigned BlockChainSync::estimateHashes()
```
gavofyork: should be `const`?

libethereum/BlockChainSync.cpp
```diff
@@ -0,0 +1,781 @@
...
+	return blockCount;
+}
+
+void BlockChainSync::requestBlocks(EthereumPeer* _peer)
+{
+	if (host().bq().knownFull())
+	{
+		clog(NetAllDetail) << "Waiting for block queue before downloading blocks";
+		m_lastActiveState = m_state;
+		pauseSync();
```
gavofyork: how can be be absolutely guaranteed is becomes unpaused if and only if it is still downloading this chain and the block queue has space?

libethereum/BlockChainSync.cpp
```diff
@@ -0,0 +1,781 @@
...
+		else
+		{
+			_peer->addRating(0);	// -1?
+			repeated++;
+		}
+	}
+
+	clog(NetMessageSummary) << dec << success << "imported OK," << unknown << "with unknown parents," << future << "with future timestamps," << got << " already known," << repeated << " repeats received.";
+
+	if (host().bq().unknownFull())
```
gavofyork: what's this about? it's very easy to have a message full of unknown blocks when syncing the chain in parallel across multiple peers.

libethereum/BlockChainSync.cpp
```diff
@@ -0,0 +1,781 @@
...
+		}
+		// Otherwise it's fine. We don't care if it's Nothing->Nothing.
+		DEV_INVARIANT_CHECK;
+		return;
+	}
+
+	clog(NetWarn) << "Invalid state transition:" << EthereumHost::stateName(_s) << "from" << EthereumHost::stateName(m_state) << ", " << (isSyncing(_peer) ? "syncing" : "holding") << (needsSyncing(_peer) ? "& needed" : "");
+}
+
+void PV60Sync::resetSyncFor(EthereumPeer* _peer, h256 _latestHash, u256 _td)
```
gavofyork: `h256 const&`

libethereum/BlockChainSync.cpp
```diff
@@ -0,0 +1,781 @@
...
+{
+	setNeedsSyncing(_peer, _latestHash, _td);
+}
+
+void PV60Sync::setNeedsSyncing(EthereumPeer* _peer, h256 _latestHash, u256 _td)
```
gavofyork: `h256 const&`, `u256 const&`

libethereum/BlockChainSync.cpp
```diff
@@ -0,0 +1,781 @@
...
+	RecursiveGuard l(x_sync);
+	DEV_INVARIANT_CHECK;
+	if (isSyncing())
+	{
+		clog(NetMessageSummary) << "Ignoring since we're already downloading.";
+		return;
+	}
+	unsigned knowns = 0;
+	unsigned unknowns = 0;
+	for (auto h: _hashes)
```
gavofyork: `auto const&`

libethereum/BlockChainSync.h
```diff
@@ -0,0 +1,213 @@
...
+	/// Request blocks from peer if needed
+	void requestBlocks(EthereumPeer* _peer);
+
+private:
+	static char const* const s_stateNames[static_cast<int>(SyncState::Size)];
+	bool invariants() const override = 0;
+	EthereumHost& m_host;
+	HashDownloadMan m_hashMan;
+
+protected:
```
gavofyork: customary to place class sections in increasing order of privacy: `public`, `protected`, then `private`.

gavofyork:
fixing the trivia locally. would still like reasoned argument that a) details all aspects of state (i.e. any data that is read/written in the class) and b) a reasoned argument when this data can never become inconsistent or result in a deadlock.

libethereum/BlockChainSync.cpp
arkpar: The event is triggered once queue has enough space for known blocks. There is another limit for unknown blocks and in case it is reached sync restarts: https://github.com/ethereum/cpp-ethereum/pull/2225/files#diff-9f5b8b62d8ddf0597e54e7c46e236653R220
Future blocks are treated as unknown

PR #2226, opened 2015-06-17
false positive test for bloom filter

gluk256:


gluk256:
never mind the evmjit

PR #2228, opened 2015-06-18
OpenCL: Always try single chunk DAG upload

LefterisJP:
- Removed the `--force-single-chunk` option

- Always attempt to create a single chunk DAG buffer in the GPU. If that
  fails then and only then switch to multiple chunks.

This change is motivated by the fact that many GPUs appear to be able to
actually allocate a lot more than what CL_DEVICE_MAX_MEM_ALLOC_SIZE
returns which proves that the results of querying the CL API on this
basically can't be trusted.

### DIFF
--- a/libethash-cl/ethash_cl_miner.cpp
+++ b/libethash-cl/ethash_cl_miner.cpp
@@ -321,7 +321,7 @@ bool ethash_cl_miner::init(
 			ETHCL_LOG("Printing program log");
 			ETHCL_LOG(program.getBuildInfo<CL_PROGRAM_BUILD_LOG>(device).c_str());
 		}
-		catch (cl::Error err)
+		catch (cl::Error const& err)
 		{
 			ETHCL_LOG(program.getBuildInfo<CL_PROGRAM_BUILD_LOG>(device).c_str());
 			return false;
@@ -334,7 +334,7 @@ bool ethash_cl_miner::init(
 			m_dagChunks.push_back(cl::Buffer(m_context, CL_MEM_READ_ONLY, _dagSize));
 			ETHCL_LOG("Created one big buffer for the DAG");
 		}
-		catch (cl::Error err)
+		catch (cl::Error const& err)
 		{
 			int errCode = err.err();
 			if (errCode != CL_INVALID_BUFFER_SIZE || errCode != CL_MEM_OBJECT_ALLOCATION_FAILURE)
@@ -345,6 +345,7 @@ bool ethash_cl_miner::init(
 				"Failed to allocate 1 big chunk. Max allocateable memory is "
 				<< result << ". Trying to allocate 4 chunks."
 			);
+			// The OpenCL kernel has a hard coded number of 4 chunks at the moment
 			m_dagChunksNum = 4;
 			for (unsigned i = 0; i < m_dagChunksNum; i++)
 			{
@@ -404,7 +405,7 @@ bool ethash_cl_miner::init(
 			m_search_buf[i] = cl::Buffer(m_context, CL_MEM_WRITE_ONLY, (c_max_search_results + 1) * sizeof(uint32_t));
 		}
 	}
-	catch (cl::Error err)
+	catch (cl::Error const& err)
 	{
 		ETHCL_LOG(err.what() << "(" << err.err() << ")");
 		return false;
@@ -498,7 +499,7 @@ void ethash_cl_miner::search(uint8_t const* header, uint64_t target, search_hook
 			pre_return_event.wait();
 #endif
 	}
-	catch (cl::Error err)
+	catch (cl::Error const& err)
 	{
 		ETHCL_LOG(err.what() << "(" << err.err() << ")");
 	}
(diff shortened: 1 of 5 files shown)


libethash-cl/ethash_cl_miner.cpp
```diff
@@ -346,26 +326,23 @@ bool ethash_cl_miner::init(
...
-		if (m_dagChunksNum == 1)
+		try
 		{
-			ETHCL_LOG("Creating one big buffer");
+			m_dagChunksNum = 1;
 			m_dagChunks.push_back(cl::Buffer(m_context, CL_MEM_READ_ONLY, _dagSize));
+			ETHCL_LOG("Created one big buffer for the DAG");
 		}
-		else
+		catch (...)
```
chfast: Please at least log the exception. It might happen that other type of exceptions are possible there. Then we have a chance to see that in logs, possibly from other users.

libethash-cl/ethash_cl_miner.cpp
```diff
@@ -346,26 +326,23 @@ bool ethash_cl_miner::init(
...
-			m_hash_kernel = cl::Kernel(program, "ethash_hash_chunks");
-			m_search_kernel = cl::Kernel(program, "ethash_search_chunks");
-		}
 
 		// create buffer for dag
-		if (m_dagChunksNum == 1)
+		try
 		{
-			ETHCL_LOG("Creating one big buffer");
+			m_dagChunksNum = 1;
```
chfast: Cannot see that from this perspective, but do m_dagChunksNum must still be a member?

libethash-cl/ethash_cl_miner.cpp
```diff
@@ -346,26 +326,23 @@ bool ethash_cl_miner::init(
...
-		else
+		catch (...)
+		{
+			cl_ulong result;
+			device.getInfo(CL_DEVICE_MAX_MEM_ALLOC_SIZE, &result);
+			ETHCL_LOG(
+				"Failed to allocate 1 big chunk. Max allocateable memory is "
+				<< result << ". Trying to allocate 4 chunks."
+			);
+			m_dagChunksNum = 4;
```
chfast: Comment why 4 have been chosen.

libethash-cl/ethash_cl_miner.cpp
LefterisJP: This is an OpenCL call that can only throw if there is not enough memory on the GPU to allocate for the buffer. There is no reason to log the exception here. If there is any other failure we do care about it will show when we try to allocate 4 chunks.

libethash-cl/ethash_cl_miner.cpp
LefterisJP: Probably not at the moment but I would like to keep them as a class member for a while since there is some more refactoring that needs to be done.

libethash-cl/ethash_cl_miner.cpp
LefterisJP: Because it's a "Safe" value. We probably need to at some point soon make it dynamic but the problem with this is that we would need to be rewritting the kernel dynamically at every work loop too.

libethash-cl/ethash_cl_miner.cpp
chfast: If you know the type of the exception and comment why it is ignored. Just to make it clear to other developers.

libethash-cl/ethash_cl_miner.cpp
chfast: I meant adding the comment in the code.

libethash-cl/ethash_cl_miner.cpp
LefterisJP: fair enough

libethash-cl/ethash_cl_miner.cpp
LefterisJP: no comment is needed. If anyone looks at the kernel he will immediately see why it's 4. When we refactor this there will be a dynamic value which will also be adequately commented.

chfast: For me that's exactly the reason for adding a comment here - you need to know/check the code in other places to understand what's going on here.

libethash-cl/ethash_cl_miner.cpp
```diff
@@ -346,26 +326,26 @@ bool ethash_cl_miner::init(
...
-		if (m_dagChunksNum == 1)
+		try
 		{
-			ETHCL_LOG("Creating one big buffer");
+			m_dagChunksNum = 1;
 			m_dagChunks.push_back(cl::Buffer(m_context, CL_MEM_READ_ONLY, _dagSize));
+			ETHCL_LOG("Created one big buffer for the DAG");
 		}
-		else
+		catch (cl::Error err)
```
chfast: Catch exceptions by reference.

libethash-cl/ethash_cl_miner.cpp
LefterisJP: To be fair, you do need to look at other places. This is calling into the GPU and uploading a kernel to be ran there. If you don't read the kernel's code then it's indeed difficult to see what's happening in general.

But if you so desperately need a comment there I can add it.

chfast: "kernel expects this number of chunks" would be enough.

PR #2234, opened 2015-06-18
VM: refactor and fix undefined behavior around data to memory copy

chfast:
std::memcpy and std::memset cannot be called with invalid pointers even when the size to be copied/set is 0.

### DIFF
--- a/libevm/VM.cpp
+++ b/libevm/VM.cpp
@@ -212,11 +212,11 @@ bytesConstRef VM::execImpl(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
 		auto size = static_cast<size_t>(_stack.back());
 		_stack.pop_back();
 
-		auto sizeToBeCopied = bigIndex + size > _data.size() ? _data.size() < bigIndex ? 0 : _data.size() - index : size;
+		size_t sizeToBeCopied = bigIndex + size > _data.size() ? _data.size() < bigIndex ? 0 : _data.size() - index : size;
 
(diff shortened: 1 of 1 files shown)


gavofyork:
does it pass all VM tests?

chfast:
All default VM and StateTest. But we should ask @CJentzsch to verify.

CJentzsch:
All tests pass.

libevm/VM.cpp
```diff
@@ -202,6 +202,24 @@ bytesConstRef VM::execImpl(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
...
+		auto index = static_cast<size_t>(bigIndex);
+		_stack.pop_back();
+		auto size = static_cast<size_t>(_stack.back());
+		_stack.pop_back();
+
+		auto sizeToBeCopied = bigIndex + size > _data.size() ? _data.size() < bigIndex ? 0 : _data.size() - index : size;
+
+		if (sizeToBeCopied > 0)
+			std::memcpy(_memory.data() + offset, _data.data() + index, sizeToBeCopied);
+		if (size - sizeToBeCopied > 0)
```
chriseth: Please don't use this in conjunction with `auto`. Depending on the signedness of the types of `size` and `sizeToBeCopied`, this might be equivalent to `size > sizeToBeCopied` or not.

chfast: Good point. Getting rid of subtraction will be enough?

PR #2236, opened 2015-06-18
Eliminate undefied behavior in RLPStream

chfast:


gavofyork:
what is undefined about this?

chfast:
`memcpy(0,0,0)` is undefined behavior. And it happens when both RLPStream and the argument _s are empty.

gavofyork:
so that's a particularly braindead part of the standard, but part of the standard it is. did you check to make sure that using this STL magic is no slower than just putting an `if (!_s.empty())` before the `memcpy`?

chfast:
No, I haven't checked that. But the code is smaller.

PR #2242, opened 2015-06-19
Assure datadir exists at start

LefterisJP:
I was getting a segfault at the writting of the config file if there was
no datadir (~/.ethereum) present. Before anything we need to make sure
that this directory exists.

### DIFF
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -274,6 +274,19 @@ void stopMiningAfterXBlocks(eth::Client* _c, unsigned _start, unsigned _mining)
 	this_thread::sleep_for(chrono::milliseconds(100));
 }
 
+void assertDatadirExists()
+{
+	try
+	{
+		bfs::exists(getDataDir()) || bfs::create_directory(getDataDir());
+	}
+	catch (const bfs::filesystem_error& e)
+	{
+		cout << "Failed to create data directory with: " << e.code().message();
+		exit(1);
+	}
+}
+
 int main(int argc, char** argv)
 {
 	// Init defaults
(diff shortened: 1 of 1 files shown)


eth/main.cpp
```diff
@@ -340,6 +341,10 @@ int main(int argc, char** argv)
 	/// Wallet password stuff
 	string masterPassword;
 
+	// make sure data dir exists
+	if (!boost::filesystem::create_directory(getDataDir()))
+		BOOST_THROW_EXCEPTION(FileError());
```
subtly: Does create_directory throw?
As this is CLI, would rather see an error message output to the user with exit() rather than a crash.

LefterisJP: We already throw a lot of those uncaught exceptions around in the CLI from inside other functions that are called in main(). But yes I agree. A message is in order

gavofyork:
this is the wrong place for it.

State & Blockchain definitely create paths they need. The likely candidate is KeyManager, but to be sure, get a backtrace from the exception and figure out what code is trying to open a file in a non-existant directory.

LefterisJP:
Yes it was the key manager. Specifically here:

```
	KeyManager keyManager;
	for (auto const& s: passwordsToNote)
		keyManager.notePassword(s);

	{
		RLPStream config(2);
		config << signingKey << beneficiary;
		writeFile(configFile, config.out());
	}
```

That `writeFile()` fails. Do you want me to move it before that? Or put it in the KeyManager class?

gavofyork:
yes.

gavofyork:
fixed with #2261.

PR #2243, opened 2015-06-19
Flush ethash error before thread terminate

LefterisJP:
If there is an error in the ethash functions, the worker thread that
called them will call terminate due to an unhandled exception. Now we
make sure that at least something appears in the log before that.

If in the future we can flush our LogOutputstream then we should use
that instead.

### DIFF
--- a/libethcore/EthashAux.cpp
+++ b/libethcore/EthashAux.cpp
@@ -122,7 +122,11 @@ EthashAux::LightAllocation::LightAllocation(h256 const& _seedHash)
 	uint64_t blockNumber = EthashAux::number(_seedHash);
 	light = ethash_light_new(blockNumber);
 	if (!light)
+	{
+		// LTODO: If at some point we know how to flush our LogOutputstream, use that one here
+		cout << "ERROR: Cache Generation failure." << endl;
 		BOOST_THROW_EXCEPTION(ExternalFunctionFailure("ethash_light_new()"));
+	}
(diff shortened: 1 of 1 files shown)


gavofyork:
there is no notion of being about to flush a log output stream, since it's entirely dependent on the ultimate backend for displaying logs. basically, assume that once the logging expression is done, it has been flushed. don't expect more control than that 'cause it ain't coming.

LefterisJP:
Basically the problem here is that ethash_full_new() was failing with no legible error, because the log stream output did not have enough time to be flushed and shown to the user. If `cout` with `endl` is used then we always see the message before the exception because it's already flushed to stdout. 

It actually was not easy to spot, since this all happened in a worker thread and the only thing that we saw was a `terminate` called inside `pthread_create()`. 

Something is definitely needed here. What would you like to do with this? Remove the TODO?

gavofyork:
`simpleDebugOut` which is getting called from the `clog`... calls `	cerr << ... << endl << flush;`. That should surely be at least as likely to give a timely output as `cout << ... << endl;`.

LefterisJP:
I also thought so, but for some reason it does not do so, at least in my machine. It's a very easy test to reproduce. Simply change the permissions of `~/.ethash` and you will see the problem.

LefterisJP:
Ah no ... damn. I figured it out. I had very low verbosity, this is why this error log was not visible in the output.

PR #2244, opened 2015-06-19
topic bloom filter

gluk256:
TopicBloomFilter is made a template

### DIFF
--- a/libwhisper/BloomFilter.cpp
+++ b/libwhisper/BloomFilter.cpp
@@ -25,9 +25,11 @@ using namespace std;
 using namespace dev;
 using namespace dev::shh;
 
+/*
 static unsigned const c_mask[] = { 1, 2, 4, 8, 16, 32, 64, 128 };
 
-void TopicBloomFilter::addRaw(AbridgedTopic const& _h)
+template <unsigned N>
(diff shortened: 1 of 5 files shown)


gluk256:
I have conducted several tests with different sizes and even with custom bloom() function, and came to conclusion that we need TopicBloomFilter to have the length of 8 bytes. The previous length (4 bytes) was not enough, because if you add 20 samples, then false positive rate exceeds 60%. Besides, it is not stable: it strongly depends on the initial pseudorandom seed, varying between 60% and 75%. If we encrease the length to 8 bytes, then false positive rate will be approximately 20% - 23%, and it does not depend on the pseudorandom seed anymore. 

You can run your own "falsePositiveRate" test. You just need to uncomment the line number 100 in bloomFilter.cpp, and change TopicBloomFilterTest type if necessary.

gluk256:
never mind the errors -- they are not mine

gluk256:
So, this is the relationship between the bloom size and false positive rate if 20 samples are added to the bloom:
4 bytes - 65%
8 bytes - 22%
16 bytes - 6%

subtly:
Need tests to go through :/

PR #2253, opened 2015-06-22
add test for suicide of coinbase

CJentzsch:
response to https://github.com/ethereum/yellowpaper/issues/142


winsvega:
i thought we had one

CJentzsch:
didn't find it, can you point to me to the test?

winsvega:
as always, it is better add than to find one )

Issue #2248, opened 2015-06-22
Assertion hit in EthereumPeer.cpp while downloading hashes


LefterisJP:
At the hash downloading stage of Blockchain syncing there is an assertion hit.

```
Program received signal SIGABRT, Aborted.
[Switching to Thread 0x7fff3d57b700 (LWP 4864)]
0x00007ffff0153528 in raise () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff0153528 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff015493a in abort () from /usr/lib/libc.so.6
#2  0x00007ffff014c3a7 in __assert_fail_base () from /usr/lib/libc.so.6
#3  0x00007ffff014c452 in __assert_fail () from /usr/lib/libc.so.6
#4  0x00007ffff5dd4a54 in dev::eth::EthereumPeer::requestHashes (this=0x7fff38816840, _lastHash=...) at /home/lefteris/ew/cpp-ethereum/libethereum/EthereumPeer.cpp:146
#5  0x00007ffff5ea3ae6 in dev::eth::PV60Sync::transition (this=0x7fff3800e510, _peer=0x7fff38816840, _s=dev::eth::SyncState::Hashes, _force=false, _needHelp=true) at /home/lefteris/ew/cpp-ethereum/libethereum/BlockChainSync.cpp:381
#6  0x00007ffff5ea4de4 in dev::eth::PV60Sync::attemptSync (this=0x7fff3800e510, _peer=0x7fff38816840) at /home/lefteris/ew/cpp-ethereum/libethereum/BlockChainSync.cpp:555
#7  0x00007ffff5ea505c in dev::eth::PV60Sync::<lambda(dev::eth::EthereumPeer*)>::operator()(dev::eth::EthereumPeer *) const (__closure=0x7fff3d578da0, _p=0x7fff38816840) at /home/lefteris/ew/cpp-ethereum/libethereum/BlockChainSync.cpp:597
[...]

PR #2254, opened 2015-06-22
p2p/sync fixes

arkpar:
* Used weak_ptr for peer->session and sync->peer
* Minor sync optimization: If a new hash is seen while downloading a block, attempt to download new block right away.
* New hashes/blocks logging


gavofyork:
tested on `gav.ethdev.com` for 24h?

arkpar:
running now

LefterisJP:
I also tested this PR and at least for the time I tried it, I did not get any of the assertions described in https://github.com/ethereum/cpp-ethereum/issues/2248.

Can't say it fixes it with 100% certainty but it seems so.

PR #2259, opened 2015-06-22
Issue: dev::bytesRef

winsvega:
dev::h256::ref::copyTo(dev::bytesRef out)
I have this function failing to copy the contents of bytes vector.
See the commit with test below.


winsvega:
this fix return the missing output from the precompiled contracts if called from transaction
but leaves qestions.
1) what if output > 32
2) what if output < 32?
3) what should be the default output value?

CJentzsch:
there is no default output, the output is excactly what is returned by the evm. If nothing is returned, than the output is nothing, an empty byte array. The size of the output bytes, need to be set to the size of the output that comes from the vm.

winsvega:
that is correct. 
but if I try to reset _out size from the precompiled contract itself it wont work
https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/Precompiled.cpp#L69

maybe change the copyTo function?

libethereum/Executive.cpp
```diff
@@ -229,7 +229,9 @@ bool Executive::execute()
 
 bool Executive::call(Address _receiveAddress, Address _senderAddress, u256 _value, u256 _gasPrice, bytesConstRef _data, u256 _gas)
 {
-	CallParameters params{_senderAddress, _receiveAddress, _receiveAddress, _gas, _value, _data, {}, {}};
+	dev::bytes defaultOutput = dev::fromHex("0000000000000000000000000000000000000000000000000000000000000000");
+	dev::bytesRef _out(&defaultOutput.at(0), 32);
```
subtly: Would &bytes.at(0) return the address of a temporarily created reference rather than the address of bytes?
Try dev::bytesRef _out(defaultOutput, 32);

PR #2261, opened 2015-06-23
create directory if not existent when writing to file

CJentzsch:
currently, when starting eth, it is aborting with an "FileError" exception when .ethereum does not exist. Because he can not write to a file in a not existing directory.


LefterisJP:
Related: https://github.com/ethereum/cpp-ethereum/pull/2242

PR #2263, opened 2015-06-23
topic advertisement

gluk256:
Exchange Protocol for Topic Advertisement, including tests.

### DIFF
--- a/libwhisper/Common.h
+++ b/libwhisper/Common.h
@@ -59,7 +59,7 @@ enum WhisperPacket
 };
 
 enum { TopicBloomFilterSize = 64 };
-enum { WhisperProtocolVersion = 2 };
+enum { WhisperProtocolVersion = 3 };
 
 using AbridgedTopic = FixedHash<4>;
 using Topic = h256;
--- a/libwhisper/Common.h
+++ b/libwhisper/Common.h
@@ -54,8 +54,7 @@ enum WhisperPacket
 {
 	StatusPacket = 0,
 	MessagesPacket,
-	AddTopicFilterPacket,
-	RemoveTopicFilterPacket,
+	UpdateTopicFilterPacket,
 	PacketCount
 };
 
--- a/libwhisper/Common.h
+++ b/libwhisper/Common.h
@@ -59,6 +59,7 @@ enum WhisperPacket
 };
 
 enum { TopicBloomFilterSize = 64 };
+enum { WhisperProtocolVersion = 2 };
 
 using AbridgedTopic = FixedHash<4>;
 using Topic = h256;
--- a/libwhisper/Common.h
+++ b/libwhisper/Common.h
@@ -54,8 +54,8 @@ enum WhisperPacket
 {
 	StatusPacket = 0,
 	MessagesPacket,
-	AddFilterPacket,
-	RemoveFilterPacket,
+	AddTopicFilterPacket,
+	RemoveTopicFilterPacket,
 	PacketCount
 };
 
--- a/libwhisper/Common.h
+++ b/libwhisper/Common.h
@@ -54,11 +54,11 @@ enum WhisperPacket
 {
 	StatusPacket = 0,
 	MessagesPacket,
-	UpdateTopicFilterPacket,
+	TopicFilterPacket,
 	PacketCount
 };
 
-enum { TopicBloomFilterSize = 8 };
+enum { TopicBloomFilterSize = 64 };
 
 using AbridgedTopic = FixedHash<4>;
 using Topic = h256;
--- a/libwhisper/Interface.h
+++ b/libwhisper/Interface.h
@@ -67,7 +67,6 @@ public:
 
 	virtual Topics const& fullTopics(unsigned _id) const = 0;
 	virtual unsigned installWatch(Topics const& _filter) = 0;
-//	virtual unsigned installWatchOnId(h256 _filterId) = 0;
 	virtual void uninstallWatch(unsigned _watchId) = 0;
 	virtual h256s peekWatch(unsigned _watchId) const = 0;
 	virtual h256s checkWatch(unsigned _watchId) = 0;
--- a/libwhisper/Interface.h
+++ b/libwhisper/Interface.h
@@ -67,7 +67,7 @@ public:
 
 	virtual Topics const& fullTopics(unsigned _id) const = 0;
 	virtual unsigned installWatch(Topics const& _filter) = 0;
-	virtual unsigned installWatchOnId(h256 _filterId) = 0;
+//	virtual unsigned installWatchOnId(h256 _filterId) = 0;
 	virtual void uninstallWatch(unsigned _watchId) = 0;
 	virtual h256s peekWatch(unsigned _watchId) const = 0;
 	virtual h256s checkWatch(unsigned _watchId) = 0;
--- a/libwhisper/WhisperHost.cpp
+++ b/libwhisper/WhisperHost.cpp
@@ -85,16 +85,9 @@ void WhisperHost::inject(Envelope const& _m, WhisperPeer* _p)
 	}
 }
 
-void WhisperHost::advertiseTopicsOfInterest()
-{
-	FixedHash<TopicBloomFilterSize> b = bloom();
-
-	for (auto i: peerSessions())
-		i.first->cap<WhisperPeer>().get()->sendTopicsOfInterest(b);
-}
-
 void WhisperHost::noteChanged(h256 _messageHash, h256 _filter)
 {
+	Guard l(m_filterLock);
 	for (auto& i: m_watches)
 		if (i.second.id == _filter)
 		{
@@ -103,27 +96,26 @@ void WhisperHost::noteChanged(h256 _messageHash, h256 _filter)
 		}
 }
 
-unsigned WhisperHost::installWatchOnId(h256 _h)
-{
-	auto ret = m_watches.size() ? m_watches.rbegin()->first + 1 : 0;
-	m_watches[ret] = ClientWatch(_h);
-	cwatshh << "+++" << ret << _h;
-	return ret;
-}
-
 unsigned WhisperHost::installWatch(shh::Topics const& _t)
 {
 	InstalledFilter f(_t);
 	h256 h = f.filter.sha3();
+	unsigned ret = 0;
 
-	Guard l(m_filterLock);
-	if (!m_filters.count(h))
-		m_filters.insert(make_pair(h, f));
+	DEV_GUARDED(m_filterLock)
+	{
+		if (!m_filters.count(h))
+			m_filters.insert(make_pair(h, f));
 
-	m_bloom.addRaw(f.filter.exportBloom());
+		ret = m_watches.size() ? m_watches.rbegin()->first + 1 : 0;
+		m_watches[ret] = ClientWatch(h);
+		cwatshh << "+++" << ret << h;
+
+		m_bloom.addRaw(f.filter.exportBloom());
+	}
+
 	noteAdvertiseTopicsOfInterest();
-
-	return installWatchOnId(h);
+	return ret;
 }
 
 h256s WhisperHost::watchMessages(unsigned _watchId)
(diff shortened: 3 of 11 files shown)


libwhisper/Common.h
```diff
@@ -54,8 +54,7 @@ enum WhisperPacket
 {
 	StatusPacket = 0,
 	MessagesPacket,
-	AddFilterPacket,
-	RemoveFilterPacket,
+	UpdateTopicFilterPacket,
```
subtly: Would TopicFilterPacket be ok instead of Update?

libwhisper/WhisperHost.cpp
```diff
@@ -85,6 +85,12 @@ void WhisperHost::inject(Envelope const& _m, WhisperPeer* _p)
 	}
 }
 
+void WhisperHost::advertizeTopicsOfInterest()
```
subtly: advertise

libwhisper/WhisperHost.cpp
```diff
@@ -114,6 +120,8 @@ unsigned WhisperHost::installWatch(shh::Topics const& _t)
 		m_filters.insert(make_pair(h, f));
 
 	m_bloom.addRaw(f.filter.exportBloom());
+	advertizeTopicsOfInterest();
```
subtly: Would recommend removing this so that advertisements are periodic. I would add that I hope we will change the periodic interval to a random one.

I suspect that doing that performing an immediate advertisement helps with tests. Accordingly, I'd recommend a noteTopicFilterChanged with a "forceAdvertisement" bool which will cause an immediate advertisement of the new topics. We can add this flag to install/uninstall and percolate it through so that changes can be tested without waiting on interval; this also gives us a tool for when we want to make changes to increase or reduce false-positive ratio.

libwhisper/WhisperHost.cpp
```diff
@@ -153,9 +161,11 @@ void WhisperHost::uninstallWatch(unsigned _i)
 	auto fit = m_filters.find(id);
 	if (fit != m_filters.end())
 	{
-		m_bloom.removeRaw(fit->second.filter.exportBloom());
 		if (!--fit->second.refCount)
 			m_filters.erase(fit);
+
+		m_bloom.removeRaw(fit->second.filter.exportBloom());
+		advertizeTopicsOfInterest();
```
subtly: Same as previous comment; immediate advertisement is optional.

libwhisper/WhisperHost.h
```diff
@@ -50,11 +50,12 @@ class WhisperHost: public HostCapability<WhisperPeer>, public Interface, public
 public:
 	WhisperHost();
 	virtual ~WhisperHost();
-
 	unsigned protocolVersion() const { return 2; }
+	void cleanup();
+	std::map<h256, Envelope> all() const { ReadGuard l(x_messages); return m_messages; }
```
subtly: Newlines between method declarations. Document methods.

libwhisper/WhisperHost.h
```diff
@@ -50,11 +50,12 @@ class WhisperHost: public HostCapability<WhisperPeer>, public Interface, public
...
+	FixedHash<TopicBloomFilterSize> const& bloom() const { return m_bloom; }
```
subtly: Is this thread-safe?  Can m_bloom be mutated while another thread is reading?
If so, but the method is only used internally and serially, then consider making method protected and setting a friend class. In either case, document thread-safety.

libwhisper/WhisperHost.h
```diff
@@ -77,8 +73,8 @@ class WhisperHost: public HostCapability<WhisperPeer>, public Interface, public
 	virtual void onStopping() override { stopWorking(); }
 
 	void streamMessage(h256 _m, RLPStream& _s) const;
-
 	void noteChanged(h256 _messageHash, h256 _filter);
+	void advertizeTopicsOfInterest();
```
subtly: Maybe better to name 'sendTopicsOfInterest()'. This is a network operation and contains no other logic (with respect to advertising).

libwhisper/WhisperPeer.cpp
```diff
@@ -32,7 +31,11 @@ using namespace dev::shh;
 WhisperPeer::WhisperPeer(Session* _s, HostCapabilityFace* _h, unsigned _i, CapDesc const&): Capability(_s, _h, _i)
 {
 	RLPStream s;
-	sealAndSend(prep(s, StatusPacket, 1) << version());
+	//sealAndSend(prep(s, StatusPacket, 1) << version());
```
subtly: It's safe to remove this. However, please also bump the protocol version for whisper capability. This will cause the capability to be disabled for older client versions; accordingly, we won't need the commented out code.

libwhisper/WhisperPeer.cpp
```diff
@@ -32,7 +31,11 @@ using namespace dev::shh;
...
+	prep(s, StatusPacket, 2);
+	s << version();
+	s << host()->bloom();
```
subtly: :+1: 
Note/todo: I'll help you test this. We might also remove the version() as it's now communicated in the handshake.

libwhisper/WhisperPeer.cpp
```diff
@@ -56,6 +59,12 @@ bool WhisperPeer::interpret(unsigned _id, RLP const& _r)
 
 		if (protocolVersion != version())
 			disable("Invalid protocol version.");
+		else
+		{
```
subtly: Backwards compatibility isn't necessary.

libwhisper/WhisperPeer.cpp
```diff
@@ -56,6 +59,12 @@ bool WhisperPeer::interpret(unsigned _id, RLP const& _r)
...
+			if (_r.itemCount() > 1) // for backwards compatibility
+				m_bloom = (FixedHash<TopicBloomFilterSize>)_r[1];
+			advertizeTopicsOfInterest();
```
subtly: Periodic advertisement can be relied on here. Recommend a noteChanged() here as well.

libwhisper/WhisperHost.h
subtly: Clarification: Suggestion is to rename WhisperPeer::advertizeTopicsOfInterest.

test/libwhisper/whisperTopic.cpp
```diff
@@ -301,4 +302,76 @@ BOOST_AUTO_TEST_CASE(asyncforwarding)
...
 
+BOOST_AUTO_TEST_CASE(topicAdvertising)
+{
+	if (test::Options::get().nonetwork)
+		return;
+
+	cnote << "Testing Topic Advertising...";
+	VerbosityHolder setTemporaryLevel(2);
+
+	Host host1("first", NetworkPreferences("127.0.0.1", 30303, false));
```
subtly: Can you update this for a random port?
NetworkPreferences p("127.0.0.1");
p.natTraversal = false;

Let me know if I can help here. We eventually need to do this w/all the tests.

gluk256:
it seems, the errors are not mine.

libwhisper/WhisperPeer.cpp
```diff
@@ -104,3 +114,13 @@ void WhisperPeer::noteNewMessage(h256 _h, Envelope const& _m)
...
 }
+
+void WhisperPeer::sendTopicsOfInterest(FixedHash<TopicBloomFilterSize> const& _bloom)
+{
+	RLPStream s;
+	prep(s, TopicFilterPacket, 1);
+	s << _bloom;
+	sealAndSend(s);
+
+	m_advertiseTopicsOfInterest = false;
```
subtly: m_advertiseTopicsOfInterest may be set to true while this method is executing and it will overwrite the value to false. Would be good to mutex m_advertiseTopicsOfInterest and set to false before sending the packet.

libwhisper/WhisperHost.cpp
```diff
@@ -105,15 +113,16 @@ unsigned WhisperHost::installWatchOnId(h256 _h)
 
 unsigned WhisperHost::installWatch(shh::Topics const& _t)
 {
-	Guard l(m_filterLock);
-
 	InstalledFilter f(_t);
 	h256 h = f.filter.sha3();
 
+	Guard l(m_filterLock);
```
subtly: Use DEV_GUARDED (see codebase for examples).

gluk256:
I can not use DEV_GUARDED in this case, because m_filterLock mutex is used to guard all the statements there.

gluk256:
apparently, the errors are not mine.

libwhisper/Interface.h
```diff
@@ -67,7 +67,7 @@ class Interface
 
 	virtual Topics const& fullTopics(unsigned _id) const = 0;
 	virtual unsigned installWatch(Topics const& _filter) = 0;
-	virtual unsigned installWatchOnId(h256 _filterId) = 0;
+//	virtual unsigned installWatchOnId(h256 _filterId) = 0;
```
LefterisJP: You can simply delete the function instead of commentint it out.

libwhisper/WhisperHost.h
```diff
@@ -50,36 +50,29 @@ class WhisperHost: public HostCapability<WhisperPeer>, public Interface, public
 public:
 	WhisperHost();
 	virtual ~WhisperHost();
-
-	unsigned protocolVersion() const { return 2; }
+	unsigned protocolVersion() const { return WhisperProtocolVersion; }
+	void cleanup(); ///< remove old messages
```
LefterisJP: I think we use this style around the codebase only for variables or class members. Or at least grepping around seems to show that. Please use `///` for function documentations.

libwhisper/WhisperHost.h
```diff
@@ -50,36 +50,29 @@ class WhisperHost: public HostCapability<WhisperPeer>, public Interface, public
...
-
 	virtual Topics const& fullTopics(unsigned _id) const override { try { return m_filters.at(m_watches.at(_id).id).full; } catch (...) { return EmptyTopics; } }
 	virtual unsigned installWatch(Topics const& _filter) override;
-	virtual unsigned installWatchOnId(h256 _filterId) override;
 	virtual void uninstallWatch(unsigned _watchId) override;
 	virtual h256s peekWatch(unsigned _watchId) const override { dev::Guard l(m_filterLock); try { return m_watches.at(_watchId).changes; } catch (...) { return h256s(); } }
 	virtual h256s checkWatch(unsigned _watchId) override { cleanup(); dev::Guard l(m_filterLock); h256s ret; try { ret = m_watches.at(_watchId).changes; m_watches.at(_watchId).changes.clear(); } catch (...) {} return ret; }
-	virtual h256s watchMessages(unsigned _watchId) override; /// returns IDs of messages, which match specific watch criteria
-
+	virtual h256s watchMessages(unsigned _watchId) override; ///< returns IDs of messages, which match specific watch criteria
```
LefterisJP: Same comment as above

test/libwhisper/bloomFilter.cpp
```diff
@@ -59,7 +58,7 @@ void testRemoveExistingBloom(TopicBloomFilterShort& _f, AbridgedTopic const& _h)
 	BOOST_REQUIRE(!_f.containsBloom(_h));
 }
 
-int calculateExpected(TopicBloomFilterTest const& f, int const n)
+double calculateExpected(TopicBloomFilterTest const& f, int const n)
```
LefterisJP: n seems to not be used, I guess you can remove

test/libwhisper/bloomFilter.cpp
```diff
@@ -111,11 +112,13 @@ BOOST_AUTO_TEST_CASE(falsePositiveRate)
 	TopicBloomFilterTest f;
 	Topic x(0xC0DEFEED); // deterministic pseudorandom value
 
-	for (int i = 1; i < 21; ++i)
+	double expectedRate = 0;
+
+	for (int i = 1; i < 50 && expectedRate < 0.5; ++i)
```
LefterisJP: float comparison like this is discouraged. In the olden days and in C we basically always compare the difference being above a trheshold called `EPSILON`.

Now C++11 has a [macro for that](http://www.cplusplus.com/reference/cmath/isless/). You can use `isless(expectedRate, 0.5)`

LefterisJP:
+1 in the looks good from me now. This should be merged.

PR #2277, opened 2015-06-24
Fix potential uint overflow in toBigEndian

LefterisJP:
The `auto` there will always evaluate to a `size_t` which is an unsigned int, so if an empty string or bytes vector is passed then we will get an overflow (underflow(?) to be exact).


gavofyork:
overflow out occur when the loop exit condition becomes true.

LefterisJP:
Note: The overflow would never happen due to the usage of `i++` in the condition but still this is a bit more readable, plus more importantly will make clang sanitizer happy.

PR #2282, opened 2015-06-25
Fixed blocks/transactions being propogated to pv61 peers only

arkpar:


LefterisJP:
Tested with an actual transaction in the network and this PR fixes the issue of not being able to send transactions to non-C++ peers.

PR #2284, opened 2015-06-25
Increase readability and make sanitizer happy

LefterisJP:
Increasing readability and making the clang sanitizer happy about the
loop condition in EthereumHost.cpp


LefterisJP:
rebased

chriseth:
Change to upward-counting loop?

LefterisJP:
Nah I think it's okay like that. It does not make it any more/less readable at least to my eyes.
```c++
for (unsigned i = 0; i < chosenSize && allowed.size(); i++)
```

PR #2290, opened 2015-06-25
Mix - fix unit testing

yann300:


yann300:
There is still a test failing, I will address this in another PR.
also TODO: need to use signal in rebuild() and newscenario() instead of wait(...)

PR #2094, opened 2015-06-05
Secp256k1

subtly:

PR #2297, opened 2015-06-26
Prep for #2094 and supporting both cryptopp and secp256k1

subtly:
Rename cryptopp Secp256k1 to Secp256k1PP. Update toPublic to use secp256k1 library and add test.


subtly:
All green!

PR #2298, opened 2015-06-26
Adjust CL miner work batch size properly

LefterisJP:
We are now propery adjusting the batch size of the OpenCL miner
depending on the execution time of the last search.

### DIFF
--- a/libethash-cl/ethash_cl_miner.cpp
+++ b/libethash-cl/ethash_cl_miner.cpp
@@ -33,6 +33,7 @@
 #include <vector>
 #include <libethash/util.h>
 #include <libethash/ethash.h>
+#include <libethcore/Ethash.h>
 #include <libethash/internal.h>
 #include "ethash_cl_miner.h"
 #include "ethash_cl_miner_kernel.h"
@@ -49,6 +50,7 @@
 #undef max
 
 using namespace std;
+using namespace dev::eth;
 
 // TODO: If at any point we can use libdevcore in here then we should switch to using a LogChannel
 #define ETHCL_LOG(_contents) cout << "[OPENCL]:" << _contents << endl
@@ -181,9 +183,9 @@ bool ethash_cl_miner::configureGPU(
 
 bool ethash_cl_miner::s_allowCPU = false;
 unsigned ethash_cl_miner::s_extraRequiredGPUMem;
-unsigned ethash_cl_miner::s_msPerBatch = CL_DEFAULT_MS_PER_BATCH;
-unsigned ethash_cl_miner::s_workgroupSize = CL_DEFAULT_LOCAL_WORK_SIZE;
-unsigned ethash_cl_miner::s_initialGlobalWorkSize = CL_DEFAULT_GLOBAL_WORK_SIZE_MULTIPLIER * CL_DEFAULT_LOCAL_WORK_SIZE;
+unsigned ethash_cl_miner::s_msPerBatch = Ethash::defaultMSPerBatch;
+unsigned ethash_cl_miner::s_workgroupSize = Ethash::defaultLocalWorkSize;
+unsigned ethash_cl_miner::s_initialGlobalWorkSize = Ethash::defaultGlobalWorkSizeMultiplier * Ethash::defaultLocalWorkSize;
 
 bool ethash_cl_miner::searchForAllDevices(function<bool(cl::Device const&)> _callback)
 {
--- a/libethash-cl/ethash_cl_miner.cpp
+++ b/libethash-cl/ethash_cl_miner.cpp
@@ -140,6 +140,7 @@ unsigned ethash_cl_miner::getNumDevices(unsigned _platformId)
 
 bool ethash_cl_miner::configureGPU(
 	unsigned _platformId,
+	unsigned _localWorkSize,
 	unsigned _globalWorkSize,
 	unsigned _msPerBatch,
 	bool _allowCPU,
@@ -147,6 +148,7 @@ bool ethash_cl_miner::configureGPU(
 	boost::optional<uint64_t> _currentBlock
 )
 {
+	s_workgroupSize = _localWorkSize;
 	s_initialGlobalWorkSize = _globalWorkSize;
 	s_msPerBatch = _msPerBatch;
 	s_allowCPU = _allowCPU;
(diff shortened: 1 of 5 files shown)


libethash-cl/ethash_cl_miner.cpp
```diff
@@ -495,19 +494,23 @@ void ethash_cl_miner::search(uint8_t const* header, uint64_t target, search_hook
 				pending.pop();
 			}
 
-/*			chrono::high_resolution_clock::duration d = chrono::high_resolution_clock::now() - t;
-			if (d > chrono::milliseconds(_msPerBatch * 10 / 9))
+			// adjust batch size depending on last search time
+			auto d = chrono::duration_cast<chrono::milliseconds>(chrono::high_resolution_clock::now() - t);
+			if (d != chrono::milliseconds(0)) // if duration is zero, we did not get in the actual search so adjust nothing
 			{
```
LefterisJP: I know this goes against coding standards but as in many other places, this gives a clang warning for potential dangling If, so it asks you to explicitly add the braces.

subtly:
Has this been tested on the GPUs we have in the office?

subtly:
I'll followup my previous comment with -- I've tested this and it doesn't work as OpenCL is rejecting the altered work size (m_batchSize).

LefterisJP:
Hm ... then this is new. I tested it in the GPU at the office and at the one I have home, and no rejection happened. We can limit the batch size by querying the limits of the batch size that OpenCL has (CL_DEVICE_MAX_WORK_GROUP_SIZE) and make sure any adjustments happens within those limits.

The problem with this is that I have learned not to trust OpenCL values as returned by `clGetDeviceInfo()`. But if as you say it can happen to a GPU that the batch size is rejected then this has to happen.

LefterisJP:
rebased on develop

libethash-cl/ethash_cl_miner.cpp
```diff
@@ -455,8 +455,7 @@ void ethash_cl_miner::search(uint8_t const* header, uint64_t target, search_hook
 		uint64_t start_nonce = uniform_int_distribution<uint64_t>()(engine);
 		for (;; start_nonce += m_batchSize)
 		{
-//			chrono::high_resolution_clock::time_point t = chrono::high_resolution_clock::now();
-
+			chrono::high_resolution_clock::time_point t = chrono::high_resolution_clock::now();
```
chfast: use auto here too?

LefterisJP: yeah why not, let me piggyback it in the next commit

LefterisJP:
This Pull Request seems to be dealing with the batchSize, better known in OpenCL as the global work size. So let's make this be "The Fix".

I noticed that in [this](https://github.com/LefterisJP/cpp-ethereum/commit/a8eb96755cdffdc7e2c6f59003bc8cb9cc971166) commit @gavofyork changed it from `1024*256` to `1024*16` along with making a first attempt to introduce dynamic global work size adjustment. 

Miners in the mining channel have reported quite a dropped hashrate due to that specific change. I suggest 2 things:

1. Revert back to the original global work size
2. Make dynamic work size adjustment optional, off by default, and activateable by an option.

Genoil:
Why try to adjust batch size on each search call? And why try to make it 100ms long? On most setups there's only one ideal batch size, that what gives the best hash rate. So instead of this you want some kind of auto tune that's present in a lot of crypto miners, that runs on first run by default and stores results in some settings file.

gavofyork:
so we cannot abort during the course of a workload. this is what causes the large "pause" times (actually just the time spent waiting for the workload to finish and the abortion condition to be caught.

we should probably allow both workload specifying and a target max abort time (100ms in this patch) until we can instrument the best approach.

Genoil:
Thanks for clarifying. On my GPU, 100ms is a ~2M workload that actually gives me the highest hashrate. A 1M workload (~50ms) gives approx. the same result, so there's quite some freedom. On Windows, there actually is a 2000ms abort time after which the driver resets itself. But I need to give the miner a 2 ^ 25 workload to crash it.

LefterisJP:
rebased

gavofyork:
fails on windows?

LefterisJP:
It's all unrelated errors due to some develop failures yesterday. Will rebase again and see.

LefterisJP:
Yet another rebase!

LefterisJP:
This should resolve #2306 by providing better defaults and also allowing for customization of both global and local size.

LefterisJP:
This should also resolve #2328 since we allow customization of the local work size (workgroup size)

gavofyork:
fails to build on windows. header issue?

LefterisJP:
Hmm I don't have a Windows build handy. @subtly do I recall correctly? This builds for you in Windows, right?

LefterisJP:
I think I may know why the windows build failed. Let it run the tests again.

LefterisJP:
Well at least now the windows build seems to pass. The OSX failures don't seem to have anything to do with me.

gavofyork:
I'd prefer not to have horrible C constants in my nice C++ file.

LefterisJP:
It's a bit more work to add constants as part of the class, and especially since they are exposed and used to also present the default values in the `--help` docstring. C macros are not so horrible. You also use them quite a bit.

But yes you are indeed correct that this should have been a c++ constant. Let me do that.

conradjohnson:
I have 2 machines, both Ubuntu 14.04, one with AMD R9 290s and one with AMD 7990s.  I can test anything y'all want to throw at them on there if you want.  Been mining on testnet with ethminer for a while now.

gavofyork:
seems the build fail is unrelated.

conradjohnson:
Seems using multiple instances for multiple GPUs, it will drop the hash on the previous process.

subtly:
This PR requires ethash-cl to link ethcore. Is that intentional?
This doesn't compile on macos and probably won't compile on windows.

libethcore/Ethash.cpp
```diff
@@ -54,6 +54,9 @@ namespace dev
 namespace eth
 {
 
+const unsigned Ethash::defaultLocalWorkSize = 64;
```
subtly: How will ethash-cl resolve this value?

LefterisJP:
@conradjohnson Can you elaborate? Are you by any chance in the Skype mining channel? We can chat about the issue there.

conradjohnson:
Can you add me to skype mining channel?

PR #2299, opened 2015-06-26
Mix: Add memory address in memory dump.

chriseth:


chriseth:
Not the best solution, but having addresses is really important.

LefterisJP:
looks good. So this is not used yet, right? I did not see any new calls to `memDumpToList()` using the 3rd argument.

LefterisJP:
looks good. So this is not used yet, right? I did not see any new calls to `memDumpToList()` using the 3rd argument.

yann300:
looks good.

PR #2307, opened 2015-06-26
windows build fix

subtly:


LefterisJP:
+1 Looks Good ... but would be nice to actually state what was wrong and is being fixed. By just looking at the PR and the testbot builds, I can see nothing wrong with the "Windows C++ Pull Requests"

PR #2309, opened 2015-06-27
DB errors in BlockChain

subtly:
Check write status for errors in BlockChain and log error so an IO issue doesn't put the database or client into an invalid state.

### DIFF
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -100,8 +100,8 @@ namespace dev
 {
 class WriteBatchNoter: public ldb::WriteBatch::Handler
 {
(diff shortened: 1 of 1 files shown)


subtly:
Tests are good -- one failure and it's network/unrelated.

libethereum/BlockChain.cpp
```diff
@@ -96,6 +96,15 @@ ldb::Slice dev::eth::toSlice(h256 const& _h, unsigned _sub)
 #endif
 }
 
+namespace dev
+{
+class WriteBatchNoter: public ldb::WriteBatch::Handler
+{
+	virtual void Put(ldb::Slice const& _key, ldb::Slice const& _value) { cout << "Put" << toHex(bytesConstRef(_key)) << "=>" << toHex(bytesConstRef(_value)); }
```
arkpar: cwarn or cerr maybe?

subtly:
@arkpar Changed it to same as in state.

PR #2310, opened 2015-06-27
Prevent syncing with busy peers

arkpar:


libethereum/BlockChainSync.cpp
```diff
@@ -649,9 +649,8 @@ void PV60Sync::noteDoneBlocks(std::shared_ptr<EthereumPeer> _peer, bool _clemenc
 		}
 		resetSync();
 		downloadMan().reset();
-		transition(_peer, SyncState::Idle);
```
arkpar: this transition is performed by the caller

PR #2313, opened 2015-06-29
Remove legacy code in blockchain.cpp

LefterisJP:


gavofyork:
feel free to move `contains` to somewhere else, but i find it infinitely more readable than the cryptic and verbose STL alternative. man i hate the STL.

LefterisJP:
But it's a function inside a .cpp file that's used in only one place which does exactly what an STL function does. I will really have to insist on this PR. I don't believe that replicating STL or not should be a matter of personal preference. It should just not happen.

gavofyork:
> feel free to move contains to somewhere else

like a .h file.

LefterisJP:
I still believe it to be unecessary but I am willing to compromise.

Rebased on develop and moved it to CommonData.h

libdevcore/CommonData.h
```diff
@@ -334,4 +334,10 @@ std::vector<T> keysOf(std::unordered_map<T, U> const& _m)
 	return ret;
 }
 
+template <class T, class V>
+bool contains(T const& _t, V const& _v)
+{
+	return std::end(_t) != std::find(std::begin(_t), std::end(_t), _v);
```
gavofyork: std::end could conceivably be non-trivial. so i'd prefer:

```c++
auto e = std::end(_t);
return e != std::find(std::begin(_t), e, _v);
```

LefterisJP:
I am pretty sure that all of the compilers we use would perform Common Subexpression Elimination on this and optimize the extra `std::end()` away.

subtly:
@LefterisJP On to getting contains() into the next C++ standard? :+1:

gavofyork:
they'd still insist on having to pass three args in (begin, end, value) rather than two (container, value) like Every. Other. Language.

PR #2329, opened 2015-06-30
Distribution Rate test introduced

gluk256:


test/libwhisper/bloomFilter.cpp
```diff
@@ -244,4 +244,53 @@ BOOST_AUTO_TEST_CASE(bloomFilterRaw)
 	BOOST_REQUIRE(!f.contains(b00110111));
 }
 
-BOOST_AUTO_TEST_SUITE_END()
\ No newline at end of file
+enum { DistributionTestSize = 8 };
```
gavofyork: `static const int DistributionTestSize = 8;` is preferred as it doesn't break GCC.

arkpar: why not just use const int?

gluk256:
fixed

test/libwhisper/bloomFilter.cpp
```diff
@@ -244,4 +244,53 @@ BOOST_AUTO_TEST_CASE(bloomFilterRaw)
...
+				}
+}
+
+BOOST_AUTO_TEST_CASE(distributionRate)
+{
+	cnote << "Testing Bloom Filter Distribution Rate...";
+
+	unsigned const c_arrSize = DistributionTestSize * 8;
+	unsigned distribution[c_arrSize];
+	memset(distribution, 0, c_arrSize * sizeof(unsigned));
```
subtly: Why not array of unsigned?  Default allocator will set values to 0 and (typically) ensures for loop won't go out of bounds.

gavofyork:
agreed with alex's comment.

gluk256:
I have fixed accordingly.
However, my tests show that std::array does not set values to 0. You still have to initialize them.

subtly:
Ah, sorry. That was std::vector. Array does have fill() :)

PR #2336, opened 2015-06-30
Fixed gcc build

arkpar:


libwhisper/Common.h
```diff
@@ -58,8 +58,8 @@ enum WhisperPacket
 	PacketCount
 };
 
-enum { TopicBloomFilterSize = 64 };
-enum { WhisperProtocolVersion = 3 };
+const int TopicBloomFilterSize = 64;
```
gavofyork: `static const int` is in keeping with elsewhere in the codebase.

PR #2338, opened 2015-06-30
Memory structs.

chriseth:
Finally, the following is possible:
```js
contract C {
  struct S { uint a; uint b; }
  function f() internal returns (S) { return S(1, 2); }
  function g() internal returns (S) { return S({b: 2, a: 1}); }
}
```

### DIFF
--- a/libsolidity/CompilerUtils.cpp
+++ b/libsolidity/CompilerUtils.cpp
@@ -481,7 +481,7 @@ void CompilerUtils::convertType(Type const& _typeOnStack, Type const& _targetTyp
 			}
 			break;
 		case DataLocation::CallData:
-			solAssert(false, "");
+			solAssert(false, "Invalid type conversion target location CallData.");
 			break;
 		}
 		break;
--- a/libsolidity/CompilerUtils.cpp
+++ b/libsolidity/CompilerUtils.cpp
@@ -436,19 +436,54 @@ void CompilerUtils::convertType(Type const& _typeOnStack, Type const& _targetTyp
 	}
 	case Type::Category::Struct:
 	{
-		//@todo we can probably use some of the code for arrays here.
 		solAssert(targetTypeCategory == stackTypeCategory, "");
 		auto& targetType = dynamic_cast<StructType const&>(_targetType);
-		auto& stackType = dynamic_cast<StructType const&>(_typeOnStack);
+		auto& typeOnStack = dynamic_cast<StructType const&>(_typeOnStack);
 		solAssert(
-			targetType.location() == DataLocation::Storage &&
-				stackType.location() == DataLocation::Storage,
-			"Non-storage structs not yet implemented."
-		);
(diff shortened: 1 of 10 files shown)


libsolidity/CompilerUtils.cpp
```diff
@@ -436,19 +436,54 @@ void CompilerUtils::convertType(Type const& _typeOnStack, Type const& _targetTyp
...
+					TypePointer targetMemberType = targetType.getMemberType(member.name);
+					solAssert(!!targetMemberType, "Member not found in target type.");
+					convertType(*member.type, *targetMemberType, true);
+					storeInMemoryDynamic(*targetMemberType, true);
+				}
+				m_context << eth::Instruction::POP << eth::Instruction::POP;
+			}
+			break;
+		case DataLocation::CallData:
+			solAssert(false, "");
```
LefterisJP: Maybe put a reason here in the assertion?

libsolidity/ExpressionCompiler.cpp
```diff
@@ -315,38 +313,66 @@ bool ExpressionCompiler::visit(FunctionCall const& _functionCall)
...
 		Expression const& firstArgument = *_functionCall.getArguments().front();
 		firstArgument.accept(*this);
 		utils().convertType(*firstArgument.getType(), *_functionCall.getType());
+		return false;
 	}
+
+	FunctionTypePointer functionType;
+	if (_functionCall.isStructConstructorCall())
+	{
+		TypeType const& type = dynamic_cast<TypeType const&>(*_functionCall.getExpression().getType());
```
LefterisJP: why not use auto in both cases? Or spell out the type in both cases? Minor grumbling .... :)

libsolidity/LValue.cpp
```diff
@@ -238,52 +240,66 @@ void StorageItem::storeValue(Type const& _sourceType, SourceLocation const& _loc
...
-					<< eth::Instruction::POP << eth::Instruction::POP;
-			else
-				m_context
-					<< eth::Instruction::SWAP2 << eth::Instruction::POP
-					<< eth::Instruction::SWAP2 << eth::Instruction::POP;
+				utils.popStackSlots(2 + sourceType.getSizeOnStack());
+			else if (sourceType.getSizeOnStack() >= 1)
+			{
+				// remove the source ref
+				solAssert(sourceStackSize <= 2, "");
```
LefterisJP: useful assertion error?

libsolidity/Types.cpp
```diff
@@ -1012,6 +1018,17 @@ bool StructType::canLiveOutsideStorage() const
 	return true;
 }
 
+unsigned StructType::getSizeOnStack() const
+{
+	switch (location())
```
LefterisJP: I guess you plan to build more on this switch right? Cause it may be an overkil if it's just for these 2.

PR #2339, opened 2015-06-30
Fixed blocks being stuck in verifying state

arkpar:


libethereum/BlockQueue.cpp
```diff
@@ -131,6 +127,7 @@ void BlockQueue::verifierBody()
 				}
 			cwarn << "BlockQueue missing our job: was there a GM?";
 			OK1:;
+			drainVerified();
```
arkpar: Here if bad block is at m_verifying[0] and m_verifying[1] contains a verified block latter will be never removed from m_verifying blocking the queue.

gavofyork: what if the bad block, `m_verifying[0]`, was on a different chain (perhaps it's a sibling) to `m_verifying[1]`?

libethereum/BlockQueue.cpp
```diff
@@ -179,6 +167,21 @@ void BlockQueue::verifierBody()
 	}
 }
 
+void BlockQueue::drainVerified()
```
gavofyork: functions that operate on the assumption there is one or more mutexes already locked should include that in their name. e.g. `drainVerified_WITH_BOTH_LOCKS`

libethereum/BlockQueue.cpp
```diff
@@ -179,6 +167,21 @@ void BlockQueue::verifierBody()
...
+{
+	while (m_verifying.size() && !m_verifying.front().blockData.empty())
```
gavofyork: language sanitizers generally prefer the always-fastest `!m_verifying.empty()` over `m_verifying.size()`.

libethereum/BlockQueue.cpp
gavofyork: prefer this to be called `purgeBadFromVerified`.

arkpar: That would be misleading. The main purpose of this function is to clear the front of m_verifying of the blocks that have been verified, no matter good or bad. Note that is is also used on line 150 for the case of good blocks.

PR #2340, opened 2015-07-01
addr member for contracts.

chriseth:


chriseth:
Postponed due to the bad naming of the new member and other reasons.

PR #2341, opened 2015-07-01
Gav

gavofyork:


libethash-cl/ethash_cl_miner.cpp
```diff
@@ -497,6 +497,7 @@ void ethash_cl_miner::search(uint8_t const* header, uint64_t target, search_hook
 				pending.pop();
 			}
 
+			(void)_msPerBatch;
```
LefterisJP: This is already there at (line 420)[https://github.com/ethereum/cpp-ethereum/pull/2341/files#diff-ba9178d44e4409d8a5228faab616a06aR420]. Build was broken when compiling with -Wextra so I added it over the weekend.

It's all removed in [this PR](https://github.com/ethereum/cpp-ethereum/pull/2298) though so better not mess with that file here.

PR #2342, opened 2015-07-01
FuzzTests: More Smart Codes

winsvega:


test/fuzzTesting/fuzzHelper.cpp
```diff
@@ -84,9 +84,17 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
 			}
 		}
 		else
+		{
+			if (info.name.find_first_of("PUSH") > 0)
```
chriseth: If I am not mistaken, this searches the string info.name for any of the letters 'P', 'U', 'S' and 'H'. Is that the intention?

test/fuzzTesting/fuzzHelper.cpp
```diff
@@ -143,16 +151,136 @@ std::string RandomCode::fillArguments(dev::eth::Instruction _opcode, RandomCodeO
 	{
 		switch (_opcode)
 		{
+		case dev::eth::Instruction::PUSH1:	code += rndByteSequence(1);	break;
```
chriseth: I guess it would be better to use something like
```
if (int(dev::eth::Instruction::PUSH1) <= int(_opcode) && int(_opcode) <= int(dev::eth::Instruction::PUSH32))
  code += rndByteSequence(int(_opcode) - int(dev::eth::Instruction::PUSH1) + 1);
```

test/fuzzTesting/fuzzHelper.cpp
```diff
@@ -143,16 +151,136 @@ std::string RandomCode::fillArguments(dev::eth::Instruction _opcode, RandomCodeO
...
+				case dev::eth::Instruction::SWAP13:
+				case dev::eth::Instruction::DUP14:	times = 14; break;
+				case dev::eth::Instruction::SWAP14:
+				case dev::eth::Instruction::DUP15:	times = 15; break;
+				case dev::eth::Instruction::SWAP15:
+				case dev::eth::Instruction::DUP16:	times = 16; break;
+				case dev::eth::Instruction::SWAP16:	times = 17; break;
+				default: times = 1;
+			}
+			for (int i = 0; i < times; i ++)
```
chriseth: Hm, this swaps/dups only constants we pushed just before the dup/swap, is that correct?
Wouldn't it be better to somehow check and ensure that the stack has a certain minimum height?

winsvega: we are creating the stack for this commands below

chriseth: Yes, but you create it up to the full height, i.e. you never swap or dup values that are results of other expressions, you only swap and dup constants.

winsvega: it might happen. cause some previous random code might have puted smth to memory.
in this switch case I am filling the stack for opcode. it is containing addresses to memory.
the actual memory could be filled with smth or not.

test/fuzzTesting/fuzzHelper.cpp
winsvega: that strange. I remember that I resolved this issue

CJentzsch:
I support Christians comments on thatm but think it is ok to merge

PR #2343, opened 2015-07-01
rpc fixes

debris:
fixed some of the functions that behaved differently than in go implementation


test/libweb3jsonrpc/webthreestubclient.h
```diff
@@ -166,43 +166,43 @@ class WebThreeStubClient : public jsonrpc::Client
             else
                 throw jsonrpc::JsonRpcException(jsonrpc::Errors::ERROR_CLIENT_INVALID_RESPONSE, result.toStyledString());
         }
-        std::string eth_getBlockTransactionCountByHash(const std::string& param1) throw (jsonrpc::JsonRpcException)
+        Json::Value eth_getBlockTransactionCountByHash(const std::string& param1) throw (jsonrpc::JsonRpcException)
```
debris: this file is autogenerated.

libethereum/ClientBase.h
```diff
@@ -149,6 +149,10 @@ class ClientBase: public Interface
 	/// Get the coinbase address
 	virtual Address address() const override;
 
+	virtual bool isKnown(h256 _hash) const override;
```
gavofyork: `h256 const&`

libethereum/ClientBase.h
```diff
@@ -149,6 +149,10 @@ class ClientBase: public Interface
...
+	virtual bool isKnown(BlockNumber _block) const override;
+	virtual bool isKnownTransaction(h256 _transactionHash) const override;
+
```
gavofyork: `h256 const&`

libethereum/Interface.h
```diff
@@ -133,12 +133,15 @@ class Interface
...
 
+	virtual bool isKnownTransaction(h256 _transactionHash) const = 0;
 	virtual Transaction transaction(h256 _transactionHash) const = 0;
 	virtual std::pair<h256, unsigned> transactionLocation(h256 const& _transactionHash) const = 0;
 	virtual h256 hashFromNumber(BlockNumber _number) const = 0;
 	virtual BlockNumber numberFromHash(h256 _blockHash) const = 0;
 	virtual int compareBlockHashes(h256 _h1, h256 _h2) const = 0;
 
+	virtual bool isKnown(BlockNumber _block) const = 0;
+	virtual bool isKnown(h256 _hash) const = 0;
```
gavofyork: `h256 const&`

libethereum/Interface.h
```diff
@@ -133,12 +133,15 @@ class Interface
 
 	// [BLOCK QUERY API]
 
+	virtual bool isKnownTransaction(h256 _transactionHash) const = 0;
```
gavofyork: `h256 const&`

gavofyork:
don't forget the `const&`s

libweb3jsonrpc/WebThreeStubServerBase.cpp
```diff
@@ -174,10 +178,14 @@ string WebThreeStubServerBase::eth_getBlockTransactionCountByHash(string const&
...
 
-string WebThreeStubServerBase::eth_getBlockTransactionCountByNumber(string const& _blockNumber)
+Json::Value WebThreeStubServerBase::eth_getBlockTransactionCountByNumber(string const& _blockNumber)
 {
 	try
 	{
+		BlockNumber blockNumber = jsToBlockNumber(_blockNumber);
+		if (!client()->isKnown(blockNumber))
+			return Json::Value(Json::nullValue);
+
```
chriseth: This might be a bit off, but can it happen that between line 186 and 189, we switch to a new fork which has a smaller block number count (but higher total difficulty) such that the block number is not valid anymore?

chriseth: Looking at the code showed that it will return a `bytes()` for the block and thus throw an RLP exception at some point. Is that bad here?

debris: > This might be a bit off, but can it happen that between line 186 and 189, we switch to a new fork which has a smaller block number count (but higher total difficulty) such that the block number is not valid anymore?

It's possible. 

> Looking at the code showed that it will return a bytes() for the block and thus throw an RLP exception at some point. Is that bad here?

We agreed with go team that for not existing blocks, we will return `null` here. Better solution requires significant refactor of `Client` and throwing/handling different kind of exceptions.

libethereum/BlockChain.h
```diff
@@ -178,6 +178,9 @@ class BlockChain
 	std::vector<unsigned> withBlockBloom(LogBloom const& _b, unsigned _earliest, unsigned _latest) const;
 	std::vector<unsigned> withBlockBloom(LogBloom const& _b, unsigned _earliest, unsigned _latest, unsigned _topLevel, unsigned _index) const;
 
+	/// Returns true if transaction is known. Thread-safe
+	bool isKnownTransaction(h256 const& _transactionHash) const { TransactionAddress ta = queryExtras<TransactionAddress, ExtraTransactionAddress>(_transactionHash, m_transactionAddresses, x_transactionAddresses, NullTransactionAddress); return !!ta; }
```
chriseth: No assignment needed.

PR #2362, opened 2015-07-02
Use size_t for RLP and decode length more carefully.

chriseth:
Length and index values now always use `size_t`.

Since RLP in theory allows to encode length values of up to `2**(2**64)`, we need to guard against the fact that they do not fit our types. Due to this reason, RLPs that contain length values of more than `std::numeric_limits<size_t>::max() - 0x101` are rejected with an "UndersizeRLP" exception (even if at a higher level, the flags ThrowOnFail / FailIfTooSmall were not given). The rationale behind this is that such an RLP has to be undersized, because it will not fit in memory. Subtracting `0x101` is done because in some circumstances, the size of the length is added to the length at a later point.

### DIFF
--- a/libdevcore/RLP.cpp
+++ b/libdevcore/RLP.cpp
@@ -246,7 +246,6 @@ void RLPStream::noteAppended(size_t _itemCount)
 				m_out[p] = (byte)(c_rlpListStart + s);
 			else if (c_rlpListIndLenZero + brs <= 0xff)
 			{
-
 				m_out[p] = (byte)(c_rlpListIndLenZero + brs);
 				byte* b = &(m_out[p + brs]);
 				for (; s; s >>= 8)
--- a/libdevcore/RLP.cpp
+++ b/libdevcore/RLP.cpp
@@ -50,7 +50,7 @@ RLP::iterator& RLP::iterator::operator++()
 	if (m_remaining)
 	{
 		m_currentItem.retarget(m_currentItem.next().data(), m_remaining);
-		m_currentItem = m_currentItem.cropped(0, RLP(m_currentItem, ThrowOnFail | FailIfTooSmall).actualSize());
+		m_currentItem = m_currentItem.cropped(0, actualSizeOfPrefix(m_currentItem));
 		m_remaining -= std::min<size_t>(m_remaining, m_currentItem.size());
 	}
 	else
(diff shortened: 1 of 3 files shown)


subtly:
To prevent regression and for interop, could you implement test(s) demonstrating issue?
Does this disallow RLP which was previously allowed?

subtly:
Please write unit test for this so before/after effects are obvious and for regression testing in case RLP is ever refactored or updated. And perhaps create a GH issue for integration/json test for other clients?

chriseth:
It is not really possible to write a test where decoding succeeded before but fails now, because it would need an RLP whose size is roughly `2**sizeof(size_t)`. I.e. it is impossible on a 64 bit machine and not easy on a 32 bit machine (let alone the fact that the build bot does not run the tests in 32 bit mode).

libdevcore/RLP.cpp
```diff
@@ -232,19 +244,22 @@ void RLPStream::noteAppended(unsigned _itemCount)
 			memmove(m_out.data() + p + encodeSize, m_out.data() + p, os - p);
 			if (s < c_rlpListImmLenCount)
 				m_out[p] = (byte)(c_rlpListStart + s);
-			else
+			else if (c_rlpListIndLenZero + brs <= 0xff)
 			{
+
```
gavofyork: random empty line?

gavofyork:
looks fine other than the minor style issue.

libdevcore/RLP.h
```diff
@@ -316,20 +319,23 @@ class RLP
...
 	/// @returns the number of bytes into the data that the payload starts.
-	unsigned payloadOffset() const { return isSingleByte() ? 0 : (1 + lengthSize()); }
+	size_t payloadOffset() const { return isSingleByte() ? 0 : (1 + lengthSize()); }
 
 	/// @returns the number of data items.
-	unsigned items() const;
+	size_t items() const;
+
+	/// @returns the "actualSize" of the RLP encoded in the largest prefix of @a _data and throws on error.
+	static size_t actualSizeOfPrefix(bytesConstRef _data) { return RLP(_data, ThrowOnFail | FailIfTooSmall).actualSize(); }
```
subtly: The size of the prefix is actualSize() - length(). Could this method be renamed to actualSize or actualSizeOfRLP?

subtly:
Please rename static method to be consistent with the method it's calling. The method does not return the "actual size of prefix". It returns the payloadoffset + payload, where payloadoffset is equivalent to the size of the prefix.

https://gist.github.com/subtly/bacbd50f6c118949688f

PR #2364, opened 2015-07-02
Issues: BlockChain::rebuild + byteRef test

winsvega:


test/libdevcore/core.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <libdevcore/CommonIO.h>
+#include <test/TestHelper.h>
+
+BOOST_AUTO_TEST_SUITE(CoreLibTests)
+
+BOOST_AUTO_TEST_CASE(byteRef)
+{	
+	cnote << "bytesRef copyTo and toString...";
+	dev::bytes originalSequence = dev::fromHex("0102030405060708091011121314151617181920212223242526272829303132");
+	dev::bytesRef _out(&originalSequence.at(0), 32);
```
gavofyork: `_out` -> `out`

test/libdevcore/core.cpp
```diff
@@ -0,0 +1,41 @@
...
+	dev::h256 ret("1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347");
```
gavofyork: `ret` is customarily used only for a value that ultimately is `return`ed. `expected` instead?

PR #2367, opened 2015-07-02
Removed incorrect blockchain sync invariant checks

arkpar:


CJentzsch:
although I cant judge the network logic, have not looked into it yet.

PR #2372, opened 2015-07-03
Use libsepc256k1 for sign, recovery, and verify.

subtly:
Sorry, not yet for windows yet.


libdevcrypto/Common.cpp
```diff
@@ -184,17 +183,38 @@ bytes dev::decryptAES128CTR(bytesConstRef _k, h128 const& _iv, bytesConstRef _ci
...
+#else
 	return s_secp256k1pp.recover(_sig, _message.ref());
+#endif
 }
 
 Signature dev::sign(Secret const& _k, h256 const& _hash)
 {
+#ifdef ETH_HAVE_SECP256K1
+	Signature s;
+	if (!secp256k1_ecdsa_sign_compact(_hash.data(), h256::size, s.data(), _k.data(), Nonce::get().data(), (int*)(s.data()+64)))
```
chriseth: Is `(int*)(s.data() + 64)` safe? Does it even work (big endian / little endian)?
Perhaps it would be better to use a local integer for it and manually put it into s after the call?

test/libp2p/peer.cpp
```diff
@@ -55,7 +55,7 @@ BOOST_AUTO_TEST_CASE(host)
 	auto node2 = host2.id();
 	host2.start();
 	
-	while (!host2.isStarted())
+	while (!host2.haveNetwork())
```
chriseth: Does not look like it should belong here.

subtly: Tests keep failing w/o it and network handshake depends on ecies :)

libdevcrypto/Common.cpp
subtly: It's working, not sure about LE. I'll update to set to a local.

test/libp2p/peer.cpp
chriseth: ok

subtly:
@chriseth Fixed issue with integer.

PR #2373, opened 2015-07-03
Catch boost exceptions and provide better diagnostics

LefterisJP:


libethereum/Utility.cpp
```diff
@@ -126,9 +126,9 @@ void dev::eth::upgradeDatabase(std::string const& _basePath)
 				}
 			}
 		}
-		catch (...)
+		catch (boost::exception const& _e)
```
chfast: You will not catch all exceptions any more.

LefterisJP:
As far as I know we put considerable effort into weeding out non-boost exceptions from our codebase. All exceptions thrown should be boost exceptions. If not they need fixing.

PR #2375, opened 2015-07-03
RLPTests: refactoring

winsvega:
So the RLP 'in' section could now be VALID or INVALID:

Example:
{
	"validRLP": {
		"in": "VALID", 
		"out": "c7c0c1c0c3c0c1c0"
	},
}
This means that the client should successfully read bytes from 'out' and create a valid RLP class from it.

{
	"invalidRLP": {
		"in": "INVALID", 
		"out": "bf0f000000000000021111"
	},
}
In this example an attempt to create RLP class from bytes from 'out' should throw an RLP exception as the bytes in 'out' section does not represent a valid RLP

### DIFF
--- a/test/libdevcore/rlp.cpp
+++ b/test/libdevcore/rlp.cpp
@@ -43,6 +43,12 @@ namespace dev
 	{	
 		void buildRLP(js::mValue& _v, RLPStream& _rlp);
 		void checkRLPAgainstJson(js::mValue& v, RLP& u);
+		enum class RlpType
+		{
+			Valid,
+			Invalid,
+			Test
+		};
 
 		void doRlpTests(json_spirit::mValue& v, bool _fillin)
 		{
@@ -55,7 +61,7 @@ namespace dev
 					continue;
 				}
 
-				std::cout << "  " << i.first << std::endl;
+				cout << "  " << i.first << endl;
 				TBOOST_REQUIRE((o.count("out") > 0));
 				TBOOST_REQUIRE((!o["out"].is_null()));
 
@@ -84,25 +90,25 @@ namespace dev
 				{
 					//Check Encode
 					TBOOST_REQUIRE((o.count("in") > 0));
-					int skipEncode = 0;
+					RlpType rlpType = RlpType::Test;
 					if (o["in"].type() == js::str_type)
 					{
 						if (o["in"].get_str() == "INVALID")
-							skipEncode = 1;
+							rlpType = RlpType::Invalid;
 						else	if (o["in"].get_str() == "VALID")
-							skipEncode = 2;
+							rlpType = RlpType::Valid;
 					}
 
-					if (skipEncode == 0)
+					if (rlpType == RlpType::Test)
 					{
 						RLPStream s;
 						dev::test::buildRLP(o["in"], s);
 						string computedText = toHex(s.out());
 
-						std::string expectedText(o["out"].get_str());
-						std::transform(expectedText.begin(), expectedText.end(), expectedText.begin(), ::tolower );
+						string expectedText(o["out"].get_str());
+						transform(expectedText.begin(), expectedText.end(), expectedText.begin(), ::tolower );
 
-						std::stringstream msg;
+						stringstream msg;
 						msg << "Encoding Failed: expected: " << expectedText << std::endl;
 						msg << " But Computed: " << computedText;
 						TBOOST_CHECK_MESSAGE(
@@ -123,7 +129,7 @@ namespace dev
 						bytes payloadToDecode = fromHex(o["out"].get_str());
 						RLP payload(payloadToDecode);
 
-						if (skipEncode == 0)
+						if (rlpType == RlpType::Test)
 							dev::test::checkRLPAgainstJson(inputData, payload);
 					}
 					catch (Exception const& _e)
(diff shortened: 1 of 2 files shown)


test/libdevcore/rlp.cpp
```diff
@@ -132,71 +192,62 @@ namespace dev
...
 
-		js::mValue& inputData = o["in"];
-		bytes payloadToDecode = fromHex(o["out"].get_str());
+	string testPath = dev::test::getTestPath();
+	testPath += "/BasicTests/RandomRLPTests";
 
-		RLP payload(payloadToDecode);
+	vector<boost::filesystem::path> testFiles;
+	boost::filesystem::directory_iterator iterator(testPath);
+	for(; iterator != boost::filesystem::directory_iterator(); ++iterator)
```
chriseth: Range-based for loop?

test/libdevcore/rlp.cpp
```diff
@@ -40,8 +40,86 @@ namespace js = json_spirit;
...
+					{
+						cnote << "rlp exception: " << _e.what();
+						was_exception = true;
+					}
+
+					if (skipEncode && was_exception)
+						continue;
+
+					if (was_exception)
+						TBOOST_ERROR("Unexpected RLP Exception!");
```
chriseth: Why not just use BOOST_CHECK_NO_THROW?

test/libdevcore/rlp.cpp
```diff
@@ -40,8 +40,86 @@ namespace js = json_spirit;
 namespace dev
 {
 	namespace test
-	{
-		static void buildRLP(js::mValue& _v, RLPStream& _rlp)
+	{	
+		void buildRLP(js::mValue& _v, RLPStream& _rlp);
+		void checkRLPAgainstJson(js::mValue& v, RLP& u);
+
+		void doRlpTests(json_spirit::mValue& v, bool _fillin)
```
chriseth: Could you make that function a bit shorter?

winsvega: In a what way?

chriseth: By creating shorter functions that are called from this function. Please also explain the meaning of the `_fillin` parameter.

winsvega: the fillin option works just as in other suites. if test provided with only 'out' section declared and fillin option set to true, It would try to create RLP class from the 'out' field. If succeed the 'in' field would be filled as VALID, if not - as INVALID.  

I would then generate random RLP byte sequences to see if they are correct with fillin option.

test/libdevcore/rlp.cpp
```diff
@@ -132,71 +225,62 @@ namespace dev
...
-
-		std::string expectedText(o["out"].get_str());
-		std::transform(expectedText.begin(), expectedText.end(), expectedText.begin(), ::tolower );
-
-		const std::string& computedText = toHex(s.out());
-
-		std::stringstream msg;
-		msg << "Encoding Failed: expected: " << expectedText << std::endl;
-		msg << " But Computed: " << computedText;
+	dev::test::executeTests("invalidRLPtest", "/BasicTests", dev::test::getFolder(__FILE__) + "/StateTestsFiller", dev::test::doRlpTests);
```
CJentzsch: why StateTestFiller?

test/libdevcore/rlp.cpp
```diff
@@ -132,71 +225,62 @@ namespace dev
...
 
-		BOOST_CHECK_MESSAGE(
-			expectedText == computedText,
-			msg.str()
-			);
-	}
 
+BOOST_AUTO_TEST_CASE(rlptest)
+{
+	dev::test::executeTests("rlptest", "/BasicTests", dev::test::getFolder(__FILE__) + "/StateTestsFiller", dev::test::doRlpTests);
```
CJentzsch: why StateTestFiller

CJentzsch:
Please make a little wiki entry on the test wiki to explain that.

winsvega:
https://github.com/ethereum/tests/wiki/RLP-Tests

test/libdevcore/rlp.cpp
```diff
@@ -40,8 +40,119 @@ namespace js = json_spirit;
...
+			for (auto& i: v.get_obj())
+			{
+				js::mObject& o = i.second.get_obj();
+				if (test::Options::get().singleTest && test::Options::get().singleTestName != i.first)
+				{
+					o.clear();
+					continue;
+				}
+
+				std::cout << "  " << i.first << std::endl;
```
CJentzsch: you have namespace std declared, so you don't need std::

test/libdevcore/rlp.cpp
```diff
@@ -40,8 +40,119 @@ namespace js = json_spirit;
...
+				else
+				{
+					//Check Encode
+					TBOOST_REQUIRE((o.count("in") > 0));
+					int skipEncode = 0;
+					if (o["in"].type() == js::str_type)
+					{
+						if (o["in"].get_str() == "INVALID")
+							skipEncode = 1;
+						else
```
CJentzsch: have `else if` on the same line

test/libdevcore/rlp.cpp
```diff
@@ -40,8 +40,119 @@ namespace js = json_spirit;
...
+					{
+						cnote << "rlp exception: " << _e.what();
+						o["in"] = "INVALID";
+					}
+				}
+				else
+				{
+					//Check Encode
+					TBOOST_REQUIRE((o.count("in") > 0));
+					int skipEncode = 0;
```
CJentzsch: why can't you use a bool for that?

winsvega: because code skiped in two diffrent scenarious. 
1) if invalid test
2) if valid test
the logic of expetions is diffrent below

test/libdevcore/rlp.cpp
winsvega: I remember it was required by the gav style to always type std::  and other namespaces
and avoid using namespace.

test/libdevcore/rlp.cpp
CJentzsch: than you should use an enum

PR #2376, opened 2015-07-03
TransactionQueue priority queue

arkpar:


libethereum/State.cpp
```diff
@@ -535,57 +535,54 @@ pair<TransactionReceipts, bool> State::sync(BlockChain const& _bc, TransactionQu
...
-									cnote << i.first << "Dropping old transaction (gas price lower)";
-									_tq.drop(i.first);
-								}
+								if (mt.nonce() < t.nonce())
+									cnote << t.sha3() << "Dropping old transaction (nonce too low)";
+								else if (mt.nonce() == t.nonce() && mt.gasPrice() <= t.gasPrice())
+									cnote << t.sha3() << "Dropping old transaction (gas price lower)";
 							}
+						}
+						_tq.drop(t.sha3());
```
arkpar: Pleas check this. I'm dropping every old transaction here, otherwise they will stay in queue and be attempted to execute again.

gavofyork: The intent of the original logic was to keep old transactions, currently in the State, with a higher gas price around, since if we don't mine this block, we will have a chance to reorder the next block with the higher-gas-price transaction and drop the one which is currently up for mining.

If the block is mined, then it may stick around in the queue (until restart, anyway), however it's very much an edge case, and is an acceptably small issue. After refactoring the tx queue, it'll be good to have a cleaning pass after each block which removes these opportunistic transactions (same sender, lower nonce).

arkpar: There are quite a lot of transactions coming from the network with outdated nonces (less than last block account state).  This function gets transactions ordered by nonce and tries to import them. If  an exception is thrown with nonce being too old, this means it won't succeed even if the state is cleared. It can only succeed if a blockchain is reorganized to a previous block, but his case is handled in Client.cpp

gavofyork: So it shouldn't keep them unless the nonce is

PR #2378, opened 2015-07-03
Initial version of the rating system

gluk256:


gluk256:
I had to write my own bloom() function, because the exsisting FixedHash::bloomPart() contains a bug (which I will fix in a separate PR). In addition, my custom function is better specialized for the Whisper protocol.

gluk256:
Also, I have fixed a bug in  Envelope::proveWork() / Envelope::workProved().

romanman:
@gluk256 : hey Vlad, do you have approve from gav for the bloom filter change ?

gluk256:
I did not change the function, I just introduced another function (to be use in Whisper exclusively). There was not bloom filter in Whisper before I started working on it, therefore it's not going to affect anything.

However, the bug in FixedHash::bloomPart() will be fixed by a static_assert, so that it will be no longer possible to go out of range.

romanman:
@gluk256 : cool I am just asking if you are sure that the function 
you just introduced will be used in Whisper and not the old one

gluk256:
That's my intention.

romanman:
@gluk256 : ok cool we will take it

gluk256:
Roman, I would advise you to wait, till it is merged.

romanman:
@gluk256 : cool

libwhisper/BloomFilter.h
```diff
@@ -83,12 +86,51 @@ void TopicBloomFilterBase<N>::removeRaw(FixedHash<N> const& _h)
 
 template <unsigned N>
 bool TopicBloomFilterBase<N>::isBitSet(FixedHash<N> const& _h, unsigned _index)
-{	
+{
 	unsigned iByte = _index / 8;
-	unsigned iBit = _index % 8;
+	unsigned iBit = _index & 0x7;
```
guanqun: Any reason to change this to an "or" operation? I know it's the same, but "% 8" is clear for me as the previous line is "/ 8".

gluk256: My consideration was efficiency.

guanqun: I think for current compiler such hand-made simplification is un-necessary. Compiler can do it for us.

guanqun:
```
➜  Repos g++ -O3 main.cpp -o main
➜  Repos cat main.cpp
unsigned foo(unsigned x)
{
	return x % 8;
}

int main()
{
	unsigned int x;
	unsigned int y = foo(x);

	return 0;
}
➜  Repos g++ -O3 main.cpp -o main
➜  Repos otool -tV main
main:
(__TEXT,__text) section
__Z3fooj:
0000000100000f80	pushq	%rbp
0000000100000f81	movq	%rsp, %rbp
0000000100000f84	andl	$0x7, %edi
0000000100000f87	movl	%edi, %eax
0000000100000f89	popq	%rbp
0000000100000f8a	retq
0000000100000f8b	nopl	(%rax,%rax)
_main:
0000000100000f90	pushq	%rbp
0000000100000f91	movq	%rsp, %rbp
0000000100000f94	xorl	%eax, %eax
0000000100000f96	popq	%rbp
0000000100000f97	retq
```

libwhisper/WhisperPeer.cpp
```diff
@@ -121,17 +121,21 @@ unsigned WhisperPeer::ratingForPeer(Envelope const& e) const
...
 
-	DEV_GUARDED(x_bloom)
-		if (e.matchesBloomFilter(m_bloom))
-			rating += BloomFilterMatchReward;
-	
-	rating += e.sha3().firstBitSet();
+	if (e.matchesBloomFilter(bloom()))
+		++rating;
+
+	rating *= 256;
```
subtly: Considering implementing ratingForPeer in envelope or host since it's used by both WhisperPeer and WhisperHost. I assume this will come later.

gluk256: It will be different in Host, because there we can add another criterium: matching one or more of our actual watches. Peer can not possibly know that and will only use bloom filter, which is only probabilistic.

libwhisper/BloomFilter.h
```diff
@@ -83,12 +86,51 @@ void TopicBloomFilterBase<N>::removeRaw(FixedHash<N> const& _h)
...
+{
+	unsigned iByte = _index / 8;
+	unsigned iBit = _index & 0x7;
+	_h[iByte] |= c_powerOfTwoBitMmask[iBit];
+}
+
+template <unsigned N>
+FixedHash<N> TopicBloomFilterBase<N>::bloom(AbridgedTopic const& _h)
+{
+	// The size of AbridgedTopic is 32 bits, and 27 of them participate in this algorithm.
```
gavofyork: `_h` is a cryptographic hash. All bits are independent and contain close to full entropy. There is no reason to utilise more bits than that required for placing the 1s in the bloom. I'm against using two algorithms for the same thing; blooms in Ethereum proper have had many more eyes and I consider them more secure.

libwhisper/BloomFilter.h
```diff
@@ -83,12 +86,51 @@ void TopicBloomFilterBase<N>::removeRaw(FixedHash<N> const& _h)
...
+
+	// We need to review the algorithm if any of the following constants will be changed.
+	static_assert(4 == AbridgedTopic::size, "wrong template parameter in TopicBloomFilterBase<N>::bloom()");
+	static_assert(3 == BitsPerBloom, "wrong template parameter in TopicBloomFilterBase<N>::bloom()");
+
+	FixedHash<N> ret;
+
+	if (TopicBloomFilterSize == N)
```
gavofyork: This new algo should be formally specified so it can be easily reviewed.

gluk256:
Specification added:
https://github.com/ethereum/wiki/wiki/Whisper-PoC-1-Protocol-Spec

Issue #2078, opened 2015-06-03
Solidity: Exponential notation for constants.


chriseth:
The following code results in an unexpected exception:
```
contract x {
    function f() {
        var a = 1e20;
    }
}
```

guanqun:
Literally  "1e20" can't be converted to cpp_int directly, that's why it crashes. If we allow this, we might need to evaluate to cpp_float first and then convert to cpp_int.

PR #2379, opened 2015-07-04
fix the typo

guanqun:
I think we're missing a "not" here. :)


gavofyork:
surely it means::

'e'/'E' must be scanned [only] as part of [a] hex number [because we don't handle exponentiation in decimal numbers]

guanqun:
Oh, I misunderstood... Thank you.

I was looking at this issue #2078, do we need to support exponentiation in decimal number?

PR #2381, opened 2015-07-04
Adding argument for peer drop multiplier

LefterisJP:


subtly:
TL;DR Let me know what you'd like to do or change and I'll submit a PR...
There's quite a bit planned here. What is this for?
Please use network preferences; this won't be exposed by web3 for quite awhile, until the network is robust enough to handle user-supplied values.

LefterisJP:
As we discussed in Skype chat maybe just document in a comment that there is a minimum amount of peers (9) and explain why, so that our users and other developers (like me :) ) can realize what's happening when looking at the code.

PR #2384, opened 2015-07-05
Txqueueasyncverify

gavofyork:


libethereum/TransactionQueue.cpp
```diff
@@ -31,12 +31,34 @@ using namespace dev::eth;
 const char* TransactionQueueChannel::name() { return EthCyan "┉┅▶"; }
 const char* TransactionQueueTraceChannel::name() { return EthCyan " ┅▶"; }
 
+TransactionQueue::TransactionQueue()
+{
+	// Allow some room for other activity
+	unsigned verifierThreads = std::max(thread::hardware_concurrency(), 3U) - 2U;
```
arkpar: do we really need more than one thread here?

libethereum/TransactionQueue.cpp
```diff
@@ -252,16 +336,20 @@ void TransactionQueue::noteGood(std::pair<h256, Transaction> const& _t)
...
-void TransactionQueue::drop(h256 const& _txHash)
+void TransactionQueue::drop(h256 const& _txHash, ImportResult _ir)
 {
 	UpgradableGuard l(m_lock);
 
 	if (!m_known.count(_txHash))
 		return;
 
+	if (m_callbacks.count(_txHash) && m_callbacks[_txHash])
+		m_callbacks[_txHash](_ir);
```
arkpar: These callbacks don't seem to be used currently

gavofyork: indeed - they probably will be in the future though.

libethereum/TransactionQueue.cpp
gavofyork: don't really see any harm. if there's 256 incoming transactions, they'll help and if there's only one, then 5 will never get woken up anyway.

PR #2386, opened 2015-07-05
Parallel transaction import/verification

arkpar:
Multithreaded transaction verification and ec recovery. This is required for the case we get a large amount  (thousands) of transactions from the network to free networking thread for other activities.

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -604,19 +604,18 @@ void Client::onChainChanged(ImportRoute const& _ir)
 		for (auto const& t: m_bc.transactions(h))
 		{
 			clog(ClientTrace) << "Resubmitting dead-block transaction " << Transaction(t, CheckTransaction::None);
-			m_tq.import(t, TransactionQueue::ImportCallback(), IfDropped::Retry);
+			m_tq.import(t, IfDropped::Retry);
 		}
 	}
 
 	// remove transactions from m_tq nicely rather than relying on out of date nonce later on.
 	for (auto const& h: _ir.liveBlocks)
-	{
 		clog(ClientTrace) << "Live block:" << h;
-		for (auto const& th: m_bc.transactionHashes(h))
-		{
-			clog(ClientTrace) << "Safely dropping transaction " << th;
-			m_tq.drop(th);
-		}
+
+	for (auto const& t: _ir.goodTranactions)
+	{
+		clog(ClientTrace) << "Safely dropping transaction " << t.sha3();
+		m_tq.dropGood(t);
 	}
 
 	if (auto h = m_host.lock())
@@ -651,7 +650,7 @@ void Client::onChainChanged(ImportRoute const& _ir)
 				for (auto const& t: m_postMine.pending())
 				{
 					clog(ClientTrace) << "Resubmitting post-mine transaction " << t;
-					auto ir = m_tq.import(t, TransactionQueue::ImportCallback(), IfDropped::Retry);
+					auto ir = m_tq.import(t, IfDropped::Retry);
 					if (ir != ImportResult::Success)
 						onTransactionQueueReady();
 				}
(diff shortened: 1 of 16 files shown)


libethereum/Client.cpp
```diff
@@ -604,19 +604,18 @@ void Client::onChainChanged(ImportRoute const& _ir)
...
-		for (auto const& th: m_bc.transactionHashes(h))
-		{
-			clog(ClientTrace) << "Safely dropping transaction " << th;
-			m_tq.drop(th);
-		}
+
+	for (auto const& t: _ir.goodTranactions)
+	{
+		clog(ClientTrace) << "Safely dropping transaction " << t.sha3();
+		m_tq.dropGood(t);
```
arkpar: In case block transaction was not seen in the queue before and there are future transactions for the same account we need to move future transactions to current here.

arkpar:
Documentation + cleaup in the following PR

chfast:
The title is intriguing, but very hard to review without any top level description.

libethereum/TransactionQueue.cpp
```diff
@@ -320,3 +358,43 @@ void TransactionQueue::clear()
...
 	m_futureSize = 0;
 }
+
+void TransactionQueue::enqueue(RLP const& _data, h512 const& _nodeId)
+{
+	unique_lock<Mutex> l(x_queue);
+	unsigned itemCount = _data.itemCount();
+	for (unsigned i = 0; i < itemCount; ++i)
+		m_unverified.emplace_back(UnverifiedTransaction(_data[i].data(), _nodeId));
+	m_queueReady.notify_all();
```
chfast: You should release the lock before notify.

libethereum/TransactionQueue.cpp
```diff
@@ -320,3 +358,43 @@ void TransactionQueue::clear()
 	m_future.clear();
 	m_futureSize = 0;
 }
+
+void TransactionQueue::enqueue(RLP const& _data, h512 const& _nodeId)
+{
+	unique_lock<Mutex> l(x_queue);
```
chfast: Why unique_lock? lock_guard is enough.

libethereum/TransactionQueue.cpp
```diff
@@ -320,3 +358,43 @@ void TransactionQueue::clear()
...
+
+void TransactionQueue::verifierBody()
+{
+	while (!m_aborting)
+	{
+		UnverifiedTransaction work;
+
+		{
+			unique_lock<Mutex> l(x_queue);
+			m_queueReady.wait(l, [&](){ return !m_unverified.empty() || m_aborting; });
```
chfast: I also need a concurrent priority queue in SmartVM. I can invest some time to extract it into a reusable form. However I would like to have a mutable one.

arkpar: This here is not a priority queue, just a simple FIFO

chfast: I was suggested by PriorityCompare. Never the less, I'm currently using FIFO to as it does not change much. But the sync mechanism is identical.

arkpar: PriorityCompare is used in a different queue, which is not concurrent (not a blocking queue)

libethereum/TransactionQueue.cpp
```diff
@@ -31,7 +31,28 @@ using namespace dev::eth;
...
+		for (unsigned i = 0; i < verifierThreads; ++i)
+			m_verifiers.emplace_back([=](){
+				setThreadName("txcheck" + toString(i));
+				this->verifierBody();
+			});
+}
+
+TransactionQueue::~TransactionQueue()
+{
+	m_aborting = true;
```
chfast: You need to lock before changing the flag.

gavofyork: To synchronise what?

chfast: The read of m_aborting in https://github.com/ethereum/cpp-ethereum/pull/2386/files#diff-37a6e0c4c4565fc12c93ad4e9f6445f6R381

PR #2387, opened 2015-07-06
Fix up rvalue references

chfast:


gavofyork:
does any of this result in substantially faster operation?

chfast:
StateTests go 0.8% faster :)

chfast:
The description is a bit misleading. The main point of this is to avoid double code copy in ExtVM creation.

PR #2390, opened 2015-07-06
FixedHash::bloomPart() bugfix: out of range

gluk256:
In case of Whisper (P=3, M=64) variable ++p would be incremented 192 times and go out of range. If we substituted (*p) with (m_data[x]), then std::array would have thrown exception. But in our case there was no warning and the program would execute normally. 
I did not do it now, because I prefer compile-time assert to run-time.

### DIFF
--- a/libdevcore/FixedHash.h
+++ b/libdevcore/FixedHash.h
@@ -31,6 +31,10 @@
 namespace dev
 {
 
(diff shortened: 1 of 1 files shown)


gluk256:
Apparently, the errors are not mine.

LefterisJP:
adding some unit tests for this would also be nice

PR #2391, opened 2015-07-06
Increase stack size needed for VM execution in Debug build type

chfast:


libethereum/ExtVM.cpp
```diff
@@ -33,7 +33,12 @@ namespace
 static unsigned const c_depthLimit = 1024;
 
 /// Upper bound of stack space needed by single CALL/CREATE execution. Set experimentally.
-static size_t const c_singleExecutionStackSize = 12 * 1024;
+static size_t const c_singleExecutionStackSize =
+#ifdef NDEBUG
+	12 * 1024;
+#else
+	64 * 1024;
```
arkpar: Can you make it platform dependent? I've noticed a lot of threads are created here on linux during full blockchain sync.

chfast: I would like not to as we need it on Windons and OSX and testing on linux only. But I will try to improve it by reusing the same thread.

gavofyork: testing on linux can be done using clang. gcc never seemed to have a problem with stack size.

libethereum/ExtVM.cpp
```diff
@@ -33,7 +33,12 @@ namespace
 static unsigned const c_depthLimit = 1024;
 
 /// Upper bound of stack space needed by single CALL/CREATE execution. Set experimentally.
-static size_t const c_singleExecutionStackSize = 12 * 1024;
+static size_t const c_singleExecutionStackSize =
+#ifdef NDEBUG
```
subtly: Is NDEBUG set by cmake, and is it available on windows?

subtly:
@chfast This is what I used on macos for debug:

```
static size_t const c_singleExecutionStackSize = 33 * 1024;

/// Standard OSX thread stack limit. Should be reasonable for other platforms too.
static size_t const c_defaultStackSize = 512 * 1024;

/// Stack overhead prior to allocation.
static size_t const c_entryOverhead = 128 * 1024;

/// On what depth execution should be offloaded to additional separated stack space.
static unsigned const c_offloadPoint = (c_defaultStackSize - c_entryOverhead) / c_singleExecutionStackSize;
```

In debug macos/clang uses about 96KB prior to execution, so setting to 128 is conservative; with the overhead value in place the 33KB for (debug) execution stack size is more accurate. I'm not sure what the best values are for release, however, the previous values worked in release.

libethereum/ExtVM.cpp
chfast: It should be set by cmake by default in all non-debug modes. The macro is a standard because the assert implementation depends on it.

chfast:
Ok, this looks reasonable. I will use these values then or you can send it as a new PR.

Any objections for disabling this for GCC (not needed on linux)?

My implementation is definitely quite crappy. Last time I checked the CPU profile report 25% of the time has been spent in thread_start function hit ~5000 times. It probably was that.

I see 2 solutions:
1. Pull the copy of Boost.Context. Or create the implementation for both 1.55 and 1.58 versions.
2. Reuse the thread for offloading. It might be a thead pool or concurrent queue implemented with condition variable. That is more portable but also more complex. And an overkill as well because we exploiting the fact that in boost's implementation of thread we can set the stack size attribute. But besides of that we don't need a thread at all.

gavofyork:
for now, we can disable for GCC and (says @subtly) for windows (since we are able to set the stack size at compilation time).

if there's no way to alter the stack size of a running thread with clang, then i'm most comfortable with option 2.

chfast:
Updated.

subtly:
@chfast Looks good -- if we remove for windows, what do we need to set windows stack size to in cmake?  Was windows working w/4MB?

subtly:
@chfast Have no problems w/merging this. We should open another PR for windows though. I'm pretty sure I wasn't having any problems on windows in debug with the stack size set to 16MB. Don't remember testing 8MB or who set it to 32MB :)

subtly:
Marking good w/the assumption that moving #ifdef check for clang doesn't have any side effects (I didn't see any and tests pass).

chfast:
I will check the Windows tomorrow and adjust the stack size needed. The good test is `testeth -t StateTests` as it checks max VM call depth possible.

chfast:
Stack size on Windows increased to 16MB. It is enough to pass all tests in debug mode.

PR #2394, opened 2015-07-06
AZ: refresh blockchain only if visible

arkpar:


alethzero/MainWin.cpp
```diff
@@ -260,6 +260,8 @@ Main::Main(QWidget *parent) :
 	m_transact->setWindowFlags(Qt::Dialog);
 	m_transact->setWindowModality(Qt::WindowModal);
 
+	connect(ui->blockChainDockWidget, &QDockWidget::visibilityChanged, [=]() { refreshBlockChain(); });
```
chfast: `[=]` is to capture `this`?

arkpar: yes

PR #2396, opened 2015-07-06
Some initial tests for libdevcore/FixedHash

LefterisJP:
Just adding a file with a few tests. We can and should use this file to add more tests for FixedHash

### DIFF
--- a/libdevcore/FixedHash.h
+++ b/libdevcore/FixedHash.h
@@ -102,7 +102,7 @@ public:
 	FixedHash operator&(FixedHash const& _c) const { return FixedHash(*this) &= _c; }
 	FixedHash operator~() const { FixedHash ret; for (unsigned i = 0; i < N; ++i) ret[i] = ~m_data[i]; return ret; }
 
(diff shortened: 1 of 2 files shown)


libdevcore/FixedHash.h
```diff
@@ -102,7 +102,7 @@ class FixedHash
 	FixedHash operator&(FixedHash const& _c) const { return FixedHash(*this) &= _c; }
 	FixedHash operator~() const { FixedHash ret; for (unsigned i = 0; i < N; ++i) ret[i] = ~m_data[i]; return ret; }
 
-	/// @returns true if all bytes in @a _c are set in this object.
+	/// @returns true if all bits in @a _c are set in this object.
```
chriseth: "true if all one-bits in @a _c are also set in this object"?

chfast:
:+1:

libdevcore/FixedHash.h
LefterisJP: good clarification

PR #2401, opened 2015-07-06
Add block gas limit tests

CJentzsch:


gavofyork:
would be good to have these test the post 830k behaviour...

CJentzsch:
This is only the filler. I did push the post 830K test to the test repo.

PR #2413, opened 2015-07-07
levelDB for whisper

gluk256:


libwhisper/WhisperDB.cpp
```diff
@@ -0,0 +1,90 @@
...
+using namespace dev;
+using namespace dev::shh;
+using namespace dev::eth;
+
+WhisperDB::WhisperDB()
+{
+	string path = Defaults::dbPath();
+	boost::filesystem::create_directories(path);
+	ldb::Options o;
+	o.create_if_missing = true;
```
arkpar: max_opened_files option should be set

gluk256: what's wrong with the default value?

arkpar: default value is 1024, too many open handles per process on linux

libwhisper/WhisperDB.cpp
```diff
@@ -0,0 +1,90 @@
...
+	delete m_db;
+}
+
+bool WhisperDB::put(dev::h256 const& _key, string const& _value)
+{
+	string s = _key.hex();
+	string cropped = s.substr(s.size() - 8);
+	leveldb::Status status = m_db->Put(m_writeOptions, s, _value);
+	if (status.ok())
+		cdebug << "Whisper DB put:" << cropped << _value;
```
subtly: Less verbose logging here and below.

libwhisper/WhisperDB.cpp
```diff
@@ -0,0 +1,90 @@
...
+	else
+		cdebug << "Whisper DB put failed:" << status.ToString() << "key:" << cropped;
```
subtly: Throw exception here.

libwhisper/WhisperDB.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <string>
+#include <boost/filesystem.hpp>
+
+#include <libdevcore/Common.h>
+#include <libdevcore/CommonData.h>
+#include <libdevcore/Exceptions.h>
+#include <libdevcore/Log.h>
+#include <libdevcore/SHA3.h>
+#include <libdevcore/FileSystem.h>
+#include "WhisperDB.h"
```
chfast: Matching header should be the first one.

libwhisper/WhisperDB.h
```diff
@@ -0,0 +1,50 @@
...
+	~WhisperDB();
+
+	bool put(dev::h256 const& _key, std::string const& _value);
+	bool erase(dev::h256 const& _key);
+	std::string get(dev::h256 const& _key) const;
+
+private:
+	ldb::ReadOptions m_readOptions;
+	ldb::WriteOptions m_writeOptions;
+	ldb::DB* m_db = nullptr;
```
chfast: Please use std::unique_ptr instead of raw pointer.

libwhisper/WhisperDB.cpp
```diff
@@ -0,0 +1,90 @@
...
+using namespace dev::shh;
+
+WhisperDB::WhisperDB()
+{
+	string path = dev::getDataDir();
+	boost::filesystem::create_directories(path);
+	ldb::Options op;
+	op.create_if_missing = true;
+	op.max_open_files = 256;
+	ldb::DB::Open(op, path + "/whisper", &m_db);
```
chfast: Check status and at least log if not ok.

libwhisper/WhisperDB.cpp
```diff
@@ -0,0 +1,90 @@
...
+WhisperDB::~WhisperDB()
+{
+	delete m_db;
+}
+
+bool WhisperDB::put(dev::h256 const& _key, string const& _value)
+{
+	string s = _key.hex();
+	string cropped = s.substr(s.size() - 8);
+	leveldb::Status status = m_db->Put(m_writeOptions, s, _value);
```
chfast: Some bad things are happening here:
- converting the hash to hex format makes it twice as long. LevelDB can store any bytes so put the hash directly.
- You don't have to use std::string as an intermediate representation of the key. LevelDB does not use std::string directly, it uses Slice type similar to our vector_ref. Try to construct the Slice directly from h256.

libwhisper/WhisperDB.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+	return status.ok();
+}
+
+string WhisperDB::lookup(dev::h256 const& _key) const
+{
+	string ret;
+	string s = _key.hex();
+	string cropped = s.substr(s.size() - 8);
+	leveldb::Status status = m_db->Get(m_readOptions, s, &ret);
```
chfast: Some bad things are happening here:
- Converting the hash to hex format makes it twice as long. LevelDB can store any bytes so put the hash directly.
- You don't have to use std::string as an intermediate representation of the key. LevelDB does not use std::string directly, it uses Slice type similar to our vector_ref. Try to construct the Slice directly from h256.

gluk256:
@subtly: we don't need the commit/rollback functionality, because the class leveldb::DB takes care of that internally. It stores the changes in memory and only saves them to disc when the size exceeds leveldb::Options::write_buffer_size (default size is 4 MB -- we can change that, if necessary). In the future we might also set leveldb::Options::block_size (approximate size of user data packed per block), depending on the average Whisper message size.

libwhisper/WhisperDB.h
```diff
@@ -0,0 +1,55 @@
...
+#include <libdevcore/db.h>
+#include <libdevcore/FixedHash.h>
+#include "Common.h"
+
+namespace dev
+{
+namespace shh
+{
+
+struct FailedToOpenLevelDB: virtual Exception { FailedToOpenLevelDB(std::string _message = std::string()): Exception(_message) {} };
```
gavofyork: `std::string const&` is more conformant with the rest of the codebase.

libwhisper/WhisperDB.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <boost/filesystem.hpp>
+#include <libdevcore/FileSystem.h>
+
+using namespace std;
+using namespace dev;
+using namespace dev::shh;
+
+WhisperDB::WhisperDB()
+{
+	string path = dev::getDataDir();
```
gavofyork: `getDataDir(".shh")`, perhaps?

libwhisper/WhisperDB.cpp
```diff
@@ -0,0 +1,69 @@
...
+	boost::filesystem::create_directories(path);
+	leveldb::Options op;
+	op.create_if_missing = true;
+	op.max_open_files = 256;
+	leveldb::DB* p = nullptr;
+	leveldb::Status status = leveldb::DB::Open(op, path + "/whisper", &p);
```
gavofyork: `"/messages"`? (then the full path is `~/.shh/messages`)

libwhisper/WhisperDB.cpp
gluk256: i ahve changed to getDataDir("shh"), because it corresponds to the default value defined in function dev::getDataDir(std::string _prefix):
     if (_prefix.empty())
		_prefix = "ethereum";

gluk256:
fixed the latest issues.

gavofyork:
seeing errors in tests. if these tests will not reliably work on the buildbot, then they should be disabled from the run as they mask potentially critical issues.

gluk256:
the remaining errors are not mine

PR #2414, opened 2015-07-07
Cleanups, fixes and download viz.

gavofyork:


CJentzsch:
[20:12:22] Jeffrey Wilcke: no change at 830.000
[20:12:26] Jeffrey Wilcke: keep as is
[20:12:56] Christoph Jentzsch: ok, as is = all blockTest follow the new rule
[20:13:08] Jeffrey Wilcke: no new rule
[20:13:24] Christoph Jentzsch: so we will stay with the old rule for Frontier?
[20:13:35] Jeffrey Wilcke: indeed, and after

CJentzsch:
I marked it with `wontfixyet`, because it seems we don't have a consensus about that change between all the clients

PR #2417, opened 2015-07-08
use new nonce to calc contract address

CJentzsch:
Currently, when no nonce is given, we calculate a nonce here https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/ClientBase.cpp#L55 .
But for calculating the address of the new contract we use the `UndefinedU256` and therefore always returning the wrong contract address. This PR fixes this by returning the new nonce through a function parameter.

### DIFF
--- a/libethereum/ClientBase.cpp
+++ b/libethereum/ClientBase.cpp
@@ -45,14 +45,14 @@ State ClientBase::asOf(BlockNumber _h) const
 	return asOf(bc().numberHash(_h));
 }
 
-h256 ClientBase::submitTransaction(TransactionSkeleton const& _t, Secret const& _secret)
+h256 ClientBase::submitTransaction(TransactionSkeleton const& _t, Secret const& _secret, u256 o_nonce)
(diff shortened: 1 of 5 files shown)


libethereum/ClientBase.cpp
```diff
@@ -45,14 +45,14 @@ State ClientBase::asOf(BlockNumber _h) const
 	return asOf(bc().numberHash(_h));
 }
 
-h256 ClientBase::submitTransaction(TransactionSkeleton const& _t, Secret const& _secret)
+h256 ClientBase::submitTransaction(TransactionSkeleton const& _t, Secret const& _secret, u256 o_nonce)
```
chriseth: If this is supposed to be an output parameter, it should be a reference.

CJentzsch:
Solved returning a pair: https://github.com/ethereum/cpp-ethereum/commit/9d5b632c5821da43fc525c2c74501e4edacce7d2

PR #2421, opened 2015-07-08
FixedHash::operator++() utilized

gluk256:


gluk256:
apparently, the errors are not mine.

PR #2422, opened 2015-07-08
Mix - bug fix. issue while getting the created contract address.

yann300:


mix/ClientModel.cpp
```diff
@@ -390,7 +390,8 @@ void ClientModel::executeSequence(vector<TransactionSettings> const& _sequence)
 				{
 					bytes param = encoder.encodedData();
 					contractCode.insert(contractCode.end(), param.begin(), param.end());
-					Address newAddress = deployContract(contractCode, transaction);
+					deployContract(contractCode, transaction);
+					Address newAddress = m_client->lastCreatedContractAddr();
```
arkpar: Better call it from inside deployContract and return Address as before.

yann300:
okay done

PR #2428, opened 2015-07-08
Start pv60 sync with 10sec delay

arkpar:


libethereum/EthereumHost.h
```diff
@@ -138,6 +138,7 @@ class EthereumHost: public p2p::HostCapability<EthereumPeer>, Worker
 	mutable Mutex x_transactions;
 	DownloadMan m_man;
 	std::unique_ptr<BlockChainSync> m_sync;
+	std::atomic<time_t> m_syncStart = { 0 };
```
gavofyork: prefer usage of `chrono`...

libethereum/EthereumHost.cpp
```diff
@@ -115,6 +116,21 @@ void EthereumHost::doWork()
 
 	foreachPeer([](std::shared_ptr<EthereumPeer> _p) { _p->tick(); return true; });
 
+	if (m_syncStart)
+	{
+		Guard l(x_sync);
+		if (!m_sync)
+		{
+			time_t now = std::chrono::system_clock::to_time_t(chrono::system_clock::now());
+			if ((now - m_syncStart) > 10)
```
gavofyork: reason for `(` `)`s?

libethereum/EthereumHost.cpp
```diff
@@ -115,6 +116,21 @@ void EthereumHost::doWork()
 
 	foreachPeer([](std::shared_ptr<EthereumPeer> _p) { _p->tick(); return true; });
 
+	if (m_syncStart)
+	{
+		Guard l(x_sync);
```
gavofyork: consider `DEV_GUARDED(x_sync)` to avoid additional `{` `}`s

libethereum/EthereumHost.h
arkpar: would require unnecessary mutex locking

PR #2430, opened 2015-07-08
ethashcl doesnt depend on ethcore

LefterisJP:
At least not anymore. Moved global constants into ethash_cl_miner class
instead of being in Ethash class.

Also `#ifdefed` some openCL related stuff to only compile when `ETHASHCL` is active.

### DIFF
--- a/libethash-cl/ethash_cl_miner.cpp
+++ b/libethash-cl/ethash_cl_miner.cpp
@@ -50,9 +50,9 @@
 
 using namespace std;
 
-const unsigned ethash_cl_miner::defaultLocalWorkSize = 64;
-const unsigned ethash_cl_miner::defaultGlobalWorkSizeMultiplier = 4096; // * CL_DEFAULT_LOCAL_WORK_SIZE
-const unsigned ethash_cl_miner::defaultMSPerBatch = 0;
+const unsigned ethash_cl_miner::c_defaultLocalWorkSize = 64;
+const unsigned ethash_cl_miner::c_defaultGlobalWorkSizeMultiplier = 4096; // * CL_DEFAULT_LOCAL_WORK_SIZE
+const unsigned ethash_cl_miner::c_defaultMSPerBatch = 0;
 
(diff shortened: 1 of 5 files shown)


libethash-cl/ethash_cl_miner.h
```diff
@@ -65,6 +65,14 @@ class ethash_cl_miner
 	void hash_chunk(uint8_t* _ret, uint8_t const* _header, uint64_t _nonce, unsigned _count);
 	void search_chunk(uint8_t const*_header, uint64_t _target, search_hook& _hook);
 
+	/* -- default values -- */
+	/// Default value of the local work size. Also known as workgroup size.
+	static const unsigned defaultLocalWorkSize;
```
gavofyork: coding standards: should either be `c_`... if not public API or CamelCase if yes.

libethash-cl/ethash_cl_miner.cpp
```diff
@@ -50,7 +49,10 @@
 #undef max
 
 using namespace std;
-using namespace dev::eth;
+
+const unsigned ethash_cl_miner::c_defaultLocalWorkSize = 64;
```
chfast: I guess it should be `unsigned const`.

libethash-cl/ethash_cl_miner.h
```diff
@@ -65,6 +65,14 @@ class ethash_cl_miner
 	void hash_chunk(uint8_t* _ret, uint8_t const* _header, uint64_t _nonce, unsigned _count);
 	void search_chunk(uint8_t const*_header, uint64_t _target, search_hook& _hook);
 
+	/* -- default values -- */
+	/// Default value of the local work size. Also known as workgroup size.
+	static const unsigned c_defaultLocalWorkSize;
```
chfast: Don't want to put the values here?

LefterisJP: Would produce undefined references due to being a static member of the class

PR #2432, opened 2015-07-08
ethminer no longer depends on boost

LefterisJP:
Introducing an extra boolean argument to remove the boost::optional
dependency. Too bad we have to wait for C++17 to get an std::optional


LefterisJP:
Rebased on develop (to get the ethcore dependency change)

Issue #2349, opened 2015-07-02
network: bind address already in use uncaught exception (crash)


gavofyork:
```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >'
  what():  bind: Address already in use
```

LefterisJP:
What's the issue for? I also get that when for example a geth runs on the same machine hence occupying the port. An error message would be much better than an uncaught exception ofcourse.

subtly:
#1398

gavofyork:
fixed.

Issue #2333, opened 2015-06-30
Worker is null


chfast:
It looks like the object that implements the Worker interface is null and when boost tries to execute virtual method doWork() the program crashes.

```
ASAN:SIGSEGV
=================================================================
==17394==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000090 (pc 0x7f572255ca0f bp 0x7f56fdf43750 sp 0x7f56fdf436e0 T45)
    #0 0x7f572255ca0e in boost::asio::detail::epoll_reactor::start_op(int, int, boost::asio::detail::epoll_reactor::descriptor_state*&, boost::asio::detail::reactor_op*, bool, bool) /usr/local/include/boost/asio/detail/impl/epoll_reactor.ipp:219:7
    #1 0x7f572255c8cd in boost::asio::detail::reactive_socket_service_base::start_op(boost::asio::detail::reactive_socket_service_base::base_implementation_type&, int, boost::asio::detail::reactor_op*, bool, bool, bool) /usr/local/include/boost/asio/detail/impl/reactive_socket_service_base.ipp:213:7
    #2 0x7f57225af9a4 in void boost::asio::detail::reactive_socket_service_base::async_receive<boost::asio::mutable_buffers_1, boost::asio::detail::read_op<boost::asio::basic_stream_socket<boost::asio::ip::tcp, boost::asio::stream_socket_service<boost::asio::ip::tcp> >, boost::asio::mutable_buffers_1, boost::asio::detail::transfer_all_t, dev::p2p::Session::doRead()::$_1::operator()(boost::system::error_code, unsigned long) const::{lambda(boost::system::error_code, unsigned long)#1}> >(boost::asio::detail::reactive_socket_service_base::base_implementation_type&, boost::asio::mutable_buffers_1 const&, int, boost::asio::detail::read_op<boost::asio::basic_stream_socket<boost::asio::ip::tcp, boost::asio::stream_socket_service<boost::asio::ip::tcp> >, boost::asio::mutable_buffers_1, boost::asio::detail::transfer_all_t, dev::p2p::Session::doRead()::$_1::operator()(boost::system::error_code, unsigned long) const::{lambda(boost::system::error_code, unsigned long)#1}>&) /usr/local/include/boost/asio/detail/reactive_socket_service_base.hpp:287:5
[...]

Issue #1398, opened 2015-03-24
Core Dump Cannot assign requested address


aperseghin:
output from running "eth" command 

Ethereum (++) 0.9.1
  Code by Gav Wood et al, (c) 2013, 2014, 2015.
  Based on a design by Vitalik Buterin.

*N* [ 10:12:24 | main ] IP Address: 192.168.1.3  = [LOCAL]
*N* [ 10:12:24 | main ] Id: e5294826d69a56a77c77566691b11f6bb27ada72bc3d27d7fc8a780314360571bfcff71e18e1d31b2e0d7b92e0f299bb479f587ae40f9163295c17cbd43b419c
*** [ 10:12:24 | main ] Opened blockchain DB. Latest: fd4af92a79c7fc2fd8bf0d342f2e832e1d4f485c85b9152d2039e03bc604fdca
*** [ 10:12:25 | main ] Opened state DB.
Transaction Signer: d134e55b80c02fd549534262f5764e3a1918e82e
Mining Benefactor: d134e55b80c02fd549534262f5764e3a1918e82e
*** [ 10:12:25 | eth ] New block on chain: Restarting mining operation.
*** [ 10:12:25 | eth ] noteChanged( { 0000000000000000000000000000000000000000000000000000000000000000 } )
*** [ 10:12:29 | p2p ] UPnP device: http://192.168.1.1:2555/upnp/cbba0323-d2bb-3569-b2d2-2df8571c54c1/desc.xml [st: urn:schemas-upnp-org:device:InternetGatewayDevice:1 ]
*N* [ 10:12:29 | p2p ] Punched through NAT and mapped local port 30303 onto external port 30303 .
*N* [ 10:12:29 | p2p ] External addr: 71.121.194.171
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >'
  **what():  bind: Cannot assign requested address**
**Aborted (core dumped)**

aperseghin:
Results from compiled version of cpp-ethereum are the same as from the repo 

root@SERVER:~/ethereum/cpp-ethereum/build/eth# ./eth -a xxxxxxxxxxxxxxxxxxxxxx -m on 
!!! [ 11:10:36 | main ] /root/ethereum/cpp-ethereum/libdevcore/CommonData.cpp(78): Throw in function int dev::fromHex(char)
[...]

PR #2438, opened 2015-07-10
Fixes for #2337, #2349, #2288, #2333, #1398.

subtly:


libp2p/Host.cpp
```diff
@@ -170,15 +184,13 @@ void Host::doneWorking()
 	// disconnect pending handshake, before peers, as a handshake may create a peer
 	for (unsigned n = 0;; n = 0)
 	{
-		{
-			Guard l(x_connecting);
+		DEV_GUARDED(x_connecting)
 			for (auto i: m_connecting)
```
chfast: You should considered iterating over the shared/weak pointers using the reference, not the copy. I.e. `for (auto& i: m_connecting)`. For performance reasons you may not want to temporary bump the shared/weak counter.

There is also "alwyas correct" variant of the for loop: `for (auto&& x: collection)`. That one will be preferred by later C++ standards by adding the syntax sugar in the form of `for(x: collection)`.

gavofyork:
compile fail on osx?

libdevcore/Worker.cpp
```diff
@@ -50,11 +50,16 @@ void Worker::startWorking()
...
-				workLoop();
-//				cnote << "Finishing up worker thread...";
-				doneWorking();
+				try
+				{
+					startedWorking();
+					workLoop();
+					doneWorking();
+				}
+				catch (std::exception const& _e)
```
arkpar: what about boost exceptions? Note that boost::current_exception_diagnostic_information() works for both

libp2p/Host.cpp
```diff
@@ -120,6 +120,15 @@ Host::~Host()
 void Host::start()
 {
 	startWorking();
+	while (isWorking() && !haveNetwork())
+		this_thread::sleep_for(chrono::milliseconds(10));
```
gluk256: i suggest a timeout here, instead of eternal cycle.

gavofyork:
tests failing.

libdevcore/Worker.cpp
subtly: Happy to change this to anything more useful.

libp2p/Host.cpp
subtly: isWorking will be set to false if/when startWorking fails due to an exception; if the exception doesn't occur then haveNetwork will return true.

libdevcore/Worker.cpp
arkpar: I mean something like this:
```
catch(...)
{
    clog(WarnChannel) << boost::current_exception_diagnostic_information();
}
```

PR #2443, opened 2015-07-10
Add minelog, a verbosity 1 logger for ethminer

LefterisJP:


ethminer/MinerAux.h
```diff
@@ -81,6 +81,12 @@ inline std::string credits()
 }
 
 class BadArgument: public Exception {};
+struct MiningChannel: public LogChannel
+{
+	static const char* name() { return EthGreen "miner"; }
+	static const int verbosity = 1;
```
chfast: Set bigger verbosity. 1 is the same as cnote.

PR #2455, opened 2015-07-13
Minor cleanup of State and Client code

arkpar:


gavofyork:
looks ok, but this is going to collide substantially with my refactoring.

libethereum/BlockChain.cpp
```diff
@@ -696,8 +696,10 @@ ImportRoute BlockChain::import(VerifiedBlockRef const& _block, OverlayDB const&
 		exit(-1);
 	}
 
-	try {
-		State canary(_db, *this, _block.info.hash(), ImportRequirements::DontHave);
+	try
+	{
+		State canary(_db, BaseState::Empty);
```
gavofyork: `BaseState::PreExisting` is marginnally more efficient if you're about to populate it anyway.

PR #2457, opened 2015-07-13
Stricter checks on RLP.

gavofyork:


libdevcore/RLP.cpp
```diff
@@ -203,7 +213,7 @@ size_t RLP::items() const
 {
 	if (isList())
 	{
-		bytesConstRef d = payload().cropped(0, length());
+		bytesConstRef d = payload();
```
winsvega: so, what about random bytes at the end?

PR #2458, opened 2015-07-14
Add JSCONSOLE to the default bundle

LefterisJP:
The default values is what the buildbot will be building and packaging. For Release now we need to also have JSCONSOLE and ETHASHCL active.

This PR will only be for JSCONSOLE. Next one will be for ETHASHCL.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ option(TESTS "Build the tests." ON)
 option(NOBOOST "No use of boost macros in test functions" OFF)
 option(EVMJIT "Build just-in-time compiler for EVM code (requires LLVM)" OFF)
 option(ETHASHCL "Build in support for GPU mining via OpenCL" OFF)
-option(JSCONSOLE "Build in javascript console" OFF)
+option(JSCONSOLE "Build in javascript console" ON)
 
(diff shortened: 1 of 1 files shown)


LefterisJP:
Trying to configure the buildbot to build succesfully with the V8 dependencly on all platforms

caktux:
ethereum-dockers repo updated with the dependencies (for both JSCONSOLE and ETHASHCL), homebrew-ethereum will need its options switched from `--with-v8-console` and `--with-gpu-mining` to `--without-...` once this is merged, just putting it here for reference and getting notified.

LefterisJP:
Acknowledged. Let me trigger a rebuild to check.

LefterisJP:
From the triggered rebuild we can see that jsconsole dependencies are installed properly on all buildbots except the 2 that fail with evmjit. This is a different error unrelated to this PR.

subtly:
Rebase/merge?

subtly:
@caktux The 'gpu' flag can, imo, safely be removed on macos w/upcoming changes. OpenCL has come w/OSX since before Mavericks (they've bundled it for years now for iphoto/imovie/etc).

LefterisJP:
why would this need a rebase? To trigger rebuild?

PR #2462, opened 2015-07-14
Save keys after deletion

### LINKED ISSUE #2256: Kill account in Alethzero does not remove it from the owned accounts list.
Once you kill an account from the tools menu, close Alethzero and re-open. It still displays the old list of accounts along with the account that was killed. 

Not entirely sure if this is the only fix for this bug but looks like although the address is deleted from the m_keyManager, it is not explicitly deleted from the ui->ourAccounts QListWidget. 
[ui->ourAccounts->takeItem(index) is missing]

arkpar:
Fixes #2256


chriseth:
Add a test? :-)

arkpar:
Done

PR #2463, opened 2015-07-14
Allow 0x prefix for address entry in AZ

arkpar:


alethzero/MainWin.cpp
```diff
@@ -638,18 +638,22 @@ pair<Address, bytes> Main::fromString(std::string const& _n) const
 	if (_n == "(Create Contract)")
 		return make_pair(Address(), bytes());
 
+	std::string n = _n;
+	if (n.find_first_of("0x") == 0)
```
chriseth: This is triggered also for a string of the form "00123".

PR #2468, opened 2015-07-14
AlethZero: Change the URL when the page changes, unless it is the URL of a Dapp.

### LINKED ISSUE #1215: Client-side routing works but doesn't update url
Not really an important issue but it would be nice if the url updated to reflect the current client-side route.

chriseth:
Fixes #1215

### DIFF
--- a/alethzero/MainWin.cpp
+++ b/alethzero/MainWin.cpp
@@ -232,6 +232,10 @@ Main::Main(QWidget *parent) :
 	{
 		ui->tabWidget->setTabText(0, ui->webView->title());
 	});
+	connect(ui->webView, &QWebEngineView::urlChanged, [=](QUrl const& _url)
+	{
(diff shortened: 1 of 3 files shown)


alethzero/MainWin.cpp
```diff
@@ -232,6 +232,10 @@ Main::Main(QWidget *parent) :
 	{
 		ui->tabWidget->setTabText(0, ui->webView->title());
 	});
+	connect(ui->webView, &QWebEngineView::urlChanged, [=](QUrl const& _url)
+	{
+		ui->urlEdit->setText(_url.toString());
```
arkpar: For dapp packages urlEdit contains ethereum url (namereg path). The webview loads pages from http://localhost in such case and that url should stay hidden.

chriseth:
It would be great if we could display something else instead.

chriseth:
Does not yet work as expected.

PR #2476, opened 2015-07-14
clear reverted block hashes from cache during imporing

debris:
After reverting head of chain I noticed that `eth_getBlockByNumber` doesn't work properly for some blocks (I made them **bold**)

......1 - 2 - 3 ... 26 - 27a - 28a
...........................\ - **27b** - **28b** - 29 - 30 - 31 - 32

When I called `eth_getBlockByNumber(27)` or `eth_getBlockByNumber(28)` I received block 27a or 28a instead of 27b and 28b. Old blockHashes were cached in `m_blockHashes`.

### DIFF
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -597,9 +597,26 @@ ImportRoute BlockChain::import(VerifiedBlockRef const& _block, OverlayDB const& 
 
 		// Most of the time these two will be equal - only when we're doing a chain revert will they not be
 		if (common != last)
+		{
 			// If we are reverting previous blocks, we need to clear their blooms (in particular, to
 			// rebuild any higher level blooms that they contributed to).
 			clearBlockBlooms(number(common) + 1, number(last) + 1);
+
+			// clear reverted blockhashes
+			DEV_WRITE_GUARDED(x_blockHashes)
+				for (auto i = route.begin(); i != route.end() && *i != common; ++i)
+				{
+					BlockInfo tbi = BlockInfo(block(*i));
+					auto h = h256(tbi.number);
+					for (auto j = m_blockHashes.begin(); j != m_blockHashes.end();)
+					{
+						if (j->first == h)
+							j = m_blockHashes.erase(j);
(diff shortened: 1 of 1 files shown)


libethereum/BlockChain.cpp
```diff
@@ -597,10 +597,27 @@ ImportRoute BlockChain::import(VerifiedBlockRef const& _block, OverlayDB const&
...
 			clearBlockBlooms(number(common) + 1, number(last) + 1);
 
+			// clear reverted blockhashes
+			DEV_WRITE_GUARDED(x_blockHashes)
+				for (auto i = route.begin(); i != route.end() && *i != common; ++i)
+				{
+					BlockInfo tbi = BlockInfo(block(*i));
+					auto h = h256(tbi.number);
+					for (auto j = m_blockHashes.begin(); j != m_blockHashes.end();)
+					{
```
gavofyork: unneeded `{`s

gavofyork:
this algorithm appears to be order(N^2)!

libethereum/BlockChain.cpp
```diff
@@ -597,10 +597,27 @@ ImportRoute BlockChain::import(VerifiedBlockRef const& _block, OverlayDB const&
...
 			// rebuild any higher level blooms that they contributed to).
 			clearBlockBlooms(number(common) + 1, number(last) + 1);
 
+			// clear reverted blockhashes
+			DEV_WRITE_GUARDED(x_blockHashes)
+				for (auto i = route.begin(); i != route.end() && *i != common; ++i)
+				{
+					BlockInfo tbi = BlockInfo(block(*i));
+					auto h = h256(tbi.number);
+					for (auto j = m_blockHashes.begin(); j != m_blockHashes.end();)
```
gavofyork: `m_blockHashes` is a map. to erase a single element from a map whose key you know, simply use `erase`. there is also a method in `BlockInfo` called `number`. this entire block can be reduced to:

```
m_blockHashes.erase(h256(number(*i)));
```

PR #2477, opened 2015-07-14
getstuff downloads llvm 3.7svn on windows

debris:


LefterisJP:
This is for the windowsjit build, yes? Has it been tested?

debris:
I think that @pawel has tested it. He requested me to upload it to the buildbot.

LefterisJP:
yea we need this dependency in Windows. If it's tested and works then it's awesome. Will also test in Windows when I get a Windows machine.

chfast:
I can test it locally if you want. But the only part that can go wrong is downloading and unpacking. The LLVM prebuilt libraries are fine.

LefterisJP:
No it should be fine. This is what I wanted to hear.

PR #2485, opened 2015-07-15
Windows install changes

chfast:


eth/CMakeLists.txt
```diff
@@ -41,9 +41,9 @@ if (JSCONSOLE)
 	target_link_libraries(${EXECUTABLE} jsconsole)
 endif()
 
-if (DEFINED WIN32 AND NOT DEFINED CMAKE_COMPILER_IS_MINGW)
-	eth_copy_dlls("${EXECUTABLE}" MHD_DLLS)
-endif()
+#if (DEFINED WIN32 AND NOT DEFINED CMAKE_COMPILER_IS_MINGW)
```
subtly: All for cleaning up cmake files but this is unrelated.

cmake/scripts/copydlls.cmake
```diff
@@ -8,10 +8,10 @@
 
 # this script is created cause we do not know configuration in multiconfiguration generators at cmake configure phase ;)
 
-if ("${CONF}" STREQUAL "Release")
-    set(DLL ${DLL_RELEASE})
-else () # Debug
+if ("${CONF}" STREQUAL "Debug")
     set(DLL ${DLL_DEBUG})
+else ()
+    set(DLL ${DLL_RELEASE})
```
subtly: Does this also fix RelWithDebInfo builds?

cmake/EthExecutableHelper.cmake
```diff
@@ -139,6 +139,12 @@ macro(eth_install_executable EXECUTABLE)
 			COMPONENT ${EXECUTABLE}
 		)
 
+		install(DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/RelWithDebInfo/"
+			DESTINATION bin
```
subtly: Why bin and not . like others?

cmake/scripts/copydlls.cmake
chfast: Yes. "debug" dependencies should be used only for Debug builds. "release" for all others.

cmake/EthExecutableHelper.cmake
chfast: The other uses the build type name, like "Release" and "Debug". See the missing "/" in the end. I find a bit awkward. "bin" fixes some DLL install problems (e.g. evmjit.dll is installed to bin).

We should've design the  installed files layout up front. But now I just want something that works.

chfast:
Thanks for checking that @subtly. I think it needs some cleaups before merging, but I want it right first, clean after that.

CMakeLists.txt
```diff
@@ -506,26 +506,26 @@ endif ()
...
+	set(CPACK_PACKAGE_NAME "Ethereum (++)")
+	set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "The Ethereum (++) Toolset")
 	set(CPACK_PACKAGE_VENDOR "ethereum.org")
 	set(CPACK_PACKAGE_DESCRIPTION_FILE "${CMAKE_CURRENT_SOURCE_DIR}/README.md")
 	set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE")
-	set(CPACK_PACKAGE_VERSION "0.9")
+	set(CPACK_PACKAGE_VERSION "0.9.29")
 	set(CPACK_GENERATOR "NSIS")
 	# seems to be not working
 	# set(CPACK_PACKAGE_ICON "${CMAKE_CURRENT_SOURCE_DIR}/alethzero/alethzero.bmp")
```
LefterisJP: These commented out stuff, should we figure them out now, or in a later PR? It would need a few cycles of testing between Cmake's CPACK and NSIS installer to see why it's not working.

cmake/FindOpenCL.cmake
```diff
@@ -124,6 +124,24 @@ endif()
...
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+
+	find_library(
+		OpenCL_LIBRARY_DEBUG
+		NAMES OpenCL_d
+	)
+
+	set(OpenCL_LIBRARIES optimized ${OpenCL_LIBRARY} debug ${OpenCL_LIBRARY_DEBUG})
+
+    # prepare dlls
```
LefterisJP: tab instead of whitespace here

CMakeLists.txt
chfast: With that enabled the installer includes components only. It needs more work than has been done as it would miss some required stuff like evmjit.dll, OpenCl.dll etc.

exp/CMakeLists.txt
```diff
@@ -30,6 +27,6 @@ target_link_libraries(${EXECUTABLE} p2p)
 if (ETHASHCL)
 	target_link_libraries(${EXECUTABLE} ethash-cl)
 	target_link_libraries(${EXECUTABLE} ethash)
-	target_link_libraries(${EXECUTABLE} ${OpenCL_LIBRARIES})
+	target_link_libraries(${EXECUTABLE} ${OpenCL_TARGET})
```
LefterisJP: What is `OpenCL_TARGET`? Can't find it in our files by grepping. Is it a cmake thing?

chfast: It used to be set to OpenCL or OpenCL_ICD.

libethash-cl/ethash_cl_miner.cpp
```diff
@@ -78,23 +78,37 @@ ethash_cl_miner::~ethash_cl_miner()
...
-	if (platforms.empty())
+	try
 	{
-		ETHCL_LOG("No OpenCL platforms found.");
-		return string();
+		cl::Platform::get(&platforms);
 	}
+	catch(cl::Error const& err)
+	{
+		#if defined(CL_PLATFORM_NOT_FOUND_KHR)
```
LefterisJP: Maybe don't indent the preprocessor commands, since they are more readable this way? I am not sure what our style guide says about that but I personally find it more readable if they are not indented. Something like:

```
#if defined(CL_PLATFORM_NOT_FOUND_KHR)
        if (err.err() == CL_PLATFORM_NOT_FOUND_KHR)
            ETHCL_LOG("No OpenCL platforms found");
        else
#endif
            throw err;
```

LefterisJP:
looks good to me. We should also figure out a working way to specify which files we would like to exclude in the final installation. Too bad that `PATTERN` can't be mixed with `COMPONENT` and `CONFIGURATION`.

But this can happen in a later PR. Let's first have a working version that does what we need ASAP and we can build on that incrementally.

chfast:
The true problem about that is we are using `install(DIRECTORY)` what is quite stupid.

PR #2487, opened 2015-07-15
Templatify method of block sealing

gavofyork:


CMakeLists.txt
```diff
@@ -434,31 +434,31 @@ add_subdirectory(libethcore)
...
 	endif ()
 endif ()
 
 if (TOOLS)
 
-	add_subdirectory(rlp)
-	add_subdirectory(abi)
-	add_subdirectory(ethvm)
-	add_subdirectory(eth)
+#	add_subdirectory(rlp)
```
chriseth: Commented out on purpose (also elsewhere)?

libethcore/BlockInfo.cpp
```diff
@@ -145,64 +130,11 @@ void BlockInfo::populateFromHeader(RLP const& _header, Strictness _s, h256 const
 		throw;
 	}
 
-	if (number > ~(unsigned)0)
+	if (m_number > ~(unsigned)0)
```
chriseth: Shouldn't this use a platform-independent type?

libethcore/BlockInfo.h
```diff
@@ -67,90 +77,200 @@ enum Strictness
...
+		BlockInfo::populateFromHeader(_header, _s);
+		BlockInfoSub::populateFromHeader(_header, _s);
+	}
+
+	void clear() { BlockInfo::clear(); BlockInfoSub::clear(); BlockInfoSub::noteDirty(); }
+	void noteDirty() const { BlockInfo::noteDirty(); BlockInfoSub::noteDirty(); }
+
+	h256 headerHash(IncludeProof _i = WithProof) const
+	{
+		RLPStream s;
```
chriseth: return `sha3(rlp(_i))`?

libethcore/Common.cpp
```diff
@@ -52,7 +56,7 @@ Network const c_network = Network::Frontier;
 Network const c_network = Network::Olympic;
 #endif
 
-const unsigned c_databaseVersion = c_databaseBaseVersion + (c_databaseVersionModifier << 8) + (ProofOfWork::revision() << 9);
+const unsigned c_databaseVersion = c_databaseBaseVersion + (c_databaseVersionModifier << 8) + (23 << 9);
```
chriseth: The literal constant looks dangerous.

libethcore/Ethash.cpp
```diff
@@ -54,58 +59,87 @@ namespace dev
...
 {
 
-const Ethash::WorkPackage Ethash::NullWorkPackage = Ethash::WorkPackage();
-
-std::string Ethash::name()
+h256 const& Ethash::BlockHeaderRaw::seedHash() const
 {
-	return "Ethash";
+	if (!m_seedHash)
+		m_seedHash = EthashAux::seedHash((unsigned)m_number);
```
chriseth: Also here: Platform-independent type.

libethcore/Common.cpp
gavofyork: indeed - temporary workaround to avoid the database being flushed; will remove on the next database update (i.e. frontier release).

libethcore/BlockInfo.cpp
gavofyork: `m_number` is `u256` so perhaps `(u256)` would be more reasonable here. removing the cast may also be an option since i guess it'll be upcast to `u256` anyway (not sure if supported on all compilers though).

PR #2489, opened 2015-07-16
Performance optimizations for msvc build

arkpar:


libdevcrypto/CMakeLists.txt
```diff
@@ -24,10 +24,8 @@ target_link_libraries(${EXECUTABLE} ${DB_LIBRARIES})
 target_link_libraries(${EXECUTABLE} ${CRYPTOPP_LIBRARIES})
 target_link_libraries(${EXECUTABLE} scrypt)
 target_link_libraries(${EXECUTABLE} devcore)
-if (NOT WIN32)
-	add_definitions(-DETH_HAVE_SECP256K1)
-	target_link_libraries(${EXECUTABLE} secp256k1)
-endif ()
+add_definitions(-DETH_HAVE_SECP256K1)
```
subtly: It's possible this may not work when we upgrade to bitcoin-core; openssl and/or different compile flags may be required.

libethereum/BlockChain.h
```diff
@@ -154,7 +155,7 @@ class BlockChain
 	UncleHashes uncleHashes() const { return uncleHashes(currentHash()); }
 	
 	/// Get the hash for a given block's number.
-	h256 numberHash(unsigned _i) const { if (!_i) return genesisHash(); return queryExtras<BlockHash, ExtraBlockHash>(h256(_i), m_blockHashes, x_blockHashes, NullBlockHash).value; }
+	h256 numberHash(unsigned _i) const { if (!_i) return genesisHash(); return queryExtras<BlockHash, uint64_t, ExtraBlockHash>(_i, m_blockHashes, x_blockHashes, NullBlockHash).value; }
```
subtly: Can _i be greater than 1<<63?

arkpar: No, since its type is unsigned here

libdevcrypto/CMakeLists.txt
arkpar: We'll add flags in such case

libethereum/BlockChain.cpp
```diff
@@ -84,17 +84,23 @@ std::ostream& dev::eth::operator<<(std::ostream& _out, BlockChain const& _bc)
...
 	static boost::thread_specific_ptr<FixedHash<33>> t_h;
 	if (!t_h.get())
 		t_h.reset(new FixedHash<33>);
 	*t_h = FixedHash<33>(_h);
 	(*t_h)[32] = (uint8_t)_sub;
 	return (ldb::Slice)t_h->ref();//(char const*)t_h.get(), 32);
-#endif
+}
+
+ldb::Slice dev::eth::toSlice(uint64_t _n, unsigned _sub)
```
subtly: This is could be a dangerous change and this method could be very useful in other places. What about casting to uint64_t inside toSlice instead of adding an additional method? Perhaps an enum argument which causes downcasts (w/default = no cast, no truncate).

subtly:
This appears to be querying by block number but hitting m_blockHashes which is BlockHashHash = std::unordered_map<h256, BlockHash>. Is m_blockHashes storing block hashes by number?

If this is by number and there won't be > 2**63 block numbers, maybe the unordered_map can be reduced as well?

libethereum/BlockChain.cpp
arkpar: Additional method is required for generic queryExtras method to work for both types. Casing to uint64_t in general case is wrong since both block hashes and numbers are stored in the database.

arkpar:
m_blockHashes is a cache map of block number to block hash. Not all of the numbers are cached though, so can't use vector here.

libethereum/BlockChain.cpp
```diff
@@ -84,17 +84,23 @@ std::ostream& dev::eth::operator<<(std::ostream& _out, BlockChain const& _bc)
 
 ldb::Slice dev::eth::toSlice(h256 const& _h, unsigned _sub)
 {
-#if ALL_COMPILERS_ARE_CPP11_COMPLIANT
```
gavofyork: any reason to remove this? the potential code allowed through c++11 once we target the newer MSVC is substantially cleaner.

gavofyork:
probably best to rebase on whizz.

libethereum/BlockChain.cpp
arkpar: Its the only place in all of the codebase this macro is currently used. And the code appears to be outdated. I'll restore and fix it.

arkpar:
rebased

PR #2490, opened 2015-07-16
Copy evmjit.dll to output on windows

arkpar:


CMakeLists.txt
```diff
@@ -409,10 +413,8 @@ if (JSCONSOLE)
 	add_subdirectory(ethconsole)
 endif ()
 
-if (NOT WIN32)
```
chfast: That fails Windows build.

eth/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ if (JSCONSOLE)
 endif()
 
 if (DEFINED WIN32 AND NOT DEFINED CMAKE_COMPILER_IS_MINGW)
-	eth_copy_dlls("${EXECUTABLE}" MHD_DLLS)
+	eth_copy_dlls("${EXECUTABLE}" MHD_DLLS EVMJIT_DLLS)
```
chfast: It does not work!!

subtly: MSVC isn't CMAKE_COMPILER_IS_MINGW :)

subtly:
Does eth need to be directly linked to evmjit? Or is evmjit header only?  If it's the former... eth doesn't have any linkage to evmjit. What all needs to link evmjit?

arkpar:
evmjit builds as dll on widows (unlike other libraries which build as static). Every executable that links it directly or indirectly needs the dll to run. This build action copies the dll to the exe output dir

eth/CMakeLists.txt
arkpar: @subtly, NOT DEFINED CMAKE_COMPILER_IS_MINGW there, so msvc fits.

chfast: In other parts "compiler match msvc" is used. And it's just simpler.

arkpar: Someone has put it there for a reason and I don't have a mingw setup to test if it works. So I'd prefer not to touch it.

gavofyork:
base? preferably on whizz branch.

debris:
looksgood

PR #2503, opened 2015-07-17
Automatic guards.

chriseth:
This is an attempt at prohibiting non-guarded access to a shared object.
The pros are, in my view:
1. you do not need two members (`x_...` and `m_...`) for the simple use-case
2. it is harder to access the object in a non-locked state
The cons are:
1. the dereference and pointer semantics are a bit strange at first
2. it is less visible what actually happens and you need to know the lifetime of temporaries
2. a unique_ptr is allocated for each lock, though that might be avoidable

### DIFF
--- a/libdevcore/Guards.h
+++ b/libdevcore/Guards.h
@@ -22,6 +22,7 @@
 #pragma once
 
 #include <mutex>
+#include <memory>
 #include <atomic>
 #include <boost/thread.hpp>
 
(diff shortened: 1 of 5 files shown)


gavofyork:
the `**` notation is indeed a little esoteric. i wonder if it may be avoidable through a set of templated operator forwards?

chriseth:
Avoided this specific use of `**`.
In general, though, I think it cannot be avoided because we need the guard as a temporary in the calling context.

mix/CodeModel.cpp
```diff
@@ -213,11 +213,11 @@ void CodeModel::reset(QVariantMap const& _documents)
 	///@todo: cancel bg job
 	Guard l(x_contractMap);
 	releaseContracts();
-	Guard pl(x_pendingContracts);
-	m_pendingContracts.clear();
+	auto contracts = *m_pendingContracts;
```
chfast: This syntax hardly matches the semantics. And looks quite awkward. Maybe just `m_pendingContracts.access()` of `m_pendingContracts.getGuard()`.

libdevcore/Guards.h
```diff
@@ -122,4 +123,50 @@ using SpinGuard = std::lock_guard<SpinLock>;
...
+public:
+	LockedObject() {}
+	LockedObject(Class&& _object): m_object(std::move(_object)) {}
+	/// Use *lockedObject to access a guard which releases the lock on destruction and allows
+	/// to access the object via ->.
+	Container operator*() { return guard<Guard>(); }
+	/// Use -> to access the object for a single, locked access.
+	Container operator->() { return guard<Guard>(); }
+
+	GuardContainer<Class, ReadGuard, MutexType> read() { return guard<ReadGuard>(); }
```
subtly: Can ReadGuard and WriteGuard safely use std::mutex (default MutexType)?

subtly:
I think wrapper classes for Container types might make sense however I don't think this makes sense as a general template. Compared to the current method it removes important information from the codebase such as scope and guard type. I don't see a problem w/current syntax ("If it's not broke...") and this won't make anything faster. It's good to put mutexes close to what they protect and wrapping an entire class is counter this. Syntax won't remove the complexity of using mutexes.

mix/CodeModel.cpp
chriseth: You can already use `.guard<GuardType>()`, I can add a non-templated function there.

chriseth:
@subtly, it is not about syntax or making things faster. It is about forcing developers to stretch if they want to access the object without locking its mutex.

subtly:
@chriseth I don't see the benefit. There have been cases w/networking where mutexes were an issue however most cases were, a) there wasn't a mutex, or b) there were two mutexes.
Compared to existing macro:
* additional call overhead
* code less expressive
* access to object is restricted
* access to mutex is restricted
* scoping not supported
* docs will become LockedObject
* subjective: more work required by developer
* subjective: false security

Just not sure what problem this fixes. It's a stretch to decide to use a mutex to start. Why is forcing a developer to stretch good here?

We could always use t_something instead of m_something for properties which are accessed by multiple threads.

gavofyork:
There have certainly been bugs where mutexes for fields which were not locked during access. Christian rightly points out that this problem would be reduced by making such behaviour more difficult. The question is whether this gain is enough to offset the reduction of readability.

Aside from the questionable syntax, the main issue I see is that it's somewhat limited to cases where there's a basic mutex locking a single field. Many, if not most, instances of guards in the codebase are with read/write locks and/or multiple fields per guard. One workaround for the latter may be to have a struct aggregating all fields protected by the lock and using that for the templated type. Not sure about the former, though.

PR #2505, opened 2015-07-17
Blockchain:rescue DB

winsvega:
unsigned overflow fix

### DIFF
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -856,7 +856,8 @@ void BlockChain::rescue(OverlayDB& _db)
 				break;
 		}
 		catch (...) {}
-		if (l == 0) break;
+		if (l == 0)
+			break;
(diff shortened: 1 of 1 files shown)


libethereum/BlockChain.cpp
```diff
@@ -856,6 +856,8 @@ void BlockChain::rescue(OverlayDB& _db)
 				break;
 		}
 		catch (...) {}
+		if (l == 0)
```
chriseth: I'm not sure this is the right thing to do. The block with number 0 was deemed invalid above, so we should actually re-generate the genesis block.

winsvega: I have a test where this cycle in this function goes l = 0;  l--;   l = 497....  
this way it would run endless.

gavofyork:
Fix made in whizz branch.

PR #2513, opened 2015-07-19
LLVM version requirements in cmake

chfast:
Currently brew installs version 3.8 as 3.7 has been branched. Version 3.7 is not available in homebrew as it has not been released yet. Ignore LLVM version requirement until 3.7 is released.


subtly:
I think homebrew only installs 3.8 if --HEAD is passed. Otherwise, 3.6.

debris:
3.8 installed via homebrew has version name `3.8.0svn` and it's incompatible with cmake's naming convention. If it was `3.8.0`, it would be found properly.

subtly:
What flags were used to install llvm from homebrew, which causes evmjit buildbot test to pass?

The issue is that by default, homebrew doesn't install 3.8, instead it installs *3.6*; removing the required cmake version will cause cmake to try and build against 3.6.

I've also installed llvm from head w/homebrew and it crashed when running w/evmjit. Looking into this.

debris:
> What flags were used to install llvm from homebrew, which causes evmjit buildbot test to pass?

It was only `--HEAD`

>The issue is that by default, homebrew doesn't install 3.8, instead it installs 3.6; removing the required cmake version will cause cmake to try and build against 3.6.

So it might be worth checking version number manually. eg.

```diff
- find_package(LLVM 3.7 REQUIRED CONFIG)
+ find_package(LLVM CONFIG)
+ if (NOT regexp_to_check_version_number 3.7/3.8svn ... )
+   message(FATAL_ERROR "Could not find a configuration file for package LLVM that is compatible
+   with requested version")
+ endif()
```

chfast:
Correct. And 3.8 is fine for the jit.

chfast:
It's all temporary. I'm waiting for LLVM 3.7 to be released next month. Then I will reconfigure the scripts again to require that specific version.

PR #2518, opened 2015-07-20
ensurePrecomputed should respect --no-precompute

LefterisJP:


LefterisJP:
This may be related to: https://github.com/ethereum/go-ethereum/issues/1501

gavofyork:
would prefer to avoid all these statics. we have the options API in `Sealer` now to allow for passing all these sorts of things.

LefterisJP:
Rebased, and will now open a PR targetting develop and using the Sealers API.

PR #2525, opened 2015-07-21
Fix clang warnings

LefterisJP:


LefterisJP:
Until github [allows us to change the target branch of a PR](https://github.com/isaacs/github/issues/18) we will need to be opening a new PR. 

So this is the clang warning PR retargetted towards latest develop:

PR #2526, opened 2015-07-21
Fixed call to contract from rpc

arkpar:


libethereum/Executive.cpp
```diff
@@ -400,10 +400,14 @@ void Executive::finalize()
 	m_gas += m_refunded;
 
 	//	cnote << "Refunding" << formatBalance(m_endGas * m_ext->gasPrice) << "to origin (=" << m_endGas << "*" << formatBalance(m_ext->gasPrice) << ")";
-	m_s.addBalance(m_t.sender(), m_gas * m_t.gasPrice());
```
arkpar: Executive is created without a transaction from Client::call and this throws an exception

libethereum/Executive.cpp
```diff
@@ -400,10 +400,14 @@ void Executive::finalize()
 	m_gas += m_refunded;
 
 	//	cnote << "Refunding" << formatBalance(m_endGas * m_ext->gasPrice) << "to origin (=" << m_endGas << "*" << formatBalance(m_ext->gasPrice) << ")";
```
chriseth: Is this comment duplicated now?

PR #2527, opened 2015-07-21
ensurePrecomputed should respect no_precompute argument V2

LefterisJP:
Continuation of #2518 using the Sealers API and targetting the develop branch.

### DIFF
--- a/libethcore/EthashSealEngine.cpp
+++ b/libethcore/EthashSealEngine.cpp
@@ -54,7 +54,8 @@ void EthashSealEngine::generateSeal(BlockInfo const& _bi)
 	m_farm.setWork(m_sealing);
 	m_farm.start(m_sealer);
 	m_farm.setWork(m_sealing);		// TODO: take out one before or one after...
-	if (option("precomputeDAG")[0] == 1)
+	bytes shouldPrecompute = option("precomputeDAG");
+	if (!shouldPrecompute.empty() && shouldPrecompute[0] == 1)
 		Ethash::ensurePrecomputed((unsigned)_bi.number());
 }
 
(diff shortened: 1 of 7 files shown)


libethcore/EthashSealEngine.cpp
```diff
@@ -54,7 +54,8 @@ void EthashSealEngine::generateSeal(BlockInfo const& _bi)
 	m_farm.setWork(m_sealing);
 	m_farm.start(m_sealer);
 	m_farm.setWork(m_sealing);		// TODO: take out one before or one after...
-	Ethash::ensurePrecomputed((unsigned)_bi.number());
+	if (option("precomputeDAG")[0] == 1)
```
arkpar: Since EthashSealEngine class does not guarantee that the option is set a check should be here.

libethereum/Client.h
```diff
@@ -140,6 +140,8 @@ class Client: public ClientBase, protected Worker
 	bool turboMining() const { return m_turboMining; }
 	/// Enable/disable GPU mining.
 	void setTurboMining(bool _enable = true) { m_turboMining = _enable; if (isMining()) startMining(); }
+    /// Enable/disable precomputing of the DAG for next epoch
+    void setShouldPrecomputeDAG(bool _precompute);
```
arkpar: Something wrong with indent here?

subtly:
Would it make more sense for this to be an Ethash configuration or option? Does this need to be a dynamically configured option?

LefterisJP:
In the beginning I went with yet another static `Ethash` member variable but Gavin asked to use the new Sealers API. See the PR to which I linked in the beginning. I don't think it's bad to have options dynamic like that, as long as it's nicely organized.

LefterisJP:
Oh ... and just like the previous PR, this should probably help in fixing the geth precompute issues seen here: https://github.com/ethereum/go-ethereum/issues/1501

subtly:
The precompute issue?  I don't understand the problem in that issue.
Run eth or geth with mining disabled. Ethminer computes dag. Guess we can always clean it up later.

LefterisJP:
The issue is that ethminer has this `--no-precompute` flag. And even when given this argument it kept precomputing the DAG.

PR #2530, opened 2015-07-21
Worker refactoring.

chriseth:
Split `m_state` into `m_requestedState` (controlled from outside) and `m_state` (controlled by the thread).

If this PR makes the system more unstable, we can postpone it to after frontier.

### DIFF
--- a/libdevcore/Worker.h
+++ b/libdevcore/Worker.h
@@ -29,37 +29,25 @@
 namespace dev
 {
 
-enum class IfRunning
-{
-	Fail,
-	Join,
-	Detach
-};
-
 enum class WorkerState
 {
-	Starting,
-	Started,
-	Stopping,
-	Stopped,
-	Killing
+	Running,			///< thread started and running
+	Stopped,			///< thread stopped, but still exists, ran be re-started
+	Terminated			///< thread does not exist
 };
 
-class Worker
+class Worker: boost::noncopyable
 {
 protected:
-	Worker(std::string const& _name = "anon", unsigned _idleWaitMs = 30): m_name(_name), m_idleWaitMs(_idleWaitMs) {}
-
-	/// Move-constructor.
-	Worker(Worker&& _m) { std::swap(m_name, _m.m_name); }
-
-	/// Move-assignment.
-	Worker& operator=(Worker&& _m) { std::swap(m_name, _m.m_name); return *this; }
+	/// Creates a new worker initially in the "Killed" state.
+	explicit Worker(std::string const& _name = "anon", unsigned _idleWaitMs = 30):
+		m_name(_name),
+		m_idleWaitMs(_idleWaitMs),
+		m_state{ WorkerState::Terminated },
+		m_requestedState{ WorkerState::Terminated }
+	{}
 
 	virtual ~Worker() { terminate(); }
-
-	/// Allows changing worker name if work is stopped.
-	void setName(std::string _n) { if (!isWorking()) m_name = _n; }
 
 	/// Starts worker thread; causes startedWorking() to be called.
 	void startWorking();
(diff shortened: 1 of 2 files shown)


subtly:
This changes semantics. If there is a problem then it needs to be corrected first before changing functionality.

What is being fixed? Can you write a test which demonstrates the race condition? Where is the current codebase hitting this condition?

This must be tested thoroughly with networking and mining before being merged.

libdevcore/Worker.h
```diff
@@ -29,74 +29,60 @@
...
-	Join,
-	Detach
-};
-
 enum class WorkerState
 {
-	Starting,
-	Started,
-	Stopping,
-	Stopped,
```
subtly: Where are killing and starting?

chriseth: `Starting` and `Killing` is now `m_requestedState == WorkerState::Started` / `WorkerState::Killing`, since "Starting" and especially "Killing" is not really the state of the worker (see description of this PR).
Perhaps it is better to view this PR not in the diff but in the full view. Note the absence of any of the special functions of `std::atomic`.

subtly:
What is being fixed? Is there a problem with the current code?

libdevcore/Worker.h
subtly: This makes sense and it's a reversion to the prior logic with support for terminate().

So the requested state has been separated from the actual state. In total, removing about two LOC and adding another atomic. Are two LOC worth the additional state? Why introduce additional state and complexity when it isn't necessary?

libdevcore/Worker.h
```diff
@@ -29,74 +29,60 @@
...
-	unsigned m_idleWaitMs = 0;
+	unsigned const m_idleWaitMs = 0;
 	
-	mutable Mutex x_work;						///< Lock for the network existance.
-	std::unique_ptr<std::thread> m_work;		///< The network thread.
-	std::atomic<WorkerState> m_state = {WorkerState::Starting};
+	mutable Mutex x_work;						///< Lock for the worker thread.
+	std::unique_ptr<std::thread> m_work;		///< The worker thread.
+	std::atomic<WorkerState> m_state;			///< The current state of the thread, controlled by the thread.
+	std::atomic<WorkerState> m_requestedState;	///< The desired state of the thread, controlled from outside.
```
subtly: Mutations of m_requestedState are binary, guarded, and reads are thread-safe. Atomic is unnecessary overkill. Worker doesn't provide, nor does it intend to provide real-time guarantees -- while(...) sleep() negates such a possibility (we do not know the value of any timer when state change is requested or enacted).

libdevcore/Worker.h
```diff
@@ -29,74 +29,60 @@
...
 
-	unsigned m_idleWaitMs = 0;
+	unsigned const m_idleWaitMs = 0;
 	
-	mutable Mutex x_work;						///< Lock for the network existance.
-	std::unique_ptr<std::thread> m_work;		///< The network thread.
-	std::atomic<WorkerState> m_state = {WorkerState::Starting};
+	mutable Mutex x_work;						///< Lock for the worker thread.
+	std::unique_ptr<std::thread> m_work;		///< The worker thread.
+	std::atomic<WorkerState> m_state;			///< The current state of the thread, controlled by the thread.
```
subtly: The "Terminated" state is never used. This should be a bool (or a different enum).

gavofyork:
Worker certainly needs a refactor and the notion of splitting actual state from requested state seems sound. I'd like to see some exhaustive tests written for it first and then the class altered to use condition variables and to include checks to make sure it's checking whether it `shouldStop` often enough or that `doWork` isn't taking too long.

PR #2535, opened 2015-07-22
After DAG load progress should be 100 to continue

LefterisJP:
Since the merging of the whizz branch we have serious problems with mining. This solves only one issue. Other problems still exist.

When the DAG already exists as a file, we never get inside the callback and as such never get to give the value of 100 to the progress variable.

This will allow us to continue with mining but it will still get stuck in the search loop, never finding a solution.

I am testing on a private chain with CPU mining to find solutions to the other issues too.


gavofyork:
invalid alteration in logic.

PR #2536, opened 2015-07-22
Update m_working Client state for private chain mining

LefterisJP:
When mining on a private chain, mining would get stuck and not
start. The reason is that the difficulty was zero. If we move the
mining state from m_premine to m_working the difficulty will get updated
properly and we will be able to mine in a private chain.

Based on #2535

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -651,6 +651,8 @@ void Client::onPostStateChanged()
 void Client::startMining()
 {
 	m_wouldMine = true;
+	DEV_WRITE_GUARDED(x_working)
+		m_working = m_preMine;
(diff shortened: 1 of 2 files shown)


libethereum/Client.cpp
```diff
@@ -651,6 +651,8 @@ void Client::onPostStateChanged()
 void Client::startMining()
 {
 	m_wouldMine = true;
+	DEV_WRITE_GUARDED(x_working)
```
gavofyork: would need `m_preMine` to be read locked at the least.

really this logic should be in `rejigMining`, though.

gavofyork:
fixed properly in latest develop.

PR #2537, opened 2015-07-22
bugfix: reap() refactored

gluk256:


test/libp2p/peer.cpp
```diff
@@ -40,6 +40,7 @@ BOOST_FIXTURE_TEST_SUITE(p2p, P2PFixture)
 
 BOOST_AUTO_TEST_CASE(host)
 {
+	cnote << "p2p/host...";
```
LefterisJP: are these left here by accident? It may become annoying if we spam `stdout` when runnign the tests.

PR #2542, opened 2015-07-22
Concurrent queue

chfast:


libdevcore/concurrent_queue.h
```diff
@@ -0,0 +1,60 @@
...
+
+namespace dev
+{
+
+/// Concurrent queue.
+/// You can push and pop elements to/from the queue. Pop will block until the queue is not empty.
+/// The default backend (_QueueT) is std::queue. It can be changed to any type that has
+/// proper push(), pop(), empty() and front() methods.
+template<typename _T, typename _QueueT = std::queue<_T>>
+class concurrent_queue
```
chriseth: Casing

chfast: vector_ref?

libevm/SmartVM.cpp
```diff
@@ -93,18 +83,11 @@ namespace
 
 		~JitWorker()
 		{
-			DEV_GUARDED(x_mutex)
-				m_finished = true;
-			m_cv.notify_one();
+			push(JitTask::createStopSentinel());
 			m_worker.join();
```
chriseth: Do I understand this correctly that the worker keeps on compiling after it reaches the end of the queue and we are waiting here in the destructor for this to finish?

chfast: Yes.

libdevcore/concurrent_queue.h
chriseth: `vector_ref` is a remnant of ancient times, but if you insist, you could probably plead to the "extending STL functionality" clause.

chfast: My intend was to make it std library extension (no other than standard headers included). But I don't care much about the names. However, it would be nice to replace that with a different external implementation at some point.

libevm/SmartVM.cpp
chfast: Extended comment: this design is quite bad, but it behaves pretty nice in this case. This queue never holds more than one element as JIT compilation finishes before transaction processing finishes.

It would be nice to add clear() here, but std::queue do not have one...

arkpar: You could use swap with an empty queue to clear

arkpar:
A few things missing to make this class generally re-usable:

* Cancel wait method.
* Handle the destruction properly when there are active waiters.
* A method to insert a range of items at once

libevm/SmartVM.cpp
chfast: If we want more than std::queue provides we should drop the adapter. And I will do that.

chfast:
@arkpar, If you need more features send patches please.

chfast:
@arkpar , or describe the feature in details. I'm quite sure you are wrong about extending the destructor.

arkpar:
I'm not suggesting adding these features right now.
What will happen now if a class is destroyed while someone is waiting in the pop method? How do I cancel queue processing without using some obscure sentinel item?

chfast:
There are 2 alternatives for stopping the consumer thread:
1. Use additional external notification solution. That is bad as you would need additional synchronization.
2. Implement a "closed" state of the queue. That is bad too, as it makes the queue and the consumer more complex.

Concurrent queue blocks if empty on consumer side. End of story. The "stop task" is the classic way of informing the consumer it has to stop.

The user is _always_ responsible for synchronization of the queue destruction. It must make sure no other thread has the reference to the queue _before_ the destruction starts. In other words it must stop the consumer first.

libdevcore/concurrent_queue.h
gavofyork: if you're going to invoke the "just extends STL functionality" clause, then it must not include any other dev headers. The point of `vector_ref` being lower-underscore was that it feels equally comfortable in a pure STL setting and thus without any dev infrastructure.

PR #2545, opened 2015-07-22
Do not accept the point at infinity as valid public key for ecrecover.

### LINKED ISSUE #2544: EC recover returns address for invalid sigs where pubkey is point at infinity
https://github.com/ethereum/tests/pull/94

This affects both txs and the precompiled EC recover contract.

YP Appendix E

chriseth:
This is only a fix for CryptoPP which is currently not used.

Partly fixes #2544


Gustav-Simonsson:
@subtly ping

Issue #2550, opened 2015-07-22
Buildbot & JIT & testing


chfast:
This a small pack of changes I would like to apply to the project. In the end it should save us some buildbot resources. I want to know your opinion.

- [x] **Build with EVM JIT by default.**

We recently enabled JSCONSOLE and OpenCL as we want ship with these features. Bots are able to build with JIT enabled. It's time to enable it by default.

- [ ] **Use `ctest`**

CTest is a CMake extension that allows defining tests inside cmake scripts. A test can be an execution of `testeth`. Someone has tried to use it before by defining a ctest test for every Boost test case. I think that's too detailed level. We should define ctest tests as a packs of test cases: core, network, whisper, transactions, transactions-with-jit, etc.
Buildbot should run tests just with executing `ctest`.
Alternatively we can create separated builtbot steps for every test subject (network, ethereum, whisper, etc).

- [ ] **Kill evmjit builders**

After previous points applied we can kill exactly 6 buildbot builders:
{Linux, OSX} ⨯ {master, develop, pull request} ⨯ {evmjit}.

PR #2552, opened 2015-07-23
Build EVM JIT by default

chfast:
Details: #2550

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -355,9 +355,14 @@ else()
 endif()
 
 if (EVMJIT)
-	if (CMAKE_SYSTEM_NAME STREQUAL "Windows" AND NOT DEFINED LLVM_DIR)
-		set(LLVM_DIR "${CMAKE_SOURCE_DIR}/extdep/install/windows/x64/share/llvm/cmake")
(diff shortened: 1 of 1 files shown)


LefterisJP:
Looks good to me as long as all buildbots pass. And it's so awesome we can now kill all the evmjit buildbots :)

chfast:
All bots green!

caktux:
Even though all checks are green, this will break the PPA builds, and for OSX users that don't have LLVM installed with --HEAD and are not installing from bottled packages...

gavofyork:
is there an alterative to waiting for august for this?

could we disable for mac, disable at compile time if llvm version too long (or disable at runtime if llvm version too low)?

i want as many platforms building with llvm by default asap.

caktux:
Disabling at compile time seems like a good solution; brew bottles should work fine and users that want to build from source (either with brew or not) can follow the instructions, but the PPA would take a little more work. Since LLVM is using its own repo (not a PPA), we can't add it for those builds; but we could just copy the packages in our own PPA instead...

chfast:
Not sure if that answers your questions, but we can make EVMJIT option somehow automatic. If not specified by user cmake can check if correct version of LLVM is available in the system.

PR #2554, opened 2015-07-23
Install NSIS.

chfast:


debris:
What will happen if NSIS is already installed?

chfast:
This is just hacking...

PR #2557, opened 2015-07-23
Parse code as data if only hex.

chriseth:
This allows to use some additional features as padding for integers
(useful because contract constructor arguments go after the code).

### DIFF
--- a/alethzero/Transact.cpp
+++ b/alethzero/Transact.cpp
@@ -226,14 +226,8 @@ static tuple<vector<string>, bytes, string> userInputToCode(string const& _user,
 	string solidity;
 	bytes data;
 	vector<string> errors;
-	if (_user.find_first_not_of("1234567890abcdefABCDEF\n\t ") == string::npos && _user.size() % 2 == 0)
-	{
-		std::string u = _user;
-		boost::replace_all_copy(u, "\n", "");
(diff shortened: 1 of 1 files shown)


alethzero/Transact.cpp
```diff
@@ -226,14 +226,8 @@ static tuple<vector<string>, bytes, string> userInputToCode(string const& _user,
...
-	if (_user.find_first_not_of("1234567890abcdefABCDEF\n\t ") == string::npos && _user.size() % 2 == 0)
-	{
-		std::string u = _user;
-		boost::replace_all_copy(u, "\n", "");
-		boost::replace_all_copy(u, "\t", "");
-		boost::replace_all_copy(u, " ", "");
-		data = fromHex(u);
-	}
+	if (_user.find_first_not_of("1234567890abcdefABCDEF\n\t ") == string::npos)
+		data  = parseData(_user);
```
gavofyork: two spaces after `data`.

gavofyork:
`parseData` is meant for parsing the data fields of a transaction, not for specifying the code of a contract-creation transaction. `userInputToCode` is strictly for contract creation transactions and usage of `parseData` in this situation would likely lead to confusion over the resultant data should whitespace be (accidentally, perhaps) introduced.

PR #2566, opened 2015-07-24
Mix - deployment - bug fix during contract call

yann300:


mix/qml/js/NetworkDeployment.js
```diff
@@ -157,7 +157,11 @@ function executeTr(blockIndex, trIndex, state, ctrAddresses, trHashes, callBack)
 		if (tr.contractId === tr.functionId)
 			rpcParams.code = codeModel.contracts[tr.contractId].codeHex + encodedParams.join("");
 		else
-			rpcParams.data = "0x" + func.qhash() + encodedParams.join("");
+        {
```
chriseth: Indentation

PR #2581, opened 2015-07-25
Make all of our memory that stores secrets auto-cleansing through use of special types.

### LINKED ISSUE #2580: SEC-2 Secrets Not Cleared from Memory

### LINKED ISSUE #2582: SEC 4 VM Instruction Callback Function Receives Un-validated  Input
The input given for the callback is perfectly "valid", in so much as it is faithful to the EVM code's requirements. It the requirements happen to be astronomical, then the debuggers and/or tracing tools should be made aware of this through the accordingly large values.

SHA3 is a reasonable suggestion, though, since this would allow an attacker (or indeed a victim through accident) to craft code which is able to appear to the debugger as though it needs little gas where it actually needs 2^256 + N.

gavofyork:
- Introduce vector_ref::cleanse, code adapted from openSSL.
- Introduce auto-cleansing secure_vector and bytesSec.
- Make Secret auto-cleansing.
- Avoid copying Secrets as much as possible.

Closes #2580 

- SHA3 gas calculation avoids overflow of `+31` through use of bigint.

Closes #2582


libethcore/Common.cpp
```diff
@@ -65,7 +65,7 @@ Network resetNetwork(Network _n)
 	c_gasLimitBoundDivisor = 1024;
 	c_minimumDifficulty = 131072;
 	c_difficultyBoundDivisor = 2048;
-	c_durationLimit = c_network == Network::Turbo ? 2 : c_network == Network::Olympic ? 8 : 12;
+	c_durationLimit = c_network == Network::Turbo ? 2 : c_network == Network::Olympic ? 8 : 13;
```
arkpar: Is this in sync with other clients?

libdevcore/Common.h
```diff
@@ -79,6 +79,16 @@ using bytes = std::vector<byte>;
 using bytesRef = vector_ref<byte>;
 using bytesConstRef = vector_ref<byte const>;
 
+template <class T>
+class secure_vector: public std::vector<T>
+{
+public:
+	template <class ... Args> secure_vector(Args&& ... _args): std::vector<T>(_args ...) {}
```
chriseth: Would `std::forward` be beneficial here?

libdevcore/FixedHash.h
```diff
@@ -214,6 +214,8 @@ class FixedHash
 		return ret;
 	}
 
+	void clear() { m_data.fill(0); }
```
chriseth: Rather use `memset_s`?

libdevcore/Common.h
```diff
@@ -79,6 +79,16 @@ using bytes = std::vector<byte>;
 using bytesRef = vector_ref<byte>;
 using bytesConstRef = vector_ref<byte const>;
 
+template <class T>
+class secure_vector: public std::vector<T>
```
chriseth: We should either force this to fixed size and prevent copying or somehow invoke `cleanse` also upon copy and resize.

chriseth:
I guess we should hold secrets in non-copyable containers and maintain shared pointers to them, to be really sure about where they are located, when they are created and destroyed.

libdevcore/FixedHash.h
arkpar: memset_s is a C11 standard (not C++). Not supported by msvc unfortunately.

gavofyork:
disabling copying seems a reasonable safeguard. once it goes into secp256k1 it's going to insecurely left in memory anyway, so it's a damage limitation exercise at best. might ultimately be best to refactor into a truly special class for which a release build implies all data is private with no accessors (other than a friend to the ecdsa_sign gateway function).

CMakeLists.txt
```diff
@@ -292,8 +295,23 @@ elseif (BUNDLE STREQUAL "miner")
...
+elseif (BUNDLE STREQUAL "default")	# development builds
+	set(SERPENT ${DECENT_PLATFORM})
+	set(SOLIDITY ON)
+	set(USENPM OFF)
+	set(GUI ON)
+	set(TOOLS ON)
+	set(TESTS ON)
+	set(FATDB ON)
+	set(ETHASHCL ON)
+	set(EVMJIT ON)
```
CJentzsch: I don't know if this is a good idea. How can we make sure that the correct version of LLVM is used?

alethzero/MainWin.cpp
```diff
@@ -847,28 +846,14 @@ void Main::readSettings(bool _skipGeometry)
...
 		m_myIdentities.clear();
 		QByteArray b = s.value("identities").toByteArray();
 		if (!b.isEmpty())
 		{
-			h256 k;
+			Secret k;
 			for (unsigned i = 0; i < b.size() / sizeof(Secret); ++i)
 			{
-				memcpy(&k, b.data() + i * sizeof(Secret), sizeof(Secret));
+				memcpy(k.writable().data(), b.data() + i * sizeof(Secret), sizeof(Secret));
```
LefterisJP: here you can have `sizeof(*k)` instead of `sizeof(Secret)`. A tiny bit more future-proof in case the type of K changes.

libethereum/BlockChain.h
```diff
@@ -292,13 +294,20 @@ class BlockChain
 protected:
 	static h256 chunkId(unsigned _level, unsigned _index) { return h256(_index * 0xff + _level); }
 
-	/// Initialise everything and open the database.
+	/// Initialise everything and ready for openning the database.
```
LefterisJP: typo in 'openning'

libwhisper/CMakeLists.txt
```diff
@@ -21,10 +21,10 @@ file(GLOB HEADERS "*.h")
 
 add_library(${EXECUTABLE} ${SRC_LIST} ${HEADERS})
 
-target_link_libraries(${EXECUTABLE} ethcore)
+#target_link_libraries(${EXECUTABLE} ethcore)
```
LefterisJP: did the commenting out go unnoticed? Maybe just remove?

LefterisJP:
I finished my reviewing. Looks good apart from the comments I made. I would really feel much more comfortable though if I could see all those modified tests pass for all buildbots.

CMakeLists.txt
gavofyork: runtime check? in any case this won't be active until the buildbots are switch to the bundle called default.

PR #2586, opened 2015-07-27
increase frontier blocktime to 13s

CJentzsch:
according to https://github.com/ethereum/common/blob/master/params.json#L8


arkpar:
same change is already in other PR:
https://github.com/ethereum/cpp-ethereum/pull/2581/files#diff-4e01855c4603e806caac0a73498394caL68

PR #2608, opened 2015-07-28
add extcodecopy stack test

CJentzsch:


chriseth:
The code looks quite complicated to me for such a test.

CJentzsch:
Yes, it come from a fuzz test, so the code has been created randomly, but hits the problem. I could add a more minimal example later (after my vacation, and days before Frontier my todo list is expoding :-) )

PR #2618, opened 2015-07-28
Clear state on dropping transactions from queue

arkpar:


libethereum/Client.cpp
```diff
@@ -650,6 +651,21 @@ void Client::resyncStateFromChain()
 	}
 }
 
+void Client::resetState()
+{
+	State newPreMine;
+	DEV_READ_GUARDED(x_preMine)
```
gavofyork: maybe add a few `{`s - we're not python :)

arkpar: This was just some wrong indent. The style for two guards on the same line take from resyncStateFromChain above.

PR #2626, opened 2015-07-29
check uncles (including difficulty and gasLimit)

CJentzsch:


CJentzsch:
(builds only fail because of timeout)

PR #2627, opened 2015-07-29
One for caktux: Bundle now only sets default, other options can override.

gavofyork:


CMakeLists.txt
```diff
@@ -231,110 +313,11 @@ if (GUI)
 endif()
 
 # Default CMAKE_BUILD_TYPE to "Release".
```
LefterisJP: Change comment since it no longer reflects the next line

CMakeLists.txt
```diff
@@ -231,110 +313,11 @@ if (GUI)
...
-	set(JSCONSOLE ON)
-	set(JSONRPC ON)
-	set(FRONTIER ON)
-endif ()
+set(CMAKE_BUILD_TYPE CACHE STRING ${D_CMAKE_BUILD_TYPE})
 
-if ("x${CMAKE_BUILD_TYPE}" STREQUAL "x")
-	set(CMAKE_BUILD_TYPE "Release")
-endif ()
+#if ("x${CMAKE_BUILD_TYPE}" STREQUAL "x")
```
LefterisJP: Why not simply remove instead of commenting out?

PR #2628, opened 2015-07-29
Make Nonce generation securely random

gavofyork:
SEC-5 Nonce Generation Uses a Deterministic Seed in `libdevcrypto/Common.cpp` on CPP Ethereum
SEC-7 Private Key Recovery from Predictable Transaction Nonces


CJentzsch:
looks good, but doesn't compile

gavofyork:
compilation breaking is in develop and has been fixed out of band.

PR #2651, opened 2015-07-30
Enable usage of llvm3.8-dev in Ubuntu

LefterisJP:
Ubuntu 14.04 replaced llvm3.7-dev with llvm3.8-dev in anticipation of llvm 3.7 release.
We are forced to use the 3.8-dev package until llvm 3.7 release.


gavofyork:
won't this fail when only 3.7 is available? it should work with both...

LefterisJP:
need someone to test with ubuntu 14.04 already having llvm 3.7-dev installed. I tested it in a clean Ubuntu where I downloaded llvm3.8-dev and it works.

gavofyork:
shooting from the hip.

PR #2658, opened 2015-07-30
Hot fix/workaround for assert failure.

gavofyork:


gavofyork:
needs properly fixing - @arkpar thoughts?

arkpar:
I'll check it more thoroughly later

PR #2664, opened 2015-07-30
windows build bugfix

gluk256:


gavofyork:
is this needed? build bots were build without, no?

LefterisJP:
I don't think this would work

gluk256:
For some reason, I could only build with this fix. However, after repeating the whole build process from scratch, I don't get this error any more. I guess, this PR should be deleted.

PR #2668, opened 2015-07-31
Attempting to fix the build in macosx

LefterisJP:


LefterisJP:
So with this PR we find out that as long as we do 
```
if (APPLE)
	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libstdc++")
	set_property(TARGET ${EXECUTABLE} PROPERTY XCODE_ATTRIBUTE_CLANG_CXX_LIBRARY "libstdc++")
endif()
```
for libjsv8 we can't have C++11 in there

PR #2669, opened 2015-07-31
Buildfix for MacOS.

chriseth:


chriseth:
Builds fine now, errors are in (unrelated) tests.

PR #2670, opened 2015-07-31
require MiningBenefactor to be set in order to mine

CJentzsch:
because of this: https://etherchain.org/account/0x0000000000000000000000000000000000000000


LefterisJP:
:+1:

PR #2683, opened 2015-08-03
Thawing

CJentzsch:
* set target BlockGasLimit to 3141592
* set default GasPrice to 50 shannon (10^9 wei = 10^-9 ether)

### DIFF
--- a/libethcore/Common.h
+++ b/libethcore/Common.h
@@ -80,6 +80,7 @@ template <> inline u256 exp10<0>()
 static const u256 ether = exp10<18>();
 static const u256 finney = exp10<15>();
 static const u256 szabo = exp10<12>();
+static const u256 shannon = exp10<9>();
 static const u256 wei = exp10<0>();
 
 using Nonce = h64;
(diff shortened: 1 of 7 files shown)


CJentzsch:
(builds only fail due to timeout)

gavofyork:
would be nice to remove all of those default arguments....

CJentzsch:
ok, 2 ways of doing it. 1) specify the default gasPrice globally (similiar to params.cpp) 2) having GasPricer objects everywhere a default gasPrice is needed. so this number only needs to be set at one place.
What do you prefer?

chriseth:
Was unable to review the merge, but the other commits look good.

CJentzsch:
on some builds this test seem to fail:
`/Users/administrator/slave/build-cpp-ethereum-osx-pr/build/test/libp2p/net.cpp:356: fatal error in "deadlineTimer": critical check fired == 1 failed
unknown location:0: fatal error in "deadlineTimer": signal: SIGABRT (application abort requested)
/Users/administrator/slave/build-cpp-ethereum-osx-pr/build/test/libp2p/net.cpp:356: last checkpoint`
But this is absolutely unrelated to this PR.

PR #2688, opened 2015-08-03
Allow explicit conversions bytes <-> string.

chriseth:


chriseth:
Only network tests fail.

PR #2694, opened 2015-08-04
Break off contract tests.

chriseth:


libsolidity/CompilerUtils.cpp
```diff
@@ -469,6 +469,13 @@ void CompilerUtils::convertType(Type const& _typeOnStack, Type const& _targetTyp
 			}
 			break;
 		}
+		case DataLocation::CallData:
+			solAssert(
+					targetType.isByteArray() &&
```
LianaHus: wrong indent here

PR #2701, opened 2015-08-04
Fix for strings as keys in mappings.

chriseth:


gavofyork:
looks at least as likely to work as it did previously. are there any tests?

chriseth:
There is no direct test, but it caused a failure in the auction registrar test.
Will add test for this specific bug when I have more time.

Issue #2696, opened 2015-08-04
Windows 7 x64 Nvidia GTX 760 2Gb mining issue


42mind:
Hello guys,

I have installed geth, synchronised with the network, downloaded latest Etehreum Win64 package. 
When I'm running ethminer with -G flag, I getting "No GPU device with sufficient memory was found. Can't GPU mine. Remove the -G ar
gument" error. Latest official Nvidia driver is installed.

Guys, do you have any ideas what to do next to resolve this issue?

Here is some useful information from the system.
https://15254b2dcaab7f5478ab-24461f391e20b7336331d5789078af53.ssl.cf1.rackcdn.com/ethereum.vanillaforums.com/editor/hf/037g3k6q06r2.png

https://15254b2dcaab7f5478ab-24461f391e20b7336331d5789078af53.ssl.cf1.rackcdn.com/ethereum.vanillaforums.com/editor/dv/d3wlhch6fb6f.png

https://15254b2dcaab7f5478ab-24461f391e20b7336331d5789078af53.ssl.cf1.rackcdn.com/ethereum.vanillaforums.com/editor/i2/bnexu2qhzqx5.png

gavofyork:
nvidia drivers tend not to be able to allocate a single 1GB+ chunk of memory, necessary to be able to mine at present. reworking this is a non-trivial undertaking; patches for a so-called "chunked" implementation are most welcome :)

42mind:
@gavofyork so it means, i should wait for a patch? :)
Btw, mobile Nvidia on Mbpro is good with mining. 
Is it any workaround? Maybe it will work on Linux? 

Thank you!

subtly:
Please try setting GPU_MAX_ALLOC_PERCENT to a large value (95 or 100; lower if main display shares GPU).

42mind:
@subtly I have tried, with no success. Thank you for your advice.

hitchcott:
I seem to be having the same issue with my GTX 960. Setting GPU_MAX_ALLOC_PERCENT does not fix it.

Using `--list-devices`:
```
[0] GeForce GTX 960
        CL_DEVICE_TYPE: GPU
        CL_DEVICE_GLOBAL_MEM_SIZE: 4294967296
        CL_DEVICE_MAX_MEM_ALLOC_SIZE: 1073741824
        CL_DEVICE_MAX_WORK_GROUP_SIZE: 1024
```

LefterisJP:
[...]

PR #2702, opened 2015-08-04
Add more info on CL device info printing

LefterisJP:
Could be able to give us quick assistance regarding what a user's OpenCL drivers see for each card as opposed to its actual specs. Just like for example in #2696


subtly:
As mentioned several times -- there isn't any debug information or parameters in code that will fix this. As I conveyed to Gustav, the users need to set the environment variable GPU_MAX_ALLOC_PERCENT to a reasonable value.

LefterisJP:
@subtly Granted, this may not help with the memory issue but having more information printed by the 
```
./eth/eth --cl-list-devices 
```
is not a bad thing.

PR #2703, opened 2015-08-04
block refactoring

CJentzsch:


libethereum/Account.cpp
```diff
@@ -20,11 +20,67 @@
...
+			BOOST_THROW_EXCEPTION(ValueTooLarge() << errinfo_comment("State value is equal or greater than 2**256") );
+		return (u256)ret;
+	};
+
+	std::unordered_map<Address, Account> ret;
+
+	js::mValue val;
+	json_spirit::read_string(_json, val);
+
+	for (auto account: val.get_obj().count("alloc") ? val.get_obj()["alloc"].get_obj() : val.get_obj())
```
CJentzsch: I changed to line from Gavs original version. When calling this function in CanonBlockChain<Ethash>::createGenesisState(), a string with genesis block description is given, not the map directly, therefore I get the "alloc" object here when its there. Could be done in a cleaner way, having another version of this function.

eth/main.cpp
```diff
@@ -904,20 +904,19 @@ void interactiveMode(eth::Client* c, std::shared_ptr<eth::TrivialGasPricer> gasP
 		}
 		else if (c && cmd == "dumptrace")
```
CJentzsch: I have not tested those changes yet. Just make it compile in I think its correct.

alethzero/MainWin.cpp
```diff
@@ -1878,7 +1879,7 @@ void Main::on_dumpBlockState_triggered()
 		{
 			f << "{" << endl;
 //			js::mObject s;
-			State state = ethereum()->state(h);
+			State state = ethereum()->block(h).state();
```
CJentzsch: not tested this, this was just a quick fix to make it compile, please review carefully.

libethereum/Client.cpp
```diff
@@ -653,7 +654,7 @@ void Client::resyncStateFromChain()
 
 void Client::resetState()
 {
-	State newPreMine;
+	Block newPreMine;
```
CJentzsch: similiar, not tested just a compile fix which I think is correct. Review carefully.

test/TestHelper.cpp
```diff
@@ -94,22 +95,43 @@ struct MissingFields : virtual Exception {};
...
+				importState(_o["post"].get_obj(), m_statePost);
+			else
+				importState(_o["postState"].get_obj(), m_statePost);
+			m_logsExpected = importLog(_o["logs"].get_array());
+		}
 	}
 }
 
+//executes an imported transacton on preState
+bytes ImportTest::executeTest()
```
CJentzsch: we will later (in another PR) refactor the test code, this will be moved, since it is not related to importing the test.

test/libethereum/blockchain.cpp
```diff
@@ -98,7 +107,7 @@ void doBlockchainTests(json_spirit::mValue& _v, bool _fillin)
 
 		// construct true blockchain
 		TransientDirectory td;
-		FullBlockChain<Ethash> trueBc(rlpGenesisBlock.out(), StateDefinition(), td.path(), WithExisting::Kill);
+		FullBlockChain<Ethash> trueBc(rlpGenesisBlock.out(), AccountMap(), td.path(), WithExisting::Kill);
 
 		if (_fillin)
```
CJentzsch: A lot of the code down from here is commented out for now. This is only for the creation of BlockChainTests, not for actually running them. This is rather complex and will get fixed later.

eth/main.cpp
```diff
@@ -904,20 +904,19 @@ void interactiveMode(eth::Client* c, std::shared_ptr<eth::TrivialGasPricer> gasP
...
 
-			dev::eth::State state = c->state(index + 1,c->blockChain().numberHash(block));
-			if (index < state.pending().size())
+			dev::eth::Block block = c->block(c->blockChain().numberHash(blockNumber));
+			if (index < block.pending().size())
 			{
-				Executive e(state, c->blockChain(), 0);
-				Transaction t = state.pending()[index];
-				state = state.fromPending(index);
+				Executive e(block, c->blockChain(), 0);
```
arkpar: Executive should operate on a pending transaction state. Use block->fromPending to get it

ethvm/main.cpp
```diff
@@ -83,10 +84,11 @@ int main(int argc, char** argv)
 	Address sender = Address(69);
 	Address origin = Address(69);
 	u256 value = 0;
-	u256 gas = state.gasLimitRemaining();
+	u256 gas = 0;
```
arkpar: use default gas limit maybe?

arkpar:
Some of the recent commits are not merged:
https://github.com/ethereum/cpp-ethereum/commit/2513561a33286254bbff68b203ff32113d86fee4#diff-82c992d147a7388a414c1154942ae0cd
https://github.com/ethereum/cpp-ethereum/commit/c30ae3216d5b7336c793b5f041b2f2e90744b7f4#diff-82c992d147a7388a414c1154942ae0cd
https://github.com/ethereum/cpp-ethereum/commit/f16cb2e23e309b36846e82de20e35e1dc18ef5d1#diff-82c992d147a7388a414c1154942ae0cd
Could be more.
All of the commits to State.cpp after e1d58ffafc73a995ffc5ed719794bb0694467a14 should be examined carefully to check if they need to be applied to Block.cpp

github also tells rebase is needed.

libethereum/Account.cpp
gavofyork: could perhaps be an additional parameter, but tbh it's essentially fine as is.

alethzero/MainWin.cpp
gavofyork: looks fine, however, i'll sort out any AZ issues.

libethereum/Client.cpp
gavofyork: looks fine.

ethvm/main.cpp
gavofyork: Alter `State` to `Block` and use the `state()` method where a `State` object is called for later.

eth/main.cpp
gavofyork: indeed; the other two lines should remain but with s/state/block/

CJentzsch: It actually is. The next line (919): Transaction t = block.pending()[index];
And the state object was not used in the code, so I removed it.

ethvm/main.cpp
CJentzsch: This initializer should not matter at all. Using Block().gasLimitRemaining() is not allowed, and at this point there is no Block object avaiable. This vaule is the remaining gas in a block, it makes no sense to initialize it with any value fdifferent from zero before we have started to build a blockchain.

gavofyork:
closing since PR will be done from blocks branch.

PR #2717, opened 2015-08-05
tests for whisper api

gluk256:


gluk256:
the error is not mine

test/libwhisper/shhrpc.cpp
```diff
@@ -44,7 +44,9 @@ namespace js = json_spirit;
 WebThreeDirect* web3;
 unique_ptr<WebThreeStubServer> jsonrpcServer;
 unique_ptr<WebThreeStubClient> jsonrpcClient;
-static uint16_t const web3port = 30333;
+static uint16_t const g_web3port = 30333;
```
arkpar: c_web3port per style standard

test/libwhisper/shhrpc.cpp
```diff
@@ -44,7 +44,9 @@ namespace js = json_spirit;
...
+static string const g_version("shhrpc-web3");
+static unsigned const g_ttl = 777000;
```
arkpar: c_ prefix

PR #2738, opened 2015-08-06
Ignore override warnings in Qt.

chriseth:


gavofyork:
might need a ping before it's actually green.

gavofyork:
*bump* would be nice to get it green before merge.

PR #2741, opened 2015-08-06
Add peer stretch parameter w/CLI option. 

subtly:


eth/main.cpp
```diff
@@ -704,6 +715,8 @@ int main(int argc, char** argv)
 			g_logVerbosity = atoi(argv[++i]);
 		else if ((arg == "-x" || arg == "--peers") && i + 1 < argc)
 			peers = atoi(argv[++i]);
+		else if ((arg == "-x" || arg == "--peer-stretch") && i + 1 < argc)
```
gavofyork: duplicate usage of `-x` will never get used.

eth/main.cpp
```diff
@@ -990,6 +1003,7 @@ int main(int argc, char** argv)
 
 	cout << ethCredits();
 	web3.setIdealPeerCount(peers);
+	web3.private_setPeerStretch(peerStretch);
```
gavofyork: either it's private or it's not, either way don't be using underscores.

subtly:
fixed

PR #2742, opened 2015-08-07
set 50 shannon as default gasPrice

CJentzsch:
this was done before, but those hardcoded numbers slipped through.

### DIFF
--- a/alethzero/MainWin.cpp
+++ b/alethzero/MainWin.cpp
@@ -859,8 +859,8 @@ void Main::readSettings(bool _skipGeometry)
 	{
 		p->readSettings(s);
 	});
-	static_cast<TrivialGasPricer*>(ethereum()->gasPricer().get())->setAsk(u256(s.value("askPrice", QString::fromStdString(boost::lexical_cast<std::string>(c_defaultGasPrice))).toString().toStdString()));
-	static_cast<TrivialGasPricer*>(ethereum()->gasPricer().get())->setBid(u256(s.value("bidPrice", QString::fromStdString(boost::lexical_cast<std::string>(c_defaultGasPrice))).toString().toStdString()));
+	static_cast<TrivialGasPricer*>(ethereum()->gasPricer().get())->setAsk(u256(s.value("askPrice", QString::fromStdString(toString(c_defaultGasPrice))).toString().toStdString()));
+	static_cast<TrivialGasPricer*>(ethereum()->gasPricer().get())->setBid(u256(s.value("bidPrice", QString::fromStdString(toString(c_defaultGasPrice))).toString().toStdString()));
 
 	ui->upnp->setChecked(s.value("upnp", true).toBool());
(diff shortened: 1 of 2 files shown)


alethzero/MainWin.cpp
```diff
@@ -860,8 +860,8 @@ void Main::readSettings(bool _skipGeometry)
 		p->readSettings(s);
 	});
 
-	static_cast<TrivialGasPricer*>(ethereum()->gasPricer().get())->setAsk(u256(s.value("askPrice", "500000000000").toString().toStdString()));
-	static_cast<TrivialGasPricer*>(ethereum()->gasPricer().get())->setBid(u256(s.value("bidPrice", "500000000000").toString().toStdString()));
+	static_cast<TrivialGasPricer*>(ethereum()->gasPricer().get())->setAsk(u256(s.value("askPrice", "50000000000").toString().toStdString()));
+	static_cast<TrivialGasPricer*>(ethereum()->gasPricer().get())->setBid(u256(s.value("bidPrice", "50000000000").toString().toStdString()));
```
CJentzsch: I want to use `c_defaultGasPrice` here. But he doesn't accept a u256, or a string and I cant convert a string into a QString. What is the common way of providing a u256 here?

arkpar: `QString::fromStdString(...)`

CJentzsch: so I need to convert u256 to std string and the to QString, right? No short cuts? since we are suing u256 a lot.

alethzero/MainWin.cpp
```diff
@@ -859,9 +859,8 @@ void Main::readSettings(bool _skipGeometry)
 	{
 		p->readSettings(s);
 	});
-
-	static_cast<TrivialGasPricer*>(ethereum()->gasPricer().get())->setAsk(u256(s.value("askPrice", "500000000000").toString().toStdString()));
-	static_cast<TrivialGasPricer*>(ethereum()->gasPricer().get())->setBid(u256(s.value("bidPrice", "500000000000").toString().toStdString()));
+	static_cast<TrivialGasPricer*>(ethereum()->gasPricer().get())->setAsk(u256(s.value("askPrice", QString::fromStdString(boost::lexical_cast<std::string>(c_defaultGasPrice))).toString().toStdString()));
+	static_cast<TrivialGasPricer*>(ethereum()->gasPricer().get())->setBid(u256(s.value("bidPrice", QString::fromStdString(boost::lexical_cast<std::string>(c_defaultGasPrice))).toString().toStdString()));
```
CJentzsch: the art of casting... (improvements welcome)

CJentzsch:
But even after the changes I made here, I still got 500 GWei as a default paramter in Alethzero when trying to do a transaction. Where is this number set?

alethzero/MainWin.cpp
gavofyork: `toString` would be a shorter and more codebase-consistent version of `boost::lexical_cast<std::string>`

PR #2745, opened 2015-08-07
Copy map before looping over it because it might be modified.

chriseth:


libdevcore/CommonData.h
```diff
@@ -349,6 +349,24 @@ std::vector<T> keysOf(std::unordered_map<T, U> const& _m)
 	return ret;
 }
 
+template<class T, class U>
+std::vector<U> valuesOf(std::map<T, U> const& _m)
+{
+	std::vector<U> ret(_m.size());
```
gavofyork: this is not equivalent to `reserve`ing, which is what you want.

libdevcore/CommonData.h
```diff
@@ -349,6 +349,24 @@ std::vector<T> keysOf(std::unordered_map<T, U> const& _m)
...
+	for (auto const& i: _m)
+		ret.push_back(i.second);
+	return ret;
+}
+
+template<class T, class U>
+std::vector<U> valuesOf(std::unordered_map<T, U> const& _m)
+{
+	std::vector<U> ret(_m.size());
```
gavofyork: likewise. should be

```
std::vector<U> ret;
ret.reserve(_m.size());
```

libdevcore/CommonData.h
chriseth: argh, right.

PR #2751, opened 2015-08-07
Don't trip up when can't change parent directory's permissions to the…

gavofyork:
… ideal.

Don't show accounts that we don't have the keys for in AZ.


chriseth:
We might display a warning at some point stating that the setup is not optimal and how to improve it.

CJentzsch:
looksgood

PR #2768, opened 2015-08-10
icap encoding and decoding tests

debris:


debris:
on macos 10.10.3, boost 1.58.0

boost's % operator is broken. This pr fixes our code.

PR #2779, opened 2015-08-10
Various net sync fixes

arkpar:


libethereum/BlockChainSync.cpp
```diff
@@ -833,7 +833,6 @@ void PV60Sync::abortSync()
 void PV60Sync::onPeerAborting()
 {
 	RecursiveGuard l(x_sync);
-	DEV_INVARIANT_CHECK;
```
arkpar: Explained with comment below

PR #2791, opened 2015-08-11
Properly set default Gas Price in Web3 server

### LINKED ISSUE #2767: Problem with transactions under eth wrong ts time=1970-01-05 ?
I am using latest ethereum ++ console under windows and sanding transactions is expansive

If I do not set gas limit and gas price there is no way to sand... I get ID from ethconsole but it never go in blockchain

If I buff gasprice and gaslimit my transfer only then go! and I mean buff a lot like 60 000 gas limit and gas price to 5-10x more then default

If I close eth ++ wallet and start Alethzero I can transfer my funds with no problems with default "cheap" fee at once

Also when I open ethconsole.exe I get:
'setTimeout not available in this environment'

I managed to copy/paste message from eth++ wallet screen after I tried transfer of 5eth:
web3.eth.sendTransaction({from: web3.eth.coinbase, to: '0x27e806e70622d2864e1b1f4426d72ffb0457c83a', value: web3.toWei(5, "ether")})

Paste from eth wallet screen:
{"eth.tx.received":{"remote_id":"664b099bGÇª","ts":"1970-01-05T00:48:48.650Z","t
x_hash":"0e0c1243GÇª"}}

i 01:22:52| New transaction 0e0c1243GÇª{27e806e7GÇª/0$5000000000000
000000+22000@10000000000000<-664b099bGÇª #8}

LefterisJP:
Fixes #2767

### DIFF
--- a/libweb3jsonrpc/WebThreeStubServerBase.cpp
+++ b/libweb3jsonrpc/WebThreeStubServerBase.cpp
@@ -239,19 +239,22 @@ string WebThreeStubServerBase::eth_getCode(string const& _address, string const&
 	}
 }
 
+void WebThreeStubServerBase::setTransactionDefaults(TransactionSkeleton & _t)
+{
+	if (!_t.from)
+		_t.from = m_ethAccounts->defaultTransactAccount();
+	if (_t.gasPrice == UndefinedU256)
+		_t.gasPrice = c_defaultGasPrice;
+	if (_t.gas == UndefinedU256)
+		_t.gas = min<u256>(client()->gasLimitRemaining() / 5, client()->balanceAt(_t.from) / _t.gasPrice);
+}
+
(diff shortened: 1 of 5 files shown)


libweb3jsonrpc/WebThreeStubServerBase.cpp
```diff
@@ -239,19 +239,22 @@ string WebThreeStubServerBase::eth_getCode(string const& _address, string const&
 	}
 }
 
+void WebThreeStubServerBase::setTransactionDefaults(TransactionSkeleton & _t)
+{
+	if (!_t.from)
+		_t.from = m_ethAccounts->defaultTransactAccount();
+	if (_t.gasPrice == UndefinedU256)
+		_t.gasPrice = c_defaultGasPrice;
```
gavofyork: should also de derived from args passed to eth

PR #2794, opened 2015-08-11
BLock2: state test filling issue

winsvega:


CJentzsch:
as far as I see, this change is already part of https://github.com/ethereum/cpp-ethereum/pull/2777, is there anything missing in https://github.com/ethereum/cpp-ethereum/pull/2777 what you add here?

winsvega:
seems to be ok

PR #2797, opened 2015-08-11
Plugins, OtherAccounts, OurAccounts

gavofyork:


libethereum/State.cpp
```diff
@@ -243,9 +243,9 @@ unordered_map<Address, u256> State::addresses() const
 void State::setRoot(h256 const& _r)
 {
 	m_cache.clear();
-	m_touched.clear();
+//	m_touched.clear();
```
arkpar: Intentional? Seems logical to clear it

alethzero/MainFace.cpp
```diff
@@ -54,6 +74,18 @@ void Plugin::addAction(QAction* _a)
...
+{
+	QAction* a = new QAction(_n, main());
+	QMenu* m = main()->findChild<QMenu*>(_menuName);
+	if (_sep)
+		m->addSeparator();
+	m->addAction(a);
+	return a;
+}
+
+std::vector<std::function<Plugin*(MainFace*)>>* MainFace::s_linkedPlugins = nullptr;
```
arkpar: Clear it in MainFace destructor to help memory leak detectors

arkpar: Also, I'm not sure this will work properly on all platforms.  It may happen that static initialization will happen after plugin registration, zeroing the pointer. I guess this does not happen because the compiler optimizes the zero static initialization to no-op since static memory is zeroed before any code is called anyway. Still better not rely on this and leave the variable uninitialized.

libethereum/State.cpp
gavofyork: yes - used when we want to change the root to an intermediate transaction in the block. still important to understand which addresses have changed during the block for `diff` to work.

alethzero/MainFace.cpp
gavofyork: standard states that all static literal initialisations take place before static non-literal initialisations.

PR #2806, opened 2015-08-12
misc platform fixes

subtly:


gavofyork:
didn't seem to fix much - all builds fail.

subtly:
Determined that open files is due to urandom device rather than leveldb.

PR #2809, opened 2015-08-12
required peers are not added to the node table

gluk256:


gluk256:
the errors are not mine

gluk256:
rebased

PR #2699, opened 2015-08-04
Correction of typos in the help output, and harmonisation of expressions, format and style.

imifos:

PR #2810, opened 2015-08-13
Fixing typos, uniformisation of help text style, replace PR #2699

imifos:
Replaces https://github.com/ethereum/cpp-ethereum/pull/2699

### DIFF
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -160,8 +160,8 @@ void help()
 		<< "    --structured-logging-url <URL>  Set the structured logging destination (currently only file:// supported)." << endl
 		<< endl
 		<< "Import/export modes:" << endl
-		<< "    --from <n>  Export only from block n; n may be a decimal, a '0x' pre-fixed hash, or 'latest'." << endl
-		<< "    --to <n>  Export only to block n (inclusive); n may be a decimal, a '0x' pre-fixed hash, or 'latest'." << endl
+		<< "    --from <n>  Export only from block n; n may be a decimal, a '0x' prefixed hash, or 'latest'." << endl
+		<< "    --to <n>  Export only to block n (inclusive); n may be a decimal, a '0x' prefixed hash, or 'latest'." << endl
 		<< "    --only <n>  Equivalent to --export-from n --export-to n." << endl
 		<< "    --dont-check  Prevent checking some block aspects. Faster importing, but to apply only when the data is known to be valid." << endl
 		<< endl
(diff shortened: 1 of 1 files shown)


eth/main.cpp
```diff
@@ -106,74 +108,74 @@ void help()
...
 		<< "Client structured logging:" << endl
-		<< "    --structured-logging  Enable structured logging (default output to stdout)." << endl
+		<< "    --structured-logging  Enable structured logging (default: output to stdout)." << endl
 		<< "    --structured-logging-format <format>  Set the structured logging time format." << endl
 		<< "    --structured-logging-url <URL>  Set the structured logging destination (currently only file:// supported)." << endl
 		<< endl
 		<< "Import/export modes:" << endl
-		<< "    --from <n>  Export only from block n; n may be a decimal, a '0x' prefixed hash, or 'latest'." << endl
-		<< "    --to <n>  Export only to block n (inclusive); n may be a decimal, a '0x' prefixed hash, or 'latest'." << endl
+		<< "    --from <n>  Export only from block n; n may be a decimal, a '0x' pre-fixed hash, or 'latest'." << endl
```
gavofyork: "prefixed" (no hyphen) is fine. see [definition of "to prefix" verb](http://www.thefreedictionary.com/prefix).

eth/main.cpp
```diff
@@ -492,7 +494,7 @@ int main(int argc, char** argv)
 				cerr << "-c is DEPRECATED. It will be removed for the Frontier. Use --client-name instead." << endl;
 			clientName = argv[++i];
 		}
-		else if ((arg == "-a" || arg == "--address" || arg == "--coinbase-address") && i + 1 < argc)
+		else if ((arg == "-a" || arg == "--address" || arg == "--etherbase-address" || arg == "--coinbase-address") && i + 1 < argc)
```
gavofyork: prefer `--beneficiary` over `--{ether,coin}base-address`.

gavofyork:
Could of minor alterations would be nice, but good otherwise :)

eth/main.cpp
imifos: Agree, I did that before my first 0xc0ffee :/

eth/main.cpp
imifos: Yeah! Totally great (/me hyperbole fan)! That would make it totally clear to users!

imifos:
Corrected the pre-fix typo and removed the coin- to etherbase modification.

Changing the coinbase to for example beneficiary (as discussed), would break compatibility and should be done the correct way, via issue etc.

gavofyork:
rebase needed.

PR #2817, opened 2015-08-13
Fixed another case of misplaced invariant check

arkpar:


gavofyork:
so these placements were probably by me - you might want to check the others from the same commit. note that `DEV_INVARIANT_CHECK` is a RAII mechanism - it should be placed at the beginning of a scope and it'll check to make sure the scope's code doesn't cause an alteration in invariant validity. if the placement of them at the end of a scope was intentional, perhaps make a non RAII variant of it or at least comment that it's there on purpose.

gavofyork:
error in ICAP tests on macos. @debris?

arkpar:
test should be fine after rebase

Issue #2770, opened 2015-08-10
SEC-013 Deep Recursion Triggers Segmentation Fault in VM


subtly:
Attackers can DOS any cpp-ethereum miner running on a system that limits the process stack memory segment to 8 megabytes (the default on many Linux systems).

PR #2818, opened 2015-08-13
Enable stack offloading for all OSs

chfast:
Possible fix for #2770. Controversial as usually.

### DIFF
--- a/libethereum/ExtVM.cpp
+++ b/libethereum/ExtVM.cpp
@@ -35,13 +35,20 @@ static unsigned const c_depthLimit = 1024;
 /// Upper bound of stack space needed by single CALL/CREATE execution. Set experimentally.
 static size_t const c_singleExecutionStackSize =
 #ifdef NDEBUG
-	12 * 1024;
+	10 * 1024;
 #else
-	33 * 1024;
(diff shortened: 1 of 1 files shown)


libethereum/ExtVM.cpp
```diff
@@ -35,13 +35,20 @@ static unsigned const c_depthLimit = 1024;
 /// Upper bound of stack space needed by single CALL/CREATE execution. Set experimentally.
 static size_t const c_singleExecutionStackSize =
 #ifdef NDEBUG
-	12 * 1024;
+	10 * 1024;
 #else
-	33 * 1024;
+	16 * 1024;
```
subtly: What system must they be lowered for?

chfast: They don't need to. But they can. I would like to keep it quite low to detect issues as early as possible and to keep offloading point as high as possible.

gavofyork: doesn't this needlessly reduce efficiency?

chfast: This increases efficiency. We're offloading on higher call depth.

PR #2820, opened 2015-08-13
new account plugin

debris:


alethzero/Main.ui
```diff
@@ -132,7 +132,7 @@
      <x>0</x>
      <y>0</y>
      <width>1617</width>
-     <height>24</height>
+     <height>22</height>
```
arkpar: Intentional? Menu looked fine to me

PR #2822, opened 2015-08-14
Async Transact dialog for gas determination.

gavofyork:


alethzero/Transact.cpp
```diff
@@ -64,6 +64,8 @@ Transact::Transact(MainFace* _c, QWidget* _parent):
 	setValueUnits(ui->valueUnits, ui->value, 0);
 
 	on_destination_currentTextChanged(QString());
+
+	startTimer(100);
```
arkpar: m_gasCalcTimer supposed to hold this timer id?

alethzero/Transact.cpp
```diff
@@ -356,23 +446,19 @@ GasRequirements Transact::determineGasRequirements()
...
 		{
-			lastGood = er;
-			if (haveUpperBound)
-				upperBound = mid;
-			else
-				haveUpperBound = true;
+			m_lastGood = lastGood;
+			m_lowerBound = m_startLowerBound;
+			m_upperBound = m_startUpperBound;
+			startTimer(0);
```
arkpar: Can this be called while the previous timer is still active? Also startTimer in contructor uses different interval

alethzero/Transact.cpp
gavofyork: yes - fixed in latest.

alethzero/Transact.cpp
gavofyork: yes - both points addressed in recent fix.

PR #2829, opened 2015-08-14
Removed duplicate random devices

arkpar:
eth keeps ~130 handles to /dev/urandom opened. One for each cpp file that uses vector_ref.h


libdevcore/Common.cpp
```diff
@@ -33,6 +33,9 @@ char const* Version = ETH_PROJECT_VERSION;
 
 const u256 UndefinedU256 = ~(u256)0;
 
+unsigned char s_cleanseCounter = 0;
+boost::random_device s_vectorRefEngine;
```
arkpar: better place for these?

gavofyork:
So `vector_ref` is meant to be header-only, hence the `static` rather than `extern`. We'll need to check that all usages of it link libdevcore and introduce `vector_ref.cpp` to merge this PR safely.

arkpar:
vector_ref::randomize is not actually used anywhere except for SecureFixedHash::random. Refactored code for SecureFixedHash::random to use FixedHash::random and removed vector_ref::randomize

libdevcore/FixedHash.h
```diff
@@ -151,16 +151,14 @@ class FixedHash
 
 	/// @returns a randomly-valued hash
 	template <class Engine>
-	static FixedHash random(Engine& _eng)
+	void randomize(Engine& _eng)
```
chfast: Doc comment is invalid.

gavofyork:
no builds?

PR #2837, opened 2015-08-15
Fix the build for gcc 5.2.0

LefterisJP:
s_notePlugin is unused and gcc 5.2.0 complains about it


gavofyork:
er. how about submitting a fix that actually addresses the issue?

LefterisJP:
Wherever this macro was used `s_notePlugin` was treated as unused and generated a warning. I am not familiar with this part of the code but grepping for it in the file showed nothing

PR #2844, opened 2015-08-16
ImportKey plugin with much better interface

gavofyork:


arkpar:
Tests failing on linux, otherwise looksgood

PR #2848, opened 2015-08-17
Eth submitWork() rate is now hex string

LefterisJP:
Is based on https://github.com/ethereum/cpp-ethereum/pull/2847.

As per the [JSON RPC API doc](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_submithashrate) submitHashrate should accept hexadecimal strings.

### DIFF
--- a/ethminer/MinerAux.h
+++ b/ethminer/MinerAux.h
@@ -37,6 +37,7 @@
 #include <libdevcore/StructuredLogger.h>
 #include <libethcore/Exceptions.h>
 #include <libdevcore/SHA3.h>
+#include <libdevcore/CommonJS.h>
 #include <libethcore/EthashAux.h>
 #include <libethcore/EthashGPUMiner.h>
 #include <libethcore/EthashCPUMiner.h>
@@ -515,7 +516,7 @@ private:
 					auto rate = mp.rate();
 					try
 					{
-						rpc.eth_submitHashrate(toHex((u256)rate, HexPrefix::Add), "0x" + id.hex());
+						rpc.eth_submitHashrate(toJS((u256)rate), "0x" + id.hex());
 					}
 					catch (jsonrpc::JsonRpcException const& _e)
 					{
--- a/ethminer/MinerAux.h
+++ b/ethminer/MinerAux.h
@@ -515,7 +515,7 @@ private:
 					auto rate = mp.rate();
 					try
 					{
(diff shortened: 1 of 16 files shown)


libweb3jsonrpc/WebThreeStubServerBase.cpp
```diff
@@ -780,9 +780,9 @@ bool WebThreeStubServerBase::eth_submitWork(string const& _nonce, string const&,
 	}
 }
 
-bool WebThreeStubServerBase::eth_submitHashrate(int _hashes, string const& _id)
+bool WebThreeStubServerBase::eth_submitHashrate(std::string const& _hashes, string const& _id)
```
LianaHus: you don't need std:: here

akstunt:
Please lets us know when this all committed. BTW not just dev branches were effected.

libweb3jsonrpc/WebThreeStubServerBase.cpp
```diff
@@ -780,9 +780,9 @@ bool WebThreeStubServerBase::eth_submitWork(string const& _nonce, string const&,
 	}
 }
 
-bool WebThreeStubServerBase::eth_submitHashrate(int _hashes, string const& _id)
+bool WebThreeStubServerBase::eth_submitHashrate(string const& _hashes, string const& _id)
 {
-	client()->submitExternalHashrate(_hashes, jsToFixed<32>(_id));
+	client()->submitExternalHashrate(stoi(_hashes, nullptr, 0), jsToFixed<32>(_id));
```
gavofyork: should use `jsToInt`.

LefterisJP:
Rebased on develop

LefterisJP:
`jsToInt()` seems to be doing exactly what [std::stoi()](http://en.cppreference.com/w/cpp/string/basic_string/stol) with base of 0 does. Maybe it would be better to simplify it?

ethminer/MinerAux.h
```diff
@@ -515,7 +515,7 @@ class MinerCLI
 					auto rate = mp.rate();
 					try
 					{
-						rpc.eth_submitHashrate((int)rate, "0x" + id.hex());
+						rpc.eth_submitHashrate(toHex((u256)rate, HexPrefix::Add), "0x" + id.hex());
```
gavofyork: `toJS`?

PR #2849, opened 2015-08-17
SEC-019 avoid overflow in vector_ref::cropped

CJentzsch:
fixes https://github.com/ethereum/cpp-ethereum/issues/2833


CJentzsch:
For some reasons one of the rlptests ends up in an endless loop because of this change...

PR #2853, opened 2015-08-17
FindEth.cmake

debris:


gavofyork:
should be in cpp-ethereum-cmake(, too)

debris:
it is:
https://github.com/ethereum/cpp-ethereum-cmake/pull/2

PR #2862, opened 2015-08-18
Tests for CommonJS

LefterisJP:
Also simplified some functions and removed some unused ones


LefterisJP:
Got some Windows failures. Will need to address them.

libdevcore/CommonJS.h
```diff
@@ -111,9 +106,21 @@ template <unsigned N> boost::multiprecision::number<boost::multiprecision::cpp_i
 
 inline u256 jsToU256(std::string const& _s) { return jsToInt<32>(_s); }
 
+/// Convert a string representation of a number to an int
+/// String can be a normal decimal number, or a hex prefixed by 0x or 0X, or an octal if prefixed by 0
+/// Returns 0 in case of failure
 inline int jsToInt(std::string const& _s)
 {
-	return std::stoi(_s, nullptr, 0);
+	int ret;
```
gavofyork: `int ret = 0;`

libdevcore/CommonJS.h
```diff
@@ -111,9 +106,21 @@ template <unsigned N> boost::multiprecision::number<boost::multiprecision::cpp_i
...
+	try
```
gavofyork: `DEV_IGNORE_EXCEPTIONS(...);`

gavofyork: `DEV_IGNORE_EXCEPTIONS(ret = std::stoi(_s, nullptr, 0));` rather than the 8 loc here.

libethcore/CommonJS.cpp
```diff
@@ -50,8 +49,8 @@ std::string prettyU256(u256 _n, bool _abridged)
 		else
 			s << _n << "(0x" << n << ")";
 	}
-	else if ((raw = fromRaw((h256)_n, &inc)).size())
-		return "\"" + raw + "\"" + (inc ? " + " + std::to_string(inc) : "");
+	else if ((raw = fromRaw((h256)_n)).size())
```
gavofyork: `!`...`.empty()` preferred over ...`.size()`

gavofyork: `!`...`.empty()` preferred over ...`.size()`

libdevcore/CommonJS.h
LefterisJP: there was no need to initialize here since it was initialized in both legs of the try/catch. But with the macro it will be needed.

LefterisJP:
Linux failure does not seem to be related

PR #2869, opened 2015-08-19
Properly autogenerate jsonrpcstub files

LefterisJP:
All files that are supposed to be auto generated by `jsonrpcstub` should now properly be generated.

Also we assume that the rescriction for lowercase names no longer applies in recent versions of `jsonrpcstub` and as such we stick to CamelCase.


arkpar:
Custom commands should have OUTPUTS and DEPENDS for dependencies tracking

libethereum/Sentinel.h
```diff
@@ -2,30 +2,26 @@
...
  */
 
-#ifndef JSONRPC_CPP_STUB_DEV_ETH_SENTINEL_H_
-#define JSONRPC_CPP_STUB_DEV_ETH_SENTINEL_H_
+#ifndef JSONRPC_CPP_STUB_SENTINEL_H_
+#define JSONRPC_CPP_STUB_SENTINEL_H_
 
 #include <jsonrpccpp/client.h>
 
-namespace dev {
```
gavofyork: why moved out of its namespace?

ethminer/CMakeLists.txt
```diff
@@ -28,6 +28,32 @@ if (JSONRPC)
 	endif()
 endif()
 
+if (ETH_JSON_RPC_STUB)
```
gavofyork: rather than copy'n'pasting all this code all over the cmake codebase, can't we just have a simple to use function?

libethereum/Sentinel.h
LefterisJP: This is an autogenerated file

ethminer/CMakeLists.txt
LefterisJP: This is a cmake function. It's a custom command. Granted it has many parameters, but this is how jsonrpcstub works in our cmake files.

gavofyork: it's 11 lines. it should be 1.

libethereum/Sentinel.h
gavofyork: try passing in a scoped class name e.g. `dev::eth::Sentinel`.

PR #2873, opened 2015-08-20
bugfixing in tests

gluk256:


chfast:
What is so important about it?

gluk256:
Netwoking tests with JS console don't work at all.

libp2p/Network.cpp
```diff
@@ -208,19 +208,17 @@ bi::tcp::endpoint Network::traverseNAT(std::set<bi::address> const& _ifAddresses
...
 	try
 	{
-		givenPort = stoi(split.at(1));
-	}
-	catch (...)
-	{
-		clog(NetWarn) << "Error resolving host address..." << LogTag::Url << _addr << ". Could not find a port after ':'";
-		return bi::tcp::endpoint();
+		if (split.size() > 1)
+			port = static_cast<unsigned>(stoi(split.at(1)));
```
chfast: I don't know how important it is, but negative port number are valid here. E.g. "1.1.1.1:-13".
You can use `std::stoul` to parse port as an unsigned value. Checking for too high values would be good too.

chfast: This is a good candidate for a unit test.

gavofyork:
looks fine. but what bugs is it supposed to fix?

test/libwhisper/shhrpc.cpp
```diff
@@ -350,8 +350,15 @@ BOOST_AUTO_TEST_CASE(server)
 	string sess2("session number two");
 	jsonrpcServer->addSession(sess2, permissions);
 	
-	b = jsonrpcServer->admin_web3_setVerbosity(5, sess1);
+	int newVerbosity = 10;
+	int oldVerbosity = g_logVerbosity;
```
chfast: That checks nothing if g_logVerbosity is 10.

subtly:
The bug is that passing "127.0.0.1" causes the method to return an empty endpoint. Thus, when trying to connect to "poc-9.ethdev.com" or "127.0.0.1", it will return "0.0.0.0".

libdevcore/Log.h
```diff
@@ -85,7 +85,7 @@ template <class Channel> bool isChannelVisible() { return isChannelVisible(&type
 /// Not thread-safe, use with caution!
 struct VerbosityHolder
 {
-	VerbosityHolder(int _temporaryValue): oldLogVerbosity(g_logVerbosity) { g_logVerbosity = _temporaryValue; }
+	VerbosityHolder(int _temporaryValue, bool _force = false): oldLogVerbosity(g_logVerbosity) { if (g_logVerbosity >= 0 || _force) g_logVerbosity = _temporaryValue; }
```
chfast: Do you plan to use the _force param? If not, do not add a feature for uncertain future needs.

gluk256: I need it for local tests.

Issue #2876, opened 2015-08-21
llvm 3.8 build problem


jpritikin:
I installed llvm from http://llvm.org/apt/trusty/ following
https://github.com/ethereum/cpp-ethereum/wiki/Building-on-Ubuntu

I checked out the master branch 9d118e62eb7 and tried to build,

````
cmake .. -DBUNDLE=minimal
make
````

It got stuck at,

````
[  0%] Built target BuildInfo.h
Linking CXX shared library libevmjit.so
/usr/bin/ld.gold: error: cannot find -lLLVMipa
collect2: error: ld returned 1 exit status
````

I can't find hte LLVMipa library. I see it in earlier versions of llvm, but not 3.8.

Please advise.

CJentzsch:
We are looking at it. As a quick workaround for now you can disable evmjit which uses the llvm lib by doing `cmake .. -DBUNDLE=minimal -DEVMJIT=0`

CJentzsch:
fixed by #2879 and #2878

PR #2878, opened 2015-08-21
fix evmjit build for llvm3.8 - PR for master branch

CJentzsch:
@chfast , please have a look.


CJentzsch:
fixes https://github.com/ethereum/cpp-ethereum/issues/2876

chfast:
I've checked that. I think it breaks some 3.7 builds. 

Can you make this change conditional? Place unchanged `set(LLVM_LIBS ...)` under `if (LLVM3_7_CONFIG)` and the changed one under `if (LLVM3_8_CONFIG)`.

`llvm-config --libs` tool can be used for that, but I still had problems with passing the output back to cmake reliably. And 3.7 will be released today, so let's make the smallest change needed.

Thanks!

chfast:
You could've placed in a couple of lines earlier. There is the if(LLVM3_7_CONFIG) already.

CJentzsch:
Was their a reason that the this line: https://github.com/ethereum/cpp-ethereum/blob/develop/evmjit/CMakeLists.txt#L46 is before the `set` .
If not I can place it in the if/else if a couple of line above.

chfast:
No reasons for that. It can be moved up.

PR #2879, opened 2015-08-21
fix evmjit build for llvm3.8 - PR for develop

CJentzsch:
same as https://github.com/ethereum/cpp-ethereum/pull/2878


gavofyork:
should be buildbots be building?

PR #2906, opened 2015-08-23
Blockchain sync resilence (SEC-021, SEC-026, SEC-023)

### LINKED ISSUE #2896: SEC-026 Unbounded Memory Usage in Block Queue

arkpar:
Fixes #2896 #2893 #2891 (partially)

### DIFF
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -299,7 +299,8 @@ void BlockChainSync::onPeerNewBlock(std::shared_ptr<EthereumPeer> _peer, RLP con
 		case ImportResult::FutureTimeUnknown:
 		case ImportResult::UnknownParent:
 		{
-			if (++_peer->m_unknownNewBlocks > c_maxPeerUknownNewBlocks)
+			_peer->m_unknownNewBlocks++;
+			if (_peer->m_unknownNewBlocks > c_maxPeerUknownNewBlocks)
 			{
 				_peer->disable("Too many uknown new blocks");
 				if (m_state == SyncState::Idle)
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -40,6 +40,7 @@ using namespace p2p;
 
 unsigned const c_chainReorgSize = 30000; /// Added to estimated hashes to account for potential chain reorganiation
 unsigned const c_hashSubchainSize = 8192; /// PV61 subchain size
+unsigned const c_maxPeerUknownNewBlocks = 1024; /// Max number of unknown new blocks peer can give us
 
 std::ostream& dev::eth::operator<<(std::ostream& _out, SyncStatus const& _sync)
 {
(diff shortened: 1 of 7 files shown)


libethereum/BlockChainSync.cpp
```diff
@@ -298,6 +299,12 @@ void BlockChainSync::onPeerNewBlock(std::shared_ptr<EthereumPeer> _peer, RLP con
 		case ImportResult::FutureTimeUnknown:
 		case ImportResult::UnknownParent:
 		{
+			if (++_peer->m_unknownNewBlocks > c_maxPeerUknownNewBlocks)
```
gavofyork: anyone else to who can't immediately tell whether `++foo->` binds first left or right? or is it just me?

libethereum/EthereumPeer.cpp
```diff
@@ -304,6 +307,11 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 			break;
 		}
 		setIdle();
+		if (itemCount > c_maxIncomingHashes)
+		{
+			disable("Too many hashes");
```
gavofyork: future PR might check simply whether the `itemCount` is at most the number we actually asked for.

libethereum/EthereumPeer.cpp
```diff
@@ -364,8 +372,15 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 	case NewBlockHashesPacket:
 	{
 		unsigned itemCount = _r.itemCount();
+
 		clog(NetMessageSummary) << "BlockHashes (" << dec << itemCount << "entries)" << (itemCount ? "" : ": NoMoreHashes");
 
+		if (itemCount > c_maxIncomingNewHashes)
```
gavofyork: there's not technically a limit on this in the protocol. we should place a limit (e.g. 32) and make other clients aware of it.

libethereum/TransactionQueue.cpp
```diff
@@ -31,6 +31,8 @@ using namespace dev::eth;
 const char* TransactionQueueChannel::name() { return EthCyan "┉┅▶"; }
 const char* TransactionQueueTraceChannel::name() { return EthCyan " ┅▶"; }
 
+const size_t c_maxVeririficationQueueSize = 8192;
```
gavofyork: one too many syllables in "verirification" :)

libethereum/TransactionQueue.cpp
```diff
@@ -370,13 +372,23 @@ void TransactionQueue::clear()
...
 {
+	bool queued = false;
 	{
 		Guard l(x_queue);
 		unsigned itemCount = _data.itemCount();
 		for (unsigned i = 0; i < itemCount; ++i)
+		{
+			if (m_unverified.size() >= c_maxVeririficationQueueSize)
+			{
+				clog(TransactionQueueChannel) << "Transaction verification queue is full. Dropping" << itemCount - i << "transactions";
```
gavofyork: this just pops later transactions, right? so still attackable by a constant stream of useless transactions? might be worth sometimes culling from the queue itself.

arkpar: For malfromed transactions the peer who sends them will eventually loose rating (and presumably banned) here: https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/EthereumHost.cpp#L386
For well-formed transactions It is more complicated. For instance an attacker could send a stream of perfectly fine transactions, with high gas price from an account with zero balance. Such transactions will only fail at execution phase.  Would it be correct to lower rating for peers that propagate them? Currently peers propagate transactions without checking if they execute first.

libethereum/EthereumPeer.cpp
arkpar: A quick change, pushed.

PR #2910, opened 2015-08-24
Drop support for cmake 2.x. Use project(VERSION) feature.

chfast:


evmjit/CMakeLists.txt
```diff
@@ -1,17 +1,9 @@
...
-else()
-	project(EVMJIT)
-	set(EVMJIT_VERSION "0.9.0.1")
-	set(EVMJIT_VERSION_MAJOR 0)
-	set(EVMJIT_VERSION_MINOR 9)
-	set(EVMJIT_VERSION_PATCH 0)
-	set(EVMJIT_VERSION_TWEAK 1)
-endif()
+cmake_policy(SET CMP0042 OLD) 	# fix MACOSX_RPATH
+cmake_policy(SET CMP0048 NEW) 	# allow VERSION argument in project()
```
debris: can we use [EthPolicy.cmake](https://github.com/ethereum/cpp-ethereum-cmake/blob/develop/EthPolicy.cmake) instead?

chfast: No. EVMJIT must not depend on any cpp-ethereum stuff.

PR #2913, opened 2015-08-24
Replace usage of time(0) by UTC equivalent

LefterisJP:
Replacing all relevant invocations of time(0) by a function that is
guaranteed to produce a UTC timestamp according to the standard and no
longer depends on the implementation.

According to [time() documentation](http://www.cplusplus.com/reference/ctime/time/):
>The value returned generally represents the number of seconds since 00:00 hours, Jan 1, 1970 UTC (i.e., the current unix timestamp). Although libraries may use a different representation of time: Portable programs should not use the value returned by this function directly, but always rely on calls to other elements of the standard library to translate them to portable types (such as localtime, gmtime or difftime).

So we have to assume that depending on the implementation `time()` may not return the UTC timestamp and as such we may get the wrong time.

Usage of `gmtime()` in our own function should amend that.

### DIFF
--- a/libdevcore/Common.h
+++ b/libdevcore/Common.h
@@ -293,10 +293,8 @@ enum class WithExisting: int
 };
 
 /// Get the current time in seconds since the epoch in UTC
-inline time_t getNowUTC()
-{
-    return mktime(gmtime((const time_t*)time(0)));
-}
+uint64_t utcTime();
+
 
 }
 
--- a/libdevcore/Common.h
+++ b/libdevcore/Common.h
@@ -292,6 +292,12 @@ enum class WithExisting: int
 	Kill
 };
 
+/// Get the current time in seconds since the epoch in UTC
+inline time_t getNowUTC()
+{
+    return mktime(gmtime((const time_t*)time(0)));
+}
+
 }
 
 namespace std
--- a/libdevcore/Common.h
+++ b/libdevcore/Common.h
@@ -295,7 +295,6 @@ enum class WithExisting: int
 /// Get the current time in seconds since the epoch in UTC
 uint64_t utcTime();
 
-
 }
 
 namespace std
--- a/libethereum/Block.cpp
+++ b/libethereum/Block.cpp
@@ -104,7 +104,7 @@ void Block::resetCurrent()
 	m_transactionSet.clear();
 	m_currentBlock = BlockInfo();
 	m_currentBlock.setCoinbaseAddress(m_beneficiary);
-	m_currentBlock.setTimestamp(max(m_previousBlock.timestamp() + 1, (u256)getNowUTC()));
+	m_currentBlock.setTimestamp(max(m_previousBlock.timestamp() + 1, (u256)utcTime()));
 	m_currentBlock.populateFromParent(m_previousBlock);
 
 	// TODO: check.
--- a/libethereum/Block.cpp
+++ b/libethereum/Block.cpp
@@ -104,7 +104,7 @@ void Block::resetCurrent()
 	m_transactionSet.clear();
 	m_currentBlock = BlockInfo();
 	m_currentBlock.setCoinbaseAddress(m_beneficiary);
-	m_currentBlock.setTimestamp(max(m_previousBlock.timestamp() + 1, (u256)time(0)));
+	m_currentBlock.setTimestamp(max(m_previousBlock.timestamp() + 1, (u256)getNowUTC()));
 	m_currentBlock.populateFromParent(m_previousBlock);
 
 	// TODO: check.
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -527,9 +527,9 @@ ImportRoute BlockChain::import(VerifiedBlockRef const& _block, OverlayDB const& 
 	}
 
 	// Check it's not crazy
-	if (_block.info.timestamp() > (u256)time(0))
+	if (_block.info.timestamp() > (u256)getNowUTC())
 	{
-		clog(BlockChainChat) << _block.info.hash() << ": Future time " << _block.info.timestamp() << " (now at " << time(0) << ")";
+		clog(BlockChainChat) << _block.info.hash() << ": Future time " << _block.info.timestamp() << " (now at " << getNowUTC() << ")";
 		// Block has a timestamp in the future. This is no good.
 		BOOST_THROW_EXCEPTION(FutureTime());
 	}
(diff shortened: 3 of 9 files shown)


libdevcore/Common.h
```diff
@@ -294,6 +294,12 @@ enum class WithExisting: int
 	Kill
 };
 
+/// Get the current time in seconds since the epoch in UTC
+inline time_t getNowUTC()
```
chfast: Can you make it returning a more familiar type like uint64_t? And move the implementation to cpp file?

LefterisJP: I'd rather leave it returning `time_t `as this is what is already returned by the time() function. Why move the implementation? It's quite a small function so I would suppose it's a prime candidate for inlining.

gavofyork: ultimately it must be converted to a `u256` for use in the header. it is probably worth doing this ASAP (i.e. here).

also, we tend not to have `get` in the names of functions and in any case "get now UTC" makes little sense when read as english. suggest `utcTime`.

chfast: By moving it to cpp you will hide the implementation details. There is also no benefits from inlining as you need to do 2 lib calls. The users will not need to include additional headers as <ctime>.

The result is usually converted to u256, but not always. @LefterisJP replaced one case where it is used as a RNG seed. And if I remember correctly, uint64_t is implicitly convertible to u256.

LefterisJP:
@gavofyork  @chfast naming is not my strong point. Renamed the function, moved implementation to source and changed the return type to `uint64_t`. As Pawel pointed `uint64_t` is implicitly converted to `u256` in many cases, while this function is also used in places where a `long unsinged int` is required so I think it's a better return value.

libethereum/Block.cpp
```diff
@@ -104,7 +104,7 @@ void Block::resetCurrent()
 	m_transactionSet.clear();
 	m_currentBlock = BlockInfo();
 	m_currentBlock.setCoinbaseAddress(m_beneficiary);
-	m_currentBlock.setTimestamp(max(m_previousBlock.timestamp() + 1, (u256)time(0)));
```
chfast: Remove unneeded casts to (u256).

libethereum/BlockChainSync.cpp
```diff
@@ -115,7 +115,7 @@ unsigned BlockChainSync::estimatedHashes() const
 {
 	BlockInfo block = host().chain().info();
 	time_t lastBlockTime = (block.hash() == host().chain().genesisHash()) ? 1428192000 : (time_t)block.timestamp();
-	time_t now = time(0);
+	time_t now = utcTime();
```
chfast: It's not time_t any more.

libethereum/BlockQueue.cpp
```diff
@@ -389,7 +389,7 @@ void BlockQueue::tick()
 
 		cblockq << "Checking past-future blocks...";
 
-		unsigned t = time(0);
+		unsigned t = utcTime();
```
chfast: Do not trunc the value. The end of 32-bit Unix time is coming: https://en.wikipedia.org/wiki/Year_2038_problem.

libwhisper/Message.h
```diff
@@ -77,7 +77,7 @@ class Envelope
 	bytes const& data() const { return m_data; }
 
 	bool matchesBloomFilter(TopicBloomFilterHash const& f) const;
-	bool isExpired() const { return m_expiry <= (unsigned)time(0); }
+	bool isExpired() const { return m_expiry <= (unsigned)utcTime(); }
```
chfast: Looks like another 32-bit issue.

LefterisJP:
As we discussed in the chat let's leave this be as it is.

libethereum/Block.cpp
LefterisJP: The cast here is needed. It won't implicitly convert to u256 for the max() function.

libethereum/BlockChain.cpp
```diff
@@ -527,9 +527,9 @@ ImportRoute BlockChain::import(VerifiedBlockRef const& _block, OverlayDB const&
 	}
 
 	// Check it's not crazy
-	if (_block.info.timestamp() > (u256)time(0))
+	if (_block.info.timestamp() > (u256)utcTime())
```
chfast: Unneeded cast.

libethereum/BlockChainSync.cpp
```diff
@@ -114,8 +114,8 @@ void BlockChainSync::onPeerStatus(std::shared_ptr<EthereumPeer> _peer)
 unsigned BlockChainSync::estimatedHashes() const
 {
 	BlockInfo block = host().chain().info();
-	time_t lastBlockTime = (block.hash() == host().chain().genesisHash()) ? 1428192000 : (time_t)block.timestamp();
-	time_t now = time(0);
+	uint64_t lastBlockTime = (block.hash() == host().chain().genesisHash()) ? 1428192000 : (time_t)block.timestamp();
```
chfast: You missed one `time_t`.

libethereum/BlockQueue.cpp
```diff
@@ -234,14 +234,14 @@ ImportResult BlockQueue::import(bytesConstRef _block, bool _isOurs)
 
 	// Check it's not in the future
 	(void)_isOurs;
-	if (bi.timestamp() > (u256)time(0)/* && !_isOurs*/)
+	if (bi.timestamp() > (u256)utcTime()/* && !_isOurs*/)
```
chfast: Unneeded cast.

libwhisper/Message.h
chfast: Cast is not needed for comparison (`<=`).

test/libethereum/blockchain.cpp
```diff
@@ -560,7 +560,7 @@ mArray importUncles(mObject const& _blObj, vector<BlockHeader>& _vBiUncles, vect
 		BlockHeader uncleBlockFromFields = constructBlock(uncleHeaderObj);
 
 		// make uncle header valid
-		uncleBlockFromFields.setTimestamp((u256)time(0));
+		uncleBlockFromFields.setTimestamp((u256)utcTime());
```
chfast: Unneeded casts.

libdevcore/Common.h
```diff
@@ -294,6 +294,10 @@ enum class WithExisting: int
 	Kill
 };
 
+/// Get the current time in seconds since the epoch in UTC
+uint64_t utcTime();
+
+
```
gavofyork: one unneeded new line.

gavofyork:
rebase?

LefterisJP:
Rebased. After rebasing since whisper and libweb3jsonrpc were moved to webthree part of the PR had to be moved to that repo:

https://github.com/ethereum/webthree/pull/17

PR #2917, opened 2015-08-25
Fixed JsonRPC dependency for ethminer 

arkpar:


LefterisJP:
Tested and it builds properly

PR #2921, opened 2015-08-26
Remove solidity and webthree directories

LefterisJP:
Now with the optional submodule design having an empty directory for either webthree or solidity is an error since Cmake will attemp to find a CMakeList.txt under it and fail.


debris:
these are not directories, but submodules. I don't know how Pawel wants to *integrate* them into the project, so they remain optional. @chfast can you explain?

LefterisJP:
Hmm I see. This is supposed to be optional. If I don't add a submodule I don't expect to even see a directory there. The directory should go away if I don't explicitly create a submodule, no?

chfast:
If you do not clone the submodules there will be no directories.

LefterisJP:
Exactly as I suspected then @chfast this must be a remainder from when we actually had solidity and webthree as directories under cpp-ethereum. Then this can be merged.

LefterisJP:
I did not clone with the `--recursive` option and I did no `git sumodule update` and the directories are still there. It seems that just adding the subdmoule would add the directory there.

chfast:
You're right. I'll fix that.

chfast:
Fixed by https://github.com/ethereum/cpp-ethereum/commit/5a64d48db8f255d12cfd1c000bb422597727e898

LefterisJP:
Thank you

PR #2927, opened 2015-08-26
Remove prefix from some admin functions

LefterisJP:
Is needed for https://github.com/ethereum/webthree/pull/14


debris:
looks good, let's merge it after ethereum/webthree#14 is rebased and merged

PR #2928, opened 2015-08-26
bugfix in p2p

gluk256:
Sometimes the tests crash in my environment (m_nodeTable is null).
Reproduction rate was approx. 20%.


subtly:
Indeed, this is a bug. I'm concerned that m_nodeTable doesn't exist within startPeerSession; would like to find out why and fix the root issue.

gavofyork:
by all means make an issue and prioritise it, but with 1.0 on its way, this should go in.

PR #2930, opened 2015-08-26
fix for #2890  :: 

subtly:


subtly:
Quickest fix was to make recommended changes.

PR #2934, opened 2015-08-27
Ignore stack protector warnings in CL header.

chfast:


LefterisJP:
Tested it and this works. It needs to be merged since https://github.com/ethereum/cpp-ethereum-cmake/pull/37 was merged. Without it we can't build due to warning generated from the stack protector due to the opencl header.

debris:
can we instead use `SYSTEM` flag when including directories? Like [here](https://github.com/ethereum/cpp-ethereum-cmake/blob/df7235ea27e9724151a2b35c4788daf37a74705f/UseEth.cmake#L83)

The change is needed here: https://github.com/ethereum/cpp-ethereum/blob/develop/libethash-cl/CMakeLists.txt#L26

```diff
- target_include_directories(${EXECUTABLE} PUBLIC ${OpenCL_INCLUDE_DIR})
+ target_include_directories(${EXECUTABLE} SYSTEM PUBLIC ${OpenCL_INCLUDE_DIR})
```

chfast:
That looks promising. Thanks. I will check that.

chfast:
Actually, we can't do that. cl.hpp is a C++ wrapper for OpenCL functions, but copied to our repo. There is not include_directories-like statement for that one.

LefterisJP:
I always wondered why don't we use the system's `cl.hpp` header. Maybe we can now switch to them and do that? They should be pesent in all systems.
@gavofyork what do you think about that?

PR #2784, opened 2015-08-10
 bugfix 995

gluk256:

PR #2935, opened 2015-08-27
probabilistic bugfix

gluk256:
Similar to my old PR: https://github.com/ethereum/cpp-ethereum/pull/2784

I can not reproduce the bug anymore, therefore I call this PR "probabilistic bugfix".
However, the following scenario is possible: m_timers.stop() is called, the corresponding variable is set immediately, but the timers are not yet stopped, because it requires to lock the mutex (it might require some time and could trigger the thread switch). Therefore, _ec variable is zero, but destructor of host object has already been called. At this point we should not continue with descovery (or whatever else), and must instead return from the corresponding function.

### DIFF
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -580,7 +580,7 @@ void NodeTable::doCheckEvictions()
 	m_timers.schedule(c_evictionCheckInterval.count(), [this](boost::system::error_code const& _ec)
 	{
 		if (_ec)
-			clog(NodeTableMessageDetail) << "Discovery timer was probably cancelled: " << _ec.value() << _ec.message();
+			clog(NodeTableMessageDetail) << "Check Evictions timer was probably cancelled: " << _ec.value() << _ec.message();
 
 		if (_ec.value() == boost::asio::error::operation_aborted || m_timers.isStopped())
 			return;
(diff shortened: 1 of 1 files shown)


libp2p/NodeTable.cpp
```diff
@@ -580,6 +580,9 @@ void NodeTable::doCheckEvictions()
 	m_timers.schedule(c_evictionCheckInterval.count(), [this](boost::system::error_code const& _ec)
 	{
 		if (_ec)
+			clog(NodeTableMessageDetail) << "Discovery timer was probably cancelled: " << _ec.value() << _ec.message();
```
subtly: Should read "Check Evictions timer" instead of Discovery.

libp2p/NodeTable.cpp
```diff
@@ -199,7 +199,7 @@ void NodeTable::doDiscover(NodeID _node, unsigned _round, shared_ptr<set<shared_
 	m_timers.schedule(c_reqTimeout.count() * 2, [this, _node, _round, _tried](boost::system::error_code const& _ec)
 	{
 		if (_ec)
-			clog(NodeTableMessageDetail) << "Discovery timer canceled: " << _ec.value() << _ec.message();
+			clog(NodeTableMessageDetail) << "Discovery timer was probably cancelled: " << _ec.value() << _ec.message();
```
subtly: canceled

gluk256: http://grammarist.com/spelling/cancel/

PR #2936, opened 2015-08-27
Fix utcTime() function

LefterisJP:
This PR is for implementing a `utcTime()` function that does not just use time(0) to return the current UTC timestamp. The reasoning behind this is that according to the [documentation](http://www.cplusplus.com/reference/ctime/time/) 

> The value returned generally represents the number of seconds since 00:00 hours, Jan 1, 1970 UTC (i.e., the current unix timestamp). Although libraries may use a different representation of time: Portable programs should not use the value returned by this function directly, but always rely on calls to other elements of the standard library to translate them to portable types (such as localtime, gmtime or difftime).

the `time(0)` function is not guaranteed to return a UTC timestamp, even though in most implementations, as we have seen, it will. Only way to do that seems to be to combine it with gmtime() as we do in this PR.

[Review on Reviewable](https://reviewable.io/reviews/ethereum/cpp-ethereum/2936)

### DIFF
--- a/libdevcore/Common.cpp
+++ b/libdevcore/Common.cpp
@@ -57,12 +57,17 @@ TimerHelper::~TimerHelper()
 		clog(TimerChannel) << m_id << chrono::duration_cast<chrono::milliseconds>(e).count() << "ms";
 }
 
+#ifdef _WIN32 // the windows version of timegm
+#define timegm _mkgmtime
+#endif
+
 uint64_t utcTime()
 {
-	// TODO: Fix if possible to not use time(0) and merge only after testing in all platforms
-	// time_t t = time(0);
-	// return mktime(gmtime(&t));
-	return time(0);
+	time_t t = time(0);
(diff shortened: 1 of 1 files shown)


LefterisJP:
This is needed because the previous implementation would actually segfault. My bad. I was under the impression that I had tested the function but in reality I was using an old build.

LefterisJP:
But there is still an issue with this. 

I get a lot of 
```
⧎ ✘  15:27:56|p2p|5ed5bd2a…|Geth/v1.0.2/linux/go1.4.2  Clock skew? Latest block is in the future
```

Working on figuring it out now.

libdevcore/Common.cpp
```diff
@@ -59,7 +59,8 @@ TimerHelper::~TimerHelper()
 
 uint64_t utcTime()
 {
-    return mktime(gmtime((const time_t*)time(0)));
```
chfast: The reason why using C-style casts is a bad idea.

LefterisJP:
As discussed offline, let's test this new implementation thoroughly in all platforms before merging.

PR #2969, opened 2015-09-10
Overlay db tests

CJentzsch:
this PR includes tests for OverlayDB class as well as a `deepkill` function. This is only a step toward state tree pruning, deleting nodes could be done better with leveldb batches.


chriseth:
Wrong repo?

CJentzsch:
yes, I need to move my PRs to libethereum

PR #3032, opened 2016-08-12
Complete the "Homecoming"

bobsummerwill:
This PR merges all the changes which we made in the following repositories back into cpp-ethereum:

- libethereum
- libweb3core
- webthree
- webthree-helpers
- webthree-umbrella

Together with some small changes to add Appveyor and TravisCI automation, plus build file changes to reflect the difference directory structure.

The resulting directory structure here is identical to that prior to the split, for ease of upstreaming of forks like the Genoil ethminer.


.travis.yml
```diff
@@ -0,0 +1,105 @@
...
+    - ccache
+    - directories: build
+install:
+    - ./scripts/install_deps.sh
+before_script:
+    - mkdir -p build
+    - cd build
+    - cmake .. -DCMAKE_BUILD_TYPE=$TRAVIS_BUILD_TYPE -DTESTS=$TESTS
+    - make -j2
+    - ../scripts/release.sh $ZIP_SUFFIX $TESTS
```
chriseth: This way, `release.sh` is run before the tests. Furthermore, `release.sh` is run even if the bulid fails.

.travis.yml
```diff
@@ -0,0 +1,105 @@
...
+script:
+    # TODO - Move these steps into tests.sh too?
+    - cd $TRAVIS_BUILD_DIR/..
+    - git clone https://github.com/ethereum/tests.git
+    - export ETHEREUM_TEST_PATH=$(pwd)/tests/
+    - cd $TRAVIS_BUILD_DIR/build
+    - ../scripts/tests.sh
```
chriseth: This file seems to be not present. In general, I would advise to only use a single script for each of the steps.

.travis.yml
bobsummerwill: I'll have a look at the ordering, but this isn't actually harmful even in that case, because `release.sh` generates the ZIP but doesn't publish it until later (after tests pass).    It only takes a few seconds, and would just result in a little redundant work and an unused ZIP on the machine.

.travis.yml
bobsummerwill: Fixed since.

scripts/build.sh
```diff
@@ -0,0 +1,35 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>
+#
+# (c) 2016 cpp-ethereum contributors.
+#------------------------------------------------------------------------------
+
```
chriseth: Please add a `set -e` here, otherwise we don't get proper error reporting and aborts.

scripts/tests.sh
```diff
@@ -0,0 +1,43 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>
+#
+# (c) 2016 cpp-ethereum contributors.
+#------------------------------------------------------------------------------
+
```
chriseth: Please also add `set -e` here.

PR #3035, opened 2016-08-16
Use EVM-C for EVMJIT

chfast:
EVMJIT part: https://github.com/ethereum/evmjit/pull/70
Moved from https://github.com/ethereum/libethereum/pull/328.


libevm/JitVM.cpp
```diff
@@ -248,47 +242,112 @@ int64_t evm_call(
 	return cost;
 }
 
+
+/// RAII wrapper for an evm instance.
+class EVM
+{
+	evm_instance* m_instance = nullptr;
```
chriseth: Members should go to the end.

libevm/JitVM.cpp
```diff
@@ -248,47 +242,112 @@ int64_t evm_call(
...
+
+public:
+	EVM(evm_query_fn _queryFn, evm_update_fn _updateFn, evm_call_fn _callFn)
+	{
+		m_instance = evm_create(_queryFn, _updateFn, _callFn);
```
chriseth: Why not `: m_instance(evm_create(...))`?

libevm/JitVM.cpp
```diff
@@ -248,47 +242,112 @@ int64_t evm_call(
...
+	}
+
+	~EVM()
+	{
+		evm_destroy(m_instance);
+	}
+
+	EVM(EVM const&) = delete;
+	EVM& operator=(EVM) = delete;
```
chriseth: Shouldn't it be `operator=(EVM const&)` or does it not matter?

libevm/JitVM.cpp
```diff
@@ -248,47 +242,112 @@ int64_t evm_call(
...
+
+	struct Result : evm_result
+	{
+		Result(evm_result const& _other)
```
chriseth: `explicit`?

libevm/JitVM.cpp
chfast: If the initialization expression is non-trivial and long as here I prefer this form to make it stand out. But does not so matter much.

libevm/JitVM.cpp
chfast: Here it does not matter. But I like implementing `operator=` with the copy&swap idiom. The implementation exploits that the param is passed by value (copied with the copy constructor) and that copy can be modified freely inside `operator=`, i.e. you can swap it with `this`.

libevm/JitVM.cpp
chriseth: Right, it does not matter much for the pointer, but still I prefer the other version because it is cleaner. As you would write
```
EVM(evm_query_fn _queryFn, evm_update_fn _updateFn, evm_call_fn _callFn):
    m_instance = evm_create(_queryFn, _updateFn, _callFn)
{
}
```

I would say it stands out in the same way.

libevm/JitVM.cpp
```diff
@@ -248,47 +242,112 @@ int64_t evm_call(
...
+
+	~EVM()
+	{
+		evm_destroy(m_instance);
+	}
+
+	EVM(EVM const&) = delete;
+	EVM& operator=(EVM) = delete;
+
+	struct Result : evm_result
```
chriseth: No space in front of `:`

libevm/JitVM.cpp
```diff
@@ -248,47 +242,112 @@ int64_t evm_call(
...
+	{
+		Result(evm_result const& _other)
+		{
+			// FIXME: Looks like iheritance from C struct is not very effective.
```
chriseth: `Result` only provides a RAII-wrapper around `evm_result`, right? In that case, I would not use inheritance.

libevm/JitVM.cpp
```diff
@@ -248,47 +242,112 @@ int64_t evm_call(
...
+			gas_left = _other.gas_left;
+			output_data = _other.output_data;
+			output_size = _other.output_size;
+			internal_memory = _other.internal_memory;
+		}
+
+		~Result()
+		{
+			// TODO: Decide when the result has to be destroyed.
+			evm_destroy_result(*this);
```
chriseth: `this` will be different from `_other` passed into the constructor, but also see the comment above.

libevm/JitVM.cpp
```diff
@@ -248,47 +242,112 @@ int64_t evm_call(
...
+		}
+	};
+
+	/// Handy wrapper for evm_execute().
+	Result execute(ExtVMFace& _ext, int64_t gas)
+	{
+		auto env = reinterpret_cast<evm_env*>(&_ext);
+		auto mode = _ext.evmSchedule().haveDelegateCall ? EVM_HOMESTEAD
+		                                                : EVM_FRONTIER;
+		return evm_execute(m_instance, env, mode, toEvmC(_ext.codeHash),
```
chriseth: Coding style does not know about alignment between multiple lines. You always have a line break directly after `(` (if you have multiple lines), have one line per list element (although some grouping into lines makes sense in cases like this one) and the closing `)` on a line of its own, at the same indentation level as the `return`.

chriseth:
Looks good apart from style and the `Result` struct.

chfast:
Can I merge it now?

bobsummerwill:
You probably need to merge from `develop` first.

I think that should then get us to `green` on the automation for the PR.

As for the code itself, I just read through Christian's feedback.    LGTM.

PR #3037, opened 2016-08-16
AppVeyor: enable EVMJIT

chfast:


bobsummerwill:
So yeah - we might well be able to have EVMJIT enabled on Appveyor, which is building significantly faster than TravisCI.

Fitting the LLVM build into the TravisCI cycle looks pretty impossible for right now.

There might be some hope with BulkBuild, though, which makes the code build much faster, and might leave "space" in those 48 mins for LLVM too.   See https://github.com/bobsummerwill/cpp-ethereum/pull/6

PR #3038, opened 2016-08-16
rpc debug calls & Rename asOf -> block

yann300:
- debug calls 
- Rename asOf -> block

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -117,22 +117,6 @@ void Client::init(p2p::Host* _extNet, std::string const& _dbPath, WithExisting _
 	startWorking();
 }
 
-Block Client::asOf(h256 const& _block) const
-{
-	try
-	{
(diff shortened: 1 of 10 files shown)


libethereum/Client.cpp
```diff
@@ -764,7 +764,7 @@ Block Client::block(h256 const& _blockHash, PopulationStatistics* o_stats) const
 	{
 		ex << errinfo_block(bc().block(_blockHash));
 		onBadBlock(ex);
-		return Block(chainParams().accountStartNonce);
+		return Block(bc());
```
chfast: Can you explain this change?

yann300: the `asOf` function (line 120) which has been replaced by `block(...)` also return `Block(bc());`.
the function you're talking about take one more parameter ( `PopulationStatistics* o_stats )` but is basically the same. so they should return the same thing in case of failure.

PR #3043, opened 2016-08-17
Added missing "set -e" in each of the new bash scripts.

bobsummerwill:
-e  Exit immediately if a command exits with a non-zero status.


chfast:
LGTM, but let's wait for CI results.

bobsummerwill:
Yep ... please just merge it when the CI completes (though do ignore failing CI if the failures are the same as the current breaks).

CC @chriseth

chfast:
There is a test failure on Linux.

Issue #3039, opened 2016-08-16
CI/Windows: tests.bat script does not check any errors


chfast:
The script https://github.com/ethereum/cpp-ethereum/blob/develop/scripts/tests.bat that is responsible to run tests on Windows does not check any return codes from the test tools. It looks that the script does not even reach the end -- https://ci.appveyor.com/project/bobsummerwill/cpp-ethereum/build/1.0.50#L1661

bobsummerwill:
The tests do reach the end and complete, but you are quite right in nothing that the error handling is missing, and we should fix that ASAP.

This is expected.   The GitHub deployment is whitelisted onto to occur from the `release` branch, so that turns into a null-op at the end there.

    Skip "GitHub" deployment as no branches matched (build branch is "develop", deploy on branches "rel

chfast:
Maybe we should switch to PowerShell on AppVeyor? It supports global flag to stop on error. http://stackoverflow.com/questions/9948517/how-to-stop-a-powershell-script-on-the-first-error

bobsummerwill:
Yes.   It would actually be better to use PowerShell than batch, which is very limited.

I've not done any PowerShell myself before.   Have you?

But yeah - we should switch.

chfast:
I have not tried either but I will.

PR #3045, opened 2016-08-17
AppVeyor CI: check if tests not fail

chfast:
Fixes testing with EVMJIT.
Resolves https://github.com/ethereum/cpp-ethereum/issues/3039.

### DIFF
--- a/scripts/tests.bat
+++ b/scripts/tests.bat
@@ -31,7 +31,7 @@ set ETHEREUM_DEPS_PATH=%4
 if "%TESTS%"=="On" (
 
     cd ..
-    git clone https://github.com/ethereum/tests.git
+    git clone --depth 1 https://github.com/ethereum/tests.git
     set ETHEREUM_TEST_PATH=%APPVEYOR_BUILD_FOLDER%\..\tests
     cd cpp-ethereum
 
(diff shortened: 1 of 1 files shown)


scripts/tests.bat
```diff
@@ -31,7 +31,7 @@ set ETHEREUM_DEPS_PATH=%4
 if "%TESTS%"=="On" (
 
     cd ..
-    git clone https://github.com/ethereum/tests.git
+    git clone --depth 1 https://github.com/ethereum/tests.git
```
bobsummerwill: Why do we need `--depth 1` here?   Are there sub-modules?   I don't see any.

bobsummerwill:
This PR could probably also benefit from a refresh from `develop`.

But yeah - looks like there are probably real unit-test failures here.    I wonder if they are specific to `evmjit`?    Could you maybe try seeing if they a replicable against the Interpreter too?    Either by tweaking this PR for a cycle or two, or by pulling just the error-code handling stuff out into its own PR?   Thanks!

scripts/tests.bat
chfast: This is about history depth. It just downloads more/less latest commit. Nothing but optimization.

chfast:
It is not about evmjit -- all tests within `testeth` pass (that requires evmjit.dll). What fails is `testweb3core` and `testweb3`.

chfast:
@bobsummerwill Can we merge it?

bobsummerwill:
We need to do something about the failing tests before enabling this.
Or all the Appveyor builds will break.

The fact that we see the issues is good, but we would just be punching ourselves in the face to turn this on without addressing those breaks.

Maybe make a new issue for each of those two breaks, detailing them?

Do you have a PC?   Maybe you can see if you can replicate them?

bobsummerwill:
Please could you refresh your branch from `develop`, and we'll see where we are on this?
We likely need to make new issues for the Appveyor breaks, and then disable those specific tests, or fix the issues, and then we can commit this.

bobsummerwill:
"Failed"..   Hmm.

Any idea what is happening here, @chfast?

@winsvega is also close to getting Appveyor driven windows testing enabled for Solidity.

It would be great if we could get both "on" this week, so we have correct and working quality gates on both projects.

See https://github.com/ethereum/solidity/pull/948 for that.
See

chfast:
Yep, testeth crashes at startup: https://github.com/ethereum/cpp-ethereum/pull/3045/files#diff-289df8186113e294030c4fc9d278bbdeR48

So it is much worse that it was when I started it.
Someone would need to check that manually on Windows. What you cannot see on AppVeyor is the popup windows explaining the problem (like "cannot find X DLL").

bobsummerwill:
Do you have a Windows setup you could check this with, please, @winsvega?

bobsummerwill:
I believe this is redundant now.

All the changes made their way into `develop` via other PRs, with the exception of updating `cpp-dependencies`, which I would suggest we defer until https://github.com/ethereum/cpp-dependencies/pull/9 and https://github.com/ethereum/cpp-dependencies/pull/3 are resolved.

PR #3046, opened 2016-08-17
Travis CI: install LLVM and build with EVMJIT

chfast:


bobsummerwill:
I don't think this will be possible in this exact form, @chfast, because the Ubuntu build isn't really fitting in the 48mins as is.

I've got a PR which removes the breaking `testweb3` tests, which might then give us time back which might make this LLVM support workable.    The `testweb3` tests seem to run very quickly on macOS, so non-broken versions of those should be quick on Ubuntu as well.

In short - we might not be able to make this change right now.    We need to have all the tests enabled and running ahead of adding LLVM to our plate.

Also, I think we should likely build both EVMJIT and non-EVMJIT configurations throughout Appveyor and TravisCI.    Want to think about how to parameterize that?   Thanks!

chfast:
I'm just trying to build with EVMJIT. Tests does not matter for now. We don't have to merge it soon.

scripts/build.sh
```diff
@@ -31,5 +31,5 @@ TESTS=$2
 # TODO Add handling for error codes
 mkdir -p build
 cd build
-cmake .. -DCMAKE_BUILD_TYPE=$1 -DTESTS=$2
+cmake .. -DCMAKE_BUILD_TYPE=$1 -DTESTS=$2 -DEVMJIT=On -DLLVM_DIR=/usr/lib/llvm-3.9/lib/cmake/llvm
```
bobsummerwill: This is a Linux-specific path which is wrong on macOS, so we'll need some conditionals in the bash for this.

bobsummerwill:
Gotcha.

So I've committed various fixes to the develop branch which get it green again.   You'll need to merge/rebase here, and then we should be in a happier place on this PR.

bobsummerwill:
Excellent - looks good.   Please merge-at-will.

The macOS builds are really tight for time, so we cannot test EVMJIT within the automation for those **yet**, but this looks just fine for available time.   Thanks!

Great to have both Interpreter and JIT under automated test.

PR #3050, opened 2016-08-17
Major simplification of CMake files

bobsummerwill:
All implicit dependencies have been replaced with explicit ones, and eth_use() is pretty much gone.
We're just using "raw CMake" and it is way simpler to follow as a consequence.
We should do the same for solidity CMake files.


bench/CMakeLists.txt
```diff
@@ -1,14 +1,25 @@
 aux_source_directory(. SRC_LIST)
+include_directories(..)
```
chfast: This is bad. Please use `target_include_directories()`.

bench/CMakeLists.txt
```diff
@@ -1,14 +1,25 @@
...
+include_directories(../utils)
 
-set(EXECUTABLE bench)
+add_executable(bench ${SRC_LIST})
 
-add_executable(${EXECUTABLE} ${SRC_LIST})
+find_package(Boost QUIET REQUIRED COMPONENTS chrono date_time)
```
chfast: It seems that they are required only on Windows. How about moving them inside `if MSVC`?

bench/CMakeLists.txt
```diff
@@ -1,14 +1,25 @@
...
+if (UNIX AND NOT APPLE)
+	target_link_libraries(bench pthread)
+endif()
 
 if (APPLE)
-	install(TARGETS ${EXECUTABLE} DESTINATION bin)
+	install(TARGETS bench DESTINATION bin)
 else()
-	eth_install_executable(${EXECUTABLE})
+	eth_install_executable(bench)
```
chfast: Just a note: install part needs some attention in the future.

doc/dependency_graph/generate.sh
```diff
@@ -1 +1,2 @@
+python ./generate.py | tred | dot -Tpng > dependency_graph.png
```
chfast: Is it duplicated intentionally?

eth/CMakeLists.txt
```diff
@@ -1,16 +1,36 @@
 aux_source_directory(. SRC_LIST)
```
chfast: Also a note: this is also bad. CMake suggests to specify file list manually, but if we want to keep it dynamic `file(GLOB)` is better.

doc/dependency_graph/generate.sh
bobsummerwill: Yes.   One generates a PNG, the other a SVG.

bobsummerwill:
Hey @chfast,
So I just submitted another big batch of cleanups on these CMake files, and let's see where that gets us to.

Still TODO - file glob patterns.    I hadn't done anything different that what was already there, but yes, it is entirely likely that what we are currently doing is "a bad thing" :-)

chfast:
This looks nice, however:

1. I'm not sure why Windows build has failed.
2. I reviewed only 50% assuming other changes are similar.
3. This boost chrono and data_time required on Windows looks suspicious. I searched the codebase I found no includes from this libraries. Can you try building without them?

bobsummerwill:
Yeah, @chfast - I will need to iterate on it a bit more.

Those libraries ARE required on Windows.   I tried removing them a while back and that doesn't work.   I suspect there is some kind of \#pragma pulling them in.

It might be related to Boost auto-linking, and so we might need MORE of this splattered around and could then remove these?   I will keep digging.

https://github.com/ethereum/cpp-ethereum/blob/develop/test/libweb3core/test/CMakeLists.txt#L26

chfast:
I'm happy about the changes so you can merge it as soon as you fix building issues.

PR #3055, opened 2016-08-18
Critical Section __m512i union

mancoast:
This improves performance using 512bit SIMD instructions

### DIFF
--- a/libethash/internal.c
+++ b/libethash/internal.c
@@ -108,6 +108,9 @@ void ethash_calculate_dag_item(
 	__m128i xmm1 = ret->xmm[1];
 	__m128i xmm2 = ret->xmm[2];
 	__m128i xmm3 = ret->xmm[3];
+#elif defined(__MIC__)
+	__m512i const fnv_prime = _mm512_set1_epi32(FNV_PRIME);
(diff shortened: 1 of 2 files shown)


bobsummerwill:
CC @gcolvin.   Please review, Greg.

Looks good to me, because it's all in \#ifdefs.

bobsummerwill:
Please could you `git push --force` this, to see if you can get a success?   Thanks!

We are up against the TravisCI time-limit.   Annoyingly.

chriseth:
@bobsummerwill please do not ask people to force-push to re-run travis. You can restart individual builds directly in travis. If you force-push, also the successful ones will be restarted.

bobsummerwill:
" If you force-push, also the successful ones will be restarted."

From force pushing just a particular PR?   Really?    That wasn't my understanding.

chriseth:
@bobsummerwill if you force-push (after amending, without it does not have any effect), you create a new commit which has to be re-run.

bobsummerwill:
Right - but only that on that single branch for that single PR, right?

Which also have the side-benefit of retaining multiple sets of logs.

PR #3065, opened 2016-08-22
test output of dots for Travis in testeth

winsvega:
test Travis 10min output bug

### DIFF
--- a/test/libethereum/test/boostTest.cpp
+++ b/test/libethereum/test/boostTest.cpp
@@ -78,6 +78,16 @@ void createRandomTest()
 	}
 }
 
+static std::atomic_bool stopTravisOut;
+void travisOut()
+{
+	while (!stopTravisOut)
+	{
+		std::this_thread::sleep_for(std::chrono::seconds(60));
+		std::cout << ".";
+	}
+}
+
 //Custom Boost Unit Test Main
 int main( int argc, char* argv[] )
 {
@@ -117,5 +127,11 @@ int main( int argc, char* argv[] )
 	for (int i = 0; i < argc; i++)
 		parameters.push_back(argv[i]);
 
-	return unit_test_main(fake_init_func, argc, argv);
+	stopTravisOut = false;
+	std::future<int> ret = std::async(unit_test_main, fake_init_func, argc, argv);
+	std::thread outputThread(travisOut);
+	int result = ret.get();
+	stopTravisOut = true;
+	outputThread.join();
+	return result;
 }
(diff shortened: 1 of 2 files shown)


test/libethereum/test/libethereum/BlockChain.cpp
```diff
@@ -537,6 +537,7 @@ BOOST_AUTO_TEST_CASE(rescue)
 
 			try
 			{
+				std::chrono::seconds(10); //try wait for block verification before rescue
```
chfast: I don't think this expression does anything.

test/libethereum/test/boostTest.cpp
```diff
@@ -117,5 +127,10 @@ int main( int argc, char* argv[] )
 	for (int i = 0; i < argc; i++)
 		parameters.push_back(argv[i]);
 
-	return unit_test_main(fake_init_func, argc, argv);
+	std::future<int> ret = std::async(unit_test_main, fake_init_func, argc, argv);
+	std::thread outputThread(travisOut);
+	int result = ret.get();
+	stopTravisOut = true;
```
chfast: Access to `stopTravisOut` must be synchronized. You probably can use atomics.

test/libethereum/test/libethereum/BlockChain.cpp
winsvega: for some reason Travis found 2 blocks but not 3 in a blockchain

test/libethereum/test/boostTest.cpp
winsvega: why? the other process just read this variable

test/libethereum/test/libethereum/BlockChain.cpp
chfast: This will not make the process to wait 10s. It will just create an object representing 10 second interval and destroy it just after created.

test/libethereum/test/boostTest.cpp
chfast: Whatever you do in separated threads you must sync memory writes/reads. In your case change to stopTravisOut may be never visible to other thread (usually not the case on x86 architecture).

test/libethereum/test/libethereum/BlockChain.cpp
winsvega: forgot to add wait

bobsummerwill:
Is this good to go, @winsvega?

You mentioned wanting to test it some more, but unless there is any risk, the best workout the code can get it on TravisCI, considering that it likely FIXES things which are currently failing.

bobsummerwill:
Great - thanks!   Let's see how this goes, eh?

PR #3069, opened 2016-08-22
EVM-C v0.2 "pylimorphic"

chfast:


libevm/JitVM.cpp
```diff
@@ -284,6 +296,7 @@ class EVM
 
 	private:
 		evm_result m_result;
+        evm_release_result_fn m_release;
```
bobsummerwill: tab/spaces wrong?

bobsummerwill:
Looks good to me.

@axic?

axic:
Looks good. We need to come up with a good way to make this code reusable in cpp-ethereum. Right now I only need to change `evmjit_get_fn_table` to `hera_get_fn_table`, but perhaps in the future evmijt wants to deviate. Still, there will be a lot of shared code, such as the query/call/update bindings.

chfast:
No problem with generalizing JitVM class -- it can take pointer `<vm-name_get_interface()` as an argument.

PR #3070, opened 2016-08-24
Greg leaves the umbrella

gcolvin:
Refactoring to support configuration of optimizations, and code for same.


libevm/VM.cpp
```diff
@@ -74,14 +50,14 @@ template <class S> S modWorkaround(S const& _a, S const& _b)
 void VM::onOperation()
 {
 	if (*m_onOp)
-		(*m_onOp)(++m_nSteps, m_PC, m_inst,
+		(*m_onOp)(++m_nSteps, PC, INST,
```
winsvega: why without m_ ?

libevm/VM.cpp
```diff
@@ -134,15 +122,12 @@ bytesConstRef VM::execImpl(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
 {
 	m_io_gas = &io_gas;
 	m_ext = &_ext;
-	m_schedule = &m_ext->evmSchedule();
+   m_schedule = &m_ext->evmSchedule();
```
winsvega: tabulation

libevm/VM.cpp
```diff
@@ -117,14 +93,26 @@ void VM::updateMem()
...
-	m_newMemSize = memNeed(*m_SP, *(m_SP - 1));
+	unsigned n = (unsigned)INST - (unsigned)Instruction::LOG0;
+	m_runGas = toUint64(m_schedule->logGas + m_schedule->logTopicGas * n + u512(m_schedule->logDataGas) * *(SP - 1));
+	m_newMemSize = memNeed(*SP, *(SP - 1));
 	updateMem();
 }
 
+void VM::fetchInstruction()
+{
+		INST =  Instruction(PC < m_code_vector.size() ? m_code[PC] : 0);
```
chriseth: Indentation is wrong and double space after `=`

libevm/VMOpt.cpp
```diff
@@ -0,0 +1,154 @@
...
+	{
+		byte inst = m_code[i];
+
+		if ((byte)Instruction::PUSH1 <= inst && inst <= (byte)Instruction::PUSH32)
+		{
+			byte n = inst - (byte)Instruction::PUSH1 + 1;
+
+			// decode pushed bytes to integral value
+			u256 val = m_code[i+1];
+			for (uint64_t j = i+2, m = n; --m; ++j)
```
chriseth: Lots of potential for off-by-one's, also there should already be code somewhere that does this. Can we re-use it?

libevm/VMOpt.cpp
```diff
@@ -0,0 +1,154 @@
...
+			}			
+#endif
+
+#ifdef EVM_REPLACE_CONST_JUMP
+			// replace JUMP or JUMPI to constant location with JUMPV or JUMPVI
+			size_t ii = i + n + 1;
+			byte op = m_code[ii];
+			if (op == (byte)Instruction::JUMP)
+			{
+				if (verifyJumpDest(val))
```
chriseth: This has logarithmic complexity.

libevmcore/Instruction.cpp
```diff
@@ -82,6 +82,8 @@ const std::map<std::string, Instruction> dev::eth::c_instructions =
 	{ "SSTORE", Instruction::SSTORE },
 	{ "JUMP", Instruction::JUMP },
 	{ "JUMPI", Instruction::JUMPI },
+	{ "JUMPV", Instruction::JUMP },
```
chriseth: I think it would be better to add these to a separate table - they are not official opcodes.

libevmcore/Instruction.h
```diff
@@ -127,16 +127,16 @@ enum class Instruction: uint8_t
...
 	DUP1 = 0x80,		///< copies the highest item in the stack to the top of the stack
-	DUP2,				///< copies the second highest item in the stack to the top of the stack
-	DUP3,				///< copies the third highest item in the stack to the top of the stack
-	DUP4,				///< copies the 4th highest item in the stack to the top of the stack
-	DUP5,				///< copies the 5th highest item in the stack to the top of the stack
-	DUP6,				///< copies the 6th highest item in the stack to the top of the stack
-	DUP7,				///< copies the 7th highest item in the stack to the top of the stack
-	DUP8,				///< copies the 8th highest item in the stack to the top of the stack
-	DUP9,				///< copies the 9th highest item in the stack to the top of the stack
+	DUP2,					///< copies the second highest item in the stack to the top of the stack
```
chriseth: Indentation.

libevm/VMOpt.cpp
```diff
@@ -0,0 +1,154 @@
...
+				val = (val << 8) | m_code[j];
+
+#ifdef EVM_USE_CONSTANT_POOL			
+			// add value to constant pool and replace PUSHn with PUSHC if room
+			if (1 < n)
+			{
+				int pool_off = pool_constant(val);
+				if (0 <= pool_off && pool_off < 256)
+				{
+					m_code[i] = (byte)Instruction::PUSHC;
```
chriseth: If you change the code here to introduce new opcodes, you also have to scan the code to check that such opcodes were not present in the actual code.

test/libethereum/test/performance/rc5.sol
```diff
@@ -0,0 +1,101 @@
+// https://people.csail.mit.edu/rivest/Rivest-rc5rev.pdf
```
chriseth: Hm, wasn't RC5 patented?

libevm/VM.cpp
gcolvin: Yes.  They were locals, and will be again if possible.

libevm/VM.cpp
gcolvin: fixed

libevm/VM.cpp
gcolvin: fixed

libevm/VMOpt.cpp
gcolvin: If you mean 112-114, yes, it's tricky code, and I suspect it's wrong, thanks. (Am debugging behind the config macros).  There is similar code in VM.cpp but it operates directly on the program counter and stack pointer.

libevmcore/Instruction.cpp
gcolvin: I thought so too, but they need to be used like regular opcodes, so they need to be in the same tables.  If ever new opcodes conflict there will be plenty of time to reassign these internally.  We could make an EIP issue to reserve a range of opcodes for this purpose, but that could delay this thing by months.  Currently these are behind a config macro that is off.

libevm/VMOpt.cpp
gcolvin: May be I'm missing something, but the outer loop is over the whole code vector - `O(n)`  The inner loop for the constant pool is over at most 32 bytes.  The `verifyJumpDest()` lookup for the jump replacement is log the number of jump destinations.  So `not n log n` unless the code is nothing but jump destinations.

libevmcore/Instruction.h
gcolvin: I can't find a problem in my editor, even with tabs made visible and counting them.  I does look bad on github though.

libevm/VMOpt.cpp
gcolvin: Hmm.  We can't throw the invalid opcode exception until it is encountered at runtime.  I think I could check here that m_code[i] is legit, and if not just leave it there.  Will make that change.

test/libethereum/test/performance/rc5.sol
gcolvin: How about that.  Fair use?  It's only performance test code, not even debugged yet.

gcolvin:
Reviewers should double-check, but I think every non-style problem is inside configuration macros that are off.  @bobsummerwill suggested using macros so I could merge in-progress code without waiting so long between PR's.

gcolvin:
Wrong - the synthetic opcodes are not just style. i do think they are OK.

libevm/VM.cpp
chriseth: Please use `m_pc`, members are members.

libevm/VMOpt.cpp
chriseth: We have to protect against the worst case which is, as you say, "the code is nothing but jump destinations". You can leave that in if it is disabled by default, but then please put a comment.

libevmcore/Instruction.cpp
chriseth: What about prefixing them? Oh, now I see that it actually says `Instruction::JUMP`, is that intended?

libevmcore/Instruction.h
chriseth: `git checkout -p origin/develop Instruction.h` provides you with an interactive tool that lets you revert single chunks.

libevm/VMOpt.cpp
chriseth: Ok! Then again: please put a comment.

test/libethereum/test/performance/rc5.sol
chriseth: I'm european, for me software patents do not exist ;-)

libevm/VM.cpp
chriseth: I don't see the fix here.

libevm/VM.cpp
```diff
@@ -151,764 +136,745 @@ bytesConstRef VM::execImpl(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
...
 			updateIOGas();
 
-			*++m_SP = m_ext->envInfo().gasLimit();
-			break;
+			--SP;
+			++PC;
+		CASE_END
 
-		case Instruction::POP:
+		CASE_BEGIN(PUSHC)
```
chriseth: Shouldn't that be "macroed out"?

libevm/VM.cpp
```diff
@@ -151,764 +136,745 @@ bytesConstRef VM::execImpl(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
...
 			onOperation();
 			updateIOGas();
 
-			--m_SP;
-			break;
+			*++SP = m_pool[m_code[++PC]];
+			++PC;
```
chriseth: Are you sure about this increment?

libevm/VM.cpp
```diff
@@ -151,764 +136,745 @@ bytesConstRef VM::execImpl(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
...
-		case Instruction::DUP15:
-		case Instruction::DUP16:
+			if (*(SP - 1))
+				PC = verifyJumpDest(*SP);
+			else
+				++PC;
+			SP -= 2;
+		CASE_END
+
+		CASE_BEGIN(JUMPV)
```
chriseth: Again: macroed out? (also for JUMPVI)

libevm/VM.cpp
```diff
@@ -151,764 +136,745 @@ bytesConstRef VM::execImpl(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
...
 			onOperation();
 			updateIOGas();
 
-			auto n = (unsigned)m_inst - (unsigned)Instruction::SWAP1 + 2;
-			auto d = *m_SP;
-			*m_SP = m_stack[(1 + m_SP - m_stack) - n];
-			m_stack[(1 + m_SP - m_stack) - n] = d;
-			break;
+			auto n = (unsigned)INST - (unsigned)Instruction::SWAP1 + 2;
+			auto d = *SP;
```
chriseth: Why not use `std::swap` here?

libevm/VMCalls.cpp
```diff
@@ -0,0 +1,186 @@
...
+using namespace dev;
+using namespace dev::eth;
+
+
+
+void VM::copyDataToMemory(bytesConstRef _data, u256*& SP)
+{
+	auto offset = static_cast<size_t>(*SP--);
+	s512 bigIndex = *SP--;
+	auto index = static_cast<size_t>(bigIndex);
```
chriseth: No need for bigindex here, right?

libevm/VMConfig.h
```diff
@@ -0,0 +1,294 @@
+///////////////////////////////////////////////////////////////////////////////
```
chriseth: License header

libevm/VMOpt.cpp
```diff
@@ -0,0 +1,154 @@
...
+	m_bounce = &VM::interpretCases;	
+	m_code_vector = m_ext->code;
+	m_code = m_code_vector.data();
+	
+	initMetrics();	
+	
+	optimize();
+}
+
+int VM::pool_constant(const u256& _con)
```
chriseth: `u256 const&`

libevm/VMOpt.cpp
```diff
@@ -0,0 +1,154 @@
...
+{
+	int i = 0, n = m_pool_vector.size();
```
chriseth: Name should be `m_poolVector`, also the type should not be part of the name. Perhaps it makes sense to single out the constant pool into its own class?

libevm/VMOpt.cpp
```diff
@@ -0,0 +1,154 @@
...
+	while (i < n)
```
chriseth: I know this is not finished code, but this needs about 60000 comparisons of `u256` variables in the worst case. Does that take long? You probably convert the input variable to a `uint8_t` early on, so probably not an issue.

libevmcore/Instruction.cpp
gcolvin: Yes, they have to look like instructions to the interpreter, not just numbers.  The only place they are used outside of the interpreter is tracing and debugging via onOperation().   There is a comment in Instruction.h, and I just put them inside macros:
```
#ifdef EVM_USE_SYNTHETIC_INSTRUCTIONS
	// these are generated by the interpreter - should never be in user code
	PUSHC = 0xad,		///< push value from constant pool
	JUMPV,				///< alter the program counter - pre-verified
	JUMPVI,				///< conditionally alter the program counter - pre-verified
#endif

```

libevm/VMOpt.cpp
gcolvin: This may turn out to be the most heavily comment module in the system ;)

libevm/VM.cpp
gcolvin: Yes, doing that.

libevm/VMOpt.cpp
gcolvin: Comment in.  Jeffrey found it to be an optimization worth doing in the go interpreter.
  https://blog.ethereum.org/2016/06/02/go-ethereums-jit-evm/

libevmcore/Instruction.h
gcolvin: Problem was wrong tab stops in my editor.  There was some other misaligned stuff already there, so I tried to fix that to.

libevm/VM.cpp
gcolvin: Existing, working code I didn't consider changing.

libevm/VMCalls.cpp
gcolvin: Not any more.  I still need to review the code for every remaining place with a too-wide type, and for uint64_t values passed to functions that take a size_t.  (You changed some size_t a while back.)

libevm/VMConfig.h
gcolvin: Aha.

libevm/VMOpt.cpp
gcolvin: Thanks.  30 year old habits extinguish slowly.

libevm/VMOpt.cpp
gcolvin: Fixed the name and some more like it.  There are three vectors using this pattern, not sure that they are worth a class.
	// space for code and pointer to data
	bytes mCodeSpace;
	byte* mCode;

	// space for stack and pointer to data
	u256s mStackSpace;
	u256* mStack;
	
	// space for constant pool and pointer to data
	u256s mPoolSpace;
	u256* mPool;
```

File: libevm/VMOpt.cpp (hunk 18 above)
Comment (gcolvin) [inline libevm/VMOpt.cpp]: if you mean `if (mPoolSpace[i] == _con)` then both are u256.   But I can reduce the number of comparisons by using a map to find the unique constants, then copy it to a vector when done.  Another possibility is to wait until encountering a PUSHn to put the constant in the pool and replace the PUSHn with PUSHC.  Once it's working I can look at that.

File: libevm/VM.cpp (hunk 10 above)
Comment (gcolvin) [inline libevm/VM.cpp]: Done.

File: libevm/VM.cpp (hunk 11 above)
Comment (gcolvin) [inline libevm/VM.cpp]: I think it's wrong, thank s.  Still debugging this code, and it should be inside macros.

**Comment by gcolvin** [comment]
And thanks @chriseth and @winsvega for the thorough review!

File: libevm/VM.cpp (hunk 3 above)
Comment (gcolvin) [inline libevm/VM.cpp]: hadn't pushed yet

File: libevm/VMOpt.cpp (hunk 17 above)
Comment (gcolvin) [inline libevm/VMOpt.cpp]: Went ahead and fixed them, as Travis is just sitting there anyway.

File: libevm/VM.cpp (hunk 1 above)
Comment (gcolvin) [inline libevm/VM.cpp]: Just noticed this, sorry.  Will be a separate commit.

File: libevm/VMOpt.cpp (hunk 17 above)
Comment (gcolvin) [inline libevm/VMOpt.cpp]: Fixed them.  Travis hadn't even started.

File: libevm/VMOpt.cpp (hunk 5 above)
Comment (chriseth) [inline libevm/VMOpt.cpp]: I am not 100% sure but in that article he talks about doing jit optimizations in the background, so in their case, it does not hurt too much if these optimizations take way too long, because the code is already executed in the foreground.

Comment (gcolvin) [inline libevm/VMOpt.cpp]: I mean that it performs well enough to be worth doing.

File: libevmcore/Instruction.cpp [hunk 19]
```diff
@@ -329,13 +334,13 @@ string dev::eth::disassemble(bytes const& _mem)
...
 		if (!isValidInstruction(_instr))
-			ret << "0x" << hex << int(_instr) << " ";
+		ret << "0x" << hex << int(_instr) << " ";
 		else
 		{
-			InstructionInfo info = instructionInfo(_instr);
-			ret << info.name << " ";
-			if (info.additional)
-				ret << "0x" << hex << _data << " ";
+		InstructionInfo info = instructionInfo(_instr);
```
chriseth: Indentation looks wrong

libevmcore/Instruction.cpp
```diff
@@ -312,13 +317,13 @@ void dev::eth::eachInstruction(
...
 		if (isValidInstruction(instr))
-			additional = instructionInfo(instr).additional;
+		additional = instructionInfo(instr).additional;
 		u256 data;
 		for (size_t i = 0; i < additional; ++i)
 		{
-			data <<= 8;
-			if (++it < _mem.end())
-				data |= *it;
+		data <<= 8;
```
chriseth: Indentation looks wrong

libevm/VM.cpp
```diff
@@ -71,17 +47,24 @@ template <class S> S modWorkaround(S const& _a, S const& _b)
...
+		if (*m_onOp)
+			(*m_onOp)(++m_nSteps, m_pc, m_op,
+				m_newMemSize > m_mem.size() ? (m_newMemSize - m_mem.size()) / 32 : uint64_t(0),
+				m_runGas, *m_io_gas, this, m_ext);
+	}
+#else
+	void VM::onOperation()
+	{
+	}
+	#define onOperation() ()
```
chriseth: I think this causes a build error

gcolvin:
@chriseth @winsvega @bobsummerwill  Can we get this merged?  I need it in before the next stuff can go in.  It builds and tests clean in the default configuration, and builds in all configurations.
Christian:  I missed the indentation problems you noticed yesterday - looks like my editor smashed a few tabs out.  I've fixed it in my branch for next time.

bobsummerwill:
If we're down to whitespace, then please let's merge this tomorrow, eh?

chriseth:
@gcolvin sure, indentation and build error were the only blockers.

PR #3074, opened 2016-08-25
fix --vmtrace for vmTests

winsvega:
connects to https://github.com/ethereum/webthree-umbrella/issues/703


chfast:
You can merge it.

PR #3225, opened 2016-08-26
Fix dead links in CONTRIBUTING.md

ryanofsky:


ryanofsky:
There's also a dead link to CONTRIBUTING.md in the roadmap blog post (https://blog.ethereum.org/2016/02/12/ethereum-dev-update-c-roadmap/)

bobsummerwill:
Thanks for reporting that, @ryanofsky, but I'm actually in the middle of moving that content into http://ethdocs.org, so this file will get deleted in the process.

I will correct the URLs in the moved content though.   Thanks again!

Do come and join us on https://gitter.im/ethereum/cpp-ethereum-development and https://gitter.im/ethereum/cpp-ethereum.    If you would be interested in contributing to the project it would be great to have you :-)

bobsummerwill:
@chriseth I guess that you will need to revector the URL in https://blog.ethereum.org/2016/02/12/ethereum-dev-update-c-roadmap/?     I'll drop a new link here when the content has been moved.

bobsummerwill:
OK - new URL is http://www.ethdocs.org/en/latest/ethereum-clients/cpp-ethereum/contributing.html.

I've removed the Solidity-specific content in there, @chriseth.    Not sure if the existing Solidity docs need some of this content, or an equivalent added?

PR #3237, opened 2016-08-30
WIP - Re-enabled OS X Yosemite builds in TravisCI

bobsummerwill:
This is to expose the unresolved 'carthage' build break.
See https://github.com/ethereum/cpp-ethereum/pull/3051


bobsummerwill:
Replaced by https://github.com/ethereum/cpp-ethereum/pull/3276.

Issue #3088, opened 2016-08-25
Eliminate CryptoPP dependency throughout the C++ codebase


bobsummerwill:
_From @bobsummerwill on December 17, 2015 18:5_

On Nov 5th, @anthony-cros said ...

Hi guys, I cross-compiled eth for ARM and tried running it on a tizen phone. I'm getting a seg fault about 2 seconds after starting the program (no logs). I went ahead and re-compiled with debug on, and here was the result (duplicate lines removed):

    Starting program: /opt/usr/media/151103215114.3/bin/eth 
    warning: Unable to find libthread_db matching inferior's thread library, thread debugging will not be available.
    warning: the debug information found in "/usr/lib/debug/lib/libc-2.20-2014.11.so.debug" does not match "/lib/libc.so.6" (CRC mismatch).
    warning: (Internal error: pc 0xb6c45424 in read in psymtab, but not in symtab.)
    Program received signal SIGSEGV, Segmentation fault.
    CryptoPP::BufferedTransformation::ChannelPut2(std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned char const*, unsigned int, int, bool) (warning: (Internal error: pc 0xb6c45424 in read in psymtab, but not in symtab.)

) at cryptlib.cpp:356
356    cryptlib.cpp: No such file or directory.
    in cryptlib.cpp

The sources for CryptoPP definitely contain a ./cryptlib.cpp and the logs of compilation don't seem to show anything unusual about this file: https://gist.github.com/anthony-cros/1540ff3393b0926ccce7

Any clues?

And @subtly replied ...    @bobsummerwill cryptopp generally isn't thread-safe; long-term a lot of changes should occur beginning with the removal of cryptopp (if its not already there)

So ... my dependency graph (https://github.com/doublethinkco/webthree-umbrella-cross/) tells me that CryptoPP has not been removed. Is there any Github issue tracking this TODO, @gavofyork? If not, I will log one, and we can put the confusion to bed. Thanks!
We haven't hit that specific seg fault on Tizen in the meantime since then, because of other ABI issues which we've been working through.

Dec 17th 2015, @subtly said ...

@bobsummerwill I'm happy to help with the removal of cryptopp. This was a high priority prior to devcon. The challenge in this regard is for windows support – if it hasn't already been resolved, secp256k1 needs just the right compile flags. I believe that whisper may also need a few changes, specifically, to use the correct form of ECIES rather than cryptopp's version.

[...]

PR #3239, opened 2016-08-30
Suppressing deprecation warnings for SHA3_256 and SHA3_512 classes in CryptoPP

bobsummerwill:
These attributes were added after the 5.6.3 release.

From that header file ...

"The Crypto++ SHA-3 implementation dates back to January 2013 when NIST
selected Keccak as SHA-3. In August 2015 NIST finalized SHA-3, and it
was a modified version of the Keccak selection. Crypto++ 5.6.2 through
5.6.4 provides the pre-FIPS 202 version of SHA-3; while Crypto++ 5.7
and above provides the FIPS 202 version of SHA-3.

See also http://en.wikipedia.org/wiki/SHA-3"

This means that we will never be able to move to the CryptoPP-5.7.x
series of releases, because Ethereum requires Keccak, not the final
SHA-3 standard algorithm.   We are planning to migrate cpp-ethereum
off CryptoPP anyway, so this is unlikely to be a long-standing issue.

See https://github.com/ethereum/cpp-ethereum/issues/3088.

### DIFF
--- a/libethash/sha3_cryptopp.cpp
+++ b/libethash/sha3_cryptopp.cpp
@@ -20,6 +20,37 @@
 * @date 2015
 */
 #include <stdint.h>
+
+#if defined(__GNUC__)
+    // Do not warn about uses of functions (see Function Attributes), variables
+    // (see Variable Attributes), and types (see Type Attributes) marked as
+    // deprecated by using the deprecated attribute.
+    //
+    // Specifically we are suppressing the warnings from the deprecation
+    // attributes added to the SHA3_256 and SHA3_512 classes in CryptoPP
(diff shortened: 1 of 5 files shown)


etam:
Question: CryptoPP provides both [SHA3](https://cryptopp.com/docs/ref/class_s_h_a3.html), which "conforms to FIPS 202 version of SHA3 using F1600 with XOF d=0x06" and [Keccak](https://cryptopp.com/docs/ref/class_keccak.html), which "uses F1600 with XOF d=0x01".

Does that mean that newer version of CryptoPP could be used if all SHA3 references were replaced by Keccak?

Issue #3240, opened 2016-08-30
Find real solutions for hanging unit-tests


bobsummerwill:
We need to find real long-term solutions for our hanging unit-tests.

We have recently added "helper thread" workarounds for issues we are hitting without our unit-tests.   They make things more tolerable in the short-term and give us some breathing space, but they don't solve the root problems.    We need to solve the root problems.
1. Mining timeouts
2. STDOUT "dots"

This new code also uses `std::future<>` which is not supported on various ARM architectures:

```
[ 75%] /home/crosseth/eth/wd/cpp-ethereum/test/libethereum/test/TestHelper.cpp: In function 'void dev::test::testMiningFunc(void (*)())':
/home/crosseth/eth/wd/cpp-ethereum/test/libethereum/test/TestHelper.cpp:1101:79: error: invalid use of incomplete type 'class std::future<void>'
   auto future = std::async(std::launch::async, &std::thread::join, &threadTest);
```

Because of [this GCC bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64735)

What we should actually do is work out the root cause of the hangs or crashes which are causing these long delays / hangs.     Maybe we create a PR where we the tests run through GDB?    Or we add additional logging, or something?

winsvega:
mining timeouts 
https://github.com/ethereum/cpp-ethereum/pull/3286

chfast:
Hanging BC tests also fixed.

PR #3241, opened 2016-08-30
WIP - Remove unit-test ".." functionality

bobsummerwill:
WIP - Remove unit-test ".." functionality.

This writes a dot to STDOUT every time 10s to keep TravisCI alive.
This change is for troubleshooting. We need to actually find the root cause.

This is for https://github.com/ethereum/cpp-ethereum/issues/3240.

### DIFF
--- a/test/boostTest.cpp
+++ b/test/boostTest.cpp
@@ -79,16 +79,6 @@ void createRandomTest()
 	}
 }
 
-static std::atomic_bool stopTravisOut;
-void travisOut()
-{
-	while (!stopTravisOut)
-	{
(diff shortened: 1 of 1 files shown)


winsvega:
in soltest chriseth using self-adjustable mining timeouts. 
in testeth mining tests have limit of execution time 

it seems that mining time depends on build-VM resources which might be allocated randomly

winsvega:
removed timeouts: 
https://github.com/ethereum/cpp-ethereum/pull/3286

bobsummerwill:
OK - updated this PR so it ONLY has the dot removal now.
Again - this is for troubleshooting, not ready to commit yet.

PR #3242, opened 2016-08-30
Update EVMJIT and EVM-C interface

chfast:


bobsummerwill:
Is this one ready to go, @chfast, @axic?

chfast:
It is ready, but no reviews yet.

libethereum/ExtVM.cpp
```diff
@@ -33,7 +33,7 @@ namespace // anonymous
 static unsigned const c_depthLimit = 1024;
 
 /// Upper bound of stack space needed by single CALL/CREATE execution. Set experimentally.
-static size_t const c_singleExecutionStackSize = 6 * 1024;
+static size_t const c_singleExecutionStackSize = 13 * 1024;
```
bobsummerwill: Please could you add some comment here about where "13" has come from.    No doubt a lot of experimental effort into that change, but all of that context is lost with this commit, so anybody in the future has no idea where "13" came from, or what led to that.

Just some comments here would be very helpful.

chfast: Simply it crashes with 12. I will add a comment how to test it.

bobsummerwill: Ha :-)   Yeah, "crash" is a pretty good rationale!

chriseth: This has to be fixed. 12 is way too large. Does it crash in evmjit or in the interpreter?

chfast: It crashes in the EVMJIT in the Release build.

chriseth: Do you think you can limit the stack size of evmjit with some easy changes?

chfast: Unfortunately I don't know where it comes from, but it is not the case for Debug builds. I can git bisect, but I don't know any tool that could analyse stack space usage.

chfast:
Blocked by https://github.com/ethereum/evmjit/issues/52

chfast:
Updated. No stack space increase needed any more.

bobsummerwill:
Nice.   What was the cause of the need for additional stack stack which you were able to resolve?

chfast:
I'm putting the `RuntimaData` struct on the stack after switched to EVM-C. But I was also able to shrink it: https://github.com/ethereum/evmjit/commit/852d8db9b3b2877e11d6824b5e06a57b4e5d2bd5.

PR #3257, opened 2016-09-01
Added crude error-code handling for Windows unit-tests

bobsummerwill:
We should really using PowerShell, not batch files, but this should at least give us accurate pass/fail.

Fixes https://github.com/ethereum/cpp-ethereum/issues/3039.

### DIFF
--- a/scripts/tests.bat
+++ b/scripts/tests.bat
@@ -48,6 +48,7 @@ if "%TESTS%"=="On" (
     testeth.exe --verbosity 2
     echo Testing EVMJIT
     testeth.exe -t VMTests,StateTests --vm jit --verbosity 2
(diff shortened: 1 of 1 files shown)


scripts/tests.bat
```diff
@@ -48,13 +48,15 @@ if "%TESTS%"=="On" (
     testeth.exe --verbosity 2
     echo Testing EVMJIT
     testeth.exe -t VMTests,StateTests --vm jit --verbosity 2
+    IF errorlevel 1 GOTO ON-ERROR-CONDITION
```
chriseth: This reminds me of `ON ERROR RESUME NEXT`.

bobsummerwill: Batch sucks, but at least this gives us error-handling.
Next step would be nuke-and-pave with PowerShell.

PR #3259, opened 2016-09-02
move webthree tests into testeth

winsvega:
remove testweb3
move tests to testeth
move sources up into test folder


bobsummerwill:
Awesome - thanks!

PR #3261, opened 2016-09-02
Cleaned up tests.bat

bobsummerwill:
I am not convinced that our Appeveyor automation is working.
Maybe it got broken by the recent merger of the 3 apps?

### DIFF
--- a/scripts/tests.bat
+++ b/scripts/tests.bat
@@ -30,22 +30,24 @@ set ETHEREUM_DEPS_PATH=%4
 
 if "%TESTS%"=="On" (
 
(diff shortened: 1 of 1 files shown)


chfast:
LGTM

bobsummerwill:
OK - so that is a bit cleaner looking, but I still don't think the tests are actually running anymore.

@chfast?  @winsvega?   Do you agree?   Look through the last few runs.    No TTY anymore.

See https://ci.appveyor.com/project/ethereum/cpp-ethereum/build/1.0.80.   That is what we want to be seeing.

PR #3262, opened 2016-09-02
Enabled EVMJIT test for Ubuntu.  More comments.

bobsummerwill:
Enabled EVMJIT unit-tests on Ubuntu (we were building LLVM and EVMJIT, but not testing them).


chfast:
It looks like increased verbosity is too much for Travis (over 10k lines of log).

bobsummerwill:
Yeah, @chfast.
I've chopped that out.
The important changes were EVMJIT runs on Ubuntu and unit-test at all for macOS Sierra.

Hopefully this will pass now.

PR #3285, opened 2016-09-05
DO NOT MERGE - Re-enable TravisCI automation for El Capitan

bobsummerwill:
Unit-tests still disabled, but let's see if we can at least build code reliably within 48 mins.


bobsummerwill:
Too close to 48 mins for comfort, so I am intentionally NOT merging just yet.

chfast:
This does not make sense to me. Building on 4 versions of OSX without testing is waste of time. Can we pick the one that is able to run tests?

Other checks can be done in nightly builds...

bobsummerwill:
We **are** running the tests on OS X Mavericks already, @chfast.

The reason for running parallel Mac versions is that they are all using different Xcode versions (and do different Clang versions).    We don't have the Homebrew hooked up again yet, but these are also required to building the bottles, which differ by version too.

I think that we'll be able to re-enable the tests for all versions when BulkBuild is done.   It's just a matter of fitting into the 48mins.    Same as with LLVM.

bobsummerwill:
We don't have **any** nightly builds at the moment, by the way.    That is another issue.

We would like do those in Azure - and indeed could do a variety of the longer-running tests in Azure in parallel with our SAAS automation.

bobsummerwill:
UPDATE - As of Sep 16th, we now have 120min timeouts configured on TravisCI.

bobsummerwill:
Note - Rebase TODO

bobsummerwill:
Turned EVERYTHING ON, and let's see where we are at now for all macOS versions.

PR #3286, opened 2016-09-05
disable mining timeout in testeth

winsvega:
Disable mining timeout but keep std output of dots


bobsummerwill:
Thanks, @winsvega!

PR #3287, opened 2016-09-05
Add a llvm repository when installing dependencies on Ubuntu xenial

pirapira:
Before this commit, on an Ubuntu xenial environment, `./script/install_deps.sh` caused error messages:
```
E: Unable to locate package llvm-3.9-dev
E: Couldn't find any package by glob 'llvm-3.9-dev'
E: Couldn't find any package by regex 'llvm-3.9-dev'
```

After this commit, the error messages disappear, the script finishes successfully,
and the rest of the build goes through.

### DIFF
--- a/scripts/install_deps.sh
+++ b/scripts/install_deps.sh
@@ -420,6 +420,8 @@ case $(uname -s) in
                     xenial)
                         #xenial
                         echo "Installing cpp-ethereum dependencies on Ubuntu Xenial Xerus (16.04)."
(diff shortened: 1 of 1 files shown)


bobsummerwill:
Thanks, @pirapira.

The Mavericks automation break above was unrelated to this simple change.

bobsummerwill:
CC @chfast.

chfast:
LGTM. Merge it. I will think of better solution in general.

Issue #3256, opened 2016-09-01
Ubuntu - Unit-test failure - Mining_3_mineBlockWithUncles


bobsummerwill:
See https://travis-ci.org/ethereum/cpp-ethereum/jobs/156891219

```
Test Case "Mining_3_mineBlockWithUncles": 
./home/travis/build/ethereum/cpp-ethereum/test/libethereum/test/libethereum/BlockChain.cpp(232): error in "Mining_3_mineBlockWithUncles": Failed test with Exception: /home/travis/build/ethereum/cpp-ethereum/libethereum/Block.cpp(558): Throw in function dev::u256 dev::eth::Block::enact(const dev::eth::VerifiedBlockRef&, const dev::eth::BlockChain&)
```

winsvega:
the build by the link is green

bobsummerwill:
Yeah - I think I re-ran it :-)

I will create a PR which re-adds this test and run that PR until the issue manifests again.

bobsummerwill:
Here -> https://github.com/ethereum/cpp-ethereum/pull/3290

PR #3290, opened 2016-09-06
WIP - Re-enable Mining_3_mineBlockWithUncles.

bobsummerwill:
This test is unreliable, so is currently disabled.

This PR re-adds it, with the intention of exposing the build break.   Then we need to fix it :-)

See https://github.com/ethereum/cpp-ethereum/issues/3256


bobsummerwill:
OK - cannot cheat this into replicating, so I will just drop the PR for now.

We will just fix the issue when we fix it.

PR #3291, opened 2016-09-06
WIP - Add initial Alpine Linux support to install_deps.sh

bobsummerwill:
Needs testing, and to add jsonrpccpp.

### DIFF
--- a/scripts/install_deps.sh
+++ b/scripts/install_deps.sh
@@ -55,6 +55,19 @@ set -e
 # Check for 'uname' and abort if it is not available.
 uname -v > /dev/null 2>&1 || { echo >&2 "ERROR - cpp-ethereum requires 'uname' to identify the platform."; exit 1; }
 
+# See http://unix.stackexchange.com/questions/92199/how-can-i-reliably-get-the-operating-systems-name
+detect_linux_distro() {
+    if [ $(command -v lsb_release) ]; then
+        DISTRO=$(lsb_release -is)
+    elif [ -f /etc/os-release ]; then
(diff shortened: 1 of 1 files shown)


bobsummerwill:
Hey @rainbeam, @VoR0220,

Please could one of you try this out on an Alpine Linux setup?

I think we're just missing `json-rpc-cpp` (which will probably need to be a cut-and-paste of the build-from-source for that package which we already have in Debian and Ubuntu), but some verification that the other packages are working would be great.    Thanks!

rainbeam:
`lsb_release` won't work in Alpine. For solidity I created a more generic os detection function: 

https://github.com/ethereum/solidity/blob/develop/scripts/install_deps.sh#L51-L61

bobsummerwill:
You did, and I just forgot to pull that code over here too.   Done now!

Thanks for that :-)

rainbeam:
You forgot to change the lsb_release invocation as well:

https://github.com/ethereum/solidity/blob/develop/scripts/install_deps.sh#L132

Also, change the script to use `sh` rather than `bash`.

Regarding dependencies, Alpine mainline ('stable') doesn't have miniupnpc, leveldb or json-rpc-cpp. When building from the umbrella I had to build these from source:

https://github.com/rainbeam/eth-static/blob/master/Dockerfile

You are also missing `curl-dev` and `libmicrottpd-dev`.

bobsummerwill:
Thanks - fixed some of those.    build-from-source bits still TODO.

curl and libmicrohttpd were not missing - just not using the -dev variants.

chfast:
Done.

Issue #3234, opened 2016-08-29
ethminer/PhoneHome.h being modified by the build process


bobsummerwill:
I'm repeatedly seeing `ethminer/PhoneHome.h` being modified as a side-effect of our `cpp-ethereum` build process. Anybody else seeing this?

Line 21. `result.isInt()` gets modified to be `result.isIntegral()`

https://github.com/ethereum/cpp-ethereum/blob/develop/ethminer/PhoneHome.h#L21

Anybody got thoughts on where that might be coming from?

chfast:
PhoneHome removed.

PR #3292, opened 2016-09-07
WIP - Homebrew automation

bobsummerwill:
This should be nearly all of the changes required to re-add Homebrew automation.

The exceptions are:
1. The hook-up in `release.sh` for the version number
2. Certificate so that the homebrew-ethereum Git commit works.
3. Uploading the `.tar.gz` to `builds.ethereum.org`.   Still haven't quite worked out how the metadata for the bottle themselves hooks up.


ethminer/PhoneHome.h
```diff
@@ -18,7 +18,7 @@ class PhoneHome : public jsonrpc::Client
             p.append(param1);
             p.append(param2);
             Json::Value result = this->CallMethod("report_benchmark",p);
-            if (result.isInt())
+            if (result.isIntegral())
```
chriseth: Why did you add this change?

chriseth:
travis is failing, looks good otherwise.

ethminer/PhoneHome.h
bobsummerwill: Ah - that was completely accidental commit.

Caused by the PhoneHome.h bug ->https://github.com/ethereum/cpp-ethereum/issues/3234

I will revert that back out, and see if I can work out where that file change is coming from within the build process.

bobsummerwill: See https://github.com/ethereum/cpp-ethereum/pull/3296.   Found the root cause.

chfast:
Obsoleted.

PR #3293, opened 2016-09-07
Optimizations

chfast:


libethash/CMakeLists.txt
```diff
@@ -29,13 +29,7 @@ if (NOT CRYPTOPP_FOUND)
 	find_package(CryptoPP 5.6.2)
 endif()
 
-if (CRYPTOPP_FOUND)
```
bobsummerwill: Great!

See also https://github.com/ethereum/cpp-ethereum/issues/3088.   If that is something which you would like to look at, that would be fantastic.

libdevcore/CommonData.cpp
```diff
@@ -35,13 +35,26 @@
...
 
-bool dev::isHex(string const& _s)
+namespace
 {
-	unsigned i = (_s.size() >= 2 && _s.substr(0, 2) == "0x") ? 2 : 0;
-	for (; i < _s.size(); ++i)
-		if (fromHex(_s[i], WhenError::DontThrow) == -1)
-			return false;
-	return true;
+int _fromHex(char _i) noexcept
```
chfast: It conflicts with `fromHex(string)`. I didn't have better name.

chriseth: We could name it `fromHexNibble` / `fromHexChar` but we can also leave it like that.

bobsummerwill: Let's change the name.    The underscore will have everybody who reads the code asking what the underscore means.

Better to just have a more explicit, longer name, which describes what exactly the function is.

chfast:
Updated. Please review.

PR #3294, opened 2016-09-07
enable onOp tracing for all users

gcolvin:
VMConfig.h is using ETH_VMTRACE to control whether to enable tracing with a provided onOp() function.  This is set with cmake -DVMTRACE for testeth, but other users of tracing and single stepping don't yet know to build this way.  So enable onOp()  unconditionally until we work this out.


gcolvin:
@bobsummerwill  Merge me please.

PR #3297, opened 2016-09-08
Drop "phone home" support

chfast:
This is very old feature to submit benchmarked mining hash rates to a public server for ranking. It never worked.


bobsummerwill:
Yes - that is even better!   Thanks.

PR #3304, opened 2016-09-12
EVMJIT: experimental exp2 optimization

chfast:


chfast:
Please do not merge it.
This is to be used by @gcolvin for performance comparison.
Because this also enables one of the optimization passes in LLVM, LLVM can change div i256 to div i160, but we don't have support for that. In other words, more works is needed here.

gcolvin:
I'm about to hit the road, will not be likely to be able to use this until Wednesday or Thursday enroute, or Sunday after I get to Shanghai.  So if you can get this ready to merge by then that would be easiest for me.  If you could merge the corresponding change to VM.cpp I would be ecstatic.  If not I can probably use the numbers I have.  Will send email with more details.  @chfast

PR #3312, opened 2016-09-18
Fix arch linux script

abueide:
Pacman must be run as sudo, makepkg cannot be run as sudo, so running script as sudo doesn't work.

### DIFF
--- a/scripts/install_deps.sh
+++ b/scripts/install_deps.sh
@@ -171,7 +171,7 @@ case $(uname -s) in
             # The majority of our dependencies can be found in the
             # Arch Linux official repositories.
             # See https://wiki.archlinux.org/index.php/Official_repositories
-            pacman -Sy --noconfirm \
+            sudo pacman -Sy --noconfirm \
                 autoconf \
                 automake \
                 gcc \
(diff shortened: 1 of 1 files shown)


bobsummerwill:
Great!   Thanks for that.

Please could you share an e-mail address for you with me?    We are in the process of re-licensing our codebase from GPLv3 to Apache 2.0 and just need a one-pager from every contributor.   I'll need to send you that.   If you wish to stay pseudonymous, that is fine.   Just need a declaration that the code is from you and that you are OK with it being used as Apache 2.0.    Even just an e-mail consent would be fine for this scope of change, but the paperwork covers you for future contributions too.   Like a CLA.

- https://bobsummerwill.com/2016/07/12/ethereum-everywhere/
- https://bobsummerwill.com/2016/07/12/c-re-licensing-plan/

abueide:
@bobsummerwill my email is abueide@protonmail.com I'm not sure that adding sudo in front of 2 lines counts as an original work, but yes I did that, and yes I consent for my future contributions to be used under the apache license as well. I'm not sure exactly what I need to do so if this isn't enough please specify what to do.

LefterisJP:
I don't like having `sudo` inside a script. It would be nice to specifically ask the user to run the script with super user privileges instead. May I suggest something like:

```bash
if [[ $UID != 0 && $EUID != 0 ]]; then
    echo "This script needs to be run with super user privileges"
    exit 1
fi
```

chriseth:
Replaced by rebased https://github.com/ethereum/cpp-ethereum/pull/3396

PR #3317, opened 2016-09-27
Fix EVMJIT call cost bug

chfast:
I moved gas cost calculations for calls to VM implementation. That fixes EVMJIT integration bug and has better scalability (number of VM implementations should be lower than number of users).

However, I had to increase stack space needed for calls because some of the logic has been moved.

### DIFF
--- a/libethereum/ExtVM.cpp
+++ b/libethereum/ExtVM.cpp
@@ -33,7 +33,7 @@ namespace // anonymous
 static unsigned const c_depthLimit = 1024;
 
 /// Upper bound of stack space needed by single CALL/CREATE execution. Set experimentally.
-static size_t const c_singleExecutionStackSize = 6 * 1024;
+static size_t const c_singleExecutionStackSize = 9 * 1024;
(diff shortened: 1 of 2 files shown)


libethereum/ExtVM.cpp
```diff
@@ -33,7 +33,7 @@ namespace // anonymous
 static unsigned const c_depthLimit = 1024;
 
 /// Upper bound of stack space needed by single CALL/CREATE execution. Set experimentally.
-static size_t const c_singleExecutionStackSize = 6 * 1024;
+static size_t const c_singleExecutionStackSize = 9 * 1024;
```
bobsummerwill: This was a constant which you had raised before in a previous commit, and then found a way to reduce again later, right, @chfast?    Given that ALL the tests are broken at the moment, I'll merge this PR as-is, but we might need to iterate on this again, eh?

chriseth:
argh, sorry, did not refresh the page :-)

PR #3323, opened 2016-09-30
Attempt to address Home-brew breaks for Mavericks in TravisCI

bobsummerwill:
- No longer doing 'brew upgrade' which was breaking for OS X Mavericks
- Removed warnings for macOS Sierra (brew install should be enough?)
- Explicitly installing indirect dependencies too.


bobsummerwill:
This approach needs more work.   Aborted.

PR #3325, opened 2016-09-30
Quick fix for OS X Mavericks breaks

bobsummerwill:
Directory changes made around the Homebrew 1.0.0 release in September 2016 are disasterous for turnaround time for `brew update`, especially on OS X Mavericks, which is missing prebuilt bottles for nearly all packages.  These changes plain broke our TravisCI automation.  This is a quick-and-dirty attempt at a workaround to get us working again.  A long-term sustainable solution will likely see us replacing the bare `brew update` with something which is tightly targeted to only update the subset of packages in Homebrew which are used in the cpp-ethereum build process, either directly or indirectly.


chfast:
Obsolated.

PR #3326, opened 2016-10-03
devcon2 attacks test fillers

winsvega:
Here is the test fillers for devcon2 attacks 
the actual tests are already added to the test repository. this PR is a source of the tests.

1. Shanghailove is a copy of real chain state and transaction on which it was executed
https://github.com/ethereum/tests/blob/develop/BlockchainTests/Homestead/bcShanghaiLove.json
2. Suicide issue test call a dozen of contracts which then suicide. but in the end it goes OOG and triggers a state revert operation. so that memory usage could grow beyond limits on some client implementations.
*SuicideIssue test can not be created with current cpp implementation. so it's file in test repository contains wrong post state and wrong lastblock hash. however if a client manage to execute this tests and does not crash then it means it is fine.  Please send me the correct post state and hashes of the suicdeIssue test. (https://github.com/ethereum/tests/blob/develop/BlockchainTests/Homestead/bcSuicideIssue.json)

3. OutOfGasPrefundedContractCreation State test
Is presented me by other developer. 

4. bcExploit contains some tests of the issues that were reported to me privately or that I found in security channel.

Please send to me a complete list of all issues that happen during the last month and devcon2 attacks.


chriseth:
I think we should not add performance test to these test suites because they do not test performance but only correctness.

winsvega:
which tests of above do you consider to be performance? 
you mean shanghai and suicide? 

shanghai better be tested on leveldb
suicide is memory test.

chriseth:
If these tests are not run for regular CI, then I'm fine with them.

PR #3330, opened 2016-10-04
Add dockerfile.

chriseth:


chriseth:
Merging so we at least have a dockerhub file.

chfast:
What do you think about have 2 layer docker solution:
1. cpp-ethereum-devenv - environment capable of building cpp-ethereum, e.g. https://github.com/ethereum/evmjit/blob/develop/docker/Dockerfile
2. cpp-ethereum with built C++ node.

PR #3344, opened 2016-10-10
EIP150 version1 gasPrice change proposal

winsvega:
Implements http://github.com/ethereum/EIPs/issues/150

 - [x] changes in gas costs
 - [ ] define transition block for the main network
 - [ ] the 63/64 rule for limiting the call stack
 - [x] test for the gas cost changes
 - [ ] test for 63/64 call gas rules

Issue #3343, opened 2016-10-10
Implement EIP 150


chriseth:
Implement EIP 150: https://github.com/ethereum/EIPs/issues/150

daniel-lucio:
Could you tag a release when this is done?

chriseth:
@daniel-lucio it does not make a lot of sense to release this without a fix to the DoS issues which we are still working on. If cpp-ethereum is able to sync to HEAD again, we will make a release.

daniel-lucio:
other silly question,   let's say i got parity or geth after this incoming fork. is ethminer still capable of working?

chriseth:
Yes, ethminer is fully independent and should work with any client. Furthermore, ethminer is mostly maintained by the community now, for example in genoil's miner: https://github.com/Genoil/cpp-ethereum

PR #3345, opened 2016-10-11
EIP150 version1-1c gasPrice change proposal

winsvega:
Implements http://github.com/ethereum/EIPs/issues/150
connects to #3343 

 - [x] squash commits and rebase on top of develop
 - [x] fix scripts/test.sh and test.bat back to develop branch
 - [x] changes in gas costs
 - [x] define transition block for the main network
 - [x] the 63/64 rule for limiting the call stack
 - [x] test for the gas cost changes
 - [x] test for 63/64 call gas rules
 - [x] If SUICIDE ~~or CALLCODE~~ hits a newly created account, it triggers an additional gas cost of 25000 (similar to CALLs)

moved from https://github.com/ethereum/cpp-ethereum/pull/3344


libethashseal/Ethash.cpp
```diff
@@ -84,10 +84,13 @@ StringHashMap Ethash::jsInfo(BlockHeader const& _bi) const
 
 EVMSchedule const& Ethash::evmSchedule(EnvInfo const& _envInfo) const
 {
-	if (_envInfo.number() >= chainParams().u256Param("frontierCompatibilityModeLimit"))
-		return HomesteadSchedule;
+	if (_envInfo.number() >= chainParams().u256Param("EIP150forkBlock"))
+		return DevconSchedule;
```
chriseth: Please make a consistent change for `EIP150`, that is its proper name.

libethashseal/Ethash.cpp
```diff
@@ -84,10 +84,13 @@ StringHashMap Ethash::jsInfo(BlockHeader const& _bi) const
...
 	else
-		return FrontierSchedule;
+		if (_envInfo.number() >= chainParams().u256Param("frontierCompatibilityModeLimit"))
```
chriseth: Do we break anything if we change this to `homsteadForkBlock`? It would be a bit more consistent.

libethashseal/GenesisInfo.cpp
```diff
@@ -119,6 +119,7 @@ R"E(
 		"accountStartNonce": "0x00",
 		"frontierCompatibilityModeLimit": "0xffffffffffffffff",
 		"daoHardforkBlock": "0xfffffffffffffff",
+		"EIP150forkBlock": "0xfffffffffffffff",
```
chriseth: CamelCase would be `EIP150ForkBlock`

libevm/VMCalls.cpp
```diff
@@ -104,13 +104,21 @@ void VM::caseCreate()
...
-	
-	if (m_ext->balance(m_ext->myAddress) >= endowment && m_ext->depth < 1024)
-		*++m_sp = (u160)m_ext->create(endowment, *m_io_gas, bytesConstRef(m_mem.data() + initOff, initSize), *m_onOp);
+
+	bool depthOk = m_schedule->isEIP150() ? true : m_ext->depth < 1024;
+	if (m_ext->balance(m_ext->myAddress) >= endowment && depthOk)
+	{
+		auto createGas = *m_io_gas;
+		if (m_schedule->isEIP150())
+			createGas -= createGas / 64;
```
chriseth: There is no agreement on this yet, right?

libevm/VMCalls.cpp
```diff
@@ -104,13 +104,21 @@ void VM::caseCreate()
...
 	uint64_t initOff = (uint64_t)*m_sp--;
 	uint64_t initSize = (uint64_t)*m_sp--;
-	
-	if (m_ext->balance(m_ext->myAddress) >= endowment && m_ext->depth < 1024)
-		*++m_sp = (u160)m_ext->create(endowment, *m_io_gas, bytesConstRef(m_mem.data() + initOff, initSize), *m_onOp);
+
+	bool depthOk = m_schedule->isEIP150() ? true : m_ext->depth < 1024;
+	if (m_ext->balance(m_ext->myAddress) >= endowment && depthOk)
+	{
+		auto createGas = *m_io_gas;
```
chriseth: Please don't use auto here.

libevm/VMCalls.cpp
```diff
@@ -104,13 +104,21 @@ void VM::caseCreate()
...
+		if (m_schedule->isEIP150())
+			createGas -= createGas / 64;
+		auto gas = createGas;
```
chriseth: Please don't use auto here.

libevm/VMCalls.cpp
```diff
@@ -142,10 +150,18 @@ bool VM::caseCallSetup(CallParameters *callParams)
 		memNeed(m_stack[(1 + m_sp - m_stack) - sizesOffset], m_stack[(1 + m_sp - m_stack) - sizesOffset - 1])
 	);
 	updateMem();
+	updateIOGas();
+
+	// "Static" const already applied. Calculate call gas.
```
chriseth: `const` -> `cost`

libevm/VMCalls.cpp
```diff
@@ -142,10 +150,18 @@ bool VM::caseCallSetup(CallParameters *callParams)
...
+	auto requestedCallGas = *m_sp;
+	auto maxAllowedCallGas = *m_io_gas - *m_io_gas / 64;
```
chriseth: Again: Please don't use auto here. Note that `m_runGas` is a 64 bit variable and `m_io_gas` is a 256 bit variable. `auto` completely hides this fact and might introduce performance and correctness issues.

Also, perhaps less confusing:
```
auto maxAllowedCallGas = *m_io_gas - (m_schedule->isEIP150() ? *m_io_gas / 64 : 0);
auto callGas = std::min(requestedCallGas, maxAllowedCallGas);
```

libevmcore/EVMSchedule.h
```diff
@@ -63,11 +63,34 @@ struct EVMSchedule
...
 	unsigned txDataNonZeroGas = 68;
 	unsigned copyGas = 3;
+
+	//EIP150
+	unsigned extcodesizeGas = 20;
+	unsigned extcodecopyGas = 20;
+	unsigned balanceGas = 20;
+	unsigned suicideGas = 0;
+
+	bool isEIP150() const { return callGas == 700; }
```
chriseth: Previously, we had booleans to trigger certain behaviour, so I think it would be better to replace this by
`bool staticCallDepthLimit = true`, set the value to `false` for the `EVMScheduleEIP150` and replace all current calls to `isEIP150()` by `!staticCallDepthLimit`.

test/libethereum/BlockchainTestsFiller/TestNetwork/bcEIP150TestFiller.json
```diff
@@ -0,0 +1,335 @@
...
+	    },
+	    "1000000000000000000000000000000000000012" : {
+		"shoulsnotexist" : "1"
+	    },
+	    "1000000000000000000000000000000000000013" : {
+		"shoulsnotexist" : "1"
+	    }
+        },
+        "pre" : {
+            "d4fe7bc31cedb7bfb8a345f31e668033056b2728" : { "balance" : "1000000", "nonce" : "0", "code" : "", "storage": {} },
```
chriseth: Please explain why we need so many accounts. Furthermore, the name of the test suggests that it tests all aspects of EIP150 although I would say it only tests one particular thing (testing all aspects in one test is also not good testing practice).

libevm/VMCalls.cpp
chfast: From 1b: "CREATE only provides 63/64 of the parent gas, rounded down, to the child call." It must be like that if we want to replace static call depth limit.

chfast: From 1b: "CREATE only provides 63/64 of the parent gas, rounded down, to the child call." It must be like that if we want to replace static call depth limit.

libevm/VMCalls.cpp
chfast: I mostly agree. However, in case of `u256 createGas = *m_io_gas` if we change `m_io_gas` to `int64_t` later on we will not auto notice the implicit conversion here. `auto` guarantees not conversion.

I will switch to explicit types.

libevmcore/EVMSchedule.h
chfast: I would avoid having redundant data in the schedule struct. But you suggestion looks fine. I would be even better to have `unsinged callDepthLimit` equal 1024 or 0. Now we have this magic number in many places. But this should not go with this PR.

libethashseal/Ethash.cpp
winsvega: in eth when parsing gson config this field is called frontierCompatibilityModeLimit

test/libethereum/BlockchainTestsFiller/TestNetwork/bcEIP150TestFiller.json
winsvega: this is transition test. it includes daohardfork. 
I think to implement more scenarious in StateTests
things are in rush now. we started with simple blockchain test that is now too complicated .

libevm/VMCalls.cpp
chriseth: Specification has changed to this in the meantime.

libevmcore/EVMSchedule.h
chriseth: yep, good suggestion

libevm/VM.cpp
```diff
@@ -186,10 +186,16 @@ void VM::interpretCases()
 
 		CASE_BEGIN(SUICIDE)
 		{
+			m_runGas = toUint64(m_schedule->suicideGas);
+			Address dest = asAddress(*m_sp);
+
+			// After EIP150 hard fork charge additional cost of sending
+			// ethers to non-existing account.
+			if (!m_schedule->staticCallDepthLimit && !m_ext->exists(dest))
```
chriseth: This requires a new bool, charging `callNewAccountGas` here is different from `staticCallDepthLimit`.
To get both readability and non-redundancy, I would suggest the following:

Add two inline functions to the schedule: `staticCallDepthLimit` and `suicideChargesNewAccountGas` and replace the `staticCallDepthLimit` by a bool member that is called `eip150Mode` and let the two inline function return the respective values depending on that member.

libevm/VMCalls.cpp
```diff
@@ -142,10 +150,19 @@ bool VM::caseCallSetup(CallParameters *callParams)
 		memNeed(m_stack[(1 + m_sp - m_stack) - sizesOffset], m_stack[(1 + m_sp - m_stack) - sizesOffset - 1])
 	);
 	updateMem();
+	updateIOGas();
+
+	// "Static" costs already applied. Calculate call gas.
+	u256 requestedCallGas = *m_sp;
+	u256 maxAllowedCallGas = *m_io_gas - *m_io_gas / 64;
+	u256 callGas = m_schedule->staticCallDepthLimit
+	               ? requestedCallGas
```
chriseth: Coding style: separators should go to the end of the line.

libevmcore/EVMSchedule.h
```diff
@@ -35,6 +35,7 @@ struct EVMSchedule
 	EVMSchedule(bool _efcd, bool _hdc, unsigned const& _txCreateGas): exceptionalFailedCodeDeposit(_efcd), haveDelegateCall(_hdc), tierStepGas(std::array<unsigned, 8>{{0, 2, 3, 5, 8, 10, 20, 0}}), txCreateGas(_txCreateGas) {}
 	bool exceptionalFailedCodeDeposit = true;
 	bool haveDelegateCall = true;
+    bool staticCallDepthLimit = true;
```
chriseth: Indentation

libevmcore/EVMSchedule.h
```diff
@@ -63,11 +64,33 @@ struct EVMSchedule
...
 
 static const EVMSchedule DefaultSchedule = EVMSchedule();
 static const EVMSchedule FrontierSchedule = EVMSchedule(false, false, 21000);
 static const EVMSchedule HomesteadSchedule = EVMSchedule(true, true, 53000);
 
+static EVMSchedule EVMScheduleEIP150()
+{
+	//EIP150
+	EVMSchedule schedule = HomesteadSchedule;
+    schedule.staticCallDepthLimit = false;
```
chriseth: Indentation.

axic:
(Why is it called EIP150 and not ERC150?)

libevm/VMCalls.cpp
```diff
@@ -155,9 +155,8 @@ bool VM::caseCallSetup(CallParameters *callParams)
 	// "Static" costs already applied. Calculate call gas.
 	u256 requestedCallGas = *m_sp;
 	u256 maxAllowedCallGas = *m_io_gas - *m_io_gas / 64;
```
chriseth: As commented on an earlier version: I think it confusing to call this `maxAllowedCallGas` even if it is not an upper bound in all code paths.

test/libethereum/BlockchainTestsFiller/TestNetwork/bcEIP150TestFiller.json
```diff
@@ -29,7 +29,9 @@
                     "0x06" : "0x01",
                     "0x07" : "0x01",
                     "0x08" : "0x01",
-                    "0x09" : "0x057ee9",
+                    "0x09" : "0x4176a",
+		    "0x0a" : "0x057ee9",
```
chriseth: Indentation

chriseth:
@winsvega we do have tests for "63/64 call gas rules", right? If yes, please update description.

winsvega:
"63/64 call gas rules" 
not yet. we need more tests for that

libethashseal/Ethash.cpp
winsvega: why git does not update this comment? 
code has been updated to use EIP150Schedule

libethashseal/GenesisInfo.cpp
winsvega: done

libevm/VMCalls.cpp
winsvega: done

libevm/VMCalls.cpp
winsvega: done

libevm/VMCalls.cpp
winsvega: done

libevm/VMCalls.cpp
winsvega: done

libevmcore/EVMSchedule.h
winsvega: changed to 
```
bool staticCallDepthLimit() const { return !eip150Mode; }
bool suicideChargesNewAccountGas() const { return eip150Mode; }
```

libevm/VM.cpp
winsvega: done

libevm/VMCalls.cpp
winsvega: done

libevmcore/EVMSchedule.h
winsvega: done

libethashseal/Ethash.cpp
chfast: This is an issue with the new GitHub review interface. On this page it shows old comments even the code has been already changed. It is better to see all changes in https://github.com/ethereum/cpp-ethereum/pull/3345/files.

axic:
@winsvega yep, the repo is called EIP, but mostly those not accepted (= part of the repo and not the issues), were referred to as ERC (Ethereum Request for Comments)

chriseth:
@winsvega please set the hard fork block to 2457000 and then I think this is ready to go.
Please focus on adding tests to the general test suite (not the cpp-specific one).

winsvega:
chriseth, rebase on top of the current develop branch is needed

chriseth:
@winsvega some tests are failing, please investigate.

libethashseal/Ethash.cpp
```diff
@@ -84,10 +84,13 @@ StringHashMap Ethash::jsInfo(BlockHeader const& _bi) const
 
 EVMSchedule const& Ethash::evmSchedule(EnvInfo const& _envInfo) const
 {
-	if (_envInfo.number() >= chainParams().u256Param("frontierCompatibilityModeLimit"))
-		return HomesteadSchedule;
+	if (_envInfo.number() >= chainParams().u256Param("EIP150ForkBlock"))
+		return EIP150Schedule;
 	else
-		return FrontierSchedule;
+		if (_envInfo.number() >= chainParams().u256Param("homsteadForkBlock"))
```
chriseth: Please do not indent.

libethashseal/Ethash.cpp
```diff
@@ -163,7 +166,7 @@ void Ethash::verify(Strictness _s, BlockHeader const& _bi, BlockHeader const& _p
 
 void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase const& _t, BlockHeader const& _bi) const
 {
-	if (_ir & ImportRequirements::TransactionSignatures && _bi.number() >= chainParams().u256Param("frontierCompatibilityModeLimit"))
```
chriseth: Didn't we agree that changing this would be too bad for other users?

libethereum/Block.cpp
```diff
@@ -371,6 +371,7 @@ pair<TransactionReceipts, bool> Block::sync(BlockChain const& _bc, TransactionQu
 					else
 					{
 						clog(StateTrace) << t.sha3() << "Temporarily no gas left in current block (txs gas > block's gas limit)";
+						_tq.drop(t.sha3());
```
chriseth: I don't think we should drop the transaction here, especially not as part of the EIP150 change.

libevmcore/EVMSchedule.h
```diff
@@ -63,11 +64,34 @@ struct EVMSchedule
 	unsigned txDataZeroGas = 4;
 	unsigned txDataNonZeroGas = 68;
 	unsigned copyGas = 3;
+
+	//EIP150
```
chriseth: Please remove this comment, those are the pre-eip150 gas values.

test/libethereum/BlockChainTests.cpp
```diff
@@ -768,6 +770,110 @@ BOOST_AUTO_TEST_CASE(bcTheDaoTest)
...
+
+BOOST_AUTO_TEST_CASE(bcForkStressTestEIP150)
+{
+	dev::test::TestBlockChain::s_sealEngineNetwork = eth::Network::EIP150Test;
+	dev::test::executeTests("bcForkStressTest", "/BlockchainTests/EIP150",dev::test::getFolder(__FILE__) + "/BlockchainTestsFiller/EIP150", dev::test::doBlockchainTests);
+}
+
+BOOST_AUTO_TEST_CASE(bcSuicideIssueEIP150)
+ {
+	//dev::test::TestBlockChain::s_sealEngineNetwork = eth::Network::EIP150Test;
```
chriseth: Please either uncomment, explain why commented or remove.

test/libethereum/BlockChainTests.cpp
```diff
@@ -780,6 +886,12 @@ BOOST_AUTO_TEST_CASE(bcForkStressTestHomestead)
 	dev::test::executeTests("bcForkStressTest", "/BlockchainTests/Homestead",dev::test::getFolder(__FILE__) + "/BlockchainTestsFiller/Homestead", dev::test::doBlockchainTests);
 }
 
+BOOST_AUTO_TEST_CASE(bcSuicideIssueHomestead)
+ {
+	//dev::test::TestBlockChain::s_sealEngineNetwork = eth::Network::HomesteadTest;
```
chriseth: Please either uncomment, explain why commented or remove.

libethashseal/Ethash.cpp
winsvega: why? last time you said the opposite

libethashseal/Ethash.cpp
winsvega: should be fine.

libethereum/Block.cpp
winsvega: other clients are in consensus with that test. 
eth cpp hang otherwise

libethashseal/Ethash.cpp
chriseth: Sure, but your comment "in eth when parsing json config this field is called frontierCompatibilityModeLimit other then that its fine" had a point, people might have such config files.

Issue #3053, opened 2016-08-18
Modify dockerfile to run as user


bobsummerwill:
The current dockerfile at https://github.com/ethereum/cpp-ethereum/blob/develop/scripts/Dockerfile is built by docker hub https://hub.docker.com/r/ethereum/client-cpp/ and the way to run it is as follows:

    sudo docker run --net=host -v ~/.ethereum:/root/.ethereum -v ~/.web3:/root/.web3/ ethereum/client-cpp

So the ethereum data file of the current user on the host is made available inside the docker container. The problem is that this way, as the default user to run inside the container is root, all files in the host directory are later owned by root.

This issue is about changing the docker file and documenting the command to run it, so that the user used inside the container coincides with the user starting the process.

chriseth:
anyone: Please feel free to take this if you know something about docker (or want to learn how to use it)!

konradkonrad:
I'll take this one :)

chriseth:
Fixed in the PR.

PR #3347, opened 2016-10-13
Improve Docker for users

konradkonrad:
Tackles #3053

This PR consists of two parts and is my entry to the hacktoberfest

1) Reduce docker image size

Having all commands in one RUN and removing /cpp-ethereum afterwards allows to
reduce the image size from 2.686 GB to 1.252 GB.

Further image size reduction could be possible by removing build dependencies.

2) Document running container as user (non-root) #3053

There is no stable way to define a user of the host-system inside the
Dockerfile. Therefore I describe how to *run* the container as the current
user. I also added a convenience script to `scripts/`, that users can install
in their PATH, that will behave just like a local installation.

The other option (that I did not follow here) would be `--userns-remap`, but
this needs more setup and is only supported in newer docker installations.

### DIFF
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -2,14 +2,16 @@ FROM ubuntu:xenial
 MAINTAINER chriseth
 ENV DEBIAN_FRONTEND noninteractive
 
-RUN apt-get -y update
-RUN apt-get upgrade -q -y
-RUN apt-get -q -y install git lsb-release sudo software-properties-common
-RUN git clone https://github.com/ethereum/cpp-ethereum.git
-RUN cpp-ethereum/scripts/install_deps.sh
-RUN mkdir cpp-ethereum/build
-RUN cd cpp-ethereum/build && cmake .. && make install
-RUN ldconfig # should actually be part of make install...
+RUN apt-get update &&\
+    apt-get -q -y install git-core sudo lsb-release software-properties-common &&\
+    git clone https://github.com/ethereum/cpp-ethereum.git &&\
+    cpp-ethereum/scripts/install_deps.sh &&\
+    mkdir cpp-ethereum/build &&\
+    cd cpp-ethereum/build &&\
+    cmake .. && \
+    make install && \
+    ldconfig &&\
+    rm -rf /cpp-ethereum
 
 EXPOSE 8545 30303
 
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -2,8 +2,9 @@
 # https://github.com/ethereum/cpp-ethereum/blob/develop/scripts/docker-eth
 # or call `docker run` like this:
 #
-##    mkdir -p ~/.ethereum ~/.web3 && docker run --rm -it \
-##    --net=host -v ~/.ethereum:/.ethereum -v ~/.web3:/.web3 \
+##    mkdir -p ~/.ethereum ~/.web3
+##    docker run --rm -it --net=host \
+##    -v ~/.ethereum:/.ethereum -v ~/.web3:/.web3 \
 ##    -e HOME=/ --user $(id -u):$(id -g) ethereum/client-cpp
 
 FROM ubuntu:xenial
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -1,3 +1,11 @@
+# For running this container as non-root user, see
+# https://github.com/ethereum/cpp-ethereum/blob/develop/scripts/docker-eth
+# or call `docker run` like this:
+#
+##    mkdir -p ~/.ethereum ~/.web3 && docker run --rm -it \
+##    --net=host -v ~/.ethereum:/.ethereum -v ~/.web3:/.web3 \
+##    -e HOME=/ --user $(id -u):$(id -g) ethereum/client-cpp
+
 FROM ubuntu:xenial
 MAINTAINER chriseth
 ENV DEBIAN_FRONTEND noninteractive
--- a/scripts/docker-eth
+++ b/scripts/docker-eth
@@ -1,0 +1,14 @@
+#!/usr/bin/env sh
+mkdir -p ~/.ethereum ~/.web3
+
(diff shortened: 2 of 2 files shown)


scripts/docker-eth
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env sh
+mkdir -p ~/.ethereum ~/.web3
+
+# check if sudo is necessary
+if ! id -nG $(whoami)|grep -qw "docker"; then SUDO='sudo'; fi
```
chriseth: no `else SUDO=''`?

scripts/Dockerfile
```diff
@@ -1,15 +1,25 @@
...
+RUN apt-get update &&\
+    apt-get -q -y install git-core sudo lsb-release software-properties-common &&\
+    git clone https://github.com/ethereum/cpp-ethereum.git &&\
+    cpp-ethereum/scripts/install_deps.sh &&\
+    mkdir cpp-ethereum/build &&\
+    cd cpp-ethereum/build &&\
+    cmake .. && \
+    make install && \
+    ldconfig &&\
+    rm -rf /cpp-ethereum
```
chriseth: If you don't mind, I think the following can be added to remove the installed dependencies (although I somehow don't think they sum up to 1GB):

```
   rm -rf /libjson-rpc-cpp
    apt-get --purge remove libargtable2-dev libedit-dev build-essential \
                    cmake \
                    git \
                    libboost-all-dev \
                    libcurl4-openssl-dev \
                    libcryptopp-dev \
                    libgmp-dev \
                    libjsoncpp-dev \
                    libleveldb-dev \
                    libmicrohttpd-dev \
                    libminiupnpc-dev \
                    libz-dev \
                    llvm-3.9-dev \
                    mesa-common-dev \
                    ocl-icd-libopencl1 \
                    opencl-headers
```

I hope that removing the `-dev` packages still retains the non-dev (i.e. runtime library) packages.

scripts/Dockerfile
```diff
@@ -1,15 +1,25 @@
+# For running this container as non-root user, see
+# https://github.com/ethereum/cpp-ethereum/blob/develop/scripts/docker-eth
+# or call `docker run` like this:
+#
+##    mkdir -p ~/.ethereum ~/.web3 && docker run --rm -it \
```
chriseth: Perhaps clearer with a line break instead of the `&&`

scripts/docker-eth
konradkonrad:
```
$ sh
$ echo $UNKNOWN_VARIABLE_123

$ 
```
resolves to nothing just fine.

bobsummerwill:
What we could really use from a Dockerfile is one built on Alpine Linux, using static linkage.

That would have the smallest possible Docker footprint, and would produce the smallest possible binaries, in a form which would work on any distro.

scripts/docker-eth
chriseth: If you assume that the variable was not previously set, that is, but they are inherited from the environment.

chriseth:
@bobsummerwill I agree, but this is just to make it work for now.

scripts/Dockerfile
konradkonrad: I tried that, and it shrinked the container by another ~200MB, but it also prevented the program from running. I would say, for a long term solution go with @bobsummerwill s idea of a static alpine based build. In the scope of this ticket, I would suggest to stick with the 1.2 GB.

scripts/Dockerfile
konradkonrad: https://github.com/ethereum/cpp-ethereum/pull/3347/commits/d4dc8ffdfabbcd9d809c65f98e0dc24a61b08416

scripts/docker-eth
konradkonrad: https://github.com/ethereum/cpp-ethereum/pull/3347/commits/0d72f8898516dc4d3ceddd58400e8135a1fc84a2

chriseth:
Thanks a lot!

scripts/Dockerfile
chfast: Unfortunately, to get anything of that you need to have install and purge in the same `RUN` command. Each `RUN` creates new image so they can only grow.

scripts/docker-eth
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env sh
+mkdir -p ~/.ethereum ~/.web3
+
+# check if sudo is necessary
+if ! id -nG $(whoami)|grep -qw "docker"; then SUDO='sudo'; else SUDO=''; fi
```
chfast: Even if you are in the docker group, the eth process will run as root.

konradkonrad: Hey @chfast thanks for your comment!
The sudo check is only about issuing the `docker run ...` command as non root (if applicable). The ownership of the process is taken care of with the
```
--user $(id -u):$(id -g)
```
bit in the proposed script. So the process in the container will have the user/group id of your current user.

scripts/docker-eth
```diff
@@ -0,0 +1,14 @@
...
+
+$SUDO docker run --rm -it \
+           --net=host \
```
konradkonrad: By the way: the need for `--net=host` makes me unhappy.

It results in publishing the jsonrpc interface on `0.0.0.0:8545`, which is pretty insecure! It is necessary though in order to make the UPnP punched nat traversal work:

Without `--net=host`:
```
  ℹ  10:19:07.397|p2p  UPnP device not found.
```

There is an issue that seems to address this issue: (https://github.com/docker/docker/issues/3043 and in particular https://github.com/docker/docker/issues/3043#issuecomment-72387449), but I still could not get it to work.

If UPnP nat traversal is considered less important than the wide open rpc interface, then `--net=host` should be replaced by:
```
    -p 127.0.0.1:8545:8545 \                                                    
    -p 0.0.0.0:30303:30303 \   
```

@chriseth let me know, if you want this addressed!

PR #3348, opened 2016-10-13
Build jsoncpp from source using CMake's ExternalProject

chfast:
After the review I will move the `jsoncpp.cmake` script to deps repo and apply the same changes in the solidity to resolve https://github.com/ethereum/solidity/issues/789.


bobsummerwill:
Is this the same pattern as you used for LLVM, @chfast?

What do you see as the merits of downloading the ZIP versus using git sub-modules?

chfast:
It is the same pattern. It is just the ExternalProject is very nice combo. It will download the source code for you, unpack it, timestamp it (it is never rebuilt), build it, "install" it. You just don't have to think if you should use git submodule, or copy the source directly, etc.

chriseth:
Build error:
```
Linking CXX executable ethminer

/home/travis/build/ethereum/cpp-ethereum/ethminer/FarmClient.h:18: error: undefined reference to 'Json::Value::operator=(Json::Value const&)'
```

Apart from this: Wonderful! Another advantage of this: The automatically attached source tarball for releases should work with this way of using dependencies in contrast to using git submodules.

chfast:
I know that. I think I forgot to update dependencies of ethminer (probably I'm not building it).
I prefer using HTTP over GIT, because then you don't need git. Besides some project might not have the source in a git repo.

@chriseth @bobsummerwill  questions for you about jsoncpp.cmake script:
- If it looks fine I will move it to the "deps" repo to allow solidity to use it. Agree?
- I named the imported library "jsoncpp". The convention is to use `::` somewhere in the name but I prefer to keep it simple. Is this fine?

chriseth:
@chfast agree on both.

chriseth:
Actually now I think this will severely break packages (debian/ubuntu, for example), unless we link statically. The makefile will install libjsoncpp into a general directory and that clashes with existing packages for libjsoncpp. Can that be solved somehow?

If not, then I am fine with using docker-only "binaries" for now with the medium-term goal of static binaries.

chfast:
jsoncpp is linked statically.

I think I will spend a bit more time on this to get json-rpc-cpp the same way and to force it to use the jsoncpp from the cmake script. Otherwise it make no sense.

chriseth:
@chfast ah ok, but I saw some "install" rules, those should probably be removed.

bobsummerwill:
Good stuff, @chfast.

All good with me, as long as we don't break things horribly in the transition.

scripts/build.sh
```diff
@@ -36,4 +36,4 @@ else
     cmake .. -DCMAKE_BUILD_TYPE=$1 -DTESTS=$2
 fi
 
-make -j2
+make -j2 VERBOSE=1
```
chriseth: Remove this again?

PR #3351, opened 2016-10-16
Remove EVMSchedule from Chain Params

chfast:
Having EVMSchedule is very misleading as it is never initialized other than default constructor and Ethash seal engine never uses it.

### DIFF
--- a/libethcore/ChainOperationParams.h
+++ b/libethcore/ChainOperationParams.h
@@ -62,7 +62,6 @@ struct ChainOperationParams
 	u256 maximumExtraDataSize = 1024;
 	u256 accountStartNonce = 0;
 	bool tieBreakingGas = true;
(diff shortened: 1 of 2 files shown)


yann300:
Maybe Ethash should use the EVMSchedule given by chain params...
currently Ethash uses chain params to return either Homestead or Frontier EVMSchedule
https://github.com/ethereum/cpp-ethereum/blob/4b805926d46bf31144e053e6898cbedd3f67e023/libethashseal/Ethash.cpp#L85

chfast:
@yann300 chain params are initialized once per blockchain (genesis block), but EVMSchedule depends also on the block number -- that's way it is implemented as a function. In other words, having fixed EVMSchedule in chain params is not generic enough. This is the reason of this change.

Issue #3353, opened 2016-10-17
Limit cache size


chriseth:
We have to limit the size of all caches that can be pruned without losing information. I'm specifically looking at the code cache of Account.
- [x] implement a global code size cache that maps code hash to code size with a simple eviction strategy (perhaps just least recently loaded, i.e. a ring buffer).
- [x] add a `codeSizeAt` function to query only the code size (and not the code itself)
- [x] split m_cache into dirty (transaction effects) and non-dirty (actual cache) cache add a routine to State to purge the non-dirty cache if it gets too large.

PR #3355, opened 2016-10-17
Introduce the code size cache.

chriseth:
This cache stores a mapping from code hash to code size in order
to avoid having to load the full code for extcodesize.

Connects to #3353

### DIFF
--- a/libethereum/CodeSizeCache.h
+++ b/libethereum/CodeSizeCache.h
@@ -1,0 +1,79 @@
+/*
+	This file is part of cpp-ethereum.
+
+	cpp-ethereum is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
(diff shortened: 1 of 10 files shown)


libethereum/CodeSizeCache.h
```diff
@@ -0,0 +1,79 @@
...
+ * @brief Simple thread-safe cache to store a mapping from code hash to code size.
+ * If the cache is full, a random element is removed.
+ */
+class CodeSizeCache
+{
+public:
+	void store(h256 const& _hash, size_t size)
+	{
+		UniqueGuard g(x_cache);
+		if (m_cache.size() > c_maxSize)
```
pirapira: Not related to the test failure, but `>=` would fit better with the name `c_maxSize`.  Currently the size can grow to `c_maxSize + 1`.

libethereum/CodeSizeCache.h
```diff
@@ -0,0 +1,79 @@
...
+	{
+		UniqueGuard g(x_cache);
+		return m_cache.at(_hash);
+	}
+
+	static CodeSizeCache& instance() { static CodeSizeCache cache; return cache; }
+
+private:
+	/// Removes a random element from the cache.
+	void removeElement()
```
axic: Wouldn't `removeRandomElement()` be more fitting?

libethereum/State.cpp
```diff
@@ -228,10 +224,36 @@ void State::ensureCached(std::unordered_map<Address, Account>& _cache, const Add
 		else
 			s = Account(state[0].toInt<u256>(), state[1].toInt<u256>(), state[2].toHash<h256>(), state[3].toHash<h256>(), Account::Unchanged);
 		bool ok;
-		tie(it, ok) = _cache.insert(make_pair(_a, s));
+		tie(it, ok) = m_cache.insert(make_pair(_a, s));
```
chfast: I just remembered something: http://en.cppreference.com/w/cpp/utility/tuple/ignore.

PR #3359, opened 2016-10-19
Optimize EXP

chfast:
In the worst case (loop-exp-32b-100k) it improves performance 10x, from 2.6 Mgas/s to 26 Mgas/s. On My Machine™.

### DIFF
--- a/libevm/VM.cpp
+++ b/libevm/VM.cpp
@@ -42,6 +42,26 @@ template <class S> S modWorkaround(S const& _a, S const& _b)
 	return (S)(s512(_a) % s512(_b));
 }
 
+/// Implementation of EXP.
+///
+/// This implements exponentiation by squaring algorithm.
+/// Is is faster than boost::multiprecision::powm() because avoids explicit
+/// mod operation.
+/// Do not inline it.
+u256 exp256(u256 _base, u256 _exponent)
+{
+	using boost::multiprecision::limb_type;
+	u256 result = 1;
+	while (_exponent)
+	{
+		if (static_cast<limb_type>(_exponent) & 1)  // If exponent is odd.
+			result *= _base;
+		_base *= _base;
+		_exponent >>= 1;
+	}
+	return result;
+}
+
 
 //
 // for tracing, checking, metering, measuring ...
@@ -324,7 +344,7 @@ void VM::interpretCases()
 			updateIOGas();
 
 			auto base = *m_sp--;
-			*m_sp = (u256)boost::multiprecision::powm((bigint)base, (bigint)expon, bigint(1) << 256);
+			*m_sp = exp256(base, expon);
 			++m_pc;
 		}
 		CASE_END


libevm/VM.cpp
```diff
@@ -42,6 +42,25 @@ template <class S> S modWorkaround(S const& _a, S const& _b)
...
+/// This implements exponentiation by squaring algorithm.
+/// Is is faster than boost::multiprecision::powm() because avoids explicit
+/// mod operation.
+/// Do not inline it.
+u256 exp(u256 _base, u256 _exponent)
+{
+	using boost::multiprecision::limb_type;
+	u256 result = 1;
+	while (_exponent) {
+		if (static_cast<limb_type>(_exponent) & 1)
```
chriseth: Is the `static_cast` required? There is also the function `bit_test`.

libevm/VM.cpp
```diff
@@ -42,6 +42,25 @@ template <class S> S modWorkaround(S const& _a, S const& _b)
...
+
+/// This implements exponentiation by squaring algorithm.
+/// Is is faster than boost::multiprecision::powm() because avoids explicit
+/// mod operation.
+/// Do not inline it.
+u256 exp(u256 _base, u256 _exponent)
+{
+	using boost::multiprecision::limb_type;
+	u256 result = 1;
+	while (_exponent) {
```
chriseth: Style (brace to the next line)

libevm/VM.cpp
chfast: It is faster, probably because & and == true is applied to native int instead of u256. But bit_test will be perfect here, thanks.

libevm/VM.cpp
chfast: Sorry. I copied it from other place and my brain just ignored it.

gcolvin:
It's been on my list see if the if the bigint casts in the EXP code were necessary, but this looks much better, thanks.  Does it also happen to handle the case where base is a power of 2?

gcolvin:
@chfast  If you need to be sure that exp() is not inlined you should move it to VMOpt.cpp.

chfast:
Thanks @gcolvin, I will check it tomorrow. I just noticed if I add `static` to the function that decreases performance. Wierd.

I'm pretty disappointed with boost multiprecision. They (usually -- we found bugs as well) produce correct results, but algorithms are very naive and polluted with all the templating. The powm one is probably too generic, not being a part of cpp_int.

BTW, not interpreter is 10x faster than JIT. And both use the same algorithm :)

gcolvin:
I guess g++ has a faster calling convention for member functions?  But 
how could interpreter be faster?

chfast:
I don't know. Maybe mul sucks.

libevm/VM.cpp
chfast: `bit_test` is a kind of macro that creates a bitmask, applies it and tests. So static_cast is the fastest way to do it as for now.

PR #3378, opened 2016-10-25
Correctly purge cache and refactor State::ensureCached

chriseth:


libethereum/State.cpp
```diff
@@ -427,39 +426,45 @@ h256 State::storageRoot(Address const& _id) const
...
 
-bytes const& State::code(Address const& _contract) const
+bytes const& State::code(Address const& _a) const
 {
-	if (!addressHasCode(_contract))
+	if (!addressHasCode(_a))
 		return NullBytes;
-	ensureCached(_contract, true, false);
-	return m_cache[_contract].code();
+	return account(_a, true)->code();
```
chfast: This is bad design because you have to search twice. Maybe the Account should have a method "fetch code".

libethereum/State.cpp
```diff
@@ -427,39 +426,45 @@ h256 State::storageRoot(Address const& _id) const
...
-		return codeSizeCache.get(codeHash);
-	else
+	if (Account const* a = account(_a))
 	{
-		size_t size = code(_contract).size();
-		codeSizeCache.store(codeHash, size);
-		return size;
+		if (a->isFreshCode())
+			return code(_a).size();
+		auto& codeSizeCache = CodeSizeCache::instance();
```
chfast: Now I can see that this global code size cache is bad idea. We should have a good account cache (as we need it anyway) and keep the code size information in the account itself.

libethereum/State.h
```diff
@@ -247,11 +250,18 @@ class State
...
+	Account const* account(Address const& _a, bool _requireCode = false) const;
+
+	/// @returns the account at the given address or a null pointer if it does not exist.
+	/// The pointer is valid until the next access to the state or account.
+	Account* account(Address const& _a, bool _requireCode = false);
+
+	/// Retrieve certain information about a given address into the cache and return
+	/// a pointer to the relevant account (or a null pointer if it does not exist).
+	/// If _requireCode is true, also load the code.
+	Account* ensureCached(Address const& _a, bool _requireCode) const;
```
chfast: You are too pedantic about `const` here. Let's keep the single function instead of 3 of them.

libethereum/State.cpp
chriseth: Can you please elaborate? It has to be global in order to speed up code size lookup for two contracts that have the same code - that is its only purpose. In addition, we can add another cache inside the account, but I'm not sure if that is an improvement because the account data structure will grow.

libethereum/State.cpp
```diff
@@ -288,126 +297,116 @@ void State::setRoot(h256 const& _r)
...
-	return it->second.balance();
 }
 
 void State::noteSending(Address const& _id)
 {
-	ensureCached(_id, false, false);
-	auto it = m_cache.find(_id);
-	if (asserts(it != m_cache.end()))
+	Account* a = account(_id);
+	if (asserts(a != nullptr))
```
chriseth: I agree, this is just a refactoring.

libethereum/State.cpp
```diff
@@ -288,126 +297,116 @@ void State::setRoot(h256 const& _r)
...
 
 void State::createContract(Address const& _address)
 {
 	m_cache[_address] = Account(requireAccountStartNonce(), balance(_address), Account::ContractConception);
 }
 
-void State::kill(Address _a)
+void State::ensureAccountExists(const Address& _address)
+{
+	if (!addressInUse(_address))
```
chriseth: Do you want to remove `addressInUse` completely?

libethereum/State.h
chriseth: I would prefer a solution with only two functions, but I think it is important to distinguish between const and non-const access to the accounts. This might save us another bug.

libethereum/State.cpp
chfast: We have the account cache and the code size cache. But the account is a tuple (nonce, balance, ptr to code, ptr to storage). It can be easily extended to tuple (nonce, balance, code size, ptr to code, ptr to storage) and the we can drop the code size cache. Just the idea, not request for changes in this PR.

libethereum/State.h
chfast: How about using const_cast and having 2 function instead of 3?

libethereum/State.cpp
chfast: It is confusing for me that we have: "exists", "is empty", and "in use" queries. My guess is "in use" should go away :).

PR #3384, opened 2016-10-27
Fix cmake replace on empty string.

chriseth:


chriseth:
@chfast could you review and merge if ok, please? The failure is a test failure and as this is a change in the build system, I don't think it is related.

Issue #3387, opened 2016-11-01
Bugs in the LLL code snippets of the testing suite


axic:
This is a follow up of #3356.

I've run the code snippets through a stricter lllc (see https://github.com/ethereum/solidity/issues/1313) and it has found a lot of potential buggy lll snippets: https://gist.github.com/axic/c245b7d7f1a3896282de77e69d2f568d

Issues:
- unterminated code (missing `)` or `}`)
- using missing instructions (`SAWP1` vs. `SWAP1`, potentially archaic ones: `PREVHASH`, `POST`, `CALLSTATELESS`, `CALLVALUE`)
- using `asm` without `()`

Some of the bugs might be false positive (such as `MSIZE` and the missing `i` variable, have to investigate).

axic:
I assume `POST` is an old name for `CALL` and `CALLSTATELESS` is an old name for `CALLCODE`?

axic:
`CALLVALUE` and `MSIZE` are methods and must be called or wrapper in an assembly block:
`(CALLVALUE)` or `(asm CALLVALUE)`.

PR #3388, opened 2016-11-01
Fix obvious bugs in the LLL sources used by the tests

axic:


axic:
Those discovered by #3387.

winsvega:
https://github.com/ethereum/cpp-ethereum/pull/3389

PR #3390, opened 2016-11-03
Debug Devcon performance improvements and merge with latest

gcolvin:
Most of the merge was straightforward.  An exception was integrating the new call gas logic in VMCalls.cpp with changes to the m_ioGas and ioGas variables.


gcolvin:
There are 6 test failures, all in bcSuicideIssueHomestead, and all are present before I apply my changes.

gcolvin:
At least turning optimizations back on doesn't introduce any new errors.

gcolvin:
The intermittent stRecursiveCreate error showed up this time.  It also shows up without my changes.

gcolvin:
@chriseth @chfast  The test failures are all pre-existing.  Could you review the code changes here so we can get this merged?

libevm/VM.cpp
```diff
@@ -129,35 +123,46 @@ void VM::logGasMem()
...
+		// trampoline to minimize depth of call stack when calling out
+		m_bounce = &VM::initEntry;
+		do
+			(this->*m_bounce)();
+		while (m_bounce);
+		
+	}
+	catch (...)
+	{
+		*io_gas = m_io_gas;
```
chfast: Exception always means OOG, so you can assign 0 here and there will be no need to update m_io_gas before throwing the exception.

libevm/VM.h
```diff
@@ -69,14 +69,14 @@ class VM: public VMFace
 
 private:
 
-	u256* m_io_gas = 0;
+	u256* io_gas = 0;
+	uint64_t m_io_gas = 0;
```
chfast: Do you use the last bit of uint64_t anywhere?

libevm/VMConfig.h
```diff
@@ -31,52 +31,48 @@ namespace eth
...
 // EVM_REPLACE_CONST_JUMP - with pre-verified jumps to save runtime lookup
 //
-// ETH_VMTRACE - set by cmake to enable tracing by testeth, ethvm and others
-//
 // EVM_TRACE - provides various levels of tracing that override ETH_VMTRACE
 
-#if true
-	#define EVM_SWITCH_DISPATCH
-#elif defined(__GNUG__)
+#if true && defined(__GNUG__)
```
chfast: This `#if true` looks messy.

libevm/VM.cpp
gcolvin: Try it and tests will fail.  The io_gas reference is used by callers to determine exactly how much gas was used before the exception.

libevm/VM.h
gcolvin: I doubt it, but I switched the code to 64 bits months before we finally decided on 63 bits.

libevm/VMConfig.h
gcolvin: It is.  You set it false to disable this GNU-only optimization.

libevm/VM.h
chfast: I just learned that signed integers are faster because overflow is UB.

gcolvin: What does 'UB' mean?  But I thought with two's complement everything but division was the same operation.  Anyway, no need to revisit uint64 versus int64 for this PR.

gcolvin: Aha.

chfast:
@chriseth are you fine merging this?

chfast:
@gcolvin Please merge changes from develop branch. You should get clean build.

chriseth:
@gcolvin if you create your pull requests by directly pushing to a custom branch on the general repository (not your private fork), then anyone can work on the branch and also e.g. rebase it.

libevm/VM.cpp
chriseth: Only exceptions that are in line with consensus should be caught here. An out-of-memory exception, for example, should not be caught. Otherwise we will get wrong and misleading results.

gcolvin: Everything caught here is rethrown.  The purpose is just to set the 256-bit _io_gas reference here, so that I don't have to set it via 256-bit arithmetic everywhere else - if I didn't catch everything here _io_gas could be wrong on some exceptions.  So the behavior should look just the same to the caller as is did before I introduced the try/catch.

libevm/VM.h
gcolvin: Oh, I might use the last bit via comparisons to 2^63-1.

gcolvin:
@chfast  I merged with develop before pushing, there are no conflicts, and the build is clean.  I can merge again and push again if you want, or we can let the PR handle the merge.
@chriseth  I'll push that way in the future.  Thanks for the tip.

gcolvin:
Moved to https://github.com/ethereum/cpp-ethereum/pull/3399

PR #3397, opened 2016-11-04
eip155 transaction hash rules

winsvega:
transaction rules by EIP155

### DIFF
--- a/libethcore/Transaction.cpp
+++ b/libethcore/Transaction.cpp
@@ -64,9 +64,14 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
 			BOOST_THROW_EXCEPTION(InvalidTransactionFormat() << errinfo_comment("transaction data RLP must be an array"));
 
 		m_data = rlp[field = 5].toBytes();
-		byte v = rlp[field = 6].toInt<byte>() - 27;
+
+		byte v = rlp[field = 6].toInt<byte>();
 		h256 r = rlp[field = 7].toInt<u256>();
 		h256 s = rlp[field = 8].toInt<u256>();
+
+		m_chainId = std::floor((float)(v - 35)/2);
+		if(m_chainId == -4 || m_chainId == 1)
+			v = rlp[field = 6].toInt<byte>() - m_chainId*2 - 35;
 
 		if (rlp.itemCount() > 9)
 			BOOST_THROW_EXCEPTION(InvalidTransactionFormat() << errinfo_comment("to many fields in the transaction RLP"));
@@ -127,8 +132,9 @@ void TransactionBase::streamRLP(RLPStream& _s, IncludeSignature _sig) const
 	else
 		_s << "";
 	_s << m_value << m_data;
+
 	if (_sig)
-		_s << (m_vrs.v + 27) << (u256)m_vrs.r << (u256)m_vrs.s;
+		_s << (m_vrs.v + m_chainId*2 + 35) << (u256)m_vrs.r << (u256)m_vrs.s;
 }
 
 static const u256 c_secp256k1n("115792089237316195423570985008687907852837564279074904382605163141518161494337");
@@ -147,3 +153,29 @@ bigint TransactionBase::gasRequired(bool _contractCreation, bytesConstRef _data,
 	return ret;
 }
 
+h256 TransactionBase::sha3(IncludeSignature _sig) const
+{
+	if (_sig == WithSignature && m_hashWith)
+		return m_hashWith;
+
+	RLPStream s;
+	if (m_chainId == -4 && _sig == WithSignature && m_type != NullTransaction) //EIP155 is On
+	{
+		//!!! code copy from streamRLP!  cant change streamRLP because it should not be affected by EIP155 proposal (it affects only hash)
+		s.appendList(9);
+		s << m_nonce << m_gasPrice << m_gas;
+		if (m_type == MessageCall)
+			s << m_receiveAddress;
+		else
+			s << "";
+		s << m_value << m_data;
+		s << (m_vrs.v + m_chainId*2 + 35) << (u256)0 << (u256)0;
+	}
+	else
+		streamRLP(s, _sig);
+
+	auto ret = dev::sha3(s.out());
+	if (_sig == WithSignature)
+		m_hashWith = ret;
+	return ret;
+}
--- a/libethcore/Transaction.cpp
+++ b/libethcore/Transaction.cpp
@@ -159,7 +159,8 @@ h256 TransactionBase::sha3(IncludeSignature _sig) const
 		return m_hashWith;
 
 	RLPStream s;
-	if (m_chainId == -4 && _sig == WithSignature && m_type != NullTransaction) //EIP155 is On
+
+	if (m_chainId == 1 /*&& _sig == WithSignature*/ && m_type != NullTransaction) //EIP155 is On
 	{
 		//!!! code copy from streamRLP!  cant change streamRLP because it should not be affected by EIP155 proposal (it affects only hash)
 		s.appendList(9);
--- a/libethcore/Transaction.cpp
+++ b/libethcore/Transaction.cpp
@@ -69,9 +69,12 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
 		h256 r = rlp[field = 7].toInt<u256>();
 		h256 s = rlp[field = 8].toInt<u256>();
 
-		m_chainId = std::floor((float)(v - 35)/2);
-		if(m_chainId == -4 || m_chainId == 1)
-			v = rlp[field = 6].toInt<byte>() - m_chainId*2 - 35;
+		if (v == 27 || v == 28)
+			m_chainId = -4;
+		else if (v == 37 || v == 38)
+			m_chainId = 1;
+
+		v = v - (m_chainId * 2 + 35);
 
 		if (rlp.itemCount() > 9)
 			BOOST_THROW_EXCEPTION(InvalidTransactionFormat() << errinfo_comment("to many fields in the transaction RLP"));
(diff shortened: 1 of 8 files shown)


libethcore/Transaction.cpp
```diff
@@ -64,9 +65,15 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
 			BOOST_THROW_EXCEPTION(InvalidTransactionFormat() << errinfo_comment("transaction data RLP must be an array"));
 
 		m_data = rlp[field = 5].toBytes();
-		byte v = rlp[field = 6].toInt<byte>() - 27;
+		m_chainID = _chainID;
+
+		byte v = rlp[field = 6].toInt<byte>();
 		h256 r = rlp[field = 7].toInt<u256>();
 		h256 s = rlp[field = 8].toInt<u256>();
+		if(v == m_chainID*2+35 || v == m_chainID*2+36)
```
chfast: Here you should calculate the transaction chain id as `txChainId = (v - 35) / 2`. Then `txChainId` either match `m_chainId` or is `-4` for Frontier transactions. If it is other value you should throw an exception to reject the transaction.

libethcore/Transaction.cpp
```diff
@@ -127,8 +134,14 @@ void TransactionBase::streamRLP(RLPStream& _s, IncludeSignature _sig) const
 	else
 		_s << "";
 	_s << m_value << m_data;
+
 	if (_sig)
-		_s << (m_vrs.v + 27) << (u256)m_vrs.r << (u256)m_vrs.s;
+	{
+		if (m_chainID >= 1)
```
chfast: Use `-4` as the default chain id to have Frontier compatibility. You should name it somewhere as `static const int frontierChainId = -4` and use instead of `-1` as default argument value. 

Please also use `chainId` name instead of `chainID`.

libethcore/Transaction.cpp
```diff
@@ -64,10 +64,15 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
 			BOOST_THROW_EXCEPTION(InvalidTransactionFormat() << errinfo_comment("transaction data RLP must be an array"));
 
 		m_data = rlp[field = 5].toBytes();
-		byte v = rlp[field = 6].toInt<byte>() - 27;
+
+		byte v = rlp[field = 6].toInt<byte>();
 		h256 r = rlp[field = 7].toInt<u256>();
 		h256 s = rlp[field = 8].toInt<u256>();
 
+		m_chainId = std::floor((float)(v - 35)/2);
```
chriseth: Please don't use float, it might create some precision problems. What about
`(int(unsigned(v)) - 35) / 2`?

libethcore/Transaction.cpp
```diff
@@ -64,10 +64,15 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
...
+		if(m_chainId == -4 || m_chainId == 1)
```
chriseth: Space after if.

libethcore/Transaction.cpp
```diff
@@ -64,10 +64,15 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
...
+			v = rlp[field = 6].toInt<byte>() - m_chainId*2 - 35;
```
chriseth: Intuitively, I would say you need `+35` here.

libethcore/Transaction.cpp
chfast: Just stick to `int` and it will be fine. Integer division is all we want here.

libethcore/Transaction.cpp
chfast: Intuitively, I think it is fine, but better to spell it like `- (m_chainId * 2 + 35)`. Please also keep spaces between binary arithmetic operators.

libethcore/Transaction.cpp
winsvega: i need X/2  to be float or double to use floor

chfast: You don't need `floor()` because integer division rounds down as floor.

winsvega: anyways it doesn't work without floor

libethcore/Transaction.cpp
```diff
@@ -147,3 +153,30 @@ bigint TransactionBase::gasRequired(bool _contractCreation, bytesConstRef _data,
...
+	{
+		//!!! code copy from streamRLP!  cant change streamRLP because it should not be affected by EIP155 proposal (it affects only hash)
+		s.appendList(9);
+		s << m_nonce << m_gasPrice << m_gas;
+		if (m_type == MessageCall)
+			s << m_receiveAddress;
+		else
+			s << "";
+		s << m_value << m_data;
+		s << (m_vrs.v + m_chainId*2 + 35) << (u256)0 << (u256)0;
```
chfast: This should be `s << m_chainId << 0 << 0`.

libethcore/Transaction.cpp
```diff
@@ -64,10 +64,18 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
 			BOOST_THROW_EXCEPTION(InvalidTransactionFormat() << errinfo_comment("transaction data RLP must be an array"));
 
 		m_data = rlp[field = 5].toBytes();
-		byte v = rlp[field = 6].toInt<byte>() - 27;
+
+		byte v = rlp[field = 6].toInt<byte>();
```
chfast: It seems that `field` var is never read. If I'm right, please remove it.

libethcore/Transaction.cpp
```diff
@@ -127,8 +135,9 @@ void TransactionBase::streamRLP(RLPStream& _s, IncludeSignature _sig) const
 	else
 		_s << "";
 	_s << m_value << m_data;
+
 	if (_sig)
-		_s << (m_vrs.v + 27) << (u256)m_vrs.r << (u256)m_vrs.s;
+		_s << (m_vrs.v + m_chainId*2 + 35) << (u256)m_vrs.r << (u256)m_vrs.s;
```
chfast: Please move this expression out to additional variable. Like `auto vOffset = m_chainId * 2 + 35`.

libethcore/Transaction.cpp
```diff
@@ -147,3 +156,29 @@ bigint TransactionBase::gasRequired(bool _contractCreation, bytesConstRef _data,
 	return ret;
 }
 
+h256 TransactionBase::sha3(IncludeSignature _sig) const
+{
+	if (_sig == WithSignature && m_hashWith)
+		return m_hashWith;
+
+	RLPStream s;
+	if (m_chainId == 1 && m_type != NullTransaction) //EIP155 is On
```
chfast: I suggest to make it useful for any valid chain id. So it should be `m_chainId > 0`.

libethcore/Transaction.cpp
```diff
@@ -147,3 +156,29 @@ bigint TransactionBase::gasRequired(bool _contractCreation, bytesConstRef _data,
...
+	{
+		//!!! code copy from streamRLP!  cant change streamRLP because it should not be affected by EIP155 proposal (it affects only hash)
+		s.appendList(9);
+		s << m_nonce << m_gasPrice << m_gas;
+		if (m_type == MessageCall)
+			s << m_receiveAddress;
+		else
+			s << "";
+		s << m_value << m_data;
+		s << m_chainId << (u256)0 << (u256)0;
```
chfast: I might be wrong, but using just `0`s should also work.

libethcore/Transaction.h
```diff
@@ -169,6 +169,7 @@ class TransactionBase
 	u256 m_gas;							///< The total gas to convert, paid for from sender's account. Any unused gas gets refunded once the contract is ended.
 	bytes m_data;						///< The data associated with the transaction, or the initialiser if it's a creation transaction.
 	SignatureStruct m_vrs;				///< The signature of the transaction. Encodes the sender.
+	int m_chainId;						///< EIP155 value
```
chfast: Please extend this comment.

libethcore/Transaction.cpp
winsvega: it looks like field is used to make it more readable 
rlp[field = 1]  istead of rlp[1]

libethcore/Transaction.cpp
```diff
@@ -147,3 +159,29 @@ bigint TransactionBase::gasRequired(bool _contractCreation, bytesConstRef _data,
...
 
+h256 TransactionBase::sha3(IncludeSignature _sig) const
+{
+	if (_sig == WithSignature && m_hashWith)
+		return m_hashWith;
+
+	RLPStream s;
+	if (m_chainId > 0 && m_type != NullTransaction) //EIP155 is On
+	{
+		//!!! code copy from streamRLP!  cant change streamRLP because it should not be affected by EIP155 proposal (it affects only hash)
```
chriseth: Why not create a helper function?

libethcore/Transaction.cpp
```diff
@@ -64,10 +64,18 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
...
+
+		byte v = rlp[field = 6].toInt<byte>();
 		h256 r = rlp[field = 7].toInt<u256>();
 		h256 s = rlp[field = 8].toInt<u256>();
 
+		if (v == 27 || v == 28)
+			m_chainId = -4;
+		else if (v == 37 || v == 38)
+			m_chainId = 1;
+
```
chriseth: What is the value of `m_chainId` in the else case?

libethcore/Transaction.cpp
winsvega: helper is fine. 
I do not think that hash calculation should be puted in streamRLP
streamRLP means stream the transaction data into RLP. 
streamRLPforEIP155 better be a helper

libethcore/Transaction.cpp
winsvega: default = 0

PR #3402, opened 2016-11-08
working performance tests

gcolvin:
Changed rc5.sol to use Rivests original data types and constants, and to use custom exp of base 2 so as to get gas requirements down.  Also more loops in loop.sol to get comparable runtime to others.


gcolvin:
@winsvega @chfast This is the working rc5.sol you've been waiting for.

PR #3403, opened 2016-11-08
testing performance and minor refactoring

gcolvin:
Test speed with and without recent improvements, moved Pawel's new exp() code to VMOpt.cpp, removed obscuring uses of auto, other housekeeping.

### DIFF
--- a/libevm/VMConfig.h
+++ b/libevm/VMConfig.h
@@ -33,17 +33,17 @@ namespace eth
 //
 // EVM_TRACE - provides various levels of tracing that override ETH_VMTRACE
 
-#if false && defined(__GNUG__)
+#if true && defined(__GNUG__)
 	#define EVM_JUMP_DISPATCH
 #else
 	#define EVM_SWITCH_DISPATCH
 #endif
 
-#if false
+#if true
 	#define EVM_REPLACE_CONST_JUMP
 #endif
 
-#if false
+#if true
 	#define EVM_USE_CONSTANT_POOL
 #endif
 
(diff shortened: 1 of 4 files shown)


libevm/VMConfig.h
```diff
@@ -33,17 +33,17 @@ namespace eth
...
 	#define EVM_SWITCH_DISPATCH
 #endif
 
-#if false
+#if true
 	#define EVM_REPLACE_CONST_JUMP
 #endif
 
-#if false
+#if true
```
chriseth: Are these leftovers?

libevm/VMConfig.h
```diff
@@ -54,7 +54,9 @@ namespace eth
 
 
 // set this to 2, 1, or 0 for more, less, or no tracing to cerr
-#define EVM_TRACE 0
+#ifndef EVM_TRACE
```
chfast: In case a macro is not defined `#if` should replace it with `0`. Please remove it, unless any compiler complains about it.

chfast:
However, clang complains about
```
/Users/travis/build/ethereum/cpp-ethereum/libevm/VM.cpp:738:19: error: unsequenced modification and access to 'm_pc' [-Werror,-Wunsequenced]
                        m_pc += m_code[++m_pc];
                             ~~        ^
```
And I agree.

libevm/VMConfig.h
gcolvin: You are right.

gcolvin:
Yes, clang is right.  I keep fixing that line, and git keeps merging it back in.

gcolvin:
As for auto - I don't like it for scalars where I do care and should know what they are.  As compared to unthinkable iterator types that just work somehow.

libevm/VMConfig.h
gcolvin: Leftovers?  I've done more testing so I'm turning them on.

gcolvin:
@chriseth @chfast  Can someone push Merge button?

gcolvin:
Thanks, @chriseth

PR #3406, opened 2016-11-09
Random cleanups

chfast:


cmake/EthCompilerSettings.cmake
```diff
@@ -71,7 +71,7 @@ if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MA
 	add_compile_options(-fPIC)
 
 	# Configuration-specific compiler settings.
-	set(CMAKE_CXX_FLAGS_DEBUG          "-O0 -g -DETH_DEBUG")
+	set(CMAKE_CXX_FLAGS_DEBUG          "-Og -g -DETH_DEBUG")
```
axic: @chriseth I think we should carry over this change to Solidity.

chriseth: Yes, good idea.

PR #3408, opened 2016-11-10
fix current pc

yann300:
strange cause it was already fixed 
https://github.com/ethereum/cpp-ethereum/commit/bb9a0cb99ceeee63c68a2127a409d82e8587a4e4

I did not tested it for now (can't build). will try soon.

### DIFF
--- a/libevm/VM.cpp
+++ b/libevm/VM.cpp
@@ -180,7 +180,6 @@ void VM::interpretCases()
 		
 		CASE_BEGIN(CREATE)
 			m_bounce = &VM::caseCreate;
-			++m_pc;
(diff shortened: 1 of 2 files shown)


yann300:
That's fine on my side (that fix the bug in remix). please review.

chriseth:
Oh, I remember we had a similar bug before. Perhaps it was re-introduced as a merge conflict.

gcolvin:
I've had at least one other fix disappear, so there may well have been a merge problem.

At some point we need a better way of providing tracing and debugging than giving users of the VM direct access to its internals.  Java provides a wire protocol.

PR #3381, opened 2016-10-26
EIP 158 + 160 protocol update

chriseth:
This implements https://github.com/ethereum/eips/issues/158 and https://github.com/ethereum/eips/issues/160 and https://github.com/ethereum/eips/issues/170

Still to do:
- [x] https://github.com/ethereum/eips/issues/155
- [x] Set transition block number in genesisInfo and tests

PR #3410, opened 2016-11-11
Limit code size (EIP170).

chriseth:
Implements https://github.com/ethereum/EIPs/issues/170


winsvega:
looks good but i think it is supposed to go to the HF branch eip158 
as it is a HF change

chriseth:
@winsvega we should merge both to develop and then create a new PR that sets the fork block number to something in the near future

chriseth:
Now part of #3381

PR #3419, opened 2016-11-21
State journal

chfast:


libethereum/Executive.h
```diff
@@ -78,6 +78,14 @@ class StandardTrace
 	DebugOptions m_options;
 };
 
+
+enum NonceRevertPolicy: bool
+{
+	keepNonce,
```
chriseth: Please make them uppercase.

libethereum/State.cpp
```diff
@@ -342,6 +342,14 @@ void State::noteSending(Address const& _id)
 		a->incNonce();
 }
 
+void State::revertIncNonce(Address const& _addr)
+{
+	auto& nonce = account(_addr)->nonce();
+	cwarn << "Revert nonce " << _addr << nonce;
```
chriseth: Perhaps check for zero?

libethereum/ExtVM.cpp
```diff
@@ -95,30 +95,78 @@ void go(unsigned _depth, Executive& _e, OnOpFunc const& _onOp)
...
 	io_gas = e.gas();
+	if (!e.newAddress())
+		m_successfulCalls.pop_back();
 	return e.newAddress();
 }
+
+void ExtVM::revert()
+{
+	clog(ExecutiveWarnChannel) << "Reverting " << myAddress;
+	for (auto it = m_successfulCalls.rbegin(); it != m_successfulCalls.rend(); ++it)
```
chriseth: I'm not sure if it is a good idea keeping a full executive around. Perhaps it is just better to store a vector of operations to revert and append the revert-vector of each successful call to the parent vector. Here, "operation to revert" is just either "(store, 0x123, 0x45)", "incnonce", "changeBalance, 0x123, 0x34", and whatever else we have.

Another solution would be to create a separate class "ReversionLog" that contains all the information needed in here and might refer to sub-logs.

libethereum/ExtVM.cpp
```diff
@@ -95,30 +95,78 @@ void go(unsigned _depth, Executive& _e, OnOpFunc const& _onOp)
...
+
+	// Restore original storage for this account. The order does not matter.
+	clog(ExecutiveWarnChannel) << "Reverting storage " << myAddress;
+	for (auto& item: m_origStorage)
+	{
+		m_s.setStorage(myAddress, item.first, item.second);
+		clog(ExecutiveWarnChannel) << "REVERT STORAGE " << myAddress << item.first << item.second;
+	}
+
+	while (m_nonceInc)
```
chriseth: Perhaps better to store `m_origNonce`?

libethereum/State.cpp
chfast: Too defensive in this case. I know it has been increased. Anyway, I'm working on improved account nonce API.

libethereum/Account.h
```diff
@@ -161,7 +165,7 @@ class Account
 	h256 codeHash() const { assert(!isFreshCode()); return m_codeHash; }
 
 	/// Sets the code of the account. Must only be called when isFreshCode() returns true.
-	void setCode(bytes&& _code) { assert(isFreshCode()); m_codeCache = std::move(_code); changed(); }
+	void setCode(bytes&& _code) { m_codeCache = std::move(_code); changed(); }
```
gumb0: You removed the assert, shouldn't the comment above be modified?

libethereum/Account.h
```diff
@@ -142,6 +144,8 @@ class Account
 	/// @returns the storage overlay as a simple hash map.
 	std::unordered_map<u256, u256> const& storageOverlay() const { return m_storageOverlay; }
 
+	void clearStorage() { m_storageOverlay.clear(); }
```
chriseth: `clearStorageChanges`?

libethereum/Account.h
chriseth: Also, why did you remove the assert? :-)

libethereum/ExtVM.cpp
```diff
@@ -95,30 +95,94 @@ void go(unsigned _depth, Executive& _e, OnOpFunc const& _onOp)
 
 bool ExtVM::call(CallParameters& _p)
 {
-	Executive e(m_s, envInfo(), m_sealEngine, depth + 1);
+	m_successfulCalls.emplace_back(m_s, envInfo(), m_sealEngine, depth + 1);
+	auto& e = m_successfulCalls.back();
```
chriseth: Please either specify the type of `e` or rename it to `executive`.

libethereum/ExtVM.cpp
```diff
@@ -95,30 +95,94 @@ void go(unsigned _depth, Executive& _e, OnOpFunc const& _onOp)
 
 bool ExtVM::call(CallParameters& _p)
 {
-	Executive e(m_s, envInfo(), m_sealEngine, depth + 1);
+	m_successfulCalls.emplace_back(m_s, envInfo(), m_sealEngine, depth + 1);
```
chriseth: Can you add a comment why we first add it to successful calls and remove it again in case it was not successful instead of just adding it when it returned successfully?

libethereum/ExtVM.cpp
```diff
@@ -95,30 +95,94 @@ void go(unsigned _depth, Executive& _e, OnOpFunc const& _onOp)
...
+	m_s.setStorage(myAddress, _n, _v);
+}
+
 h160 ExtVM::create(u256 _endowment, u256& io_gas, bytesConstRef _code, OnOpFunc const& _onOp)
 {
-	Executive e(m_s, envInfo(), m_sealEngine, depth + 1);
+	// Every CREATE increases this account nonce, no matter if it succeeds.
+	++m_nonceInc;
+
+	m_successfulCalls.emplace_back(m_s, envInfo(), m_sealEngine, depth + 1);
```
chriseth: Some comment as for `call` also applies here.

libethereum/State.cpp
```diff
@@ -336,6 +336,13 @@ void State::incNonce(Address const& _addr)
 		m_cache[_addr] = Account(requireAccountStartNonce() + 1, 0);
 }
 
+void State::setNonce(Address const& _addr, u256 const& _nonce)
```
gumb0: Was this previously declared but not implemented?

libethereum/ExtVM.cpp
```diff
@@ -95,30 +95,94 @@ void go(unsigned _depth, Executive& _e, OnOpFunc const& _onOp)
...
+	// Revert nonce if dumped.
+	if (m_nonceInc > 0)
+		m_s.setNonce(myAddress, m_s.getNonce(myAddress) - m_nonceInc);
+
+	if (m_selfdestructBeneficiary)
+	{
+//		clog(ExecutiveWarnChannel) << "REVERT SELFDESTRUCT touch " << myAddress << m_selfdestructBeneficiary;
+		m_s.untouch(m_selfdestructBeneficiary);
+	}
+
```
gumb0: What happens with ExtVM instance after revert() is called?
At this moment we have everything already reverted, but also m_successfulCalls, m_origStorage, m_nonceInc, m_selfdestructBeneficiary having some values.

libethereum/State.h
```diff
@@ -211,9 +214,13 @@ class State
 	/// Sets the code of the account. Must only be called during / after contract creation.
 	void setCode(Address const& _address, bytes&& _code) { m_cache[_address].setCode(std::move(_code)); }
 
+	void clearStorage(Address const& _addr) { m_cache[_addr].clearStorage(); }
+
 	/// Delete an account (used for processing suicides).
 	void kill(Address _a);
 
+	bool isAlive(Address const& _addr) { auto a = account(_addr); return a && a->isAlive(); }
```
gumb0: const method?

libethereum/Account.h
chfast: In general, I'm using it to also clear the code when reverting. There are alternatives to this, but I believe this API is too protective.

libethereum/ExtVM.cpp
chfast: Because when using `emplace_back` the Executive is created only once, and never moved.

libethereum/ExtVM.cpp
chfast: Good point. I think the contract of usage of `revert()` should be use only once. We can make second `revert()` doing nothing, but that is too defensive in my opinion and need additional care.

libethereum/State.cpp
chfast: Exactly. This method was introduces just before this PR, but I forgot to actually provide the implementation. See https://github.com/ethereum/cpp-ethereum/pull/3444.

test/libdevcrypto/crypto.cpp
```diff
@@ -145,7 +145,7 @@ BOOST_AUTO_TEST_CASE(common_encrypt_decrypt)
 BOOST_AUTO_TEST_CASE(cryptopp_cryptopp_secp256k1libport)
 {
 #if ETH_HAVE_SECP256K1
-	secp256k1_context_t* ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN | SECP256K1_CONTEXT_VERIFY);
+//	auto ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN | SECP256K1_CONTEXT_VERIFY);
```
chfast: This should not be here,

chfast:
Refactored that a bit. Also uses less memory.

libethereum/State.h
```diff
@@ -210,9 +213,13 @@ class State
 	/// Sets the code of the account. Must only be called during / after contract creation.
 	void setCode(Address const& _address, bytes&& _code) { m_cache[_address].setCode(std::move(_code)); }
 
+	void clearStorage(Address const& _addr) { m_cache[_addr].clearStorageChanges(); }
```
gumb0: This probably should be named clearStorageChanges, too

libethereum/Executive.cpp
```diff
@@ -459,7 +480,63 @@ void Executive::finalize()
...
+void Executive::revert()
+{
+	if (m_ext)
+		m_ext->sub.clear();
+
+	eth::revert(m_s, m_orig);
+	m_orig.address = {};
+}
+
+void dev::eth::revert(State& _state, AccountSnapshot const& _changes)
```
chfast: It was used by both Executive and ExtVM at some point, but I guess it not any more. To be fixed soon...

libethereum/State.h
chfast: Yep. Automatic rename sometimes sucks...

libethereum/State.h
```diff
@@ -161,6 +161,9 @@ class State
 	/// Check if the address is in use.
 	bool addressInUse(Address const& _address) const;
 
+	bool isTouched(Address const& _address) const { auto a = account(_address); return !a || a->isDirty(); }
+	void untouch(Address const& _address) { m_cache[_address].untouch(); }
```
gumb0: I think we can also add _address to m_unchangedCacheEntries here

chfast:
All comments addressed. The free function `revert()` stayed as internal because it is recursively called.

libethereum/Executive.h
```diff
@@ -78,6 +78,21 @@ class StandardTrace
 	DebugOptions m_options;
 };
 
+
+struct AccountSnapshot
```
gumb0: Maybe this is AccountChanges then? and m_orig is m_changes?

libethereum/Executive.cpp
```diff
@@ -459,7 +480,64 @@ void Executive::finalize()
...
+			// The the account was alive before CREATE (prefund) we have to
+			// reset some params. This is not very precise but should work in
+			// real live networks where we don't anticipate hash collisions.
+			_state.setNonce(_changes.address, 0);
+			_state.setCode(_changes.address, {});
+			_state.clearStorageChanges(_changes.address);
+		} else
+			// If the account was not existing before we can safely kill it.
+			_state.kill(_changes.address);
+	} else
```
chriseth: else on next line

libethereum/Executive.h
```diff
@@ -200,6 +218,8 @@ class Executive
 
 	bigint m_gasCost;
 	SealEngineFace const& m_sealEngine;
+
+	AccountSnapshot m_orig;
```
chriseth: Please document

libethereum/Executive.h
```diff
@@ -172,24 +187,27 @@ class Executive
...
 	/// @returns true iff the operation ended with a VM exception.
 	bool excepted() const { return m_excepted != TransactionException::None; }
 
 	/// Collect execution results in the result storage provided.
 	void setResultRecipient(ExecutionResult& _res) { m_res = &_res; }
 
+	/// Revert all changes made to the state by this execution.
+	void revert();
+
+	AccountSnapshot takeSnapshot() { return std::move(m_orig); }
```
chriseth: Please document. Is `m_orig` usable after calling `takeSnapshot()`?

libethereum/Executive.h
```diff
@@ -78,6 +78,21 @@ class StandardTrace
 	DebugOptions m_options;
 };
 
+
+struct AccountSnapshot
+{
+	bool exists = false;
```
gumb0: And this could be existedBefore for example

libethereum/ExtVM.cpp
```diff
@@ -95,30 +94,62 @@ void go(unsigned _depth, Executive& _e, OnOpFunc const& _onOp)
...
 		e.accrueSubState(sub);
 	}
 	_p.gas = e.gas();
 
-	return !e.excepted();
+	if (e.excepted())
+		return false;
+
+	// FIXME: Make sure move constructor is really used.
+	m_orig.children.emplace_back(e.takeSnapshot());
```
chriseth: The name of `takeSnapshot` would suggest to store the current state of the sub-call, but I think it stores the previous state, right? Perhaps we should call it something like `reversionLog`?
Also I think `m_orig`, `takeSnapshot` and `AccountSnapshot` should all have similar names.

libethereum/Executive.h
```diff
@@ -78,6 +78,21 @@ class StandardTrace
...
+	bool isCreation = false;
+	int nonceInc = 0;
+	Address address;
```
gumb0: Also it seems somewhat confusing that address here serves two purposes - could be value receiver, could be created address. I suggest at least a comment here about that

The same for caller member

libethereum/ExtVM.cpp
```diff
@@ -95,30 +94,62 @@ void go(unsigned _depth, Executive& _e, OnOpFunc const& _onOp)
...
+	return true;
 }
 
 size_t ExtVM::codeSizeAt(dev::Address _a)
 {
 	return m_s.codeSize(_a);
 }
 
+void ExtVM::setStore(u256 _n, u256 _v)
```
chriseth: Perhaps `setStorage`? And please also change the names of the local variables to something understandable. Let's not repeat the mistakes that were done in the past...

libethereum/ExtVM.cpp
```diff
@@ -95,30 +94,62 @@ void go(unsigned _depth, Executive& _e, OnOpFunc const& _onOp)
...
+	m_orig.children.emplace_back(e.takeSnapshot());
 	return e.newAddress();
 }
+
+void ExtVM::suicide(Address _a)
+{
+	// FIXME: What if _a is 0?
+	if (!m_s.isTouched(_a))
+		m_orig.selfdestructBeneficiary = _a;
+	// TODO: Why transfer is no used here?
```
chriseth: Please just use it.

libethereum/ExtVM.cpp
```diff
@@ -95,30 +94,62 @@ void go(unsigned _depth, Executive& _e, OnOpFunc const& _onOp)
...
 	io_gas = e.gas();
+	if (!e.newAddress())
+		return {};
+	m_orig.children.emplace_back(e.takeSnapshot());
 	return e.newAddress();
 }
+
+void ExtVM::suicide(Address _a)
+{
+	// FIXME: What if _a is 0?
```
chriseth: Does it make a difference?

libethereum/ExtVM.cpp
gumb0: It is setting one value in storage, I'd say it's setStorageValue

libethereum/Executive.h
chfast: The `AccountSnapshot` should be merged with the `Account` later on (at least tried to be merged). But renaming it to "AccountChanges" make no sense -- it can contain some account data, in out case it is used to keep "original" data of a changed account.

libethereum/Executive.h
chfast: Actually it is the same purpose -- this is the address of currently processing account (it is newly created address for CREATE, it receives sent ethers, etc). The caller it the message sender.

libethereum/ExtVM.cpp
chfast: Now not, old comment.

libethereum/ExtVM.cpp
chfast: This is just moved implementation from the header file. I will not change the name in this PR.

libethereum/ExtVM.cpp
chfast: I like "reversion log", but actually AccountSnapshot does not care what it is used for. Anyway, I will try to do something about it.

libethereum/ExtVM.cpp
chfast: Using it breaks some tests :).

Issue #3367, opened 2016-10-21
Implement journal


chriseth:
It has to be easy to revert changes to the state during a call. Currently, this is realized by making a full copy of the transaction state changes (stored in `State::m_cache`) at each call, which obviously does not scale similarly as the gas costs.

The following strategy would scale better (also implemented by geth):

Only keep a single instance of `State::m_cache` across calls and instead, store a log of changes that is replayed in reverse for a revert.

More specifically, keep a vector of changes to the state (a change is any call to State or Account that performs a modification that would result in an action in the `commit` function) and store what would have to be done during a revert of that change. Instead of copying back the saved copy of `m_cache` during the `revert` function, replay the log in reverse and apply it to `m_cache`.

This scales better than the previous solution if one accepts that the cost for a revert is already paid by the modifying opcode (i.e. half of the cost goes to the effect, half goes to the revert).

PR #3420, opened 2016-11-22
Cleanups pre State Journal

chfast:
A pack of non-functional changes in preparation for State Journal in #3419.

See also #3367.

### DIFF
--- a/libethereum/ExtVM.h
+++ b/libethereum/ExtVM.h
@@ -42,7 +42,7 @@ class ExtVM: public ExtVMFace
 {
 public:
 	/// Full constructor.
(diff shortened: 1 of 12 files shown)


libethereum/ExtVM.h
```diff
@@ -69,12 +69,6 @@ class ExtVM: public ExtVMFace
 	/// Read address's balance.
 	virtual u256 balance(Address _a) override final { return m_s.balance(_a); }
 
-	/// Subtract amount from account's balance.
-	virtual void subBalance(u256 _a) override final { m_s.subBalance(myAddress, _a); }
```
chriseth: Are these unused?

chfast: Not used. For `subBalance` it is always used withing `transferBalance`.

Issue #3156, opened 2016-08-26
Enable static builds of binaries


bobsummerwill:
_From @rainbeam on April 30, 2016 22:41_

Following on from ethereum/webthree-helpers#157, enable static linking of some more binaries by switching `add_executable` to `eth_simple_add_executable`.

This is a master issue. Target binaries:
- [x] `eth` 
  - [x] wrap `add_exe` ethereum/webthree#163
- [x] `solc` ethereum/solidity#528
- [x] `soltest` ^
- [x] `lllc` ^
  - [x] fix macro arguments ethereum/webthree-helpers#164
- [ ] `ethminer` 
  - [x] wrap `add_exe` ethereum/libethereum#248
  - [ ] macro changed.   dynamic working.   static still has `-lcurl` ordering issue
- [x] Fix accidental REQUIRED introduced onto curl for dynamic builds (https://github.com/ethereum/webthree-helpers/pull/165)
- [ ] Usage of `STATIC_LINKING` across platforms
  - [x] Test static linkage on Alpine Linux
  - [ ] Test static linkage on Debian (working, you said?) (no, curl link problems)
  - [ ] Test static linkage on Ubuntu
  - [x] Test static linkage on OS X (partial static-linkage working).
    - [ ] Suppress some [linker warnings](http://stackoverflow.com/questions/4929255/building-static-libraries-on-mac-using-cmake-and-gcc)
  - [ ] Test static linkage on Windows

_Copied from original issue: ethereum/webthree-umbrella#495_

bobsummerwill:
_From @rainbeam on April 30, 2016 23:54_

Just running a local build now to sanity check the above build with static linking.

bobsummerwill:
_From @rainbeam on May 1, 2016 0:20_

Everything works except `ethminer` which fails on linking with a curl problem:

```
[100%] Linking CXX executable ethminer
/src/built/lib/libjsonrpccpp-client.a(httpclient.cpp.o): In function `jsonrpc::HttpClient::HttpClient(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
httpclient.cpp:(.text+0x14d): undefined reference to `curl_easy_init'
/src/built/lib/libjsonrpccpp-client.a(httpclient.cpp.o): In function `jsonrpc::HttpClient::~HttpClient()':
httpclient.cpp:(.text+0x1e4): undefined reference to `curl_easy_cleanup'
[...]

PR #3429, opened 2016-11-25
Support static builds.

chriseth:


chriseth:
@chfast shall I close this one or does it provide any value?

chfast:
We can merge it, if you want to continue later. What are the current issues?

chriseth:
The current issues are that the binary does not find a library. We can continue later. The main idea was not a static library but just a smaller docker image.

axic:
Can we have this option in Solidity too?

chriseth:
Strangely, solidity works without this option: https://hub.docker.com/r/ethereum/solc/

chfast:
@chriseth Please specify the requirements in a separated issue. I can take it from there (I usually like build tweaks).

chriseth:
@chfast I think we can just continue here: https://github.com/ethereum/cpp-ethereum/issues/3156

PR #3437, opened 2016-12-02
Ropsten genesis config fix

gumb0:


libethereum/State.cpp
```diff
@@ -332,10 +332,10 @@ u256 State::balance(Address const& _id) const
 void State::noteSending(Address const& _id)
 {
 	Account* a = account(_id);
-	if (asserts(a != nullptr))
```
chriseth: Did that cause an exception before? Please check where it was caught and increase the logging there.

PR #3445, opened 2016-12-06
libethereum: mitigate `testeth` crash in Mac OS

pirapira:
This PR tries to solve the test failure in Mac OS X.

### DIFF
--- a/libethereum/ExtVM.cpp
+++ b/libethereum/ExtVM.cpp
@@ -33,7 +33,7 @@ namespace // anonymous
 static unsigned const c_depthLimit = 1024;
 
 /// Upper bound of stack space needed by single CALL/CREATE execution. Set experimentally.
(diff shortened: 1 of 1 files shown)


pirapira:
Now trying to reduce the size while keeping the tests clean.

libethereum/ExtVM.cpp
```diff
@@ -33,7 +33,7 @@ namespace // anonymous
 static unsigned const c_depthLimit = 1024;
 
 /// Upper bound of stack space needed by single CALL/CREATE execution. Set experimentally.
-static size_t const c_singleExecutionStackSize = 11 * 1024;
+static size_t const c_singleExecutionStackSize = 14 * 1024;
```
gcolvin: Whatever it takes.

PR #3446, opened 2016-12-06
Disable account creation if there's none

gumb0:


chriseth:
I think "signingKey" can be completely removed.

gumb0:
The only effect I see so far after removing account generation is the output like this:
```
Transaction Signer: XE50000000000000000000000000000000 (00000000-0000-0000-0000-000000000000 - 00000000)
Mining Beneficiary: XE50000000000000000000000000000000 (00000000-0000-0000-0000-000000000000 - 00000000)
Foundation: XE55PXQKKK4B9BYPBGT1XCYW6R5ELFAT6EM (00000000-0000-0000-0000-000000000000 - de0b2956)
```

chriseth:
Yes, I would also remove the line starting with `Foundation` and the one starting with `Transaction Signer` - there should be no default transaction signer.

eth/main.cpp
```diff
@@ -962,10 +959,7 @@ int main(int argc, char** argv)
 	for (auto const& s: passwordsToNote)
 		keyManager.notePassword(s);
 
-	writeFile(configFile, rlpList(signingKey, author));
-
-	if (sessionKey)
-		signingKey = sessionKey;
+	writeFile(configFile, rlpList(author, author));
```
gumb0: Not sure about this, probably we can't change the format of this file easily? Or do we need it at all

gumb0: Anyway this author is being read in the beginning from this file if written previously

And then passed to eth::Client::setAuthor (which sets it to some "preSeal" block)

gumb0: Ok, that's what I did here

gumb0:
Removed --import-session-secret, --sign-key, --session-sign-key options,

PR #3453, opened 2016-12-08
Fix possible exploit related to account cache updating

gumb0:
Improve the logic for selecting the random element to remove from cache:
generate random element index instead of generating the random address and
finding the closest one in cache


chfast:
Very nice indeed.

PR #3456, opened 2016-12-09
ouput transaction RLP info when caught an exception

winsvega:
more info on invalid transaction

### DIFF
--- a/libethcore/Transaction.cpp
+++ b/libethcore/Transaction.cpp
@@ -26,6 +26,7 @@
 #include <libevmcore/EVMSchedule.h>
 #include <libethcore/Exceptions.h>
 #include "Transaction.h"
+
 using namespace std;
 using namespace dev;
 using namespace dev::eth;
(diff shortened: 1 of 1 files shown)


libethcore/Transaction.cpp
```diff
@@ -89,7 +90,7 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
 	}
 	catch (Exception& _e)
 	{
-		_e << errinfo_name("invalid transaction format") << BadFieldError(field, toHex(rlp[field].data().toBytes()));
```
chriseth: Can you please also add a comment to line 77/78 that contains the value of 'v'?

winsvega: exception on that line would be caught by line 92

winsvega: why? toString(rlp)  will output whole transaction fields including v

PR #3457, opened 2016-12-10
test: workaround for crash in Mac

pirapira:
The crash was observed at: https://travis-ci.org/ethereum/cpp-ethereum/jobs/182600890


test/TestHelper.cpp
```diff
@@ -486,10 +486,10 @@ int ImportTest::compareStates(State const& _stateExpect, State const& _statePost
 					TestOutputHelper::testName() + "Check State: " << a.first <<  ": incorrect storage [" << s.first << "] = " << toHex(stateStorage[s.first]) << ", expected [" << s.first << "] = " << toHex(s.second));
 
 				//Check for unexpected storage values
-				stateStorage = _stateExpect.storage(a.first);
```
pirapira: The crash happened in this assignment.  I need to dig this further.  For the time being, I just define a new variable.

pirapira:
Now I see no test failures in Mac.

Issue #3459, opened 2016-12-11
Windows 7 build error


bnrmaster:
I met an error：
c:\users\administrator\cpp-ethereum\libweb3jsonrpc\UnixSocketServer.h(28): fatal error C1083: can not open include file: “sys/un.h”: No such file or directory

when trying to build the "web3jsonrpc" project using VS 2015 on Window 7

Can anyone give me some advice？

chriseth:
I just created a pull request that might fix your problem.

bnrmaster:
It works,thx for your answer! @chriseth

PR #3460, opened 2016-12-12
Fix includes for windows.

chriseth:
Fixes https://github.com/ethereum/cpp-ethereum/issues/3459

### DIFF
--- a/libweb3jsonrpc/UnixSocketServer.cpp
+++ b/libweb3jsonrpc/UnixSocketServer.cpp
@@ -19,7 +19,7 @@ along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
 *   Arkadiy Paronyan <arkadiy@ethdev.com>
 * @date 2015
 */
-#ifndef WIN32
+#if !defined(_WIN32)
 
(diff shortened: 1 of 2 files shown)


libweb3jsonrpc/UnixSocketServer.cpp
```diff
@@ -19,7 +19,7 @@ along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
 *   Arkadiy Paronyan <arkadiy@ethdev.com>
 * @date 2015
 */
-#ifndef WIN32
+#if !defined(_WIN32)
```
chfast: When checking compiler and build flags, I'm suggesting using `#if _WIN32`. When not defined the macro is substituted by 0.

chfast: When it is defined it is defined to 1. So I think it is a bit better to expect it to be 1 than `#if defined(_WIN32)` to be true for `_WIN32` set to 0.

chriseth: In all other places, we use the `defined(_WIN32)` style.

Issue #3376, opened 2016-10-25
Optimization: Cache information about non-existing accounts


chriseth:
The cache inside state should store information about non-existing accounts. Currently, it only stores information about existing accounts, so access to a non-existing acocunt always has to go to the database.

chfast:
I would assign -1 for not-existing accounts.

chriseth:
Not sure about the type information there. But perhaps we can still combine that with the missing protection against attacks when clearing m_cache, i.e. extend `m_unchangedCacheEntries` by a data structure that maps from address to 
 - this is a non-existing account
 - this is an account that was not modified
and allow this datastructure to be purged more efficiently.

PR #3461, opened 2016-12-12
Cache the addresses of accessed non-existing accounts

gumb0:
See #3376 
This optimization should help with the attacks that repeatedly access non-existing account addresses

This PR is based on #3419 by @chfast and includes commits from it

### DIFF
--- a/libethereum/State.cpp
+++ b/libethereum/State.cpp
@@ -64,6 +64,7 @@ State::State(State const& _s):
 	m_state(&m_db, _s.m_state.root(), Verification::Skip),
 	m_cache(_s.m_cache),
 	m_unchangedCacheEntries(_s.m_unchangedCacheEntries),
+	m_nonExistingAccountsCache(_s.m_nonExistingAccountsCache),
 	m_touched(_s.m_touched),
 	m_accountStartNonce(_s.m_accountStartNonce)
 {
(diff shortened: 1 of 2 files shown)


libethereum/State.cpp
```diff
@@ -367,17 +376,23 @@ void State::subBalance(Address const& _id, bigint const& _amount)
...
 }
 
 void State::ensureAccountExists(const Address& _address)
 {
 	if (!addressInUse(_address))
-		m_cache[_address] = Account(requireAccountStartNonce(), 0, Account::NormalCreation);
+		createInCache(_address, Account(requireAccountStartNonce(), 0, Account::NormalCreation));
+}
+
+void State::createInCache(Address const& _address, Account const&& _account)
```
chriseth: It will be created on disk in the end, so perhaps just call that `createAccount`?

libethereum/State.cpp
```diff
@@ -367,17 +376,23 @@ void State::subBalance(Address const& _id, bigint const& _amount)
 
 void State::createContract(Address const& _address, bool _incrementNonce)
 {
-	m_cache[_address] = Account(
+	createInCache(_address, Account(
 		requireAccountStartNonce() + (_incrementNonce ? 1 : 0),
 		balance(_address),
-		Account::ContractConception
+		Account::ContractConception)
```
chriseth: Closing parentheses go to the next line.

gumb0:
Rebased it on current develop

PR #3462, opened 2016-12-12
libethashseal split genesis

winsvega:
split genesis info into separate files


libethashseal/CMakeLists.txt
```diff
@@ -2,6 +2,17 @@ aux_source_directory(. SRC_LIST)
 
 file(GLOB HEADERS "*.h")
 
+macro (add_sources)
```
chriseth: Why do we need a special macro for that?

winsvega:
I've just copied this way from libethereum/test folder. but it can be done like that without macro. (review)

chriseth:
Wonderful, thanks!

PR #3463, opened 2016-12-13
remove inconsistency in genesis config

winsvega:
need review for other modules (vm, eth)


libethashseal/GenesisInfo.h
```diff
@@ -34,10 +34,10 @@ namespace eth
 enum class Network
 {
 	//Olympic = 0,			///< Normal Olympic chain.
-	Frontier = 1,			///< Normal Frontier/Homestead chain.
-	//Morden = 2,				///< Normal Morden chain.
+	MainNetwork = 1,		///< Normal Frontier/Homestead/DAO/EIP150/EIP158 chain.
+	//Morden = 2,			///< Normal Morden chain.
 	Ropsten = 3,			///< New Ropsten Test Network
-	Test = 70,				///< Normal Frontier/Homestead chain without all the premine.
+	TransitionnetTest = 70,	///< Normal Frontier/Homestead/DAO/EIP150/EIP158 chain without all the premine.
```
chfast: It would be nice if the test network configs could have been defined separately, somewhere in test/. But this is not request for change, just comment for future considerations.

winsvega: to do it this way we might need to add a compilation flag to libethashseal so it would include files from cpp-ethereum/test

PR #3466, opened 2016-12-16
Implementation of EthereumPeer::requestNodeData & EthereumPeer::requestReceipts

gumb0:
Supporting a calls to PV63 methods which are going to be used in Fast Sync.

PR #3470, opened 2016-12-21
Session & EthereumPeer refactoring & unit tests for EthereumPeer

gumb0:
Refactoring trying to achieve better testability of EthereumPeer class.
Currently work in progress.

This PR is based on #3466 and first commit is from there.

### DIFF
--- a/libethereum/EthereumHost.cpp
+++ b/libethereum/EthereumHost.cpp
@@ -49,6 +49,63 @@ const char* EthereumHostTrace::name() { return EthPurple "^" EthGray "  "; }
 const char* EthereumHostTrace::name() { return EthPurple "⧫" EthGray " "; }
 #endif
 
+namespace
+{
+class EthereumPeerObserver : public EthereumPeerObserverFace
+{
+public:
+	explicit EthereumPeerObserver(EthereumHost& _host) : m_host(_host) {}
+
+	void onPeerStatus(std::shared_ptr<EthereumPeer> _peer) override
+	{
+		m_host.onPeerStatus(_peer);
+	}
+
+	void onPeerTransactions(std::shared_ptr<EthereumPeer> _peer, RLP const& _r) override
+	{
+		m_host.onPeerTransactions(_peer, _r);
+	}
+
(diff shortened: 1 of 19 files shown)


libp2p/Session.h
```diff
@@ -158,11 +193,25 @@ class Session: public std::enable_shared_from_this<Session>
...
 	std::shared_ptr<Framing> getFraming(uint16_t _protocolID);
 	void multiplexAll();
 };
 
+template <class PeerCap>
+std::shared_ptr<PeerCap> capabilityFromSession(SessionFace const& _session, u256 const& _version = PeerCap::version())
+{ 
+	try 
+	{ 
+		return std::static_pointer_cast<PeerCap>(_session.capabilities().at(std::make_pair(PeerCap::name(), _version)));
```
chriseth: This is only used in networking code, right? So we could spend the extra CPU cycles and use `dynamic_pointer_cast` for added safety. What do you think?

gumb0: That would basically check that our initialization code correctly registers Capability classes with correct names. Maybe some debug-only check would be enough? Like adding `assert(dynamic_pointer_cast<PeerCap>(cap))`

The result value returned from this function is almost never checked for nullptr anyway

libethereum/EthereumHost.cpp
```diff
@@ -49,14 +49,278 @@ const char* EthereumHostTrace::name() { return EthPurple "^" EthGray "  "; }
...
+private:
+	EthereumHost& m_host;
+};
+
+class EthereumHostData: public EthereumHostDataFace
+{
+public:
+	EthereumHostData(BlockChain const& _chain, OverlayDB const& _db): m_chain(_chain), m_db(_db) {}
+
+	pair<bytes, unsigned> blockHeaders(RLP const& _blockId, unsigned _maxHeaders, u256 _skip, bool _reverse) const override
```
gumb0: This code is moved here from `EthereumPeer::interpret()`

gumb0:
Rebased on latest develop

PR #3471, opened 2016-12-22
EVM-C update

chfast:
This mostly updates the EVM-C interface. Changes there comes mostly from lesson learned from integrating EVMJIT with geth. The C API is more explicit to avoid some C ABI complex rules like returning or passing structs by value.

Please do not merge. I want to do it only when EVMJIT part is accepted.


chfast:
One more develop merge to be sure...

PR #3473, opened 2016-12-27
Fix data race in concurrent queues

chfast:
These concurrent queues could have been better implemented (keeping the both conditions under single lock), but I don't have time to that right now.

Fixes https://github.com/ethereum/cpp-ethereum/issues/3182.


winsvega:
we've tried to make those variables atomic - didn't help

chfast:
This at least satisfies the thread sanitizer.

PR #3474, opened 2016-12-27
Travis CI: install latest cmake

chfast:


scripts/install_cmake.sh
```diff
@@ -0,0 +1,16 @@
...
+mkdir -p build
+cd build
+
+OS_NAME=$(uname -s)
+CMAKE_NAME=cmake-3.7.1-$OS_NAME-x86_64
+CMAKE_BIN=$(pwd)/$CMAKE_NAME/bin
+
+if ! $CMAKE_BIN/cmake --version; then
+echo "Downloading $CMAKE_NAME"
+curl -s https://cmake.org/files/v3.7/$CMAKE_NAME.tar.gz | tar xz
```
chriseth: Can we add a shasum check here? Ah and please add quotes around the variable.

chfast: Ok, will try :)

chfast:
This works on Linux, but not on MacOS, probably because the homebrew update happens after it and overwrites the PATH. However, upgrading cmake on MacOS is not critical because homebrew delivers quite recent one.

chfast:
@chriseth how it looks now?

chriseth:
@chfast build error on macos: https://travis-ci.org/ethereum/cpp-ethereum/jobs/188673314

chfast:
Ok, I refactored it. I'm installing cmake to ~/.local and Travis is picking it up automatically. I also disabled it for MacOS as it has not effect, but the script should work on MacOS as well.

scripts/install_cmake.sh
```diff
@@ -0,0 +1,37 @@
...
+
+BIN=$PREFIX/bin
+
+if test -f $BIN/cmake && ($BIN/cmake --version | grep -q "$VERSION"); then
+    echo "CMake $VERSION already installed in $BIN"
+else
+    FILE=cmake-$VERSION-$OS-x86_64.tar.gz
+    URL=https://cmake.org/files/v3.7/$FILE
+    ERROR=0
+    cd /tmp
```
chriseth: Could you do that in a subshell - I don't  like scripts modifying the working directory.

Oh and perhaps `mktemp -d` would be useful, just in case the downloaded file already exists in `/tmp/`. Ah and the other benefit of `mktemp` is that in respects a different `$TMP`.

chfast: It is in called in a subshell, but how to do it here?

chfast:
Ready?

PR #3479, opened 2017-01-04
initial (disabled) implementation of jumps and subroutines

gcolvin:
This implements the new opcodes for jumps and subroutines.  It does not implement validation.  Opcodes are disabled by default, and will halt with bad instruction.


gcolvin:
The current error is:
```C:/projects/cpp-ethereum/test/TestHelper.cpp(581): error: in "StateTestsGeneral/stRevertTest": chfastRevertPrefoundEmptyOOG Expected another postState hash! EIP158 data: 0 gas: 0 val: 0```
I have no idea what it means.  It showed up and went away again on Linux earlier.
```/home/travis/build/ethereum/cpp-ethereum/test/TestHelper.cpp(581): error in "stRevertTest": chfastRevertPrefoundEmptyOOG Expected another postState hash! EIP158 data: 0 gas: 0 val: 0``` 

I don't think it has anything to do with the code I just added, which isn't even running.

@chfast @winsvega

File: libevm/VM.cpp [hunk 1]
```diff
@@ -815,7 +815,93 @@ void VM::interpretCases()
 			m_sp -= 2;
 		CASE_END
 
+		CASE_BEGIN(JUMPTO)
+		THROW_IF_SUBS_DISABLED();
+			ON_OP();
+			updateIOGas();
+			{
+				uint64_t dest       = m_code[m_pc++];
```
chriseth: I guess the new specification still assumes that bytecode is suffixed with an infinite number of zeros. Does the access here still work with our "enlarge the code just by a bunch of bytes" fix?

libevm/VM.cpp
```diff
@@ -815,7 +815,93 @@ void VM::interpretCases()
...
+				m_pc =    dest;
+			}
+		CASE_END
+
+		CASE_BEGIN(JUMPIF)
+		THROW_IF_SUBS_DISABLED();
+			ON_OP();
+			updateIOGas();
+			if (*m_sp--) {
+				uint64_t dest       = m_code[m_pc++];
```
chriseth: Can we use a big endian decoding helper function here and in all the other copies of this code? :-)

libevm/VM.cpp
```diff
@@ -815,7 +815,93 @@ void VM::interpretCases()
...
+				dest =  (dest << 8) | m_code[m_pc++];
+				m_pc =    dest;
+			}
+		CASE_END
+
+		CASE_BEGIN(JUMPIF)
+		THROW_IF_SUBS_DISABLED();
+			ON_OP();
+			updateIOGas();
+			if (*m_sp--) {
```
chriseth: Braces to the next line.

libevm/VM.cpp
```diff
@@ -815,7 +815,93 @@ void VM::interpretCases()
...
+				uint64_t dest       = m_code[m_pc++];
+				dest =  (dest << 8) | m_code[m_pc++];
+				dest =  (dest << 8) | m_code[m_pc++];
+				dest =  (dest << 8) | m_code[m_pc++];
+				m_pc =   dest;
+			} else
```
chriseth: Else to the next line.

libevm/VM.cpp
```diff
@@ -815,7 +815,93 @@ void VM::interpretCases()
...
+				++m_pc;
+			--m_sp;
```
chriseth: I think this should not be here.

libevm/VM.cpp
```diff
@@ -815,7 +815,93 @@ void VM::interpretCases()
...
+			}
+		CASE_END
+
+		CASE_BEGIN(JUMPSUBV)
+		THROW_IF_SUBS_DISABLED();
+			ON_OP();
+			updateIOGas();
+			{
+				*++m_rp = m_pc;
+				byte n = *(byte*)m_pc;
```
chriseth: Can this be correct?

libevm/VM.cpp
```diff
@@ -815,7 +815,93 @@ void VM::interpretCases()
...
+				uint64_t i = uint64_t(*m_sp--);
+				if (i >= n) i = n - 1;
+				i = 1 + i * 4;
```
chriseth: Please either add some comments about how jumpsubv works or use variable names that are not single letters :-)

libevm/VM.h
```diff
@@ -101,13 +103,18 @@ class VM: public VMFace
 	u256s m_stackSpace;
 	u256* m_stack;
 	
+	// space for return stack and pointer to data
+	std::vector<uint64_t> m_returnSpace;
+	uint64_t* m_return;
```
chriseth: Please initialize.

libevm/VMOpt.cpp
```diff
@@ -132,8 +75,8 @@ void VM::optimize()
 				
 		// make synthetic ops in user code trigger invalid instruction if run
 		if (op == Instruction::PUSHC ||
```
chriseth: The correct style here would be (replacing four spaces by tabs)
```
if (
    op == Instruction::PUSHC ||
    op == Instruction::JUMPC ||
    op == Instruction::JUMPCI
)
{
```

libevmcore/Instruction.cpp
```diff
@@ -164,152 +173,160 @@ const std::map<std::string, Instruction> dev::eth::c_instructions =
...
+	{ Instruction::GASPRICE,     { "GASPRICE",       0,     0,    1,  false,       BaseTier } },
+	{ Instruction::EXTCODESIZE,  { "EXTCODESIZE",    0,     1,    1,  false,       SpecialTier } },
+	{ Instruction::EXTCODECOPY,  { "EXTCODECOPY",    0,     4,    0,  true,        SpecialTier } },
+	{ Instruction::BLOCKHASH,    { "BLOCKHASH",      0,     1,    1,  false,       ExtTier } },
+	{ Instruction::COINBASE,     { "COINBASE",       0,     0,    1,  false,       BaseTier } },
+	{ Instruction::TIMESTAMP,    { "TIMESTAMP",      0,     0,    1,  false,       BaseTier } },
+	{ Instruction::NUMBER,       { "NUMBER",         0,     0,    1,  false,       BaseTier } },
+	{ Instruction::DIFFICULTY,   { "DIFFICULTY",     0,     0,    1,  false,       BaseTier } },
+	{ Instruction::GASLIMIT,     { "GASLIMIT",       0,     0,    1,  false,       BaseTier } },
+	{ Instruction::JUMPTO,       { "JUMPTO",         0,     1,    0,  true,        VeryLowTier } },
```
chriseth: This should have an "add" value of 4, because it has a 4-byte immediate argument, similar for other opcodes.

libevm/VM.cpp
gcolvin: We enlarge it enough for a full PUSH32, so it should.

libevm/VM.cpp
gcolvin: Good idea, unless we change to a different encoding.  @Arachnid didn't like this one.  But a helper is still a good idea.

libevm/VM.cpp
gcolvin: OK ;->

libevm/VM.cpp
gcolvin: I think I messed up only one brace and one else this time.

libevm/VM.h
gcolvin: Already done in constructor.

libevmcore/Instruction.cpp
gcolvin: Gotcha.

libevm/VM.cpp
gcolvin: I don't think so, and I think I made the same mistake in all of the JUMP cases.

gcolvin:
Code for handling JUMPDEST, BEGINSUB, and BEGINDATA scanning is in place.  Everything compiles, but the EVM_JUMPS_AND_SUBS macro is turned off for this merge.

gcolvin:
@chriseth @chfast  I think this is ready to merge.

chriseth:
@chfast could you please review this?

libevm/VM.cpp
```diff
@@ -815,7 +848,68 @@ void VM::interpretCases()
 			m_sp -= 2;
 		CASE_END
 
+#ifdef EVM_JUMPS_AND_SUBS
+		CASE_BEGIN(JUMPTO)
+			ON_OP();
+			updateIOGas();
+			m_pc = decode_jump_dest(m_code, m_pc);
```
chfast: Have you handled out of bound code access issue somewhere else (like with push data)?

libevm/VM.cpp
```diff
@@ -815,7 +848,68 @@ void VM::interpretCases()
...
+		CASE_END
+
+		CASE_BEGIN(JUMPIF)
+			ON_OP();
+			updateIOGas();
+			if (*m_sp)
+			{
```
chfast: Style: no {

libevm/VM.cpp
```diff
@@ -815,7 +848,68 @@ void VM::interpretCases()
...
+			ON_OP();
+			updateIOGas();			
+			m_pc = decode_jumpv_dest(m_code, m_pc, m_sp);
+		CASE_END
+
+		CASE_BEGIN(JUMPSUB)
+			ON_OP();
+			updateIOGas();
+			{
+				*++m_rp = m_pc++;
```
chfast: To be honest these _pointer tricks_ annoy me a bit. Can't we make it more explicit using vectors, push_backs, etc?

libevm/VM.h
```diff
@@ -65,7 +65,9 @@ class VM: public VMFace
 	bytes const& memory() const { return m_mem; }
 	u256s stack() const { assert(m_stack <= m_sp + 1); return u256s(m_stack, m_sp + 1); };
 
-	VM(): m_stackSpace(1025), m_stack(m_stackSpace.data() + 1), m_pool(256) {};
+	VM():
+		m_stackSpace(1025), m_stack(m_stackSpace.data() + 1),
+		m_returnSpace(1025), m_return(m_returnSpace.data() + 1) {};
```
chfast: Style: `{};` to the next line.

libevm/VMOpt.cpp
```diff
@@ -61,153 +61,140 @@ void VM::copyCode()
...
+				 (byte)op <= (byte)Instruction::PUSH32)
 		{
-			i += (byte)op - (byte)Instruction::PUSH1 + 1;
+			pc += (byte)op - (byte)Instruction::PUSH1 + 1;
 		}
-		
+#ifdef EVM_JUMPS_AND_SUBS
+		else if (op == Instruction::JUMPTO || op == Instruction::JUMPIF || op == Instruction::JUMPSUB)
+		{
+			pc += 4 + 1;
```
chfast: How about `1 + 4` here?

libevm/VMOpt.cpp
```diff
@@ -61,153 +61,140 @@ void VM::copyCode()
...
+		}
+		else if (op == Instruction::JUMPV || op == Instruction::JUMPSUBV)
+		{
+			++pc;
+			pc += 4 * m_code[pc];  // number of 4-byte dests followed by table
```
chfast: And here (to be consitent with previous one) `pc += 1 + 4 * m_code[pc]`?

libevm/VMOpt.cpp
```diff
@@ -61,153 +61,140 @@ void VM::copyCode()
...
 			// add value to constant pool and replace PUSHn with PUSHC if room
 			if (1 < nPush)
 			{
 				TRACE_PRE_OPT(1, i, op);
-				int pool_off = poolConstant(val);
-				if (0 <= pool_off && pool_off < 256)
+				byte h = ((hash >> 8) ^ hash) & 0xff;
+				if (m_pool[h] == 0)
+					m_pool[h] = val;
+				if (m_pool[h] == val)
```
chfast: In general, I'm lost in this constant pool section. Hope it works :)

libevm/VM.cpp
```diff
@@ -44,6 +44,40 @@ template <class S> S modWorkaround(S const& _a, S const& _b)
 
 
 //
+// for decoding destinations of JUMPTO, JUMPV, JUMPSUB and JUMPSUBV
+//
+
+uint64_t VM::decode_jump_dest(const byte* const _code, uint64_t& _pc)
```
chfast: Style: camelCase and `byte const* const`.

libevm/VM.cpp
```diff
@@ -44,6 +44,40 @@ template <class S> S modWorkaround(S const& _a, S const& _b)
...
+{
+	// turn 4 MSB-first bytes in the code into a native-order integer
```
chfast: Can you move general comments about the function above the function definition (Doxygen style)?

libevm/VM.cpp
```diff
@@ -44,6 +44,40 @@ template <class S> S modWorkaround(S const& _a, S const& _b)
...
+	uint64_t dest      = _code[_pc++];
+	dest = (dest << 8) | _code[_pc++];
+	dest = (dest << 8) | _code[_pc++];
+	dest = (dest << 8) | _code[_pc++];
+	return dest;
+}
+
+uint64_t VM::decode_jumpv_dest(const byte* const _code, uint64_t& _pc, u256*& _sp)
```
chfast: Style: same as the previous function.
Also, `u256*&` is a bit scary. Can you add a comment about how it behave?

libevm/VMOpt.cpp
gcolvin: I went with consistently using ++m_PC to skip over the bytecode.

libevm/VMOpt.cpp
gcolvin: It passes tests.  I added comments, as it's really pretty simple.

libevm/VM.cpp
gcolvin: That would be the comment on line 47.

libevm/VM.cpp
gcolvin: I'm not exactly sure how it works anymore either, and it is scary.  But both the PC and the SP get modified by a jump, so I need some way to do that.

libevm/VM.cpp
gcolvin: I think the  push data is enough to cover this case.

libevm/VM.h
gcolvin: If you insist  ;->

libevm/VM.cpp
gcolvin: For now have moved the extension of the code array to after the jumpdest/beginsub scan, so we know how many extra bytes there are.

libevm/VM.h
chriseth: I still think it is better to put a `=nullptr` here in case someone creates a new constructor and forgets to initialize that member.

libevmcore/Instruction.cpp
```diff
@@ -164,152 +173,160 @@ const std::map<std::string, Instruction> dev::eth::c_instructions =
...
+	{ Instruction::EXTCODECOPY,  { "EXTCODECOPY",    0,     4,    0,  true,        SpecialTier } },
+	{ Instruction::BLOCKHASH,    { "BLOCKHASH",      0,     1,    1,  false,       ExtTier } },
+	{ Instruction::COINBASE,     { "COINBASE",       0,     0,    1,  false,       BaseTier } },
+	{ Instruction::TIMESTAMP,    { "TIMESTAMP",      0,     0,    1,  false,       BaseTier } },
+	{ Instruction::NUMBER,       { "NUMBER",         0,     0,    1,  false,       BaseTier } },
+	{ Instruction::DIFFICULTY,   { "DIFFICULTY",     0,     0,    1,  false,       BaseTier } },
+	{ Instruction::GASLIMIT,     { "GASLIMIT",       0,     0,    1,  false,       BaseTier } },
+	{ Instruction::JUMPTO,       { "JUMPTO",         4,     1,    0,  true,        VeryLowTier } },
+	{ Instruction::JUMPIF,       { "JUMPIF",         4,     2,    0,  true,        LowTier } },
+	{ Instruction::JUMPV,        { "JUMPV",          4,     1,    0,  true,        MidTier } },
```
chriseth: I guess the `4` here is not accurate since the immediate data of that opcode is dynamic, is it?

libevm/VMOpt.cpp
```diff
@@ -49,175 +49,237 @@ void VM::initMetrics()
...
+			op == Instruction::JUMPIF ||
+			op == Instruction::JUMPSUB)
+		{
+			++pc;
+			pc += 4;
+		}
+		else if (op == Instruction::JUMPV || op == Instruction::JUMPSUBV)
+		{
+			++pc;
+			pc += 4 * m_ext->code[pc];  // number of 4-byte dests followed by table
```
chriseth: Just for added safety, can we use something like  `4 * (unsigned char) m_ext->code[pc]`?

libevm/VMOpt.cpp
```diff
@@ -49,175 +49,237 @@ void VM::initMetrics()
...
+			m_beginSubs.push_back(pc);
+		}
+		else if (op == Instruction::BEGINDATA)
+		{
+			break;
+		}
+#endif
 	}
+	
+	copyCode(pc - nBytes);
```
chriseth: What happens here if there is a JUMPV that has 255 destinations (but the code does not actually contain those) and pc goes off the code due to that?

libevm/VM.h
gcolvin: There is only a default constructor.  I don't think you can inherit from this class without the derived constructor calling the inherited default constructor.  But I'll look at it after I have clean tests again.

gcolvin: I intend to rewrite this anyway.

libevmcore/Instruction.cpp
gcolvin: RIght.  The 4 probably is a copy-paste error, and JUMPSUBV has the same problem.  The minimum would be 5. I hadn't considered how much trouble the variable-length data might cause

libevm/VMOpt.cpp
gcolvin: Probably a SEGV?  Though the code will probably have already failed validation.  But for this and other reasons I think we may need a more robust encoding.

gcolvin: Actually, a PUSH code without enough data would have the same problem.  And worse - the copyCode() doesn't copy the data section at the end - what was I thinking?  But it can't, because it might all be zeros and all be gone.  So BEGINDATA needs a length argument.  And in the meantime it's back to nBytes+33.

gcolvin:
@chfast @chriseth  I think we should merge this.  It could be better, but I've tried to answer most of your points, and this PR is getting way too old.  I think everything that could matter to correctness is safely behind macros, and can be addressed going forward.  I just want to get a first pass at implementing our EIP merged.

libevm/VM.h
gcolvin: Major cleanup.

libevm/VMOpt.cpp
gcolvin: Simple, but wrong.  Broke it out into a class to clarify things.

libevm/VMOpt.cpp
chfast: I think it is fine. It will use `int` type for this expression. We also have AddressSanitizer to help.

libevm/VMOpt.cpp
```diff
@@ -49,175 +49,237 @@ void VM::initMetrics()
 	done = true;
 }
 
-void VM::copyCode()
+void VM::copyCode(int _extraBytes)
```
chfast: Please use `size_t` to match the type of `size()` later on.

libevm/VM.h
```diff
@@ -77,7 +75,7 @@ class VM: public VMFace
 	static std::array<InstructionMetric, 256> c_metrics;
 	static void initMetrics();
 	static u256 exp256(u256 _base, u256 _exponent);
-	void copyCode();
+	void copyCode(int);
```
chfast: Add param name here.

PR #3482, opened 2017-01-04
Remove OpenCL support

chfast:


ethminer/MinerAux.h
```diff
@@ -150,35 +145,6 @@ class MinerCLI
...
-			try {
-				m_msPerBatch = stol(argv[++i]);
-			}
-			catch (...)
-			{
-				cerr << "Bad " << arg << " option: " << argv[i] << endl;
-				BOOST_THROW_EXCEPTION(BadArgument());
-			}
-#endif // ETH_ETHASHCL
 		else if (arg == "--list-devices")
```
chriseth: Shouldn't we remove this, too?

chfast: Probably yes. We don't want to "support" old flags, do we?

chriseth: I think if we are concerned about compatibility, we should be more concerned about compatibility with geth than with older versions of eth.

PR #3483, opened 2017-01-04
Fix state journal

chfast:
When reverting changes of a message also untouch accounts touched by CREATE.

### DIFF
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -515,6 +515,8 @@ void revertAccountChanges(State& _state, AccountRevertLog const& _changes)
 			_state.setNonce(_changes.address, 0);
 			_state.setCode(_changes.address, {});
 			_state.clearStorageChanges(_changes.address);
+			// Also untouch it because it classify for being cleared any more.
+			_state.untouch(_changes.address);
 		}
 		else
 			// If the account was not existing before we can safely kill it.


libethereum/Executive.cpp
```diff
@@ -515,6 +515,8 @@ void revertAccountChanges(State& _state, AccountRevertLog const& _changes)
 			_state.setNonce(_changes.address, 0);
 			_state.setCode(_changes.address, {});
 			_state.clearStorageChanges(_changes.address);
+			// Also untouch it because it classify for being cleared any more.
+			_state.untouch(_changes.address);
```
gumb0: What happened if it stayed touched? Why did affect only EIP158 ?

chfast: Yes, only EIP158. In EIP158 when an account is "touched" and empty it must be removed at the end of transaction.

gumb0: I see.

Maybe we can here just move the `if (!_changes.existed) _state.untouch(_changes.address)` from lines 531-532 to outside of `if (_changes.isCreation)` (to line 534)

Then it's not repeated

gumb0: Oh no, there're opposite conditions, sorry

winsvega:
hm. seems it have fixed general state test chfastRevert.
still strange why that test didn't worked on my local cpp

PR #3487, opened 2017-01-09
ExternalProject Crypto++: fix lib location on some GNU/Linux systems

chfast:


abueide:
builds correctly and fixes my issue with compiling master on arch linux

PR #3488, opened 2017-01-09
New version of debug_storageRangeAt

gumb0:
Modifying storageRangeAt according to comments in https://github.com/ethereum/go-ethereum/pull/3407

### DIFF
--- a/libdevcore/TrieDB.h
+++ b/libdevcore/TrieDB.h
@@ -452,15 +452,14 @@ public:
 
 	void remove(bytesConstRef _key) { Super::remove(sha3(_key)); }
 
-	//friend class iterator;
-
-	class iterator : public GenericTrieDB<_DB>::iterator
+	// iterates over <key, value> pairs
+	class iterator: public GenericTrieDB<_DB>::iterator
 	{
 	public:
 		using Super = typename GenericTrieDB<_DB>::iterator;
 
 		iterator() { }
-		iterator(FatGenericTrieDB const* _trie): Super(_trie) { }
+		iterator(FatGenericTrieDB const* _trie) : Super(_trie) { }
 
 		typename Super::value_type at() const
 		{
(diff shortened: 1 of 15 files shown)


libdevcore/TrieDB.h
```diff
@@ -462,16 +462,13 @@ class FatGenericTrieDB: private SpecificTrieDB<GenericTrieDB<_DB>, h256>
 		iterator() { }
 		iterator(FatGenericTrieDB const* _trie): Super(_trie) { }
 
-		typename Super::value_type at() const
+		bytes key() const
```
gumb0: This iterator is used only in State to get all the storage of account I think

gumb0: I was not right, it is used in State also to have a map of addresses to storage and we need to iterate easily over it getting addresses (not hashes)

I will better add a second type of iterator here

gumb0:
Now output of storageRangeAt is similar to this:

```js
{ complete: false,
  storage:
   [ { hashedKey: '0x0e0686cf556ee1885c6e6349eaa494a545c6580591ca4ed5ddee6c7d924ec595',
       key: '0x1577871241893300e556b8798934ee70cfe051b0943db23eea9a5e2e918eaa50',
       value: '0x3439623565383666616265323136366200000000000000000000000000000000' },
     { hashedKey: '0x2277b2835c91a9c163a4555d97a9028bf04d4b9f3d0f9a3cfdedcd9eef9200ca',
       key: '0x027e3a08d18c1d70aa5909473f76aded3f86a903bf98709f94fa056cc662651a',
       value: '0x012707' },
     { hashedKey: '0x301828f1b7d481a5a8c7a60602555780c11f4219f81eef7b61197c35af82eb31',
       key: '0x657c1a534f9c7ed61ba46924b6eb083a0c74da2ad2d90af08ce6217f1893ddf2',
       value: '0x01531e' } ] }
```

PR #3491, opened 2017-01-10
Update install_deps.sh script: remove cryptopp

chfast:


chfast:
install_deps.cmake is alread updated: https://github.com/ethereum/cpp-dependencies/blob/f0289e2f0da04566661c30cbf6a4af1ac4b388e0/install_deps.cmake

chriseth:
hm, it still has `opencl-icd` there. But merging this first.

chfast:
Ah, right... but this was about cryptopp :) I will do another round for OpenCL :)

PR #3499, opened 2017-01-13
State Journal v2

chfast:
I added some description in the State.h header. If that's is not enough I'm open for suggestions how to improve it.

~28 failures are expected in state test -- the tests are about account address collision (CREATE overwriting existing account). The behavior in this case is not defined. Updated tests are ready, but we don't want to merge them before this PR not to break other PRs and develop.

### DIFF
--- a/libethereum/Account.cpp
+++ b/libethereum/Account.cpp
@@ -82,19 +82,19 @@ AccountMap dev::eth::jsonToAccountMap(std::string const& _json, u256 const& _def
 
 			if (haveCode)
 			{
-				ret[a] = Account(nonce, balance, Account::ContractConception);
+				ret[a] = Account(nonce, balance);
 				if (o["code"].type() == json_spirit::str_type)
 				{
 					if (o["code"].get_str().find("0x") != 0)
 						cerr << "Error importing code of account " << a << "! Code needs to be hex bytecode prefixed by \"0x\".";
 					else
-						ret[a].setCode(fromHex(o["code"].get_str().substr(2)));
+						ret[a].setNewCode(fromHex(o["code"].get_str().substr(2)));
 				}
 				else
 					cerr << "Error importing code of account " << a << "! Code field needs to be a string";
 			}
 			else
-				ret[a] = Account(nonce, balance, Account::NormalCreation);
+				ret[a] = Account(nonce, balance);
 
 			if (haveStorage)
 				for (pair<string, js::mValue> const& j: o["storage"].get_obj())
@@ -105,7 +105,7 @@ AccountMap dev::eth::jsonToAccountMap(std::string const& _json, u256 const& _def
 		{
 			(*o_mask)[a] = AccountMask(haveBalance, haveNonce, haveCode, haveStorage, shouldNotExists);
 			if (!haveStorage && !haveCode && !haveNonce && !haveBalance && shouldNotExists) //defined only shouldNotExists field
-				ret[a] = Account(0, 0, Account::NormalCreation);
+				ret[a] = Account(0, 0);
 		}
 
 		if (o_precompiled && o.count("precompiled"))
--- a/libethereum/Account.cpp
+++ b/libethereum/Account.cpp
@@ -21,17 +21,22 @@
 
 #include "Account.h"
 #include <json_spirit/JsonSpiritHeaders.h>
-#include <libethcore/Common.h>
 #include <libethcore/ChainOperationParams.h>
 #include <libethcore/Precompiled.h>
+
 using namespace std;
 using namespace dev;
 using namespace dev::eth;
+
+void Account::setNewCode(bytes&& _code)
+{
+	// FIXME: Make code hash lazy (calculated when first needed)?
+	m_codeCache = std::move(_code);
+	m_hasNewCode = true;
+	m_codeHash = sha3(m_codeCache);
+}
+
 namespace js = json_spirit;
-
-#pragma GCC diagnostic ignored "-Wunused-variable"
-
-const h256 Account::c_contractConceptionCodeHash;
 
 uint64_t toUnsigned(js::mValue const& _v)
 {
--- a/libethereum/Account.cpp
+++ b/libethereum/Account.cpp
@@ -30,7 +30,6 @@ using namespace dev::eth;
 
 void Account::setNewCode(bytes&& _code)
 {
-	// FIXME: Make code hash lazy (calculated when first needed)?
 	m_codeCache = std::move(_code);
 	m_hasNewCode = true;
 	m_codeHash = sha3(m_codeCache);
--- a/libethereum/Account.h
+++ b/libethereum/Account.h
@@ -144,9 +144,6 @@ public:
 	/// @returns the storage overlay as a simple hash map.
 	std::unordered_map<u256, u256> const& storageOverlay() const { return m_storageOverlay; }
 
-	/// Clear all changes made to the storage since last commit.
-	void clearStorageChanges() { m_storageOverlay.clear(); }
-
 	/// Set a key/value pair in the account's storage. This actually goes into the overlay, for committing
 	/// to the trie later.
 	void setStorage(u256 _p, u256 _v) { m_storageOverlay[_p] = _v; changed(); }
--- a/libethereum/Account.h
+++ b/libethereum/Account.h
@@ -105,9 +105,6 @@ public:
 	/// during creation phase.
 	bool isEmpty() const { return nonce() == 0 && balance() == 0 && codeHash() == EmptySHA3; }
 
-	/// @returns the balance of this account. Can be altered in place.
-	u256& balance() { return m_balance; }
-
 	/// @returns the balance of this account.
 	u256 const& balance() const { return m_balance; }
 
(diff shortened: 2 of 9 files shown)


libethereum/Executive.cpp
```diff
@@ -255,14 +255,16 @@ bool Executive::call(Address _receiveAddress, Address _senderAddress, u256 _valu
...
 {
-	// Always remember the sender, needed for revert.
-	m_revertLog.caller = _p.senderAddress;
-
 	// If external transaction.
 	if (m_t)
 		// Increment associated nonce for sender.
 		m_s.incNonce(_p.senderAddress);
 
+	m_savepoint = m_s.savepoint();
```
chriseth: Usual db terminology uses "checkpoint"

libethereum/Executive.cpp
```diff
@@ -486,63 +483,48 @@ void Executive::finalize()
...
-
-		if (!_changes.existed)
-			_state.untouch(_changes.address);
-	}
-
-	if (_changes.selfdestructBeneficiary)
-		_state.untouch(_changes.selfdestructBeneficiary);
-}
-}
+//namespace
```
chriseth: Remove if finished

libethereum/State.cpp
```diff
@@ -378,6 +407,14 @@ u256 State::storage(Address const& _id, u256 const& _key) const
 		return 0;
 }
 
+void State::setStorage(Address const& _contract, u256 const& _location,
+                       u256 const& _value)
```
chriseth: Style: indentation

libethereum/State.cpp
```diff
@@ -378,6 +407,14 @@ u256 State::storage(Address const& _id, u256 const& _key) const
...
+{
+	Change change = {ChangeKind::storage, _contract, _location, storage(_contract, _location), 0};
+	m_changeLog.push_back(std::move(change));
```
chriseth: Why not emplace_back?

libethereum/State.cpp
```diff
@@ -499,6 +536,50 @@ bool State::isTrieGood(bool _enforceRefs, bool _requireNoLeftOvers) const
...
+			untouch(change.address);
+			break;
+		case ChangeKind::create:
+//			std::cerr << "Revert Create " << change.address << "\n";
+			kill(change.address);
+			break;
+		case ChangeKind::prefound_create:
+//			std::cerr << "Revert prefound " << change.address << "\n";
+			setCode(change.address, {});
+			untouch(change.address);  // FIXME: add prefound and CREATE revert in single transaction.
```
chriseth: What if account was pre-funded and then created in the same transaction? Is it fine to untouch it here?

libethereum/State.h
```diff
@@ -98,6 +98,25 @@ DEV_SIMPLE_EXCEPTION(IncorrectAccountStartNonceInState);
...
 class SealEngineFace;
 
+enum class ChangeKind
+{
+	balance,
+	storage,
+	nonce,
+	touched,
+	create,
+	prefound_create
```
chriseth: `prefunded_create`?

libethereum/State.h
```diff
@@ -98,6 +98,25 @@ DEV_SIMPLE_EXCEPTION(IncorrectAccountStartNonceInState);
 
 class SealEngineFace;
 
+enum class ChangeKind
+{
+	balance,
```
chriseth: I think these are capitalized in all other places, but I might be wrong.

libethereum/State.h
```diff
@@ -263,6 +278,10 @@ class State
 	u256 const& requireAccountStartNonce() const;
 	void noteAccountStartNonce(u256 const& _actual);
 
+	using ChangeLog = std::vector<Change>;
+	size_t savepoint();
```
chriseth: const?

libethereum/State.h
```diff
@@ -263,6 +278,10 @@ class State
 	u256 const& requireAccountStartNonce() const;
 	void noteAccountStartNonce(u256 const& _actual);
 
+	using ChangeLog = std::vector<Change>;
```
chriseth: Does it need to be public? Can it be closer to where it is used?

gcolvin:
@chfast I was just beginning to study this code with an eye to integrating validation, but don't understand it much yet.  Could you explain a bit about how it works and how you are changing it?  I suspect that conceptually it is much simpler than the C++.

libethereum/State.h
```diff
@@ -98,6 +98,40 @@ DEV_SIMPLE_EXCEPTION(IncorrectAccountStartNonceInState);
...
 
+
+
+namespace detail
+{
+struct Change
+{
+	enum Kind: int
+	{
+		balance,
```
gumb0: Coding standards 3.b Enum members start with uppercase letter

gumb0: Also 6.j Prefer enum class to straight enum.

libethereum/Executive.cpp
chriseth: Ok, I give in :-)

libethereum/State.h
chfast: Changed to uppercase, but I would like to keep typed enum, because this avoids another namespace nesting. In the end, it is a class internal to State API.

libethereum/State.cpp
```diff
@@ -530,23 +539,22 @@ void State::revert(size_t _savepoint)
...
-		case Change::create:
-			account.kill();
-			break;
-		case Change::prefund_create:
-			// FIXME: add prefound and CREATE revert in single transaction.
+		case Change::Create:
+			// Clear the deployed code (if any) and mark as not dirty.
+			// FIXME: probably wrong for accounts prefunded in the same
+			//        transaction. Separate Create and SetCode events.
 			account.setCode({});
```
gumb0: Previously it was account.kill() for Change::Create
So where do we kill the account now?

chfast: Kill will not be needed (it was used to clear the account -- set all elements to null). But I'm working on the change to separate true account create (there was no account, now it is) from the code deployment... Should be ready today.

chfast:
@gcolvin I extended the docs a bit... (see State.h).

libethereum/Account.h
```diff
@@ -152,29 +140,24 @@ class Account
...
+	/// @returns the hash of the account's code.
+	h256 codeHash() const { return m_codeHash; }
 
-	/// @returns true if we are either in the contract-conception state or if the account's code is not
-	/// empty.
-	bool codeBearing() const { return m_codeHash != EmptySHA3; }
-
-	/// @returns the hash of the account's code. Must only be called when isFreshCode() returns false.
-	h256 codeHash() const { assert(!isFreshCode()); return m_codeHash; }
+	bool isFreshCode() const { return m_hasNewCode; }
```
chriseth: Previously, this function returned true even during creation. Now it only returns true after code has been deployed.

chfast: Yes. You cannot tell now if an account is in "creation" phase. But that information should not be needed.

For the Changelog we need accounts modification to be more atomic, not packed into transnational changes. So you can not add code to any existing account -- the API will allow you to even replace the existing code with a new one... From my experience in this area, it is easier to maintain it by doing set of independent changes (e.g. create empty account, add balance, increase nonce, set code) than having API restricted transactions (e.g. create an account with this nonce, this balance, and allow setting code later).

chfast:
Ready for final review. Read description.

libethereum/Executive.cpp
```diff
@@ -255,13 +256,16 @@ bool Executive::call(Address _receiveAddress, Address _senderAddress, u256 _valu
 
 bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address const& _origin)
 {
-	// Always remember the sender, needed for revert.
-	m_revertLog.caller = _p.senderAddress;
-
 	// If external transaction.
 	if (m_t)
+	{
+		// FIXME: some unrevertable changes occupy change log.
```
chriseth: Is this still an issue?

libethereum/Executive.cpp
```diff
@@ -283,64 +287,48 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
 		m_gas = _p.gas;
 		if (m_s.addressHasCode(_p.codeAddress))
```
chriseth: I'm not sure if that changes anything, but this function will now return true during the creation phase of `_p.codeAddress`, while it returned false previously.

libethereum/State.h
```diff
@@ -98,10 +98,68 @@ DEV_SIMPLE_EXCEPTION(IncorrectAccountStartNonceInState);
...
+		Create,
+
+		/// New code was added to an account (by "create" message execution).
+		NewCode,
+
+		/// Account was touched for the first time.
+		Touch
+	};
+
+	Kind kind;        ///< The kind of the change.
```
chriseth: This struct is 84 bytes long, does it make sense to shorten it for "cheap" operations?

libethereum/Account.cpp
```diff
@@ -21,17 +21,22 @@
...
+
 using namespace std;
 using namespace dev;
 using namespace dev::eth;
-namespace js = json_spirit;
 
-#pragma GCC diagnostic ignored "-Wunused-variable"
+void Account::setNewCode(bytes&& _code)
+{
+	// FIXME: Make code hash lazy (calculated when first needed)?
```
chriseth: I don't think this is necessary. We have to compute the hash unless the creation is reverted.

libethereum/State.h
chfast: I considered this. To achieve this we would need to use some polymorphic interfaces what means heap allocation and additional dereference (not CPU cache friendly). I can benchmark this, but as a separated task.

What we can do at low cost is to use `h256` (32 bytes) instead of `u256` (40 bytes).

libethereum/Executive.cpp
chfast: I checked what's there exactly -- it is balance change that paid for the transaction.

chfast:
Any pending requests here?

chriseth:
Please merge and synchronize with merging test changes.

PR #3501, opened 2017-01-16
Fix crash when connecting to node with older eth protocol version

gumb0:


libethereum/EthereumHost.cpp
```diff
@@ -640,7 +640,9 @@ shared_ptr<Capability> EthereumHost::newPeerCapability(shared_ptr<SessionFace> c
 {
 	auto ret = HostCapability<EthereumPeer>::newPeerCapability(_s, _idOffset, _cap, _capID);
 
-	capabilityFromSession<EthereumPeer>(*_s)->init(
+	auto cap = capabilityFromSession<EthereumPeer>(*_s, _cap.second);
```
gumb0: We requested here the new capability just registered on the previous line, but forgot to specify its version.

libethereum/EthereumHost.cpp
```diff
@@ -640,7 +640,9 @@ shared_ptr<Capability> EthereumHost::newPeerCapability(shared_ptr<SessionFace> c
 {
 	auto ret = HostCapability<EthereumPeer>::newPeerCapability(_s, _idOffset, _cap, _capID);
 
-	capabilityFromSession<EthereumPeer>(*_s)->init(
```
gumb0: We requested here the new capability just registered on the previous line, but forgot to specify its version.

libethereum/EthereumHost.cpp
```diff
@@ -640,7 +640,9 @@ shared_ptr<Capability> EthereumHost::newPeerCapability(shared_ptr<SessionFace> c
...
+	auto cap = capabilityFromSession<EthereumPeer>(*_s, _cap.second);
+	assert(cap);
```
chriseth: So if the version does not match, an exception is thrown here, right? Can you check that this also happens in release mode?

PR #3502, opened 2017-01-17
add array filters for expect section in general state tests

winsvega:
this option is used when 
./testeth -t StateTestsGeneral/stTestCase --filltests --checkstate -d 0 -g 0 -v 1

to set specific array values to check in expected results. 
cause in general tests there could be like tens of tests (because of the different transaction fields) and sometime it is needed to produce post state check just for a few of them

### DIFF
--- a/test/TestHelper.cpp
+++ b/test/TestHelper.cpp
@@ -567,6 +567,10 @@ void ImportTest::checkGeneralTestSection(json_spirit::mObject const& _expects, v
 			string trInfo = netIdToString(t.netId) + " data: " + toString(t.dataInd) + " gas: " + toString(t.gasInd) + " val: " + toString(t.valInd);
 			if (_expects.count("result"))
 			{
+				Options const& opt = Options::get();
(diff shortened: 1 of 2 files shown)


test/TestHelper.cpp
```diff
@@ -566,6 +566,10 @@ void ImportTest::checkGeneralTestSection(json_spirit::mObject const& _expects, v
 			string trInfo = netIdToString(t.netId) + " data: " + toString(t.dataInd) + " gas: " + toString(t.gasInd) + " val: " + toString(t.valInd);
 			if (_expects.count("result"))
 			{
+				Options const& opt = Options::get();
+				if ((opt.test_d != -1 && opt.test_d != t.dataInd) || (opt.test_g != -1 && opt.test_g != t.gasInd) || (opt.test_v != -1 && opt.test_v != t.valInd))
```
chfast: Please, split this line, it's too long.

test/TestHelper.h
```diff
@@ -241,6 +241,9 @@ class Options
 	std::string singleTestFile;
 	std::string singleTestName;
 	std::string singleTestNet;
+	int test_d; ///< GeneralState data
```
chfast: Use longer name, `testData` or `data`...

chfast:
@winsvega are you going to address my comments?

PR #3507, opened 2017-01-17
libevmcore, libevm: turn enum Tier into an enum class

pirapira:
Just turning `enum Tier` into `enum class Tier`.

I made a similar change in Solidity, and learned that `cpp-ethereum` has the same code: https://github.com/ethereum/solidity/pull/1562


libevmcore/Instruction.h
```diff
@@ -228,7 +228,7 @@ inline Instruction logInstruction(unsigned _number)
 	return Instruction(unsigned(Instruction::LOG0) + _number);
 }
 
-enum Tier
+enum class Tier : unsigned
```
chfast: Can you also rename the elements by removing the `Tier` suffix?

pirapira: That's a good idea.

gcolvin:
#3479 is merged, so you are good to go @pirapira, except that there are conflicts now, as I feared.

pirapira:
I solved conflicts and took in @chfast 's suggestion.

chfast:
LGTM

PR #3510, opened 2017-01-18
EVM-C v3

chfast:
This implements EVM-C v3 on the cpp-ethereum side, specified https://github.com/ethereum/evmjit/issues/110.

EVMJIT part: https://github.com/ethereum/evmjit/pull/112.


axic:
I'm not really familiar with cpp-ethereum internals so cannot comment much on this.

libevm/JitVM.cpp
```diff
@@ -21,9 +21,9 @@ inline evm_uint160be toEvmC(Address _addr)
 	return *reinterpret_cast<evm_uint160be*>(&_addr);
 }
 
-inline Address fromEvmC(evm_uint160be _addr)
+inline Address fromEvmC(evm_uint160be const* _addr)
```
gumb0: Could we pass the const reference here (and in `asUint`) and reinterpret_cast the reference?
I think it'd be cleaner to dereference the pointer on the calling side, the thing like `asUint(ptr)` gave me a microsecond of confusion, it's not immediately clear that the function converts the value pointed to and not the pointer

libevm/JitVM.cpp
```diff
@@ -34,110 +34,65 @@ inline evm_uint256be toEvmC(h256 _h)
...
 		break;
 	}
-	case EVM_CALL_DEPTH:
-		o_result->int64 = env.depth;
-		break;
-	}
 }
 
-void evm_update(
+void updateState(
```
gumb0: Why do you rename this but not `evm_query`?
and also `evm_call` & `evm_get_tx_context` are snake_case, while `updateState` & `getBlockHash` are camelCase

libevm/JitVM.cpp
chfast: Yes, that make sense. I recently have noticed they look bad. I will also try to use `std::copy` instead of `reinterpret_cast` if proven to generate the same code.

libevm/JitVM.cpp
chfast: With this PR I renamed some changed function to match our naming convention. I will do the same for the rest.

libevm/JitVM.cpp
```diff
@@ -153,34 +108,48 @@ void evm_update(
...
+	auto value = asUint(&_msg->value);
+	bytesConstRef input{_msg->input, _msg->input_size};
 
-	if (_kind == EVM_CREATE)
+	if (_msg->kind == EVM_CREATE)
 	{
 		assert(_outputSize == 20);
-		u256 gas = _gas;
+		u256 gas = _msg->gas;
+		// TODO: How it knows who the sender is?
```
gumb0: for `create` ExtVM gets it from `myAddress` member initialized in constructor

libevm/JitVM.cpp
```diff
@@ -16,129 +16,83 @@ static_assert(sizeof(Address) == sizeof(evm_uint160be),
 static_assert(alignof(Address) == alignof(evm_uint160be),
               "Address types alignment mismatch");
 
-inline evm_uint160be toEvmC(Address _addr)
+inline evm_uint160be toEvmC(Address const& _addr)
 {
-	return *reinterpret_cast<evm_uint160be*>(&_addr);
+	return *reinterpret_cast<evm_uint160be const*>(&_addr);
```
gumb0: Can we do just `return reinterpret_cast<evm_uint160be const&>(_addr);` ?

libevm/JitVM.cpp
```diff
@@ -16,129 +16,83 @@ static_assert(sizeof(Address) == sizeof(evm_uint160be),
...
+	(void) _addr;
 	auto &env = *reinterpret_cast<ExtVMFace*>(_opaqueEnv);
+	assert(fromEvmC(*_addr) == env.myAddress);
 	switch (_key)
 	{
 	case EVM_SSTORE:
 	{
-		auto index = asUint(_arg1->uint256be);
-		auto value = asUint(_arg2->uint256be);
+		auto index = fromEvmC(_arg1->uint256be);
```
gumb0: `auto` & `fromEvmC` combination is a little confusing here, it's not clear to what we convert

libevm/JitVM.cpp
```diff
@@ -190,12 +158,12 @@ int64_t evm_call(
...
-	params.gas = _gas;
-	params.apparentValue = _kind == EVM_DELEGATECALL ? env.value : value;
-	params.valueTransfer = _kind == EVM_DELEGATECALL ? 0 : params.apparentValue;
-	params.senderAddress = _kind == EVM_DELEGATECALL ? env.caller : env.myAddress;
-	params.codeAddress = fromEvmC(*_address);
-	params.receiveAddress = _kind == EVM_CALL ? params.codeAddress : env.myAddress;
+	params.gas = _msg->gas;
+	params.apparentValue = value;
+	params.valueTransfer = _msg->kind == EVM_DELEGATECALL ? 0 : params.apparentValue;
+	params.senderAddress = fromEvmC(_msg->sender);
```
gumb0: Why did we get rid of conditional here?

gumb0:
Windows build fails on `std::copy(addr.begin(), addr.end(), _outputData)`

chfast:
Fixed. Fancy array iterator MSVC has :)

gumb0:
GCC is not happy now :)

gumb0:
Maybe something just like `std::copy(addr.data(), addr.data() + Address::size, _outputData)` ?

chfast:
I will fix it :)

libdevcore/FixedHash.h
```diff
@@ -141,6 +141,12 @@ class FixedHash
 	/// @returns a constant byte pointer to the object's data.
 	byte const* data() const { return m_data.data(); }
 
+	/// @returns begin iterator.
+	auto begin() const -> typename std::array<byte, N>::const_iterator { return m_data.begin(); }
```
gumb0: Shouldn't we call these cbegin & cend ?

chfast: I don't think some, `cbegin()` is only useful when you have a non-const object, with `begin() const` and `begin()` and you want to enforce getting const iterator.

PR #3514, opened 2017-01-20
libethcore: SealEngine chooses the schedule based on the parameters set in the RPC

pirapira:
This is needed to test https://github.com/ethereum/solidity/pull/1577


pirapira:
@chriseth   Now I have duplicate `Ethash::evmSchedule()` and `SealEngine::evmSchedule()`.  Which one shall I keep?  I still don't have the big picture.

chriseth:
I guess the cleanest solution would be to implement the function in `SealEngineBase` and remove all other implementations (including the one in SealEngineFace).

pirapira:
@chriseth that doesn't work.  `Ethash` inherits `SealEngineFace` directly.

chriseth:
Does it make sense to inherit from SealEngineBase instead? :)

pirapira:
That does make sense.  I'm treading carefully just because I do not know the current design choices.

PR #3519, opened 2017-01-24
script for running all tests on azureVM

winsvega:


scripts/runalltests.sh
```diff
@@ -0,0 +1,37 @@
+workdir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/testslog
```
chriseth: Please add `#!/usr/bin/env bash` as a first line.

chriseth: I usually use `workdir="$(dirname "$0")/testslog"` - why would you use the extra `pwd` step?

scripts/runalltests.sh
```diff
@@ -0,0 +1,37 @@
...
+testdir=$workdir/../../../tests
```
chriseth: Put `$workdir` in quotes everywhere.

scripts/runalltests.sh
```diff
@@ -0,0 +1,37 @@
...
+
+if [ -d $workdir ]; then
```
chriseth: Ah, now I see. If you only use workdir to store a single temporary file, I think using something like
`logfile="$(mktemp)"` is much easier, better portable and does not break if you run it multiple times.

scripts/runalltests.sh
```diff
@@ -0,0 +1,37 @@
...
+fi
+mkdir $workdir
+exec &> $workdir/testlog.txt
+cd $testdir
+
+echo "Fetching the test repository:"
+git pull
+echo "Test Repository last commit hash:"
+git rev-parse HEAD
+cd ..
```
chriseth: So this script somehow assumes that it is run from within a different repository? What if we change this to just use `git clone github.com/...` instead of relying on being part of a certain repository that is already checked out? If we clone it (we can do it with `--depth 1` to save bandwidth and time), we can always run it everywhere and even on a different server if needed, as long as it supports sending mail.
If we do that, then we would need
`workdir="$(mktemp -d)"` - which creates a new temporary directory.

scripts/runalltests.sh
```diff
@@ -0,0 +1,37 @@
...
+cd cpp-ethereum
+echo "Fetching the cpp-ethereum develop:"
+git pull
+echo "cpp-ethereum last commit hash:"
+git rev-parse HEAD
+cd build
+echo "Make cpp-ethereum develop:"
+make -j8
+
+exec &>> $workdir/testlog.txt
```
chriseth: What is the purpose of this?

scripts/runalltests.sh
winsvega: this line is from stackoverflow

scripts/runalltests.sh
winsvega: originally it was intended to store multiple log files and an archive of this files

scripts/runalltests.sh
winsvega: so you suggest to clone the repos on a fly? 
but keep this script as a part of cpp-ethereum. 
as it is now script should update itself by calling git pull

scripts/runalltests.sh
winsvega: print stderror and stdout  to the log file

PR #3521, opened 2017-01-26
Tests - add some new tests, big cleanup of old ones

gcolvin:
This looks bigger and scarier than it is.  The many deletions are mostly .bin and .evm files that were just clutter.  I added a few new tests, added the version pragma to the old ones, added some macros to VM.cpp to guard some horrible hacks for making the VM pretend that it could do some native 64-bit operations, not meter gas, not check stack...


gcolvin:
@chriseth @chriseth @wanderer @axic 

The numbers below are from the new  `mul*` series of tests here.  The C version (and original Solidity version) have hundreds of successive `x *= CONSTANT` statements in their main loop.  The Solidity version now has inline assembly to replace those all with `DUP2 MUL`.
* `mul0.sol` is the main loop, empty
* `mul64.sol` is the main loop with multiplies, where `x` does not exceed `2**64 - 1`
* `mul256.sol` is the main loop with mulltiplies, where `x` does not exceed `2**256 - 1`

From the numbers below I gather that:

1) The overhead for the empty loop in `mul0` is minimal.
2) The gas-per-second gets worse by 22% as the arithmetic exceeds 64 bits
3) Execution time gets better by 37% when code irrelevant to actually interpreting is removed
4) Execution time gets better by 63% when native 64-bit operations are used
5) Execution time gets better by 50% when C code is used instead of EVM

So for the `mul64.sol` benchmark, a full-on EVM 1.5 with all overheads minimized or eliminated and 64-calculations approaching native speed we can hope see execution time go from 4.9 to 1.8 secs - a 63% improvement, and C is only twice as fast.

(I don't know that actual native arithmetic is needed for these results, as the boost library is not very good.  And with a better library narrow operations might not look so much better than wider ones, so the question of whether narrow operators are needed so as to assign cheaper gas prices is not answered.)

**Current EVM, using boost wide arithmetic**

$ ./ethvm test mul0.bin
gas used: 27500116
gas/sec: 1.009e+08
exec time: 0.272422

$ ./ethvm test mul64.bin
gas used: 804000116
gas/sec: 1.636e+08
exec time: 4.913561

$ ./ethvm test mul256.bin
gas used: 622500116
gas/sec: 1.275e+08
exec time: 4.882710

**All gas metering, stack checking, and tracing removed**

$ ./ethvm test mul64.bin
exec time: 3.075962

**MUL and DUP hacked to use native 64-bit operations**

$ ./ethvm test mul64.bin
exec time: 1.837255

**Coded in C**

$ time ./mul64c

real	0m0.924s
user	0m0.920s
sys	0m0.000s

_**Note: none of the compilations are optimized as both the Solidy and C compilers take the sequence of multiplications and turn it into one constant.**_

PR #3522, opened 2017-01-26
Fast sync - downloading headers

gumb0:
Work in progess


libdevcore/RangeMask.h
```diff
@@ -71,6 +71,16 @@ class RangeMask
 		return ret;
 	}
 
+	/// @returns a range containing at most _items smallest elements.
```
chriseth: `@returns a range that starts with the first element and ends @a _items later or at the first excluded element`?

libethereum/EthereumPeer.cpp
```diff
@@ -220,8 +220,13 @@ void EthereumPeer::tick()
 	auto s = session();
 	time_t  now = std::chrono::system_clock::to_time_t(chrono::system_clock::now());
 	if (s && (now - m_lastAsk > 10 && m_asking != Asking::Nothing))
+	{
 		// timeout
 		s->disconnect(PingTimeout);
+
+		if (m_observer)
+			m_observer->onPeerRequestTimeout(dynamic_pointer_cast<EthereumPeer>(shared_from_this()), m_asking);
```
chriseth: Why do you need a dynamic pointer cast here?

libethereum/FastSync.cpp
```diff
@@ -0,0 +1,164 @@
...
+	}
+	else
+		syncPeer(_peer);
+}
+
+void FastSync::syncPeer(std::shared_ptr<EthereumPeerFace> _peer)
+{
+	Guard guard(m_downloadingHeadersMutex);
+	BlockNumberRange const nextRange = m_headersToDownload.lowestRange(c_maxRequestHeaders);
+	if (nextRange.first == nextRange.second)
```
chriseth: Is there some kind of empty method?

libethereum/FastSync.cpp
```diff
@@ -0,0 +1,164 @@
...
+		// nothing left to download
+		// TODO add to the pool of free peers
+		return;
+
+	_peer->requestBlockHeaders(nextRange.first, nextRange.second - nextRange.first, 0, false);
```
chriseth: Also here, perhaps a length method?

libethereum/FastSync.cpp
```diff
@@ -0,0 +1,164 @@
...
+	}
+}
+
+
+FastSync::BlockNumberRangeMask FastSync::saveDownloadedHeaders(RLP const& _r)
+{
+	BlockNumberRangeMask downloadedRangeMask(allHeadersRange());
+	try
+	{
+		for (size_t i = 0; i < _r.itemCount(); ++i)
```
chriseth: Can you use a range-based for loop here?

libethereum/EthereumPeer.cpp
gumb0: Because enable_shared_from_this is in the base class (Capability), so shared_from_this() here gives us shared_ptr<Capabilty>
It is this way when handling all responses in this class

libethereum/FastSync.cpp
gumb0: Yes, it turns out I can, I'll change it

libethereum/FastSync.cpp
gumb0: Yes, changed

libethereum/FastSync.cpp
gumb0: No length method since this range is just an std::pair

libethereum/FastSync.cpp
```diff
@@ -0,0 +1,168 @@
...
+namespace
+{
+unsigned const c_maxRequestHeaders = 1024;
+}
+
+FastSync::FastSync()
+{
+}
+
+void FastSync::onPeerStatus(shared_ptr<EthereumPeerFace> _peer)
```
chfast: Do you agree passing share_ptr to these callbacks is bad idea? I tried to simplify it some time ago, but BlockChainSync takes ownership of some peers. Is this the case for FastSync?

gumb0: FastSync saves them only as weak_ptr's now.
BlockChainSync does the same, I don't see yet where it takes ownership...

libethereum/EthereumHost.h
```diff
@@ -72,7 +73,6 @@ class EthereumHost: public p2p::HostCapability<EthereumPeer>, Worker
 	void reset();
 
 	bool isSyncing() const;
-	bool isBanned(p2p::NodeID const& _id) const { return !!m_banned.count(_id); }
```
gumb0: This was not used

chfast:
Ad your planned changes for `BlockQueue`, maybe it should be changed into a generic `Validator` (or `Verifier`) that is capable of verifying different types of items...

dustpan28645 (approved):

Issue #3435, opened 2016-12-02
cpp-ethereum for debian


thehumans35:
Hi,
after following the instruction on http://ethdocs.org/en/latest/ethereum-clients/cpp-ethereum/building-from-source/linux.html

git clone --recursive https://github.com/ethereum/cpp-ethereum.git
cd cpp-ethereum
./scripts/install_deps.sh

I got on bash: 
W: Fehlschlag beim Holen von http://ppa.launchpad.net/ethereum/ethereum-dev/ubuntu/dists/jessie/main/binary-amd64/Packages  404  Not Found
W: Fehlschlag beim Holen von http://ppa.launchpad.net/ethereum/ethereum-dev/ubuntu/dists/jessie/main/binary-i386/Packages  404  Not Found
W: Fehlschlag beim Holen von http://ppa.launchpad.net/ethereum/ethereum/ubuntu/dists/jessie/main/binary-amd64/Packages  404  Not Found
W: Fehlschlag beim Holen von http://ppa.launchpad.net/ethereum/ethereum/ubuntu/dists/jessie/main/binary-i386/Packages  404  Not Found

Looking for ubuntu folder, but I am running debian on my notebook?
Any instructions?
Thank you.

E: Einige Indexdateien konnten nicht heruntergeladen werden. Sie wurden ignoriert oder alte an ihrer Stelle benutzt.

chriseth:
Sorry, we barely support ubuntu now. You could try to use the docker image, that should work best.

thehumans35:
Dear Chriseth,

thank you for your message, could you please send me a link for docker image.


Best regards,

Hakan

________________________________
Von: chriseth <notifications@github.com>
Gesendet: Freitag, 2. Dezember 2016 10:28:46
An: ethereum/cpp-ethereum
Cc: thehumans35; Author
Betreff: Re: [ethereum/cpp-ethereum] cpp-ethereum for debian (#3435)


Sorry, we barely support ubuntu now. You could try to use the docker image, that should work best.

-

chriseth:
Here are the instructions: http://www.ethdocs.org/en/latest/ethereum-clients/cpp-ethereum/installing-binaries/docker.html
It seems they do not reflect the latest changes, you need to replace "cpp-client" by "client-cpp".

krzyszp:
[...]

PR #3524, opened 2017-01-26
Improve install_deps.sh script

chfast:
1. I removed git and cmake from install_deps.sh script. They must be installed manually.
2. Switched to POSIX sh
3. Linux distribution detection tries to use `/etc/*-release` files, `lsb_release` is not available in docker images by default.
4. Adding `ethereum-ppa` is not needed, because we don't depend on system cryptopp.
5. This should fix Debian builds. https://github.com/ethereum/cpp-ethereum/issues/3435.

Tested on docker images of:
- Debian Jessie,
- Fedora 25,
- CentOS 7.


chfast:
Thanks for helping with this. I also moved LLVM-related packages to Travis-only section.

PR #3531, opened 2017-01-30
azure test script: make report at the begining

winsvega:


scripts/runalltests.sh
```diff
@@ -27,13 +27,20 @@ cd test
...
-exec &> /dev/null
-#cd $workdir
-#rm results.zip
-#zip results.zip *.txt
-date=$(date +%Y-%m-%d)
+#Prepare Header of the report
+exec &> $workdir/testreport.txt
+exectime=$(echo "$timeend - $timestart" | bc)
+echo "REPORT"
+echo "Test execution time: "$exectime
```
chriseth: Can you change this to `"Test execution time: $exectime s"`?

scripts/runalltests.sh
```diff
@@ -27,13 +27,20 @@ cd test
...
+cat $workdir/testerror.txt
+cat $workdir/testreport.txt | cat - $workdir/testlog.txt > temp && mv temp $workdir/testlog.txt
```
chriseth: Wouldn't `cat $workdir/testreport.txt $workdir/testlog.txt > $workdir/testlog.txt` do the same?

winsvega: cat: input file is output file

chriseth: Ah ok, but you can just use a new file.

PR #3537, opened 2017-02-01
fill general state tests as blockchain tests

winsvega:
allows to fill StateTestsGeneral as the blockchain tests using --fillchain option 

please review the code

### DIFF
--- a/test/TestHelper.cpp
+++ b/test/TestHelper.cpp
@@ -77,9 +77,10 @@ void printHelp()
 	cout << setw(30) << "--wallet" << setw(25) << "Enable wallet tests" << std::endl;
 	cout << setw(30) << "--all" << setw(25) << "Enable all tests" << std::endl;
 
-	cout << std::endl << "Test Creation" << std::endl;
+	cout << std::endl << "Test Generation" << std::endl;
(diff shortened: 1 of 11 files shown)


test/TestHelper.h
```diff
@@ -157,6 +157,9 @@ class ImportTest
 		eth::Network netId;
 	};
 	std::vector<transactionToExecute> m_transactions;
+	typedef std::pair<eth::State, eth::AccountMaskMap> stateAndMap;
```
chfast: Please use `using` instead of `typedef`: `using stateAndMap = std::pair<...`. And name the types with capital letter.

test/TestHelper.cpp
```diff
@@ -1113,6 +1233,8 @@ Options::Options(int argc, char** argv)
 		}
 		else if (arg == "--filltests")
 			fillTests = true;
+		else if (arg == "--fillchain")
+			fillBlockchain = true;
```
chfast: Be consistent with names. If the param name is `--fillchain` name the variable `fillchain` (or `fillChain`, but I prefer exactly the same name).

chfast:
Thank you. Please make sure the CI is green and merge.

PR #3540, opened 2017-02-02
Fix consensus failure during synchronization

gumb0:
Optimizer shouldn't write to the code stored in ExtVMFace, it should modify only its own copy instead.

### DIFF
--- a/libevm/VMOpt.cpp
+++ b/libevm/VMOpt.cpp
@@ -63,14 +63,16 @@ void VM::copyCode(int _extraBytes)
 
 void VM::optimize()
 {
-	size_t pc, nBytes = m_ext->code.size();
+	copyCode(33);
+
+	size_t const nBytes = m_ext->code.size();
 
 	// build a table of jump destinations for use in verifyJumpDest
 	
 	TRACE_STR(1, "Build JUMPDEST table")
-	for (pc = 0; pc < nBytes; ++pc)
+	for (size_t pc = 0; pc < nBytes; ++pc)
 	{
-		Instruction op = Instruction(m_ext->code[pc]);
+		Instruction op = Instruction(m_code[pc]);
 		TRACE_OP(2, pc, op);
 				
 		// make synthetic ops in user code trigger invalid instruction if run
@@ -81,7 +83,7 @@ void VM::optimize()
 		)
 		{
 			TRACE_OP(1, pc, op);
-			m_ext->code[pc] = (byte)Instruction::BAD;
+			m_code[pc] = (byte)Instruction::BAD;
 		}
 
 		if (op == Instruction::JUMPDEST)
@@ -107,7 +109,7 @@ void VM::optimize()
 		else if (op == Instruction::JUMPV || op == Instruction::JUMPSUBV)
 		{
 			++pc;
-			pc += 4 * m_ext->code[pc];  // number of 4-byte dests followed by table
+			pc += 4 * m_code[pc];  // number of 4-byte dests followed by table
 		}
 		else if (op == Instruction::BEGINSUB)
 		{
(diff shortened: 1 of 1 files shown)


libevm/VMOpt.cpp
```diff
@@ -63,14 +63,16 @@ void VM::copyCode(int _extraBytes)
 
 void VM::optimize()
 {
+	copyCode(33);
```
chriseth: Due to line 112 we might have to copy much more. Perhaps we can just append the zeros after the first pass?

gumb0: I changed it here to the way it was before recent modifications -- reserving `codeSize + 33` bytes, copying all the code.
I'm not sure it's correct, and what was the purpose of reserving extra `pc - nBytes + 33` bytes (maybe it was optimization to reserve less in case the for loop hits the break on line 120?)

gumb0: In which case does it need much more? The loop continues while ` pc < nBytes` and the only case I see when `pc` ends up > `nBytes` is if `pc` is incremented in one of the lines 09, 107, 112 and the loop ends on that.

gumb0: Probably we could resize m_codeSpace after the loop if it's really needed... Yeah, that was what you meant I think, ok I'll add it.

libevm/VMOpt.cpp
```diff
@@ -119,9 +121,13 @@ void VM::optimize()
 		}
 #endif
 	}
-	
-	copyCode(pc - nBytes + 33);
 
+	if (pc > nBytes)
+	{
+		m_codeSpace.resize(m_codeSpace.size() + pc - nBytes);
```
chriseth: I think it is better to create a helper function (which is also used by `copyCode`) that does the following:
```
m_codeSpace.resize(size);
m_code = m_codeSpace.data()
```

libevm/VMOpt.cpp
```diff
@@ -119,9 +126,10 @@ void VM::optimize()
 		}
 #endif
 	}
-	
-	copyCode(pc - nBytes + 33);
 
+	if (pc > nBytes)
+		resizeCodeSpace(m_codeSpace.size() + pc - nBytes);
+
```
gcolvin: I think this is not needed at all, as noted above.  Was the reason for creating resizeCodeSpace() so we could reuse it here?

libevm/VMOpt.cpp
```diff
@@ -119,9 +126,10 @@ void VM::optimize()
 		}
 #endif
 	}
-	
-	copyCode(pc - nBytes + 33);
```
gcolvin: Somehow when I merged this code a line got left behind.  At VMOpt:125, just inside the BEGINDATA case, it should say
> pc = nBytes;

I think the implication is that this loop cannot in fact make the data bigger, I just thought once that the BEGINDATA payload might.

libevm/VMOpt.cpp
gcolvin: It won't need more than 33 extra bytes ever with the current optimizations.  But it needs them here, not later, thanks.

libevm/VMOpt.cpp
```diff
@@ -57,20 +57,27 @@ void VM::copyCode(int _extraBytes)
...
 	m_codeSpace = m_ext->code;
-	m_codeSpace.resize(extendedSize);
+	resizeCodeSpace(extendedSize);
+}
+
+void VM::resizeCodeSpace(size_t _size)
+{
+	m_codeSpace.resize(_size);
 	m_code = m_codeSpace.data();
 }
```
gumb0: No, it was just extracting these two lines into separate method, since it was needed twice. But I'll remove it since you say it's not needed, thanks

libevm/VMOpt.cpp
gumb0: Ok, so now we won't need this `pc = nBytes` since we don't use `pc` after the loop.
And I moved `pc` declaration into the for loop, because it won't be needed outside.

gumb0:
It's still approved and passes tests, so I'm merging it

PR #3541, opened 2017-02-03
Add the packaging metadata to build the cpp-ethereum snap

elopio:
This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds that then can be distributed to many users through the Ubuntu store.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -7,8 +7,6 @@ grade: devel
 confinement: devmode
 
 apps:
-  bench:
-    command: bench
   eth:
     command: eth
   ethkey:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,34 @@
+name: cpp-ethereum
+version: master
+summary: blockchain app platform
+description: Ethereum C++ project
+
+grade: devel
+confinement: devmode
+
+apps:
+  bench:
+    command: bench
+  eth:
+    command: eth
+  ethkey:
+    command: ethkey
+  ethminer:
+    command: ethminer
+  ethvm:
+    command: ethvm
+  rlp:
+    command: rlp
+
+parts:
+  cpp-ethereum:
+    source: .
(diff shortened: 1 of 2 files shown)


elopio:
You can find more information about snapcraft here: http://snapcraft.io/

This is just an initial snap to start the discussion. It builds, and runs, but needs more testing and polishing.

To test it in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone --recursive https://github.com/elopio/cpp-ethereum
    $ cd cpp-ethereum
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous --devmode

(dangerous because the snap you build yourself doesn't come signed from the trusted store. devmode to only warn about confinement, which helps to define the required permissions properly. Once you push it to the store, your users will just have to do $ sudo snap install cpp-ethereum)

If you have any questions or comments, please let me know.

chfast:
Wow, nice. Looks good to me.

Just one question: is the path `snap/snapcraft.yml` the "standard" location for the file?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,34 @@
...
+version: master
+summary: blockchain app platform
+description: Ethereum C++ project
+
+grade: devel
+confinement: devmode
+
+apps:
+  bench:
+    command: bench
```
chfast: I'm not sure this should be exported. It is just some internal benchmark tool.

elopio:
Thanks for the review @chfast !

`snap/snapcraft.yml` is the default location. You can actually make snapcraft work from any location as long as the `source` is correct.

chriseth:
That's nice! I would prefer the file to be in e.g. `dist/snap/snapcraft.yaml`, then we can do similar things for debian, etc.

elopio:
@chriseth I've moved the yaml. Please note that this requires snapcraft to be called from dist/snap.

Issue #3520, opened 2017-01-25
Implement EIP5 - better gas for call output


chfast:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-5.md
https://github.com/ethereum/EIPs/issues/8

Implementing https://github.com/ethereum/cpp-ethereum/issues/3318 first would be helpful.

chfast:
This is dead, right @chriseth ?

chriseth:
Yes.

PR #3549, opened 2017-02-08
Rework returning output from VMs

### LINKED ISSUE #3318: Redesign VM output management
I would like to propose change in the VM-related API. Now the user have to provide the output buffer where the VM will copy the output (it comes from current CALL semantics).
I want to change that to return the output buffer reference with attached virtual destructor to the user. The user is responsible to copy whatever it wants and release the buffer.

This should be aligned with:
1. The way how CREATE works.
2. https://github.com/ethereum/EIPs/issues/8.
3. Changes planned in EVM-C: aligning external and internal calls.

chfast:
This PR introduces `owning_bytes_ref` type -- description in the docs.

This resolves #3318 (Executive might require changes as well) and goes towards #3520.

The big win here is that we don't have to keep a VM instance around to access the output. The output is just returned from `exec()` method so VMs are "stateless". Check JitVM and SmartVM to see the simplifications.


chfast:
Sorry, I just pushed more similar changes, now around Executive / ExtVM. They are not so beautiful any more. Have to go now...

chfast:
This is blocked by old boost on Ubuntu 14.04 (including Travis CI).

chfast:
Unfortunately, I had to change the code around precompiled contracts, what make it harder to review.

libevm/ExtVMFace.h
```diff
@@ -37,6 +38,46 @@ namespace dev
...
 
+/// Reference to a slice of buffer that also owns the buffer.
+///
+/// This is extension to the concept C++ STL library names as array_view
+/// (also known as gsl::span, array_ref, here vector_ref) -- reference to
+/// continuous non-modifiable memory. The extension makes the object also owning
+/// the referenced buffer.
+///
+/// This type is used by VMs to return output coming from RETURN instruction.
+/// To avoid memory copy, a VM returns its whole memory + the information what
```
chriseth: Is this still valid?

libevm/ExtVMFace.h
```diff
@@ -37,6 +38,46 @@ namespace dev
...
+/// part of this memory is actually the output. This simplifies the VM design,
+/// because there are multiple options how the output will be used (can be
+/// ignored, part of it copied, or all of it copied). The decision what to do
+/// with it was moved out of VM interface making VMs "stateless".
+///
+/// The type is movable, but not copyable. Default constructor available.
```
chriseth: So the difference to `bytes` is only that it is not copyable? Or is the other benefit that it supports the `vector_ref` interface?

chfast: It's the opposite -- it's extended `vector_ref`. With simple `bytes` you cannot state e.g. "this is a bytes buffer of 100 bytes, but interesting part is a slice from byte 13 of size 22".

I have in mind another change, to have polymorphic buffer implementation to handle the buffers coming from different VM implementations (like from EVMJIT, where it is allocated by `std::malloc`).

libevm/ExtVMFace.h
chfast: What exactly?

chriseth: That we keep the full memory, but now I understand that yes :-)

PR #3552, opened 2017-02-09
Cleanups

chfast:


libethereum/BlockChain.cpp
```diff
@@ -181,9 +180,6 @@ void BlockChain::init(ChainParams const& _p, std::string const& _path)
 	m_sealEngine.reset(m_params.createSealEngine());
 	m_genesis.clear();
 	genesis();
-
-	// remove the next line real soon. we don't need to be supporting this forever.
```
chriseth: Only took two weeks.

libwebthree/Support.cpp
```diff
@@ -111,7 +111,6 @@ void Support::hintSHA256(h256 const& _content, h256 const& _sha256, Secret const
 Address Support::icapRegistrar() const
```
chriseth: Is this used somewhere?

chfast: My IDE says yes, but it might be wrong. Have not check details.

chfast:
I also removed this Support thing.

PR #3557, opened 2017-02-10
Fix deadlocks in sync

gumb0:
1. Fix counting the total size of known blocks in BlockQueue

`BlockQueue` keeps track the total size & count of known blocks (stored in `m_unverified`, `m_unverified`, `m_verifying`), increases it when new blocks arrive, correctly decreases when blocks are imported into blockchain, but forgets to decrease when blocks are removed when being marked as bad (e.g. when verifying fails in `BlockQueue::verifierBody()`).
This led to a deadlock similiar to the one in https://github.com/ethereum/cpp-ethereum/issues/3182 when `m_knownSize` exceeds the limit after many bad blocks arrive, and sync is paused after that forever.

This fix encapsulates all counting of total size into separate class `SizedBlockQueue` instead of keeping it in `m_knownSize`.

2. The similar refactoring for `m_unknownSize` & `m_unknowCount`.

3. Fix resetting the sync
BlockChainSync should switch SyncState::NotSynced when sync is being reset due to chain reorganization.

### DIFF
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -742,7 +742,7 @@ void BlockChainSync::resetSync()
 	m_bodySyncPeers.clear();
 	m_headerIdToNumber.clear();
 	m_syncingTotalDifficulty = 0;
-	m_state = SyncState::Idle;
+	m_state = SyncState::NotSynced;
 }
 
 void BlockChainSync::restartSync()
@@ -755,7 +755,6 @@ void BlockChainSync::restartSync()
 	m_startingBlock = host().chain().number();
 	m_lastImportedBlock = m_startingBlock;
 	m_lastImportedBlockHash = host().chain().currentHash();
-	m_state = SyncState::NotSynced;
 }
 
 void BlockChainSync::completeSync()
--- a/libethereum/BlockQueue.cpp
+++ b/libethereum/BlockQueue.cpp
@@ -46,9 +46,7 @@ size_t const c_maxUnknownSize = 512 * 1024 * 1024; // Block size can be ~50kb
 
 BlockQueue::BlockQueue():
 	m_unknownSize(0),
-	m_knownSize(0),
-	m_unknownCount(0),
-	m_knownCount(0)
+	m_unknownCount(0)
 {
 	// Allow some room for other activity
 	unsigned verifierThreads = std::max(thread::hardware_concurrency(), 3U) - 2U;
@@ -90,8 +88,6 @@ void BlockQueue::clear()
 	m_future.clear();
 	m_unknownSize = 0;
 	m_unknownCount = 0;
-	m_knownSize = 0;
-	m_knownCount = 0;
 	m_difficulty = 0;
 	m_drainingDifficulty = 0;
 }
@@ -104,19 +100,19 @@ void BlockQueue::verifierBody()
 
 		{
 			unique_lock<Mutex> l(m_verification);
-			m_moreToVerify.wait(l, [&](){ return !m_unverified.empty() || m_deleting; });
+			m_moreToVerify.wait(l, [&](){ return !m_unverified.isEmpty() || m_deleting; });
 			if (m_deleting)
 				return;
-			swap(work, m_unverified.front());
-			m_unverified.pop_front();
+			work = m_unverified.dequeue();
+
 			BlockHeader bi;
 			bi.setSha3Uncles(work.hash);
 			bi.setParentHash(work.parentHash);
-			m_verifying.emplace_back(move(bi));
+			m_verifying.enqueue(move(bi));
 		}
 
 		VerifiedBlock res;
-		swap(work.block, res.blockData);
+		swap(work.blockData, res.blockData);
 		try
 		{
 			res.verified = m_bc->verifyBlock(&res.blockData, m_onBad, ImportRequirements::OutOfOrderChecks);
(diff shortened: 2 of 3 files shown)


libethereum/BlockQueue.h
```diff
@@ -67,6 +67,117 @@ enum class QueueStatus
 
 std::ostream& operator<< (std::ostream& os, QueueStatus const& obj);
 
+template<class T>
+class SizedBlockQueue
+{
+public:
+	std::size_t count() const
```
chriseth: Would it be better readable to put short one-liners on a single line?
The argument would be that the class then easily fits a single page and its behaviour can more easily be grasped.

libethereum/BlockQueue.h
```diff
@@ -67,6 +67,117 @@ enum class QueueStatus
...
+
+	void clear()
+	{
+		m_queue.clear();
+		m_size = 0;
+	}
+
+	void enqueue(T&& _t)
+	{
+		m_size += _t.blockData.size();
```
chriseth: Any exception safety that has to be obeyed here?

libethereum/BlockQueue.h
```diff
@@ -67,6 +67,117 @@ enum class QueueStatus
...
+		return m_queue.front().verified.info.sha3Uncles();
+	}
+
+	T const& next() const
+	{
+		return m_queue.front();
+	}
+
+	template<class Op>
+	void forEach(Op _op) const
```
chriseth: Would it be better to implement a boost range compatible interface so that `boost::for_each(queue, []{..})` can be used as easily as any other boost range algorithms?

chfast:
Is this ready to review? If yes, mark with [needs review] label.

libethereum/BlockQueue.h
gumb0: Yeah, that's more common to our codebase, although I find one-liners less readable (and and harder debuggable)
Ok, I'll change it, fitting to a page is a good argument

libethereum/BlockQueue.h
gumb0: Move constructor shouldn't throw, so I think it's fine here like this.

libethereum/BlockQueue.h
gumb0: That, I guess, means just implementing begin() & end() ?
(never saw yet boost range algorithms used in our code)
Yeah, that would be more idiomatic, I'll try it.

gumb0:
@chfast This part can be reviewed, but I'd like to do some more refactoring here (see description), so I haven't put [needs review] yet

libethereum/BlockQueue.h
gumb0: Turns out this is not needed at all, I've just forgotten to remove it

libethereum/BlockChainSync.cpp
```diff
@@ -742,7 +742,7 @@ void BlockChainSync::resetSync()
 	m_bodySyncPeers.clear();
 	m_headerIdToNumber.clear();
 	m_syncingTotalDifficulty = 0;
-	m_state = SyncState::Idle;
+	m_state = SyncState::NotSynced;
```
gumb0: I'm not 100% sure it won't break anything but it doesn't look right at all.

Currently this `BlockChainSync::resetSync()` may be called during sync from https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/BlockChainSync.cpp#L601
Then `BlockChainSync::isSyncing()` reports that sync is complete and `Client` begins to import new transactions from TransactionQueue here: https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/Client.cpp#L676 
This shouldn't happen until sync is really complete.

libethereum/BlockQueue.h
gumb0: Although, yeah, it would be nicer to emplace_back first, then update m_size, I'll change it

libethereum/BlockChainSync.cpp
chfast: I've seen this behavior.

libethereum/BlockQueue.h
```diff
@@ -113,7 +263,7 @@ class BlockQueue: HasInvariants
 	h256 firstUnknown() const { ReadGuard l(m_lock); return m_unknownSet.size() ? *m_unknownSet.begin() : h256(); }
 
 	/// Get some infomration on the current status.
-	BlockQueueStatus status() const { ReadGuard l(m_lock); Guard l2(m_verification); return BlockQueueStatus{m_drainingSet.size(), m_verified.size(), m_verifying.size(), m_unverified.size(), m_future.size(), m_unknown.size(), m_knownBad.size()}; }
+	BlockQueueStatus status() const { ReadGuard l(m_lock); Guard l2(m_verification); return BlockQueueStatus{m_drainingSet.size(), m_verified.count(), m_verifying.count(), m_unverified.count(), m_future.count(), m_unknown.count(), m_knownBad.size()}; }
```
chfast: Wow. This is a long line. Maybe it's time to move it to .cpp file?

libethereum/BlockQueue.h
```diff
@@ -67,6 +67,156 @@ enum class QueueStatus
...
 
+template<class T>
+class SizedBlockQueue
+{
+public:
+	std::size_t count() const { return m_queue.size(); }
+
+	std::size_t size() const { return m_size; }
+	
+	bool isEmpty() const { return m_queue.empty(); }
```
chfast: Why not `empty()` as standard containers?

libethereum/BlockQueue.cpp
```diff
@@ -237,14 +217,12 @@ ImportResult BlockQueue::import(bytesConstRef _block, bool _isOurs)
 	// Check it's not in the future
 	if (bi.timestamp() > utcTime() && !_isOurs)
 	{
-		m_future.insert(make_pair((unsigned)bi.timestamp(), make_pair(h, _block.toBytes())));
+		m_future.insert((unsigned)bi.timestamp(), h, _block.toBytes());
 		char buf[24];
 		time_t bit = (unsigned)bi.timestamp();
```
chfast: I'm not sure using `unsigned` in place of `time_t` is good idea. Maybe use `time_t` directly in `m_future`?

libethereum/BlockQueue.h
gumb0: It is more in accordance with our coding standards for bool getters, it is not a standard container and `empty()` is ambiguous as can be interpreted as "empty it"

libethereum/BlockQueue.h
gumb0: yeah, ok

libethereum/BlockQueue.cpp
gumb0: Will do

PR #3558, opened 2017-02-13
prepare metropolis

winsvega:
connects to #3561

### DIFF
--- a/libethashseal/GenesisInfo.cpp
+++ b/libethashseal/GenesisInfo.cpp
@@ -61,6 +61,7 @@ h256 const& dev::eth::genesisStateRoot(Network _n)
 	case Network::FrontierTest: return c_genesisStateRootFrontierTest;
 	case Network::HomesteadTest: return c_genesisStateRootHomesteadTest;
 	case Network::EIP150Test: return c_genesisStateRootEIP150Test;
+	case Network::EIP158Test: return c_genesisStateRootEIP158Test;
 	case Network::MetropolisTest: return c_genesisStateRootMetropolisTest;
 	default:
 		throw std::invalid_argument("Invalid network value");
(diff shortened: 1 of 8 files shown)


libethashseal/GenesisInfo.cpp
```diff
@@ -59,7 +61,7 @@ h256 const& dev::eth::genesisStateRoot(Network _n)
 	case Network::FrontierTest: return c_genesisStateRootFrontierTest;
 	case Network::HomesteadTest: return c_genesisStateRootHomesteadTest;
 	case Network::EIP150Test: return c_genesisStateRootEIP150Test;
-	case Network::EIP158Test: return c_genesisStateRootEIP158Test;
```
chfast: Why EIP158 removed here?

libethashseal/genesis/metropolisTest.cpp
```diff
@@ -0,0 +1,60 @@
...
+		MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+		GNU General Public License for more details.
+
+		You should have received a copy of the GNU General Public License
+		along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#include "../GenesisInfo.h"
+
+static dev::h256 const c_genesisStateRootMetropolisTest;
+static std::string const c_genesisInfoMetropolisTest = std::string() +
```
chfast: The part `std::string() +` make no sense.

chfast:
@winsvega, can you clean up the constants like `c_genesisInfoMetropolisTest`. My opinion is to keep them as `char const[]`, i.e. c-string literal `""`. You can make the `std::string` when needed to be parsed. In this case we don't waste heap memory all the time, and avoid dynamic initialization.

But I can accept `std::string` if you don't agree, just make it look beautiful.

After that, you can merge it.

Issue #3535, opened 2017-01-31
Fix eth solo build


chfast:
This should work:
```
make eth
make install
```

- [x] `make eth` is broken -- BuildInfo.h header is not generated.
- [x] If only eth has been built, `make install` should only install eth. For this we might want to mark other install targets as `OPTIONAL`.

chriseth:
Sorry, but what is this issue about?

chfast:
Updated description.

chriseth:
Ok, now it is clearer - I though this was a bug and the proposed solution was `make eth; make install` ;-)

Issue #3300, opened 2016-09-10
Race condition with BuildInfo.h generation


bobsummerwill:
There is a race condition with `buildinfo.h` generation within the build system.

It only manifests when building with very broad parallelism (`-j8` or `-j16`, for example).

The issue occurs because the makefile generated by the CMake process creates a phony target for that generation step, but does not generate any explicit dependencies onto that phony target from the targets for the C++ code compilation which use that generated header file.

With only 2 or 4 cores, that doesn't seem to be a problem.   I assume that the processes are launched in parallel but the codegen step nearly always completes in time.    With 8 or 16 processes racing, one of the C++ processes often gets to requiring that header file before it is actually generated.

We need to work out how to tweak our CMake files so that it generate those dependency edges in the makefile.

hobi82:
Quick question: Is this the same issue I see on my build system (Windows 10 1607, CMake, Visual C++ 2015):

---snip
**command used: src\github.com\ethereum\cpp-ethereum\build>cmake --build . --config RelWithDebInfo**

Build FAILED.

"C:\Users\hobi\src\github.com\ethereum\cpp-ethereum\build\ALL_BUILD.vcxproj" (default target) (1) ->
"C:\Users\hobi\src\github.com\ethereum\cpp-ethereum\build\bench\bench.vcxproj" (default target) (3) ->
"C:\Users\hobi\src\github.com\ethereum\cpp-ethereum\build\libdevcore\devcore.vcxproj" (default target) (4) ->
(ClCompile target) ->
  **C:\Users\hobi\src\github.com\ethereum\cpp-ethereum\libdevcore\Common.cpp(25): fatal error C1083: Cannot open include
file: 'cpp-ethereum/BuildInfo.h': No such file or directory [C:\Users\hobi\src\github.com\ethereum\cpp-ethereum\build\l**
ibdevcore\devcore.vcxproj]

    0 Warning(s)
    1 Error(s)
---snap

If yes, how can I fix that? I was following this wiki entry to setup my build system: http://www.ethdocs.org/en/latest/ethereum-clients/cpp-ethereum/building-from-source/windows.html

thank you!

hwxie:
I have the exact same issue as hobi82 with same OS. Is there a flag I can use to limit to building by 2 cores?

bobsummerwill:
@hobi82 @hwxie It could well be the same issue, yes.

[...]

PR #3565, opened 2017-02-16
Probably better "make install"

chfast:
- Resolves BuildInfo.h dependency issue by placing explicit libdevcore -> BuildInfo.h dependency. Fixes https://github.com/ethereum/cpp-ethereum/issues/3300.
- Marks all targets except eth as OPTIONAL for install. This means you can do `make eth; make install` to install only `eth`. Resolves https://github.com/ethereum/cpp-ethereum/issues/3535. However, the issue is `make install` will not work without previous `make eth` or `make`.


chriseth:
Nice! Can you do the same thing for solidity? :-)

chfast:
Yes, sure. Create an issue and assign me.

PR #3566, opened 2017-02-16
Send daily test mail to more people.

chriseth:


chriseth:
@winsvega can you manually trigger the run once so that we see this is working?

Issue #3161, opened 2016-08-26
Clang - Can we get rid of -Qunused-arguments?


bobsummerwill:
_From @bobsummerwill on May 9, 2016 19:17_

See https://github.com/ethereum/webthree-helpers/blob/develop/cmake/EthCompilerSettings.cmake#L127.

What happens if we remove this?   Can we just fix the underlying issue?

_Copied from original issue: ethereum/webthree-umbrella#504_

Issue #3160, opened 2016-08-26
Clang - Can we get rid of -Wno-unused-function and -Wno-dangling-else?


bobsummerwill:
_From @bobsummerwill on May 9, 2016 19:16_

See https://github.com/ethereum/webthree-helpers/blob/develop/cmake/EthCompilerSettings.cmake#L107.

Can we remove these?   What happens?   What code needs these?   Can we just fix the code?

_Copied from original issue: ethereum/webthree-umbrella#503_

bobsummerwill:
_From @chfast on May 12, 2016 21:28_

Remove -> build -> fix code -> commit.

bobsummerwill:
I know.   This is a reminder to try that at some stage.

Issue #3158, opened 2016-08-26
Arch Linux - Localize -permissive warning suppressions


bobsummerwill:
_From @bobsummerwill on May 9, 2016 19:7_

CC @Legogris, @tymat, @gcolvin, @chfast.

To work around warnings in Boost which are showing up on Arch Linux and FreeBSD, I just added a "-permissive" in our GCC and Clang builds.

See https://github.com/ethereum/webthree-helpers/pull/171.

We should actually take the time to track down exactly where the issues manifest, and add localized push/pop pragmas to the code which has the includes.     Or better yet, work out what the warnings are and get them fixed.    Maybe we need a newer Boost?   Maybe we need some extra defines?

The aim is to remove https://github.com/ethereum/webthree-helpers/blob/develop/cmake/EthCompilerSettings.cmake#L55.

_Copied from original issue: ethereum/webthree-umbrella#500_

bobsummerwill:
_From @ijcoe6ru on June 28, 2016 3:14_

The build error is because of boost. It is fixed in https://github.com/boostorg/multiprecision/commit/f9c8f9ec091ad232c0a291904f7839d665d098e0 or boost-1.61.0.

bobsummerwill:
Thanks for the info, @ijcoe6ru :-)

When I get a chance, I will try to add more focussed pragmas around this, maybe even with version-conditionals.    Knowing where the root issue is certainly helps a lot!

bobsummerwill:
_From @ijcoe6ru on July 1, 2016 13:10_

I did a search here:

https://software.opensuse.org/package/boost

As of July 1st 12:56 UTC, among openSUSE, SUSE, Debian, RedHat, CentOS, Ubuntu, and CentOS, only OpenSUSE 42, SLE-12, SLE-11, CentOS7, and CentOS6 have boost 1.61.0. Some other versions of openSUSE have the package boost_1_61.

This build error was reported to occur on Arch Linux, but Arch Linux's boost is still 1.60.0. The aur packages boost-build-git and boost-build-nightly are later than 1.61.0.

Since many distributions don't have boost 1.61.0, is it OK to use a fixed version of `cpp_int.hpp` in Ethereum's source?

bobsummerwill:
As in "ophan a copy of the code into Ethereum's source"?

[...]

PR #3568, opened 2017-02-16
Compiler flags cleanups

chfast:
Fixes https://github.com/ethereum/cpp-ethereum/issues/3160.
Fixes https://github.com/ethereum/cpp-ethereum/issues/3158.
Fixes https://github.com/ethereum/cpp-ethereum/issues/3161.


bobsummerwill:
Fantastic! :-)

chriseth:
Actually I just wondered whether we might need those function if we want to replace cryptopp by secp256k1 in the future, but it's probably quite easy to re-introduce them.

bobsummerwill:
And to fix the warnings in that code, if it does end up being needed!

PR #3570, opened 2017-02-17
Docker update and fixes

chfast:
This updates the Dockerfile not to install boost and to build only eth.
Also secp256k1 initialization is moved to eth main thread to avoid stack overflow.

### DIFF
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -13,10 +13,6 @@ FROM alpine
 
 RUN apk add --no-cache \
         libstdc++ \
-        boost-filesystem \
-        boost-random \
-        boost-regex \
-        boost-thread \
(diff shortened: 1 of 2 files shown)


scripts/Dockerfile
```diff
@@ -26,13 +22,13 @@ RUN apk add --no-cache \
         cmake \
         g++ \
         make \
-        curl-dev boost-dev libmicrohttpd-dev \
+        linux-headers curl-dev libmicrohttpd-dev \
```
chriseth: `linux-headers`? Do we really need that?

chfast: Yes. Boost includes at least `<linux/version.h>`. The package is removed after the build.

PR #3574, opened 2017-02-17
Fix generating headers response when start block is unknown

gumb0:


libethereum/EthereumHost.cpp
```diff
@@ -185,12 +185,13 @@ class EthereumHostData: public EthereumHostDataFace
...
-			<< ", skip: " << _skip << ", reverse: " << _reverse << ")";
+				<< ", maxHeaders: " << _maxHeaders
+				<< ", skip: " << _skip << ", reverse: " << _reverse << ")";
 
-			if (!_reverse)
+			if (!m_chain.isKnown(blockHash))
+				blockHash = {};
+			else if (!_reverse)
 			{
 				auto n = m_chain.number(blockHash);
```
gumb0: I've caught this while syncing with Ropsten. When we get a request with start block hash which we don't know yet, `m_chain.number(blockHash)` returns `(unsigned)-1`, then all the following calculations are incorrect of course, and we end up in the long loop collecting a lot of headers.

PR #3576, opened 2017-02-19
Validator

gcolvin:
Validation is implemented as a method of the VM for now, in order to use its functions and macros for traversing and decoding bytecode. Could factor that into a common base, but at the cost of VM performance.  Note that all of this new code is hidden behind the EVM_JUMPS_AND_SUBS macro, currently defined false in VMConfig.h.  When defined true the code compiles, but cannot run without further support yet to be merged.


libevm/VMValidate.cpp
```diff
@@ -0,0 +1,159 @@
...
+void VM::validate(ExtVMFace& _ext)
+{
+	m_ext = &_ext;
+	initEntry();
+	size_t PC;
+	byte OP;
+	for (PC = 0; (OP = m_code[PC]); ++PC)
+		if  (OP == byte(Instruction::BEGINSUB))
+			validateSubroutine(PC, m_return, m_stack);
+		else if (OP == byte(Instruction::BEGINDATA))
```
chriseth: I think you have to check for multi-byte instructions in this loop, don't you?

gcolvin: Yes.  Thanks for the reminder.

gcolvin:
I would like to understand why when I went to bed last night this PR was ready for automatic merge and all checks passed, and today it will not merge and no checks pass.  This is apparently due to days-old changes to VM.h and VM.cpp.  This despite doing, as the commit history shows, "Merge branch 'develop' of https://github.com/ethereum/cpp-ethereum into validator".  Between that and the time zones this PR may waste days of our time rather that the hours it deserves.  Very discouraging.

chfast:
Because of my merge of another PR. My merge commit is 3 hours old.

I can resolve conflicts for you. They are easy, I checked.

gcolvin:
Aha.  Git hates me, it really does.  In a year now I have not had a PR go smoothly, no matter what I do.  The resolution was easy, so I did it, thanks.

gcolvin:
Appveyor has failed.  I am quite clueless:
Cloning into 'C:/projects/cpp-ethereum/evmjit'...
fatal: destination path 'C:/projects/cpp-ethereum/deps' already exists and is not an empty directory.
fatal: clone of 'https://github.com/ethereum/cpp-dependencies' into submodule path 'C:/projects/cpp-ethereum/deps' failed
Failed to clone 'deps' a second time, aborting

gcolvin:
@chfast  @chriseth  Requestes changes are in, merged with develop and tested on my end, pushed and testing on github now.  Appveyor is failing in ways that look like configuration issues, or at least not to do with the code.

chfast:
I also restarted the AppVeyor. It looks like there is a race condition between git checkout and cache restore.

gcolvin:
The logic tries to follow the EIP as closely as it can, but you are right, testing will tell.  I have a version that is further along that I am debugging.  If the top level is `validateSubroutine()` that is ok, but it leaves it to the caller to traverse the bytecode, which is tricky.  It seemed the usual case would be to pass a whole contract to `validate()`, but I made both functions public.  (Thanks for the restart.)

gcolvin:
@chriseth Can you merge this?

PR #3577, opened 2017-02-20
Fix sync deadlock

gumb0:


libethereum/BlockChain.cpp
```diff
@@ -432,6 +432,11 @@ tuple<ImportRoute, bool, unsigned> BlockChain::sync(BlockQueue& _bq, OverlayDB c
 				std::move(std::begin(r.goodTranactions), std::end(r.goodTranactions), std::back_inserter(goodTransactions));
 				++count;
 			}
+			catch (dev::eth::AlreadyHaveBlock const&)
```
gumb0: Without this the exception goes to `catch (Exception& ex)` below, where this block is marked as bad (but it's not bad, just already added to the blockchain). After that BlockQueue considers all child blocks bad, too, and the sync doesn't progress anymore.

(Still not clear why the same block is being imported twice, probably it shouldn't happen)

PR #3578, opened 2017-02-20
Fix build info for eth

chfast:


chriseth:
Why didn't this cause a compiler error?

chfast:
Macros :)

PR #3580, opened 2017-02-21
callback from blockchain when new block is imported

winsvega:


libethereum/BlockChain.h
```diff
@@ -292,6 +292,9 @@ class BlockChain
 	/// Change the function that is called with a bad block.
 	template <class T> void setOnBad(T const& _t) { m_onBad = _t; }
 
+	/// Change the function that is called when a new block is imported
+	template <class T> void setOnBlockImport(T const& _t) { m_onBlockImport = _t; }
```
chfast: There is no need to have a template here.

winsvega: what about the previous one?

libethereum/BlockChainSync.cpp
```diff
@@ -177,6 +177,12 @@ BlockChainSync::~BlockChainSync()
 	abortSync();
 }
 
+void BlockChainSync::onBlockImported(BlockHeader const& _info)
+{
+	string number = dev::toString(_info.number());
```
gumb0: Conversion to string is not needed, just `static_cast<unsigned>(_info.number())`

libethereum/BlockChainSync.h
```diff
@@ -26,6 +26,7 @@
 
 #include <libdevcore/Guards.h>
 #include <libethcore/Common.h>
+#include "libethcore/BlockHeader.h"
```
gumb0: Use angle brackets like surrounding #include's

libethereum/Client.cpp
```diff
@@ -103,15 +103,18 @@ void Client::init(p2p::Host* _extNet, std::string const& _dbPath, WithExisting _
...
 	bc().setOnBad([=](Exception& ex){ this->onBadBlock(ex); });
 
+
 	if (_forceAction == WithExisting::Rescue)
 		bc().rescue(m_stateDB);
 
 	m_gp->update(bc());
 
 	auto host = _extNet->registerCapability(make_shared<EthereumHost>(bc(), m_stateDB, m_tq, m_bq, _networkId));
+	bc().setOnBlockImport([=](BlockHeader const& _info){ host->onBlockImported(_info); });
```
gumb0: Do you need this to be here after `registerCapability` and not above just after `setOnBad` calls ?

libethereum/EthereumHost.h
```diff
@@ -36,6 +36,8 @@
 #include <libdevcore/OverlayDB.h>
 #include "CommonNet.h"
 #include "EthereumPeer.h"
+#include "BlockChainSync.h"
+#include "libethcore/BlockHeader.h"
```
gumb0: angle brackets

libethereum/BlockChain.h
chfast: Use `std::function` as the parameter type. In previous line as well.

libethereum/BlockChainSync.cpp
```diff
@@ -177,6 +177,12 @@ BlockChainSync::~BlockChainSync()
...
+	m_lastImportedBlock = atoi(number.c_str());
```
gumb0: `m_lastImportedBlockHash` needs to be updated every time `m_lastImportedBlock` is updated

libethereum/Client.cpp
winsvega: it works with 'host'

libethereum/BlockChainSync.cpp
```diff
@@ -177,6 +177,12 @@ BlockChainSync::~BlockChainSync()
 	abortSync();
 }
 
+void BlockChainSync::onBlockImported(BlockHeader const& _info)
+{
+	m_lastImportedBlock = static_cast<unsigned>(_info.number());
```
gumb0: These values are not atomic and used from network-message-handling thread simultaneously with importing, so I think it should be under `RecursiveGuard l(x_sync);`

gumb0: Also, I would add a comment here about why we need this handler, since it's not obvious.

libethereum/Client.cpp
gumb0: I would better move it close to `setOnBad` and use `m_host` instead of `host`

winsvega:
it does not work with m_host.
error: base operand of ‘->’ has non-pointer type ‘std::weak_ptr<dev::eth::EthereumHost>’
  bc().setOnBlockImport([=](BlockHeader const& _info){ m_host->onBlockImported(_info); });

gumb0:
Sure, you need to lock `weak_ptr` to get `shared_ptr` like
 `if (auto h = m_host.lock()) h->onBlockImported(_info);`

libethereum/Client.cpp
```diff
@@ -103,15 +103,22 @@ void Client::init(p2p::Host* _extNet, std::string const& _dbPath, WithExisting _
...
+
 	if (_forceAction == WithExisting::Rescue)
 		bc().rescue(m_stateDB);
 
 	m_gp->update(bc());
 
 	auto host = _extNet->registerCapability(make_shared<EthereumHost>(bc(), m_stateDB, m_tq, m_bq, _networkId));
 	m_host = host;
+
+	bc().setOnBlockImport([=](BlockHeader const& _info){
```
gumb0: Now you can move it above to line 105, because m_host exists there, and this callback can't be called, I think, until `startWorking()` call in the end of this method, and anyway if condition in lambda won't be true until m_host is set.

libethereum/BlockChainSync.cpp
```diff
@@ -836,7 +844,11 @@ bool BlockChainSync::invariants() const
 	if (!isSyncing() && !m_bodies.empty())
 		BOOST_THROW_EXCEPTION(FailedInvariant() << errinfo_comment("Got bodies while not syncing"));
 	if (isSyncing() && m_host.chain().number() > 0 && m_haveCommonHeader && m_lastImportedBlock == 0)
+	{
+		std::cerr << "Have common header " << m_haveCommonHeader;
```
chriseth: Please remove in the final version.

libethereum/BlockChainSync.cpp
```diff
@@ -177,6 +177,15 @@ BlockChainSync::~BlockChainSync()
 	abortSync();
 }
 
+void BlockChainSync::onBlockImported(BlockHeader const& _info)
+{
+	//if a block has been added via mining or other block import function
+	//through RPC, then we should count it as a last imported block
+	RecursiveGuard l(x_sync);
+	m_lastImportedBlock = static_cast<unsigned>(_info.number());
```
gumb0: I'm thinking now, that this could potentially break the sync.

This is called now during the sync, too, right?
`m_lastImportBlock` keeps track of the latest block we pushed into `BlockQueue` (for verification), and this can be further ahead than the blocks currently being imported into `BlockChain`, so this callback I guess can set m_lastImportedBlock to some earlier value.

So I think let's update `m_lastImportedBlock` & `m_lastImportedHash` only if the new number is greater

gumb0: Also, please add `m_highestBlock = max(m_lastImportedBlock, m_highestBlock)` - now I see that this is what `onPeerNewBlock` does after successful import and here the same should happen I suppose

winsvega: https://github.com/winsvega/cpp-ethereum/blob/5540e2520b32ff5efb75630e8dade2a85d1cb38d/libethereum/BlockChain.cpp#L867

I think this is the place where the latest best block is called. 
so it is not just import. but a successful block being added as a new top block.

gumb0: `m_lastImportedBlock` is the last block pushed to `BlockQueue` and during sync `BlockQueue` has block numbers further ahead of those currently being imported in `BlockChain`

libethereum/BlockChain.cpp
```diff
@@ -864,6 +864,8 @@ ImportRoute BlockChain::import(VerifiedBlockRef const& _block, OverlayDB const&
 		{
 			newLastBlockHash = _block.info.hash();
 			newLastBlockNumber = (unsigned)_block.info.number();
+			if (m_onBlockImport)
+				m_onBlockImport(_block.info);
```
chriseth: Would it be better to call this callback after we have written everything to the database (i.e. at the end of this method)?

Issue #3583, opened 2017-02-22
cmake .. fails on fedora


lhirlimann:
running latest fedora25 cmake fails : 
[ludo@Oulanl cpp-ethereum]$ mv build build.one
[ludo@Oulanl cpp-ethereum]$ mkdir build
[ludo@Oulanl cpp-ethereum]$ cd build
[ludo@Oulanl build]$ cmake .. -DBUNDLE=miner
-- The C compiler identification is GNU 6.3.1
-- The CXX compiler identification is GNU 6.3.1
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found CURL: /usr/lib64/libcurl.so  
-- Found mhd: /usr/lib64/libmicrohttpd.so  
CMake Error at CMakeLists.txt:50 (include):
  include could not find load file:

    deps/boost.cmake

------------------------------------------------------------------------
-- Configuring cpp-ethereum
------------------------------------------------------------------------
-- CMake 3.6.2 (/usr/bin/cmake)
-- CMAKE_BUILD_TYPE Build type                               RelWithDebInfo
-- TARGET_PLATFORM  Target platform                          Linux
-- STATIC BUILD                                              
--------------------------------------------------------------- features
--                  Hardware identification support          
--                  HTTP Request support                     TRUE
-- VMTRACE          VM execution tracing                     OFF
-- PROFILING        Profiling support                        OFF
-- FATDB            Full database exploring                  ON
-- ROCKSDB          Prefer rocksdb to leveldb                OFF
-- PARANOID         -                                        OFF
-- MINIUPNPC        -                                        ON
------------------------------------------------------------- components
-- TESTS            Build tests                              ON
-- EVMJIT           Build LLVM-based JIT EVM                 OFF
------------------------------------------------------------------------

-- Found leveldb: /usr/lib64/libleveldb.so  
-- Cpuid headers: CPUID_INCLUDE_DIR-NOTFOUND
[...]

PR #3584, opened 2017-02-22
CMake: check if git submodules are initialized

chfast:
This should help newcomers to resolve uninitialized git submodules issue without our attention, avoid questions like https://github.com/ethereum/cpp-ethereum/issues/3583.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,10 +44,13 @@ include(EthExecutableHelper)
 include(EthDependencies)
 include(EthUtils)
 
+if (NOT EXISTS ${CMAKE_SOURCE_DIR}/deps/boost.cmake)
+    message(FATAL_ERROR "Git submodules not initialized, execute:\n  git submodule update --init")
+endif()
+include(deps/boost.cmake)
 include(deps/jsoncpp.cmake)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -44,10 +44,13 @@ include(EthExecutableHelper)
 include(EthDependencies)
 include(EthUtils)
 
+if (NOT EXISTS ${CMAKE_SOURCE_DIR}/deps/boost.cmake)
+    message(FATAL_ERROR "Git submodules not initialized, execute:\n  git submodule update --init")
```
winsvega: thats good. I always forgot that command syntax

CMakeLists.txt
```diff
@@ -44,10 +44,13 @@ include(EthExecutableHelper)
 include(EthDependencies)
 include(EthUtils)
 
+if (NOT EXISTS ${CMAKE_SOURCE_DIR}/deps/boost.cmake)
```
gumb0: Maybe deps/.git would be more general?

chfast: Yes, good idea.

gcolvin:
Thanks!

PR #3585, opened 2017-02-22
Fix infinite loop during import from rlp file

gumb0:


eth/main.cpp
```diff
@@ -1108,10 +1108,11 @@ int main(int argc, char** argv)
 			}
 		}
 
-		while (web3.ethereum()->blockQueue().items().first + web3.ethereum()->blockQueue().items().second > 0)
+		bool moreToImport = true;
+		while (moreToImport)
 		{
 			this_thread::sleep_for(chrono::seconds(1));
-			web3.ethereum()->syncQueue(100000);
+			tie(ignore, moreToImport, ignore) = web3.ethereum()->syncQueue(100000);
```
gumb0: `Client` uses this value returned from `syncQueue` to determine if regular network sync needs to continue, import here should better do the same, I think.

eth/main.cpp
```diff
@@ -1108,10 +1108,11 @@ int main(int argc, char** argv)
 			}
 		}
 
-		while (web3.ethereum()->blockQueue().items().first + web3.ethereum()->blockQueue().items().second > 0)
```
gumb0: This check was remaining true forever in case when 1 block is added to BlockQueue, is valid but with unknown parent.

eth/main.cpp
gumb0: It is basically the same as checking only `web3.ethereum()->blockQueue().items().first > 0`

holiman:
Just to confirm; this fixes my issue when running imports through hive

PR #3587, opened 2017-02-22
Integrate libsnark

chriseth:
Implements https://github.com/ethereum/EIPs/pull/212 and https://github.com/ethereum/EIPs/pull/213
Fixes https://github.com/ethereum/cpp-ethereum/issues/3619#event-985648352

This is still work in progress.

 - [x] change point encoding
 - [x] remove unnecessary stuff
 - [x] make this work for truncated input
 - [x] we can probably patch libsnark to remove all cpp files outside the directory `src/algebra/curves/alt_bn128/` - this requires patching the makefile.
 - [x] add source hash for external dependency
 - [x] move unittest file to the right place
 - [x] clean up git history
 - [x] check edge cases for pairing


chriseth:
@chfast please take over from here :-)

winsvega:
please rebase

winsvega:
please rebase on develop so to point this PR to a test branch

chfast:
Progress: it builds on Windows, but tests fail.

libethashseal/genesis/metropolisTest.cpp
```diff
@@ -57,7 +57,10 @@ R"E(
 		"0000000000000000000000000000000000000002": { "wei": "1", "precompiled": { "name": "sha256", "linear": { "base": 60, "word": 12 } } },
 		"0000000000000000000000000000000000000003": { "wei": "1", "precompiled": { "name": "ripemd160", "linear": { "base": 600, "word": 120 } } },
 		"0000000000000000000000000000000000000004": { "wei": "1", "precompiled": { "name": "identity", "linear": { "base": 15, "word": 3 } } },
-		"0000000000000000000000000000000000000005": { "wei": "1", "precompiled": { "name": "modexp" } }
+		"0000000000000000000000000000000000000005": { "wei": "1", "precompiled": { "name": "modexp" } } },
```
pirapira: This line has one too many `}`.  Really.

pirapira:
There is an input that crashes `libff` (Pairing with the zero-element of G2).  The fix is https://github.com/pirapira/cpp-ethereum/commit/958a69c675b17a27ef016ffc4df78efbf96b83ab

pirapira:
I prepared some tests for snark integration https://github.com/ethereum/tests/pull/171/files

libdevcrypto/LibSnark.cpp
```diff
@@ -0,0 +1,212 @@
...
+using namespace dev::crypto;
+
+namespace
+{
+
+void initLibSnark()
+{
+	// This is hackish, but we really want to use `static` variable for lock
+	// free thread-safe initialization.
+	static bool initialized = (libff::alt_bn128_pp::init_public_params(), true);
```
pirapira: So is `init_public_params()` supposed to be called just once?

pirapira: Maybe we should first say `if (!initialized)` in the beginning.  Maybe we need to compare-and-swap `initialized`.

chfast: Yes, good idea.
You can use [std::call_ones](http://en.cppreference.com/w/cpp/thread/call_once) as a start.

pirapira: That's something new to me.  Let me try this.

chfast: Alternatively, you can wrap the function to add some logs to make sure it is called, and it is called once.

libdevcrypto/LibSnark.cpp
```diff
@@ -0,0 +1,217 @@
...
+	// is too short.
+	h256 xbin(_data, h256::AlignLeft);
+	if (u256(xbin) >= u256(fromLibsnarkBigint(libff::alt_bn128_Fq::mod)))
+		BOOST_THROW_EXCEPTION(InvalidEncoding());
+	return toLibsnarkBigint(xbin);
+}
+
+libff::alt_bn128_G1 decodePointG1(dev::bytesConstRef _data)
+{
+	libff::alt_bn128_Fq X = decodeFqElement(_data.cropped(0));
```
pirapira: I think it's better to specify the length as `cropped(0, 32)`.

libdevcrypto/LibSnark.cpp
```diff
@@ -0,0 +1,217 @@
...
+	return x;
+}
+
+libff::alt_bn128_Fq decodeFqElement(dev::bytesConstRef _data)
+{
+	// h256::AlignLeft ensures that the h256 is zero-filled on the right if _data
+	// is too short.
+	h256 xbin(_data, h256::AlignLeft);
+	if (u256(xbin) >= u256(fromLibsnarkBigint(libff::alt_bn128_Fq::mod)))
+		BOOST_THROW_EXCEPTION(InvalidEncoding());
```
pirapira: I'll add a more informative message here (especially the value of the `modulo`).

libdevcrypto/LibSnark.cpp
pirapira: No, now I see that the code is supposed to work even when `_data` is shorter than 64 bytes.

libdevcrypto/LibSnark.cpp
```diff
@@ -0,0 +1,221 @@
...
+}
+
+libff::bigint<libff::alt_bn128_q_limbs> toLibsnarkBigint(h256 const& _x)
+{
+	libff::bigint<libff::alt_bn128_q_limbs> x;
+	auto const N = x.N;
+	constexpr auto L = sizeof(x.data[0]);
+	for (unsigned i = 0; i < N; i++)
+		for (unsigned j = 0; j < L; j++)
+			x.data[N - 1 - i] |= uint64_t(_x[i * L + j]) << (8 * (L - 1 - j));
```
pirapira: This might be a problem when `x.data[0]` is wider than 64bit, but I'm not sure if we should go for
```
            x.data[N - 1 - i] |= (std::remove_reference<decltype(x.data[0])>::type)(_x[i * L + j]) << (8 * (L - 1 - j));
```

pirapira: I guess I should just say `libff::mp_limb_t`.

pirapira:
On my Mac with `The CXX compiler identification is AppleClang 8.1.0.8020042`, I'm seeing

```
/Users/yh/shared/cpp-ethereum/deps/src/snark/src/algebra/curves/alt_bn128/alt_bn128_g1.cpp:411:9: fatal error: invalid operands to binary expression ('std::ostream' (aka 'basic_ostream<char>') and 'alt_bn128_Fq' (aka 'Fp_model<alt_bn128_q_limbs, alt_bn128_modulus_q>'))
    out << copy.X << OUTPUT_SEPARATOR << copy.Y;
    ~~~ ^  ~~~~~~
```

pirapira:
The error message is the same as https://github.com/scipr-lab/libff/issues/4

test/libdevcrypto/LibSnark.cpp
```diff
@@ -0,0 +1,470 @@
...
+		Fq("21341087976609916409401737322664290631992568431163400450267978471171152600502"),
+		Fq("2942165230690572858696920423896381470344658299915828986338281196715687693170"),
+		Fq::one()
+	);
+
+
+	cout << "Verifying..." << endl;
+	bool verified = libsnark::r1cs_ppzksnark_verifier_strong_IC(_vk, primary_input, proof);
+	cout << "Verified: " << verified << endl;
+}
```
pirapira: Why is this commented out?  Does it make sense to turn this into a BOOST test case?

test/libdevcrypto/LibSnark.cpp
```diff
@@ -0,0 +1,470 @@
...
+	cout << "\t\tif (!Pairing.pairingProd2(proof.C, vk.C, Pairing.negate(proof.C_p), Pairing.P2())) return false;" << endl;
+	cout << "\t\tif (!Pairing.pairingProd3(proof.K, vk.gamma, Pairing.negate(Pairing.add(vk_x, Pairing.add(proof.A, proof.C))), vk.gammaBeta2, Pairing.negate(vk.gammaBeta1), proof.B)) return false;" << endl;
+	cout << "\t\tif (!Pairing.pairingProd3(" << endl;
+	cout << "\t\t\t\tPairing.add(vk_x, proof.A), proof.B," << endl;
+	cout << "\t\t\t\tPairing.negate(proof.H), vk.Z," << endl;
+	cout << "\t\t\t\tPairing.negate(proof.C), Pairing.P2()" << endl;
+	cout << "\t\t)) return false;" << endl;
+	cout << "\t\treturn true;" << endl;
+	cout << "\t}" << endl;
+}
```
pirapira: Why is this commented out?  Does it make sense to turn this into a C++ template?

libdevcrypto/LibSnark.cpp
```diff
@@ -26,54 +23,51 @@
...
-			libff::alt_bn128_pp::init_public_params();
-		});
+	static bool s_initialized = []() noexcept
+	{
+		libff::inhibit_profiling_info = true;
+		libff::inhibit_profiling_counters = true;
+		libff::alt_bn128_pp::init_public_params();
+		return true;
+	}();
+	(void)s_initialized;
```
pirapira: I'm wondering why we need this boolean that we never read.  Maybe doing this is enough?
```
[]() noexcept
{
		libff::inhibit_profiling_info = true;
		libff::inhibit_profiling_counters = true;
		libff::alt_bn128_pp::init_public_params();
}();
```

Or, do we even need the closure?

test/libdevcrypto/LibSnark.cpp
pirapira: The above comment is a leftover of something obsolete.

libdevcrypto/LibSnark.cpp
chfast: We need any `static` variable to use thread-safe static initialization (aka magic statics). We can use `call_once` here explicitly, but static vars are handled better by compilers.
What do you think? Go back to `call_once`?

pirapira: No need to go back; I just wondered.

pirapira:
In the EIP PR, I asked for some clarification about the input length. https://github.com/ethereum/EIPs/pull/212#discussion_r121099919

I checked some other clients, and all of them required `input_length % 192 == 0` for the pairing checker.

libdevcrypto/LibSnark.cpp
```diff
@@ -0,0 +1,213 @@
...
+	return b;
+}
+
+h256 fromLibsnarkBigint(libff::bigint<libff::alt_bn128_q_limbs> const& _b)
+{
+	auto const N = _b.N;
+	constexpr size_t L = sizeof(_b.data[0]);
+	static_assert(sizeof(mp_limb_t) == L, "Unexpected limb size in libff::bigint.");
+	h256 x;
+	for (size_t i = 0; i < N; i++)
```
pirapira: I get the following error.  Maybe it helps to specify the type of `N`.
```
libdevcrypto/LibSnark.cpp:68:23: error: comparison of integers of different signs: 'size_t' (aka 'unsigned long') and 'const long' [-Werror,-Wsign-compare]
        for (size_t i = 0; i < N; i++)
```

libdevcrypto/LibSnark.cpp
```diff
@@ -0,0 +1,213 @@
...
+	return libff::alt_bn128_Fq2(
+		decodeFqElement(_data.cropped(32)),
+		decodeFqElement(_data.cropped(0))
+	);
+}
+
+libff::alt_bn128_G2 decodePointG2(dev::bytesConstRef _data)
+{
+	libff::alt_bn128_Fq2 x = decodeFq2Element(_data);
+	libff::alt_bn128_Fq2 y = decodeFq2Element(_data.cropped(64));
```
pirapira: `x` and `y` can be `const`.

libdevcrypto/LibSnark.cpp
```diff
@@ -0,0 +1,213 @@
...
+	if (x == libff::alt_bn128_Fq2::zero() && y == libff::alt_bn128_Fq2::zero())
+		return libff::alt_bn128_G2::zero();
+	libff::alt_bn128_G2 p(x, y, libff::alt_bn128_Fq2::one());
+	if (!p.is_well_formed())
```
pirapira: `p` can also be `const`.

libdevcrypto/LibSnark.cpp
```diff
@@ -0,0 +1,213 @@
...
+	// Input: list of pairs of G1 and G2 points
+	// Output: 1 if pairing evaluates to 1, 0 otherwise (left-padded to 32 bytes)
+
+	size_t constexpr pairSize = 2 * 32 + 2 * 64;
+	size_t const pairs = _in.size() / pairSize;
+	if (pairs * pairSize != _in.size())
+	{
+		// Invalid length.
+		return {false, bytes{}};
+	}
```
pirapira: Style: we don't have braces around a single-line body of `if`.

libdevcrypto/LibSnark.cpp
```diff
@@ -0,0 +1,213 @@
...
+	try
+	{
+		initLibSnark();
+		libff::alt_bn128_Fq12 x = libff::alt_bn128_Fq12::one();
+		for (size_t i = 0; i < pairs; ++i)
+		{
+			dev::bytesConstRef pair = _in.cropped(i * pairSize, pairSize);
+			libff::alt_bn128_G2 p = decodePointG2(pair.cropped(2 * 32));
```
pirapira: `pair` and `p` can be `const`.

libdevcrypto/LibSnark.cpp
```diff
@@ -0,0 +1,213 @@
...
+		libff::alt_bn128_G1 const p1 = decodePointG1(_in);
+		libff::alt_bn128_G1 const p2 = decodePointG1(_in.cropped(32 * 2));
+		return {true, encodePointG1(p1 + p2)};
+	}
+	catch (InvalidEncoding const&)
+	{
+	}
+	catch (...)
+	{
+		cwarn << "Internal exception from libsnark. Forwarding as precompiled contract failure.";
```
pirapira: Perhaps, we should look into the library, and see which exception is supposed to happen.

chfast:
Ok, I will handle these.

libdevcrypto/LibSnark.cpp
chfast: It does not throw any exceptions.

pirapira: Should we silence unexpected exceptions here?

chfast: I prefer not to. It's better to terminate the node in this case than running it in invalid state mode. Agree?

pirapira: I agree.

chfast:
Great! Waiting for CI tests...

pirapira:
All green.

PR #3588, opened 2017-02-22
Code coverage

chfast:


.travis.yml
```diff
@@ -111,7 +111,8 @@ before_script:
     - if type ccache; then ccache -s; fi
 script:
     - cd $TRAVIS_BUILD_DIR/build && ../scripts/tests.sh $TRAVIS_TESTS
-
+after_success:
+    - cd $TRAVIS_BUILD_DIR/build && bash <(curl -s https://codecov.io/bash) -a '>/dev/null 2>&1'
```
chriseth: This basically grants the people at codecov access to our deployment keys. Is that bad?

chfast: I didn't think about it. So maybe we get the script from fixed Git commit? 

https://raw.githubusercontent.com/codecov/codecov-bash/f81a7b47aef0fad3900e887a7ad9c96c0d12c47f/codecov

chfast:
The codecov script is not taken from GH and fixed commit. How it looks now?

PR #3592, opened 2017-02-24
Success condition for precompiled contracts.

chriseth:


libethereum/Executive.cpp
```diff
@@ -262,9 +262,9 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
...
 	m_savepoint = m_s.savepoint();
 
-	if (m_sealEngine.isPrecompiled(_p.codeAddress))
+	if (m_sealEngine.isPrecompiled(_p.codeAddress, m_envInfo.number()))
 	{
-		bigint g = m_sealEngine.costOfPrecompiled(_p.codeAddress, _p.data);
+		bigint g = m_sealEngine.costOfPrecompiled(_p.codeAddress, _p.data, m_envInfo.number());
 		if (_p.gas < g)
 		{
 			m_excepted = TransactionException::OutOfGasBase;
```
chriseth: @chfast do you know why this does not do `m_gas = 0`?

chfast: Probably setting excepted flag is enough... As you can see we have probably too much layers here :).
Try setting it to 0. If no broken tests we can keep it.

chriseth:
I used `pair<bool,bytes>` because of the upcoming EIP for the revert opcode.

chfast:
How is this related?

chriseth:
The revert opcode allows returning data in the failure case. So in the future, it looks like we have the following parameters:
 - failure or not
 - returned data
 - remaining gas (because we do not out-of-gas for all failure cases)

PR #3593, opened 2017-02-24
testeth refactoring

winsvega:
split large TestHelper.cpp into separate files.

there was several class realizations in one single TestHelper.cpp making it about 1000+ lines long. 
So the point of this PR is to separate those realization. like ImportTest, TestOutputHelper, Options  into different files and headers.


test/CMakeLists.txt
```diff
@@ -9,11 +9,12 @@ macro (add_sources)
 	endforeach()
 endmacro()
 
+add_subdirectory(libtesteth)
 add_subdirectory(fuzzTesting)
 add_subdirectory(libethcore)
 add_subdirectory(libethereum)
-if (NOT EMSCRIPTEN)		
-    add_subdirectory(libtestutils)		
+if (NOT EMSCRIPTEN)
```
chfast: You can remove EMSCRIPTEN check. It was for solidity.

test/fuzzTesting/fuzzHelper.cpp
```diff
@@ -19,12 +19,12 @@
...
  */
 
-#include "fuzzHelper.h"
-
 #include <chrono>
 #include <boost/random.hpp>
 #include <boost/filesystem/path.hpp>
 #include <libevmcore/Instruction.h>
+#include <test/libtesteth/TestOutputHelper.h>
+#include <test/fuzzTesting/fuzzHelper.h>
```
chfast: There is a rule to include matching header (name.cpp -> name.hpp) as the first include (as it was here). I think we should keep it.

test/libtesteth/TestOutputHelper.h
```diff
@@ -0,0 +1,57 @@
...
+
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/** @file TestOutputHelper.h
```
chfast: Don't repeat file name after `@file`!

winsvega: no filenames after @file  ?

chfast: It is optional and should be only needed when documenting **other** file. https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdfile

PR #3594, opened 2017-02-27
Use int64 for gas in Transaction::baseGasRequired

chfast:
This is small change towards using int64 for gas type where possible. I will send small changes overtime not to scare @gcolvin.

### DIFF
--- a/libethcore/Transaction.cpp
+++ b/libethcore/Transaction.cpp
@@ -166,6 +166,10 @@ void TransactionBase::checkChainId(int chainId) const
 int64_t TransactionBase::baseGasRequired(bool _contractCreation, bytesConstRef _data, EVMSchedule const& _es)
 {
 	int64_t g = _contractCreation ? _es.txCreateGas : _es.txGas;
+
+	// Calculate the cost of input data.
+	// No risk of overflow by using int64 until txDataNonZeroGas is quite small
+	// (the value not in billions).
 	for (auto i: _data)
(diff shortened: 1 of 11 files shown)


libethcore/Transaction.cpp
```diff
@@ -163,12 +163,12 @@ void TransactionBase::checkChainId(int chainId) const
 		BOOST_THROW_EXCEPTION(InvalidSignature());
 }
 
-bigint TransactionBase::gasRequired(bool _contractCreation, bytesConstRef _data, EVMSchedule const& _es, u256 const& _gas)
+int64_t TransactionBase::baseGasRequired(bool _contractCreation, bytesConstRef _data, EVMSchedule const& _es)
 {
-	bigint ret = (_contractCreation ? _es.txCreateGas : _es.txGas) + _gas;
+	int64_t g = _contractCreation ? _es.txCreateGas : _es.txGas;
 	for (auto i: _data)
```
chriseth: Why will this not overflow the sum?

chfast: The 63-bit precision is big enough. The cost is 68 gas units per byte, so you will need 115 PB of input to overflow.

I also dream about using Undefined Behavior Sanitizer (aka UBSan) at some point. It will detect signed integer overflows at runtime. That's also a reason for using signed integers for gas calculations.

chfast: I thinks I should add a comment about it. And it is consensus code -- Executive uses it.

I will add the comment and merge.

Issue #3546, opened 2017-02-07
Account abstraction - EIP 86


chriseth:
Implement https://github.com/ethereum/EIPs/issues/86

TODO
 - [x] Transactions with signature (0, 0, 0) (ie. v = r = s = 0 0) are valid and msg.sender is set to uint120(-1) during execution.
 - [x] Accept those transactions if they are coming in from blocks, but do not forward such transactions if they are coming in not yet in a block.

gumb0:
Reminder: we need to disable incrementing the nonce of the sender in case of zero signature

pirapira:
Reminder: we need to change the way the addresses of new contracts are calculated.

Reminder: we need to add CREATE_P2SH instruction.

yann300:
@pirapira yep I think I could do that next week.

chfast:
What do we do with this?

pirapira:
@chfast we leave it until there is a decision whether Constantinople includes this or not.

pirapira:
@chfast since there is nothing to do currently, we can close this issue, I think.

chfast:
Based on AllCoreDev https://github.com/ethereum/pm/issues/52, the EIP-86 is probably not going to be accepted in this "big" variant.

PR #3597, opened 2017-02-27
EIP86

yann300:
https://github.com/ethereum/cpp-ethereum/issues/3546

### DIFF
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -79,6 +79,8 @@ Public SignatureStruct::recover(h256 const& _hash) const
 
 Address dev::ZeroAddress = Address();
 
+Address dev::LastAddress = Address("0xffffffffffffffffffffffffffffffffffffffff");
+
 Public dev::toPublic(Secret const& _secret)
 {
 	bytes o(65);
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -67,6 +67,13 @@ bool dev::SignatureStruct::isValid() const noexcept
 	return true;
 }
 
+bool dev::SignatureStruct::zeroSignature() const noexcept
+{
+	if (r == h256(0) && s == h256(0) && v == 0)
+		return true;
+	return false;
+}
+
 Public SignatureStruct::recover(h256 const& _hash) const
 {
 	return dev::recover((Signature)*this, _hash);
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -69,9 +69,7 @@ bool dev::SignatureStruct::isValid() const noexcept
 
 bool dev::SignatureStruct::zeroSignature() const noexcept
 {
-	if (r == h256(0) && s == h256(0) && v == 0)
-		return true;
-	return false;
+	return (r == h256(0) && s == h256(0) && v == 0);
 }
 
 Public SignatureStruct::recover(h256 const& _hash) const
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -67,11 +67,6 @@ bool dev::SignatureStruct::isValid() const noexcept
 	return true;
 }
 
-bool dev::SignatureStruct::isZeroSignature() const noexcept
-{
-	return (!r && !s && !v);
-}
-
 Public SignatureStruct::recover(h256 const& _hash) const
 {
 	return dev::recover((Signature)*this, _hash);
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -69,7 +69,7 @@ bool dev::SignatureStruct::isValid() const noexcept
 
 bool dev::SignatureStruct::isZeroSignature() const noexcept
 {
-	return (r == h256(0) && s == h256(0) && v == 0);
+	return (!r && !s && !v);
 }
 
 Public SignatureStruct::recover(h256 const& _hash) const
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -79,7 +79,7 @@ Public SignatureStruct::recover(h256 const& _hash) const
 
 Address dev::ZeroAddress = Address();
 
-Address dev::LastAddress = Address("0xffffffffffffffffffffffffffffffffffffffff");
+Address dev::MaxAddress = Address("0xffffffffffffffffffffffffffffffffffffffff");
 
 Public dev::toPublic(Secret const& _secret)
 {
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -67,7 +67,7 @@ bool dev::SignatureStruct::isValid() const noexcept
 	return true;
 }
 
-bool dev::SignatureStruct::zeroSignature() const noexcept
+bool dev::SignatureStruct::isZeroSignature() const noexcept
 {
 	return (r == h256(0) && s == h256(0) && v == 0);
 }
--- a/libdevcrypto/Common.h
+++ b/libdevcrypto/Common.h
@@ -54,7 +54,7 @@ struct SignatureStruct
 	bool isValid() const noexcept;
 
 	/// @returns true if r = 0 && s = 0 && v =0
-	bool zeroSignature() const noexcept;
+	bool isZeroSignature() const noexcept;
 
 	/// @returns the public part of the key that signed @a _hash to give this sig.
 	Public recover(h256 const& _hash) const;
--- a/libdevcrypto/Common.h
+++ b/libdevcrypto/Common.h
@@ -53,6 +53,9 @@ struct SignatureStruct
 	/// @returns true if r,s,v values are valid, otherwise false
 	bool isValid() const noexcept;
 
+	/// @returns true if r = 0 && s = 0 && v =0
+	bool zeroSignature() const noexcept;
+
 	/// @returns the public part of the key that signed @a _hash to give this sig.
 	Public recover(h256 const& _hash) const;
 
--- a/libdevcrypto/Common.h
+++ b/libdevcrypto/Common.h
@@ -72,7 +72,7 @@ using Address = h160;
 extern Address ZeroAddress;
 
 /// The last address.
-extern Address LastAddress("0xffffffffffffffffffffffffffffffffffffffff");
+extern Address LastAddress;
 
 /// A vector of Ethereum addresses.
 using Addresses = h160s;
--- a/libdevcrypto/Common.h
+++ b/libdevcrypto/Common.h
@@ -72,7 +72,7 @@ using Address = h160;
 extern Address ZeroAddress;
 
 /// The last address.
-extern Address LastAddress;
+extern Address MaxAddress;
 
 /// A vector of Ethereum addresses.
 using Addresses = h160s;
--- a/libdevcrypto/Common.h
+++ b/libdevcrypto/Common.h
@@ -71,6 +71,9 @@ using Address = h160;
 /// The zero address.
 extern Address ZeroAddress;
 
+/// The last address.
+extern Address LastAddress("0xffffffffffffffffffffffffffffffffffffffff");
+
 /// A vector of Ethereum addresses.
 using Addresses = h160s;
 
--- a/libdevcrypto/Common.h
+++ b/libdevcrypto/Common.h
@@ -53,9 +53,6 @@ struct SignatureStruct
 	/// @returns true if r,s,v values are valid, otherwise false
 	bool isValid() const noexcept;
 
-	/// @returns true if r = 0 && s = 0 && v =0
-	bool isZeroSignature() const noexcept;
-
 	/// @returns the public part of the key that signed @a _hash to give this sig.
 	Public recover(h256 const& _hash) const;
 
--- a/libethashseal/Ethash.cpp
+++ b/libethashseal/Ethash.cpp
@@ -150,8 +150,6 @@ void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase co
 
 	if (_ir & ImportRequirements::TransactionSignatures)
 	{
-		if (_bi.number() >= chainParams().u256Param("homsteadForkBlock"))
-			_t.checkLowS();
 		if (_bi.number() >= chainParams().u256Param("EIP158ForkBlock"))
 		{
 			int chainID(chainParams().u256Param("chainID"));
--- a/libethashseal/Ethash.cpp
+++ b/libethashseal/Ethash.cpp
@@ -157,6 +157,8 @@ void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase co
 		}
 		else
 			_t.checkChainId(-4);
+		if (_bi.number() < chainParams().u256Param("EIP86ForkBlock") && _t.hasZeroSignature())
+			BOOST_THROW_EXCEPTION(InvalidSignature());
 	}
 	// Unneeded as it's checked again in Executive. Keep it here since tests assume it's checked.
 	if (_ir & ImportRequirements::TransactionBasic && _t.gasRequired(evmSchedule(EnvInfo(_bi))) > _t.gas())
--- a/libethashseal/Ethash.cpp
+++ b/libethashseal/Ethash.cpp
@@ -157,7 +157,7 @@ void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase co
 		}
 		else
 			_t.checkChainId(-4);
-		if (_bi.number() < chainParams().u256Param("EIP86ForkBlock") && _t.hasZeroSignature())
+		if (_bi.number() < chainParams().u256Param("metropolisForkBlock") && _t.hasZeroSignature())
 			BOOST_THROW_EXCEPTION(InvalidSignature());
 	}
 	// Unneeded as it's checked again in Executive. Keep it here since tests assume it's checked.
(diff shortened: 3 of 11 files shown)


libdevcrypto/Common.cpp
```diff
@@ -58,6 +58,9 @@ class Secp256k1Context
 
 bool dev::SignatureStruct::isValid() const noexcept
 {
+	if (zeroSign()) {
```
chriseth: Please be careful, this function is used inside the ecrecover precompile whose behaviour should not be changed.
I think it might be better to keep `isValid()` returning false and treat the zero-signature as a special case from the outside.

libdevcrypto/Common.cpp
```diff
@@ -67,6 +70,14 @@ bool dev::SignatureStruct::isValid() const noexcept
 	return true;
 }
 
+bool dev::SignatureStruct::zeroSign() const noexcept
```
chriseth: `zeroSignature`?

libdevcrypto/Common.cpp
```diff
@@ -67,6 +70,14 @@ bool dev::SignatureStruct::isValid() const noexcept
...
+{
+	if (r == h256(0) && s == h256(0) && v == 0) {
```
chriseth: Style: Don't use braces for single-line body and if you use braces, move them to the next line.

libethcore/Transaction.h
```diff
@@ -142,6 +142,8 @@ class TransactionBase
 	/// @returns the signature of the transaction. Encodes the sender.
 	SignatureStruct const& signature() const { return m_vrs; }
 
+	bool isForwardingCall() const { return m_vrs.zeroSign(); }
```
chriseth: Not sure if this should be called a forwarding call. Perhaps just `hasZeroSignature`?

libdevcrypto/Common.cpp
```diff
@@ -67,6 +67,13 @@ bool dev::SignatureStruct::isValid() const noexcept
 	return true;
 }
 
+bool dev::SignatureStruct::zeroSignature() const noexcept
+{
+	if (r == h256(0) && s == h256(0) && v == 0)
```
chriseth: What about `return r == h256(0) && s == h256(0) && v == 0;`?

libethcore/Transaction.cpp
```diff
@@ -112,9 +112,9 @@ Address const& TransactionBase::sender() const
 	if (!m_sender)
 	{
 		auto p = recover(m_vrs, sha3(WithoutSignature));
```
chriseth: If we have a zero-signature, this calls `recover` for each call of `sender`. Please cache in both situations.

libethcore/Transaction.cpp
```diff
@@ -112,9 +112,9 @@ Address const& TransactionBase::sender() const
...
-		if (!p)
-			BOOST_THROW_EXCEPTION(InvalidSignature());
-		m_sender = right160(dev::sha3(bytesConstRef(p.data(), sizeof(p))));
+		if (p) {
```
chriseth: No braces for single statement.

libethereum/Executive.cpp
```diff
@@ -198,25 +198,28 @@ void Executive::initialize(Transaction const& _transaction)
...
+			m_excepted = TransactionException::InvalidSignature;
+			throw;
+		}
+		if (m_t.nonce() != nonceReq)
+		{
+			clog(ExecutiveWarnChannel) << "Invalid Nonce: Require" << nonceReq << " Got" << m_t.nonce();
+			m_excepted = TransactionException::InvalidNonce;
+			BOOST_THROW_EXCEPTION(InvalidNonce() << RequirementError((bigint)nonceReq, (bigint)m_t.nonce()));
+		}
+	} else {
```
chriseth: This should also have some activation block number. Also: Braces on their own line, no braces for single statement.

libethereum/TransactionQueue.cpp
```diff
@@ -100,6 +100,8 @@ ImportResult TransactionQueue::check_WITH_LOCK(h256 const& _h, IfDropped _ik)
 
 ImportResult TransactionQueue::import(Transaction const& _transaction, IfDropped _ik)
 {
+	if (t.hasZeroSignature())
+		return ImportResult::ZeroSignature;
```
chriseth: I think we cannot abort that early, at least we have to check whether the transaction is already known. Furthermore, this does not actually import the transaction. Or is that the intention?

libethereum/TransactionQueue.cpp
```diff
@@ -57,34 +57,8 @@ TransactionQueue::~TransactionQueue()
...
-//			cdebug << "Importing" << t;
-			ir = manageImport_WITH_LOCK(h, t);
-		}
-		catch (...)
-		{
-			return ImportResult::Malformed;
-		}
-	}
-	return ir;
+	Transaction t = Transaction(_transactionRLP, CheckTransaction::Everything);
```
gumb0: `Transaction` constructor may throw if invalid RLP data is given.
In the previous version it was caught and `ImportResult::Malformed` returned

libdevcrypto/Common.h
```diff
@@ -53,6 +53,9 @@ struct SignatureStruct
 	/// @returns true if r,s,v values are valid, otherwise false
 	bool isValid() const noexcept;
 
+	/// @returns true if r = 0 && s = 0 && v =0
+	bool zeroSignature() const noexcept;
```
gumb0: `isZeroSignature` would better follow coding standards

libethereum/Executive.cpp
yann300: Do we need an activation block number here? cause we don't accept tx that are not already in a block.

libethereum/Executive.cpp
```diff
@@ -198,24 +198,29 @@ void Executive::initialize(Transaction const& _transaction)
...
+		}
+		if (m_t.nonce() != nonceReq)
+		{
+			clog(ExecutiveWarnChannel) << "Invalid Nonce: Require" << nonceReq << " Got" << m_t.nonce();
+			m_excepted = TransactionException::InvalidNonce;
+			BOOST_THROW_EXCEPTION(InvalidNonce() << RequirementError((bigint)nonceReq, (bigint)m_t.nonce()));
+		}
 	}
+	else
+		m_t.forceSender(LastAddress); // https://github.com/ethereum/EIPs/issues/86
```
gumb0: Could we instead put this logic inside `Transaction`, that is make `Transaction::sender()` return LastAddress in case `hasZeroSignature()` ?

test/libethereum/TransactionTests.cpp
```diff
@@ -63,7 +63,7 @@ void doTransactionTests(json_spirit::mValue& _v, bool _fillin)
 			try
 			{
 				Transaction txFromFields(rlpStream.out(), CheckTransaction::Everything);
-				if (!txFromFields.signature().isValid())
+				if (!txFromFields.signature().isValid() && !txFromFields.signature().isZeroSignature())
```
gumb0: I think we can put `isZeroSignature()` check inside `isValid()`, then we won't have to duplicate it three times (here and in `TransactionBase` constructor)

libethashseal/Ethash.cpp
```diff
@@ -157,6 +157,8 @@ void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase co
 		}
 		else
 			_t.checkChainId(-4);
+		if (_bi.number() < chainParams().u256Param("metropolisForkBlock") && _t.hasZeroSignature())
```
chriseth: Is such a signature actually invalid now or does it just resolve to a weird sender address?

libethcore/SealEngine.cpp
```diff
@@ -43,8 +43,10 @@ void SealEngineFace::populateFromParent(BlockHeader& _bi, BlockHeader const& _pa
 	_bi.populateFromParent(_parent);
 }
 
-void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, TransactionBase const& _t, BlockHeader const&) const
+void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, TransactionBase const& _t, BlockHeader const& _bi) const
 {
+	if (_bi.number() < chainParams().u256Param("metropolisForkBlock") && _t.hasZeroSignature())
```
chriseth: Same coment as above applies here.

libethashseal/Ethash.cpp
yann300: it is invalid now (before the fork) ... but the sender will be now set to 0xffffff...

pirapira: @chriseth it's unlikely, but I haven't tried.

gumb0: According to current `TransactionBase` constructor now `v` can't be 0 (will be rejected as invalid signature)

gumb0: We should better leave this check only in `SealEngineFace::verifyTransaction` (base class) and call `SealEngineFace::verifyTransaction(_ir, _t, _bi)` in the beginning of `Ethash::verifyTransaction`

gumb0: (Make sure then that `_t.checkLowS()` is called only after homestead, because now it's not the case in `SealEngineFace::verifyTransaction`)

libethashseal/Ethash.cpp
```diff
@@ -146,6 +146,8 @@ void Ethash::verify(Strictness _s, BlockHeader const& _bi, BlockHeader const& _p
 
 void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase const& _t, BlockHeader const& _bi) const
 {
+	SealEngineFace::verifyTransaction(_ir, _t, _bi);
+
 	if (_ir & ImportRequirements::TransactionSignatures)
 	{
 		if (_bi.number() >= chainParams().u256Param("homsteadForkBlock"))
```
gumb0: And this check can be removed now, since we check it in the base class

yann300: ok

libethereum/TransactionQueue.cpp
```diff
@@ -57,34 +57,16 @@ TransactionQueue::~TransactionQueue()
...
-			t = Transaction(_transactionRLP, CheckTransaction::Everything);
-			UpgradeGuard ul(l);
-//			cdebug << "Importing" << t;
-			ir = manageImport_WITH_LOCK(h, t);
-		}
-		catch (...)
-		{
-			return ImportResult::Malformed;
-		}
+		Transaction(_transactionRLP, CheckTransaction::Everything);
```
chfast: Eee... I guess you should assign it to `t`.

libethereum/TransactionQueue.cpp
```diff
@@ -57,34 +57,16 @@ TransactionQueue::~TransactionQueue()
...
+	}
+	catch (...)
```
chfast: Please catch some type. Do we know what is expected from Transaction constructor? If that set is small, can we move the `return ...` here too?

libethereum/Executive.cpp
```diff
@@ -198,23 +198,26 @@ void Executive::initialize(Transaction const& _transaction)
...
-		clog(ExecutiveWarnChannel) << "Invalid Nonce: Require" << nonceReq << " Got" << m_t.nonce();
-		m_excepted = TransactionException::InvalidNonce;
-		BOOST_THROW_EXCEPTION(InvalidNonce() << RequirementError((bigint)nonceReq, (bigint)m_t.nonce()));
+		// Avoid invalid transactions.
+		u256 nonceReq;
+		try
+		{
+			nonceReq = m_s.getNonce(m_t.sender());
+		}
+		catch (...)
```
chfast: Please try to catch specific exception.

yann300: well `sender()` and `getNonce()` are not throwing specific exception

libethereum/TransactionQueue.cpp
yann300: yep :)

libethereum/Executive.cpp
gumb0: `sender()` throws `InvalidSignature`, you can catch it like `catch (InvalidSignature const&)`

chfast: `State::getNonce()` does not throw, so don't try to catch anything.

gumb0: @chfast `sender()` throws here

libethcore/Transaction.cpp
```diff
@@ -111,10 +112,14 @@ Address const& TransactionBase::sender() const
...
-		auto p = recover(m_vrs, sha3(WithoutSignature));
-		if (!p)
-			BOOST_THROW_EXCEPTION(InvalidSignature());
-		m_sender = right160(dev::sha3(bytesConstRef(p.data(), sizeof(p))));
+		if (hasZeroSignature())
+			m_sender = MaxAddress;
+		else
+		{
+			auto p = recover(m_vrs, sha3(WithoutSignature));
+			if (p)
```
chfast: Important: When can `recover()` fail?

If the question is answered, you can do this in single line: `m_sender = recover(...)`. In case of failure `m_sender` will be still "null".

libethereum/Executive.cpp
gumb0: Oh yeah, it does not not, I was looking on the old version, sorry

libethcore/Transaction.cpp
yann300: I pushed changes

libethereum/TransactionQueue.cpp
```diff
@@ -57,34 +57,16 @@ TransactionQueue::~TransactionQueue()
...
-			t = Transaction(_transactionRLP, CheckTransaction::Everything);
-			UpgradeGuard ul(l);
-//			cdebug << "Importing" << t;
-			ir = manageImport_WITH_LOCK(h, t);
-		}
-		catch (...)
-		{
-			return ImportResult::Malformed;
-		}
+		t = Transaction(_transactionRLP, CheckTransaction::Everything);
```
chfast: Move `t` declaration here. As `Transaction t(_transactionRLP, CheckTransaction::Everything);`.

libethcore/Transaction.cpp
```diff
@@ -111,10 +112,16 @@ Address const& TransactionBase::sender() const
...
-			BOOST_THROW_EXCEPTION(InvalidSignature());
-		m_sender = right160(dev::sha3(bytesConstRef(p.data(), sizeof(p))));
+		if (hasZeroSignature())
+			m_sender = MaxAddress;
+		else
+		{
+			auto p = recover(m_vrs, sha3(WithoutSignature));
+			if (!p)
+				BOOST_THROW_EXCEPTION(InvalidSignature());
+			else
```
chfast: Don't put `else` if you return or throw inside `if`.

libdevcrypto/Common.cpp
```diff
@@ -67,13 +67,20 @@ bool dev::SignatureStruct::isValid() const noexcept
 	return true;
 }
 
+bool dev::SignatureStruct::isZeroSignature() const noexcept
+{
+	return (r == h256(0) && s == h256(0) && v == 0);
```
chfast: You can save us some time by not putting `0` in the constructor. The default constructor creates "null" hash. You can do one of 4:
```
r == h256()
r == h256{}
r == {}
!r
```
Up to you. C++.

libethcore/SealEngine.cpp
```diff
@@ -43,9 +43,11 @@ void SealEngineFace::populateFromParent(BlockHeader& _bi, BlockHeader const& _pa
 	_bi.populateFromParent(_parent);
 }
 
-void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, TransactionBase const& _t, BlockHeader const&) const
+void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, TransactionBase const& _t, BlockHeader const& _bi) const
 {
-	if (_ir & ImportRequirements::TransactionSignatures)
+	if (_bi.number() < chainParams().u256Param("metropolisForkBlock") && _t.hasZeroSignature())
```
chriseth: Why don't we check for `_ir &ImportRequirements::TransactionSignatures` here?
Also: Is the zero signature considered invalid already now because of the v value?

test/libethereum/TransactionQueue.cpp
```diff
@@ -31,6 +31,20 @@ using namespace dev::test;
 
 BOOST_FIXTURE_TEST_SUITE(TransactionQueueSuite, TestOutputHelper)
 
+BOOST_AUTO_TEST_CASE(TransactionEIP86)
```
chfast: Yey. New unit test! Perfect!

winsvega:
having 
*** 131 failures are detected in the test module "Master Test Suite"

on transaction tests

yann300:
tests fixed

libethcore/SealEngine.cpp
```diff
@@ -43,9 +43,11 @@ void SealEngineFace::populateFromParent(BlockHeader& _bi, BlockHeader const& _pa
 	_bi.populateFromParent(_parent);
 }
 
-void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, TransactionBase const& _t, BlockHeader const&) const
+void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, TransactionBase const& _t, BlockHeader const& _bi) const
 {
-	if (_ir & ImportRequirements::TransactionSignatures)
+	if ((_ir & ImportRequirements::TransactionSignatures) && _bi.number() < chainParams().u256Param("metropolisForkBlock") && _t.hasZeroSignature())
```
chriseth: I know I asked that before, but I'm not sure I got a reply: Why is this not a consensus change before the metropolis fork block?

gumb0: Previously zero signature before metropolis was rejected in TransactionBase constructor during `m_vrs.isValid()` check.

This new check here is needed because now zero signature is always accepted in TransactionBase constructor without getting to `m_vrs.isValid()` check.
So without this new check it would be consensus change before metropolis (consensus about zero signature transactions)

chriseth: ok, thanks!

chfast:
It does not compile on Windows!!!!!!!!!!!!!!!!!!!!!!!!!

PR #3599, opened 2017-02-28
Add streaming operator for bytes.

chriseth:
This allows `BOOST_CHECK_EQUAL` to be used on bytes.

I don't know why the template for vector is not used, but this way it is nicer anyway.

### DIFF
--- a/libdevcore/CommonIO.h
+++ b/libdevcore/CommonIO.h
@@ -37,6 +37,7 @@
 #include <iostream>
 #include <chrono>
 #include "Common.h"
+#include "CommonData.h"
 #include "Base64.h"
 
 namespace dev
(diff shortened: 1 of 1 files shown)


libdevcore/CommonIO.h
```diff
@@ -76,6 +77,7 @@ std::string memDump(bytes const& _bytes, unsigned _width = 8, bool _html = false
 template <class S, class T> struct StreamOut { static S& bypass(S& _out, T const& _t) { _out << _t; return _out; } };
 template <class S> struct StreamOut<S, uint8_t> { static S& bypass(S& _out, uint8_t const& _t) { _out << (int)_t; return _out; } };
 
+inline std::ostream& operator<<(std::ostream& _out, bytes const& _e);
```
chfast: Why additional declaration?

chriseth: Right, was mislead by the other declarations.

chfast: Otherwise, looks good.

chriseth:
Actually it seems this does not solve the problem. Retrying after moving the operator to the global namespace.

chriseth:
OK, the problem with logging in boost test seems to be a more complicated issue. This change is useful nevertheless.

PR #3603, opened 2017-03-01
secp256k1 upgrade

chfast:
This is "just" libsecp256k1 upgrade to the commit mentioned in the CMake file. I also added some simple tests that helped me to update the API usages of the library.

I had to remove one test that was flaky. Seems that Crypto++ and libsecp256k1 don't agree about possible v values in the signature in the DER format.


chfast:
Of course, I forgot to set flags for MSVC...

chriseth:
Why did you add libsecp256k1 instead of using cmake dependencies?

chfast:
I replaced the old code. We had a copy of it exactly in `utils/secp256k1`. I didn't wanted to mix these two changes (upgrade + export to ExteranlProject), but looks like I should've. Or at least revert the order.

The issue with ExternalProject is secp256k1 does not support CMake, so it looks we will have inject CMakeLists.txt there... but should work. Should I do it here?

chriseth:
Yes, please do it in one go.

chfast:
Yey, I did it!

The issues is I had to include generated `ecmult_static_context.h` file. Maybe I will fix is some day.

I need to check Windows now.

cmake/secp256k1/src/ecmult_static_context.h
```diff
@@ -0,0 +1,1160 @@
+#ifndef _SECP256K1_ECMULT_STATIC_CONTEXT_
```
chriseth: Why is this file needed?

chfast: It is generated by a tool included in secp256k1. Instead of computing the context at runtime, it is computed at build time and include as static data.

chriseth: Why can't it be computed at build time but has to be part of the source?

chriseth: Ah now I see, you say you will maybe fix it some time. Ok.

chfast: I actually did it. Will push it soon....

chriseth: Wohoo!

chfast:
So beautiful now...

chriseth:
`+167 −8,673` - that's how I like it!

PR #3604, opened 2017-03-01
Add coverage report to alltests builds

chfast:


chriseth:
Is it possible to add a link to the result to the email?

chfast:
So... CodeCov UI does not have the option to filter by flags yet. See https://github.com/codecov/support/issues/235. You can attach link to the commit the report will be merged in. E.g. https://codecov.io/gh/ethereum/cpp-ethereum/commit/3e293a1133ea4fb9752227001201a2ddd5fc7e9e. But the "alltests" flag is also configured as "not joined". I'm not sure what will happen. Let's wait 2-3 days to get first reports...

PR #3605, opened 2017-03-02
runalltests.sh: workdir & other improvements

chfast:
Now you can run it like this:
```
cd /tmp
bash <( curl -s https://raw.githubusercontent.com/ethereum/cpp-ethereum/runalltests-rm-dirs/scripts/runalltests.sh)
```


winsvega:
hm it does work. but not with file operations
no log attached into email

PR #3606, opened 2017-03-02
make missing test file a warning

winsvega:
with this PR we could add test fillers before a PR that contains changes is merged into develop
thus just having warning messages instead of errors of missing tests

once new PR is merged we could fill actual tests and get rid of the warnings.

on the other hand if some tests are really missing or smth is screwed with testeth setup 
it would be just warnings for tests with fillers

### DIFF
--- a/test/libtesteth/TestHelper.cpp
+++ b/test/libtesteth/TestHelper.cpp
@@ -474,7 +474,12 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
 		cnote << "TEST " << name << ":";
 		json_spirit::mValue v;
 		string s = asString(dev::contents(testPath + "/" + name + ".json"));
-		BOOST_REQUIRE_MESSAGE(s.length() > 0, "Contents of " + testPath + "/" + name + ".json is empty. Have you cloned the 'tests' repo branch develop and set ETHEREUM_TEST_PATH to its path?");
+		BOOST_WARN_MESSAGE(s.length() > 0, "Missing test file!");
+		if(s.length() == 0)
+		{
+			TestOutputHelper::printWarn("Contents of " + testPath + "/" + name + ".json is empty. Have you cloned the 'tests' repo branch develop and set ETHEREUM_TEST_PATH to its path?");
+			return;
+		}
 		json_spirit::read_string(s, v);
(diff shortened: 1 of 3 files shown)


winsvega:
that would mean to change all the output system again. 
as it is now I just redirect all the standart output while tests are running 
to output log data through TestOutputHelper
then restore standart output again

test/libtesteth/TestHelper.cpp
```diff
@@ -474,7 +474,12 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
 		cnote << "TEST " << name << ":";
 		json_spirit::mValue v;
 		string s = asString(dev::contents(testPath + "/" + name + ".json"));
-		BOOST_REQUIRE_MESSAGE(s.length() > 0, "Contents of " + testPath + "/" + name + ".json is empty. Have you cloned the 'tests' repo branch develop and set ETHEREUM_TEST_PATH to its path?");
+		BOOST_WARN_MESSAGE(s.length() > 0, "Missing test file!");
```
chfast: Remove this warning. There is no use for it.

test/libtesteth/TestHelper.cpp
```diff
@@ -474,7 +474,12 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
...
+		if(s.length() == 0)
```
chfast: Fix style.

chfast: It also should be `s.empty()`.

test/libtesteth/TestHelper.cpp
```diff
@@ -474,7 +474,12 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
...
+		{
+			TestOutputHelper::printWarn("Contents of " + testPath + "/" + name + ".json is empty. Have you cloned the 'tests' repo branch develop and set ETHEREUM_TEST_PATH to its path?");
```
chfast: Can you make the message shorter? E.g.
```
file.json is empty. Set correct path in ETHEREUM_TEST_PATH environment variable.
```

chfast:
Can you address requested changes?

PR #3608, opened 2017-03-02
testpath for runalltests

winsvega:
easier option for https://github.com/ethereum/cpp-ethereum/pull/3605


chfast:
https://github.com/ethereum/cpp-ethereum/pull/3605 is not about fixing only this issue. It also fixes the issue I'm having issues locally...

PR #3610, opened 2017-03-03
New dependencies layout

chfast:


cmake/ProjectJsonCpp.cmake
```diff
@@ -0,0 +1,37 @@
...
+    URL https://github.com/open-source-parsers/jsoncpp/archive/1.7.7.tar.gz
+    URL_HASH SHA256=087640ebcf7fbcfe8e2717a0b9528fff89c52fcf69fa2a18cc2b538008098f97
+    CMAKE_COMMAND ${JSONCPP_CMAKE_COMMAND}
+    CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
+               # Build static lib but suitable to be included in a shared lib.
+               -DCMAKE_POSITION_INDEPENDENT_CODE=${BUILD_SHARED_LIBS}
+               -DJSONCPP_WITH_TESTS=Off
+               -DJSONCPP_WITH_PKGCONFIG_SUPPORT=Off
+    LOG_CONFIGURE 1
+    # Overwtire build and install commands to force Release build on MSVC.
```
chriseth: Typo :-)

cmake/scripts/helpers.cmake
```diff
@@ -0,0 +1,90 @@
+get_filename_component(ROOT_DIR "${CMAKE_CURRENT_LIST_DIR}/../.." ABSOLUTE)
+
+set(CACHE_DIR "${ROOT_DIR}/deps/src")
+set(PACKAGES_DIR "${ROOT_DIR}/packages")
+
+function(download URL DST_FILE STATUS)
```
chriseth: Why does the functionality offered by cmake not suffice here?

chfast: To be honest, I don't remember. It was my first contact with CMake scripting (this file is copied from cpp-dependencies, only needed by cpp-ethereum on Windows). It looks I added some kind of checking if the file is already there. ExternalProject does it nicely, but pure `file(DOWNLOAD ...)` maybe not.

chriseth:
Ok, good to go from my side.

chfast:
Ok. Let's spin it up.

Issue #3617, opened 2017-03-03
EIP 166: replay protection via high-order bits of nonce


pirapira:
Using the high-order bits of nonce to encode the chain ID https://github.com/ethereum/EIPs/issues/166

gumb0:
This EIP is no longer in the list planned for Metropolis https://github.com/ethereum/EIPs/blob/master/README.md

PR #3620, opened 2017-03-06
EIP166 - replay protection via high-order bits of nonce

gumb0:
https://github.com/ethereum/cpp-ethereum/issues/3617
https://github.com/ethereum/EIPs/issues/166

- [x] Add check for CHAIN_ID in transaction validation for transactions in the block
- [x] Add the check for pending transactions
- [x] When checking against sender nonce, use only lower 64 bits of tx nonce
- [x] Creating nonce for `eth.sendTransaction` RPC call

### DIFF
--- a/libethashseal/Ethash.cpp
+++ b/libethashseal/Ethash.cpp
@@ -158,9 +158,18 @@ void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase co
 		else
 			_t.checkChainId(-4);
 	}
-	// Unneeded as it's checked again in Executive. Keep it here since tests assume it's checked.
-	if (_ir & ImportRequirements::TransactionBasic && _t.baseGasRequired(evmSchedule(EnvInfo(_bi))) > _t.gas())
-		BOOST_THROW_EXCEPTION(OutOfGasIntrinsic());
+	if (_ir & ImportRequirements::TransactionBasic)
+	{
+		if (_bi.number() >= chainParams().u256Param("metropolisForkBlock"))
+		{
+			unsigned const nonceChainId(chainParams().u256Param("nonceChainID"));
+			_t.checkNonceChainId(nonceChainId);
+		}
+
+		// Unneeded as it's checked again in Executive. Keep it here since tests assume it's checked.
+		if (_t.baseGasRequired(evmSchedule(EnvInfo(_bi))) > _t.gas())
+			BOOST_THROW_EXCEPTION(OutOfGasIntrinsic());
+	}
 }
 
 u256 Ethash::childGasLimit(BlockHeader const& _bi, u256 const& _gasFloorTarget) const
--- a/libethashseal/Ethash.cpp
+++ b/libethashseal/Ethash.cpp
@@ -166,7 +166,7 @@ void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase co
 			_t.checkNonceChainId(nonceChainId);
 		}
 
-		// Unneeded as it's checked again in Executive. Keep it here since tests assume it's checked.
+		// TODO: Unneeded as it's checked again in Executive. Keep it here since tests assume it's checked.
 		if (_t.baseGasRequired(evmSchedule(EnvInfo(_bi))) > _t.gas())
 			BOOST_THROW_EXCEPTION(OutOfGasIntrinsic());
 	}
--- a/libethcore/Transaction.cpp
+++ b/libethcore/Transaction.cpp
@@ -31,6 +31,11 @@ using namespace std;
 using namespace dev;
 using namespace dev::eth;
 
+namespace
+{
+int const c_chainIdInNonceBits = 64;
+}
+
 TransactionBase::TransactionBase(TransactionSkeleton const& _ts, Secret const& _s):
 	m_type(_ts.creation ? ContractCreation : MessageCall),
 	m_nonce(_ts.nonce),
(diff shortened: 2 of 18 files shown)


gumb0:
I think some tests from https://github.com/ethereum/tests/blob/develop/TransactionTests/ttTransactionTest.json are failing now because they have transactions with nonce 0xfff...ff

chfast:
Just a side question: is it possible to reject transactions with gas > 2^63-1 straight away when parsing?

gumb0:
@chfast I think it's possible, since there're currently some checks in `TransactionBase` contructor (like checking signature validity and chainId encoded in v value)

libethereum/ClientBase.cpp
```diff
@@ -44,7 +45,13 @@ pair<h256, Address> ClientBase::submitTransaction(TransactionSkeleton const& _t,
 	TransactionSkeleton ts(_t);
 	ts.from = toAddress(_secret);
 	if (_t.nonce == Invalid256)
+	{
 		ts.nonce = max<u256>(postSeal().transactionsFrom(ts.from), m_tq.maxNonce(ts.from));
+		if (postSeal().info().number() >= bc().chainParams().u256Param("metropolisForkBlock"))
+		{
```
yann300: curving brace

libethashseal/Ethash.cpp
```diff
@@ -158,9 +158,18 @@ void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase co
 		else
 			_t.checkChainId(-4);
 	}
-	// Unneeded as it's checked again in Executive. Keep it here since tests assume it's checked.
-	if (_ir & ImportRequirements::TransactionBasic && _t.baseGasRequired(evmSchedule(EnvInfo(_bi))) > _t.gas())
-		BOOST_THROW_EXCEPTION(OutOfGasIntrinsic());
+	if (_ir & ImportRequirements::TransactionBasic)
```
yann300: That function is called by the `verifyBlock` , 
but seems transactions that comes from `onPeerTransaction` get accepted and thus relayed even if the `chainId` is wrong.
Would it make sense to check for the replay protection also at the 'network' layer?

gumb0: For unconfirmed transactions the check is in Executive.
(so they are checked when when we try to execute them and put into pending block)

gumb0: We can't check that on network level, because the check depends on which block the transaction will be put into
(depends on the block number)

gumb0: So currently we propagate all the transactions with incorrect chainID-in-nonce (which is fine, since they are currently valid)
We can add the check for chainID-in-nonce to TransactionQueue::import() later after the Metropolis fork, then we'll stop to propagate them.

test/libethashseal/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+cmake_policy(SET CMP0015 NEW)
```
chfast: You don't know what this is about, do you? Remove it.

test/libethashseal/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
...
+
+aux_source_directory(. SRCS)
```
chfast: `file(GLOB SRCS "*.cpp")`.

libethashseal/Ethash.cpp
```diff
@@ -158,9 +158,18 @@ void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase co
...
+	{
+		if (_bi.number() >= chainParams().u256Param("metropolisForkBlock"))
+		{
+			unsigned const nonceChainId(chainParams().u256Param("nonceChainID"));
+			_t.checkNonceChainId(nonceChainId);
+		}
+
+		// Unneeded as it's checked again in Executive. Keep it here since tests assume it's checked.
```
chfast: Can you add `TODO:` prefix for this comment.

libethcore/Transaction.cpp
```diff
@@ -31,6 +31,12 @@ using namespace std;
 using namespace dev;
 using namespace dev::eth;
 
+namespace
+{
+int const c_chainIdInNonceBits = 64;
+u256 const c_nonceLowMask = 0xffffffffffffffff;
```
chfast: I'm sure you can use primitive int type here, boost should have an overloading for it.

test/libethashseal/CMakeLists.txt
gumb0: No idea, just copied from another dir 😆

libethcore/Transaction.cpp
gumb0: Sure it has, but it converts it to multiprecision internally anyway, so will we save anything?

chfast: We will save at least dynamic initialization of this global.

gumb0: Buying it for additional initialization time for processing each transaction?
I'm not really convinced

chfast: So I checked :) https://gist.github.com/chfast/dc7cfb070947f703e57c25dc3cbb79f7.

- Both versions are very similar, they call `eval_bitwise_...` template function (big unrolled code in the middle).
- `u256 c_nonceLowMask` does not require dynamic initialization. See `_ZN12_GLOBAL__N_1L10lowMask256E`.
- The version with `uint64_t` is a bit better because does not require global constant loading.

That was not to prove you ware wrong, I was really curious. I would slightly prefer native type in the constant, but you can decide. Let me know.

gumb0: So it inlines the int value in code instead of using variable?

Ok, there's little difference, I'll change it, int looks slightly nicer in code for me, too

gumb0:
Rebased & review comments addressed

libethcore/Transaction.cpp
chfast: Global does not have to be created because it has local scope to linker `namespace {}`. Compiler usually produces code like `mov -1 to reg0`. The constant is included in the instruction itself.

chfast:
Is this ready to be merged?

pirapira:
This EIP faded away.

PR #3624, opened 2017-03-07
remove old state tests

winsvega:
move bStates coverage to time consuming tests


test/libethereum/Block.cpp
```diff
@@ -50,6 +50,8 @@ BOOST_AUTO_TEST_CASE(bStructures)
 
 BOOST_AUTO_TEST_CASE(bStates)
 {
+	if (!dev::test::Options::get().quadratic)
```
gumb0: I don't know much about how Boost.Test operate, but couldn't we do in a more dynamic way, like having external config files with a list of tests to run ?
So that we won't need to recompile tests every time we move tests between fast/slow suites ?

winsvega: cmake scans test files for boost test suites and then it all gets executed.
`./testeth -t <suitename>/<casename> `

some tests are disabled by default. so you could run it by
`./testeth --memory --quadratic --wallet` 

or just use all option
`./testeth --all`

chfast: I think "labels" is what we want in this case: http://www.boost.org/doc/libs/1_63_0/libs/test/doc/html/boost_test/tests_organization/tests_grouping.html

They are not dynamic (external config file), but at least we don't need to create our version of it.
But that should go into a separated PR.

PR #3630, opened 2017-03-10
Replace Crypto++'s SHA256 (partial)

chfast:


chriseth:
Can you run a quick performance check on the two?

chfast:
The scrypt version is at least 2x faster than Crypto++ version.

chriseth:
Nice! This also brings sha256 more in line with sha3 and ecrecover: https://gist.github.com/chriseth/4168b56bfe638cae8da1945dd988600b

chfast:
Now SHA256 is taken from secp256k1.

I did noob mistake before testing performance impact of this change. The true performance increase is ~39% for the set of 2 tests from #3632.

PR #3634, opened 2017-03-13
Replace Crypto++ ECIES

chfast:


axic:
Please add a test for PBKDF2.

libdevcrypto/Common.cpp
```diff
@@ -126,9 +117,40 @@ void dev::encryptECIES(Public const& _k, bytesConstRef _plain, bytes& o_cipher)
...
+	bytes mKey(32);
+	secp256k1_sha256_finalize(&ctx, mKey.data());
+
+	auto iv = h128::random();
+	bytes cipherText = encryptSymNoAuth(SecureFixedHash<16>(eKey), iv, _plain);
+	if (cipherText.empty())
+		return;
+
+	bytes msg(1 + Public::size + h128::size + cipherText.size() + 32);
+	msg[0] = 0x04;
```
axic: Probably I would create a helper which creates a SEC.1 key.

SEC.1 (http://www.secg.org/sec1-v2.pdf) section 2.3.3 describes the rules.

libdevcrypto/Common.cpp
```diff
@@ -138,10 +160,50 @@ bool dev::decryptECIES(Secret const& _k, bytesConstRef _cipher, bytes& o_plainte
 
 bool dev::decryptECIES(Secret const& _k, bytesConstRef _sharedMacData, bytesConstRef _cipher, bytes& o_plaintext)
 {
-	bytes io = _cipher.toBytes();
-	if (!Secp256k1PP::get()->decryptECIES(_k, _sharedMacData, io))
+	// interop w/go ecies implementation
+
+	// io_cipher[0] must be 2, 3, or 4, else invalidpublickey
+	if (_cipher.empty() || _cipher[0] < 2 || _cipher[0] > 4)
```
axic: Probably better to have a class for SEC.1 key (mentioned above) and just use validations on that class here.

libdevcrypto/Common.cpp
```diff
@@ -104,19 +102,12 @@ Address dev::toAddress(Address const& _from, u256 const& _nonce)
 
 void dev::encrypt(Public const& _k, bytesConstRef _plain, bytes& o_cipher)
 {
-	bytes io = _plain.toBytes();
-	Secp256k1PP::get()->encrypt(_k, io);
-	o_cipher = std::move(io);
+	encryptECIES(_k, _plain, o_cipher);
```
gumb0: `encrypt` & `decrypt` are now the same as `encryptECIES` & `decryptECIES`, do we still need both?

chfast: I will take a look. Probably not.

chfast:
@axic requested some wrappers for Pubkey, Seckey to be added to handle conversion to/from secp256k1 library format nicely. I think it is good idea and I will start working on it soon.

Issue #3618, opened 2017-03-03
EIP 140: REVERT instruction


pirapira:
Implementing REVERT instruction as part of the Metropolis changing.  It causes exceptional halting but returns data.

The newest document:
* https://github.com/ethereum/EIPs/pull/206

gumb0:
Implementation in EVMJIT is still needed

PR #3635, opened 2017-03-13
EIP140 - Implement REVERT opcode

gumb0:
https://github.com/ethereum/cpp-ethereum/issues/3618
https://github.com/ethereum/EIPs/pull/206


libevm/VM.cpp
```diff
@@ -233,6 +233,24 @@ void VM::interpretCases()
...
+				throwBadInstruction();
+
+			m_newMemSize = memNeed(*m_SP, *(m_SP - 1));
+			updateMem();
+			ON_OP();
+			updateIOGas();
+
+			uint64_t b = (uint64_t)*m_SP--;
+			uint64_t s = (uint64_t)*m_SP--;
+			owning_bytes_ref output{ move(m_mem), b, s };
```
chriseth: `return` above just assigns to `m_output` - why did you choose to put it into the exception?

gumb0: Because `m_output` is returned to Executive through return value of `VM::exec()`

And we need a way to both return output data and to tell Executive to revert the state.

We could either return pair/tuple from `VM::exec()` or throw exception with data.

gumb0: Exception is not ideal, since it's not really an exceptional situation, just another valid opcode.
But I think the code is cleaner this way (and after all REVERT is conceptually described as 'cheap throw')

libevmcore/Instruction.h
```diff
@@ -188,6 +188,7 @@ enum class Instruction: uint8_t
 	CALLCODE,           ///< message-call with another account's code only
 	RETURN,             ///< halt execution returning output data
 	DELEGATECALL,       ///< like CALLCODE but keeps caller's value and sender
+	REVERT = 0xfd,      ///< stop execution and revert state changes, without consuming all provided gas
```
gumb0: Forgot to fill `c_instructions`, `c_instructionInfo` in Instruction.cpp ...

libevm/VM.cpp
chfast: I think we should use `pair<bool, bytes>` (extended `std::optional`), as @chriseth did in precompiled already. We can name this time, and add convenient methods similar to `std::optional`. I think nothing from boost/std directly matches our case here.

If this is a lot refactoring, it can be done in a later PR. Ideally, implement the return type change first, then EIP140.

gcolvin: I've got a refactoring underway at https://github.com/ethereum/cpp-ethereum/tree/downward-dog, but it doesn't look like it will conflict with this one.

gcolvin: Talk of a bigger refactoring reminds me that I've never liked the interface to the VM - the way it's a function pretending to be an object, the way that function takes its ration of gas by reference...  I attempted to eliminate the _io_gas reference once and gave up, as it had insinuated itself too deep in the code.

chfast: Yep. I agree with that. The amount of gas left should be return as well. I'm trying to address that in https://github.com/ethereum/evmjit/blob/develop/include/evm.h#L70-L106, but EVM-C seems to be always postponed.

gumb0: Ok, I'll do it here. (I was not aware of those precompiled changes)

gumb0:
EVMJIT build is failing, forgot to check it locally...

chfast:
If you need help with it let me know.

test/tools/libtestutils/Common.cpp
```diff
@@ -44,7 +44,7 @@ std::string dev::test::getTestPath()
 	if (ptestPath == NULL)
 	{
 		ctest << " could not find environment variable ETHEREUM_TEST_PATH \n";
-		testPath = "../../../tests";
+		testPath = "../../test/jsontests";
```
gumb0: This path won't work on Windows 😞 (there's one subdir more for executables)

winsvega: you could use --testpath

gumb0: We need to change Appveyor's script then?

winsvega: yes

gumb0:
Possible checks/edge cases to pay attention when testing this:
- check that state changes are reverted
- check revert in contract creation code
- check that sender nonce is incremented even after revert
- check that sender nonce is incremented even after revert inside CREATE call
- check that no instructions after REVERT are executed
- check that output data is really returned

.gitmodules
```diff
@@ -5,3 +5,4 @@
 [submodule "test/tests"]
 	path = test/jsontests
 	url = https://github.com/ethereum/tests.git
+	branch = revertTests
```
winsvega: lets not merge before the test coverage

chfast: This is not right. "branch" is only convenient feature that allows you to do `git submodule update --remote`, nothing more. We definitely don't want this to be merged into our default branch.

libevm/VM.cpp
```diff
@@ -233,6 +233,25 @@ void VM::interpretCases()
...
+				throwBadInstruction();
+
+			m_newMemSize = memNeed(*m_SP, *(m_SP - 1));
+			updateMem();
+			ON_OP();
+			updateIOGas();
+
+			uint64_t b = (uint64_t)*m_SP--;
+			uint64_t s = (uint64_t)*m_SP--;
+			m_output = owning_bytes_ref{ std::move(m_mem), b, s };
```
chfast: Style: if you use `{}` to construct an object use it the same way as `()`.

libevm/VMFace.h
```diff
@@ -33,6 +33,13 @@ ETH_SIMPLE_EXCEPTION_VM(OutOfGas);
 ETH_SIMPLE_EXCEPTION_VM(OutOfStack);
 ETH_SIMPLE_EXCEPTION_VM(StackUnderflow);
 
+struct VmExecResult
+{
+	VmExecResult(owning_bytes_ref&& _output, bool _revertNeeded) : output(std::move(_output)), revertNeeded(_revertNeeded) {}
+
+	owning_bytes_ref output;
+	bool revertNeeded = false;
```
chfast: This just a comment, but the name for this flag is not right. A VM user does not care about reverts, but cares if the call succeeded or failed. `status` would be better name. Also compare with similar returned type in precompileds.

gumb0: Hm, but then we'll have two ways to signal `VM::exec()` failure - through exceptions and success return flag - which is not good at all.

Are we planning to get rid of throwing exceptions in favor of returned result then?

chfast: That would be nice I think. I'm sure @gcolvin agrees :).
This is classic case "should a parser throw exception if syntax is invalid"? IMHO, not, because it is nothing exceptional about detecting syntex errors in parser. It is its _normal_ job. But that's more philosophical...

Anyway. Throw an exception in this PR. Refactoring of this should be separated.

gumb0: That was the first version of this PR 😃 Ok, I'll try to revive those commits

gumb0:
I revived the version when REVERT results in the special exception (containing output data) being thrown from VM to Executive and made the CALL ended with REVERT be considered failed (pushing 0 to caller's stack).

libevm/VMFace.h
```diff
@@ -25,14 +25,30 @@ namespace dev
 namespace eth
 {
 
-struct VMException: virtual Exception {};
-#define ETH_SIMPLE_EXCEPTION_VM(X) struct X: virtual VMException { const char* what() const noexcept override { return #X; } }
+struct VMException: Exception {};
```
gumb0: Any reason why these were virtually inherited? It's not like they are going to be inherited several times here in libevm and `VMException` and children don't have data members anyway!

I'm changing it because gcc was issuing a warning for `RevertInstruction` move assignment, which is unsafe with virtual base (virtual base can be move-assigned several times if inherited several times)

gumb0:
Rebased on develop

libevm/VMFace.h
chfast: This is Boost suggestion: http://www.boost.org/doc/libs/1_63_0/libs/exception/doc/using_virtual_inheritance_in_exception_types.html

chfast:
Should I take it from here to implement that in EVMJIT?

PR #3637, opened 2017-03-14
more keystore tests

winsvega:
according to this page
https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition

### DIFF
--- a/test/libdevcrypto/SecretStore.cpp
+++ b/test/libdevcrypto/SecretStore.cpp
@@ -32,6 +32,7 @@ using namespace dev::test;
 
 namespace js = json_spirit;
 namespace fs = boost::filesystem;
+namespace utf = boost::unit_test;
 
 BOOST_AUTO_TEST_SUITE(Crypto)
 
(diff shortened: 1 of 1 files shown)


test/libdevcrypto/SecretStore.cpp
```diff
@@ -215,5 +216,122 @@ BOOST_AUTO_TEST_CASE(recode)
...
+		BOOST_CHECK_EQUAL(uuid, h128("3198bc9c66725ab3d9954942343ae5b6"));
+		BOOST_CHECK(contentsString(importFile) == keyData);
+		BOOST_CHECK_EQUAL(priv, toHex(store.secret(uuid, [&](){ return password; }).makeInsecure()));
+		BOOST_CHECK_EQUAL(store.keys().size(), 1);
+		BOOST_CHECK_EQUAL(store.address(uuid), Address("008aeeda4d805471df9b2a5b0f38a0c3bcba786b"));
+	}
+	fs::remove(importFile);
+}
+
+BOOST_AUTO_TEST_CASE(vector_Scrypt2, *utf::expected_failures(2))
```
chfast: Why 2 failures here?

test/libdevcrypto/SecretStore.cpp
```diff
@@ -215,5 +216,122 @@ BOOST_AUTO_TEST_CASE(recode)
 	}
 }
 
+BOOST_AUTO_TEST_CASE(vector_PBKDF2SHA256, *utf::expected_failures(1))
```
chfast: What this `vector_` prefix in the test name suppose to mean? Name it `KeyImport` or something like that.

test/libdevcrypto/SecretStore.cpp
winsvega: old test format

chfast: I don't understand. Does it mean it does not import it at all?

PR #3638, opened 2017-03-14
Experimental: AES replacement

chfast:


libdevcrypto/AES.cpp
```diff
@@ -28,9 +28,264 @@
 
 using namespace std;
 using namespace dev;
-using namespace dev::crypto;
 using namespace CryptoPP;
 
+namespace
+{
+
+// This code is modified version of https://github.com/kokke/tiny-AES128-C.
```
chriseth: Please mention that it was released under the "unlicense", roughly equivalent to a release to the public domain.

libdevcrypto/AES.cpp
```diff
@@ -28,9 +28,264 @@
...
+		roundKeys[(i * 4) + 2] = key[(i * 4) + 2];
+		roundKeys[(i * 4) + 3] = key[(i * 4) + 3];
+	}
+
+	// All other round keys are found from the previous round keys.
+	for(; (i < (Nb * (Nr + 1))); ++i)
+	{
+		for (int j = 0; j < 4; ++j)
+		{
+			tempa[j] = roundKeys[(i-1) * 4 + j];
```
chriseth: `i` is bounded by `Nb * (Nr + 1)` and `j` is bounded by `4` - is there any easier way to (perhaps in an automated way) check that this does not go out of bounds with respect to `roundKeys`?

chfast: I hoped I will not have to decrypt this indexing :D

chriseth: Sorry, but this just looks like dynamite to me. Buffers passed as plain pointers with sizes resulting from formulas including weird constants, it is just too easy to confuse two constants here, and we don't really know how much this library is used, tested, etc.

chfast:
I agree with @chriseth now. We should spend some time to find good AES implementation for both 128 and 256 (presale wallet) size.

pirapira:
@chfast what's the status here?

PR #3643, opened 2017-03-15
storageRangeAt return storage as object

yann300:


yann300:
already asked the go guy that's how they will do

yann300:
specification change:
`complete` flag is no longer used. 
Instead of that `nextKey` will contain the next key if it applies

libweb3jsonrpc/Debug.cpp
```diff
@@ -142,22 +142,21 @@ Json::Value Debug::debug_storageRangeAt(string const& _blockHashOrNumber, int _t
...
-			keyValue["hashedKey"] = toCompactHex(it->first, HexPrefix::Add, 1);
+			std::string hashedKey = toCompactHex(it->first, HexPrefix::Add, 1);
 			keyValue["key"] = toCompactHex(it->second.first, HexPrefix::Add, 1);
 			keyValue["value"] = toCompactHex(it->second.second, HexPrefix::Add, 1);
 
-			ret["storage"].append(keyValue);
+			ret["storage"][hashedKey] = keyValue;
 		}
+		if (it != storage.end())
+			ret["nextKey"] = toCompactHex((++it)->first, HexPrefix::Add, 1);
```
gumb0: We could move this into if body inside the loop (before `break`), then we won't need to define `it` outside of loop

PR #3644, opened 2017-03-15
split json and unit tests

winsvega:
testeth refactoring

new folder structure
`test/testtool/jsontests/`  - json test parsers logic
`test/testtool` - testeth source
`test/unittests` - coverage tests


chfast:
The description does not match the code. My view is like that (it is more how the code looks, than the description):

- test/
   - tools/ (C++ code supporting testing)
   - unittests/ (C++ Boost.Test unit tests)
   - jsontests/ (git submodule to tests repo)

In test/tools we can do additional grouping like libtestutils/ libjsontests/ testeth/, but I'm not sure this is necessary. My opinion is good file names are enough, less CMake config, etc. Even if we want some grouping, let's move what we have right now there to see how it looks.

Bad thing about this PR is you started moving files around without firstly submitting the idea (as a test description) to review.

Smaller issues I've found:

- [ ] A lost unit test: https://github.com/winsvega/cpp-ethereum/blob/develop/test/tools/external-dependencies/boost.cpp
- [ ] This files looks wrong: https://github.com/winsvega/cpp-ethereum/blob/develop/test/tools/libtesteth/CMakeLists.txt

winsvega:
Right. the idea of moving those files was discussed in the internal channel.

chfast:
I think a chat discussion is not enough. I would like to see for example a GH issue describing exactly what you are going to do.

test/CMakeLists.txt
```diff
@@ -1,34 +1,12 @@
...
-add_subdirectory(libdevcrypto)
-add_subdirectory(libp2p)
-add_subdirectory(libdevcore)
-add_subdirectory(libweb3jsonrpc)
-add_subdirectory(libwhisper)
-add_subdirectory(libevm)
-add_subdirectory(external-dependencies)
-
-set(SRC_LIST ${SRC_LIST} ${SRC})
+set(SRC_LIST ${SRC})
```
chfast: Use of one them names from the beginning.

test/tools/libtestutils/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # Don't re-use SRC_LIST, because it has a special meaning within the test folder,
 # with some dynamic aggregation of unit-testing going on.
 aux_source_directory(. LOCAL_SRC_LIST)
-include_directories(../..)
+include_directories(../../..)
```
chfast: Use target_include_directories().

test/CMakeLists.txt
```diff
@@ -1,34 +1,12 @@
 
-aux_source_directory(. SRC_LIST)
+file(GLOB_RECURSE SRC *.cpp *.sol)
```
chfast: Why `*.sol` files are also searched here?

winsvega: so to see it in the project

chfast:
The build fails on Linux and Windows: cannot find libtestutils.

winsvega:
do we use it somewhere I cant see. I think we could remove libtestutils

Issue #3197, opened 2016-08-26
Fedora - Add support in install_deps.sh


bobsummerwill:
_From @bobsummerwill on July 8, 2016 3:0_

Hey @3esmit and @tflux99 and @giact!

(moved from https://github.com/ethereum/webthree-umbrella/issues/371)

I have been working on an infrastructure-as-code alternative to these tedious per-distro sets of instructions, which will "go live" when we switch back to cpp-ethereum.

You can see my work-in-progress on that here:
- https://github.com/bobsummerwill/cpp-ethereum/blob/merge_repos/.travis.yml
- https://github.com/bobsummerwill/cpp-ethereum/blob/merge_repos/install_deps.sh

Now for the TravisCI bit, we're stuck with just Ubuntu Trusty automation for the time being, but we might well be able to expand that via Docker in the future.

Whatever the case, I would like to expand this install_deps.sh pattern pattern to cover more distros, so that our ethdocs.org instructions for the distros which we have bothered to do that can be simplified to, with the steps being executed directly and committed in with the code itself.
- git clone --recursive
- ./install_deps.sh
- cmake
- make

So maybe for Fedora we can take that approach instead, rather than bothering with written instructions?

Please could help me out and copy https://github.com/bobsummerwill/cpp-ethereum/blob/merge_repos/install_deps.sh locally and run it, does it go into the right conditional?

I suspect that it won't work, and that we will need to do something different for Fedora detection.  Thanks!

I suspect it won't, and that we will need to do something fancier, as per http://unix.stackexchange.com/questions/92199/how-can-i-reliably-get-the-operating-systems-name

Thanks!

_Copied from original issue: ethereum/webthree-umbrella#631_

bobsummerwill:
So I've had a go at writing install_deps.sh steps for Fedora.

See https://github.com/bobsummerwill/cpp-ethereum/blob/merge_repos/install_deps.sh#L269

Please could somebody be my guinea-pig?

Also, does `lsb_release` work for you?    If so, what does it show?   Thanks!

bobsummerwill:
BTW, @tflux99, in your Fedora 22 instructions you said GCC 5.3 didn't work, but I don't have any details on the failures you were hitting.    I think it should work, so I've gone without any override for that.    Let's see what that yields, and if there are build breaks then we can work through them.   Thanks!

bobsummerwill:
_From @giact on July 8, 2016 20:8_

I have been summoned! 😁

[...]

PR #3646, opened 2017-03-16
package cmake was missing for fedora

kmokeddem:
this is for ethereum/cpp-ethereum#3197


chfast:
This is on purpose. Users have to install git and cmake themselves.

chfast:
Can you review this change in the README file and instructions how to build the project included there?

kmokeddem:
The README is fine regarding fedora. 

It might be useful to make a mention of needing the cmake3 package and cmake3 for the correct version of cmake on redhat, but likely anyone who's building from source can figure that out.

PR #3649, opened 2017-03-17
Eip86transaction tests

winsvega:


libethashseal/Ethash.cpp
```diff
@@ -152,8 +152,13 @@ void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase co
 	{
 		if (_bi.number() >= chainParams().u256Param("EIP158ForkBlock"))
 		{
-			int chainID(chainParams().u256Param("chainID"));
-			_t.checkChainId(chainID);
+			if (_bi.number() >= chainParams().u256Param("metropolisForkBlock") && _t.hasZeroSignature())
+				_t.checkChainId(0);
```
yann300: well i am not that's right. I don't think you have to modify that.

winsvega: otherwise it will throw on line 
`_t.checkChainId(chainID);`

yann300: if you have zero signature, you can extract the chain Id
`chain_id = signature.v`
like `_t.checkChainId(_t.signature.v);`

gumb0: @winsvega Why do you need this change?
Zero signature still has chainId in `v` and should be checked

yann300: well no not let me check something

gumb0: @winsvega if it throws, you have incorrect chainId in your `v` in zero signature

yann300: I think that's fine to keep the previous code
`
int chainID(chainParams().u256Param("chainID"));
_t.checkChainId(chainID);
`

in the constructor of `TransactionBase`, m_chain is set to sign.v in case of zero signature

winsvega: the issue is that chain params has chainid = 1  
and zero sig transaction has chainid = 0 
meaning that we wont accept zerosig transaction on any chain if it's id != 0

winsvega: we could however change 
`void TransactionBase::checkChainId(int chainId) const`
to return true if transaction has zero signature and chain id of 0 
but i think it is better to check it again metropolis fork block

yann300: zero sig transaction should have `v == 1` also

libethashseal/genesis/mainNetworkTest.cpp
```diff
@@ -27,7 +27,7 @@ R"E(
 		"daoHardforkBlock": "0x1d4c00",
 		"EIP150ForkBlock": "0x259518",
 		"EIP158ForkBlock": "0x28d138",
-		"metropolisForkBlock": "0xfffffffffffffff",
+		"metropolisForkBlock": "0x2dc6c0",
```
yann300: what is 0x2dc6c0 ?

winsvega: 3 000 000  metropolis transition block. just for the time being (in test network)

Issue #3332, opened 2016-10-05
Use-after-free when closing eth


chfast:
```
=================================================================
==6164==ERROR: AddressSanitizer: heap-use-after-free on address 0x625000059bc8 at pc 0x7fd141367c15 bp 0x7fd12a7219a0 sp 0x7fd12a721990
READ of size 1 at 0x625000059bc8 thread T14 (p2p)
    #0 0x7fd141367c14 in std::_Function_handler<void (boost::system::error_code const&), dev::p2p::DeadlineOps::reap()::{lambda(boost::system::error_code const&)#1}>::_M_invoke(std::_Any_data const&, boost::system::error_code const&) (/home/chfast/Projects/ethereum/build/cpp-ethereum/asan/libp2p/libp2p.so+0x199c14)
    #1 0x7fd1412f8266 in boost::asio::detail::wait_handler<std::function<void (boost::system::error_code const&)> >::do_complete(boost::asio::detail::task_io_service*, boost::asio::detail::task_io_service_operation*, boost::system::error_code const&, unsigned long) (/home/chfast/Projects/ethereum/build/cpp-ethereum/asan/libp2p/libp2p.so+0x12a266)
    #2 0x7fd14132200f in dev::p2p::Host::doneWorking() (/home/chfast/Projects/ethereum/build/cpp-ethereum/asan/libp2p/libp2p.so+0x15400f)
    #3 0x7fd140f41c64 in dev::Worker::startWorking()::{lambda()#1}::operator()() const [clone .constprop.236] (/home/chfast/Projects/ethereum/build/cpp-ethereum/asan/libdevcore/libdevcore.so+0x8cc64)
    #4 0x7fd1403a2c7f  (/usr/lib/x86_64-linux-gnu/libstdc++.so.6+0xb8c7f)
    #5 0x7fd141e7d6f9 in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x76f9)
    #6 0x7fd13fb08b5c in clone (/lib/x86_64-linux-gnu/libc.so.6+0x106b5c)

0x625000059bc8 is located 8904 bytes inside of 8912-byte region [0x625000057900,0x625000059bd0)
freed by thread T14 (p2p) here:
    #0 0x7fd14280fb2a in operator delete(void*) (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x99b2a)
[...]

PR #3651, opened 2017-03-17
issues/3332 use-after-free

kmokeddem:
calling m_timer.reset() detroys m_ioService which is later used in doneWorking(). 

https://github.com/ethereum/cpp-ethereum/issues/3332


chfast:
Hey @kmokeddem, 
thanks for checking this one.

You mentions a test for this. Can it be included?

chfast:
It does not fix the issue I think. Also some tests don't want to finish after the change.

I think the issue is more about `m_nodeTable.reset()` line a bit above your change.

Do you know how to build with ASan? Add option `-DSANITIZE=address` to cmake config.

PR #3653, opened 2017-03-20
Integrate hunter and use it for Boost

chfast:


chfast:
@gcolvin @gumb0 @pirapira Can you please test this PR for me? I need feedback about user experience.

pirapira:
I guess I'm added for Mac.

pirapira:
Seems to be working for me on a Mac (just upgraded to Sierra, 10.12.4).

chfast:
@gumb0 Windows/Debug fixed.

gumb0:
Debug now builds fine. I tried to build RelWithDebInfo and got errors about failing to find jsonrpccpp headers
```
C:\dev\cpp-ethereum\libweb3jsonrpc\AdminEth.cpp(1): fatal error C1083: Cannot open include file: 'jsonrpccpp/common/exception.h': No such file or directory

```

chfast:
Yeah... partly fixed. But will also require EVMJIT update. Tomorrow :)

chfast:
It looks like all fixed now :)

gumb0:
Looks like working now

chfast:
Security concerns have been raised during the meeting in Berlin about downloading sources and binaries from unknown sources. I've researched the issue and will publish the extended description in a separate document. But the short answer: we can build from sources as they are "protected" by hashes known up-front and cache the outputs "locally". Downloading binaries from not our cache server is more risky and we should disable this option.

chfast:
I'd like to merge this and give it a try. In case of troubles I will revert the change.

PR #3655, opened 2017-03-21
Transaction tests

winsvega:
see the commit titles

### DIFF
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -180,6 +180,11 @@ void Executive::initialize(Transaction const& _transaction)
 {
 	m_t = _transaction;
 
+	BlockHeader bh;
+	bh.setNumber(m_envInfo.number());
+	bh.setGasLimit(m_envInfo.gasLimit());
+	m_sealEngine.verifyTransaction(ImportRequirements::Everything, m_t, bh);
+
 	// Avoid transactions that would take us beyond the block gas limit.
 	u256 startGasUsed = m_envInfo.gasUsed();
 	if (startGasUsed + (bigint)m_t.gas() > m_envInfo.gasLimit())
--- a/libethereum/State.cpp
+++ b/libethereum/State.cpp
@@ -518,6 +518,10 @@ std::pair<ExecutionResult, TransactionReceipt> State::execute(EnvInfo const& _en
 		onOp = Executive::simpleTrace(); // override tracer
 #endif
 
+	BlockHeader bh;
+	bh.setNumber(_envInfo.number());
+	_sealEngine.verifyTransaction(ImportRequirements::Everything, _t, bh);
+
 	// Create and initialize the executive. This will throw fairly cheaply and quickly if the
 	// transaction is bad in any way.
 	Executive e(*this, _envInfo, _sealEngine);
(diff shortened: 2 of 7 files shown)


libethereum/State.cpp
```diff
@@ -518,6 +518,11 @@ std::pair<ExecutionResult, TransactionReceipt> State::execute(EnvInfo const& _en
 		onOp = Executive::simpleTrace(); // override tracer
 #endif
 
+	BlockHeader bh;
+	bh.setNumber(_envInfo.number());
+	bh.setGasLimit(_envInfo.gasLimit());
+	_sealEngine.verifyTransaction(ImportRequirements::Everything, _t, bh);
```
gumb0: Most of the checks inside `verifyTransaction` are duplicated in `Executive::initialize` I think

So I think we should better move this `verifyTransaction` call into `Executive::initialize`, removing the checks there.
I'm not sure if it's easy, when I tried to do this refactoring, I think saving the exception to `m_excepted` was something hard to deal with, then I gave up.

But just adding the call to `verifyTransaction` here doesn't seem right

libethereum/Executive.cpp
```diff
@@ -198,26 +198,23 @@ void Executive::initialize(Transaction const& _transaction)
 		BOOST_THROW_EXCEPTION(OutOfGasBase() << RequirementError((bigint)m_baseGasRequired, (bigint)m_t.gas()));
 	}
 
-	if (!m_t.hasZeroSignature())
+	// Avoid invalid transactions.
```
gumb0: Nonce shouldn't be checked for zero signature transaction, I think

libethereum/Executive.cpp
```diff
@@ -180,6 +180,11 @@ void Executive::initialize(Transaction const& _transaction)
 {
 	m_t = _transaction;
 
+	BlockHeader bh;
+	bh.setNumber(m_envInfo.number());
+	bh.setGasLimit(m_envInfo.gasLimit());
+	m_sealEngine.verifyTransaction(ImportRequirements::Everything, m_t, bh);
```
gumb0: `Executive::initialize` saves the exceptions to `m_excepted` member for some reason, so I guess we need to do the same for exceptions thrown here from `verifyTransactions`

So we need here
1. Wrap this in try/catch, catching `Exception`, filling `m_excepted` and rethrowing
2. If you add catch clause for `OutOfGasIntrinsic`, you could remove the check in lines 197-204, because it's already checked in `verifyTransaction`. The catch clause should do the same as current check inside `if (m_baseGasRequired > m_t.gas())`

libethereum/Executive.cpp
```diff
@@ -180,6 +180,11 @@ void Executive::initialize(Transaction const& _transaction)
 {
 	m_t = _transaction;
 
+	BlockHeader bh;
```
gumb0: This constructing of fake header is not so good, you assume that number and gasLimit are the only things that `verifyTransaction` is going to need, i.e. relying on internal implementation of the called function, which can change in the future and break this code...

Ideally we'd like to use the real block header here
@chfast do you have any advice for this? Maybe we could save the block header as `Executive` member?

winsvega: that was the issue I was talking about in gitter channel

yann300: sorry that was maybe already answered, but why do we need to call `verifyTransaction` at this point?

gumb0: @yann300 It makes sense for example for unconfirmed transactions coming over network to validate them before executing

PR #3656, opened 2017-03-21
Tests repo as a submodule

winsvega:
this would allow us to change travis script to fetch to a certain commit from the test repo 
and run specific tests on a PR
testeth could take the testpath with --testpath option

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,6 @@
 	path = evmjit
 	url = https://github.com/ethereum/evmjit
 	branch = develop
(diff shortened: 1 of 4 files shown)


chfast:
I realized the travis config is missing here. We don't have to clone tests repo any more (as the checkout of cpp-ethereum will do this for us).

winsvega:
lets 
do it with Andrei's PR 
https://github.com/ethereum/cpp-ethereum/pull/3635

PR #3658, opened 2017-03-23
Re-enable broken unit tests

chfast:
The "rescue" test is still marked with 1 expected failure.


chfast:
@winsvega what do you think?

test/unittests/libethereum/BlockChain.cpp
```diff
@@ -476,7 +442,7 @@ BOOST_AUTO_TEST_CASE(insertException)
 	}
 }
 
-BOOST_AUTO_TEST_CASE(rescue)
+BOOST_AUTO_TEST_CASE(rescue, *utf::expected_failures(1))
```
winsvega: is this test constantly failing?

test/unittests/libethereum/BlockChain.cpp
```diff
@@ -80,39 +79,6 @@ BOOST_AUTO_TEST_CASE(opendb)
 	BOOST_CHECK_EXCEPTION(BlockChain bc2(p, tempDirBlockchain.path(), WithExisting::Verify), DatabaseAlreadyOpen, is_critical);
 }
 
-//(rebuild)
```
chfast: I don't know. It was dead for so long so I deleted it.

test/unittests/libethereum/BlockChain.cpp
chfast: No, but fails with some good probability. And it is not about synchronization (see removed sleep), I waited many minutes and it still can fail. Boost is fine with tests expected to fail actually not to fail.

PR #3659, opened 2017-03-24
EIP86

yann300:
- Stop execution (if nonce || value || gasPrice != 0 in case of ZeroSignTrasaction)
 - Don't incr nonce if Max address

### DIFF
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -348,7 +348,7 @@ Secret Nonce::next()
 	return sha3(~m_value);
 }
 
-void ecdh::agree(Secret const& _s, Public const& _r, Secret& o_s)
+bool ecdh::agree(Secret const& _s, Public const& _r, Secret& o_s)
 {
 	auto* ctx = getCtx();
 	static_assert(sizeof(Secret) == 32, "Invalid Secret type size");
@@ -356,12 +356,16 @@ void ecdh::agree(Secret const& _s, Public const& _r, Secret& o_s)
 	std::array<byte, 65> serializedPubKey{{0x04}};
 	std::copy(_r.asArray().begin(), _r.asArray().end(), serializedPubKey.begin() + 1);
 	auto r = secp256k1_ec_pubkey_parse(ctx, &rawPubkey, serializedPubKey.data(), serializedPubKey.size());
-	(void)r;
+	if (r != 1)
+	{
+		return false;
+	}
 	assert(r == 1);
 	std::array<byte, 33> compressedPoint;
 	r = secp256k1_ecdh_raw(ctx, compressedPoint.data(), &rawPubkey, _s.data());
 	assert(r == 1);  // TODO: This should be "invalid secret key" exception.
 	std::copy(compressedPoint.begin() + 1, compressedPoint.end(), o_s.writable().data());
+	return true;
 }
 
 bytes ecies::kdf(Secret const& _z, bytes const& _s1, unsigned kdByteLen)
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -356,6 +356,7 @@ void ecdh::agree(Secret const& _s, Public const& _r, Secret& o_s)
 	std::array<byte, 65> serializedPubKey{{0x04}};
 	std::copy(_r.asArray().begin(), _r.asArray().end(), serializedPubKey.begin() + 1);
 	auto r = secp256k1_ec_pubkey_parse(ctx, &rawPubkey, serializedPubKey.data(), serializedPubKey.size());
+	(void)r;
 	assert(r == 1);
 	std::array<byte, 33> compressedPoint;
 	r = secp256k1_ecdh_raw(ctx, compressedPoint.data(), &rawPubkey, _s.data());
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -348,7 +348,7 @@ Secret Nonce::next()
 	return sha3(~m_value);
 }
 
-bool ecdh::agree(Secret const& _s, Public const& _r, Secret& o_s)
+void ecdh::agree(Secret const& _s, Public const& _r, Secret& o_s)
 {
 	auto* ctx = getCtx();
 	static_assert(sizeof(Secret) == 32, "Invalid Secret type size");
@@ -356,16 +356,11 @@ bool ecdh::agree(Secret const& _s, Public const& _r, Secret& o_s)
 	std::array<byte, 65> serializedPubKey{{0x04}};
 	std::copy(_r.asArray().begin(), _r.asArray().end(), serializedPubKey.begin() + 1);
 	auto r = secp256k1_ec_pubkey_parse(ctx, &rawPubkey, serializedPubKey.data(), serializedPubKey.size());
-	if (r != 1)
-	{
-		return false;
-	}
 	assert(r == 1);
 	std::array<byte, 33> compressedPoint;
 	r = secp256k1_ecdh_raw(ctx, compressedPoint.data(), &rawPubkey, _s.data());
 	assert(r == 1);  // TODO: This should be "invalid secret key" exception.
 	std::copy(compressedPoint.begin() + 1, compressedPoint.end(), o_s.writable().data());
-	return true;
 }
 
 bytes ecies::kdf(Secret const& _z, bytes const& _s1, unsigned kdByteLen)
--- a/libethcore/SealEngine.cpp
+++ b/libethcore/SealEngine.cpp
@@ -47,6 +47,12 @@ void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, Transactio
 {
 	if ((_ir & ImportRequirements::TransactionSignatures) && _bi.number() < chainParams().u256Param("metropolisForkBlock") && _t.hasZeroSignature())
 		BOOST_THROW_EXCEPTION(InvalidSignature());
+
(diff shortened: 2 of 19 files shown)


winsvega:
could we merge at least first 3 commits from this 
https://github.com/ethereum/cpp-ethereum/pull/3655  
?

libethereum/Executive.cpp
```diff
@@ -218,18 +218,23 @@ void Executive::initialize(Transaction const& _transaction)
...
+		bigint totalCost = m_t.value() + gasCost;
+		if (m_s.balance(m_t.sender()) < totalCost)
+		{
+			clog(ExecutiveWarnChannel) << "Not enough cash: Require >" << totalCost << "=" << m_t.gas() << "*" << m_t.gasPrice() << "+" << m_t.value() << " Got" << m_s.balance(m_t.sender()) << "for sender: " << m_t.sender();
+			m_excepted = TransactionException::NotEnoughCash;
+			BOOST_THROW_EXCEPTION(NotEnoughCash() << RequirementError(totalCost, (bigint)m_s.balance(m_t.sender())) << errinfo_comment(m_t.sender().abridged()));
+		}
+		m_gasCost = (u256)gasCost;  // Convert back to 256-bit, safe now.
+	}
+	else if (m_t.gasPrice() != 0 || m_t.value() != 0 || m_t.nonce() != 0)
```
gumb0: I hope that this check won't be needed when we merge the changes in `Executive` from https://github.com/ethereum/cpp-ethereum/pull/3655
(then `verifyTransaction` will be called here, which has this check)

yann300: right this check will be done by `verifyTransaction`

libethereum/State.cpp
```diff
@@ -271,7 +271,9 @@ u256 State::balance(Address const& _id) const
 
 void State::incNonce(Address const& _addr)
 {
-	if (Account* a = account(_addr))
+	if (_addr == MaxAddress)
```
gumb0: I would better put this special logic into `Executive`, because it's more "the logic of executing special kind of transactions" than "the logic of incrementing the nonce in the state", and it would be closer to other special treatment of zero sig transactions.

If you want to avoid duplication in several places where `incNonce` is called, make one wrapper method with this check

winsvega:
please rebase your PR
and point the submodule jsontests to
https://github.com/ethereum/tests/tree/zeroTransaction

yann300:
I really want this PR to include only:
- Stop execution (if nonce || value || gasPrice != 0 in case of ZeroSignTrasaction)
- Don't incr nonce if Max address

Tests you want to add fit in this scope?
(note that I did not add any test yet - need to do that)

libethereum/Executive.cpp
gumb0: we can remove this now?

libethereum/Executive.cpp
```diff
@@ -180,6 +180,8 @@ void Executive::initialize(Transaction const& _transaction)
 {
 	m_t = _transaction;
 
+	m_sealEngine.verifyTransaction(ImportRequirements::Everything, _transaction, m_envInfo);
```
gumb0: @winsvega We still need to wrap this in try/catch, see my comment in https://github.com/ethereum/cpp-ethereum/pull/3655

test/tools/libtesteth/BlockChainHelper.cpp
```diff
@@ -545,6 +545,22 @@ TestTransaction TestTransaction::defaultTransaction(u256 const& _nonce, u256 con
...
 }
 
+TestTransaction TestTransaction::defaultZeroTransaction(u256 const& _gasLimit, bytes const& _data)
+{
+	json_spirit::mObject txObj;
+	txObj["data"] = toHex(_data);
+	txObj["gasLimit"] = toString(_gasLimit);
+	txObj["gasPrice"] = toString(0);
+	txObj["nonce"] = toString(0);
+	txObj["v"] = toString(1);
```
yann300: `chainId` is always 1 ?

winsvega: this is just the test version. yes we could define an argument in this function for chainID
by default chainID must be = 1 in network test.

libdevcrypto/Common.cpp
```diff
@@ -356,6 +356,7 @@ void ecdh::agree(Secret const& _s, Public const& _r, Secret& o_s)
 	std::array<byte, 65> serializedPubKey{{0x04}};
 	std::copy(_r.asArray().begin(), _r.asArray().end(), serializedPubKey.begin() + 1);
 	auto r = secp256k1_ec_pubkey_parse(ctx, &rawPubkey, serializedPubKey.data(), serializedPubKey.size());
+	(void)r;
```
chriseth: Can we instead perform assertion checking also in release mode?

libethcore/SealEngine.cpp
```diff
@@ -43,11 +43,17 @@ void SealEngineFace::populateFromParent(BlockHeader& _bi, BlockHeader const& _pa
...
 
-void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, TransactionBase const& _t, BlockHeader const& _bi) const
+void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, TransactionBase const& _t, EnvInfo const& _env) const
 {
-	if ((_ir & ImportRequirements::TransactionSignatures) && _bi.number() < chainParams().u256Param("metropolisForkBlock") && _t.hasZeroSignature())
+	if ((_ir & ImportRequirements::TransactionSignatures) && _env.number() < chainParams().u256Param("metropolisForkBlock") && _t.hasZeroSignature())
 		BOOST_THROW_EXCEPTION(InvalidSignature());
-	if (_bi.number() >= chainParams().u256Param("homsteadForkBlock") && (_ir & ImportRequirements::TransactionSignatures))
+
+	if ((_ir & ImportRequirements::TransactionBasic) && _env.number() >= chainParams().u256Param("metropolisForkBlock") &&
```
chriseth: Can you format this according to the coding style, please?

libethereum/BlockChain.cpp
```diff
@@ -1535,7 +1535,7 @@ VerifiedBlockRef BlockChain::verifyBlock(bytesConstRef _block, std::function<voi
 			try
 			{
 				Transaction t(d, (_ir & ImportRequirements::TransactionSignatures) ? CheckTransaction::Everything : CheckTransaction::None);
-				m_sealEngine->verifyTransaction(_ir, t, h);
+				m_sealEngine->verifyTransaction(_ir, t, EnvInfo(h));
```
chriseth: What is the difference between EnvInfo and only the block header? Why do we need env info?

winsvega: we use m_sealEngine->verifyTransaction in State where there are no blockheader. so not to construct fake blockheader at there Yann replaced the function to accept EnvInfo (which is used at State)

libethereum/Executive.cpp
```diff
@@ -180,6 +180,26 @@ void Executive::initialize(Transaction const& _transaction)
 {
 	m_t = _transaction;
 
+	try
+	{
+		m_sealEngine.verifyTransaction(ImportRequirements::Everything, _transaction, m_envInfo);
+	}
+	catch (InvalidSignature ex)
```
gumb0: Better catch the exception by const reference, then it's not copied
`catch (InvalidSignature const&)`

libethereum/Executive.cpp
```diff
@@ -180,6 +180,26 @@ void Executive::initialize(Transaction const& _transaction)
...
+	{
+		m_excepted = TransactionException::InvalidSignature;
+		BOOST_THROW_EXCEPTION(ex);
```
gumb0: and rethrow just by `throw;`

libethereum/Executive.cpp
```diff
@@ -260,7 +280,8 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
 		// FIXME: changelog contains unrevertable balance change that paid
 		//        for the transaction.
 		// Increment associated nonce for sender.
-		m_s.incNonce(_p.senderAddress);
+		if (_p.senderAddress != MaxAddress) // IEP86
```
gumb0: typo: EIP

libethereum/Executive.cpp
```diff
@@ -316,7 +337,8 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
 bool Executive::create(Address _sender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin)
 {
 	u256 nonce = m_s.getNonce(_sender);
-	m_s.incNonce(_sender);
+	if (_sender != MaxAddress) // IEP86
```
gumb0: typo: EIP

libethereum/State.cpp
```diff
@@ -297,7 +297,8 @@ void State::addBalance(Address const& _id, u256 const& _amount)
 		// Increase the account balance. This also is done for value 0 to mark
 		// the account as dirty. Dirty account are not removed from the cache
 		// and are cleared if empty at the end of the transaction.
-		a->addBalance(_amount);
+		if (_id != MaxAddress) // IEP86
```
gumb0: typo: EIP

gumb0:
I'm still not sure whether we should have better save full `BlockHeader` in `EnvInfo` instead of having `EnvInfo` parameter in `verifyTransaction`, but I'm ok with merging it now like this if tests pass.

winsvega:
smth is wrong with the test run. check that submodule is set.

PR #4022, opened 2017-03-29
reverse stack direction, simplify stack logic

gcolvin:
Many little changes here, mostly in service of having the stack grow down rather than up and moving the adjustment and checking of the stack pointer into the CASE macro.  So instead of this for ADD
```
*(m_SP - 1) += *m_SP;
--m_SP;
```
we get this
```
m_SPP[0] = m_SP[0] + m_SP[1];
```
which is easier to read and corresponds to the yellow paper notation, where SP is the current stack pointer, and SP' is the next one.  It gets much easier to read in some of the more complex operations.  In the old days the array notation might have cost a little performance, but compilers are too smart for that now, especially when the indexes are all small constants.

### DIFF
--- a/libevm/VM.cpp
+++ b/libevm/VM.cpp
@@ -57,22 +57,21 @@ uint64_t VM::decodeJumpDest(const byte* const _code, uint64_t& _pc)
 	return dest;
 }
 
-uint64_t VM::decodeJumpvDest(const byte* const _code, uint64_t& _pc, u256*& _sp)
+uint64_t VM::decodeJumpvDest(const byte* const _code, uint64_t& _pc, byte _voff)
 {
 	// Layout of jump table in bytecode...
 	//     byte opcode
 	//     byte n_jumps
 	//     byte table[n_jumps][4]
 	//	
-	uint64_t i = uint64_t(*_sp--);  // byte on stack indexes into jump table
 	uint64_t pc = _pc;
 	byte n = _code[++pc];           // byte after opcode is number of jumps
-	if (i >= n) i = n - 1;          // if index overflow use default jump
-	pc += i * 4;                    // adjust pc to index destination in table
+	if (_voff >= n) _voff = n - 1;  // if offset overflows use default jump
+	pc += _voff * 4;                // adjust inout pc before index destination in table
 	
 	uint64_t dest = decodeJumpDest(_code, pc);
 	
-	_pc += 1 + n * 4;               // adust input _pc to opcode after table 
+	_pc += 1 + n * 4;               // adust inout _pc to opcode after table 
 	return dest;
 }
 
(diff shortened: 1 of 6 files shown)


gcolvin:
This line compiles and tests fine on Linux and Mac.  It very mysteriously will not compile on Windows. 
```
void copyDataToMemory(bytesConstRef _data, u256* _SP);
```
Stripping out the cruft, the errors are:
```
VM.h(149): error C2143: syntax error: missing ')' before 'constant'
VM.h(149): error C2143: syntax error: missing ';' before 'constant'
VM.h(149): error C2059: syntax error: 'constant'
VM.h(149): error C2059: syntax error: ')'
VM.h(149): error C2238: unexpected token(s) preceding ';'
```
Maybe it will be obvious to someone, but I need to sleep for now.

libevm/VM.cpp
```diff
@@ -87,15 +86,31 @@ void VM::onOperation()
...
 }
+#if EVM_HACK_ON_OPERATION
+	#define onOperation()
+#endif
 
-void VM::checkStack(unsigned _removed, unsigned _added)
+//
+// set current SP to SP', adjust SP' per _removed and _added items
+//
+void VM::adjustStack(unsigned _removed, unsigned _added)
```
chfast: I would suggest using `int` type here. 2 reasons: we need to do arithmetic in the function, so unsigned type is not good for this. Moreover, unsigned is sometimes pessimization because of _defined_ overflow.

libevm/VM.h
```diff
@@ -97,19 +101,17 @@ class VM: public VMFace
...
-	byte* m_code = nullptr;
+	// space for code
+	bytes m_code;
 
-	// space for stack and pointer to data
-	u256 m_stackSpace[1025];
-	u256* m_stack = m_stackSpace + 1;
-	ptrdiff_t stackSize() { return m_SP - m_stack; }
+	// space for data stack, grows towards smaller addresses from the end
+	u256 m_stack[1024];
```
chfast: I'm not sure it will be very useful here, but you can use `std::array<u256, 1024>` instead of `u256[1024]`. It is a wrapper for a classic array, but with interface similar to other containers. So you will get a methods like `.begin()`. On the other hand, `std::begin()` function works with `u256[1024]` type.

gcolvin: Since I'm not using the stack with any STL algorithms I agree that std::array is not very useful.  (It's funny, because I was a big advocate for these safe wrapper classes, and authored some of them, but I don't use them much.)

libevm/VM.cpp
gcolvin: I'll look at it.  I think the `unsigned` types go back to the older code this came from, and they get passed to throwBadStack() function which expects unsigned.  Overall I don't think the code is always using the optimal data types.  (Not sure I understand the pessimization problem - _removed and _added are not going to overflow/)

libevm/VM.h
chfast: `unsigned` type has to support value wrap (`int` not, because in signed types overflow is undefined behavior). So if compiler cannot prove that overflow with `unsigned` cannot happen it has to generate additional code to support the wrap, because `unsigned` is 32-bit and we are on 64-bit architecture :)

Issue #3647, opened 2017-03-16
CodeCov: disable fixing data


chfast:
Add option `-X fix` to codecov script config in .travis.yml.

narendravardi:
Hi Paweł,

I am a newbie and want to work on this issue. Can I work on this?

chfast:
Hi Narendra,
yes you can take it.

narendravardi:
@chfast,

I have raised a pull request for this issue. Please review and let me know if any changes required.

chfast:
Fixed.

PR #4024, opened 2017-03-31
Adding option -X fix to codecov script config in .travis.yml #3647

narendravardi:
Adding option -X fix to codecov script config in .travis.yml #3647


narendravardi:
@chfast thank you very much for accepting! :)

Issue #3615, opened 2017-03-03
EIP101/198: bigint arithmetics precompiled contracts


pirapira:
Adding precompiled contracts that do unsigned big-integer arithmetics, following EIP 101.

* EIP https://github.com/ethereum/EIPs/issues/101
* EIP (PR) https://github.com/ethereum/EIPs/pull/198

PR #4028, opened 2017-04-03
EIP198 - Bigint modexp precompiled contract

gumb0:
https://github.com/ethereum/cpp-ethereum/issues/3615
https://github.com/ethereum/EIPs/pull/198

- [x] Contract code implementation 
- [x] Calculating the cost
- [x] Optimize code implementation if `boost::multiprecision::powm` is too slow


gumb0:
Looks like the `boost::multiprecision::powm` is not slower that the manual implemetation, since we don't know the modulo length and can't get rid of intermediate modulo operations (like we do for the `EXP` opcode implementation)

gcolvin:
I don't recall the history here, but why didn't we just add a MODEXP opcode?

chfast:
This is actually a good question, @gcolvin. Should we go to EIPs with it to ask for clarification when a features go as opcodes and when as precompiled contracts...

chriseth:
@gcolvin for this particular function, the reason to choose a precompiled contract over an opcode was because the arguments can be larger than 32 bits and it would be weird to have such an opcode fetch from memory. Furthermore, I think the idea is that as much as possible should go to precompiled contracts instead of opcodes, and precompiled contracts have to be pure functions.

PR #4033, opened 2017-04-04
transaction tests

winsvega:
do not include sender to the final test if transaction is invalid

### DIFF
--- a/test/tools/jsontests/TransactionTests.cpp
+++ b/test/tools/jsontests/TransactionTests.cpp
@@ -85,6 +85,8 @@ void doTransactionTests(json_spirit::mValue& _v, bool _fillin)
 				//Transaction is InValid
 				cnote << "Transaction Exception: " << diagnostic_information(_e);
 				o.erase(o.find("transaction"));
+				if (o.count("sender") > 0)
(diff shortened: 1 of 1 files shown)


test/tools/jsontests/TransactionTests.cpp
```diff
@@ -85,6 +85,8 @@ void doTransactionTests(json_spirit::mValue& _v, bool _fillin)
 				//Transaction is InValid
 				cnote << "Transaction Exception: " << diagnostic_information(_e);
 				o.erase(o.find("transaction"));
+				if (o.count("sender") > 0)
```
gumb0: Not sure I understand what do these changes do. It influences the generated json files?

chfast: It influences the way json tests are generated (we can expect update in the jsontests submodule after it).

winsvega:
So I got a report from Arakdy. He said that presents of 'sender' field in transaction tests that are expected to be invalid kind of misleading so I already fixed the tests and this is the fix for transaction tests filler.

chfast:
Ok, thanks.

Issue #4031, opened 2017-04-03
Buffer overflow in testeth --filltests


chfast:
```
=================================================================
==13617==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6020000daf58 at pc 0x55917e2b9aee bp 0x7f27890c9930 sp 0x7f27890c9920
READ of size 8 at 0x6020000daf58 thread T1
    #0 0x55917e2b9aed in dev::test::ImportTest::exportTest(std::vector<unsigned char, std::allocator<unsigned char> > const&) /home/chfast/Projects/ethereum/cpp-ethereum/test/tools/libtesteth/ImportTest.cpp:579
    #1 0x55917e21b3da in dev::test::doStateTests(json_spirit::Value_impl<json_spirit::Config_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&, bool) /home/chfast/Projects/ethereum/cpp-ethereum/test/tools/jsontests/StateTests.cpp:71
    #2 0x55917e161386 in std::_Function_handler<void (json_spirit::Value_impl<json_spirit::Config_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&, bool), void (*)(json_spirit::Value_impl<json_spirit::Config_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&, bool)>::_M_invoke(std::_Any_data const&, json_spirit::Value_impl<json_spirit::Config_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&, bool&&) /usr/include/c++/6/functional:1740
    #3 0x55917e16aff2 in std::function<void (json_spirit::Value_impl<json_spirit::Config_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&, bool)>::operator()(json_spirit::Value_impl<json_spirit::Config_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&, bool) const /usr/include/c++/6/functional:2136
[...]

PR #4034, opened 2017-04-04
fix detection of a transaction that has failed on expect section

winsvega:
connects to #4031

### DIFF
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -589,7 +589,6 @@ int ImportTest::exportTest(bytes const& _output)
 			m_testObject.erase(m_testObject.find("expect"));
 		}
 
-		size_t k = 0;
 		std::map<string, json_spirit::mArray> postState;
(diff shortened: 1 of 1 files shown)


test/tools/libtesteth/ImportTest.cpp
```diff
@@ -600,11 +599,13 @@ int ImportTest::exportTest(bytes const& _output)
...
 			obj2["hash"] = toHex(m_transactions[i].postState.rootHash().asBytes());
-			if (stateIndexesToPrint.size())
-			if (i == stateIndexesToPrint[k] && Options::get().checkstate)
+
+			//Print the post state if transaction has failed on expect section
+			if (Options::get().checkstate)
 			{
-				obj2["postState"] = fillJsonWithState(m_transactions[i].postState);
-				k++;
+				auto errored = std::find(std::begin(stateIndexesToPrint), std::end(stateIndexesToPrint), i);
```
chfast: Shouldn't it be named `auto it` as find returns iterator?

Issue #3613, opened 2017-03-03
EIP 98: removing medstate from transaction receipts


pirapira:
Implementing EIP 98 as part of the Metropolis changes:

* https://github.com/ethereum/EIPs/issues/98

PR #4035, opened 2017-04-11
EIP98 - Remove state root hash from transaction receipt

gumb0:
Implementing Option 1 of EIP - removing state root hash from Receipt RLP

https://github.com/ethereum/cpp-ethereum/issues/3613
https://github.com/ethereum/EIPs/issues/98

- [x] Save zero hash instead of root state hash
- [x] Remove root state hash instead of saving zero hash
- [x] Implement debug RPC methods (trace, storageRangeAt) without relying on state root in receipt
- [x] Fix tests relying on state root in receipts

### DIFF
--- a/libethereum/Block.h
+++ b/libethereum/Block.h
@@ -202,9 +202,10 @@ public:
 	/// Get the bloom filter of a particular transaction that happened in the block.
 	LogBloom const& logBloom(unsigned _i) const { return receipt(_i).bloom(); }
 
-	/// Get the State root hash immediately after the given number of pending transactions have been applied.
+	/// Get the State root hash immediately after all previous transactions before transaction @a _i have been applied.
 	/// If (_i == 0) returns the initial state of the block.
 	/// If (_i == pending().size()) returns the final state of the block, prior to rewards.
+	/// Returns zero hash if intermediate state root is not available in the receipt (the case after EIP98)
 	h256 stateRootBeforeTx(unsigned _i) const;
 
 	// State-change operations
--- a/libethereum/Block.h
+++ b/libethereum/Block.h
@@ -202,10 +202,10 @@ public:
 	/// Get the bloom filter of a particular transaction that happened in the block.
 	LogBloom const& logBloom(unsigned _i) const { return receipt(_i).bloom(); }
 
-	/// Get the State immediately after the given number of pending transactions have been applied.
+	/// Get the State root hash immediately after the given number of pending transactions have been applied.
 	/// If (_i == 0) returns the initial state of the block.
 	/// If (_i == pending().size()) returns the final state of the block, prior to rewards.
-	State fromPending(unsigned _i) const;
+	h256 stateRootBeforeTx(unsigned _i) const;
 
 	// State-change operations
 
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -775,28 +775,6 @@ Block Client::block(h256 const& _blockHash, PopulationStatistics* o_stats) const
 	}
 }
 
-State Client::state(unsigned _txi, h256 const& _blockHash) const
-{
-	try
-	{
-		return block(_blockHash).fromPending(_txi);
-	}
-	catch (Exception& ex)
-	{
-		ex << errinfo_block(bc().block(_blockHash));
-		onBadBlock(ex);
-		return State(chainParams().accountStartNonce);
-	}
-}
-
-eth::State Client::state(unsigned _txi) const
-{
-	DEV_READ_GUARDED(x_postSeal)
-		return m_postSeal.fromPending(_txi);
-	assert(false);
-	return State(chainParams().accountStartNonce);
-}
-
 void Client::flushTransactions()
 {
 	doWork();
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -158,11 +158,19 @@ Executive::Executive(Block& _s, LastHashes const& _lh, unsigned _level):
 }
 
 Executive::Executive(State& _s, Block const& _block, unsigned _txIndex, BlockChain const& _bc, unsigned _level):
-	m_s(_s = _block.fromPending(_txIndex)),
+	m_s(_s = _block.state()),
 	m_envInfo(_block.info(), _bc.lastHashes(_block.info().parentHash()), _txIndex ? _block.receipt(_txIndex - 1).gasUsed() : 0),
 	m_depth(_level),
 	m_sealEngine(*_bc.sealEngine())
 {
+	u256 const rootHash = _block.stateRootBeforeTx(_txIndex);
+	if (rootHash)
+		m_s.setRoot(rootHash);
+	else
+	{
+		m_s.setRoot(_block.stateRootBeforeTx(0));
+		m_s.executeBlockTransactions(_block, _txIndex, _bc.lastHashes(_block.info().parentHash()), *_bc.sealEngine());
+	}
 }
 
 u256 Executive::gasUsed() const
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -157,20 +157,12 @@ Executive::Executive(Block& _s, LastHashes const& _lh, unsigned _level):
 {
 }
 
-Executive::Executive(State& _s, Block const& _block, unsigned _txIndex, BlockChain const& _bc, unsigned _level):
-	m_s(_s = _block.state()),
+Executive::Executive(State& io_s, Block const& _block, unsigned _txIndex, BlockChain const& _bc, unsigned _level):
+	m_s(createIntermediateState(io_s, _block, _txIndex, _bc)),
 	m_envInfo(_block.info(), _bc.lastHashes(_block.info().parentHash()), _txIndex ? _block.receipt(_txIndex - 1).gasUsed() : 0),
 	m_depth(_level),
 	m_sealEngine(*_bc.sealEngine())
 {
-	u256 const rootHash = _block.stateRootBeforeTx(_txIndex);
-	if (rootHash)
-		m_s.setRoot(rootHash);
-	else
-	{
-		m_s.setRoot(_block.stateRootBeforeTx(0));
-		m_s.executeBlockTransactions(_block, _txIndex, _bc.lastHashes(_block.info().parentHash()), *_bc.sealEngine());
-	}
 }
 
 u256 Executive::gasUsed() const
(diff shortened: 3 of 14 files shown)


libethereum/Client.cpp
```diff
@@ -775,28 +775,6 @@ Block Client::block(h256 const& _blockHash, PopulationStatistics* o_stats) const
 	}
 }
 
-State Client::state(unsigned _txi, h256 const& _blockHash) const
```
gumb0: These were not used

libethereum/Block.h
```diff
@@ -202,10 +202,11 @@ class Block
 	/// Get the bloom filter of a particular transaction that happened in the block.
 	LogBloom const& logBloom(unsigned _i) const { return receipt(_i).bloom(); }
 
-	/// Get the State immediately after the given number of pending transactions have been applied.
+	/// Get the State root hash immediately before after all previous transactions before transaction @a _i have been applied.
```
pirapira: "before after" reads strange.

gumb0: Thanks, fixed

libethereum/Executive.cpp
```diff
@@ -141,6 +141,25 @@ string StandardTrace::json(bool _styled) const
 	return _styled ? Json::StyledWriter().write(m_trace) : Json::FastWriter().write(m_trace);
 }
 
+namespace
+{
+
+State& initIntermediateState(State& _s, Block const& _block, unsigned _txIndex, BlockChain const& _bc)
```
chriseth: Can you name this `o_s` to signal that it will be modified?

libethereum/Executive.cpp
```diff
@@ -158,7 +177,7 @@ Executive::Executive(Block& _s, LastHashes const& _lh, unsigned _level):
 }
 
 Executive::Executive(State& _s, Block const& _block, unsigned _txIndex, BlockChain const& _bc, unsigned _level):
```
chriseth: Also here, this should be named `o_s` - but I see this is not part of the PR.

libethereum/State.cpp
```diff
@@ -539,7 +540,27 @@ std::pair<ExecutionResult, TransactionReceipt> State::execute(EnvInfo const& _en
...
+		TransactionReceipt(startGasUsed + e.gasUsed(), e.logs()) :
+		TransactionReceipt(rootHash(), startGasUsed + e.gasUsed(), e.logs());
+	return make_pair(res, receipt);
+}
+
+void State::executeBlockTransactions(Block const& _block, unsigned _txCount, LastHashes const& _lastHashes, SealEngineFace const& _sealEngine)
+{
+	u256 gasUsed = 0;
+	for (unsigned i = 0; i < _txCount; ++i)
+	{
```
chriseth: The whole for-loop body should be a helper function also used elsewhere - or are the cases there more complicated?

libethereum/TransactionReceipt.cpp
```diff
@@ -44,7 +54,13 @@ TransactionReceipt::TransactionReceipt(h256 _root, u256 _gasUsed, LogEntries con
 
 void TransactionReceipt::streamRLP(RLPStream& _s) const
 {
-	_s.appendList(4) << m_stateRoot << m_gasUsed << m_bloom;
+	if (m_stateRoot)
```
chriseth: Can you document this is the header? Perhaps best to put that into the class description.

libweb3jsonrpc/Debug.cpp
```diff
@@ -136,7 +142,15 @@ Json::Value Debug::debug_storageRangeAt(string const& _blockHashOrNumber, int _t
 		Block block = m_eth.block(blockHash(_blockHashOrNumber));
 
 		unsigned const i = ((unsigned)_txIndex < block.pending().size()) ? (unsigned)_txIndex : block.pending().size();
-		State state = block.fromPending(i);
+		State state(block.state());
+		u256 const rootHash = block.stateRootBeforeTx(_txIndex);
+		if (rootHash)
```
chriseth: I have seen this block of code before :-)

gumb0: Extracted this to common function `dev::eth::createIntermediateState` in State.cpp

libethereum/State.cpp
gumb0: Yeah, the other case - `State::execute` needs to get gas, logs, and return value from `Executive`...
But I've moved the common part (initialize-execute-go-finalize) to separate helper method

libethereum/TransactionReceipt.cpp
gumb0: Added a comment

libethereum/State.cpp
```diff
@@ -617,3 +639,17 @@ std::ostream& dev::eth::operator<<(std::ostream& _out, State const& _s)
...
+State& dev::eth::createIntermediateState(State& o_s, Block const& _block, unsigned _txIndex, BlockChain const& _bc)
+{
+	o_s = _block.state();
+	u256 const rootHash = _block.stateRootBeforeTx(_txIndex);
+	if (rootHash)
+		o_s.setRoot(rootHash);
+	else
+	{
+		o_s.setRoot(_block.stateRootBeforeTx(0));
+		o_s.executeBlockTransactions(_block, _txIndex, _bc.lastHashes(_block.info().parentHash()), *_bc.sealEngine());
```
chriseth: Do we need a check for dereferencing sealEngine here?

libethereum/Executive.cpp
```diff
@@ -157,8 +157,8 @@ Executive::Executive(Block& _s, LastHashes const& _lh, unsigned _level):
 {
 }
 
-Executive::Executive(State& _s, Block const& _block, unsigned _txIndex, BlockChain const& _bc, unsigned _level):
-	m_s(_s = _block.fromPending(_txIndex)),
+Executive::Executive(State& o_s, Block const& _block, unsigned _txIndex, BlockChain const& _bc, unsigned _level):
```
chfast: I guess the convention is `io_s`, however I'm not big fun of it.

libethereum/State.cpp
chfast: I wanted to move it to a separate function for a long time.

libethereum/State.cpp
```diff
@@ -539,7 +535,33 @@ std::pair<ExecutionResult, TransactionReceipt> State::execute(EnvInfo const& _en
...
 	}
 
-	return make_pair(res, TransactionReceipt(rootHash(), startGasUsed + e.gasUsed(), e.logs()));
+	TransactionReceipt const receipt = _envInfo.number() >= _sealEngine.chainParams().u256Param("metropolisForkBlock") ?
+		TransactionReceipt(startGasUsed + e.gasUsed(), e.logs()) :
+		TransactionReceipt(rootHash(), startGasUsed + e.gasUsed(), e.logs());
+	return make_pair(res, receipt);
+}
+
+void State::executeTransaction(Executive& _e, Transaction const& _t, OnOpFunc const& _onOp)
```
chfast: It does not depend on State, can we make it a free function somewhere in Executive.h?

libethereum/State.cpp
chfast: I don't think so. I've spot it before, it should not be a pointer type.

test/unittests/libethereum/Block.cpp
```diff
@@ -59,7 +59,7 @@ BOOST_AUTO_TEST_CASE(bStates)
 		OverlayDB const& genesisDB = genesisBlock.state().db();
 		BlockChain const& blockchain = testBlockchain.interface();
 
-		State stateBofore = testBlockchain.topBlock().state();
+		h256 stateBefore = testBlockchain.topBlock().state().rootHash();
```
chfast: Rename to `stateRootBefore`?

test/unittests/libethereum/Block.cpp
```diff
@@ -76,15 +76,15 @@ BOOST_AUTO_TEST_CASE(bStates)
 
 		Block block2 = blockchain.genesisBlock(genesisDB);
 		block2.populateFromChain(blockchain, testBlock.blockHeader().hash());
-		State stateAfterInsert = block2.fromPending(0); //get the state of blockchain on previous block
-		BOOST_REQUIRE(ImportTest::compareStates(stateBofore, stateAfterInsert) == 0);
+		h256 stateAfterInsert = block2.stateRootBeforeTx(0); //get the state of blockchain on previous block
+		BOOST_REQUIRE_EQUAL(stateBefore, stateAfterInsert);
```
chfast: Just a notice, `BOOST_REQUIRE` is overused in tests. I think `BOOST_CHECK` is usually better.

gumb0: I'd say finer-grained tests with a single BOOST_REQUIRE in each one is better than large tests with a series of BOOST_CHECKs

libethereum/State.cpp
gumb0: Maybe better internal in Executive.cpp then?

libethereum/Executive.cpp
gumb0: In some sense it's not really used as input, it is assigned immediately another State in `createIntermediateState`. On the other hand it initializes `m_state` member... Ok, I think I'll rename it

libethereum/State.cpp
chfast: If possible, yes.

test/unittests/libethereum/Block.cpp
chriseth: Always use CHECK if it is possible to continue the test in case it fails, and only REQUIRE if it would not make sense to continue the test.

chfast: I think @gumb0 understand the difference. His point is that in case of failure in CHECK boost will output a long list of errors instead of single one per test.

libethereum/State.cpp
gumb0: Yeah, it's always non-null in practice

libethereum/State.cpp
gumb0: I meant here in State.cpp, sorry, moved it now.

test/unittests/libethereum/Block.cpp
gumb0: Yeah, here it doesn't make sense to proceed with other checks when the first check fails, they are dependent

chfast:
Great job, as usual!

pirapira:
I was asked to review this, and I intend to read the Metropolis changes in all major clients.  I added this to my pending list with priority H and deadline June 15.

Issue #4020, opened 2017-03-28
eth crashes with Assertion `r == 1' failed.


moneroexamples:
I compiled the current master and run it on antergos (arch). I had the same issue before, but not had time to look into more. So now I did and this is fulll trace obtained using gdb:

```
[mwo@ant eth]$ gdb ./eth
GNU gdb (GDB) 7.12.1
Copyright (C) 2017 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./eth...done.
(gdb) run
Starting program: /home/mwo/cpp-ethereum/build/eth/eth 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
cpp-ethereum, a C++ Ethereum client
[New Thread 0x7ffff07fb700 (LWP 4203)]
[New Thread 0x7fffefffa700 (LWP 4204)]
[New Thread 0x7fffef56e700 (LWP 4205)]
[New Thread 0x7fffeed6d700 (LWP 4206)]
[New Thread 0x7fffee0c5700 (LWP 4207)]
cpp-ethereum 1.3.0
  By cpp-ethereum contributors, (c) 2013-2016.
  See the README for contributors and credits.
[New Thread 0x7fffed8c4700 (LWP 4210)]
[New Thread 0x7fffed0c3700 (LWP 4211)]
  ℹ  09:36:44.404|p2p  UPnP device not found.
eth: /home/mwo/cpp-ethereum/libdevcrypto/Common.cpp:359: void dev::crypto::ecdh::agree(const Secret&, const Public&, dev::Secret&): Assertion `r == 1' failed.
 ⚡   09:36:44.411|eth  void dev::p2p::Host::start() 8023 ms
[...]

PR #4040, opened 2017-04-11
Fix ECDH agree

chfast:
We must consider secret and public key being invalid inputs of ecdh::agree().

Fixes https://github.com/ethereum/cpp-ethereum/issues/4020.


libdevcrypto/ECDHE.cpp
```diff
@@ -31,6 +31,7 @@ void ECDHE::agree(Public const& _remote, Secret& o_sharedSecret) const
 		BOOST_THROW_EXCEPTION(InvalidState());
 	
 	m_remoteEphemeral = _remote;
-	ecdh::agree(m_ephemeral.secret(), m_remoteEphemeral, o_sharedSecret);
+	if (!ecdh::agree(m_ephemeral.secret(), m_remoteEphemeral, o_sharedSecret))
+		BOOST_THROW_EXCEPTION(InvalidState());
```
gumb0: You should probably create a new exception type for this, since this looks completely different from the error checked above

chfast: Ok, I will refactor that. Thanks.

chfast:
So I removed the ECDHE to get rid of "invalid state" exception there. And I added InvalidPublic exception to inform about invalid pubkey passed for ECDH. Still there is a lot of mix conventions how to handle errors...

libdevcrypto/ECDHE.cpp
```diff
@@ -1,37 +0,0 @@
...
- */
-
-#include "ECDHE.h"
-
-using namespace dev;
-using namespace dev::crypto;
-
-void ECDHE::agree(Public const& _remote, Secret& o_sharedSecret) const
-{
-	if (m_remoteEphemeral)
```
chriseth: Is this check still in place?

chfast: This check was never effective, because this is agreed during p2p handshake and cannot happen again for the same peer (the same remote public key).

chriseth: I have the feeling that this might be a protection against an attack where you retrieve the private key by performing multiple key agreements. If possible, I would like to keep it as an internal assertion, even (or especially) if it is never triggered by the current code.

chfast:
Now I'm f*cked! I did more than requested and p2p is broken now. So I either revert it and apply simple fix requested by @gumb0 or try to fix it :).

libdevcrypto/ECDHE.cpp
chfast: Fixed. And assert included.

chfast:
Can we merge it soon?

chfast:
I moved the exception to the p2p namespace. The comments also explain where it is caught. The handshake will just fail nicely, what is good.

libp2p/RLPxHandshake.h
```diff
@@ -58,8 +58,6 @@ class RLPXHandshake: public std::enable_shared_from_this<RLPXHandshake>
 	/// Setup outbound connection.
 	RLPXHandshake(Host* _host, std::shared_ptr<RLPXSocket> const& _socket, NodeID _remote): m_host(_host), m_remote(_remote), m_originated(true), m_socket(_socket), m_idleTimer(m_socket->ref().get_io_service()) { crypto::Nonce::get().ref().copyTo(m_nonce.ref()); }
 
-	~RLPXHandshake() {}
```
gumb0: The class has a virtual protected function, shouldn't it have virtual destructor?

chfast: It has because the base class has. The one I removed was actually hiding the virtual one.

chfast: Changed to virtual. Sorry.

Issue #4037, opened 2017-04-11
Switch DB implementation to rocksdb


chfast:
I would like to try rocksdb instead of leveldb. The most important advantage is actually Windows support. They promise to have the same API and we had rocksdb support before (as an option).

Before starting we should add some performance test to it, like block imports, to compare later on.

Also, do #4036 first to have nice build experience from the beginning.

gumb0:
@halfalicious See also my work in https://github.com/ethereum/aleth/pull/4844

halfalicious:
No longer needed as we've switched to pulling in LevelDB via Hunter so we're using the same implementation on all platforms.

PR #4043, opened 2017-04-15
Repair support for RocksDB

je4d:
Further integrate the RocksDB option into cmake, and use the ldb abstratcion in parts of the code that were using leveldb directly


chfast:
Hi @je4d,
Can you tell us why are you working on this?

je4d:
I was trying to sync the blockchain, and at some point it slowed to a crawl. It seemed to be spending most of the time reading from leveldb whilst executing contracts. I read something that indicated that rocksdb may be better suited to eth's workload, so I decided to give it a go, and I had to make changes to get it to build. That's about it.

chfast:
Great. I created https://github.com/ethereum/cpp-ethereum/issues/4037 some time ago, so your work is very helpful.

libdevcore/CMakeLists.txt
```diff
@@ -9,7 +9,13 @@ target_include_directories(devcore SYSTEM PUBLIC ${BOOST_INCLUDE_DIR})
 add_dependencies(devcore BuildInfo.h)
 
 if (NOT EMSCRIPTEN)
-    find_package(LevelDB)
-    target_include_directories(devcore SYSTEM PUBLIC ${LEVELDB_INCLUDE_DIRS})
-    target_link_libraries(devcore Boost::Filesystem Boost::Random Boost::Thread ${LEVELDB_LIBRARIES})
+    if (ROCKSDB)
```
chfast: Can you separate common dependencies from conditional leveldb / rocksdb ones? Also `if (NOT EMSCRIPTEN)` can be remove as it is not needed any more.

libdevcore/dbfwd.h
```diff
@@ -0,0 +1,34 @@
...
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/** @file dbfwd.h
+ * @author Jeff Snyder <jeff-github@caffeinated.me.uk>
```
chfast: We don't continue adding `@file` section if no additional description included, but it is fine if you want keep you name here. In any case remove `dbfwd.h` argument from `@file` tag and the `@date` tag.

PR #4044, opened 2017-04-18
Declare MaxAddress and ZeroAddress as const

pirapira:
I'm now looking into the Metropolis changes.  Let me warm up with trivialities.


chfast:
I wander if could go with `constexpr` here...

pirapira:
Good idea.  I'll try.

pirapira:
@chfast The constructors of `Address` = `FixedHash` are not so `constexpr`ish.  They use `memcpy` and so to initialize the buffer.

chfast:
If not possible to be done quickly give up. But it would be nice task for refactoring.

pirapira:
I give up for now.

PR #4046, opened 2017-04-19
Eip86 CREATE

yann300:
- Set the address of any contract created through a creation transaction to equal sha3(NULL_SENDER + sha3(init code)) % 2**160, where + represents concatenation, replacing the earlier address formula of sha3(rlp.encode([sender, nonce]))
 - Create a new opcode at 0xfb, CREATE_P2SH, which sets the creation address to sha3(sender + sha3(init code)) % 2**160. If a contract at that address already exists, fails and returns 0 as if the init code had run out of gas.


libethereum/Executive.cpp
```diff
@@ -391,6 +400,8 @@ bool Executive::go(OnOpFunc const& _onOp)
 			auto vm = _onOp ? VMFactory::create(VMKind::Interpreter) : VMFactory::create();
 			if (m_isCreation)
 			{
+				if (m_s.addressHasCode(m_newAddress))
```
pirapira: I think this happens only after Metropolis (although I doubt I can trigger this right now).

winsvega:
I'will start working on the tests tomorrow.

test/tools/jsontests/vm.cpp
```diff
@@ -281,6 +282,23 @@ eth::OnOpFunc FakeExtVM::simpleTrace() const
 	};
 }
 
+MetropolisFakeExtVM::MetropolisFakeExtVM(EnvInfo const& _envInfo, unsigned _depth):
+	FakeExtVM(_envInfo, _depth)
+{}
+
+h160 MetropolisFakeExtVM::create(u256 _endowment, u256& io_gas, bytesConstRef _init, OnOpFunc const&, Instruction _creationType)
+{
+	(void)_creationType;
```
pirapira: Do you need this line?

libethereum/Executive.cpp
pirapira: I remember there is a test case about this...

test/tools/jsontests/vm.cpp
yann300: if not it will build (not used parameter are treated as error)

pirapira: This parameter is used below.

libethereum/Executive.cpp
yann300: That stuff will basically never return true before Metropolis. but yeah that worth testing only after metro

test/tools/jsontests/vm.cpp
yann300: well wait... you're right :) don't need this line

pirapira:
Looks OK except the Travis failure.

libevm/VM.cpp
```diff
@@ -211,6 +211,12 @@ void VM::interpretCases()
 		//
 		// Call-related instructions
 		//
+
+		CASE(CREATE_PSH)
+		{
+			m_bounce = &VM::caseCreate;
```
gumb0: It should be available only after Metropolis and generate BadInstruction before. See for example how REVERT handles this

gumb0:
Also the changes are needed in libevmcore/Instruction.cpp - add new instuction to `c_instructionInfo` - it's used to calculate instruction price; and to `c_instructions` - it's used by vmtrace

See my changes for REVERT https://github.com/ethereum/cpp-ethereum/pull/3635/files#diff-eebc659bb21aa390ba67700a3586fb13

test/tools/jsontests/vm.h
```diff
@@ -80,5 +80,12 @@ class FakeExtVM: public eth::ExtVMFace
 	u256 execGas;
 };
 
+class MetropolisFakeExtVM: public FakeExtVM
```
gumb0: This class looks not used

yann300: yep we will try to remove the entire fakeExtVM with @winsvega

libevm/ExtVMFace.h
```diff
@@ -298,7 +298,7 @@ class ExtVMFace
 	virtual void suicide(Address) { sub.suicides.insert(myAddress); }
 
 	/// Create a new (contract) account.
-	virtual h160 create(u256, u256&, bytesConstRef, OnOpFunc const&) { return h160(); }
+	virtual h160 create(u256, u256&, bytesConstRef, OnOpFunc const&, Instruction) { return h160(); }
```
gumb0: I would better make `OnOpFunc` parameter the last one, because it's kind of different than all the others - all others are configuring the creation and this one is a callback

yann300:
will move to https://github.com/ethereum/cpp-ethereum/pull/4051

PR #4048, opened 2017-04-19
fix out of range call to block's receipts

winsvega:
configure blockchain tests mining
make all fields hex and prefixed with 0x

related test update:
https://github.com/ethereum/tests/pull/154

### DIFF
--- a/libethashseal/Ethash.cpp
+++ b/libethashseal/Ethash.cpp
@@ -162,9 +162,10 @@ void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase co
 		BOOST_THROW_EXCEPTION(OutOfGasIntrinsic());
 
 	// Avoid transactions that would take us beyond the block gas limit.
+	// Should check only actual gas used by transaction
 	u256 startGasUsed = _env.gasUsed();
-	if (startGasUsed + (bigint)_t.gas() > _env.gasLimit())
-		BOOST_THROW_EXCEPTION(BlockGasLimitReached() << RequirementError((bigint)(_env.gasLimit() - startGasUsed), (bigint)_t.gas()));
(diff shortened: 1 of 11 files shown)


libethereum/Block.h
```diff
@@ -191,16 +191,16 @@ class Block
 	h256Hash const& pendingHashes() const { return m_transactionSet; }
 
 	/// Get the transaction receipt for the transaction of the given index.
-	TransactionReceipt const& receipt(unsigned _i) const { return m_receipts[_i]; }
+	TransactionReceipt const& receipt(unsigned _i) const { return m_receipts.at(_i); }
```
gumb0: Is there a catch for `std::out_of_range` or at least `std::exception` everywhere where it's called?

winsvega: there is a catch somewhere - it will just print an error message to the error log.

chfast: In case of JSON RPC, shouldn't we response with a JSON error, not with command line log?

winsvega:
that bug was discovered during test framework changes. so...
regarding to RPC we'll have to look at the behaviour when sending an actual RPC request.

libethashseal/Ethash.cpp
```diff
@@ -163,7 +163,7 @@ void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase co
 
 	// Avoid transactions that would take us beyond the block gas limit.
 	u256 startGasUsed = _env.gasUsed();
-	if (startGasUsed + (bigint)_t.gas() > _env.gasLimit())
+	if (startGasUsed + _t.gas() > _env.gasLimit())
```
chriseth: We have to guard against overflow somehow.

PR #4051, opened 2017-04-21
Eip86 CREATE

yann300:


libethereum/Executive.h
```diff
@@ -78,6 +78,11 @@ class StandardTrace
 	DebugOptions m_options;
 };
 
+enum CreationContext: short {
```
gumb0: - Prefer `enum class` to unscoped `enum`
- Brace on separate line
- Better use the full word `Transaction` instead of `Tx`

libevmcore/EVMSchedule.h
```diff
@@ -39,6 +39,7 @@ struct EVMSchedule
 	bool eip150Mode = false;
 	bool eip158Mode = false;
 	bool haveRevert = false;
+	bool haveCreatePsh = false;
```
gumb0: Better rename to `haveCreateP2sh`

libevmcore/Instruction.cpp
```diff
@@ -313,6 +314,7 @@ static const std::map<Instruction,  InstructionInfo> c_instructionInfo =
 	{ Instruction::LOG3,         { "LOG3",           0,     5,     0,  true,       Tier::Special } },
 	{ Instruction::LOG4,         { "LOG4",           0,     6,     0,  true,       Tier::Special } },
 	{ Instruction::CREATE,       { "CREATE",         0,     3,     1,  true,       Tier::Special } },
+	{ Instruction::CREATE_P2SH,   { "CREATE_P2SH",     0,     3,     1,  true,       Tier::Special } },
```
axic: I believe this doesn't have 2 inputs?

yann300: why?

yann300: like value, offset, size

yann300:
@chfast @gumb0 I just  rebase, please merge if you're happy with it

yann300:
@winsvega we need to plug this PR with the eip86 tests also

libethereum/Executive.cpp
```diff
@@ -418,6 +439,14 @@ bool Executive::go(OnOpFunc const& _onOp)
 			m_excepted = toTransactionException(_e);
 			revert();
 		}
+		catch (AddressAlreadyUsed const _e)
```
gumb0: Catch by const reference, then exception won't be copied here
`catch (AddressAlreadyUsed const& _e)`

libethereum/Executive.h
```diff
@@ -78,6 +78,7 @@ class StandardTrace
 	DebugOptions m_options;
 };
 
+enum class CreationContext: short;
```
gumb0: I think you don't need and shouldn't duplicate `: short` here in the forward declaration (this is another nice feature of `enum class`)

libevm/ExtVMFace.h
```diff
@@ -235,6 +235,8 @@ class EnvInfo
 	u256 m_gasUsed;
 };
 
+enum class CreationContext: short;
```
gumb0: The same: you don't need the underlying type in forward declaration

test/tools/jsontests/vm.cpp
```diff
@@ -37,9 +38,28 @@ FakeExtVM::FakeExtVM(EnvInfo const& _envInfo, unsigned _depth):			/// TODO: XXX:
 	ExtVMFace(_envInfo, Address(), Address(), Address(), 0, 1, bytesConstRef(), bytes(), EmptySHA3, _depth)
 {}
 
-h160 FakeExtVM::create(u256 _endowment, u256& io_gas, bytesConstRef _init, OnOpFunc const&)
+h160 FakeExtVM::create(u256 _endowment, u256& io_gas, bytesConstRef _init, CreationContext _creationType, OnOpFunc const&)
 {
-	Address na = right160(sha3(rlpList(myAddress, get<1>(addresses[myAddress]))));
+	unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(eth::Network::MainNetworkTest)).createSealEngine());
```
gumb0: Maybe let's not change this `FakeExtVM` until we need it?

libevm/JitVM.h
```diff
@@ -7,7 +7,7 @@ namespace dev
 {
 namespace eth
 {
-
+
```
gumb0: Unneeded change, just revert this file

libethereum/Executive.h
```diff
@@ -150,7 +151,7 @@ class Executive
 
 	/// Set up the executive for evaluating a bare CREATE (contract-creation) operation.
 	/// @returns false iff go() must be called (and thus a VM execution in required).
-	bool create(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _code, Address _originAddress);
+	bool create(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _code, Address _originAddress, CreationContext _creationType);
```
gumb0: I would actually do this not with an `enum` parameter, but by having three separate interface methods, something like `create`, `createOpcode`, `createP2shOpcode`. Each would calculate new address in its own way and then call the internal implementation function with the rest of the code.

Branching he logic on enum is more error-prone and less maintainable, because if we add a new enum value, we need to search for all the places where we branch on this enum. We'd also get a nicer smaller functions and less nested `if` constructions, if we extract the address calculation from the current big `create`

Sorry for not pointing this out earlier, it's probably not that critical and you could leave it like this if you think it's better

test/tools/jsontests/vm.cpp
yann300: apparently we already use `FakeExtVM`

libethereum/Executive.h
```diff
@@ -150,7 +149,10 @@ class Executive
 
 	/// Set up the executive for evaluating a bare CREATE (contract-creation) operation.
 	/// @returns false iff go() must be called (and thus a VM execution in required).
+	bool executeCreate(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _code, Address _originAddress);
```
gumb0: Make it private if it's not used outside of class

libethereum/Executive.cpp
```diff
@@ -306,19 +306,41 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
...
+	}
+	else
+		return createOpcode(_txSender, _endowment, _gasPrice, _gas, _init, _origin);
+}
+
+bool Executive::createOpcode(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin)
+{
+	u256 nonce = m_s.getNonce(_txSender);
+	if (_txSender != MaxAddress) // EIP86
+		m_s.incNonce(_txSender);
```
gumb0: So now the nonce can be incremented only for CREATE and for create tx before Metropolis, but neither for CREATE_P2SH nor for create tx after Metropolis. Is this correct?

yann300: well that interesting. I think there's no need to increment the nonce... I've asked...

libethereum/Executive.h
yann300: done

libevmcore/Instruction.cpp
```diff
@@ -165,6 +165,7 @@ const std::map<std::string,  Instruction> dev::eth::c_instructions =
 	{ "LOG3", Instruction::LOG3 },
 	{ "LOG4", Instruction::LOG4 },
 	{ "CREATE", Instruction::CREATE },
+	{ "CREATE_P2SH", Instruction::CREATE_P2SH },
```
gumb0: It's renamed now to `CREATE2`, could you rename it everywhere in code? (including `Executive::createP2shOpcode` method)

libethereum/Executive.cpp
```diff
@@ -373,6 +395,9 @@ bool Executive::go(OnOpFunc const& _onOp)
 			auto vm = _onOp ? VMFactory::create(VMKind::Interpreter) : VMFactory::create();
 			if (m_isCreation)
 			{
+				if (m_envInfo.number() >= m_sealEngine.chainParams().u256Param("metropolisForkBlock") && m_s.addressHasCode(m_newAddress))
```
gumb0: This will affect both creation transaction and CREATE2, but this rule is specified only for CREATE2 in EIP. I've asked for the clarification there.

gumb0: This is fine according to Vitalik's clarification

libethereum/Executive.cpp
```diff
@@ -306,19 +306,45 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
...
 
-bool Executive::create(Address _sender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin)
+bool Executive::create(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin)
 {
-	u256 nonce = m_s.getNonce(_sender);
-	if (_sender != MaxAddress) // EIP86
-		m_s.incNonce(_sender);
+	if (m_envInfo.number() >= m_sealEngine.chainParams().u256Param("metropolisForkBlock"))
+	{
+		if (_txSender != MaxAddress) // EIP86
```
gumb0: Move it into `executeCreate`, then you  don't need to repeat it 3 times

libethereum/Executive.h
```diff
@@ -151,6 +150,8 @@ class Executive
 	/// Set up the executive for evaluating a bare CREATE (contract-creation) operation.
 	/// @returns false iff go() must be called (and thus a VM execution in required).
 	bool create(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _code, Address _originAddress);
+	bool createOpcode(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _code, Address _originAddress);
+	bool create2Opcode(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _code, Address _originAddress, bytes _salt);
```
gumb0: Salt is just a 32-byte number, pass it as `u256`

libevmcore/EVMSchedule.h
```diff
@@ -39,6 +39,7 @@ struct EVMSchedule
 	bool eip150Mode = false;
 	bool eip158Mode = false;
 	bool haveRevert = false;
+	bool haveCreateP2sh = false;
```
gumb0: rename to `haveCreate2`

libevm/VMCalls.cpp
```diff
@@ -117,17 +117,32 @@ void VM::caseCreate()
...
+		initSize = (uint64_t)m_SP[2];
+	}
+	else
+	{
+		salt = toBigEndian(m_SP[1]);
+		initOff = (uint64_t)m_SP[2];
+		initSize = (uint64_t)m_SP[3];
+	}
 
+	if (m_OP == Instruction::CREATE2)
```
gumb0: What is this? You already assigined `salt` at line 130

libevmcore/Instruction.h
```diff
@@ -182,11 +182,12 @@ enum class Instruction: uint8_t
 	JUMPC,              ///< alter the program counter - pre-verified
 	JUMPCI,             ///< conditionally alter the program counter - pre-verified
 
-	CREATE = 0xf0,      ///< create a new account with associated code
+	CREATE = 0xf0,      ///< create a new account with associated code. (since EIP86: sha3((MaxAddress + sha3(code))
 	CALL,               ///< message-call into an account
 	CALLCODE,           ///< message-call with another account's code only
 	RETURN,             ///< halt execution returning output data
 	DELEGATECALL,       ///< like CALLCODE but keeps caller's value and sender
+	CREATE2 = 0xfb,		///< create a new account with associated code. sha3((sender + sha3(code))
```
gumb0: Formula in comment is outdated now

gumb0:
Looks like a bunch of tests are failing now

libethereum/Executive.cpp
```diff
@@ -306,19 +306,42 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
...
+	if (m_envInfo.number() >= m_sealEngine.chainParams().u256Param("metropolisForkBlock"))
+	{
+		m_newAddress = right160(sha3(MaxAddress.asBytes() + sha3(_init).asBytes()));
+		return executeCreate(_txSender, _endowment, _gasPrice, _gas, _init, _origin);
+	}
+	else
+		return createOpcode(_txSender, _endowment, _gasPrice, _gas, _init, _origin);
+}
+
+bool Executive::createOpcode(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin)
```
pirapira: `txSender` sounds like the sender of the transaction, but I guess this address points to the contract executing `CREATE`.

libethereum/Executive.cpp
```diff
@@ -306,19 +306,42 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
...
+{
+	u256 nonce = m_s.getNonce(_txSender);
+	m_newAddress = right160(sha3(rlpList(_txSender, nonce)));
+	return executeCreate(_txSender, _endowment, _gasPrice, _gas, _init, _origin);
+}
+
+bool Executive::create2Opcode(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin, u256 _salt)
```
pirapira: Same remark as above about the name of the variable `_txSender`.

libethereum/Executive.cpp
```diff
@@ -306,19 +306,42 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
...
+bool Executive::executeCreate(Address _sender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin)
 {
-	u256 nonce = m_s.getNonce(_sender);
 	if (_sender != MaxAddress) // EIP86
 		m_s.incNonce(_sender);
 
 	m_savepoint = m_s.savepoint();
 
-	m_isCreation = true;
+	m_isCreation = true;
```
pirapira: I see some spaces :cactus: :cactus:

pirapira:
There is a [new rule](https://github.com/ethereum/EIPs/pull/208/files#diff-95b306e6e6d2d6dbb0c5fd3ac4ea45c4R28) that forbids some overwriting.

For example, [this line](https://github.com/ethereum/cpp-ethereum/pull/4051/files#diff-a3d49683ea9c6074fda16d145b8192f8R357) needs to change.

libethereum/Executive.cpp
```diff
@@ -306,19 +306,42 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
...
-bool Executive::create(Address _sender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin)
+bool Executive::create(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin)
+{
+	if (m_envInfo.number() >= m_sealEngine.chainParams().u256Param("metropolisForkBlock"))
+	{
+		m_newAddress = right160(sha3(MaxAddress.asBytes() + sha3(_init).asBytes()));
+		return executeCreate(_txSender, _endowment, _gasPrice, _gas, _init, _origin);
+	}
+	else
+		return createOpcode(_txSender, _endowment, _gasPrice, _gas, _init, _origin);
```
pirapira: It looks a bit strange that `m_newAddress` is set above but not here.

libethereum/Executive.cpp
pirapira: This morning it became clearer "nonempty code OR non-empty nonce" https://github.com/ethereum/EIPs/pull/208/files#diff-95b306e6e6d2d6dbb0c5fd3ac4ea45c4R28  So, a check about the nonce needs to be added here.

libethereum/Executive.cpp
yann300: `m_newAddress` is set in `createOpcode`. `executeCreate` is the private function that is called by every `create...`

libethereum/Executive.cpp
```diff
@@ -418,6 +444,14 @@ bool Executive::go(OnOpFunc const& _onOp)
 			m_excepted = toTransactionException(_e);
 			revert();
 		}
+		catch (AddressAlreadyUsed const& _e)
+		{
+			// EIP86
+			clog(StateSafeExceptions) << "Address already used. " << diagnostic_information(_e);
+			m_gas = 0;
+			m_excepted = toTransactionException(_e);
+			revert();
```
pirapira: I thought, in this case, the topmost stack element should become zero, and the execution should continue.

libethereum/Executive.cpp
pirapira: I suspected something like that, but I was lazy to look up assignments into `m_newAddress`.   It's kind of far away, so I might want to change this in the future (that is, if you keep this as it is).

test/tools/jsontests/vm.cpp
```diff
@@ -37,9 +38,20 @@ FakeExtVM::FakeExtVM(EnvInfo const& _envInfo, unsigned _depth):			/// TODO: XXX:
...
+	/*
+	 * EIP86:
+	 *  creation from:
+	 *   - CREATE: always create using nonce
+	 *   - CREATE2: declared after metropolis fork, create using code
+	 */
+	Address na;
+	if (_op == Instruction::CREATE)
+		na = right160(sha3(rlpList(myAddress, get<1>(addresses[myAddress]))));
+	else if (_op == Instruction::CREATE2)
```
pirapira: Maybe you want to check it's already Metropolis?

libevmcore/Instruction.h
```diff
@@ -182,11 +182,12 @@ enum class Instruction: uint8_t
 	JUMPC,              ///< alter the program counter - pre-verified
 	JUMPCI,             ///< conditionally alter the program counter - pre-verified
 
-	CREATE = 0xf0,      ///< create a new account with associated code
+	CREATE = 0xf0,      ///< create a new account with associated code. (since EIP86: sha3((MaxAddress + sha3(code))
```
pirapira: The address calculation of `CREATE` does not change.

libethereum/Executive.cpp
yann300: well that how OutOfGas is handled right now... true that this is maybe not wanted (even for outOfGas)

pirapira: Ah, I see, we are already running the init code.

axic:
Is there a way to squash some of the 39 commits into a couple? :)

libevm/JitVM.cpp
```diff
@@ -1,16 +1,18 @@
...
 
 namespace dev
 {
 namespace eth
 {
+	
 namespace
 {
-
+
```
axic: Extra whitespace?

libethereum/Executive.cpp
```diff
@@ -373,6 +391,11 @@ bool Executive::go(OnOpFunc const& _onOp)
 			auto vm = _onOp ? VMFactory::create(VMKind::Interpreter) : VMFactory::create();
 			if (m_isCreation)
 			{
+				if (m_envInfo.number() >= m_sealEngine.chainParams().u256Param("metropolisForkBlock") &&
+					m_s.addressHasCode(m_newAddress) &&
+					(m_s.code(m_newAddress).size() != 0 || m_s.getNonce(m_newAddress)))
```
gumb0: This doesn't look right. We need to fail if either code is non-empty or nonce is non-zero, right?
Then it should be `(m_s.addressHasCode(m_newAddress) || m_s.getNonce(m_newAddress) != 0)`

libethereum/Executive.cpp
gumb0: Yeah, this will revert the current frame execution, then the calling frame will deal with putting 0 on stack

libethereum/Executive.h
```diff
@@ -151,6 +150,8 @@ class Executive
 	/// Set up the executive for evaluating a bare CREATE (contract-creation) operation.
 	/// @returns false iff go() must be called (and thus a VM execution in required).
 	bool create(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _code, Address _originAddress);
+	bool createOpcode(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _code, Address _originAddress);
```
gumb0: Better rename `_txSender` to `_sender` here, since  you renamed it in cpp

libethereum/Executive.cpp
```diff
@@ -331,11 +354,6 @@ bool Executive::create(Address _sender, u256 _endowment, u256 _gasPrice, u256 _g
 	// Schedule _init execution if not empty.
 	if (!_init.empty())
 		m_ext = make_shared<ExtVM>(m_s, m_envInfo, m_sealEngine, m_newAddress, _sender, _origin, _endowment, _gasPrice, bytesConstRef(), _init, sha3(_init), m_depth);
-	else if (m_s.addressHasCode(m_newAddress))
```
gumb0: Why do you remove this? I guess this can break a bunch of tests that check consensus around address collision. 
This is the branch for empty init code, so it's not contract creation and I think this should work as earlier.

libethereum/Executive.cpp
```diff
@@ -331,12 +352,13 @@ bool Executive::create(Address _sender, u256 _endowment, u256 _gasPrice, u256 _g
...
+	else if (m_s.addressHasCode(m_newAddress) && m_envInfo.number() < m_sealEngine.chainParams().u256Param("metropolisForkBlock"))
 		// Overwrite with empty code in case the account already has a code
 		// (address collision -- not real life case but we can check it with
 		// synthetic tests).
 		m_s.setNewCode(m_newAddress, {});
 
+
```
pirapira: An extra line.

libevm/VMCalls.cpp
```diff
@@ -117,8 +117,20 @@ void VM::caseCreate()
...
-	uint64_t initSize = (uint64_t)m_SP[2];
+	uint64_t initOff;
+	uint64_t initSize;
+	u256 salt;
+	if (m_OP == Instruction::CREATE)
+	{
+		initOff = (uint64_t)m_SP[1];
+		initSize = (uint64_t)m_SP[2];
+	}
+	else
```
pirapira: I think it's a bit safer to check `m_OP == Instruction::CREATE2`.

pirapira: (To prevent errors when the third create opcode is introduced.)

test/tools/jsontests/vm.cpp
```diff
@@ -37,9 +38,27 @@ FakeExtVM::FakeExtVM(EnvInfo const& _envInfo, unsigned _depth):			/// TODO: XXX:
...
+	 *   - CREATE: always create using nonce
+	 *   - CREATE2: declared after metropolis fork, create using code
+	 */
+	Address na;
+	if (envInfo().number() >= se->chainParams().u256Param("metropolisForkBlock"))
+	{
+		if (_op == Instruction::CREATE)
+			na = right160(sha3(rlpList(myAddress, get<1>(addresses[myAddress]))));
+		else if (_op == Instruction::CREATE2)
+			na = right160(sha3(myAddress.asBytes() + toBigEndian(salt) + sha3(_init).asBytes()));
```
pirapira: Maybe append
```
else
    <assertion failure>;
```
to make sure `na` is always set.

libethereum/Executive.cpp
```diff
@@ -306,9 +306,30 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
...
+bool Executive::createOpcode(Address _sender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin)
 {
 	u256 nonce = m_s.getNonce(_sender);
+	m_newAddress = right160(sha3(rlpList(_sender, nonce)));
+	return executeCreate(_sender, _endowment, _gasPrice, _gas, _init, _origin);
+}
+
+bool Executive::create2Opcode(Address _sender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin, u256 _salt)
+{
+	m_newAddress = right160(sha3(_sender.asBytes() + toBigEndian(_salt) + sha3(_init).asBytes()));
```
pirapira: Maybe it's a good thing to define a function that computes the new address for CREATE2, in order to avoid duplication here and in jitvm.

pirapira: Not in jitvm, but in here: https://github.com/ethereum/cpp-ethereum/pull/4051/files#diff-6a118ab066bfd6ba1615e8c47d7a6ff7R56

libevm/JitVM.cpp
```diff
@@ -1,16 +1,17 @@
...
 
 namespace dev
 {
 namespace eth
 {
+	
 namespace
-{
-
+{
```
pirapira: The diff around here looks a bit weird.

pirapira:
Please rebase.

gumb0:
Moved the check for contract address collision (implementing requirement https://github.com/ethereum/EIPs/pull/208/files#diff-95b306e6e6d2d6dbb0c5fd3ac4ea45c4R28) from `Executive::go` to `Executive::executeCreate`, because it should be checked always even when init code is empty (in which case `Executive::executeCreate` won't be called)

test/tools/jsontests/vm.cpp
```diff
@@ -37,9 +38,28 @@ FakeExtVM::FakeExtVM(EnvInfo const& _envInfo, unsigned _depth):			/// TODO: XXX:
...
-	Address na = right160(sha3(rlpList(myAddress, get<1>(addresses[myAddress]))));
+	unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(eth::Network::MainNetworkTest)).createSealEngine());
+	/*
+	 * EIP86:
+	 *  creation from:
+	 *   - CREATE: always create using nonce
+	 *   - CREATE2: declared after metropolis fork, create using code
+	 */
+	Address na;
+	if (envInfo().number() >= se->chainParams().u256Param("metropolisForkBlock"))
```
gumb0: We should remove this code, it's not needed by anyone, FakeExtVM is used only  with Frontier rules in VM tests only

yann300:
reopened https://github.com/ethereum/cpp-ethereum/pull/4172

libethereum/Executive.cpp
```diff
@@ -306,9 +306,30 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
 	return !m_ext;
 }
 
-bool Executive::create(Address _sender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin)
+bool Executive::create(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin)
```
pirapira: Maybe use `u256 const&` instead of `u256` because now these arguments are passed deeper and deeper.  Also `Address`.

PR #4054, opened 2017-04-23
Implement bitwise shifting (EIP145)

axic:
Implements https://github.com/ethereum/EIPs/pull/215


libevm/VM.cpp
```diff
@@ -573,6 +573,47 @@ void VM::interpretCases()
 		}
 		NEXT
 
+#ifdef EVM_USE_BITSHIFT
+		CASE(SHL)
+		{
+			ON_OP();
+			updateIOGas();
+
+			/// TODO: confirm shift >= 256 results in 0
```
chfast: We should add unit tests for edge cases in boost::multiprecision.

axic: That is done in #4064. So we should add appropriate boundary checks here (and masking).

libevm/VM.cpp
```diff
@@ -573,6 +573,48 @@ void VM::interpretCases()
...
 		NEXT
 
+#ifdef EVM_USE_BITSHIFT
+		CASE(SHL)
+		{
+			ON_OP();
+			updateIOGas();
+
+			/// This workarounds a bug in Boost, ...
+			u256 mask = (u256(1) << (256 - m_SP[1])) - 1;
```
pirapira: I don't know what happens when `m_SP[1]` is more than 256.
* Is the underflown subtraction defined?
* Isn't the shift very slow when `256 - m_SP[1]` is a big number?

axic: The EIP says `If the shift amount (`arg2`) is greater or equal 256 the result is 0.` so the above code should be updated accordingly.

axic:
Need to swap operands as per latest changes.

gcolvin:
@axic @chfast There is substantial overlap with my #4569.  This one is the one that should be pulled.  However,  you will want libevm/Instruction.h and Instruction.cpp from #4569.  Also convention I am using for an experimental feature like this is to place
```
+#ifndef EIP_145
 +	#define EIP_145 false
 +#endif
```
in libevm/VMConfig.h and wall off the code for the feature with `#if EIP_145`.

gcolvin:
@axic @chfast Inspecting our code there are many differences.  For starters.  I am taking the item on the top of the stack `SP[0]` as the amount to shift by, which is how I read the spec. E.g.
> The `SHL` instruction (shift left) pops 2 values from the stack, `arg1` and `arg2`, and pushes on the stack the second popped value `arg2` shifted to the left by the number of bits in the first popped value `arg1`.  So I think that's wrong.

Otherwise our code is just different.
* For SHL I'm not working around whatever Boost bug you are, but I don't know if my approach triggers the bug.
* For SHR I explicitly test for a shift amount >= 256, which might not be needed.
* For SAR I use two shifts and a mask rather than division and exponentiation.

So perhaps my code is wrong, but please look.

libevm/VM.cpp
```diff
@@ -621,6 +621,62 @@ void VM::interpretCases()
...
+				if (shiftee & hibit)
+					m_SPP[0] = allbits;
+				else
+					m_SPP[0] = 0;
+			}
+			else
+			{
+				uint64_t amount = uint64_t(m_SP[0]);
+			    m_SPP[0] = shiftee >> amount;
+			    m_SPP[0] |= allbits << (256 - amount);
```
chfast: I think you have to check `shiftee & hibit` here as well.

libevm/VM.cpp
```diff
@@ -621,6 +621,66 @@ void VM::interpretCases()
...
+#ifdef EVM_USE_BITSHIFT
+		CASE(SHL)
+		{
+			ON_OP();
+			updateIOGas();
+
+			if (m_SP[0] >= 256)
+				m_SPP[0] = 0;
+			else
+				m_SPP[0] = m_SP[1] << uint64_t(m_SP[0]);
```
axic: Could also just use `unsigned` and not the specific `uint64_t` since it always fits into 8 bits.

libevm/VM.cpp
```diff
@@ -621,6 +621,66 @@ void VM::interpretCases()
...
+				m_SPP[0] = m_SP[1] >> uint64_t(m_SP[0]);
+		}
+		NEXT
+
+		CASE(SAR)
+		{
+			ON_OP();
+			updateIOGas();
+
+			static u256 const hibit = u256(1) << 255;
```
axic: I think I'd prefer a slower, but definitely correct implementation for this case to ensure everything is working properly. Once multiple clients agree on tests, it can be optimised for speed.

libevm/VM.cpp
```diff
@@ -621,6 +621,53 @@ void VM::interpretCases()
...
+			u256 shift = m_SP[0];
+			if (shift >= 256)
+			{
+				if (value >= 0)
+					m_SPP[0] = 0;
+				else
+					m_SPP[0] = s2u(-1);
+			}
+			else
+				m_SPP[0] = s2u(divWorkaround(value, exp256(2, shift)));
```
axic: This was broken due to EVM DIV not rounding the same way.

axic:
I wonder if it is possible to do the RC5 changes in a separate pull request as it doesn't seem related to this PR?

libevm/VMConfig.h
```diff
@@ -25,9 +25,13 @@ namespace eth
 //
 // interpreter configuration macros for development, optimizations and tracing
 //
+// EIP_145                - bitwise shifting
```
axic: Out of curiosity, why is `EIP_145` better/more expressive than `EVM_USE_BITSHIFT` ?

axic:
Where are the unit tests being run? What did change in `jsontests` submodule?

libevm/VMConfig.h
gcolvin: It tells the reader where to find the spec for the feature.  And it's consistent with the others.

gcolvin:
The relationship is the change to use the solidity shift operators rather than exponentiation.  The shifts in rc5 are data-dependent, and will be a good stress test.

gcolvin:
If you mean libevm/tests/unittests/performance the tests there are run manually with tests.mk - testing performance on virtualized machines is useless.  shift-tests.asm doesn't really belong there so far as being a performance test, but I don't know where else to put it.

gcolvin:
As far as I know nothing changed in jsontests; I just did `git submodule update --init`.  But apparently git thinks something changed, and I'm guessing doing `git commit -a ...` might have caused the problem?

gcolvin:
@axic It appears my attempt to remove jsontests from the PR failed, although it did allow for an automatic merge again.  I will refrain from further attempts to back this out and, with apologies, leave it to you.

libevm/VM.cpp
gcolvin: Done.

libevm/VM.cpp
gcolvin: Do you think the code I have now is mathematically incorrect?  Or are you worried about bugs in boost's library?

pirapira:
Please rebase.

axic:
@gumb0 @pirapira @chfast is it possible to include a `genesis.json` "enable option" field in this PR?

gumb0:
It should be possible, you'd need to add `EIP145ForkBlok` to `ChainOperationParams`, parse it in `ChainParams.cpp`, also add a flag to `EVMSchedule` and define `EIP145Schedule` and check the flag in EVM

libevm/VMConfig.h
chfast: I prefer using a meaning names, because I'm usually confused by numbers.
As a compromise, can it be `EIP145_BITSHIFT` or `EVM_EIP145_BITSHIFT`?

axic:
@gumb0 would you be able to implement that? Otherwise this PR can be merged now as well, but having an option to enable it runtime would be nice so that it can be used in solidity testing.

libevm/VMConfig.h
gcolvin: I much prefer the compromise.

gumb0:
I'll try to implement it some time soon

chfast:
@gumb0, I started to play with this. Nothing is ready yet, if you want help let me know.

axic:
Well alternatively we could include this under "constantinople".

axic:
@pirapira @gumb0 @chfast I'd say merge this as is and add runtime support in a separate PR (either via `constantinople` or a new schedule).

pirapira:
@axic `testeth --filltests` needs this feature under "Constantinople" schedule, in order to add GeneralStateTests about it.  I think `testeth` can include this branch as a submodule and do some kind of modifications there.

axic:
How about adding support for the constantinople schedule in this PR then? Shouldn't take too long.

axic:
Or could merge this and help @tcsiwula finish #4838

axic:
@gumb0 @chfast I think this is ready for merging now.

axic:
Merged this as every test succeeded except the code coverage :)

axic:
@gcolvin @tcsiwula thank you for your help finishing this!

axic:
@winsvega @pirapira @holiman everything's ready for writing bitwise shifting state tests now 😉

gcolvin:
@axic Sorry to be late to the game.  I believe the runtime tests in VM.cpp can more efficiently be handled in VM::optimize by conditionally replacing the unimplemented instructions with Instruction::INVALID. 
https://github.com/ethereum/cpp-ethereum/blob/a3d93a8fb80c2620449e4d399507dd5ac77ecaae/libevm/VMOpt.cpp#L86

axic:
@gcolvin it seems that the `optimize` function is not even used for the byzantium opcodes. I think it may be a good idea to implement that in a separate PR.

gcolvin:
@axic Unless jump destinations are no longer verified the loop I pointed at is needed, and conditionally invalid opcodes can be tested for there, whether or not my work on bytecode optimization is being abandoned.  Once replaced with INVALID these opcodes will be detected by the opcode dispatch without incurring extra overhead on every opcode execution.  It would be small change to this PR, or a small new PR, but I haven't time to make it right now.

axic:
>  it seems that the optimize function is not even used for the byzantium opcodes. 

... aka it doesn't even replace Byzantium opcodes with `INVALID` on non-Byzantium mode.

gcolvin:
Also Known As what?  I'm not understanding you, sorry.

axic:
The code you have linked to does not replace any Byzantium opcode with the invalid opcode. Not sure how can I write this sentence differently.

gcolvin:
The 'aka' abbreviation threw me off.  I'm saying that the code I linked to should be changed to replace Byzantium opcodes with INVALID in non-Byzantium mode.

gcolvin:
Something like this, and remove the tests in VM.cpp
```
	for (size_t pc = 0; pc < nBytes; ++pc)
	{
		Instruction op = Instruction(m_code[pc]);
		TRACE_OP(2, pc, op);

		if (
			op == Instruction:: SHL ||
			op == Instruction::SHR ||
			op == Instruction::SAR
		)
		{
			if (!m_schedule->haveBitwiseShifting)
				m_code[pc] = (byte)Instruction::INVALID;
		}
	. . . 
```

axic:
I understand how it can be added, all I am saying that it is not present for `returndatacopy` and `returndatasize` at the moment.

As a result, this isn't something which was missed in this PR, rather than it is a new improvement, which can be done separately.

gcolvin:
Aha.  I didn't understand what you didn't understand ;)  I agree it can be done separately.  I just wanted to make note of it.  I haven't had much time for this code, but try to notice when changes go in.

Issue #3616, opened 2017-03-03
EIP214: STATICCALL instruction


pirapira:
Implementing the STATICCALL instruction as specified in EIP 214.

The newest doc is https://github.com/ethereum/EIPs/pull/214

PR #4057, opened 2017-04-24
EIP214: STATICCALL

yann300:
Resolves https://github.com/ethereum/cpp-ethereum/issues/3616.

https://github.com/ethereum/EIPs/pull/214

### DIFF
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -240,7 +240,7 @@ bool Executive::execute()
 
 bool Executive::call(Address _receiveAddress, Address _senderAddress, u256 _value, u256 _gasPrice, bytesConstRef _data, u256 _gas)
 {
-	CallParameters params(_senderAddress, _receiveAddress, _receiveAddress, _value, _value, _gas, _data, {});
+	CallParameters params{_senderAddress, _receiveAddress, _receiveAddress, _value, _value, _gas, _data, {}};
 	return call(params, _gasPrice, _senderAddress);
 }
 
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -240,7 +240,7 @@ bool Executive::execute()
 
 bool Executive::call(Address _receiveAddress, Address _senderAddress, u256 _value, u256 _gasPrice, bytesConstRef _data, u256 _gas)
 {
-	CallParameters params{_senderAddress, _receiveAddress, _receiveAddress, _value, _value, _gas, _data, {}};
+	CallParameters params(_senderAddress, _receiveAddress, _receiveAddress, _value, _value, _gas, _data, {});
 	return call(params, _gasPrice, _senderAddress);
 }
 
@@ -297,7 +297,7 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
 		{
 			bytes const& c = m_s.code(_p.codeAddress);
 			h256 codeHash = m_s.codeHash(_p.codeAddress);
-			m_ext = make_shared<ExtVM>(m_s, m_envInfo, m_sealEngine, _p.receiveAddress, _p.senderAddress, _origin, _p.apparentValue, _gasPrice, _p.data, &c, codeHash, m_depth);
+			m_ext = make_shared<ExtVM>(m_s, m_envInfo, m_sealEngine, _p.receiveAddress, _p.senderAddress, _origin, _p.apparentValue, _gasPrice, _p.data, &c, codeHash, m_depth, _p.staticCall);
 		}
 	}
 
--- a/libethereum/ExtVM.h
+++ b/libethereum/ExtVM.h
@@ -43,8 +43,8 @@ class ExtVM: public ExtVMFace
 {
 public:
 	/// Full constructor.
-	ExtVM(State& _s, EnvInfo const& _envInfo, SealEngineFace const& _sealEngine, Address _myAddress, Address _caller, Address _origin, u256 _value, u256 _gasPrice, bytesConstRef _data, bytesConstRef _code, h256 const& _codeHash, unsigned _depth = 0):
-		ExtVMFace(_envInfo, _myAddress, _caller, _origin, _value, _gasPrice, _data, _code.toBytes(), _codeHash, _depth), m_s(_s), m_sealEngine(_sealEngine)
+	ExtVM(State& _s, EnvInfo const& _envInfo, SealEngineFace const& _sealEngine, Address _myAddress, Address _caller, Address _origin, u256 _value, u256 _gasPrice, bytesConstRef _data, bytesConstRef _code, h256 const& _codeHash, unsigned _depth = 0, bool _staticCall = false):
+		ExtVMFace(_envInfo, _myAddress, _caller, _origin, _value, _gasPrice, _data, _code.toBytes(), _codeHash, _depth, _staticCall), m_s(_s), m_sealEngine(_sealEngine)
 	{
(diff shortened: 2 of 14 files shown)


libevm/VM.cpp
```diff
@@ -214,19 +214,25 @@ void VM::interpretCases()
 		
 		CASE(CREATE)
 		{
+			if (m_schedule->haveStaticCall && m_ext->staticCall)
+				throwBadInstruction();
```
chfast: Should be `throwDisallowedStateChange()`.

libevm/VM.cpp
```diff
@@ -214,19 +214,25 @@ void VM::interpretCases()
...
-
-			// Pre-homestead
-			if (!m_schedule->haveDelegateCall)
-				throwBadInstruction();
-
+		CASE(STATICCALL)
 		CASE(CALL)
 		CASE(CALLCODE)
 		{
+			// Pre-homestead
```
chfast: Remove this comment. It does not provided any information any more, as we are so used to hard-forks.

libevm/VMCalls.cpp
```diff
@@ -159,11 +167,13 @@ bool VM::caseCallSetup(CallParameters *callParams, bytesRef& o_output)
 {
 	m_runGas = toInt63(m_schedule->callGas);
 
-	if (m_OP == Instruction::CALL && !m_ext->exists(asAddress(m_SP[1])))
+	callParams->staticCall = (m_OP == Instruction::STATICCALL || m_ext->staticCall);
+
+	if ((m_OP == Instruction::CALL || m_OP == Instruction::STATICCALL) && !m_ext->exists(asAddress(m_SP[1])))
```
chfast: Should we charge for new account creation in case of STATICCALL? @chriseth @vbuterin?

chfast:
clang has showed an important issue. We missed something...

libethereum/ExtVM.h
```diff
@@ -94,6 +94,7 @@ class ExtVM: public ExtVMFace
 private:
 	State& m_s;  ///< A reference to the base state.
 	SealEngineFace const& m_sealEngine;
+	bool m_staticCall;
```
chfast: This is not needed.

libevm/ExtVMFace.h
```diff
@@ -201,6 +201,7 @@ struct CallParameters
 	u256 apparentValue;
 	u256 gas;
 	bytesConstRef data;
+	bool staticCall;
```
chfast: Initialize it here too: `bool staticCall = false`.

libevm/VM.cpp
```diff
@@ -214,19 +214,24 @@ void VM::interpretCases()
 		
 		CASE(CREATE)
 		{
+			if (m_schedule->haveStaticCall && m_ext->staticCall)
```
gumb0: Do we really need a check for `m_schedule->haveStaticCall` here and in other places?
I think it's redundant, it can't be a static call if don't have it 😄

winsvega:
CALLDATA for staticcall is not transfered to the subcall
strange gasUsed count when using different values for memory arguments of STATICCALL

chfast:
Do you want me to help with rebase and cleaning up the history?

libethereum/Executive.cpp
```diff
@@ -240,7 +240,7 @@ bool Executive::execute()
 
 bool Executive::call(Address _receiveAddress, Address _senderAddress, u256 _value, u256 _gasPrice, bytesConstRef _data, u256 _gas)
 {
-	CallParameters params{_senderAddress, _receiveAddress, _receiveAddress, _value, _value, _gas, _data, {}};
+	CallParameters params(_senderAddress, _receiveAddress, _receiveAddress, _value, _value, _gas, _data, {});
```
chfast: This seems to be unnecessary change.

libevm/VM.cpp
chfast: I agree. Keep only `m_ext->staticCall` check.
You can add before `if`: `assert(!staticCall || haveStaticCall)` which checks implication `staticCall => haveStaticCall`.

libevm/VM.cpp
```diff
@@ -265,6 +270,9 @@ void VM::interpretCases()
 
 		CASE(SUICIDE)
 		{
+			if (m_schedule->haveStaticCall && m_ext->staticCall)
```
chfast: Same here.

libevm/VM.cpp
```diff
@@ -340,6 +348,9 @@ void VM::interpretCases()
 
 		CASE(LOG0)
 		{
+			if (m_schedule->haveStaticCall && m_ext->staticCall)
```
chfast: Same here.

libevm/VMCalls.cpp
```diff
@@ -159,14 +167,16 @@ bool VM::caseCallSetup(CallParameters *callParams, bytesRef& o_output)
...
```
chfast: STATICCALL does not have a value argument, so this check does not apply to it.
cc @pirapira.

pirapira: That's right.

test/tools/jsontests/StateTests.cpp
```diff
@@ -178,6 +178,7 @@ BOOST_AUTO_TEST_CASE(stRevertTest){}
 
 //Metropolis Tests
 BOOST_AUTO_TEST_CASE(stStackTests){}
+BOOST_AUTO_TEST_CASE(stStaticCall){}
```
chfast: @winsvega you forgot to update jsontests.

libevm/VMCalls.cpp
```diff
@@ -182,14 +189,17 @@ bool VM::caseCallSetup(CallParameters *callParams, bytesRef& o_output)
...
-	if (m_OP == Instruction::CALL && !m_ext->exists(asAddress(m_SP[1])))
+	callParams->staticCall = (m_OP == Instruction::STATICCALL || m_ext->staticCall);
+
+	Address destinationAddr = asAddress(m_SP[1]);
+	if (m_OP == Instruction::CALL && !m_ext->exists(destinationAddr))
 		if (m_SP[2] > 0 || m_schedule->zeroValueTransferChargesNewAccountGas())
 			m_runGas += toInt63(m_schedule->callNewAccountGas);
 
-	if (m_OP != Instruction::DELEGATECALL && m_SP[2] > 0)
+	if ((m_OP == Instruction::CALL || m_OP == Instruction::CALLCODE) && m_SP[2] > 0)
```
chriseth: I hope this won't get us into trouble: We change for value transfer for callcode, although we do not consider it state changing.

chriseth:
Does it make sense to add a check somewhere which verifies that the state root is unchanged after a staticcall? Is that possible from an architectural standpoint?

libevm/VMCalls.cpp
pirapira: What would be the test case for this???

pirapira: CALLCODE with positive value transfer... probably we already have it...

chfast:
@chriseth, it might be hard because we don't update the state root until the end of a transaction.

libevm/VMCalls.cpp
pirapira: Added a test case in the Metropolis testing table https://docs.google.com/spreadsheets/d/1xat7UI8GtB4ZGVdlK5_XQSHJZaMThi4SrlcL8XMZb5Q/edit#gid=1412823514&range=C21

winsvega:
we could check that state root is not changed in tests.

PR #4058, opened 2017-04-24
fix precompiled defenition in genesis.json

winsvega:
there was a missing ','  at the genesis json for tests


libethashseal/genesis/mainNetworkTest.cpp
```diff
@@ -55,7 +55,8 @@ R"E(
 		"0000000000000000000000000000000000000001": { "precompiled": { "name": "ecrecover", "linear": { "base": 3000, "word": 0 } } },
 		"0000000000000000000000000000000000000002": { "precompiled": { "name": "sha256", "linear": { "base": 60, "word": 12 } } },
 		"0000000000000000000000000000000000000003": { "precompiled": { "name": "ripemd160", "linear": { "base": 600, "word": 120 } } },
-		"0000000000000000000000000000000000000004": { "precompiled": { "name": "identity", "linear": { "base": 15, "word": 3 } } }
+		"0000000000000000000000000000000000000004": { "precompiled": { "name": "identity", "linear": { "base": 15, "word": 3 } } },
+		"0000000000000000000000000000000000000005": { "precompiled": { "name": "modexp" } }
```
pirapira: Is the mainNetworkTest already supposed to include metropolis changes?

winsvega: mainNetworkTest is to be similar as mainNetwork config with same transition block numbers but without presale genesis state

PR #4059, opened 2017-04-24
Fix eth_call RPC method

gumb0:
This change reflects the fact that we now have both zero signature transactions (eip86, these should be invalid before Metropolis) and actually unsigned transactions created by eth_call. This is expressed by having `boost::optional<SignatureStruct>` member in the `TransactionBase` class

### DIFF
--- a/libethcore/SealEngine.cpp
+++ b/libethcore/SealEngine.cpp
@@ -54,7 +54,7 @@ void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, Transactio
 		(_t.value() != 0 || _t.gasPrice() != 0 || _t.nonce() != 0))
 			BOOST_THROW_EXCEPTION(InvalidZeroSignatureTransaction() << errinfo_got((bigint)_t.gasPrice()) << errinfo_got((bigint)_t.value()) << errinfo_got((bigint)_t.nonce()));
 
-	if (_env.number() >= chainParams().u256Param("homsteadForkBlock") && (_ir & ImportRequirements::TransactionSignatures))
+	if (_env.number() >= chainParams().u256Param("homsteadForkBlock") && (_ir & ImportRequirements::TransactionSignatures) && _t.hasSignature())
 		_t.checkLowS();
 }
 
--- a/libethcore/Transaction.cpp
+++ b/libethcore/Transaction.cpp
@@ -73,7 +73,7 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
 		m_vrs = SignatureStruct{ r, s, v };
 
 		if (hasZeroSignature())
-			m_chainId = m_vrs.v;
+			m_chainId = m_vrs->v;
 		else
 		{
 			if (v > 36)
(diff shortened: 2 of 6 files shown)


libethcore/Transaction.cpp
```diff
@@ -130,6 +133,14 @@ Address const& TransactionBase::sender() const
 	return m_sender;
 }
 
+SignatureStruct const& TransactionBase::signature() const
+{ 
+	if (!m_vrs)
```
yann300: does it make sense to throw here?
has  `Transaction` support now to be unsigned, it would maybe make sense to return null here

libethcore/Transaction.cpp
```diff
@@ -153,8 +164,11 @@ void TransactionBase::streamRLP(RLPStream& _s, IncludeSignature _sig, bool _forE
 
 	if (_sig)
 	{
+		if (!m_vrs)
```
yann300: don't we need to check for `is_initialized` here?

libethcore/Transaction.cpp
gumb0: What do you mean by null? Returning default-constructed `SignatureStruct` would mean `[0, 0, 0]` signature which is now considered a special-case valid signature (different thing from not having `m_vrs` at all)

So I think throwing is fine, since this method shouldn't be called for unsigned transactions at all.

libethcore/Transaction.cpp
gumb0: This is exactly this check, because `operator bool()` for `optional` calls `is_initialized`

libethcore/Transaction.cpp
gumb0: We could return empty `optional<SignatureStruct>`, but I don't thinks it's any better, the user still would need to check if the returned value is empty or not. And if he forgets he'll just get a crash instead of uncaught exception

yann300: yeah by null I guess I meant  an uninittialized struct.
Having this function to throw is not something that I would have expected if i would use it, 
but if the others are fine by throwing here, that's ok

libethcore/SealEngine.cpp
```diff
@@ -54,7 +54,7 @@ void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, Transactio
 		(_t.value() != 0 || _t.gasPrice() != 0 || _t.nonce() != 0))
 			BOOST_THROW_EXCEPTION(InvalidZeroSignatureTransaction() << errinfo_got((bigint)_t.gasPrice()) << errinfo_got((bigint)_t.value()) << errinfo_got((bigint)_t.nonce()));
 
-	if (_env.number() >= chainParams().u256Param("homsteadForkBlock") && (_ir & ImportRequirements::TransactionSignatures))
+	if (_env.number() >= chainParams().u256Param("homsteadForkBlock") && (_ir & ImportRequirements::TransactionSignatures) && _t.hasSignature())
```
yann300: do we need to check for `!hasZeroSignature` as well?

gumb0: I guess not necessary, because `checkLowS` considers zero-sig transaction valid.

libethcore/Transaction.h
```diff
@@ -136,13 +138,14 @@ class TransactionBase
...
+	/// @returns true if the transaction was signed
+	bool hasSignature() const { return m_vrs.is_initialized(); }
 
-	/// @returns the signature of the transaction. Encodes the sender.
-	SignatureStruct const& signature() const { return m_vrs; }
+	/// @returns true if the transaction was signed with zero signature
+	bool hasZeroSignature() const { return m_vrs && !m_vrs->s && !m_vrs->r; }
 
-	bool hasZeroSignature() const { return !m_vrs.s && !m_vrs.r; }
+	/// @returns the signature of the transaction. Encodes the sender.
```
yann300: it does not encode the sender right?
could we just add a comment saying that this function throw if transaction is unsigned?

gumb0: yeah, this is the old confusing comment. That didn't meant the process of encoding, but that the signature has a sender encoded in it. I'll change this and add a not about exception

yann300:
looks good
@chfast @chriseth ?

PR #4060, opened 2017-04-25
VM: Remove unnecessary check for output buffer size

chfast:
Unless the intention was to check for the bool flag, this check is not needed (copyTo will not copy anything if the output is empty).


chfast:
Changed. Thanks a lot.

PR #4062, opened 2017-04-25
EIP211: RETURNDATA

chfast:
https://github.com/ethereum/EIPs/pull/211


chfast:
@winsvega can you generate tests for this?

libevm/VM.h
```diff
@@ -104,6 +104,9 @@ class VM: public VMFace
 	// space for code
 	bytes m_code;
 
+	/// Space for return data.
```
pirapira: Why three `/`s?  (I was not looking for them but I saw them.)

libevmcore/Instruction.cpp
```diff
@@ -204,6 +54,8 @@ static const std::map<Instruction,  InstructionInfo> c_instructionInfo =
 	{ Instruction::ADDMOD,       { "ADDMOD",         0,     3,    1,  false,       Tier::Mid } },
 	{ Instruction::MULMOD,       { "MULMOD",         0,     3,    1,  false,       Tier::Mid } },
 	{ Instruction::SIGNEXTEND,   { "SIGNEXTEND",     0,     2,    1,  false,       Tier::Low } },
+	{ Instruction::RETURNDATASIZE,{"RETURNDATASIZE", 0,     0,    1,  false,       Tier::Low } },
+	{ Instruction::RETURNDATACOPY,{"RETURNDATACOPY", 0,     3,    0,  false,       Tier::Low } },
```
pirapira: Why do the Tiers look different from those of `CALLDATASIZE` and `CALLDATACOPY`?

libevm/VM.h
chfast: `///` starts a Doxygen comment.

libevmcore/Instruction.cpp
chfast: I think I was so focused on setting the number of args right I forgot about the costs.

libevmcore/Instruction.cpp
```diff
@@ -204,6 +54,8 @@ static const std::map<Instruction,  InstructionInfo> c_instructionInfo =
 	{ Instruction::ADDMOD,       { "ADDMOD",         0,     3,    1,  false,       Tier::Mid } },
 	{ Instruction::MULMOD,       { "MULMOD",         0,     3,    1,  false,       Tier::Mid } },
 	{ Instruction::SIGNEXTEND,   { "SIGNEXTEND",     0,     2,    1,  false,       Tier::Low } },
+	{ Instruction::RETURNDATASIZE,{"RETURNDATASIZE", 0,     0,    1,  false,       Tier::VeryLow } },
```
gumb0: Are the prices still not right? `Tier::VeryLow` is 3, but `RETURNDATASIZE`  should be 2, which is `Tier::Base`, like `CALLDATASIZE`
And `RETURNDATACOPY` should be `Tier::VeryLow` ?

libevmcore/Instruction.cpp
```diff
@@ -204,6 +54,8 @@ static const std::map<Instruction,  InstructionInfo> c_instructionInfo =
...
+	{ Instruction::RETURNDATACOPY,{"RETURNDATACOPY", 0,     3,    0,  false,       Tier::Low } },
```
gumb0: I guess it should have `true` for side-effects like `CALLDATACOPY`

libevmcore/Instruction.cpp
chfast: Of course. I'm really sorry, I must have been distracted. BTW, "side effect" is not used, but this should be addressed in #4072.

libevm/VMCalls.cpp
```diff
@@ -139,12 +142,20 @@ void VM::caseCreate()
...
+
+	// Clear the return data buffer. This will not free the memory.
+	m_returnData.clear();
+
 	bytesRef output;
 	if (caseCallSetup(callParams.get(), output))
 	{
 		std::pair<bool, owning_bytes_ref> callResult = m_ext->call(*callParams);
 		callResult.second.copyTo(output);
+		m_returnData.assign(callResult.second.begin(), callResult.second.end());
```
gumb0: Could we define `m_returnData` as `owning_bytes_ref` and then move here instead of copying?

(if not, using helpers like `owning_bytes_ref::toBytes` probably would make this expression simpler)

chfast: I'm not a big fun of helpers like `toBytes`, usually C++ std lib should be enough. But the question is: should we keep the full memory buffer (owning_bytes_ref) or copy only the return buffer.

libevm/VMCalls.cpp
```diff
@@ -153,11 +153,20 @@ void VM::caseCall()
...
 	{
-		std::pair<bool, owning_bytes_ref> callResult = m_ext->call(*callParams);
-		callResult.second.copyTo(output);
-		m_returnData.assign(callResult.second.begin(), callResult.second.end());
-
-		m_SPP[0] = callResult.first ? 1 : 0;
+		bool success = false;
+		owning_bytes_ref outputRef;
+		std::tie(success, outputRef) = m_ext->call(*callParams);
+		outputRef.copyTo(output);
```
pirapira: Am I mistaken that `output` is only written to but not read from?  (This didn't change in this commit, but I started wondering now.)

pirapira: Ah, it's the caller's memory!

chfast: Magic. The `output` is actually a memory reference initialized in `caseCallSetup()`. This is exactly CALL output buffer, might be of length 0 depending on CALL arguments. Here we copy the data there according to what the CALL requested.

libevm/VMCalls.cpp
chfast: Anyway, refactored.

chfast:
Big thanks for reviewing this.
Should we wait for tests and finalization of the spec (opcode numbers are to be changes, and decision about copying out of buffer)?

gumb0:
I would better merge now, then fix what changed later

libevm/VM.cpp
```diff
@@ -715,6 +727,20 @@ void VM::interpretCases()
...
+		{
+			if (!m_schedule->haveReturnData)
+				throwBadInstruction();
+
+			m_copyMemSize = toInt63(m_SP[2]);
+			updateMem(memNeed(m_SP[0], m_SP[2]));
+			ON_OP();
+			updateIOGas();
+
+			copyDataToMemory(&m_returnData, m_SP);
```
axic: Not strictly connected to this PR, but probably it would be better to have the offset + size as explicit parameters in `copyDataToMemory`.

libevm/VMCalls.cpp
```diff
@@ -120,6 +120,9 @@ void VM::caseCreate()
 	uint64_t initOff = (uint64_t)m_SP[1];
 	uint64_t initSize = (uint64_t)m_SP[2];
 
+	// Clear the return data buffer. This will not free the memory.
+	m_returnData.clear();
+
```
pirapira: When the contract initialization executes `REVERT`, `m_returnData` has to be set.

pirapira: See the last sentence on this line: https://github.com/ethereum/EIPs/pull/211/files#diff-fd8c393953fe66162d8bb518d87bfb57R32

(That's something I learned while doing YP.)

chfast: Implemented.

libevm/VMCalls.cpp
```diff
@@ -130,7 +130,11 @@ void VM::caseCreate()
 		if (!m_schedule->staticCallDepthLimit())
 			createGas -= createGas / 64;
 		u256 gas = createGas;
-		m_SPP[0] = (u160)m_ext->create(endowment, gas, bytesConstRef(m_mem.data() + initOff, initSize), m_onOp);
+		h160 addr;
+		owning_bytes_ref revertOutput;
+		std::tie(addr, revertOutput) = m_ext->create(endowment, gas, bytesConstRef(m_mem.data() + initOff, initSize), m_onOp);
+		m_SPP[0] = (u160)addr;
+		m_returnData = revertOutput.toBytes();
```
pirapira: When the creation succeeds, does this set the deployed code in the buffer?

chfast: No. Executive handles this. For CREATE, the VM output is only taken in case of REVERT. https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/Executive.cpp#L411.

pirapira: Ah, I see.  I've opened the same file in the editor but I was not quite seeing it.

chfast: That's my role to explain the changes.

libevm/JitVM.cpp
```diff
@@ -149,12 +149,15 @@ int64_t call(
 		u256 gas = _msg->gas;
 		// ExtVM::create takes the sender address from .myAddress.
 		assert(fromEvmC(_msg->sender) == env.myAddress);
-		auto addr = env.create(value, gas, input, {});
+		h160 addr;
+		owning_bytes_ref revertOutput;
+		std::tie(addr, revertOutput) = env.create(value, gas, input, {});
```
gumb0: `revertOutput` is not used after `create`, is this correct?

chfast: At the moment it is. EVM-C does not support "revert output" yet so there is not option to pass this back to EVMJIT. But EVM-C v4 changes are in progress.

libevmcore/Instruction.h
```diff
@@ -47,6 +47,9 @@ enum class Instruction: uint8_t
 	EXP,                ///< exponential operation
 	SIGNEXTEND,         ///< extend length of signed integer
 
+	RETURNDATASIZE = 0x0d,  ///< size of data returned from previous call
+	RETURNDATACOPY = 0x0e,  ///< copy data returned from previous call to memory
```
pirapira: The opcodes seem to have changed: https://github.com/ethereum/solidity/pull/2275#discussion_r116934325

chfast: Not in the EIP PR.

libevm/JitVM.cpp
gumb0: I'd use `std::ignore` while this is the case.

libevm/VM.h
gumb0: Maybe clarify that this is for the data _returned from the subcall_.
Because now we have m_output described as "return bytes" and m_returnedData described as "return data", this is confusing if you don't know what's it all about

libevm/VMCalls.cpp
```diff
@@ -120,14 +120,21 @@ void VM::caseCreate()
...
 	if (m_ext->balance(m_ext->myAddress) >= endowment && m_ext->depth < 1024)
 	{
 		*m_io_gas_p = m_io_gas;
 		u256 createGas = *m_io_gas_p;
 		if (!m_schedule->staticCallDepthLimit())
 			createGas -= createGas / 64;
 		u256 gas = createGas;
-		m_SPP[0] = (u160)m_ext->create(endowment, gas, bytesConstRef(m_mem.data() + initOff, initSize), m_onOp);
+		h160 addr;
+		owning_bytes_ref revertOutput;
```
gumb0: Maybe name it just `output`? Because `revertOutput` is kind of assuming how `create` works internally and that only `REVERT` can return data. Which is the case now, but could change in the furure

gumb0:
Looks good now, just a couple of minor comments.

gumb0:
@yann300 take a look at changes in ` ExtVM::create` here, you'll probably need to rebase on that if we merge this first

chfast:
Ok, I will try to apply suggestions as soon as possible.

chfast:
Ready to land?

pirapira:
@chriseth is changing the EIP.

chfast:
@gumb0 @pirapira @chriseth I'm for merging this as is, and implement the final decision about throwing on out-of-range reads when the decision is made. If you agree, please press the Merge button.

PR #4063, opened 2017-04-25
allow future blocks in tests

winsvega:
metropolis transition test

### DIFF
--- a/libethashseal/genesis/eip158Test.cpp
+++ b/libethashseal/genesis/eip158Test.cpp
@@ -40,7 +40,7 @@ R"E(
 		"registrar" : "0xc6d9d2cd449a754c494264e1809c50e34d64562b",
 		"networkID" : "0x1",
 		"chainID": "0x01",
-		"allowFutureBlocks" : "1"
+		"allowFutureBlocks" : true
 	},
 	"genesis": {
 		"nonce": "0x0000000000000042",
--- a/libethashseal/genesis/eip158Test.cpp
+++ b/libethashseal/genesis/eip158Test.cpp
@@ -39,7 +39,8 @@ R"E(
 		"blockReward": "0x4563918244F40000",
 		"registrar" : "0xc6d9d2cd449a754c494264e1809c50e34d64562b",
 		"networkID" : "0x1",
-		"chainID": "0x01"
+		"chainID": "0x01",
+		"allowFutureBlocks" : "1"
 	},
 	"genesis": {
 		"nonce": "0x0000000000000042",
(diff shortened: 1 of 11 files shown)


chfast:
Can you give any description what this is about?

libethashseal/genesis/eip158Test.cpp
```diff
@@ -39,7 +39,8 @@ R"E(
 		"blockReward": "0x4563918244F40000",
 		"registrar" : "0xc6d9d2cd449a754c494264e1809c50e34d64562b",
 		"networkID" : "0x1",
-		"chainID": "0x01"
+		"chainID": "0x01",
+		"allowFutureBlocks" : "1"
```
chfast: So we have values like `0x1`, `0x01` and `1`. WTF? Why it cannot be `true` / `false`?

winsvega: because there is no specification for json file format.

libethereum/ChainParams.cpp
```diff
@@ -66,8 +66,16 @@ ChainParams ChainParams::loadConfig(string const& _json, h256 const& _stateRoot)
 	cp.tieBreakingGas = params.count("tieBreakingGas") ? params["tieBreakingGas"].get_bool() : true;
 	cp.blockReward = u256(fromBigEndian<u256>(fromHex(params["blockReward"].get_str())));
 	for (auto i: params)
-		if (i.first != "accountStartNonce" && i.first != "maximumExtraDataSize" && i.first != "blockReward" && i.first != "tieBreakingGas")
+	{
+		if (i.first != "accountStartNonce" && i.first != "maximumExtraDataSize" && i.first != "blockReward" && i.first != "tieBreakingGas" && i.first != "allowFutureBlocks")
```
chfast: This line does not fit on my FullHD display. Can you split it please?

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -461,7 +461,6 @@ void overwriteBlockHeaderForTest(mObject const& _blObj, TestBlock& _block, Chain
 			BlockHeader parentHeader = importedBlocks.at(importedBlocks.size() - 1).blockHeader();
 			tmp.setTimestamp(toInt(ho["RelTimestamp"]) + parentHeader.timestamp());
 			tmp.setDifficulty(((const Ethash*)sealEngine)->calculateDifficulty(tmp, parentHeader));
-			this_thread::sleep_for(chrono::seconds((int)toInt(ho["RelTimestamp"])));
```
chfast: Was this the cause of the BC tests slowdown?

winsvega: when filling the tests this line actually waits this amounts of seconds so the forged block wont throw a FutureTime exception.

chfast:
You ware suppose to wait for Windows build!

PR #4064, opened 2017-04-26
Boost shift tests

chfast:


test/unittests/external-dependencies/boost.cpp
```diff
@@ -37,4 +37,139 @@ BOOST_AUTO_TEST_CASE(u256_overflow_test)
...
+	BOOST_CHECK_EQUAL(k >> 253, 0b11);
+	BOOST_CHECK_EQUAL(k >> 254, 0b1);
+	BOOST_CHECK_EQUAL(k >> 255, 0);
+	BOOST_CHECK_EQUAL(k >> 256, 0);
+	BOOST_CHECK_EQUAL(k >> ~uint32_t(0), 0);
+
+	// Division equivalence.
+
+	// Built-in type:
+	volatile auto d = int64_min;
```
gumb0: What is this `volatile` magic for ?

chfast: Nothing important any more. I wanted to check if runtime results are the same as compile-time results so I put `volatile` to prevent the compiler to compute it up front. I'm not sure it works.

chfast:
I will change both. The `utf` is suggested by boost docs, but I agree the name is not clear.

Issue #3612, opened 2017-03-03
EIP96: Block hash and state root in a special contract


pirapira:
Implementing EIP 96 as part of the Metropolis changes.

* EIP 96 https://github.com/ethereum/EIPs/issues/96

chfast:
I think for now we can only implement "the infrastructure" for it, as the bytecode to be deployed is not ready.

PR #4066, opened 2017-05-03
EIP96 - blockhash refactoring

gumb0:
https://github.com/ethereum/cpp-ethereum/issues/3612
https://github.com/ethereum/EIPs/pull/210

- [x] Contract creation
- [x] A call to update stored block hashes before processing transactions in each block
- [x] BLOCKHASH implementation using the call to contract
- [x] Change BLOCKHASH cost after Metropolis


libdevcrypto/Common.cpp
```diff
@@ -67,6 +67,8 @@ const Address dev::ZeroAddress = Address();
 
 const Address dev::MaxAddress = Address("0xffffffffffffffffffffffffffffffffffffffff");
 
+const Address dev::SuperuserAddress = Address("0xfffffffffffffffffffffffffffffffffffffffe");
```
axic: This has been renamed to `system`.

gumb0: Thanks, I'll rename it

libethereum/ExtVM.cpp
```diff
@@ -134,3 +134,22 @@ void ExtVM::suicide(Address _a)
 	m_s.subBalance(myAddress, m_s.balance(myAddress));
 	ExtVMFace::suicide(_a);
 }
+
+h256 ExtVM::blockHash(u256 _number)
+{
+	if (envInfo().number() < m_sealEngine.chainParams().u256Param("metropolisForkBlock") + 256)
```
axic: I think it was agreed last time that the opcode remains limited to the last 256 blocks.

gumb0: So where the code or comment disagrees with this?

axic: Right now the comment disagrees with the code. If you update the code to match the current EIP (limited to 256) then the comment is fine again :)

gumb0: I don't see the disagreement.
This line in particular just checks whether the current block (not the block requested) is before or after Metro+256 (according to this spec https://github.com/ethereum/EIPs/pull/210/files#diff-e02a92c2fb96c1a1bfb05e4c6e2ef5daR34 )

If we are before Metro+256, we return, as comment says, "hash of a block if within the last 256 blocks, or h256() otherwise."
Otherwise we do the same but through the call.

axic: I do not see that it is limited to the last 256 blocks through the call below after Metropolis?

gumb0: Oh, I see, I was assuming that the contract itself (the new version of it) will return 0 if requested block is out of bounds.
Yeah, anyway it will makes sense to avoid making a call in this case, I'll change it.
Thanks for noticing this!

axic: The contract has no easy way knowing where the call came from and its code may be different too in the future. Additionally the spec will/should say that the opcode itself filters for <=256.

gumb0:
>  don't pass last hashes to Executive after Metropolis, saving performance
This should better go into separate PR, because looks like some non-trivial refactoring is needed there

libevmcore/EVMSchedule.h
```diff
@@ -108,6 +109,7 @@ static const EVMSchedule EIP158Schedule = []
 static const EVMSchedule MetropolisSchedule = []
 {
 	EVMSchedule schedule = EIP158Schedule;
+	schedule.blockhashGas = 800;
```
pirapira: It's uncertain to me if `800` should be applied from the beginning of `Metropolis` from 256 blocks into `Metropolis`.  I asked a question https://github.com/ethereum/EIPs/pull/210/files#r117211219

pirapira: And I will ask in the allcoredev meeting https://github.com/ethereum/pm/issues/14#issuecomment-302364003

libethereum/Block.cpp
```diff
@@ -692,6 +693,29 @@ void Block::performIrregularModifications()
...
+	}
+
+	if (blockNumber >= metropolisForkBlock)
+	{
+		u256 const nonce = transactionsFrom(SystemAddress);
+		u256 const gas = 1000000;
+		Transaction tx(0, 0, gas, c_blockhashContractAddress, m_previousBlock.hash().asBytes(), nonce);
+		tx.forceSender(SystemAddress);
+		// passing empty lastHashes - assuming the contract doesn't use BLOCKHASH
+		m_state.execute(EnvInfo(info(), {}, 0), *m_sealEngine, tx, Permanence::Committed);
```
pirapira: I guess this doesn't increment the nonce.

libethereum/ExtVM.cpp
```diff
@@ -134,3 +134,26 @@ void ExtVM::suicide(Address _a)
...
+h256 ExtVM::blockHash(u256 _number)
+{
+	u256 const currentNumber = envInfo().number();
+
+	if (_number >= currentNumber || _number < (std::max<u256>(256, currentNumber) - 256))
+		return h256();
+
+	if (currentNumber < m_sealEngine.chainParams().u256Param("metropolisForkBlock") + 256)
+	{
+		assert(envInfo().lastHashes().size() > (unsigned)(currentNumber - 1 - _number));
```
pirapira: The assertion seems to fail if `currentNumber == 0` and `_number` == `0xff..ff` (which is `0 - 1`).  I think `lastHashes().size()` should be zero in that case, and the right hand side evaluates to zero.

libethereum/Block.cpp
gumb0: Hm, I think it does. Shouldn't it?

pirapira: I don't know if it should or not.  So I need to ask.

libethereum/ExtVM.cpp
gumb0: We won't get here because of the check at line 142, requested `_number` is always less than `currentNumber `here

gumb0: Also `currentNumber` is never 0 here, because you can't execute anything in the genesis block

libethereum/Block.cpp
pirapira: I asked https://github.com/ethereum/EIPs/pull/210/files#r117260179 https://github.com/ethereum/pm/issues/14#issuecomment-302417899

libethereum/ExtVM.cpp
pirapira: OK, seems valid.

gumb0:
@pirapira Thanks for the review! I need to wait until the tests pass here anyway

pirapira:
Please rebase.

libethereum/Block.cpp
```diff
@@ -692,6 +693,29 @@ void Block::performIrregularModifications()
...
+		m_state.createContract(c_blockhashContractAddress);
+		m_state.setNewCode(c_blockhashContractAddress, bytes(c_blockhashContractCode));
+		m_state.commit(State::CommitBehaviour::KeepEmptyAccounts);
+	}
+
+	if (blockNumber >= metropolisForkBlock)
+	{
+		u256 const nonce = transactionsFrom(SystemAddress);
+		u256 const gas = 1000000;
+		Transaction tx(0, 0, gas, c_blockhashContractAddress, m_previousBlock.hash().asBytes(), nonce);
```
chfast: Again out crappy vector_ref interface, but I think you can go from h256 to bytes_ref without going through std::vector. It might look like `{hash.begin(), hash.size()}`. Or improve vector_ref.

libevmcore/EVMSchedule.h
chfast: Yes, this seem to be strange. Shouldn't we just charge for the contract execution?

test/tools/jsontests/vm.cpp
```diff
@@ -53,6 +53,13 @@ std::pair<bool, eth::owning_bytes_ref> FakeExtVM::call(CallParameters& _p)
 	return {true, eth::owning_bytes_ref{}};  // Return empty output.
 }
 
+h256 FakeExtVM::blockHash(u256 _number)
+{
+	return _number < envInfo().number() && _number >= (std::max<u256>(256, envInfo().number()) - 256) ?
```
chfast: I would go with `if else`, not `? :` :)

libethereum/Block.cpp
gumb0: `Transaction` constructor takes `bytes`, not `bytes_ref`. Do you want to change that?

chfast: No, that's fine then.

gumb0:
Rebased and addressed comments.

libethashseal/genesis/metropolisTransitionTest.cpp
```diff
@@ -0,0 +1,63 @@
...
+		"metropolisForkBlock": "0x02",
+		"minGasLimit": "0x1388",
+		"maxGasLimit": "7fffffffffffffff",
+		"tieBreakingGas": false,
+		"gasLimitBoundDivisor": "0x0400",
+		"minimumDifficulty": "0x020000",
+		"difficultyBoundDivisor": "0x0800",
+		"durationLimit": "0x0d",
+		"blockReward": "0x4563918244F40000",
+		"registrar" : "0xc6d9d2cd449a754c494264e1809c50e34d64562b",
```
chriseth: Can we remove this from the codebase?

libethcore/Common.cpp
```diff
@@ -53,6 +53,9 @@ const unsigned c_databaseVersionModifier = 0;
 
 const unsigned c_databaseVersion = c_databaseBaseVersion + (c_databaseVersionModifier << 8) + (23 << 9);
 
+const Address c_blockhashContractAddress(0xf0);
+const bytes c_blockhashContractCode(fromHex("0x73fffffffffffffffffffffffffffffffffffffffe33141561006a5760014303600035610100820755610100810715156100455760003561010061010083050761010001555b6201000081071515610064576000356101006201000083050761020001555b5061013e565b4360003512151561008457600060405260206040f361013d565b61010060003543031315156100a857610100600035075460605260206060f361013c565b6101006000350715156100c55762010000600035430313156100c8565b60005b156100ea576101006101006000350507610100015460805260206080f361013b565b620100006000350715156101095763010000006000354303131561010c565b60005b1561012f57610100620100006000350507610200015460a052602060a0f361013a565b600060c052602060c0f35b5b5b5b5b"));
```
chriseth: Can we use an assert to check the hash of this code against other implementations? Or will the code hash be part of the block tests anyway?

chriseth: Might there be any global variable initialization race condition problems here?

libethereum/Block.cpp
gumb0: Performing the call to update the blockhash contract should be similar to CALL from EVM, not to external transaction and should not increment SYSTEM_ADDRESS's nonce.

So the fix is needed here.

pirapira:
The Travis test failure looks related to the blockhash https://travis-ci.org/ethereum/cpp-ethereum/jobs/237208475#L1908

libethashseal/genesis/metropolisTransitionTest.cpp
gumb0: I'll remove it from this file in this PR and from other files in separate PR

libethcore/Common.cpp
gumb0: 1. Yeah, blockchain tests check that the contract with correct code exists in the state

2. There shouldn't be as long as we don't use these constants for other static variable initialization in another module. Which we shouldn't do.

libethashseal/genesis/metropolisTransitionTest.cpp
gumb0: I'll remove it in a separate PR

test/tools/jsontests/StateTests.cpp
```diff
@@ -115,25 +116,49 @@ class generaltestfixture
...
 
-		boost::filesystem::directory_iterator iterator(fillersPath);
-		for(; iterator != boost::filesystem::directory_iterator(); ++iterator)
+		dirIt iterator(fillersPath);
+		for(; iterator != dirIt(); ++iterator)
 			if (boost::filesystem::is_regular_file(iterator->path()) && iterator->path().extension() == ".json")
 			{
 				string fileboost = iterator->path().filename().string();
-				dev::test::executeTests(fileboost, "/GeneralStateTests/"+_folder, "/GeneralStateTestsFiller/"+_folder, dev::test::doStateTests);
+				if (fileboost.rfind("BCFiller.json") != std::string::npos)
```
gumb0: @winsvega could you explain what happens here and why do you need to hardcode json file name?

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -102,16 +101,28 @@ void checkBlocks(TestBlock const& _blockFromFields, TestBlock const& _blockFromR
 
 void doBlockchainTests(json_spirit::mValue& _v, bool _fillin)
 {
-	if (!Options::get().fillchain) //fill blockchain through state tests
+	bool isRunFromStateTests = true;
+	string suitename = boost::unit_test::framework::get<boost::unit_test::test_suite>(boost::unit_test::framework::current_test_case().p_parent_id).p_name;
+	if (suitename.find("StateTests") == string::npos)
```
gumb0: Is there any way to pass the flag/enum to this function telling that we're in a special case of StateTests instead of querying suite name from boost? Why is `doBlockchainTests` called from StateTests suite anyway?

test/tools/jsontests/StateTests.cpp
winsvega: so not to convert 100 of tests into blockchain we decided to add a marker to such tests to be run as blockchain but being state tests. 
I had an option in mind - to read file contents and then look for an option to run this test as blockchain.
then I realized that to do that I will have to parse whole json file and that would add to execution time. 
there for we have this hardcoded thing. 

the mere fact that state tests are being run as blockchain is a workaround. it was not designed to be run as blockchain tests.

test/tools/jsontests/StateTests.cpp
```diff
@@ -115,25 +116,49 @@ class generaltestfixture
...
+				{
+					//Skip filling this tests as state tests
+					if (Options::get().filltests && !Options::get().fillchain)
+					{
+						TestOutputHelper::incTestCount(2);
```
winsvega: as state test was not designed to run as a blockchain test. I have to overrun the test counter with this to get 100% tests executed output.

test/tools/jsontests/BlockChainTests.cpp
winsvega: read above

test/tools/jsontests/StateTests.cpp
pirapira: @winsvega can you keep a list of these names in a file in the `tests` repository?

gumb0:
After discussion with @winsvega he agreed to try a different approach with fixing tests for this - rolling back current workaround changes in testeth code and just convert affected json files

libevmcore/EVMSchedule.h
pirapira: This can stay as it is.  In this previous allcoredev meeting, it was agreed that the gas schedule changes at the Metropolis block (inclusive), not later.

gumb0:
@chfast Don't spend your time on that yet, @winsvega plans now a different approach

test/tools/jsontests/vm.cpp
```diff
@@ -53,6 +53,14 @@ std::pair<bool, eth::owning_bytes_ref> FakeExtVM::call(CallParameters& _p)
 	return {true, eth::owning_bytes_ref{}};  // Return empty output.
 }
 
+h256 FakeExtVM::blockHash(u256 _number)
+{
+	if (_number < envInfo().number() && _number >= (std::max<u256>(256, envInfo().number()) - 256))
+		return sha3(toString(_number));
+	else
```
chfast: There should be no `else`.

PR #4074, opened 2017-05-11
State diff

winsvega:
add --statediff option when filling or running tests

### DIFF
--- a/libethereum/State.cpp
+++ b/libethereum/State.cpp
@@ -550,7 +550,7 @@ std::pair<ExecutionResult, TransactionReceipt> State::execute(EnvInfo const& _en
 			removeEmptyAccounts = _envInfo.number() >= _sealEngine.chainParams().u256Param("EIP158ForkBlock");
 			commit(removeEmptyAccounts ? State::CommitBehaviour::RemoveEmptyAccounts : State::CommitBehaviour::KeepEmptyAccounts);
 			break;
-		case Permanence::Uncommited:
+		case Permanence::Uncommitted:
 			break;
 	}
 
--- a/libethereum/State.cpp
+++ b/libethereum/State.cpp
@@ -41,7 +41,6 @@
 using namespace std;
 using namespace dev;
 using namespace dev::eth;
-using namespace dev::eth::detail;
 namespace fs = boost::filesystem;
 
 const char* StateSafeExceptions::name() { return EthViolet "⚙" EthBlue " ℹ"; }
@@ -541,12 +540,18 @@ std::pair<ExecutionResult, TransactionReceipt> State::execute(EnvInfo const& _en
 	u256 const startGasUsed = _envInfo.gasUsed();
 	executeTransaction(e, _t, onOp);
 
-	if (_p == Permanence::Reverted)
-		m_cache.clear();
-	else
+	bool removeEmptyAccounts = false;
+	switch (_p)
 	{
-		bool removeEmptyAccounts = _envInfo.number() >= _sealEngine.chainParams().u256Param("EIP158ForkBlock");
-		commit(removeEmptyAccounts ? State::CommitBehaviour::RemoveEmptyAccounts : State::CommitBehaviour::KeepEmptyAccounts);
+		case Permanence::Reverted:
+			m_cache.clear();
+			break;
+		case Permanence::Committed:
+			removeEmptyAccounts = _envInfo.number() >= _sealEngine.chainParams().u256Param("EIP158ForkBlock");
+			commit(removeEmptyAccounts ? State::CommitBehaviour::RemoveEmptyAccounts : State::CommitBehaviour::KeepEmptyAccounts);
+			break;
+		case Permanence::Uncommited:
+			break;
 	}
 
 	TransactionReceipt const receipt = _envInfo.number() >= _sealEngine.chainParams().u256Param("metropolisForkBlock") ?
--- a/libethereum/State.h
+++ b/libethereum/State.h
@@ -83,7 +83,7 @@ enum class Permanence
 {
 	Reverted,
 	Committed,
-	Uncommited	///< Uncommited state for change log readings in tests.
+	Uncommitted	///< Uncommitted state for change log readings in tests.
 };
 
 #if ETH_FATDB
--- a/libethereum/State.h
+++ b/libethereum/State.h
@@ -82,7 +82,8 @@ enum class BaseState
 enum class Permanence
 {
 	Reverted,
-	Committed
+	Committed,
+	Uncommited	///< Uncommited state for change log readings in tests.
 };
 
 #if ETH_FATDB
@@ -96,9 +97,6 @@ DEV_SIMPLE_EXCEPTION(IncorrectAccountStartNonceInState);
 
 class SealEngineFace;
 class Executive;
-
-namespace detail
-{
 
 /// An atomic state changelog entry.
 struct Change
@@ -142,8 +140,7 @@ struct Change
 	{}
 };
 
-}
-
+using ChangeLog = std::vector<Change>;
 
 /**
  * Model of an Ethereum state, essentially a facade for the trie.
@@ -309,6 +306,8 @@ public:
 	/// Revert all recent changes up to the given @p _savepoint savepoint.
 	void rollback(size_t _savepoint);
 
+	ChangeLog const& changeLog() const { return m_changeLog; }
+
 private:
 	/// Turns all "touched" empty accounts into non-alive accounts.
 	void removeEmptyAccounts();
@@ -336,7 +335,7 @@ private:
 	u256 m_accountStartNonce;
 
 	friend std::ostream& operator<<(std::ostream& _out, State const& _s);
-	std::vector<detail::Change> m_changeLog;
+	ChangeLog m_changeLog;
 };
 
 std::ostream& operator<<(std::ostream& _out, State const& _s);
--- a/test/tools/libtesteth/FillJsonFunctions.cpp
+++ b/test/tools/libtesteth/FillJsonFunctions.cpp
@@ -1,0 +1,218 @@
+/*
+	This file is part of cpp-ethereum.
+
+	cpp-ethereum is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/** @file
+ * Helper functions to work with json::spirit and test files
+ */
+#include <libethereum/Transaction.h>
+#include <libdevcore/CommonData.h>
+
+#include <test/tools/libtesteth/TestHelper.h>
+#include <test/tools/libtesteth/TestOutputHelper.h>
+#include <test/tools/libtesteth/Options.h>
+#include <test/tools/libtesteth/ImportTest.h>
+
+using namespace std;
+using namespace dev;
+using namespace dev::eth;
+using namespace dev::test;
+
+namespace dev
+{
+namespace test
+{
+
+
+json_spirit::mObject fillJsonWithTransaction(Transaction const& _txn)
+{
+	json_spirit::mObject txObject;
+	txObject["nonce"] = toCompactHex(_txn.nonce(), HexPrefix::Add, 1);
+	txObject["data"] = _txn.data().size() ? toHex(_txn.data(), 2, HexPrefix::Add) : "";
+	txObject["gasLimit"] = toCompactHex(_txn.gas(), HexPrefix::Add, 1);
+	txObject["gasPrice"] = toCompactHex(_txn.gasPrice(), HexPrefix::Add, 1);
+	txObject["r"] = toCompactHex(_txn.signature().r, HexPrefix::Add, 1);
+	txObject["s"] = toCompactHex(_txn.signature().s, HexPrefix::Add, 1);
+	txObject["v"] = toCompactHex(_txn.signature().v + 27, HexPrefix::Add, 1);
+	txObject["to"] = _txn.isCreation() ? "" : "0x" + toString(_txn.receiveAddress());
+	txObject["value"] = toCompactHex(_txn.value(), HexPrefix::Add, 1);
+	ImportTest::makeAllFieldsHex(txObject);
+	return txObject;
+}
+
+json_spirit::mObject fillJsonWithStateChange(State const& _stateOrig, eth::State const& _statePost, std::vector<dev::eth::detail::Change> const& _changeLog)
+{
+	json_spirit::mObject oState;
+	if (!_changeLog.size())
+		return oState;
+
+	using changeKind = eth::detail::Change::Kind;
+	using changeStruct = eth::detail::Change;
+	eth::detail::ChangeLog changeLog = _changeLog;
+
+	//Sort the vector by address field
+	std::sort(changeLog.begin(), changeLog.end(),
+		[ ]( const changeStruct& lhs, const changeStruct& rhs )
+		{
+			return lhs.address < rhs.address;
+		}
+	);
+
+	std::ostringstream log;
+	json_spirit::mObject o;
+
+	size_t i = 0;
+	while (i < changeLog.size())
+	{
+		json_spirit::mArray agregatedBalance;
+		json_spirit::mArray agregatedStorage;
+		std::map<string, ostringstream> logInfo;
+		Address currentAddress = changeLog.at(i).address;
+		u256 tmpBalance = _stateOrig.balance(currentAddress);
+
+		while (i < changeLog.size() && changeLog[i].address == currentAddress)
+		{
+			//Agregate changes for the same addresses
+			string after;
+			string before;
+			json_spirit::mArray record;
+			changeStruct change = changeLog.at(i);
+			switch (change.kind)
+			{
+				case changeKind::NewCode:
+					//take the original and final code only
+					before = toHex(_stateOrig.code(change.address), 2, HexPrefix::Add);
+					after = toHex(_statePost.code(change.address), 2, HexPrefix::Add);
+					record.push_back(before);
+					record.push_back("->");
+					record.push_back(after);
+					o["code"] = record;
+					logInfo["code"] << "'code' : ['" + before + "', '->', '" + after + "']" << std::endl;
+				break;
+				case changeKind::Nonce:
+					//take the original and final nonce only
+					before = toCompactHex(_stateOrig.getNonce(change.address), HexPrefix::Add, 1);
+					after = toCompactHex(_statePost.getNonce(change.address), HexPrefix::Add, 1);
+					record.push_back(before);
+					record.push_back("->");
+					record.push_back(after);
+					o["nonce"] = record;
+					logInfo["nonce"] << "'nonce' : ['" + before + "', '->', '" + after + "']" << std::endl;
+				break;
+				case changeKind::Balance:
+					before = toCompactHex(tmpBalance, HexPrefix::Add, 1);
+					after = toCompactHex(change.value, HexPrefix::Add, 1);
+					record.push_back(before);
+					record.push_back("+=");
+					record.push_back(after);
+					tmpBalance += change.value;
+					agregatedBalance.push_back(record);
+					logInfo["balance"] << "'balance' : ['" + before + "', '+=', '" + after + "']" << std::endl;
+				break;
+				case changeKind::Touch:
+					o["hasBeenTouched"] = "true";
+					logInfo["hasBeenTouched"] << "'hasBeenTouched' : ['true']" << std::endl;
+				break;
+				case changeKind::Storage:
+					//take the original and final storage only
+					before = toCompactHex(change.key, HexPrefix::Add, 1) + " : " + toCompactHex(_stateOrig.storage(change.address, change.key), HexPrefix::Add, 1);
+					after = toCompactHex(change.key, HexPrefix::Add, 1) + " : " + toCompactHex(_statePost.storage(change.address, change.key), HexPrefix::Add, 1);
+					record.push_back(before);
+					record.push_back("->");
+					record.push_back(after);
+					agregatedStorage.push_back(record);
+					o["storage"] = agregatedStorage;
+					logInfo["storage"] << "'storage' : ['" + before + "', '->', '" + after + "']" << std::endl;
+				break;
+				case changeKind::Create:
+					o["newlyCreated"] = "true";
+					logInfo["newlyCreated"] << "'newlyCreated' : ['true']" << std::endl;
+				break;
+				default:
+					o["unknownChange"] = "true";
+					logInfo["unknownChange"] << "'unknownChange' : ['true']" << std::endl;
+				break;
+			}
+			++i;
+		}
+
+		//summaraize changes
+		if (agregatedBalance.size())
+		{
+			json_spirit::mArray record;
+			string before = toCompactHex(_stateOrig.balance(currentAddress), HexPrefix::Add, 1);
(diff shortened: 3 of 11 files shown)


libethereum/State.cpp
```diff
@@ -541,6 +541,7 @@ std::pair<ExecutionResult, TransactionReceipt> State::execute(EnvInfo const& _en
 	u256 const startGasUsed = _envInfo.gasUsed();
 	executeTransaction(e, _t, onOp);
 
+	res.changeLog = m_changeLog;
```
gumb0: I don'like that this way we will have this in `ExecutionResult` (paying with additional memory consumption) even when we don't need it (we need it only for tests and maybe vmtraces)

@chfast what do you think?

winsvega: could be under cmake flag --VMTRACE

libdevcore/CommonData.h
```diff
@@ -172,9 +179,9 @@ inline std::string toHex(u256 val, HexPrefix prefix = HexPrefix::DontAdd)
 	return (prefix == HexPrefix::Add) ? "0x" + str : str;
 }
 
-inline std::string toCompactHex(u256 val, HexPrefix prefix = HexPrefix::DontAdd, unsigned _min = 0)
+inline std::string toCompactHex(u256 val, HexPrefix prefix = HexPrefix::DontAdd, PrintZero _min = PrintZero::DontPrint)
```
gumb0: So many places where we pass `PrintZero::Print` here, is it ever used really with `PrintZero::DontPrint` ?

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -629,6 +654,10 @@ int ImportTest::exportTest(bytes const& _output)
 				if (it != std::end(stateIndexesToPrint))
 					obj2["postState"] = fillJsonWithState(m_transactions[i].postState);
 			}
+
+			if (Options::get().statediff)
```
gumb0: Does this go into generated json files?

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -124,19 +125,126 @@ eth::Network stringToNetId(string const& _netname)
...
 
+json_spirit::mObject fillJsonWithStateChange(State const& _stateOrig, eth::State const& _statePost, std::vector<dev::eth::detail::Change> const& _changeLog)
+{
+	json_spirit::mObject oState;
+	typedef dev::eth::detail::Change::Kind changeKind;
+	typedef std::vector<dev::eth::detail::Change> logVector;
+	std::vector<dev::eth::detail::Change> changeLog = _changeLog;
+
+	std::ostringstream log;
+	while(changeLog.size())
```
gumb0: Why do you need to iterate in this weird way instead of just `for (auto& change : changeLog)` ?

gumb0:
It probably would be nice to have this feature not only for tests, but also for regular `eth` operation, like we have vmtrace for both... Would it be difficult to output state diff to log in `eth` when, say, VMTRACE is enabled ?

libethereum/State.cpp
chfast: I don't like it either. This must go to `dev::test` namespace unless proven it is not possible to be done this way.

And big no to build flags. Let's find a way to do it nicely without affecting not-test related code.

libdevcore/CommonData.h
chfast: This looks bad. Can you describe what the use cases for this function are? If tests needs something different maybe it will be better for `dev::test` to have a separated function for this case.

chfast:
Ok, let's take is slowly.
Can we separate the changes related to `toCompactHex` from the "state diff" one first?

libethereum/State.cpp
gumb0: I had in mind adding another callback function for collecting state diffs as `State::execute` parameter.

chfast: Is the "state diff" the same as "state journal" or "state changelog"? If yes, let's stick to a single name. We can add a function to export the changelog next to the functions like `savepoint()` or `revert()`.

gumb0: I think it's essentially the same, let's call it change log.

We need it after transaction execution (`executeTransaction` call on line 542), but before `commit` which clears changelog.

The getter function would work if we save the copy before commit, or don't clear it.

winsvega: yes, getter function would be nice. 
state.executeTransaction()
state.getChangelog()

libdevcore/CommonData.h
winsvega: there is a person who constantly asking me to change hex format in our tests. I've changed it here to PrintZero::DontPrint because that is what the meaning of that "1" was in the code. 
but the changes are not final.

test/tools/libtesteth/ImportTest.cpp
winsvega: when using with --filltests yes.

test/tools/libtesteth/TestHelper.cpp
winsvega: I use this because of the changeLog.erase

test/tools/libtesteth/ImportTest.cpp
gumb0: Do we really need to write it into files and incerase their size and complexity? It's only for human readers, right?

Maybe just leave it like a log option for `testeth` run ?

test/tools/libtesteth/TestHelper.cpp
gumb0: Why do you need to erase while iterating over it?

gumb0: If you need group changes by the addresses, you could sort `changeLog` vector by `address` field, then do all the processing in one iteration over the vector

winsvega:
refactored the log function. 
I've removed changeLog clean function from commit in state. and moved it to the begining  of execute.
will see how tests work. if there is an issue we could dublicate m_changeLog variable in State just for  getChangeLog() method

test/tools/libtesteth/FillJsonFunctions.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+	std::ostringstream log;
+	json_spirit::mObject o;
+
+	json_spirit::mArray agregatedBalance;
+	json_spirit::mArray agregatedStorage;
+	changeStruct popChange = changeLog.at(0);
+	std::map<string, ostringstream> logInfo;
+	u256 tmpBalance = _stateOrig.balance(popChange.address);
+	for (size_t i = 0; i < changeLog.size(); i++)
```
gumb0: This loop probably works correctly, but I think expressing it something like this would be clearer:
```
size_t i = 0;
while (i < changeLog.size())
{
	Addres currentAddress = changeLog[i].address;

	string before;
	string after;
	json_spirit::mArray record;
	json_spirit::mArray agregatedBalance;
	json_spirit::mArray agregatedStorage;
	std::map<string, ostringstream> logInfo;

	while (i < changeLog.size() && changeLog[i].address == currentAddress)
	{
		// handle changeLog[i]

		++i;
	}
}
```

pirapira:
I think ChangLog.md can talk about the new option.

test/tools/libtesteth/FillJsonFunctions.cpp
```diff
@@ -0,0 +1,228 @@
...
+	{
+		if (_map.size() && _map.find(a.first) == _map.end())
+			continue;
+
+		json_spirit::mObject o;
+		if (mapEmpty || _map.at(a.first).hasBalance())
+			o["balance"] = toCompactHex(_state.balance(a.first), HexPrefix::Add, 1);
+		if (mapEmpty || _map.at(a.first).hasNonce())
+			o["nonce"] = toCompactHex(_state.getNonce(a.first), HexPrefix::Add, 1);
+		{
```
pirapira: Why is a `{` here?  The `if` statement has ended in the last line.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -637,6 +667,8 @@ int ImportTest::exportTest(bytes const& _output)
 			obj[it->first] = it->second;
 
 		m_testObject["post"] = obj;
+
+
```
pirapira: Spurious lines?

libethereum/StateDetail.h
```diff
@@ -0,0 +1,75 @@
...
+
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/** @file StateDetail.h
```
chfast: Why did you move it to another file?

winsvega:
the first solution was to put changeLog into execution results. that required separation of StateDetail from State

libethereum/StateDetail.h
chfast: You have to move it back.

test/tools/libtesteth/FillJsonFunctions.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/** @file
```
chfast: What is the point of not having these function in TestHelper.cpp?

winsvega:
to split the large file with lots of functions into smaller ones

chfast:
This just makes it compile twice as long as before.

winsvega:
I stick to that guy's opinion:
"f I find that an implementation file is thousands of lines long, that's usually a sign that there's too much there and I need to break it up."
https://stackoverflow.com/questions/531133/should-i-put-many-functions-into-one-file-or-more-or-less-one-function-per-fi

test/tools/libtesteth/FillJsonFunctions.cpp
chfast: @winsvega ?

libethereum/State.cpp
```diff
@@ -224,7 +225,6 @@ void State::commit(CommitBehaviour _commitBehaviour)
 	if (_commitBehaviour == CommitBehaviour::RemoveEmptyAccounts)
 		removeEmptyAccounts();
 	m_touched += dev::eth::commit(m_cache, m_state);
-	m_changeLog.clear();
```
chfast: `commit()` have to clear the changelog. See comments below.

test/tools/libtesteth/ImportTest.cpp
chfast: Answer for this?

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -477,7 +422,10 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
 	}
 	try
 	{
-		cnote << "TEST " << name << ":";
+		if ((Options::get().singleTest && Options::get().singleTestName == name)
+			||	!Options::get().singleTest)
```
chfast: WTF is this tab in the middle?

winsvega:
staediff also work when filling tests producing output to the test .json for debug

libethereum/State.cpp
```diff
@@ -223,9 +224,12 @@ void State::commit(CommitBehaviour _commitBehaviour)
 {
 	if (_commitBehaviour == CommitBehaviour::RemoveEmptyAccounts)
 		removeEmptyAccounts();
+
 	m_touched += dev::eth::commit(m_cache, m_state);
-	m_changeLog.clear();
+	if (m_changeLog.size() != 0)
+		m_changeLogResult = m_changeLog;
```
chfast: This is bad and inefficient (both fixable), but read below.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -93,6 +93,7 @@ bytes ImportTest::executeTest()
 				eth::Network network = networks[j];
 				std::pair<eth::State, ImportTest::execOutput> out = executeTransaction(network, m_envInfo, m_statePre, m_transactions[i].transaction);
```
chfast: You should try to improve `executeTransaction()` function first.
1. Do you need preState after this function. If not do not copy the State.
2. Can it also return changelog if you need it?
3. Use `std::tie()` to unpack tuple.

libethereum/State.cpp
```diff
@@ -223,9 +223,10 @@ void State::commit(CommitBehaviour _commitBehaviour)
 {
 	if (_commitBehaviour == CommitBehaviour::RemoveEmptyAccounts)
 		removeEmptyAccounts();
+
 	m_touched += dev::eth::commit(m_cache, m_state);
-	m_changeLog.clear();
 	m_cache.clear();
+	m_changeLog.clear();
```
chfast: Revert all changes here.

libethcore/ChainOperationParams.h
```diff
@@ -15,8 +15,6 @@
 	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
 */
 /** @file ChainOperationsParams.h
- * @author Gav Wood <i@gavwood.com>
```
chfast: Do not do this unrelated changes.

libethereum/State.cpp
```diff
@@ -541,12 +542,18 @@ std::pair<ExecutionResult, TransactionReceipt> State::execute(EnvInfo const& _en
...
-		m_cache.clear();
-	else
+	bool removeEmptyAccounts;
+	switch (_p)
 	{
-		bool removeEmptyAccounts = _envInfo.number() >= _sealEngine.chainParams().u256Param("EIP158ForkBlock");
-		commit(removeEmptyAccounts ? State::CommitBehaviour::RemoveEmptyAccounts : State::CommitBehaviour::KeepEmptyAccounts);
+		case Permanence::Reverted:
+			m_cache.clear();
+		break;
```
chfast: `break` must be indented.

libethereum/State.cpp
```diff
@@ -541,12 +542,18 @@ std::pair<ExecutionResult, TransactionReceipt> State::execute(EnvInfo const& _en
...
+		case Permanence::Committed:
+			removeEmptyAccounts = _envInfo.number() >= _sealEngine.chainParams().u256Param("EIP158ForkBlock");
+			commit(removeEmptyAccounts ? State::CommitBehaviour::RemoveEmptyAccounts : State::CommitBehaviour::KeepEmptyAccounts);
+		break;
+		default:
```
chfast: Use `Unfinished` instead of `default`. Compilers will be satisfied by having cases for all options.

libethereum/State.h
```diff
@@ -15,8 +15,6 @@
 	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
 */
 /** @file State.h
- * @author Gav Wood <i@gavwood.com>
```
chfast: Unrelated change.

libethereum/State.h
```diff
@@ -82,7 +80,8 @@ enum class BaseState
 enum class Permanence
 {
 	Reverted,
-	Committed
+	Committed,
+	Unfinished
```
chfast: Documentation missing.

libethereum/State.h
```diff
@@ -142,8 +141,7 @@ struct Change
 	{}
 };
 
-}
-
+} //detail
```
chfast: Ugly formatting. Put 2 spaces before `//`, and space after `//`. Also better comment: `// namespace detail`.

libethereum/State.h
```diff
@@ -336,7 +336,7 @@ class State
 	u256 m_accountStartNonce;
 
 	friend std::ostream& operator<<(std::ostream& _out, State const& _s);
-	std::vector<detail::Change> m_changeLog;
+	std::vector<detail::Change> m_changeLog;	///< Working changeLog
```
chfast: 1. Working means the one that is not broken. You mean current, or change log of currently executing transaction?
2. "changeLog" is not an English word. You mean "change log"?
3. Missing full stop.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -52,13 +53,14 @@ ImportTest::ImportTest(json_spirit::mObject& _o, bool isFiller, testType testTem
 
 bytes ImportTest::executeTest()
 {
+	using results = std::tuple<eth::State, ImportTest::ExecOutput, ImportTest::ChangeLog>;
```
chfast: Type names must start with uppercase letter. Should be `Results`.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -52,13 +53,14 @@ ImportTest::ImportTest(json_spirit::mObject& _o, bool isFiller, testType testTem
...
 	if (m_testType == testType::StateTests)
```
chfast: Is this case needed any more?

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -91,8 +93,9 @@ bytes ImportTest::executeTest()
 					continue;
 
 				eth::Network network = networks[j];
-				std::pair<eth::State, ImportTest::execOutput> out = executeTransaction(network, m_envInfo, m_statePre, m_transactions[i].transaction);
-				m_transactions[i].postState = out.first;
+				results out = executeTransaction(network, m_envInfo, m_statePre, m_transactions[i].transaction);
```
chfast: Use `std::tie()` to assign `postState` and `changeLog` and ignore tuple element of index 1.

test/tools/libtesteth/ImportTest.h
```diff
@@ -54,15 +54,17 @@ class ImportTest
...
 
 	eth::State m_statePre;
 	eth::State m_statePost;
 	eth::LogEntries m_logs;
 	eth::LogEntries m_logsExpected;
 
 private:
-	typedef std::pair<eth::ExecutionResult, eth::TransactionReceipt> execOutput;
-	std::pair<eth::State, execOutput> executeTransaction(eth::Network const _sealEngineNetwork, eth::EnvInfo const& _env, eth::State const& _preState, eth::Transaction const& _tr);
+	typedef std::vector<dev::eth::detail::Change> ChangeLog;
```
chfast: This must go to State.h. Also use it in State class declaration.

test/tools/libtesteth/ImportTest.h
```diff
@@ -54,15 +54,17 @@ class ImportTest
...
+	typedef std::pair<eth::ExecutionResult, eth::TransactionReceipt> ExecOutput;
```
chfast: Use `using` instead of `typedef` to make a type alias.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -477,7 +422,9 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
 	}
 	try
 	{
-		cnote << "TEST " << name << ":";
+		if ((Options::get().singleTest && Options::get().singleTestName == name) || !Options::get().singleTest)
+			cnote << "TEST " << name << ":";
```
chfast: Verbosity 2 outputs test names. Why do you need this?

winsvega: so that when you use --statediff you would see only the name of the test you are running with --singletest option. 
other names make it hard to see the changeLog in console when there are many tests

libethereum/State.cpp
```diff
@@ -541,12 +541,18 @@ std::pair<ExecutionResult, TransactionReceipt> State::execute(EnvInfo const& _en
 	u256 const startGasUsed = _envInfo.gasUsed();
 	executeTransaction(e, _t, onOp);
 
-	if (_p == Permanence::Reverted)
-		m_cache.clear();
-	else
+	bool removeEmptyAccounts;
```
chfast: This variable must be initialized.

libethereum/State.h
```diff
@@ -82,7 +82,8 @@ enum class BaseState
 enum class Permanence
 {
 	Reverted,
-	Committed
+	Committed,
+	Unfinished	///< uncommited state for changeLog readings
```
chfast: In documenting comments always use full sentences. Must be `Uncommited state for change log readings in tests.".
Seeing the comment, shouldn't we change `Unfinished` to `Uncommited`?

libethereum/State.h
```diff
@@ -142,8 +143,9 @@ struct Change
 	{}
 };
 
-}
+typedef std::vector<Change> ChangeLog;
```
chfast: 1. `using ChangeLog = std::vector<Change>;
2. It must not be in `detail` namespace. Put it in `eth` namespace directly. Now the `detail` namespace is pointless as we now make the change log available outside of State.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -55,10 +56,13 @@ bytes ImportTest::executeTest()
...
 	{
 		eth::Network network = eth::Network::MainNetwork;
-		std::pair<eth::State, ImportTest::execOutput> out = executeTransaction(network, m_envInfo, m_statePre, m_transaction);
-		m_statePost = out.first;
-		m_logs = out.second.second.log();
-		return out.second.first.output;
+		LogEntries emptyLogs;
+		ExecutionResult emptyRes;
+		ImportTest::ExecOutput execOut = make_pair(emptyRes, TransactionReceipt(h256(), u256(), emptyLogs));
+		std::tie(m_statePost, execOut, std::ignore) =
```
chfast: Here this change was not necessary as ExecutionResult does not have default constructor, but it can stay.

winsvega: we will remove this part anyway at some point

libethereum/State.h
```diff
@@ -82,7 +82,8 @@ enum class BaseState
 enum class Permanence
 {
 	Reverted,
-	Committed
+	Committed,
+	Uncommited	///< Uncommited state for change log readings in tests.
```
chfast: Typo, `Uncommitted`.

test/tools/libtesteth/FillJsonFunctions.cpp
```diff
@@ -0,0 +1,218 @@
...
+	return txObject;
+}
+
+json_spirit::mObject fillJsonWithStateChange(State const& _stateOrig, eth::State const& _statePost, std::vector<dev::eth::Change> const& _changeLog)
+{
+	json_spirit::mObject oState;
+	if (!_changeLog.size())
+		return oState;
+
+	using changeKind = eth::Change::Kind;
```
chfast: Type aliases must start from upper letter.

test/tools/libtesteth/FillJsonFunctions.cpp
```diff
@@ -52,19 +52,19 @@ json_spirit::mObject fillJsonWithTransaction(Transaction const& _txn)
...
 {
 	json_spirit::mObject oState;
 	if (!_changeLog.size())
 		return oState;
 
-	using changeKind = eth::detail::Change::Kind;
-	using changeStruct = eth::detail::Change;
-	eth::detail::ChangeLog changeLog = _changeLog;
+	using ChangeKind = eth::Change::Kind;
+	using ChangeStruct = eth::Change;
```
chfast: This make no sense. You can use `Change` (because `using namespace dev::eth`), why you want to replace it with `ChangeStruct`.

Similar `ChangeKind` instead of `Change::Kind`?

test/tools/libtesteth/FillJsonFunctions.cpp
```diff
@@ -52,19 +52,19 @@ json_spirit::mObject fillJsonWithTransaction(Transaction const& _txn)
...
+	eth::ChangeLog changeLog = _changeLog;
 
 	//Sort the vector by address field
 	std::sort(changeLog.begin(), changeLog.end(),
-		[ ]( const changeStruct& lhs, const changeStruct& rhs )
+		[ ]( const ChangeStruct& lhs, const ChangeStruct& rhs )
```
chfast: Too much spaces here. Should be: `[](const ChangeStruct& lhs, const ChangeStruct& rhs)`.

test/tools/libtesteth/FillJsonFunctions.cpp
winsvega: because we use mutation approach on this PR

PR #4075, opened 2017-05-11
EVM-C v4

chfast:
EVMJIT part: https://github.com/ethereum/evmjit/pull/118

This includes hardcore vector created with placement new operator. This so more a proof of concept, we can reconsider replacing it. But this is also fun!

### DIFF
--- a/libevm/JitVM.cpp
+++ b/libevm/JitVM.cpp
@@ -131,50 +131,55 @@ void getBlockHash(evm_uint256be* o_hash, evm_env* _envPtr, int64_t _number)
 	*o_hash = toEvmC(env.blockHash(_number));
 }
 
+void create(evm_result* o_result, ExtVMFace& _env, evm_message const* _msg) noexcept
+{
+	u256 gas = _msg->gas;
+	u256 value = fromEvmC(_msg->value);
+	bytesConstRef init = {_msg->input, _msg->input_size};
+	// ExtVM::create takes the sender address from .myAddress.
+	assert(fromEvmC(_msg->sender) == _env.myAddress);
+
+	// TODO: EVMJIT does not support RETURNDATA at the moment, so
+	//       the output is ignored here.
+	h160 addr;
+	std::tie(addr, std::ignore) = _env.create(value, gas, init, {});
+	o_result->gas_left = static_cast<int64_t>(gas);
+	o_result->release = nullptr;
+	if (addr)
+	{
+		o_result->code = EVM_SUCCESS;
+		// Use reserved data to store the address.
+		static_assert(sizeof(o_result->reserved.data) >= addr.size,
+			"Not enough space to store an address");
+		std::copy(addr.begin(), addr.end(), o_result->reserved.data);
+		o_result->output_data = o_result->reserved.data;
+		o_result->output_size = addr.size;
+	}
+	else
+	{
+		o_result->code = EVM_FAILURE;
+		o_result->output_data = nullptr;
+		o_result->output_size = 0;
+	}
+}
+
 void call(evm_result* o_result, evm_env* _opaqueEnv, evm_message const* _msg) noexcept
 {
 	assert(_msg->gas >= 0 && "Invalid gas value");
 	auto &env = reinterpret_cast<ExtVMFace&>(*_opaqueEnv);
-	u256 value = fromEvmC(_msg->value);
-	bytesConstRef input{_msg->input, _msg->input_size};
 
+	// Handle CREATE separately.
 	if (_msg->kind == EVM_CREATE)
-	{
-		u256 gas = _msg->gas;
-		// ExtVM::create takes the sender address from .myAddress.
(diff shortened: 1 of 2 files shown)


libevm/JitVM.cpp
```diff
@@ -131,12 +132,13 @@ void getBlockHash(evm_uint256be* o_hash, evm_env* _envPtr, int64_t _number)
...
-int64_t call(
-	evm_env* _opaqueEnv,
-	evm_message const* _msg,
-	uint8_t* _outputData,
-	size_t _outputSize
-) noexcept
+void releaseVectorInPayload(evm_result const* _result)
+{
+	auto* b = reinterpret_cast<bytes const*>(&_result->payload.bytes);
+	b->~bytes();
```
chriseth: This is very unusual and most of the time dangerous. If there is no other way than to call the destructor manually, please at least add a comment.

chfast: I added a comments about it, and now the "destructor" code is very close to the "constructor" code (using non-capturing lambda).

libevm/JitVM.cpp
```diff
@@ -173,15 +177,35 @@ int64_t call(
...
+	// when finished with it.
+
+	// First assign reference. References are not invalidated when vector
+	// of bytes is moved. See `.takeBytes()` below.
+	o_result->output_data = output.data();
+	o_result->output_size = output.size();
+
+	// Place a new vector of bytes containing output in result's reserved memory.
+	static_assert(sizeof(bytes) <= sizeof(o_result->reserved), "Vector is too big");
+	new(&o_result->reserved) bytes(output.takeBytes());
```
gumb0: Does this placement new magic give us any significant advantage?
We could just save a pointer to dynamically allocated `bytes`, like `o_result->context = new bytes(output.takeBytes())`, then just `delete` it in release function.

Or even `o_result->context = new owning_bytes_ref(move(output))`, then we don't need `owning_bytes_ref::takeBytes`

chfast: Yes, both options are good and safe.

The advantage of placement new is there is no dynamic memory allocation. We use some memory of evm_result object. I really wanted to try if it is possible (this affects evm_result design). But I'm happy to use a safer option at this point.

libevm/JitVM.cpp
```diff
@@ -154,13 +148,23 @@ int64_t call(
...
 		std::tie(addr, std::ignore) = env.create(value, gas, input, {});
-		auto gasLeft = static_cast<int64_t>(gas);
+		o_result->gas_left = static_cast<int64_t>(gas);
+		o_result->release = nullptr;
 		if (addr)
-			std::copy(addr.begin(), addr.end(), _outputData);
+		{
+			o_result->code = EVM_SUCCESS;
+			// Use reserved data to store the address.
+			std::copy(addr.begin(), addr.end(), o_result->reserved.data);
```
gumb0: There was an `assert` to check that `addr` fits into `_outputData`, which you deleted. Maybe add here `assert` or `static_assert` that `o_result->reserved.data` is at least 20 bytes

chfast: Will be done. Good advice.

libevm/JitVM.cpp
```diff
@@ -127,25 +127,19 @@ void getTxContext(evm_tx_context* result, evm_env* _opaqueEnv) noexcept
...
 	*o_hash = toEvmC(env.blockHash(_number));
 }
 
-int64_t call(
-	evm_env* _opaqueEnv,
-	evm_message const* _msg,
-	uint8_t* _outputData,
-	size_t _outputSize
-) noexcept
+void call(evm_result* o_result, evm_env* _opaqueEnv, evm_message const* _msg) noexcept
```
gumb0: This function is now over 70 lines long, and doing two very distinct tasks, performing create and call. Consider splitting it into at least 2 internal functions

libevm/JitVM.cpp
gumb0: Yeah, there's one allocation less and one indirection less, but in general such esoteric tricks should better be used only when we're really sure we need them. 

In general placement new should better not be used outside of lower level memory-managing utility classes like containers, I guess.

But hm, ok if you wanted to try it  :) I'm ok with leaving it like this for now

chfast:
Requested changes applied :).

libevm/JitVM.cpp
```diff
@@ -127,61 +127,90 @@ void getTxContext(evm_tx_context* result, evm_env* _opaqueEnv) noexcept
...
 	*o_hash = toEvmC(env.blockHash(_number));
 }
 
-int64_t call(
-	evm_env* _opaqueEnv,
-	evm_message const* _msg,
-	uint8_t* _outputData,
-	size_t _outputSize
-) noexcept
+void create(evm_result* o_result, ExtVMFace& env, evm_message const* _msg) noexcept
```
gumb0: _env

PR #4077, opened 2017-05-12
Improve bStates unit test

gumb0:


test/unittests/libethereum/Block.cpp
```diff
@@ -108,10 +117,6 @@ BOOST_AUTO_TEST_CASE(bStates)
 	{
 		BOOST_ERROR("Failed test with Exception: " << _e.what());
 	}
-	catch(...)
```
gumb0: @winsvega  Don't  do this inside `BOOST_AUTO_TEST_CASE`. boost.test framework catches any uncaught exceptions and reports it itself.

`BOOST_REQUIRE_` etc. macros throw their internal exceptions themselves, they are caught is such `catch(...)` here and this leads to confusing error messages.

gumb0:
@winsvega  Please try to run tests with these changes on the nightly tests server.

test/unittests/libethereum/Block.cpp
```diff
@@ -50,14 +50,20 @@ BOOST_AUTO_TEST_CASE(bStructures)
 
 BOOST_AUTO_TEST_CASE(bStates)
 {
+	// this test does full Ethash mining
+	if (!dev::test::Options::get().quadratic)
```
winsvega: tested on server run. works fine.

chfast:
Do I get it right, that the "fix" in this case means "disable the test"?

gumb0:
@chfast  Not exactly :) We disable it for Travis (as it was before my previous changes) since it's time-consuming, and there's a fix for it when it runs on nightly test server

chfast:
Ok, but it means I'm not going to test it usually.
Anyway, are any parts of the test more interesting than others so we could extract them to other test cases?

chfast:
It also looks like some code in Block.cpp does not have "default" coverage now: https://codecov.io/gh/ethereum/cpp-ethereum/pull/4077/changes.

gumb0:
I think what we could and should do here is not use chain params with ethash SealEngine, but use NoProof seal engine, which is specifically test-oriented and fast.

I'm doing the similar thing now for Metropolis BLOCKHASH-related tests in this same file, I'll try to modify this test later then. The we could enable it and get the coverage

PR #4088, opened 2017-05-22
Disable builds with UPnP by default

chfast:


chfast:
I pushed 2 more changes:
- remove miniupnpc from install_deps scripts,
- (not related) do not test EVMJIT on AppVeyor (we build without EVMJIT so tasting it make no sense).

Ready to land?

Issue #4078, opened 2017-05-14
CMake build failed (on Windows)


EzBar:
Hi,
I'm trying to build the latest version (develop branch, db8a4a2d9b) using CMake 3.8.1.
I'm getting the following error: 

"E:\Ethereum\cpp-ethereum\build\ALL_BUILD.vcxproj" (default target) (1) ->
"E:\Ethereum\cpp-ethereum\build\eth\eth.vcxproj" (default target) (11) ->
"E:\Ethereum\cpp-ethereum\build\libethashseal\ethashseal.vcxproj" (default target) (13) ->
"E:\Ethereum\cpp-ethereum\build\libethereum\ethereum.vcxproj" (default target) (15) ->
"E:\Ethereum\cpp-ethereum\build\libevm\evm.vcxproj" (default target) (16) ->
(ClCompile target) ->
  e:\ethereum\cpp-ethereum\libethereum\Executive.h(22): fatal error C1083: Cannot open include file: 'json/json.h': No such file or directory (compiling source file E:\Ethereum\cpp-ethereum\libevm\VM.cpp) [
E:\Ethereum\cpp-ethereum\build\libevm\evm.vcxproj]
  e:\ethereum\cpp-ethereum\libethereum\Executive.h(22): fatal error C1083: Cannot open include file: 'json/json.h': No such file or directory (compiling source file E:\Ethereum\cpp-ethereum\libevm\VMOpt.cpp
) [E:\Ethereum\cpp-ethereum\build\libevm\evm.vcxproj]
  e:\ethereum\cpp-ethereum\libethereum\Executive.h(22): fatal error C1083: Cannot open include file: 'json/json.h': No such file or directory (compiling source file E:\Ethereum\cpp-ethereum\libevm\VMCalls.c
pp) [E:\Ethereum\cpp-ethereum\build\libevm\evm.vcxproj]
  e:\ethereum\cpp-ethereum\libethereum\Executive.h(22): fatal error C1083: Cannot open include file: 'json/json.h': No such file or directory (compiling source file E:\Ethereum\cpp-ethereum\libevm\VMValidat
e.cpp) [E:\Ethereum\cpp-ethereum\build\libevm\evm.vcxproj]

    92 Warning(s)
    4 Error(s)

I've followed the instructions on the README file. I'm running Windows 10 with Visual Studio 2015 installed.
Any suggestion?

Best,
Ez

chfast:
[...]

PR #4094, opened 2017-05-25
Fix jsoncpp includes in evm library

chfast:
Fixes https://github.com/ethereum/cpp-ethereum/issues/4078.

### DIFF
--- a/libevm/CMakeLists.txt
+++ b/libevm/CMakeLists.txt
@@ -19,13 +19,11 @@ file(GLOB HEADERS "*.h")
 
 add_library(evm ${SOURCES} ${HEADERS})
 
-find_package(Eth)
-
-target_link_libraries(evm ${Eth_ETHCORE_LIBRARIES})
-target_link_libraries(evm ${Eth_EVMCORE_LIBRARIES})
-target_include_directories(evm PRIVATE .. ${JSONCPP_INCLUDE_DIR})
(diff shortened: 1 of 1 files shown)


libevm/CMakeLists.txt
```diff
@@ -19,13 +19,11 @@ file(GLOB HEADERS "*.h")
 
 add_library(evm ${SOURCES} ${HEADERS})
 
-find_package(Eth)
-
-target_link_libraries(evm ${Eth_ETHCORE_LIBRARIES})
-target_link_libraries(evm ${Eth_EVMCORE_LIBRARIES})
-target_include_directories(evm PRIVATE .. ${JSONCPP_INCLUDE_DIR})
+# TODO: We have horrible cilcular dependency: evm depends also on ethereum
+#       library. Here we only explicitly add jsoncpp to fix includes.
```
winsvega: what about cmake variables like ${Eth_ETHCORE_LIBRARIES} ? 

I do like explicit dependencies declarations though

chfast: Instead of `${Eth_ETHCORE_LIBRARIES}` I'm using `ethcore`. Much simpler and we don't depend on external helper functions/macros. This is what everyone with basic CMake skill will understand. Knowing where  `${Eth_ETHCORE_LIBRARIES}` comes from is much harder.

PR #4095, opened 2017-05-26
Export state change log in state tests

chfast:


test/tools/libtesteth/ImportTest.cpp
```diff
@@ -55,10 +55,10 @@ bytes ImportTest::executeTest()
 	if (m_testType == testType::StateTests)
 	{
 		eth::Network network = eth::Network::MainNetwork;
-		std::pair<eth::State, ImportTest::execOutput> out = executeTransaction(network, m_envInfo, m_statePre, m_transaction);
-		m_statePost = out.first;
-		m_logs = out.second.second.log();
-		return out.second.first.output;
+		auto out = executeTransaction(network, m_envInfo, m_statePre, m_transaction);
```
winsvega: auto makes code harder to read. and intelsense does not work with auto.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -91,8 +91,8 @@ bytes ImportTest::executeTest()
 					continue;
 
 				eth::Network network = networks[j];
-				std::pair<eth::State, ImportTest::execOutput> out = executeTransaction(network, m_envInfo, m_statePre, m_transactions[i].transaction);
-				m_transactions[i].postState = out.first;
+				std::tie(m_transactions[i].postState, std::ignore, std::ignore) =
```
winsvega: will std::tie(m_transactions[i].postState      be assigned after execute transaction ? 
because postState would be ready after execution

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -183,16 +183,16 @@ bytes ImportTest::executeTest()
...
 {
-	eth::State initialState = _preState;
 	try
 	{
 		unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(_sealEngineNetwork)).createSealEngine());
 		bool removeEmptyAccounts = m_envInfo.number() >= se->chainParams().u256Param("EIP158ForkBlock");
-		ImportTest::execOutput execOut = initialState.execute(_env, *se.get(), _tr);
-		initialState.commit(removeEmptyAccounts ? State::CommitBehaviour::RemoveEmptyAccounts : State::CommitBehaviour::KeepEmptyAccounts);
-		return std::pair<eth::State, ImportTest::execOutput>(initialState, execOut);
+		ImportTest::ExecOutput execOut = _state.execute(_env, *se.get(), _tr);
```
winsvega: hard to see that _state is actually a copy of preState. wouldnt it be better if ? 
executeTransactino( ..   const& _preState .. )
{
State state = preState;
...
}

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -183,16 +183,16 @@ bytes ImportTest::executeTest()
...
+		ChangeLog changeLog = _state.changeLog();
```
winsvega: the issue is still here.  changeLog being cleaned in state at _state.execute  function line 553

winsvega: what Permanence does ?  if tis not affecting the results perhaps we should use this flag.

test/tools/libtesteth/ImportTest.cpp
chfast: This is exactly as it was.

test/tools/libtesteth/ImportTest.cpp
chfast: No. This is bad practice in C++. If you need a copy of an argument you make it by passing the argument by value. We can change the name of argument from `_state` to `state` to indicate this.

test/tools/libtesteth/ImportTest.cpp
chfast: I see. Then you have to work on this more.

PR #4096, opened 2017-05-26
CMake: simplify options declarations + other fixes

chfast:
This simplifies CMake options declaration to make sure they are cached variables (issue discovered in #4088).


chfast:
CMake changes are just: use CMake's `option()` instead of combination of our legacy macros.

matthod:
Fix seems to work. Builds OK now. Thanks for quick work on this!

PR #4098, opened 2017-05-26
Precompiled: change gas costs for modexp

pirapira:
This reflects https://github.com/ethereum/EIPs/commit/7138d542c43d1ff6305d79ea6ef19e6c6a61db15 and the comment https://github.com/ethereum/EIPs/commit/7138d542c43d1ff6305d79ea6ef19e6c6a61db15#commitcomment-22292809.

I will add tests once the EIP text is updated for [this comment](https://github.com/ethereum/EIPs/commit/7138d542c43d1ff6305d79ea6ef19e6c6a61db15#commitcomment-22292809).


libethcore/Precompiled.cpp
```diff
@@ -132,15 +132,31 @@ ETH_REGISTER_PRECOMPILED(modexp)(bytesConstRef _in)
 	return {true, ret};
 }
 
+namespace
+{
+	bigint expLengthAdjust(size_t _baseLength, size_t _expLength, bytesConstRef _in)
+	{
+		size_t const numConsideredBytes { min(_expLength, size_t(32)) };
```
gumb0: If using braces for initialization, use the same amount of spaces as with parentheses.
So `size_t const numConsideredBytes{min(_expLength, size_t(32))};`

pirapira: That makes sense.

pirapira:
I'll add a few more tests tomorrow.

libethcore/Precompiled.cpp
```diff
@@ -132,15 +132,34 @@ ETH_REGISTER_PRECOMPILED(modexp)(bytesConstRef _in)
 	return {true, ret};
 }
 
+namespace
+{
+	bigint expLengthAdjust(bigint _baseLength, bigint _expLength, bytesConstRef _in)
+	{
+		size_t const numConsideredBytes(min(_expLength, {32}));
+		bigint expConsideredBytes{};
+		if (96 + _baseLength == bigint(size_t(96 + _baseLength)))
```
gumb0: What is this `if` for?

pirapira: For detecting overflows.  Say, `_baseLength` is `2^64 - 96`, and we will be reading the exponent from the wrong position.

gumb0: Then I guess we need to add checks that lengths fit into `size_t` into modexp implementation itself.

Also I think I would use `std::numeric_limits` for that instead of C-style-cast tricks

libethcore/Precompiled.cpp
```diff
@@ -132,15 +132,34 @@ ETH_REGISTER_PRECOMPILED(modexp)(bytesConstRef _in)
 	return {true, ret};
 }
 
+namespace
+{
+	bigint expLengthAdjust(bigint _baseLength, bigint _expLength, bytesConstRef _in)
+	{
+		size_t const numConsideredBytes(min(_expLength, {32}));
```
gumb0: I'd say this `numConsideredBytes` trick and having one return statement makes it pretty hard to read.

I'd express this somewhat closer to EIP spec with something like this:
```
if (_expLength <= 32)
{
	bigint const exp(parseBigEndianRightPadded(_in, 96 + _baseLength, expLength));
	return exp ? msb(exp) : 0;
}
else
{
	bigint const expFirstWord(parseBigEndianRightPadded(_in, 96 + _baseLength, 32));
	size_t const highestBit = expFirstWord ? msb(expFirstWord) : 0;
	return 8 * (_expLength - 32) + highestBit;
}
```

gumb0: Also, better pass `_expOffset` into this function instead of `_baseLength` and leave `expOffset = baseLength + 96` calculation in the calling function

pirapira: OK.  That's closer to what I wrote first.

I also think I should change `parseBigEndianRightPadded` to take `bigint` instead of `size_t`.

I'll make these changes after adding tests.

libethcore/Precompiled.cpp
pirapira: I added an assert in `modexp` implementation.  I converted these checks with `std::numeric_limits`.

libethcore/Precompiled.cpp
pirapira: Done.  Now I look into Travis failure (maybe I need to update the json tests).

gumb0: Yeah, cost change affects state, so you definitely need to regenerate state tests for Metropolis

libethcore/Precompiled.cpp
```diff
@@ -99,48 +99,77 @@ ETH_REGISTER_PRECOMPILED(identity)(bytesConstRef _in)
 
 // Parse _count bytes of _in starting with _begin offset as big endian int.
 // If there's not enough bytes in _in, consider it infinitely right-padded with zeroes.
-bigint parseBigEndianRightPadded(bytesConstRef _in, size_t _begin, size_t _count)
+bigint parseBigEndianRightPadded(bytesConstRef _in, bigint _begin, bigint _count)
 {
-	if (_begin > _in.count())
+	if (_begin > numeric_limits<size_t>::max())
```
gumb0: Hmm, seems like these two if's don't change much. Maybe having parameters as `bigint` would be enough?

if `_begin > numeric_limits<size_t>::max()` it won't pass the check `if (begin > _in.count())` just below.

Price calculation uses at most `_count=32` and if any length is greater than `numeric_limits<size_t>::max()` the price would be too big and we shouldn't get to implementation.

libethcore/Precompiled.cpp
```diff
@@ -99,48 +99,77 @@ ETH_REGISTER_PRECOMPILED(identity)(bytesConstRef _in)
 
 // Parse _count bytes of _in starting with _begin offset as big endian int.
 // If there's not enough bytes in _in, consider it infinitely right-padded with zeroes.
-bigint parseBigEndianRightPadded(bytesConstRef _in, size_t _begin, size_t _count)
+bigint parseBigEndianRightPadded(bytesConstRef _in, bigint _begin, bigint _count)
```
gumb0: I think we should better pass `bigint`s by const reference, copying could be expensive.

libethcore/Precompiled.cpp
```diff
@@ -99,48 +99,77 @@ ETH_REGISTER_PRECOMPILED(identity)(bytesConstRef _in)
...
 		return 0;
 
 	// crop _in, not going beyond its size
-	bytesConstRef cropped = _in.cropped(_begin, min(_count, _in.count() - _begin));
+	bytesConstRef cropped = _in.cropped(begin, min(count, _in.count() - begin));
 
 	bigint ret = fromBigEndian<bigint>(cropped);
 	// shift as if we had right-padding zeroes
-	ret <<= 8 * (_count - cropped.count());
+	ret <<= 8 * (count - cropped.count());
```
gumb0: I'd leave the assert before this line to check that the right-hand operand of shift doesn't overflow size_t

libethcore/Precompiled.cpp
```diff
@@ -99,48 +99,77 @@ ETH_REGISTER_PRECOMPILED(identity)(bytesConstRef _in)
...
 
 ETH_REGISTER_PRECOMPILED(modexp)(bytesConstRef _in)
 {
-	size_t const baseLength(parseBigEndianRightPadded(_in, 0, 32));
-	size_t const expLength(parseBigEndianRightPadded(_in, 32, 32));
-	size_t const modLength(parseBigEndianRightPadded(_in, 64, 32));
+	bigint const baseLength(parseBigEndianRightPadded(_in, 0, 32));
+	bigint const expLength(parseBigEndianRightPadded(_in, 32, 32));
+	bigint const modLength(parseBigEndianRightPadded(_in, 64, 32));
+	assert(modLength <= numeric_limits<size_t>::max()); // Otherwise gas should be too expensive
```
gumb0: And I'd added asserts with similar checks for baseLength & expLength

pirapira: I cannot do that for `expLength` because when both `modLength` and `baseLength` are zero, the gas cost is zero even for a very big `expLength`.

I will add one for `baseLength`.

pirapira: (Perhaps, I should add a special case for both `modLength` and `baseLength` being zero.)

libethcore/Precompiled.cpp
pirapira: This shift on `bigint` and it's supposed to overflow the range of `size_t` sometimes.  Especially when this routing is parsing `base`, `mod` or `exp` (not their lengths).

libethcore/Precompiled.cpp
pirapira: Done locally, will appear here.

libethcore/Precompiled.cpp
pirapira: Even if `_begin > numeric_limits<size_t>::max()`, when `_begin` is one more than the max value, `begin` would be zero.  So I leave these tests.

gumb0: No, I think you should compare `if (_begin > _in.count())` instead of `if (begin > _in.count())` and get rid of at least the first test.

In fact having the huge `_begin` should not be a problem for this function, since we know that input is infinitely right-padded with zeroes, it should return 0. It's not like this is an invalid `_begin` argument

pirapira: OK.  I'll do that for `_begin`.  

For `_count`, I'll change the condition to assertion.  If `_count` is bigger than the max `size_t`, the resulting number wouldn't fit in the memory.

libethcore/Precompiled.cpp
gumb0: Right-hand operand to shift is not a bigint, it's probably `size_t`
This makes sense because shifting to more that `size_t` max would require extraordinary amount of memory

pirapira: I see.

pirapira: So I added an assertion.

libethcore/Precompiled.cpp
```diff
@@ -99,48 +99,78 @@ ETH_REGISTER_PRECOMPILED(identity)(bytesConstRef _in)
...
-	size_t const expLength(parseBigEndianRightPadded(_in, 32, 32));
-	size_t const modLength(parseBigEndianRightPadded(_in, 64, 32));
+	bigint const baseLength(parseBigEndianRightPadded(_in, 0, 32));
+	bigint const expLength(parseBigEndianRightPadded(_in, 32, 32));
+	bigint const modLength(parseBigEndianRightPadded(_in, 64, 32));
+	assert(modLength <= numeric_limits<size_t>::max() / 8); // Otherwise gas should be too expensive.
+	assert(baseLength <= numeric_limits<size_t>::max() / 8); // Otherwise, gas should be too expensive.
+	if (modLength == 0 && baseLength == 0)
+		return {true, bytes{}}; // This is a special case where expLength can be very big.
+	assert(expLength << numeric_limits<size_t>::max() / 8);
```
gumb0: `<=`

pirapira:
@gumb0 @chfast ready to merge?

libethcore/Precompiled.cpp
```diff
@@ -99,48 +99,78 @@ ETH_REGISTER_PRECOMPILED(identity)(bytesConstRef _in)
 
 // Parse _count bytes of _in starting with _begin offset as big endian int.
 // If there's not enough bytes in _in, consider it infinitely right-padded with zeroes.
-bigint parseBigEndianRightPadded(bytesConstRef _in, size_t _begin, size_t _count)
+bigint parseBigEndianRightPadded(bytesConstRef _in, bigint const& _begin, bigint const& _count)
 {
 	if (_begin > _in.count())
 		return 0;
+	assert(_count <= numeric_limits<size_t>::max() / 8); // Otherwise, the return value would not fit in the memory.
```
chfast: This is still absurd big number. Hardware uses ~40 bits for addresses :)

pirapira: I know, but the era of non-volatile memory is coming!

pirapira:
I'll fix the conflict.

pirapira:
The test failure looks a bit similar to what I saw in `snark` branch.

PR #4100, opened 2017-05-26
Remove ethminer

chfast:
Remove support for mining in farm mode, including ethminer executable.


chfast:
This also removes CURL dependency. I'm waiting for builds and tests...

PR #4105, opened 2017-05-31
Parse the "allowFutureBlocks" in a more flexible manner.

chriseth:
This is required for testing Solidity.

### DIFF
--- a/libethereum/ChainParams.cpp
+++ b/libethereum/ChainParams.cpp
@@ -73,7 +73,11 @@ ChainParams ChainParams::loadConfig(string const& _json, h256 const& _stateRoot)
 		else
 		{
 			if (i.first == "allowFutureBlocks")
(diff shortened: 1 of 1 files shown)


chfast:
I don't understand why this is needed. This also means that `0` of `"0"` would represent `true`.
If you have a good reason go ahead and merge it, but it might hit us some day in the future.

chriseth:
We made a breaking change to how this string is parsed. Previously, the presence of the key in the json would activate the feature. This was changed to "the feature is active if the key is present in the json, but the json will not be parsed if the value is not a bool". Both are obivously wrong, so I changed it to "If the key is present and the value is a bool, its value determines whether the feature is active or not and if it not a bool, the feature is active regardless of the value". So this change improves backwards compatibility while also actually using the value supplied.
We could try parsing the string, but as there is no documentation or specification for the feature anyway, ...

PR #4106, opened 2017-06-02
reimplement const pool and reenable const jumps and jump dispatch opt.

gcolvin:
With a maximum size on code the constant pool can now be a vector of all constants, quickly constructed and not exploitable at runtime - unlike the previous small table and hash table.  Also turned back on the other optimizations, and added new opcodes to the dispatch table and debugged it.


libevm/VM.cpp
```diff
@@ -15,8 +15,6 @@
 	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
 */
 /** @file VM.cpp
```
chfast: Remove the full comment section.

libevm/VM.h
```diff
@@ -15,8 +15,6 @@
 	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
 */
 /** @file VM.h
```
chfast: Remove full comment section.

test/unittests/performance/tests.mk
```diff
@@ -26,9 +26,9 @@ endif
 	$(TEST_ETHVM) $@
 	$(TEST_EVM) $@ run
 
-% : %.c
-	$(CCC) $* $^
-	$(STATS) ./$*
+mul64c: mul64c.c
```
chfast: Can you add some description in the beginning of this file how to use it?

gcolvin: OK, will sneak into some semi-relevant PR.

libevm/VM.cpp
gcolvin: That is, the entire copyright section?  Or just section starting with `/**`  ?

PR #4111, opened 2017-06-03
change and add opcodes to conform with EIP 615

gcolvin:
Instructions were unchanged since before the EIP.


libevmcore/Instruction.h
```diff
@@ -184,6 +175,17 @@ enum class Instruction: uint8_t
 	JUMPC,              ///< alter the program counter - pre-verified
 	JUMPCI,             ///< conditionally alter the program counter - pre-verified
 
+	JUMPTO = 0xb0,      ///< alter the program counter to a jumpdest
```
gumb0: The comments for JUMPTO & JUMPV are now equal (as are JUMPSUB & JUMSUBV). Some distinguishing note would be helpful

gcolvin: Cut an and paste - wii fix it soon.  Thanks

PR #4114, opened 2017-06-05
random blockchain tests. Parse all files in bcRandom folder.

winsvega:


test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -102,14 +102,20 @@ void checkBlocks(TestBlock const& _blockFromFields, TestBlock const& _blockFromR
 
 void doBlockchainTests(json_spirit::mValue& _v, bool _fillin)
 {
-	if (!Options::get().fillchain) //fill blockchain through state tests
+	bool skipTestOutput = false;
+	string casename = boost::unit_test::framework::current_test_case().p_name;
+	if (casename == "bcRandom" || Options::get().fillchain)
+		skipTestOutput = true;
+
+	if (!skipTestOutput)
```
gumb0: This variable seems to have reverted meaning. Shouldn't it be when `skipTestOutput == true`, don't do `TestOutputHelper::initTest` ?

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -102,14 +102,20 @@ void checkBlocks(TestBlock const& _blockFromFields, TestBlock const& _blockFromR
...
 		TestOutputHelper::initTest(_v);
+
 	for (auto& i: _v.get_obj())
 	{
 		string testname = i.first;
 		json_spirit::mObject& o = i.second.get_obj();
 
-		if (!Options::get().fillchain)
+		if (!skipTestOutput)
 		if (!TestOutputHelper::passTest(o, testname))
```
gumb0: This line should be one level further indented. Or use `&&` instead of nested `if`

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -407,7 +413,9 @@ void doBlockchainTests(json_spirit::mValue& _v, bool _fillin)
 			ImportTest::compareStates(postState, blockchain.topBlock().state());
 		}
 	}//for tests
-	TestOutputHelper::finishTest();
+
+	if (!skipTestOutput)
+		TestOutputHelper::finishTest();
```
gumb0: Previously This was executed when `Options::get().fillchain == flase`, now it's not, is this correct?

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -102,14 +102,20 @@ void checkBlocks(TestBlock const& _blockFromFields, TestBlock const& _blockFromR
 
 void doBlockchainTests(json_spirit::mValue& _v, bool _fillin)
 {
-	if (!Options::get().fillchain) //fill blockchain through state tests
+	bool skipTestOutput = false;
+	string casename = boost::unit_test::framework::current_test_case().p_name;
+	if (casename == "bcRandom" || Options::get().fillchain)
```
gumb0: I don't like the test case names being hardcoded deep in the test framework code. Could we maybe have a dynamic option in `Options::get()` that has a list of test case names for which we skip output ?

test/tools/jsontests/BlockChainTestsBoost.cpp
```diff
@@ -173,4 +173,36 @@ BOOST_AUTO_TEST_CASE(userDefinedFile)
 	dev::test::userDefinedTest(dev::test::doBlockchainTests);
 }
 BOOST_AUTO_TEST_SUITE_END()
+
+//BlockChainTests
+BOOST_AUTO_TEST_SUITE(BlockChainTestsRandom)
+
+BOOST_AUTO_TEST_CASE(bcRandom)
```
gumb0: Do I understand the general idea correctly - this is a special-case code which is executed for RandomTests dir instead of the regular code in `doBlockchainTests` ?
Then perhaps it should be also in BlockChainTests.cpp, not here?

test/tools/jsontests/BlockChainTests.cpp
winsvega: it is still being executed when fillchain == false. 
it should be executed only in blockchain tests and not in bcRandom

test/tools/jsontests/BlockChainTestsBoost.cpp
winsvega: BlockChainTestsBoost.cpp   
file contains boost test cases for blockchain tests

test/tools/jsontests/BlockChainTestsBoost.cpp
```diff
@@ -173,4 +173,35 @@ BOOST_AUTO_TEST_CASE(userDefinedFile)
...
+{
+	using Bdit = boost::filesystem::directory_iterator;
+	std::string fillersPath = dev::test::getTestPath() + "/src/BlockchainTestsFiller/RandomTests";
+
+	Bdit iterator_tmp(fillersPath);
```
gumb0: this iterator can be declared inside `for`. Also don't use underscores in names.

test/tools/jsontests/BlockChainTestsBoost.cpp
```diff
@@ -173,4 +173,35 @@ BOOST_AUTO_TEST_CASE(userDefinedFile)
...
+	int fileCount = 0;
+	for(; iterator_tmp != Bdit(); ++iterator_tmp)
```
gumb0: Also this is simpler expressed with something like
```
auto fileCount = count_if(Bdit(fillersPath), Bdit(), [](Bdit it) { 
    return boost::filesystem::is_regular_file(it->path()) && it->path().extension() == ".json";  });
```

test/tools/libtesteth/TestHelper.h
```diff
@@ -166,6 +166,7 @@ void doTransactionTests(json_spirit::mValue& _v, bool _fillin);
 void doStateTests(json_spirit::mValue& v, bool _fillin);
 void doVMTests(json_spirit::mValue& v, bool _fillin);
 void doBlockchainTests(json_spirit::mValue& _v, bool _fillin);
+void doBlockchainTestsInternal(json_spirit::mValue& _v, bool _fillin);
```
gumb0: It's not really internal since it's in the header file. I'd name this one maybe just `doBlockchainTests` and another one `doBlockchainTestsWithLog` or something

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -61,7 +61,7 @@ void TestOutputHelper::initTest(json_spirit::mValue& _v)
 	m_currTest = 0;
 }
 
-bool TestOutputHelper::passTest(json_spirit::mObject& _o, std::string& _testName)
+bool TestOutputHelper::passTest(std::string& _testName)
```
gumb0: Pass `_testName` by const reference

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -102,16 +102,24 @@ void checkBlocks(TestBlock const& _blockFromFields, TestBlock const& _blockFromR
...
 	{
 		string testname = i.first;
 		json_spirit::mObject& o = i.second.get_obj();
 
-		if (!Options::get().fillchain)
-		if (!TestOutputHelper::passTest(o, testname))
+		//Select test by name if --singletest is set and not filling state tests as blockchain
+		if (!TestOutputHelper::passTest(testname) && !Options::get().fillchain)
+		{
+			o.clear(); //don't add irrelevant tests to the final file when filling
```
pirapira: I'm puzzled why `o.clear()` is necessary.  After this line, `continue` will destroy `o` anyway.

pirapira: Ah, is it the case that the destructor of `json_sprit::mObject` performs some writing out?

pirapira: Ah, it changes the json element in `_v` somehow.

winsvega: o is taken by & from _v.  so if you fill a single test from json file that contains many tests.
the other tests wont be filled.

test/tools/libtesteth/TestHelper.h
winsvega: that would require to change a lot of other lines allover the testeth
any other name suggestions?

gumb0: Well, another option: `doBlockchainTest` & `doBlockchainTestWithoutLog` maybe

test/tools/jsontests/StateTests.cpp
```diff
@@ -117,20 +117,18 @@ class generaltestfixture
...
+
 		if (dev::test::Options::get().filltests)
 			fileCount *= 2; //tests are checked when filled and after they been filled
 		dev::test::TestOutputHelper::initTest(fileCount);
 
-		boost::filesystem::directory_iterator iterator(fillersPath);
-		for(; iterator != boost::filesystem::directory_iterator(); ++iterator)
-			if (boost::filesystem::is_regular_file(iterator->path()) && iterator->path().extension() == ".json")
+		for(Bdit it(fillersPath); it != Bdit(); ++it)
+			if (boost::filesystem::is_regular_file(it->path()) && it->path().extension() == ".json")
```
gumb0: This code is repeated at least 4 times, maybe make a small helper like `isJsonFile(boost::filesystem::directory_entry const&)`

test/tools/jsontests/BlockChainTestsBoost.cpp
```diff
@@ -173,4 +173,31 @@ BOOST_AUTO_TEST_CASE(userDefinedFile)
...
+	using Bdit = boost::filesystem::directory_iterator;
+	std::string fillersPath = dev::test::getTestPath() + "/src/BlockchainTestsFiller/RandomTests";
+
+	auto fileCount = count_if(Bdit(fillersPath), Bdit(), [](boost::filesystem::directory_entry const& dirent) {
+		return boost::filesystem::is_regular_file(dirent.path()) && dirent.path().extension() == ".json";  });
+
+	//bcRandom tests are generated from random state tests and have 1 test case * 5 forks in each file
+	dev::test::TestOutputHelper::initTest(fileCount * 5);
+
+	for(Bdit it(fillersPath); it != Bdit(); ++it)
```
gumb0: Space is needed after `for` keyword according to coding standards

test/tools/jsontests/StateTests.cpp
```diff
@@ -117,20 +117,18 @@ class generaltestfixture
...
+			return boost::filesystem::is_regular_file(dirent.path()) && dirent.path().extension() == ".json";  });
+
 		if (dev::test::Options::get().filltests)
 			fileCount *= 2; //tests are checked when filled and after they been filled
 		dev::test::TestOutputHelper::initTest(fileCount);
 
-		boost::filesystem::directory_iterator iterator(fillersPath);
-		for(; iterator != boost::filesystem::directory_iterator(); ++iterator)
-			if (boost::filesystem::is_regular_file(iterator->path()) && iterator->path().extension() == ".json")
+		for(Bdit it(fillersPath); it != Bdit(); ++it)
```
gumb0: space after for

test/tools/jsontests/StateTests.cpp
gumb0: And maybe another one for counting the number of json files in directory (this code is also duplicated now)

test/tools/jsontests/BlockChainTestsBoost.cpp
```diff
@@ -173,4 +172,26 @@ BOOST_AUTO_TEST_CASE(userDefinedFile)
...
+{
+	std::string fillersPath = dev::test::getTestPath() + "/src/BlockchainTestsFiller/RandomTests";
+
+	std::vector<boost::filesystem::path> files = test::getJsonFiles(fillersPath);
+	int fileCount = files.size();
+
+	//bcRandom tests are generated from random state tests and have 1 test case * 5 forks in each file
+	dev::test::TestOutputHelper::initTest(fileCount * 5);
+
+	for (size_t i = 0; i < files.size(); i++)
```
gumb0: You could use range for here then
`for (auto const& file: files)`

PR #4117, opened 2017-06-06
EVM-C: Split state update functions

chfast:
EVMJIT: https://github.com/ethereum/evmjit/pull/119.

This change splits state update combo callback function into 3 functions. The old combo callback was used to handle SSTORE, SELFDESTRUCT and LOG. Now each opcode is handled independently. This means more callback functions to implement, but they are short and simple. And no additional switch needed.

### DIFF
--- a/libevm/JitVM.cpp
+++ b/libevm/JitVM.cpp
@@ -99,12 +99,21 @@ void updateState(
 		env.setStore(index, value);    // Interface uses native endianness
 		break;
 	}
-	case EVM_SELFDESTRUCT:
-		// Register selfdestruction beneficiary.
-		env.suicide(fromEvmC(_arg1->address));
-		break;
 	}
 }
+
+void selfdestruct(
+	evm_env* _opaqueEnv,
+	evm_uint160be const* _addr,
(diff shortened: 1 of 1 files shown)


libevm/JitVM.cpp
```diff
@@ -76,41 +76,51 @@ void queryState(
...
+	size_t _dataSize,
+	evm_uint256be const _topics[],
+	size_t _numTopics
+) noexcept
+{
+	(void) _addr;
+	auto &env = reinterpret_cast<ExtVMFace&>(*_opaqueEnv);
+	assert(fromEvmC(*_addr) == env.myAddress);
+	h256 const* pTopics = reinterpret_cast<h256 const*>(_topics);
+	env.log({pTopics, pTopics + _numTopics}, {_data, _dataSize});
```
gumb0: This is little confusing, not clear what exactly is being passed to `log`, I'd probable name the constructed types explicitly, like
`env.log(h256s{pTopics, pTopics + _numTopics}, bytesConstRef{_data, _dataSize});`

PR #4122, opened 2017-06-08
SIMD

gcolvin:
Portable implementations of SIMD operations for VM, including ocpodes, instruction info, and interpreter cases.  Ugliest hack is the VM::StackItem union - needed because boost u256 type can't go in union directly, but can be copied as bits.


gcolvin:
Attempt to rebase on this branch made a hopeless mess.  Started over on a new branch.

PR #4124, opened 2017-06-08
Returndatacopy throws when offset+size is bigger than the size of returndata buffer

pirapira:
This follows the [change in the EIP](https://github.com/ethereum/EIPs/pull/211/commits/28f1709777cbd99dc96dcdbbef8c532674bfbc86#diff-fd8c393953fe66162d8bb518d87bfb57R43).

### DIFF
--- a/libevm/VM.cpp
+++ b/libevm/VM.cpp
@@ -763,6 +763,9 @@ void VM::interpretCases()
 		{
 			if (!m_schedule->haveReturnData)
 				throwBadInstruction();
+			bigint const endOfAccess{bigint(m_SP[1]) + bigint(m_SP[2])};
+			if (m_returnData.size() < endOfAccess)
+				throwBufferOverrun(endOfAccess);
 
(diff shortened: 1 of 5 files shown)


libevm/VM.cpp
```diff
@@ -763,6 +763,9 @@ void VM::interpretCases()
 		{
 			if (!m_schedule->haveReturnData)
 				throwBadInstruction();
+			bigint const endOfAccess{bigint(m_SP[1]) + bigint(m_SP[2])};
```
chfast: I would avoid explicit constructor here to still pretend bigint works like regular int.
```
bigint const endOfAccess = bigint(m_SP[1]) + bigint(m_SP[2]);
```

pirapira: Let me pretend that way.

pirapira:
@gumb0 @chriseth @chfast  ready to merge?

chfast:
Yes.

Issue #4084, opened 2017-05-19
Don't ask for lastHashes from Blockchain after Metropolis


gumb0:
Follow-up to https://github.com/ethereum/cpp-ethereum/pull/4066

Starting from block `METROPOLIS_FORK_BLOCK+256` BLOCKHASH opcode relies on the call to contract instead of asking `Blockchain` and we could skip creating `lastHashes` array, saving performace.

gumb0:
Refactoring work started in the branch https://github.com/ethereum/cpp-ethereum/tree/lasthashes-refactoring

PR #4126, opened 2017-06-08
LastHashes refactoring

gumb0:
Follow-up to https://github.com/ethereum/cpp-ethereum/pull/4066
Resolves https://github.com/ethereum/cpp-ethereum/issues/4084

Refactoring basically replaces `LastHashes` vector constructed by `Blockchain` class and passed to `Executive` when we need to execute some VM code with `LastBlockHashesFace` interface provided by `BlockChain` class.
The new interface constructs exactly the same vector, and the behavior before Metropolis is not changed.

After Metropolis BLOCKHASH doesn't access this interface (performs a call to contract instead) and this way we avoid constructing unnecessary hash vector.

### DIFF
--- a/ethvm/main.cpp
+++ b/ethvm/main.cpp
@@ -19,10 +19,7 @@
  * @date 2014
  * EVM Execution tool.
  */
-#include <fstream>
-#include <iostream>
-#include <ctime>
-#include <boost/algorithm/string.hpp>
+
 #include <libdevcore/CommonIO.h>
 #include <libdevcore/RLP.h>
 #include <libdevcore/SHA3.h>
@@ -30,10 +27,15 @@
 #include <libethereum/Block.h>
 #include <libethereum/Executive.h>
 #include <libethereum/ChainParams.h>
+#include <libethereum/LastBlockHashesFace.h>
 #include <libethashseal/GenesisInfo.h>
 #include <libethashseal/Ethash.h>
 #include <libevm/VM.h>
 #include <libevm/VMFactory.h>
+#include <boost/algorithm/string.hpp>
+#include <fstream>
+#include <iostream>
+#include <ctime>
 using namespace std;
 using namespace dev;
 using namespace eth;
@@ -118,6 +120,13 @@ enum class Mode
 	Test
 };
 
+class LastBlockHashes: public eth::LastBlockHashesFace
+{
+public:
+	h256s precedingHashes(h256 const& /* _mostRecentHash */) const override { return h256s(256, h256()); }
+	void clear() override {}
+};
+
 int main(int argc, char** argv)
 {
 	setDefaultOrCLocale();
@@ -132,9 +141,9 @@ int main(int argc, char** argv)
 	u256 gasPrice = 0;
 	bool styledJson = true;
 	StandardTrace st;
-	EnvInfo envInfo;
 	Network networkName = Network::MainNetwork;
-	envInfo.setGasLimit(maxBlockGasLimit());
+	BlockHeader blockHeader; // fake block to be executed in
+	blockHeader.setGasLimit(maxBlockGasLimit());
 	bytes data;
 	bytes code;
 
@@ -178,15 +187,15 @@ int main(int argc, char** argv)
 		else if (arg == "--gas-price" && i + 1 < argc)
 			gasPrice = u256(argv[++i]);
 		else if (arg == "--author" && i + 1 < argc)
-			envInfo.setAuthor(Address(argv[++i]));
+			blockHeader.setAuthor(Address(argv[++i]));
 		else if (arg == "--number" && i + 1 < argc)
-			envInfo.setNumber(u256(argv[++i]));
+			blockHeader.setNumber(u256(argv[++i]));
 		else if (arg == "--difficulty" && i + 1 < argc)
-			envInfo.setDifficulty(u256(argv[++i]));
+			blockHeader.setDifficulty(u256(argv[++i]));
 		else if (arg == "--timestamp" && i + 1 < argc)
-			envInfo.setTimestamp(u256(argv[++i]));
+			blockHeader.setTimestamp(u256(argv[++i]));
 		else if (arg == "--gas-limit" && i + 1 < argc)
-			envInfo.setGasLimit(u256(argv[++i]).convert_to<int64_t>());
+			blockHeader.setGasLimit(u256(argv[++i]).convert_to<int64_t>());
 		else if (arg == "--value" && i + 1 < argc)
 			value = u256(argv[++i]);
 		else if (arg == "--network" && i + 1 < argc)
@@ -274,6 +283,8 @@ int main(int argc, char** argv)
 	state.addBalance(sender, value);
 
 	unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(networkName)).createSealEngine());
+	LastBlockHashes lastBlockHashes;
+	EnvInfo const envInfo(blockHeader, lastBlockHashes, 0);
 	Executive executive(state, envInfo, *se);
 	ExecutionResult res;
 	executive.setResultRecipient(res);
@@ -313,7 +324,7 @@ int main(int argc, char** argv)
 
 	if (mode == Mode::Statistics)
 	{
-		cout << "Gas used: " << res.gasUsed << " (+" << t.baseGasRequired(se->evmSchedule(envInfo)) << " for transaction, -" << res.gasRefunded << " refunded)" << endl;
+		cout << "Gas used: " << res.gasUsed << " (+" << t.baseGasRequired(se->evmSchedule(envInfo.number())) << " for transaction, -" << res.gasRefunded << " refunded)" << endl;
 		cout << "Output: " << toHex(output) << endl;
 		LogEntries logs = executive.logs();
 		cout << logs.size() << " logs" << (logs.empty() ? "." : ":") << endl;
--- a/libdevcore/vector_ref.h
+++ b/libdevcore/vector_ref.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "db.h"
+
 #include <cstring>
 #include <cassert>
 #include <type_traits>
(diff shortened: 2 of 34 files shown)


libdevcore/vector_ref.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "db.h"
```
gumb0: This vector_ref.h file uses `DEV_LDB` macro defined in db.h
This came up after reordering of #include's in another file

libethashseal/Ethash.h
```diff
@@ -45,7 +45,7 @@ class Ethash: public SealEngineBase
 
 	StringHashMap jsInfo(BlockHeader const& _bi) const override;
 	void verify(Strictness _s, BlockHeader const& _bi, BlockHeader const& _parent, bytesConstRef _block) const override;
-	void verifyTransaction(ImportRequirements::value _ir, TransactionBase const& _t, EnvInfo const& _env) const override;
+	void verifyTransaction(ImportRequirements::value _ir, TransactionBase const& _t, BlockHeader const& _header, u256 const& _startGasUsed) const override;
```
gumb0: The reason for this is that I find it weird that transaction validation depends on `EnvInfo` which is an enviroment info for VM execution.
Also `EnvInfo` now contains a pointer to `LastBlockHashesFace`, which shouldn't be needed by transation validation at all.

libethereum/Client.h
```diff
@@ -100,10 +100,6 @@ class Client: public ClientBase, protected Worker
 	/// Queues a block for import.
 	ImportResult queueBlock(bytes const& _block, bool _isSafe = false);
 
-	using Interface::call; // to remove warning about hiding virtual function
-	/// Makes the given call. Nothing is recorded into the state. This cheats by creating a null address and endowing it with a lot of ETH.
-	ExecutionResult call(Address _dest, bytes const& _data = bytes(), u256 _gas = 125000, u256 _value = 0, u256 _gasPrice = 1 * ether, Address const& _from = Address());
```
gumb0: This is not used

libethereum/Interface.h
```diff
@@ -95,13 +95,6 @@ class Interface
 	ExecutionResult call(Secret const& _secret, u256 _value, Address _dest, bytes const& _data, u256 _gas, u256 _gasPrice, BlockNumber _blockNumber, FudgeFactor _ff = FudgeFactor::Strict) { return call(toAddress(_secret), _value, _dest, _data, _gas, _gasPrice, _blockNumber, _ff); }
 	ExecutionResult call(Secret const& _secret, u256 _value, Address _dest, bytes const& _data, u256 _gas, u256 _gasPrice, FudgeFactor _ff = FudgeFactor::Strict) { return call(toAddress(_secret), _value, _dest, _data, _gas, _gasPrice, _ff); }
 
-	/// Does the given creation. Nothing is recorded into the state.
```
gumb0: These were not used

ethvm/main.cpp
```diff
@@ -313,7 +313,7 @@ int main(int argc, char** argv)
 
 	if (mode == Mode::Statistics)
 	{
-		cout << "Gas used: " << res.gasUsed << " (+" << t.baseGasRequired(se->evmSchedule(envInfo)) << " for transaction, -" << res.gasRefunded << " refunded)" << endl;
+		cout << "Gas used: " << res.gasUsed << " (+" << t.baseGasRequired(se->evmSchedule(envInfo.number())) << " for transaction, -" << res.gasRefunded << " refunded)" << endl;
```
winsvega: will this affect VMTests ?

gumb0: Definitely not, this change is just giving `evmSchedule` the only thing it needs - block number instead of the whole EnvInfo

libethcore/SealEngine.h
```diff
@@ -75,7 +75,7 @@ class SealEngineFace
 	ChainOperationParams const& chainParams() const { return m_params; }
 	void setChainParams(ChainOperationParams const& _params) { m_params = _params; }
 	SealEngineFace* withChainParams(ChainOperationParams const& _params) { setChainParams(_params); return this; }
-	virtual EVMSchedule const& evmSchedule(EnvInfo const&) const = 0;
+	virtual EVMSchedule const& evmSchedule(u256 const&) const = 0;
```
chfast: Name the parameter?

libethcore/SealEngine.h
```diff
@@ -105,7 +105,7 @@ class SealEngineBase: public SealEngineFace
 			m_onSealGenerated(ret.out());
 	}
 	void onSealGenerated(std::function<void(bytes const&)> const& _f) override { m_onSealGenerated = _f; }
-	EVMSchedule const& evmSchedule(EnvInfo const&) const override;
+	EVMSchedule const& evmSchedule(u256 const&) const override;
```
chfast: Name the parameter?

libevm/ExtVMFace.h
```diff
@@ -228,30 +227,32 @@ class EnvInfo
 {
 public:
 	EnvInfo() {}
-	EnvInfo(BlockHeader const& _current, LastHashes const& _lh = LastHashes(), u256 const& _gasUsed = u256()):
+	EnvInfo(BlockHeader const& _current, std::shared_ptr<LastBlockHashesFace const> _lh, u256 const& _gasUsed):
```
chfast: I can't to figure out the reason this is the shared_ptr.

gumb0: I tried to get rid of it and pass `LastBlockHashesFace const&` around, but I gave up because some test code need setting & resetting this interface into `EnvInfo` (calling `EnvInfo::setLastHashes()`)

gumb0: This code to be exact: https://github.com/ethereum/cpp-ethereum/pull/4126/files#diff-e01cd53eebd917cc35354ffc7719d8adR295

Probably there's a way to change this test code somehow... modifying ´EnvInfo´ is weird in general

libethashseal/Ethash.h
gcolvin: I am happy to see these dependencies go.

libevm/ExtVMFace.h
gcolvin: I don't like the shared_ptr here either, but can't tell you how to fix it.

libevm/ExtVMFace.h
```diff
@@ -228,30 +227,32 @@ class EnvInfo
...
 
 	void setNumber(u256 const& _v) { m_headerInfo.setNumber(_v); }
 	void setAuthor(Address const& _v) { m_headerInfo.setAuthor(_v); }
 	void setTimestamp(u256 const& _v) { m_headerInfo.setTimestamp(_v); }
 	void setDifficulty(u256 const& _v) { m_headerInfo.setDifficulty(_v); }
 	void setGasLimit(u256 const& _v) { m_headerInfo.setGasLimit(_v); }
-	void setLastHashes(LastHashes&& _lh) { m_lastHashes = _lh; }
-
+	void setLastHashes(std::shared_ptr<LastBlockHashesFace const> _lh) { m_lastHashes = _lh; }
+
```
pirapira: Spurious space.

chfast:
I like the direction it is going. And I also want less dependencies on test framework. Just can we use unique PTR instead?

gumb0:
@chfast Hm, making it `unique_ptr` means we'll construct the new instance of this `LastBlockHashes` implementation (https://github.com/ethereum/cpp-ethereum/pull/4126/files#diff-4bd28474d53dacc284d9432a1a785c10R128) on every tx execution. Which kind of defeats the purpose of caching the returned hashes vector in `LastBlockHashes::m_lastHashes`

gumb0:
I'm trying to replace `shared_ptr<LastBlockHashesFace const>` with `LastBlockHashesFace const&`.
If it doesn't work out, I'll go with `unique_ptr`

If it does work out, it probably better should go into separate PR

ethvm/main.cpp
```diff
@@ -274,6 +283,8 @@ int main(int argc, char** argv)
 	state.addBalance(sender, value);
 
 	unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(networkName)).createSealEngine());
+	LastBlockHashes lastBlockHashes;
+	EnvInfo const envInfo(blockHeader, lastBlockHashes, 0);
```
gumb0: Looks like previously ethvm didn't work for the code containing BLOCKHASH. Now any BLOCKHASH will return 0

gumb0:
I've replaced `shared_ptr<LastBlockHashes const>` with `LastBlockHashes const&` and made `EnvInfo` struct immutable (now when test code needs to set a new last hashes, it recreates `EnvInfo`)

Tests are green now, please review the changes @chfast @gcolvin @pirapira @winsvega 
(I somehow messed up making it a separate commit and it's all now in one forced-pushed commit, sorry)

libethereum/BlockChain.cpp
```diff
@@ -122,6 +122,42 @@ class WriteBatchNoter: public ldb::WriteBatch::Handler
...
+
+class LastBlockHashes: public LastBlockHashesFace
+{
+public:
+	explicit LastBlockHashes(BlockChain const& _bc): m_bc(_bc) {}
+
+	h256s precedingHashes(h256 const& _mostRecentHash) const override
+	{
+		Guard l(m_lastHashesMutex);
+		if (m_lastHashes.empty() || m_lastHashes.back() != _mostRecentHash)
```
pirapira: Do you mean `front` instead of `back`?

In a following line I see `m_lastHashes[0] = _mostRecentHash`.

libethereum/BlockChain.cpp
```diff
@@ -392,19 +429,6 @@ string BlockChain::dumpDatabase() const
 	return ss.str();
 }
 
-LastHashes BlockChain::lastHashes(h256 const& _parent) const
-{
-	Guard l(x_lastLastHashes);
-	if (m_lastLastHashes.empty() || m_lastLastHashes.back() != _parent)
```
pirapira: I see the `back` is inherited from here.

libethereum/BlockChain.cpp
gumb0: Yeah, looks like it was incorrect in the original code, nice catch!

PR #4139, opened 2017-06-12
update the source of the snap package

elopio:
Now that the snapcraft.yaml was moved to the root of the project, the source needs to be updated too.


elopio:
Thanks for the merge @chfast!
On our side, to get this working on build.snapcraft.io we need to fix https://github.com/canonical-websites/build.snapcraft.io/issues/497

PR #4142, opened 2017-06-13
add filter option to getJsonFiles function

winsvega:


test/tools/libtesteth/TestHelper.cpp
```diff
@@ -198,13 +198,18 @@ void replaceLLLinState(json_spirit::mObject& _o)
...
+std::vector<boost::filesystem::path> getJsonFiles(std::string const& _dirPath, std::string const& _pattern)
 {
 	vector<boost::filesystem::path> jsonFiles;
 	using Bdit = boost::filesystem::directory_iterator;
 	for (Bdit it(_dirPath); it != Bdit(); ++it)
+	{
 		if (boost::filesystem::is_regular_file(it->path()) && it->path().extension() == ".json")
-			jsonFiles.push_back(it->path());
+		{
+			if (_pattern.empty() || it->path().filename().string().find(_pattern.c_str()) != std::string::npos)
```
chfast: 1. You don't need `.c_str()` here.
2. Can you split the line. It is hard to read.

test/tools/libtesteth/TestHelper.h
```diff
@@ -121,7 +121,7 @@ class ZeroGasPricer: public eth::GasPricer
 };
 
 // helping functions
-std::vector<boost::filesystem::path> getJsonFiles(std::string const& _dirPath);
+std::vector<boost::filesystem::path> getJsonFiles(std::string const& _dirPath, std::string const& _pattern = std::string());
```
chfast: `= {}` should be enough.

test/tools/libtesteth/TestHelper.cpp
gumb0: If this function is supposed to return only one file when `_pattern` is non-empty, it's better in this case just to check if the file is there instead of iterating the directory.

pirapira:
Why is the linker complaining in the Travis build?

winsvega:
strange it was fine. maybe retrigger

pirapira:
I did restart.

pirapira:
In Travis, `ld.gold` is not happy.

pirapira:
@chfast is this good now?

test/tools/jsontests/StateTests.cpp
```diff
@@ -117,7 +117,8 @@ class generaltestfixture
 	{
 		std::string fillersPath = test::getTestPath() + "/src/GeneralStateTestsFiller/" + _folder;
 
-		std::vector<boost::filesystem::path> files = test::getJsonFiles(fillersPath);
+		string filter = test::Options::get().singleTestName.empty() ? string() : test::Options::get().singleTestName + "Filler";
```
chfast: For consistency you should use `std::` prefix here.

PR #4151, opened 2017-06-15
Revert bug fix

### LINKED ISSUE #4130: revert bug
We discovered that cpp does wrong revert of the state in case of address collision. 

The test is here (with wrong cpp results):
https://github.com/ethereum/tests/blob/cffb263d6c9075e1396e72d5d8febc21329b6600/BlockchainTests/GeneralStateTests/stRevertTest/RevertDepthCreateAddressCollision.json#L642-L665

The source for this test is here:
https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stRevertTest/RevertDepthCreateAddressCollisionFiller.json#L188-L205

It appears to be that cpp won't restore the code of the original account after collision overwrite and then OOG revert. (case d0g0 - double oog). Python is failing the test.

pirapira:
This fixes #4130 .

Before Metropolis, when an address collision happens, an existing non-empty code could be overwritten.  The journaling system didn't remember the old code, so it could not recover the old code.  This PR fixes that problem.

### DIFF
--- a/libethereum/State.cpp
+++ b/libethereum/State.cpp
@@ -450,8 +450,8 @@ bytes const& State::code(Address const& _addr) const
 
 void State::setNewCode(Address const& _address, bytes&& _code)
 {
+	m_changeLog.emplace_back(_address, code(_address));
 	m_cache[_address].setNewCode(std::move(_code));
-	m_changeLog.emplace_back(Change::NewCode, _address);
 }
 
 h256 State::codeHash(Address const& _a) const
@@ -512,7 +512,10 @@ void State::rollback(size_t _savepoint)
 			m_cache.erase(change.address);
 			break;
 		case Change::NewCode:
-			account.resetCode();
+			if (change.oldCode.empty())
+				account.resetCode();
+			else
+				account.setNewCode(std::move(change.oldCode));
 			break;
 		case Change::Touch:
 			account.untouch();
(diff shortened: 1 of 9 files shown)


libethereum/State.cpp
```diff
@@ -512,7 +512,10 @@ void State::rollback(size_t _savepoint)
 			m_cache.erase(change.address);
 			break;
 		case Change::NewCode:
-			account.resetCode();
+			if (change.oldCode.empty())
```
chfast: Here I would go with `account.setNewCode(std::move(change.oldCode));` all the time. Moving empty bytes is not more costly than reset.

pirapira: Fixed.

libethereum/State.cpp
```diff
@@ -512,10 +512,7 @@ void State::rollback(size_t _savepoint)
 			m_cache.erase(change.address);
 			break;
 		case Change::NewCode:
-			if (change.oldCode.empty())
-				account.resetCode();
-			else
-				account.setNewCode(std::move(change.oldCode));
+			account.setNewCode(std::move(change.oldCode));
```
chfast: Hm... after these changes it looks we should also rename `.setNewCode()` to just `.setCode()`.

pirapira: That's true.  I'll add a commit to this PR.

pirapira: Done.

chfast:
The best!

pirapira:
All green!

PR #4152, opened 2017-06-15
_info section in tests

winsvega:
this adds a new section to the tests 
```
        "_info" : {
            "comment" : "Simple test",
            "filledwith" : "cpp-1.3.0+commit.6be76b64.Linux.g++",
            "source" : "/src/GeneralStateTestsFiller/stExample/add11Filler.json"
        },
```

### DIFF
--- a/test/tools/libtesteth/TestHelper.cpp
+++ b/test/tools/libtesteth/TestHelper.cpp
@@ -216,7 +216,9 @@ std::string executeCmd(std::string const& _command)
 		BOOST_ERROR("Failed to run " + _command);
 	if (fgets(output, sizeof(output) - 1, fp) == NULL)
 		BOOST_ERROR("Reading empty result for " + _command);
-	pclose(fp);
+	int exitCode = pclose(fp);
+	if (exitCode != 0)
+		BOOST_ERROR("The command '" + _command + "' exited with " + toString(exitCode) + " code.");
 	return boost::trim_copy(string(output));
 }
 
@@ -468,11 +470,12 @@ void addClientInfo(json_spirit::mValue& _v, std::string const& _testSource)
 		if (o.count("_info"))
 		{
 			json_spirit::mObject& existingInfo = o["_info"].get_obj();
-			comment = existingInfo["comment"].get_str();
+			if (existingInfo.count("comment"))
+				comment = existingInfo["comment"].get_str();
 		}
 
 		//prepare the relative src path
-		string source = _testSource.substr(_testSource.find("/src/"), _testSource.length());
+		string source = _testSource.substr(_testSource.rfind("/src/"), _testSource.length());
 
 		clientinfo["filledwith"] = "cpp-ethereum rev-head: " + head;
 		clientinfo["source"] = source;
--- a/test/tools/libtesteth/TestHelper.cpp
+++ b/test/tools/libtesteth/TestHelper.cpp
@@ -208,6 +208,18 @@ std::vector<boost::filesystem::path> getJsonFiles(std::string const& _dirPath)
 	return jsonFiles;
 }
 
+std::string executeCmd(std::string const& _command)
+{
+	char output[1024];
+	FILE *fp = popen(_command.c_str(), "r");
+	if (fp == NULL)
+		BOOST_ERROR("Failed to run " + _command);
+	if (fgets(output, sizeof(output) - 1, fp) == NULL)
+		BOOST_ERROR("Reading empty result for " + _command);
+	pclose(fp);
+	return boost::trim_copy(string(output));
+}
+
 string compileLLL(string const& _code)
 {
 	if (_code == "")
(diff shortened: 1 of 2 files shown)


test/tools/libtesteth/TestHelper.cpp
```diff
@@ -208,6 +208,18 @@ std::vector<boost::filesystem::path> getJsonFiles(std::string const& _dirPath)
...
+{
+	char output[1024];
+	FILE *fp = popen(_command.c_str(), "r");
+	if (fp == NULL)
+		BOOST_ERROR("Failed to run " + _command);
+	if (fgets(output, sizeof(output) - 1, fp) == NULL)
+		BOOST_ERROR("Reading empty result for " + _command);
+	pclose(fp);
+	return boost::trim_copy(string(output));
+}
```
pirapira: Why is this function called `execute` while it just reads a string?

pirapira: Ah, no `popen` might mean "pipe" something...

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -208,6 +208,18 @@ std::vector<boost::filesystem::path> getJsonFiles(std::string const& _dirPath)
...
 
+std::string executeCmd(std::string const& _command)
+{
+	char output[1024];
+	FILE *fp = popen(_command.c_str(), "r");
+	if (fp == NULL)
+		BOOST_ERROR("Failed to run " + _command);
+	if (fgets(output, sizeof(output) - 1, fp) == NULL)
+		BOOST_ERROR("Reading empty result for " + _command);
+	pclose(fp);
```
pirapira: The return value of `pclose` should be checked to see if the command returned `0`.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -452,6 +456,31 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
...
 }
 
+void addClientInfo(json_spirit::mValue& _v, std::string const& _testSource)
+{
+	for (auto& i: _v.get_obj())
+	{
+		json_spirit::mObject& o = i.second.get_obj();
+		json_spirit::mObject clientinfo;
+
+		string head = executeCmd("git rev-parse HEAD");
```
pirapira: @chfast I guess this can be done nicely with `buildinfo.cmake` trick?

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -452,6 +456,31 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
...
+		string comment;
+		if (o.count("_info"))
+		{
+			json_spirit::mObject& existingInfo = o["_info"].get_obj();
+			comment = existingInfo["comment"].get_str();
```
pirapira: Maybe throw an error if "comment" does not exist.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -419,6 +422,7 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
 
 			json_spirit::read_string(s, v);
 			doTests(v, true);
+			addClientInfo(v, testfilename);
```
pirapira: I think `_fillerPathAppendix` should be passed into this function.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -452,6 +456,31 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
...
+		string head = executeCmd("git rev-parse HEAD");
+		string comment;
+		if (o.count("_info"))
+		{
+			json_spirit::mObject& existingInfo = o["_info"].get_obj();
+			comment = existingInfo["comment"].get_str();
+		}
+
+		//prepare the relative src path
+		string source = _testSource.substr(_testSource.find("/src/"), _testSource.length());
```
pirapira: This does not work with my setup.  My test fillers are in `~/src/tests/src/GeneralStateTestsFiller/stLogTests`.  The first `/src/` is something different from what's sought here.

test/tools/libtesteth/TestHelper.cpp
winsvega: this way it wont get the filename

pirapira: Then `_fillerPathAppendix + "/" + name + nameEnding`.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -452,6 +456,31 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
...
+
+		clientinfo["filledwith"] = "cpp-ethereum rev-head: " + head;
```
gumb0: Let's have only commit hash as value here, without "cpp-ethereum rev-head", this way it's easier parseble and wastes less disk space.
So something like
```
clientinfo["filledwith_rev"] = head;
```

test/tools/libtesteth/TestHelper.cpp
gumb0: Yeah, there should be a nicer way, I don't like launching a separate process for this, even less for doing it in the loop (launching a process for each file to get the same string)

gumb0: Turns out we already have this in file build/include/BuildInfo.h

test/tools/libtesteth/TestHelper.cpp
winsvega: rfind should work

pirapira: I don't fight for this.

winsvega:
we need at least one green build

PR #4154, opened 2017-06-16
The optimized code for GTX1060 and CMP4 (may also be useful for other Nvidian cards)

davilizh:
The performance is improved from 22579336 to 23907532 on a flashed GTX 1060 with 2 GPCs 9 TPCs (the product chip should have 10 TPCs), and from 32320941 to 34162453 on a Flashed CMP4 with tRRD=4. 
Optimizations include:
**1. ethash_cuda_miner_kernel.cu**
We have commented out "launch_bounds" in the code. launch_bound is discussed in http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#axzz4fzSzZc9p in detail.
**2. dagger_shuffle.cuh**
We moved around and reduced variable definitions to the minimum required. The compiler should have been able to do this analysis, but it never hurts to help out the compiler.
The state in compute_hash of dagger_shuffle.cuh is modified.
We simplify the nested if/else blocks into a switch statement.
We simplify control flow. Remove the conditional from the inner loop so all threads calculate the value, and then all threads use a __shfl to read thread t's value (throwing away the rest of the threads' calculated value).
We increase the total number of LDGs to increase occupancy. We define PARALLEL_HASH to let each warp have PARALLEL_HASH LDGs in-flight at a time, not 1 at a time, which is the original case.
Every thread is the master for calculating one hash value. Each thread initializes its version of state using keccak_f1600_init. Then in the main loop: When i=0 threads 0-7 copy the values of thread 0's state[0-7] into each threads' shuffle[0-7], do the main computation, and then thread 0 captures the result of shuffle[0-3] into state[8-11]. On the next loop when i=1 threads 0-7 copy the values of thread 1's state[0-7] into each threads' shuffle[0-7], do the main computation, and then thread 1 captures the result of shuffle[0-3] into state[8-11].
With the modification this is changed so that if PARALLEL_HASH=2: When i=0 threads 0-7 copy the values of thread 0's state[0-7] into each threads' shuffle[0][0-7] and thread 1's state[0-7] into each threads' shuffle[1][0-7]. They do the main computation on these 2 shuffle vectors in parallel. Then thread 0 captures the result of shuffle[0][0-3] into its state[8-11] and thread 1 captures the result of shuffle[1][0-3] into its state[8-11].
**3. keccak.cuh**
Since the input argument uint2 *s is changed in dagger_shuffle.cuh, we have to modify keccak_f1600_init and keccak_f1600_final in keccak.cuh accordingly.


davilizh:
Sorry, I made a wrong command. Can anyone help to delete it?

PR #4166, opened 2017-06-19
Testeth t option before separator

### LINKED ISSUE #4163: testeth should check -t '.../...' appears before the separator '--'

pirapira:
This fixes #4163 and does a little bit more.


winsvega:
boost already will throw if a custom option comes before --

pirapira:
That simplifies a lot.

PR #4167, opened 2017-06-19
[WIP] Removing ccache to see if that fixes the build failure

pirapira:
If that fixes the build, I will further try upgrading the ccache version.


pirapira:
The build still fails.  So it's not about ccache.

PR #4172, opened 2017-06-19
EIP86 - CREATE2

yann300:


libethereum/Executive.cpp
```diff
@@ -306,9 +306,30 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
 	return !m_ext;
 }
 
-bool Executive::create(Address _sender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin)
+bool Executive::create(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _init, Address _origin)
```
pirapira: Maybe use `u256 const&` because now the arguments are passed deeper and deeper.  Also `Address const&`.

gumb0:
Please rebase to fix linux build on travis

winsvega:
should work now. 
we could merge this. and then add more tests for CREATE2, ZeroTransaction and CreateTransactions

libethereum/Executive.cpp
```diff
@@ -363,9 +363,9 @@ bool Executive::executeCreate(Address _sender, u256 _endowment, u256 _gasPrice,
 	}
 	else
 	{
-		if ((m_s.addressHasCode(m_newAddress) || m_s.getNonce(m_newAddress) != 0))
+		if ((m_s.addressHasCode(m_newAddress) || m_s.getNonce(m_newAddress) > 1))
```
gumb0: @winsvega  I don't get this change, this means we won't fail when nonce == 1 ?

libethereum/Executive.cpp
```diff
@@ -331,12 +352,25 @@ bool Executive::create(Address _sender, u256 _endowment, u256 _gasPrice, u256 _g
...
+	{
+		if (m_s.addressHasCode(m_newAddress))
+			// Overwrite with empty code in case the account already has a code
+			// (address collision -- not real life case but we can check it with
+			// synthetic tests).
+			m_s.setCode(m_newAddress, {});
+	}
+	else
+	{
+		if ((m_s.addressHasCode(m_newAddress) || m_s.getNonce(m_newAddress) > 1))
```
gumb0: @winsvega I don't get this change, this means we won't fail when nonce == 1 ?

test/tools/jsontests/vm.cpp
```diff
@@ -38,9 +39,28 @@ FakeExtVM::FakeExtVM(EnvInfo const& _envInfo, unsigned _depth):			/// TODO: XXX:
 	ExtVMFace(_envInfo, Address(), Address(), Address(), 0, 1, bytesConstRef(), bytes(), EmptySHA3, false, _depth)
 {}
 
-std::pair<h160, eth::owning_bytes_ref> FakeExtVM::create(u256 _endowment, u256& io_gas, bytesConstRef _init, OnOpFunc const&)
+std::pair<h160, eth::owning_bytes_ref> FakeExtVM::create(u256 _endowment, u256& io_gas, bytesConstRef _init, Instruction _op, u256 salt, OnOpFunc const&)
 {
-	Address na = right160(sha3(rlpList(myAddress, get<1>(addresses[myAddress]))));
+	unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(eth::Network::MainNetworkTest)).createSealEngine());
```
gumb0: This change should be removed, nobody needs it, `FakeExtVM` is used only with frontier rules

libethereum/Executive.cpp
winsvega: hmm. thats strange.

winsvega: I see so here is the issue. the new address is being created with default nonce = 1. 
then it is being checked for existance. this is not a reliable way to check that the address is already exist.

pirapira: The EIP [says](https://github.com/ethereum/EIPs/pull/208/files#diff-95b306e6e6d2d6dbb0c5fd3ac4ea45c4R28), check for `hasCode || nonce != 0`.

(OK, it says "non-empty nonce" which is weird.)

winsvega: the EIP is correct. however at that point in code the adress is somehow already present at the state
as if we create an address then check if its existing

gumb0: Ah I see, we've already incremented the nonce on line 350

gumb0: Then I guess this check should better be done before nonce incrementation. Probably just after incrementing sender's nonce

libethereum/Executive.cpp
```diff
@@ -331,12 +353,25 @@ bool Executive::create(Address _sender, u256 _endowment, u256 _gasPrice, u256 _g
...
+			// synthetic tests).
+			m_s.setCode(m_newAddress, {});
+	}
+	else
+	{
+		if (accountAlreadyExist)
+		{
+			clog(StateSafeExceptions) << "Address already used: " << m_newAddress;
+			m_gas = 0;
+			m_excepted = TransactionException::AddressAlreadyUsed;
```
pirapira: Should we also clear `m_newAddress`?

winsvega: I guess revert do the logic

pirapira: Yes it does.

libethereum/Executive.cpp
```diff
@@ -331,12 +353,25 @@ bool Executive::create(Address _sender, u256 _endowment, u256 _gasPrice, u256 _g
...
+			revert();
```
pirapira: In this case, I think true must be returned from this function, indicating that the VM execution is not necessary.

libethereum/Executive.h
```diff
@@ -150,10 +149,12 @@ class Executive
 
 	/// Set up the executive for evaluating a bare CREATE (contract-creation) operation.
 	/// @returns false iff go() must be called (and thus a VM execution in required).
-	bool create(Address _txSender, u256 _endowment, u256 _gasPrice, u256 _gas, bytesConstRef _code, Address _originAddress);
+	bool create(Address const& _txSender, u256 const& _endowment, u256 const& _gasPrice, u256 const& _gas, bytesConstRef _code, Address const& _originAddress);
+	bool createOpcode(Address const& _sender, u256 const& _endowment, u256 const& _gasPrice, u256 const& _gas, bytesConstRef _code, Address const& _originAddress);
+	bool create2Opcode(Address const& _sender, u256 const& _endowment, u256 const& _gasPrice, u256 const& _gas, bytesConstRef _code, Address const& _originAddress, u256 const& _salt);
```
pirapira: Please document the return value of the two new functions (it is not obvious).

libethereum/Executive.h
```diff
@@ -183,6 +184,8 @@ class Executive
 	void revert();
 
 private:
+	bool executeCreate(Address const& _txSender, u256 const& _endowment, u256 const& _gasPrice, u256 const& _gas, bytesConstRef _code, Address const& _originAddress);
```
pirapira: Please document the return value.

libethereum/Executive.cpp
winsvega: ask @yann300  about this

pirapira: I made the change.

pirapira:
@winsvega please review my commit.

Issue #4159, opened 2017-06-16
No rule to make target '../deps/lib/libmpir.a'


moneroexamples:
The issue is based on this issue: https://github.com/moneroexamples/compile-cpp-ethereum-on-arch/issues/1

problem on arch/manjaro:

```
make[2]: *** No rule to make target '../deps/lib/libmpir.a', needed by 'rlp/rlp'.  Stop.
make[2]: *** Waiting for unfinished jobs....
make[2]: *** No rule to make target '../deps/lib/libmpir.a', needed by 'bench/bench'.  Stop.
make[2]: *** Waiting for unfinished jobs....
```

And this is copy and paste of the comment explaining how manually it can be fixed:

The problem is that `libmpir.a` is in `deps/lib64` rather than in `deps/lib`. So the compilation blows up as its looking for `libmpir.a` in wrong folder.

I will be making issues on cpp-ethereum about this  shortly. In the meantime, have to manually fix it:

edit file `~/cpp-ethereum/cmake/ProjectMPIR.cmake`

and change: 
```
set(MPIR_LIBRARY ${INSTALL_DIR}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}mpir${CMAKE_STATIC_LIBRARY_SUFFIX})
```
into 

```
set(MPIR_LIBRARY ${INSTALL_DIR}/lib64/${CMAKE_STATIC_LIBRARY_PREFIX}mpir${CMAKE_STATIC_LIBRARY_SUFFIX})
```

and re-compile.

mariouriarte:
Thank you, it helped me a lot

SingingBush:
@moneroexamples thanks. after pulling the changes for #4115 I ran into this problem also. changing the cmake file was the last piece of the puzzle for compiling on Fedora 26.

moneroexamples:
@SingingBush 

Glad it helped.

PR #4173, opened 2017-06-19
Fix MPIR project configuration

chfast:
Fixes https://github.com/ethereum/cpp-ethereum/issues/4159.


goblin:
This actually breaks the build for me, I got:

```
make[2]: *** No rule to make target '../deps/lib/x86_64-linux-gnu/libmpir.a', needed by 'eth/eth'.  Stop.
```

For me, `libmpir.a` ended up in `../deps/lib/`, not `../deps/lib/x86_64-linux-gnu/`.

I'm not sure how I ran cmake, since it was done automatically by debhelper. To reproduce, on Debian Stretch, issue `dh_make` to build a stub debian package, and then use `debuild -uc -us` to try and build it.

I also had to make a similar change in `cmake/ProjectCryptopp.cmake` (line 171, probably also 169, but I'm not using MSVC).

PR #4178, opened 2017-06-20
snark: Add ECADD and ECMUL benchmark tests

chfast:


pirapira:
I think I'm not getting in the way of #4201.

PR #4180, opened 2017-06-20
testeth: catch InvalidOption exception into a message and exit(1)

pirapira:
The indentation change confused the diff but I just added a `try` ... `catch`.

### DIFF
--- a/test/tools/libtesteth/boostTest.cpp
+++ b/test/tools/libtesteth/boostTest.cpp
@@ -38,6 +38,7 @@
 
 #include <clocale>
 #include <stdlib.h>
+#include <iostream>
 #include <test/tools/libtesteth/TestHelper.h>
 #include <boost/version.hpp>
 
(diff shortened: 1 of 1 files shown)


test/tools/libtesteth/boostTest.cpp
```diff
@@ -117,37 +118,44 @@ void setDefaultOrCLocale()
...
-		framework::master_test_suite().add(ts1);
+			//add random tests suite
+			test_suite* ts1 = BOOST_TEST_SUITE("RandomTestCreationSuite");
+			ts1->add(BOOST_TEST_CASE(&createRandomTest));
+			framework::master_test_suite().add(ts1);
+		}
+	}
+	catch (dev::test::Options::InvalidOption const& e)
+	{
+		std::cerr << e.what() << std::endl << std::flush;
```
gumb0: `std::endl` does the flushing itself, no need for `std::flush` 😄

pirapira: How many keystrokes have I wasted in my life...

Issue #4085, opened 2017-05-19
Improve Block unit tests


gumb0:
We could make the tests in test/unittests/libethereum/Block.cpp use NoProof `SealEngine` instead of regular Ethash (by using special config instead of usual `c_genesisInfoFrontierTest`)

Then they would not require full ethash mining, will be faster and we could enable for usual Travis test run the disabled ones.

PR #4182, opened 2017-06-21
Run Block unit tests without PoW-mining

gumb0:
Resloves https://github.com/ethereum/cpp-ethereum/issues/4085


winsvega:
could we add a new field to genesis.json config file  that is specifying minimum difficulty for the mining? 
that way we could make all blockchain test work much faster

PR #4185, opened 2017-06-21
testeth errors when an odd-length hex string encodes a byte sequence

pirapira:
For `code` and `data`, `testeth` now expects an even-length hex strings.

Whenever a sequence of bytes is encoded as hex, the result's length is even.

### DIFF
--- a/test/tools/libtesteth/TestHelper.cpp
+++ b/test/tools/libtesteth/TestHelper.cpp
@@ -261,7 +261,7 @@ string compileLLL(string const& _code)
 void checkHexHasEvenLength(string const& _str)
 {
 	if (_str.size() % 2)
-		BOOST_ERROR("An odd-length hex string represents a byte sequence: " + _str);
+		BOOST_ERROR(TestOutputHelper::testName() + " An odd-length hex string represents a byte sequence: " + _str);
 }
 
 bytes importCode(json_spirit::mObject& _o)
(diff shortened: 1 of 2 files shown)


test/tools/libtesteth/TestHelper.cpp
```diff
@@ -249,18 +253,25 @@ string compileLLL(string const& _code)
...
 	result = "0x" + result;
+	checkHexHasEvenLength(result);
 	return result;
 #endif
 }
 
+void checkHexHasEvenLength(string const& _str)
+{
+	if (_str.size() % 2)
+		BOOST_ERROR("An odd-length hex string represents a byte sequence: " + _str);
```
winsvega: you should use TestOutputHelper::  to output current test case and test name

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -249,18 +253,25 @@ string compileLLL(string const& _code)
...
+
 bytes importCode(json_spirit::mObject& _o)
 {
 	bytes code;
 	if (_o["code"].type() == json_spirit::str_type)
 		if (_o["code"].get_str().find("0x") != 0)
 			code = fromHex(compileLLL(_o["code"].get_str()));
 		else
-			code = fromHex(_o["code"].get_str().substr(2));
+			code = importByteArray(_o["code"].get_str());
```
winsvega: what about the case above? it still uses fromHex

winsvega:
so there was no error after refilling corrected pairing test ?

test/tools/libtesteth/TestHelper.cpp
pirapira: In the above case, `compileLLL` checks the length of the returned string.

pirapira:
@winsvega there were some other errors, but I fixed all of them in https://github.com/ethereum/tests/pull/190

PR #4195, opened 2017-06-22
revive randomcode option

winsvega:
usage:
``` ./testeth -- --randomcode <int>  ```

generates smart random EVM code

### DIFF
--- a/test/tools/fuzzTesting/fuzzHelper.cpp
+++ b/test/tools/fuzzTesting/fuzzHelper.cpp
@@ -26,6 +26,24 @@
 #include <test/tools/fuzzTesting/fuzzHelper.h>
 #include <test/tools/libtesteth/TestOutputHelper.h>
 
+using namespace dev;
+const static std::array<eth::Instruction, 14> invalidOpcodes{
+	eth::Instruction::INVALID,
+	eth::Instruction::PUSHC,
+	eth::Instruction::JUMPC,
+	eth::Instruction::JUMPCI,
+	eth::Instruction::JUMPTO,
+	eth::Instruction::JUMPIF,
+	eth::Instruction::JUMPSUB,
+	eth::Instruction::JUMPV,
+	eth::Instruction::JUMPSUBV,
+	eth::Instruction::BEGINSUB,
+	eth::Instruction::BEGINDATA,
+	eth::Instruction::RETURNSUB,
+	eth::Instruction::PUTLOCAL,
+	eth::Instruction::GETLOCAL
+};
+
 namespace dev
 {
 namespace test
@@ -43,12 +61,6 @@ boostIntGenerator RandomCode::randOpLengGen = boostIntGenerator(gen, opLengDist)
 boostIntGenerator RandomCode::randOpMemrGen = boostIntGenerator(gen, opMemrDist);
 boostIntGenerator RandomCode::randUniIntGen = boostIntGenerator(gen, uniIntDist);
 boostUInt64Generator RandomCode::randUInt64Gen = boostUInt64Generator(gen, uInt64Dist);
-
-const std::array<eth::Instruction, 14> invalidOpcodes{ eth::Instruction::INVALID,
-		eth::Instruction::PUSHC,eth::Instruction::JUMPC,eth::Instruction::JUMPCI,eth::Instruction::JUMPTO,
-		eth::Instruction::JUMPIF,eth::Instruction::JUMPSUB,eth::Instruction::JUMPV,eth::Instruction::JUMPSUBV,
-		eth::Instruction::BEGINSUB,eth::Instruction::BEGINDATA,eth::Instruction::RETURNSUB,eth::Instruction::PUTLOCAL,
-		eth::Instruction::GETLOCAL};
 
 int RandomCode::recursiveRLP(std::string& _result, int _depth, std::string& _debug)
 {
--- a/test/tools/fuzzTesting/fuzzHelper.cpp
+++ b/test/tools/fuzzTesting/fuzzHelper.cpp
@@ -22,6 +22,7 @@
 #include <chrono>
 #include <boost/random.hpp>
 #include <boost/filesystem/path.hpp>
+#include <boost/assign/list_of.hpp>
 #include <libevmcore/Instruction.h>
 #include <test/tools/fuzzTesting/fuzzHelper.h>
 #include <test/tools/libtesteth/TestOutputHelper.h>
@@ -41,6 +42,12 @@ boostIntGenerator RandomCode::randOpCodeGen = boostIntGenerator(gen, opCodeDist)
 boostIntGenerator RandomCode::randOpLengGen = boostIntGenerator(gen, opLengDist);
 boostIntGenerator RandomCode::randUniIntGen = boostIntGenerator(gen, uniIntDist);
 boostUInt64Generator RandomCode::randUInt64Gen = boostUInt64Generator(gen, uInt64Dist);
+
+const std::vector<eth::Instruction> RandomCode::invalidOpcodes = boost::assign::list_of(eth::Instruction::INVALID) \
+		(eth::Instruction::PUSHC) (eth::Instruction::JUMPC) (eth::Instruction::JUMPCI) (eth::Instruction::JUMPTO) \
+		(eth::Instruction::JUMPIF) (eth::Instruction::JUMPSUB) (eth::Instruction::JUMPV) (eth::Instruction::JUMPSUBV) \
+		(eth::Instruction::BEGINSUB) (eth::Instruction::BEGINDATA) (eth::Instruction::RETURNSUB) (eth::Instruction::PUTLOCAL) \
+		(eth::Instruction::GETLOCAL);
 
 int RandomCode::recursiveRLP(std::string& _result, int _depth, std::string& _debug)
 {
@@ -208,7 +215,7 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
 	std::string code;
 
 	//random opCode amount
-	boostIntDistrib sizeDist (0, _maxOpNumber);
+	boostIntDistrib sizeDist (1, _maxOpNumber);
 	boostIntGenerator rndSizeGen(gen, sizeDist);
 	int size = (int)rndSizeGen();
 
@@ -218,13 +225,21 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
 	for (auto i = 0; i < size; i++)
 	{
 		uint8_t opcode = weightsDefined ? randOpCodeWeight() : randOpCodeGen();
-		dev::eth::InstructionInfo info = dev::eth::instructionInfo((dev::eth::Instruction) opcode);
+		eth::Instruction inst = (eth::Instruction) opcode;
+		eth::InstructionInfo info = eth::instructionInfo(inst);
 
-		if (info.name.find("INVALID_INSTRUCTION") != std::string::npos)
+		if (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
+			|| std::find(invalidOpcodes.begin(), invalidOpcodes.end(), inst) != invalidOpcodes.end())
 		{
-			//Byte code is yet not implemented
-			if (_options.useUndefinedOpCodes == false)
+			if (_options.useUndefinedOpCodes == true)
 			{
+				std::string byte = toCompactHex(opcode);
+				code += (byte == "") ? "00" : byte;
+				std::cerr << info.name << std::endl;
+			}
+			else
+			{
+				//Byte code is yet not implemented. do not count it.
 				i--;
 				continue;
 			}
@@ -233,13 +248,13 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
 		{
 			if (info.name.find("PUSH") != std::string::npos)
 				code += toCompactHex(opcode);
-			code += fillArguments((dev::eth::Instruction) opcode, _options);
-		}
+			code += fillArguments((eth::Instruction) opcode, _options);
 
-		if (info.name.find("PUSH") == std::string::npos)
-		{
-			std::string byte = toCompactHex(opcode);
-			code += (byte == "") ? "00" : byte;
+			if (info.name.find("PUSH") == std::string::npos)
+			{
+				std::string byte = toCompactHex(opcode);
+				code += (byte == "") ? "00" : byte;
+			}
 		}
 	}
 	return code;
@@ -273,9 +288,12 @@ void RandomCode::refreshSeed()
 
 std::string RandomCode::getPushCode(std::string const& _hex)
 {
-	int length = _hex.length() / 2;
+	std::string hexVal = _hex;
+	if (hexVal.empty())
+		hexVal = "00";
+	int length = hexVal.length() / 2;
 	int pushCode = 96 + length - 1;
-	return toCompactHex(pushCode) + _hex;
+	return toCompactHex(pushCode) + hexVal;
 }
 
 std::string RandomCode::getPushCode(int _value)
@@ -284,13 +302,13 @@ std::string RandomCode::getPushCode(int _value)
 	return getPushCode(hexString);
 }
 
-std::string RandomCode::fillArguments(dev::eth::Instruction _opcode, RandomCodeOptions const& _options)
+std::string RandomCode::fillArguments(eth::Instruction _opcode, RandomCodeOptions const& _options)
 {
-	dev::eth::InstructionInfo info = dev::eth::instructionInfo(_opcode);
+	eth::InstructionInfo info = eth::instructionInfo(_opcode);
 
 	std::string code;
 	bool smart = false;
-	unsigned num = info.args;
+	unsigned argsNum = info.args;
 	int rand = randUniIntGen() % 100;
 	if (rand < _options.smartCodeProbability)
 		smart = true;
@@ -298,24 +316,24 @@ std::string RandomCode::fillArguments(dev::eth::Instruction _opcode, RandomCodeO
 	if (smart)
 	{
 		//PUSH1 ... PUSH32
-		if (dev::eth::Instruction::PUSH1 <= _opcode && _opcode <= dev::eth::Instruction::PUSH32)
+		if (eth::Instruction::PUSH1 <= _opcode && _opcode <= eth::Instruction::PUSH32)
 		{
-		  code += rndByteSequence(int(_opcode) - int(dev::eth::Instruction::PUSH1) + 1);
+		  code += rndByteSequence(int(_opcode) - int(eth::Instruction::PUSH1) + 1);
 		  return code;
 		}
 
 		//SWAP1 ... SWAP16 || DUP1 ... DUP16
-		bool isSWAP = (dev::eth::Instruction::SWAP1 <= _opcode && _opcode <= dev::eth::Instruction::SWAP16);
-		bool isDUP = (dev::eth::Instruction::DUP1 <= _opcode && _opcode <= dev::eth::Instruction::DUP16);
+		bool isSWAP = (eth::Instruction::SWAP1 <= _opcode && _opcode <= eth::Instruction::SWAP16);
+		bool isDUP = (eth::Instruction::DUP1 <= _opcode && _opcode <= eth::Instruction::DUP16);
 
 		if (isSWAP || isDUP)
 		{
 			int times = 0;
 			if (isSWAP)
-				times = int(_opcode) - int(dev::eth::Instruction::SWAP1) + 2;
+				times = int(_opcode) - int(eth::Instruction::SWAP1) + 2;
 			else
 			if (isDUP)
-				times = int(_opcode) - int(dev::eth::Instruction::DUP1) + 1;
+				times = int(_opcode) - int(eth::Instruction::DUP1) + 1;
 
 			for (int i = 0; i < times; i ++)
 				code += getPushCode(randUniIntGen() % 32);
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: test/tools/fuzzTesting/fuzzHelper.cpp [hunk 1]
```diff
@@ -42,6 +43,12 @@ boostIntGenerator RandomCode::randOpLengGen = boostIntGenerator(gen, opLengDist)
 boostIntGenerator RandomCode::randUniIntGen = boostIntGenerator(gen, uniIntDist);
 boostUInt64Generator RandomCode::randUInt64Gen = boostUInt64Generator(gen, uInt64Dist);
 
+const std::vector<eth::Instruction> RandomCode::invalidOpcodes = boost::assign::list_of(eth::Instruction::INVALID) \
```
gumb0: Why not just initialize with an initializer list in braces?
`const std::vector<eth::Instruction> RandomCode::invalidOpcodes{eth::Instruction::INVALID, eth::Instruction::PUSHC, ...};`

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -218,13 +225,21 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
...
+		eth::Instruction inst = (eth::Instruction) opcode;
+		eth::InstructionInfo info = eth::instructionInfo(inst);
 
-		if (info.name.find("INVALID_INSTRUCTION") != std::string::npos)
+		if (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
+			|| std::find(invalidOpcodes.begin(), invalidOpcodes.end(), inst) != invalidOpcodes.end())
 		{
-			//Byte code is yet not implemented
-			if (_options.useUndefinedOpCodes == false)
+			if (_options.useUndefinedOpCodes == true)
```
gumb0: Just `if (_options.useUndefinedOpCodes)` is simpler

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -208,7 +215,7 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
 	std::string code;
 
 	//random opCode amount
-	boostIntDistrib sizeDist (0, _maxOpNumber);
+	boostIntDistrib sizeDist (1, _maxOpNumber);
```
gumb0: We don't want to generate code with `STOP` ?

winsvega: we don't want to generate empty code. this is about how many opcode will be generated

gumb0: Ah yeah, got it

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -233,13 +248,13 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
 		{
 			if (info.name.find("PUSH") != std::string::npos)
 				code += toCompactHex(opcode);
-			code += fillArguments((dev::eth::Instruction) opcode, _options);
-		}
+			code += fillArguments((eth::Instruction) opcode, _options);
```
gumb0: You've already converted `opcode` into `Instruction` in `inst` variable

test/tools/fuzzTesting/fuzzHelper.cpp
gumb0: Maybe these are not invalid, but "skipped" or "unsupported"?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -218,13 +225,21 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
...
-		if (info.name.find("INVALID_INSTRUCTION") != std::string::npos)
+		if (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
+			|| std::find(invalidOpcodes.begin(), invalidOpcodes.end(), inst) != invalidOpcodes.end())
 		{
-			//Byte code is yet not implemented
-			if (_options.useUndefinedOpCodes == false)
+			if (_options.useUndefinedOpCodes == true)
+			{
+				std::string byte = toCompactHex(opcode);
+				code += (byte == "") ? "00" : byte;
```
gumb0: IIRC the last parameter for `toCompactHex` already does this - pass 2 there to get always at least 2 symbols

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: Greg implemented those. but it is only exist in cpp 
https://github.com/ethereum/cpp-ethereum/pull/4111

gumb0: I know, that's what I'm saying - they are not really invalid, but not supported by the protocol yet

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -218,13 +225,21 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
...
+				std::cerr << info.name << std::endl;
```
gumb0: No error happened, maybe log to some other channel, not to `stderr` ?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -367,27 +405,61 @@ std::string RandomCode::fillArguments(dev::eth::Instruction _opcode, RandomCodeO
 }
 
 
-//Ramdom Code Options
-RandomCodeOptions::RandomCodeOptions() : useUndefinedOpCodes(false), smartCodeProbability(50)
+//Default Ramdom Code Options
```
gumb0: Random

test/tools/fuzzTesting/fuzzHelper.cpp
gumb0: Don't we still need to `fillArguments` in this case?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -284,38 +302,38 @@ std::string RandomCode::getPushCode(int _value)
...
 		smart = true;
 
 	if (smart)
 	{
 		//PUSH1 ... PUSH32
-		if (dev::eth::Instruction::PUSH1 <= _opcode && _opcode <= dev::eth::Instruction::PUSH32)
+		if (eth::Instruction::PUSH1 <= _opcode && _opcode <= eth::Instruction::PUSH32)
 		{
-		  code += rndByteSequence(int(_opcode) - int(dev::eth::Instruction::PUSH1) + 1);
+		  code += rndByteSequence(int(_opcode) - int(eth::Instruction::PUSH1) + 1);
```
gumb0: Indent by tab

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -357,7 +395,7 @@ std::string RandomCode::fillArguments(dev::eth::Instruction _opcode, RandomCodeO
 	}
 
 	if (smart == false)
```
gumb0: This line should better be `else`

winsvega: it should not. if smart opcode is not defined in switch then smart = false again

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: where?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -325,29 +343,49 @@ std::string RandomCode::fillArguments(dev::eth::Instruction _opcode, RandomCodeO
 
 		switch (_opcode)
 		{
-		case dev::eth::Instruction::CREATE:
+		case eth::Instruction::EXTCODECOPY:
+			code += getPushCode(randUniIntGen() % 32);  //memstart2
```
gumb0: Would be much better to define special distributions (like maybe `MemAddressDist`, `MemLenDist`) and generators (`MemAddressGen`, `MemLemGen`) on top of the file and then not repeat `% 32` and `% 128` many times

(also this modulo operation results in non-uniform distribution)

test/tools/fuzzTesting/fuzzHelper.cpp
gumb0: Spaces are currenly used for indentation here

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: addresses are either random hex or taken from the possible adresses array.

test/tools/fuzzTesting/fuzzHelper.cpp
gumb0: Ah yeah, I see, weird

test/tools/fuzzTesting/fuzzHelper.cpp
gumb0: Random memory address is generated here.
What I'm asking is make it look like `code += getPushCode(randMemAddresstGen());  //memstart2`

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -367,27 +405,61 @@ std::string RandomCode::fillArguments(dev::eth::Instruction _opcode, RandomCodeO
...
+	setWeight(eth::Instruction::SSTORE, 70);
+	setWeight(eth::Instruction::CALL, 70);
+	setWeight(eth::Instruction::CALLCODE, 70);
+	setWeight(eth::Instruction::DELEGATECALL, 70);
+	setWeight(eth::Instruction::STATICCALL, 70);
+	setWeight(eth::Instruction::EXTCODECOPY, 70);
+	setWeight(eth::Instruction::EXTCODESIZE, 70);
+
+	//some smart addresses for calls
+	addAddress(Address("0xffffffffffffffffffffffffffffffffffffffff"));
```
gumb0: Instead of these `push_back`s you could initialize this vector with initializer list in braces { .. }
```
RandomCodeOptions::RandomCodeOptions(): 
  useUndefinedOpCodes(false), 
  smartCodeProbability(100),
  addressList{Address("0xffffffffffffffffffffffffffffffffffffffff"), Address("0x1000000000000000000000000000000000000000"), ...}
```

winsvega: I think it is nicer the way it is now.

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -273,91 +280,114 @@ void RandomCode::refreshSeed()
 
 std::string RandomCode::getPushCode(std::string const& _hex)
 {
-	int length = _hex.length() / 2;
+	std::string hexVal = _hex;
+	if (hexVal.empty())
+		hexVal = "00";
+	int length = hexVal.length() / 2;
```
pirapira: I think there should be a check about `hexVal.length() % 2`.

winsvega: why? 

len (010203040506) =  12 
but it is 6 bytes

12%2 = 0

pirapira: When `_hex` has an odd length, no push code should be available.

winsvega: this function asuume it has always correct hex on input.

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -35,13 +35,19 @@ boost::random::mt19937 RandomCode::gen;
...
 boostIntDistrib RandomCode::uniIntDist = boostIntDistrib (0, 0x7fffffff);
-boostUint64Distrib RandomCode::uInt64Dist = boostUint64Distrib (0, std::numeric_limits<uint64_t>::max());
+boostUint64 RandomCode::uInt64Dist = boostUint64 (0, std::numeric_limits<uint64_t>::max());
 
 boostIntGenerator RandomCode::randOpCodeGen = boostIntGenerator(gen, opCodeDist);
 boostIntGenerator RandomCode::randOpLengGen = boostIntGenerator(gen, opLengDist);
 boostIntGenerator RandomCode::randUniIntGen = boostIntGenerator(gen, uniIntDist);
 boostUInt64Generator RandomCode::randUInt64Gen = boostUInt64Generator(gen, uInt64Dist);
 
+const std::vector<eth::Instruction> RandomCode::invalidOpcodes{ eth::Instruction::INVALID, \
```
gumb0: `const` should be after type name

\ symbols are not needed

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: we could add a check if you think its necessary

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -273,91 +280,114 @@ void RandomCode::refreshSeed()
...
+		break;
+		case eth::Instruction::EXTCODESIZE:
+			code += getPushCode(toString(_options.getRandomAddress()));//address
+		break;
+		case eth::Instruction::CREATE:
 			//(CREATE value mem1 mem2)
-			code += getPushCode(randUniIntGen() % 128);  //memlen1
-			code += getPushCode(randUniIntGen() % 32);   //memlen1
-			code += getPushCode(randUniIntGen());		 //value
+			code += getPushCode(randUniIntGen());	//memlen1
```
gumb0: But now you generate here the values 0..0x7fffffff instead of 0..127

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -273,91 +280,114 @@ void RandomCode::refreshSeed()
...
+			code += getPushCode(randUniIntGen());	//memlen1
```
gumb0: And here 0..0x7fffffff instead of 0..31

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -218,13 +224,17 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
...
+		eth::InstructionInfo info = eth::instructionInfo(inst);
 
-		if (info.name.find("INVALID_INSTRUCTION") != std::string::npos)
+		if (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
+			|| std::find(invalidOpcodes.begin(), invalidOpcodes.end(), inst) != invalidOpcodes.end())
 		{
-			//Byte code is yet not implemented
-			if (_options.useUndefinedOpCodes == false)
+			if (_options.useUndefinedOpCodes)
+				code += toCompactHex(opcode, HexPrefix::DontAdd, 1);
```
gumb0: Still not clear: why don't we call `fillArguments` in this case?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -233,13 +243,10 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
...
 				code += toCompactHex(opcode);
-			code += fillArguments((dev::eth::Instruction) opcode, _options);
-		}
+			code += fillArguments(inst, _options);
 
-		if (info.name.find("PUSH") == std::string::npos)
-		{
-			std::string byte = toCompactHex(opcode);
-			code += (byte == "") ? "00" : byte;
+			if (info.name.find("PUSH") == std::string::npos)
```
gumb0: Well, here's the suggestion to make this clearer
```
if (info.name.find("PUSH") != std::string::npos)
{
  code += toCompactHex(opcode, HexPrefix::DontAdd, 1);
  code += fillArguments(inst, _options);
}
else
{
  code += fillArguments(inst, _options);
  code += toCompactHex(opcode, HexPrefix::DontAdd, 1);
}
```
(or maybe even move the code for generating args to PUSH from `fillArguments` into another function, because it's weird that the same function should be called in different order for different use cases)

test/tools/fuzzTesting/fuzzHelper.cpp
gumb0: assert would be enough I think

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: because UndefinedOpCodes are like random bytes. we dont have any record on how many arguments to use for it

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: it is not. that the way how you use the push opcode

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: thats fine I guess. it is 0....2GB

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: thats fine I guess. it is 0....2GB

test/tools/fuzzTesting/fuzzHelper.cpp
gumb0: We do at least for the ones in `invalidOpcodes`, `c_instructionInfo` table has info on how many arguments are needed.

There's no sense in adding these opcodes into the code without arguments, if EVM supports them, it probably will try to interpret current stack as arguments

gumb0: Maybe we don't need these option at all?

test/tools/fuzzTesting/fuzzHelper.cpp
gumb0: Well, that means it will generate code accessing 1GB of memory on average for each instruction.

I guess the idea of generating addresses and lengths 0..127 and 0..32 was to produce the code closer to real life scenarios

test/tools/fuzzTesting/fuzzHelper.h
```diff
@@ -97,6 +97,7 @@ class RandomCode
 	/// Generate random int64
 	static std::string randomUniIntHex(u256 _maxVal = 0);
 	static u256 randomUniInt(u256 _maxVal = 0);
+	static std::vector<eth::Instruction> const invalidOpcodes;
```
chfast: Let's move it to .cpp file only. This is no need for keep this set in header.

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -35,13 +35,19 @@ boost::random::mt19937 RandomCode::gen;
...
 boostIntDistrib RandomCode::uniIntDist = boostIntDistrib (0, 0x7fffffff);
-boostUint64Distrib RandomCode::uInt64Dist = boostUint64Distrib (0, std::numeric_limits<uint64_t>::max());
+boostUint64 RandomCode::uInt64Dist = boostUint64 (0, std::numeric_limits<uint64_t>::max());
 
 boostIntGenerator RandomCode::randOpCodeGen = boostIntGenerator(gen, opCodeDist);
 boostIntGenerator RandomCode::randOpLengGen = boostIntGenerator(gen, opLengDist);
 boostIntGenerator RandomCode::randUniIntGen = boostIntGenerator(gen, uniIntDist);
 boostUInt64Generator RandomCode::randUInt64Gen = boostUInt64Generator(gen, uInt64Dist);
 
+const std::vector<eth::Instruction> RandomCode::invalidOpcodes{ eth::Instruction::INVALID,
```
chfast: Use `std::array` here. And reformat code into a nice list.

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: there one sense: to see what would happen. so if flag is defined it will randomly add a byte that does not represent a valid opcode. that was the initial idea.

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: yes. so it to be executable. what we need is a distribution from 0...2GB  but with higher probability on first 10 MB

test/tools/libtesteth/Options.cpp
```diff
@@ -234,8 +236,23 @@ Options::Options(int argc, char** argv)
 		}
 		else if (arg == "--statediff")
 			statediff = true;
+		else if (arg == "--randomcode")
+		{
+			throwIfNoArgumentFollows();
+			int maxCodes = atoi(argv[++i]);
+			if (maxCodes > 1000)
```
pirapira: Check for a negative as well.

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -34,14 +34,22 @@ namespace test
...
-boostUint64Distrib RandomCode::uInt64Dist = boostUint64Distrib (0, std::numeric_limits<uint64_t>::max());
+boostUint64 RandomCode::uInt64Dist = boostUint64 (0, std::numeric_limits<uint64_t>::max());
 
 boostIntGenerator RandomCode::randOpCodeGen = boostIntGenerator(gen, opCodeDist);
 boostIntGenerator RandomCode::randOpLengGen = boostIntGenerator(gen, opLengDist);
+boostIntGenerator RandomCode::randOpMemrGen = boostIntGenerator(gen, opMemrDist);
 boostIntGenerator RandomCode::randUniIntGen = boostIntGenerator(gen, uniIntDist);
 boostUInt64Generator RandomCode::randUInt64Gen = boostUInt64Generator(gen, uInt64Dist);
 
+const std::array<eth::Instruction, 14> invalidOpcodes{ eth::Instruction::INVALID,
```
chfast: Fix the formatting. Maybe put the item each in separate line.

chfast:
You have merged this with build failure: https://travis-ci.org/ethereum/cpp-ethereum/jobs/247065035.

PR #4196, opened 2017-06-22
testeth errors when a general state test filler contains more than one test.

### LINKED ISSUE #4092: testeth should raise an error when a test filler contains more than one test
Currently, a test filler file is supposed to contain only one test.  This issue keeps track if changing `testeth` so that it raises an error when a file contains more than one test.  Also, if the name of the test is different from the name of the file (minus `Filler.json`), `testeth` should raise an error.

pirapira:
Fixes #4092.

### DIFF
--- a/test/tools/jsontests/StateTests.cpp
+++ b/test/tools/jsontests/StateTests.cpp
@@ -41,6 +41,9 @@ namespace dev {  namespace test {
 
 void doStateTests(json_spirit::mValue& _v, bool _fillin)
 {
+	BOOST_REQUIRE_MESSAGE(!_fillin || _v.get_obj().size() == 1,
+		TestOutputHelper::testFileName() + " A GeneralStateTest filler should contain only one test.");
+
 	for (auto& i: _v.get_obj())
 	{
 		string testname = i.first;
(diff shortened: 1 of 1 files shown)


test/tools/jsontests/StateTests.cpp
```diff
@@ -41,6 +41,9 @@ namespace dev {  namespace test {
 
 void doStateTests(json_spirit::mValue& _v, bool _fillin)
 {
+	BOOST_REQUIRE_MESSAGE(!_fillin || _v.get_obj().size() == 1,
+		TestOutputHelper::testFileName() + " A GeneralStateTest filler should contain only one test.");
```
winsvega: _fillin == true means it is a filler

pirapira: Exactly.
* when `_fillin == true`, this assertion checks the number of tests in the file.
* when `_fillin == false`, this assertion is skipped.

The strict checking is required when doing the filling, not when running existing tests.

winsvega:
here is the problem with PRs. 
if we merge this. then we will have to rebase all PRs that change tests  commit hash and wait again for rebuild on travis and appveyour

Issue #4199, opened 2017-06-23
STATICCALL gives stipend to a callee


chfast:
Buggy code: https://github.com/ethereum/cpp-ethereum/blob/develop/libevm/VMCalls.cpp#L251-L252

We have to refactor not to use random access like `m_SP[...]` all over the place.

PR #4200, opened 2017-06-23
Fix STATICCALL

chfast:
Fixes https://github.com/ethereum/cpp-ethereum/issues/4199.

### DIFF
--- a/libevm/VMCalls.cpp
+++ b/libevm/VMCalls.cpp
@@ -207,6 +207,10 @@ void VM::caseCall()
 
 bool VM::caseCallSetup(CallParameters *callParams, bytesRef& o_output)
 {
+	// Make sure the params were properly initialized.
+	assert(callParams->valueTransfer == 0);
+	assert(callParams->apparentValue == 0);
+
 	m_runGas = toInt63(m_schedule->callGas);
 
 	callParams->staticCall = (m_OP == Instruction::STATICCALL || m_ext->staticCall);
@@ -260,11 +264,9 @@ bool VM::caseCallSetup(CallParameters *callParams, bytesRef& o_output)
 		callParams->valueTransfer = m_SP[2];
 		callParams->apparentValue = m_SP[2];
 	}
-	else
-	{
+	else if (m_OP == Instruction::DELEGATECALL)
+		// Forward VALUE.
 		callParams->apparentValue = m_ext->value;
-		callParams->valueTransfer = 0;
-	}
 
 	uint64_t inOff = (uint64_t)inputOffset;
 	uint64_t inSize = (uint64_t)inputSize;
--- a/libevm/VMCalls.cpp
+++ b/libevm/VMCalls.cpp
@@ -211,15 +211,17 @@ bool VM::caseCallSetup(CallParameters *callParams, bytesRef& o_output)
 
 	callParams->staticCall = (m_OP == Instruction::STATICCALL || m_ext->staticCall);
 
+	bool const haveValueArg = m_OP == Instruction::CALL || m_OP == Instruction::CALLCODE;
+
 	Address destinationAddr = asAddress(m_SP[1]);
 	if (m_OP == Instruction::CALL && !m_ext->exists(destinationAddr))
 		if (m_SP[2] > 0 || m_schedule->zeroValueTransferChargesNewAccountGas())
 			m_runGas += toInt63(m_schedule->callNewAccountGas);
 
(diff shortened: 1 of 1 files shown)


chfast:
This also requires JSON tests update. I can't do it because I have issues with compiling LLL code.

libevm/VMCalls.cpp
```diff
@@ -211,15 +211,17 @@ bool VM::caseCallSetup(CallParameters *callParams, bytesRef& o_output)
...
 	if (m_OP == Instruction::CALL && !m_ext->exists(destinationAddr))
 		if (m_SP[2] > 0 || m_schedule->zeroValueTransferChargesNewAccountGas())
 			m_runGas += toInt63(m_schedule->callNewAccountGas);
 
-	if ((m_OP == Instruction::CALL || m_OP == Instruction::CALLCODE) && m_SP[2] > 0)
+	if (haveValueArg && m_SP[2] > 0)
 		m_runGas += toInt63(m_schedule->callValueTransferGas);
 
-	size_t sizesOffset = (m_OP == Instruction::DELEGATECALL || m_OP == Instruction::STATICCALL) ? 2 : 3;
+	size_t const sizesOffset = haveValueArg ? 3 : 2;
```
chriseth: Did you forget a `!` here or was there a bug before?

libevm/VMCalls.cpp
```diff
@@ -248,27 +250,26 @@ bool VM::caseCallSetup(CallParameters *callParams, bytesRef& o_output)
...
-		callParams->apparentValue = m_ext->value;
-		callParams->valueTransfer = 0;
+		callParams->valueTransfer = m_SP[2];
+		callParams->apparentValue = m_SP[2];
 	}
 	else
 	{
-		callParams->apparentValue = callParams->valueTransfer = m_SP[2];
-		inOutOffset = 1;
+		callParams->apparentValue = m_ext->value;
```
chriseth: Do we want to forward the value in case of staticcall? I would say thiscase should only apply to delegatecall.

libevm/VMCalls.cpp
chfast: I also swapped the 2 and 3.

libevm/VMCalls.cpp
chfast: You are probably right. In case of STATICALL the value should not be forwarded and be 0 all the time.
This might be another bug. @winsvega Can you look for some test cases for this?

winsvega: I have those on the list
actually callvalue has some strange value at this test 
https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stStaticCall/static_callToDelCallOpCodeCheckFiller.json#L43

winsvega: and here 
https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stStaticCall/static_callToStaticOpCodeCheckFiller.json#L43

chfast: All of them are quite complex. We need a simple test with a transaction having no-zero value, then staticcall and check of the callvalue inside this staticcall.

Similar one with a CALL instead of transaction.

winsvega: this test does exactly just that
just add more values to the value list

chfast: I'm not sure what you mean by "this test". You referenced 2 tests before. If you mean https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stStaticCall/static_callToStaticOpCodeCheckFiller.json#L43, this tests does other weird things like: checking for ORIGIN, multiple levels of different calls, etc. I'm not sure any of them check if the value from transaction is forwarded in STATICCALL.

Also, I don't know what you mean by "just add more values to the value list".

pirapira: I will add tests.

1. call transaction into A with non-zero value; A performs STATICCALL to B; B stores the call value
2. call transaction into Q; Q calls into A with non-zero value; A performs STATICCALL to B; B stores the call value

In both cases, the storage slot of `B` would initially contain some non-zero value, so that we can be sure that the value is overwritten by zero.

pirapira:
I added the two tests.

pirapira:
Let me have a look at these
* `static_callToDelCallOpCodeCheck Expected another postState hash`
* `static_callToStaticOpCodeCheck Expected another postState hash`

chfast:
Ready to go?

pirapira:
@winsvega sounds not happy about the tests I changed.

pirapira:
I changed the tests and rendered them useless.  I will apply correct fixes.

PR #4201, opened 2017-06-26
SIMD

gcolvin:
This is a portable implementation of the https://github.com/ethereum/EIPs/issues/616 SIMD proposal.
* The proposal itself has had a lot of attention as a result of trying to implement it, and because the wasm simd proposal surfaced.
* The SIMD code itself is all in the new VMSIMD.cpp.  The rest is just hooking it up.
* The new opcodes and their implementations are walled off from production use with an EIP_616 macro set false in Instructions.h.
* Yesterday I gave up on the fairly clean design I had been using - defining stack slots as a union of u256 and the simd vectors - because so much code changed accessing union members that rebasing failed miserably.  The code became harder to read as well.
* Instead I leave u256 as the slot type, so no code expecting that has to change.  I punch through that type only in VMSIMD.cpp to store raw arrays on the same memory otherwise used by the u256 stack slots.


gcolvin:
@winsvega  clang won't compile the std::array initializer in fuzzHelper, but gcc and vcc will.
```
/Users/travis/build/ethereum/cpp-ethereum/test/tools/fuzzTesting/fuzzHelper.cpp:31:2: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
```
C++11 requires extra braces, C++14 doesn't.  So I added the silly braces.  They didn't break gcc, will see soon if the heal clang.

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,669 @@
...
+		}
+		break;
+	default: throwBadInstruction();
+	}
+}
+
+void VM::wtov(uint8_t _b, const u256& _in, u256& _out)
+
+{
+	const uint8_t n = pow2N((_b) & 0xf), t = (_b) >> 4;
```
gumb0: This calculation is duplicated all over the place, create a couple of helper functions to get number of elements and type

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,669 @@
...
+
+// conversion functions to overlay vectors on storage for u256 stack slots
+// a dirty trick but it keeps the SIMD types from polluting the rest of the VM
+// so at least assert there's room for the trick, and use typedefs for some safety
+static_assert(sizeof(uint64_t[4]) <= sizeof(u256), "stack slot too narrow for SIMD");
+typedef uint64_t a64x4 [4];		
+typedef uint32_t a32x8 [8];
+typedef uint16_t a16x16[16];
+typedef uint8_t  a8x32 [32];
+      a64x4&  v64x4 (      u256& _stack_item) { return (a64x4&) *(a64x4*) &_stack_item; }
```
gumb0: Declare these functions `inline` maybe?

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,669 @@
...
+		for (int j = n, v = 0, i = 0; 0 < n; ++i)
+		{
+			v |= p[--j];
+			v8x32(m_SPP[0])[i] = v;
+		}
+		break;
+	case 1:
+		for (int j = n, v = 0, i = 0; 0 < n; ++i)
+		{
+			v |= p[--j];
```
gumb0: `p[--j]` is `uint8_t`, so you bit-or it with lowest 8 bits of `v` and then on the next line you shift right 8 bits, loosing what you've just read from `p` ? What am I getting wrong?

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,669 @@
...
+	m_SPP[0] = 0;
+	switch (t)
+	{
+	case 0:
+		for (int i = 0; i < n; ++i) {
```
gumb0: Braces on separate lines

gumb0:
This is an impressive amount of work, I wouldn't mind someone helping with the review 😃 @chfast @chriseth

gumb0:
I haven't finished looking through all of the VMSIMD.cpp yet, will try to continue tomorrow

libevm/VM.cpp
```diff
@@ -14,8 +14,6 @@
 	You should have received a copy of the GNU General Public License
 	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
 */
-/** @file VM.cpp
```
gcolvin: I think (could be wrong) Pawel asked me to remove them.  I took a lot out.

gcolvin: OK

libevm/VM.cpp
```diff
@@ -643,7 +657,296 @@ void VM::interpretCases()
...
+			xand(m_code[++m_PC]); ++m_PC;
+		}
+		CONTINUE
+         
+		CASE(XXOR)
+		{
+			ON_OP();
+			updateIOGas();
+
+			xoor(m_code[++m_PC]); ++m_PC;
```
gcolvin: I think xoor is a typo and only one of them is needed, thanks.

libevm/VM.cpp
```diff
@@ -643,7 +657,296 @@ void VM::interpretCases()
...
+			ON_OP();
+			updateIOGas();
+
+			xpush(m_code[++m_PC]); ++m_PC;
+		}
+		CONTINUE
+
+		CASE(XPUT)
+		{
+			uint8_t b = ++m_PC;
```
gcolvin: Probably because I forgot.  Thanks.

libevm/VMSIMD.cpp
gcolvin: I'm familiar with this idiom and why it works.  I'm not sure one reinterpret cast would work, though it might compile.   I'm more concerned with the undefined behavior that this trick opens up in principle.

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,669 @@
...
+namespace dev
+{
+namespace eth
+{
+
+// conversion functions to overlay vectors on storage for u256 stack slots
+// a dirty trick but it keeps the SIMD types from polluting the rest of the VM
+// so at least assert there's room for the trick, and use typedefs for some safety
+static_assert(sizeof(uint64_t[4]) <= sizeof(u256), "stack slot too narrow for SIMD");
+typedef uint64_t a64x4 [4];
```
gcolvin: Indeed.  Some standards prefer typedef for this purpose - convenient names in a small scope.

libevm/VMSIMD.cpp
gcolvin: Good idea.  (Though compilers are pretty good now.)

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,669 @@
...
+static_assert(sizeof(uint64_t[4]) <= sizeof(u256), "stack slot too narrow for SIMD");
+typedef uint64_t a64x4 [4];		
+typedef uint32_t a32x8 [8];
+typedef uint16_t a16x16[16];
+typedef uint8_t  a8x32 [32];
+      a64x4&  v64x4 (      u256& _stack_item) { return (a64x4&) *(a64x4*) &_stack_item; }
+      a32x8&  v32x8 (      u256& _stack_item) { return (a32x8&) *(a32x8*) &_stack_item; }
+      a16x16& v16x16(      u256& _stack_item) { return (a16x16&)*(a16x16*)&_stack_item; }
+      a8x32&  v8x32 (      u256& _stack_item) { return (a8x32&) *(a8x32*) &_stack_item; }
+const a64x4&  v64x4 (const u256& _stack_item) { return (a64x4&) *(a64x4*) &_stack_item; }
```
gcolvin: Another three-decade old habit I need to break, writing code that looks wrong but is right. :(
 (K&R, Bjarne's books, and all editions of the C and C++ programming standards put the const first.)

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,669 @@
...
+void VM::xrol (uint8_t _b) { EVALXOPU(ROL, _b); }
+void VM::xror (uint8_t _b) { EVALXOPU(ROR, _b); }
+
+uint8_t pow2N(uint8_t n)
+{
+	static uint8_t exp[6] = { 1, 2, 4, 8, 16, 32 };
+	return exp[n];
+}
+
+void VM::vtow(uint8_t _b, const u256& _in, u256& _out)
```
gcolvin: It's a big object, and I wanted to be sure it wasn't copied on return.

libevm/VMSIMD.cpp
gcolvin: Good idea.

libevm/VMSIMD.cpp
gcolvin: Another bad habit.  Messed up less often than usual in this code...

gcolvin:
Thanks @gumb0 Andrei

libevm/VMSIMD.cpp
gcolvin: Looks like I pasted the mstore() code as a starting template and then failed to reverse the shifts.  Thanks.

libevm/VM.cpp
```diff
@@ -115,6 +113,22 @@ void VM::adjustStack(unsigned _removed, unsigned _added)
 #endif
 }
 
+void VM::sstoreGas()
+{
+	if (m_ext->staticCall)
```
gcolvin: I copied the existing code in SSTORE exactly.  Could move this one piece to SSTORE and to XSTORE?

gcolvin: Yeah, it's included as a preamble to lots of functions that way.  Somebody slipped them in while I wasn't looking ;)

libevm/VM.cpp
```diff
@@ -643,7 +657,296 @@ void VM::interpretCases()
...
+			++m_PC;
+		}
+		CONTINUE
+	         
+		CASE(XMUL)
+		{
+			ON_OP();
+			updateIOGas();
+
+			xmul(m_code[++m_PC]); ++m_PC;
```
gcolvin: Created one function for advancing the PC and fetching the simd type.

libevm/VM.cpp
gcolvin: Wrong.  Seems `xor` is now a keyword, so I called the function `xoor`, and it needs a case to call it too.

libevm/VM.cpp
gcolvin: Oh - ON_OP() just because I forgot.  The rest because they have complicated gas functions I haven't written yet.

libevm/VMSIMD.cpp
gcolvin: Yep, that one function looks to be the only place.  A record!

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,687 @@
...
+namespace dev
+{
+namespace eth
+{
+
+// conversion functions to overlay vectors on storage for u256 stack slots
+// a dirty trick but it keeps the SIMD types from polluting the rest of the VM
+// so at least assert there's room for the trick, and use wrappers for some safety
+static_assert(sizeof(uint64_t[4]) <= sizeof(u256), "stack slot too narrow for SIMD");
+typedef uint64_t a64x4 [4];
```
gumb0: `using a64x4 = uint64_t[4];` is more readable and does the same, I still recommend changing it

libevm/VMSIMD.cpp
gumb0: I'm pretty sure `return reinterpret_cast<a64x4&>(_stack_item);` will do the same.

libevm/VMSIMD.cpp
gumb0: It's guaranteed by the standard to be moved, so in general we should avoid output parameters in C++11 and further

(Anyway the naming convention for output params is o_parameterName)

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,687 @@
...
+		}
+		break;
+	default: throwBadInstruction();
+	}
+}
+
+void VM::xget(uint8_t _b, uint8_t _c)
+{
+	// n type t elements in source vector, m type u in get indexes
+	const uint8_t n = nElem(_b), t = elemType(_b);
```
gumb0: Is `n` used anywhere in this function?

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,687 @@
...
+		}
+
+	default: throwBadInstruction();
+	}
+}
+
+void VM::xput(uint8_t _b, uint8_t _c)
+{
+	// n type t elements in source vector, m type u in put index
+	const uint8_t n = nElem(_b), t = elemType(_b);
```
gumb0: The same here, looks like `n` is not used, shouldn't it be?

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,687 @@
...
+	// for every index in the mask copy out the indexed value in one of the sources
+	switch (t)
+	{
+	case 0:
+		for (int i = 0; i < n; ++i) {
```
gumb0: braces on separate lines

libevm/VMSIMD.cpp
gcolvin: Will consider.  typedef introduces a new name for an existing type, which is often not what people actually want.  But it's exactly what I want.

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,687 @@
...
+
+// conversion functions to overlay vectors on storage for u256 stack slots
+// a dirty trick but it keeps the SIMD types from polluting the rest of the VM
+// so at least assert there's room for the trick, and use wrappers for some safety
+static_assert(sizeof(uint64_t[4]) <= sizeof(u256), "stack slot too narrow for SIMD");
+typedef uint64_t a64x4 [4];		
+typedef uint32_t a32x8 [8];
+typedef uint16_t a16x16[16];
+typedef uint8_t  a8x32 [32];
+inline a64x4        &v64x4 (u256       &_stack_item) { return (a64x4&) *(a64x4*) &_stack_item; }
```
gcolvin: Not sure I understand you, but this code is currently behind macros and not compiled, so if I've made that sort of mistake it won't catch until I take time to drop the guards and compile the new code.

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,687 @@
...
+#define DIV( x1, x2) ((x1) / (x2))
+#define MOD( x1, x2) ((x1) % (x2))
+#define LT(  x1, x2) ((x1) < (x2))
+#define GT(  x1, x2) ((x1) > (x2))
+#define EQ(  x1, x2) ((x1) == (x2))
+#define ZERO(x1, x2) ((x1) == 0)
+#define AND( x1, x2) ((x1) & (x2))
+#define OR(  x1, x2) ((x1) | (x2))
+#define XOR( x1, x2) ((x1) ^ (x2))
+#define NOT( x1, x2) (~x1)
```
gcolvin: Yes, thanks.

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,687 @@
...
+}
+
+inline uint8_t nElem(uint8_t _b)
+{
+	return pow2N((_b) & 0xf);
+}
+
+inline uint8_t elemType(uint8_t _b)
+{
+	return (_b) >> 4;
```
gcolvin: Wasn't compiled - see above :{

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,687 @@
...
+
+{
+	const uint8_t n = nElem(_b), t = elemType(_b);
+	switch (t)
+	{
+	case 0:
+		for (int i = n-1; 0 <= i; --i)
+		{ 
+			v8x32(_out) [i] |= (uint8_t )(_in & 0xff);
+			_in  << 8;
```
gcolvin: probably

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,687 @@
...
+	default: throwBadInstruction();
+	}
+}
+
+void VM::xmload (uint8_t _b)
+{
+	// n bytes of type t elements in memory vector
+	// goes onto stack element by element, LSB first
+	uint8_t *p = m_mem.data() + toInt15(m_SP[0]);
+	const uint8_t n = nElem(_b), t = elemType(_b);
```
gcolvin: I think that only works on LSB-first machines, and there is no portable way to know whether you are on one.  So that becomes part of the non-portable optimizations.

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,687 @@
...
+	switch (t)
+	{
+	case 0:
+		for (int j = n, v = 0, i = 0; 0 < n; ++i)
```
gcolvin: I'll check it out, thanks.  Loops should all have the same structure.

libevm/VMSIMD.cpp
gcolvin: Probably not.  Just a line cut and pasted into all of the other functions.  Will remove it when I split all those lines and make the const backwards.

libevm/VMSIMD.cpp
gcolvin: Same cut and paste issue.

libevm/VMSIMD.cpp
gcolvin: Arrgh...

libevm/VMSIMD.cpp
gumb0: No, it's correct, it's just formatting issue - we put ampersand next to return type/variable type, not next to the name (see Coding standards 6c)

gcolvin: Oh, right.  That's how I usually do it, messed up here.

libevm/VM.cpp
gcolvin: @chriseth ?

libevm/VMSIMD.cpp
gcolvin: I'm seeing why I broke my rule and did it the old-fashioned way . In this case - with everything else vertically aligned - not aligning the ampersands is ugly and hard to read.

hard-fault:
Hi,
I have a problem with building the code base.
I followed all the steps given on the ethereum git home page.
But ran into issue with this command.

cmake --build . # Build all default targets.

/Users/shreyasm/Desktop/Github/cpp-ethereum/build
bash-3.2$ cmake --build .
[ 0%] Built target BuildInfo.h
[ 1%] Performing download step (download, verify and extract) for 'boost'
-- verifying file...
file='/Users/shreyasm/Desktop/Github/cpp-ethereum/deps/src/boost_1_63_0.tar.gz'
-- SHA256 hash of
/Users/shreyasm/Desktop/Github/cpp-ethereum/deps/src/boost_1_63_0.tar.gz
does not match expected value
expected: 'fe34a4e119798e10b8cc9e565b3b0284e9fd3977ec8a1b19586ad1dec397088b'
actual: 'efe2318e2797d664f9da43f38aa56440cd755b8ae81ba90edfdde4bc422d5711'
-- File already exists but hash mismatch. Removing...
-- Downloading...
dst='/Users/shreyasm/Desktop/Github/cpp-ethereum/deps/src/boost_1_63_0.tar.gz'
timeout='none'
-- Using src='https://github.com/ethereum/cpp-dependencies/releases/download/cache/boost_1_63_0.tar.gz'

It is stuck in that step for more than 24Hrs.

Please help me with this.

I am looking forward to fix some bugs.

Thanks,
Shreyas Badiger

libevm/VMSIMD.cpp
gcolvin: done

gcolvin:
Shreyhas, you may find help here: https://gitter.im/ethereum/cpp-ethereum-development

gcolvin:
VCC is right, GCC and I are wrong, and I'm backing out attempt at configurable instruction set.

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,722 @@
...
+	// n bytes of type t elements in stack vector
+	// goes onto memory by element, LSB first
+	uint8_t *p = m_mem.data() + toInt15(m_SP[0]);
+	uint8_t const n = nElem(_b);
+	uint8_t const t = elemType(_b);
+
+	switch (t)
+	{
+	case 0:
+		for (int j = n, v = 0, i = 0; 0 < n; ++i)
```
gumb0: Here the loops are still different for different cases, this one looks like indefinite loop

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,722 @@
...
+void VM::xsload(uint8_t _b)
+{
+	u256 w = m_ext->store(m_SP[0]);
+	m_SPP[0] = wtov(_b, w);
+}
+
+void VM::xsstore(uint8_t _b)
+{
+	u256 w;
+	w = vtow(_b, m_SP[1]);
```
gumb0: Assign on the previoud (declaration) line

libevm/VMSIMD.cpp
gcolvin: Yeah, I make that mistake in four places - it should be 'i < n'.  But why am I counting up there but down after?   I realized I couldn't remember how these worked - so set them aside as I fixed easier things. Now I have this problem preventing compilation with the EIP's off - but only on Windows.

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,722 @@
...
+{
+	u256 out;
+	uint8_t const n = nElem(_b);
+	uint8_t const t = elemType(_b);
+	switch (t)
+	{
+	case 0:
+		for (int i = n-1; 0 <= i; --i)
+		{ 
+			v8x32(out) [i] |= (uint8_t )(_in & 0xff);
```
gumb0: In other places you do `% 4`, `% 16` etc., maybe let's do it this way insead of `& 0xff`, `& 0xffff` etc. here for consistency, too

gcolvin: I'd rather stick with the bitwise operator for masking off bytes - it's the standard idiom, and using `%` would just be confusing.  (Also hugely more efficient unoptimized.)  But I'll sleep on it.   Soon.

test/unittests/libdevcrypto/LibSnark.cpp
```diff
@@ -338,6 +334,155 @@ BOOST_AUTO_TEST_CASE(pairingNullInput)
 	BOOST_CHECK(!r.first);
 }
 
+BOOST_AUTO_TEST_CASE(generateRandomPoints)
```
gumb0: I don't get why the tests for EC stuff are being added in this PR? Some anomaly during rebase?

gumb0: Of it's actually testing SIMD operations, then probably better move it to separate file

gcolvin: Oh damn, @gumb0   Git pulled irrelevant transactions into this PR when I pushed my changes.  I don't know whether it will ignore them as redundant when the final merge comes.
@chriseth do you know?
Worst case - when we are ready to merge I start a new PR.

gcolvin:
@gumb0  The commit currently needing review (and not displayed on this page) is https://github.com/ethereum/cpp-ethereum/pull/4201/commits/b4bacee60db5ccfc410833501dac6aefc7ddd213

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,722 @@
...
+void VM::xnot (uint8_t _b) { EVALXOPU(NOT, _b); }
+void VM::xshr (uint8_t _b) { EVALXOPU(SHR, _b); }
+void VM::xsar (uint8_t _b) { EVALXOPS(SHR, _b); }
+void VM::xshl (uint8_t _b) { EVALXOPU(SHL, _b); }
+void VM::xrol (uint8_t _b) { EVALXOPU(ROL, _b); }
+void VM::xror (uint8_t _b) { EVALXOPU(ROR, _b); }
+
+inline uint8_t pow2N(uint8_t n)
+{
+	static uint8_t exp[6] = { 1, 2, 4, 8, 16, 32 };
```
gumb0: We need to handle here `n` possibly being > 5 (throw exception?), since you pass here `(_b) & 0xf`, it can be up to 15

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,722 @@
...
+	{
+	case 0:
+		for (int i = n-1; 0 <= i; --i)
+		{ 
+			out << 8;
+			out |= v8x32(_in) [i];
+		}
+		break;
+	case 1:
+		for (int i = n-1; 0 <= i; --i)
```
gumb0: I think we also need to figure out how to check that n is within allowed limits here and in other cases.
If I understand this correctly the code can contain encoding like "32 items 16 bits wide", then this loop will go out of bounds of the array returned by `v16x16(_in)`.

gumb0: Maybe instead of `nElem` & `elemType` we can have one function returning a pair and validating if it's allowed combination encoded?

libevm/VMSIMD.cpp
gcolvin: See below

libevm/VMSIMD.cpp
gcolvin: See below.

libevm/VMSIMD.cpp
gcolvin: done

libevm/VMSIMD.cpp
gcolvin: fixed - need to count down to reverse LSP/MSB

libevm/VMSIMD.cpp
gcolvin: nElem() and elemType() get their input from validated code.

libevm/VMSIMD.cpp
gcolvin: fixed

libevm/VMSIMD.cpp
gumb0: I see, then we could add an assert here for this assumption

libevm/VMSIMD.cpp
gumb0: Ok, makes sense, here it's masking, in other places you calculate array index in some range - % is more clear there

gumb0: I think you could use simple assignment here instead of `|=` ?
You access different output vector elements on each iteration and they are all initially 0

libevm/VMSIMD.cpp
gumb0: Still don't see it fixed, did you push the fix?

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,722 @@
...
+	// n bytes of type t elements in memory vector
+	// goes onto stack element by element, LSB first
+	uint8_t* p = m_mem.data() + toInt15(m_SP[0]);
+	uint8_t const n = nElem(_b);
+	uint8_t const t = elemType(_b);
+
+	switch (t)
+	{
+	case 0:
+		for (int j = n, v = 0, i = 0; 0 < n; ++i)
```
gumb0: The same here as in `xmstore` - the loop is indefinite and differs from cases 2 & 3

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,722 @@
...
+		throwBadInstruction();
+	}
+	return out;
+}
+
+void VM::xmload (uint8_t _b)
+{
+	// n bytes of type t elements in memory vector
+	// goes onto stack element by element, LSB first
+	uint8_t* p = m_mem.data() + toInt15(m_SP[0]);
```
gumb0: this could be made `uint_8 const*`

libevm/VMSIMD.cpp
gumb0: Don't you assume that uint16_t, uint32_t, uint64_t are LSB-first when you access vectors on stack through arrays like uint16_t[16] etc. ?

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,722 @@
...
+			v <<= 8;
+			v |= p[--j];
+			v16x16(m_SPP[0])[i] = v;
+		}
+		break;
+	case 2:
+		for (int j = n,  v = 0, i = n - 1; 0 <= i; --i)
+		{
+			v |= p[--j];
+			v <<= 8;
```
gumb0: These repeated shifts suck 😆 Maybe let's create some helper function like `int readFromMemory(uint_t* _p, int _countBytes)` and do this in a loop inside it ?

gumb0: Also, if you leave it like this I think `v = 0` should be done inside loop body, each iteration get a new int from memory

libevm/VMSIMD.cpp
gcolvin: Good idea.  Note that the switches on the element type have a default case that throws.

gcolvin: Note that the switches on element type have a default case that throws, so that at least is OK.

libevm/VMSIMD.cpp
gcolvin: I'll look at this.  To an extent, once you validate code you trust the validation.  That allows for better optimization.

libevm/VMSIMD.cpp
gcolvin: done

libevm/VMSIMD.cpp
gcolvin: The shift idiom is the standard way to portably pack a variable with bytes.  I don't see a reason to hide it, or not to unroll the look here.  Might unroll some others.
Yes, initialize v at top of loop, thanks.

gumb0: Shift itself is fine, but why do you think repeating it 8 times is better looking than a loop?

gcolvin: Doesn't look better - or worse - to me either way.  Just maybe faster unrolled.  Plus it's only the 8-liner that would take less lines as a loop.  So I'm more inclined to unroll some other loops than to roll these up.  But the compiler can probably handle it either way.

libevm/VMSIMD.cpp
gcolvin: Yes, assignment is better, thanks.

gcolvin:
This branch is gitted to death.  Work has moved to ethereum/cpp-ethereum/pull/4233.

gcolvin:
Trashed and closed.  Work continues here - https://github.com/ethereum/cpp-ethereum/pull/4233

PR #4227, opened 2017-07-06
WIP - Download snapshot through Parity's warp protocol

gumb0:
This is still in PoC-stage, downloads only from single peer, needs improvements in many minor details and tests.

It's better to start this pinning to one of Parity's boot nodes:
```
./eth --download-snapshot snapshot_dir -v 9 --peerset required:016b20125f447a3b203a3cae953b2ede8ffe51290c071e7599294be84317635730c397b8ff74404d6be412d539ee5bb5c3c700618723d3b53958c92bd33eaa82@159.203.210.80:30303 --pin --no-discovery
```

(other bootnodes for main net are here: https://github.com/paritytech/parity/blob/master/ethcore/res/ethereum/foundation.json#L182 )

### DIFF
--- a/libethereum/WarpPeerCapability.cpp
+++ b/libethereum/WarpPeerCapability.cpp
@@ -45,6 +45,8 @@ bool WarpPeerCapability::interpret(unsigned _id, RLP const& _r)
 		{
 		case WarpStatusPacket:
 		{
+			// TODO check that packet has enough elements
+
 			// Packet layout:
 			// [ version:P, state_hashes : [hash_1:B_32, hash_2 : B_32, ...],  block_hashes : [hash_1:B_32, hash_2 : B_32, ...], 
 			//		state_root : B_32, block_number : P, block_hash : B_32 ]
--- a/libethereum/WarpPeerCapability.cpp
+++ b/libethereum/WarpPeerCapability.cpp
@@ -1,0 +1,175 @@
+/*
+	This file is part of cpp-ethereum.
+
+	cpp-ethereum is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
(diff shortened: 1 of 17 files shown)


libethereum/WarpPeerCapability.cpp
```diff
@@ -0,0 +1,175 @@
...
+			// Packet layout:
+			// [ version:P, state_hashes : [hash_1:B_32, hash_2 : B_32, ...],  block_hashes : [hash_1:B_32, hash_2 : B_32, ...], 
+			//		state_root : B_32, block_number : P, block_hash : B_32 ]
+			m_protocolVersion = _r[0].toInt<unsigned>();
+			m_networkId = _r[1].toInt<u256>();
+			m_totalDifficulty = _r[2].toInt<u256>();
+			m_latestHash = _r[3].toHash<h256>();
+			m_genesisHash = _r[4].toHash<h256>();
+			m_snapshotHash = _r[5].toHash<h256>();
+			m_snapshotNumber = _r[6].toInt<u256>();
```
chriseth: What happens if the index goes "out of bounds"?

libethereum/WarpPeerCapability.cpp
```diff
@@ -0,0 +1,175 @@
...
+	std::shared_ptr<SessionFace> s = session();
+	if (!s)
+		return false; // Expired
+
+	if (m_genesisHash != _genesisHash)
+	{
+		disable("Invalid genesis hash");
+		return false;
+	}
+	if (find(_protocolVersions.begin(), _protocolVersions.end(), m_protocolVersion) == _protocolVersions.end())
```
chriseth: I know it's a little less efficient on the average case, but `_procotolVersions.count(m_protocolVersion)` is much better readable and less error-prone.

gumb0:
Closing in favor of approach of https://github.com/ethereum/cpp-ethereum/pull/4622

PR #4233, opened 2017-07-07
new branch for PR, attempted vcc fix, other cleanup

gcolvin:
The specification is here - ethereum/EIPs#616
The previous PR and review is here - ethereum/cpp-ethereum/pull/4201

@gumb0  This is the latest code for review.
@pirapira  SIMD vectors and operations are typed.  I would be very interested in what you have to say about how well the correct use of the types can be validated and otherwise worked with formally.


libevm/VM.cpp
```diff
@@ -643,7 +655,460 @@ void VM::interpretCases()
...
+		}
+		CONTINUE
+
+		CASE(JUMPSUB)
+		{
+			ON_OP();
+			updateIOGas();
+			*m_RP++ = m_PC++;
+			m_PC = decodeJumpDest(m_code.data(), m_PC);
+			}
```
gumb0: too many braces

libevm/VM.cpp
```diff
@@ -643,7 +655,460 @@ void VM::interpretCases()
...
+		CONTINUE
+
+		CASE(XPUT)
+		{
+			ON_OP();
+			updateIOGas();
+
+			uint8_t b = ++m_PC;
+			uint8_t c = ++m_PC;
+			xput(m_code[b], m_code[c]); ++m_PC;
```
gumb0: each statement on separate line

libevm/VM.cpp
```diff
@@ -643,7 +655,460 @@ void VM::interpretCases()
...
+		CONTINUE
+
+		CASE(XGET)
+		{
+			ON_OP();
+			updateIOGas();
+
+			uint8_t b = ++m_PC;
+			uint8_t c = ++m_PC;
+			xget(m_code[b], m_code[c]); ++m_PC;
```
gumb0: each statement on separate line

libevm/VM.cpp
```diff
@@ -1096,7 +1491,7 @@ void VM::interpretCases()
 #if EVM_HACK_DUP_64
 			*(uint64_t*)m_SPP = *(uint64_t*)(m_SP + n);
 #else
-			m_SPP[0] = m_SP[n];
+			new(m_SPP) u256(m_SP[n]);
```
gumb0: What is this change for?

libevm/VM.h
```diff
@@ -184,7 +185,69 @@ class VM: public VMFace
...
+	void xsstore (uint8_t);
+	void xvtowide(uint8_t);
+	void xwidetov(uint8_t);
+	void xpush   (uint8_t);
+	void xput    (uint8_t, uint8_t);
+	void xget    (uint8_t, uint8_t);
+	void xswizzle(uint8_t);
+	void xshuffle(uint8_t);
+	
+	u256 vtow(uint8_t _b, const u256& _in);
```
gumb0: `u256 const&`

libevm/VM.h
```diff
@@ -184,7 +185,69 @@ class VM: public VMFace
...
+	void wtov(uint8_t _b, u256 _in, u256& _o_out);
```
gumb0: Make this return `u256`, since you've made it for `vtow`.
Pass `_in` by const reference

libevm/VMConfig.h
```diff
@@ -22,17 +22,28 @@ namespace eth
...
-// EVM_USE_CONSTANT_POOL  - 256 constants unpacked and ready to assign to stack
+// EVM_USE_CONSTANT_POOL  - constants unpacked and ready to assign to stack
 //
-// EVM_REPLACE_CONST_JUMP - with pre-verified jumps to save runtime lookup
+// EVM_REPLACE_CONST_JUMP - pre-verified jumps to save runtime lookup
 //
 // EVM_TRACE              - provides various levels of tracing
 
+#ifndef EIP_615
+	#define EIP_615 false
```
gumb0: I would prefer defining these in .cmake files instead.
Then we'd be able to build with or without it just by changing cmake command line.

Or maybe it's not even necessary to modify .cmake files. maybe it would be enough just to delete these #defines, then try to build with `cmake .. -DEIP_615=true` ?

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+{
+	return pow2N((_b) & 0xf);
+}
+
+inline uint8_t elemType(uint8_t _b)
+{
+	return (_b) >> 4;
+}
+
+// in must be by reference because it is really just memory for a vector
```
gumb0: was that comment about output param? there's no output param anymore, comment may be removed, I guess

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+	default:
+		throwBadInstruction();
+	}
+}
+
+void VM::xmload (uint8_t _b)
+{
+	// n bytes of type t elements in memory vector
+	// goes onto stack element by element, LSB first
+	uint8_t* p = m_mem.data() + toInt15(m_SP[0]);
```
gumb0: can be made `uint8_t const*`

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+			int v = 0;
+			v = v8x32(m_SPP[0])[i];
+			p[--j] = (uint8_t)v;
+		}
+		break;
+	case 1:
+		for (int j = n,  i = n - 1; 0 <= i; --i)
+		{
+			int v = 0;
+			v = v8x32(m_SPP[0])[i];
```
gumb0: I guess this should be  `v = v16x16(m_SPP[0])[i];` ?

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+			p[--j] = (uint8_t)v;
+			v >>= 8;
+			p[--j] = (uint8_t)v;
+		}
+		break;
+	case 2:
+		for (int j = n,  i = n - 1; 0 <= i; --i)
+		{
+			int v = 0;
+			v = v8x32(m_SPP[0])[i];
```
gumb0: `v32x8` ?

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+			p[--j] = (uint8_t)v;
+			v >>= 8;
+			p[--j] = (uint8_t)v;
+		}
+		break;
+	case 3:
+		for (int j = n,  i = n - 1; 0 <= i; --i)
+		{
+			int v = 0;
+			v = v8x32(m_SPP[0])[i];
```
gumb0: `v64x4` ?

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+
+		switch (u)
+		{
+		case 0:
+			for (int i = 0; i < m; ++i)
+				v8x32 (m_SPP[0])[i] = v8x32(m_SP[0])[v8x32 (m_SP[1])[i] % 32];
+			break;
+		case 1:
+			for (int i = 0; i < m; ++i)
+				v16x16(m_SPP[0])[i] = v8x32(m_SP[0])[v16x16(m_SP[1])[i] % 16];
```
gumb0: Hm, not sure if this is correct, shouldn't source vector and result vector have the same type (which is `t`) ?
That is, I think left-hand side here should be `v8x32(m_SPP[0])[i]`

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+			throwBadInstruction();
+		}
+
+	case 1:
+
+		switch (u)
+		{
+		case 0:
+			for (int i = 0; i < m; ++i)
+				v8x32 (m_SPP[0])[i] = v16x16(m_SP[1])[v8x32 (m_SP[0])[i] % 32];
```
gumb0: The same problem here, assigning `uint16_t` to `uint8_t` doesn't look correct.
Left-hand side should be `v16x16`  ?

libevm/VMConfig.h
gcolvin: Presumably these will stay off in production until EIPs are accepted, and then be removed.  So in the meantime it's good to be able to switch configurations with minimal recompilation.  That's my main consideration.  cmake files might work. Rebuilding with with cmake at the tope would not.

libevm/VM.h
gcolvin: CHeck comments on implementation.  Arguments and returns for this and vtow() are as they need to be/

libevm/VM.cpp
gcolvin: Weird.  This compiled.

libevm/VM.cpp
gcolvin: ok

libevm/VM.cpp
gcolvin: ok

libevm/VM.cpp
gcolvin: The stack slot being copied into may no longer hold a u256, so we construct a new one in the memory, rather than assign.  I added a comment.

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+#define GT(  x1, x2) ((x1) > (x2))
+#define EQ(  x1, x2) ((x1) == (x2))
+#define ZERO(x1, x2) ((x1) == 0)
+#define AND( x1, x2) ((x1) & (x2))
+#define OR(  x1, x2) ((x1) | (x2))
+#define XOR( x1, x2) ((x1) ^ (x2))
+#define NOT( x1, x2) (~x1)
+#define SHR( x1, x2) ((x1) >> (x2))
+#define SHL( x1, x2) ((x1) << (x2))
+#define ROL( x1, x2) (((x1) << x2)|((x1) >> (sizeof(x1) * 8 - (x2))))
```
gcolvin: thanks

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+#define ROR( x1, x2) (((x1) >> x2)|((x1) << (sizeof(x1) * 8 - (x2))))
```
gcolvin: thanks

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+#define DIV( x1, x2) ((x1) / (x2))
+#define MOD( x1, x2) ((x1) % (x2))
+#define LT(  x1, x2) ((x1) < (x2))
+#define GT(  x1, x2) ((x1) > (x2))
+#define EQ(  x1, x2) ((x1) == (x2))
+#define ZERO(x1, x2) ((x1) == 0)
+#define AND( x1, x2) ((x1) & (x2))
+#define OR(  x1, x2) ((x1) | (x2))
+#define XOR( x1, x2) ((x1) ^ (x2))
+#define NOT( x1, x2) (~x1)
```
gcolvin: thanks

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+void VM::xoor (uint8_t _b) { EVALXOPU(OR,  _b); }
+void VM::xxor (uint8_t _b) { EVALXOPU(XOR, _b); }
+void VM::xnot (uint8_t _b) { EVALXOPU(NOT, _b); }
+void VM::xshr (uint8_t _b) { EVALXOPU(SHR, _b); }
+void VM::xsar (uint8_t _b) { EVALXOPS(SHR, _b); }
+void VM::xshl (uint8_t _b) { EVALXOPU(SHL, _b); }
+void VM::xrol (uint8_t _b) { EVALXOPU(ROL, _b); }
+void VM::xror (uint8_t _b) { EVALXOPU(ROR, _b); }
+
+inline uint8_t pow2N(uint8_t n)
```
gcolvin: sigh

libevm/VMSIMD.cpp
gcolvin: I regret I couldn't make these work as templates and avoid all these parentheses.

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+{
+	static uint8_t exp[6] = { 1, 2, 4, 8, 16, 32 };
```
gcolvin: Do we ever turn asserts on?  Where?  There are a lot of them that could be added to double-check everything for this EIP and 615 that is supposed to be validated.

libevm/VMSIMD.cpp
gcolvin: For this function it's _in that must be by reference.  You can't just copy the u256 in this situation.

libevm/VMSIMD.cpp
gcolvin: sigh

libevm/VM.h
gcolvin: sigh

libevm/VMSIMD.cpp
gcolvin: yep, thanks

libevm/VMSIMD.cpp
gcolvin: yep, thanks

libevm/VMSIMD.cpp
gcolvin: yep, thanks

libevm/VMSIMD.cpp
gcolvin: Spec says they can be different.  We could take up in the EIP issue whether that is a good idea.

libevm/VMSIMD.cpp
gcolvin: Same answer as above.

libevm/VM.cpp
gcolvin: No - this is EIP_615 code, which I haven't compiled for a while.

libevm/VMConfig.h
gcolvin: Looking at the code again, I still like this pattern.  It allows for external configuration (via cmake or whatever) - if a macro has a definition when compiled it is respected, and if it doesn't an appropriate default is chosen.  During development I just change the defaults to quickly switch configurations.

libevm/VM.cpp
gcolvin: Now it compiles.  Thanks.

libevm/VMSIMD.cpp
gumb0: Hm, ok, but the code contains only two bytes after XGET opcode, for the type of source vector and the type of get vector (indices vector).
Then this implementation considers result type to be equal get vector type, I find it weird.

For the most general implementation we probably should have 3 bytes following XGET opcode, for types of all 3 vectors?
I'll add a comment to EIP

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+
+		switch (u)
+		{
+		case 0:
+			for (int i = 0; i < m; ++i)
+				v8x32 (m_SPP[0])[v8x32(m_SP[1])[i] % 32] = v8x32(m_SP[0])[i];
+			break;
+		case 1:
+			for (int i = 0; i < m; ++i)
+				v16x16(m_SPP[0])[v8x32(m_SP[1])[i] % 16] = v8x32(m_SP[0])[i];
```
gumb0: And this implementation of XPUT considers source vector type to be the same as index vector type (which is `t`)
Why not consider source vector type be the same as result vector (which is `u`) ?

libevm/VM.cpp
```diff
@@ -1096,7 +1492,9 @@ void VM::interpretCases()
 #if EVM_HACK_DUP_64
 			*(uint64_t*)m_SPP = *(uint64_t*)(m_SP + n);
 #else
-			m_SPP[0] = m_SP[n];
+			// the stack slot being copied into may no longer hold a u256
+			// so we construct a new one in the memory, rather than assign
+			new(m_SPP) u256(m_SP[n]);
```
gumb0: So do I understand this correctly, that this is done to avoid copy assignment operator interpreting the data as `u256` and somehow corrupting it, since it can be not `u256` at all?

But this placement new still calls the copy constructor, and copy constructor will still interpret it as `u256`.

Maybe we should `memcpy` or `std:copy` the bytes here? Whis is against the rules since uint256 is not a POD...
But I don't see how placement new helps here

libevm/VM.h
gumb0: Ok, the convention for output params is `o_out`, without leading underscore

gcolvin: Ok.  But I wonder why the inconsistency?  Either all arguments should have a leading underscore or they shouldn't.

libevm/VMSIMD.cpp
gumb0: I see now, XGET really replaces each index in the stack word with a value from source vector, that's why index vector type and result vector type are the same.

libevm/VMSIMD.cpp
gumb0: Here I'm still confused, it reads only two vectors from the stack, but XPUT is supposed to take 3 vectors?

gcolvin: As discussed at the EIP I need to clean up the definition, and then I can be sure I have implemented the definition.  But this looks wrong, and your suggestion sound right, thanks.

libevm/VM.cpp
gcolvin: It's not the source that is the problem - it's a u256.  It's the destination that's the problem - it might be a vector.  The assignment operator would treat the destination as a u256, which might blow up.  Placement new treats it as raw memory.  (Then we are just down to the problem that the destination's destructor is not called.  I think we will need XPOP and XPUSH to fix that.)

gumb0: But source can  also be a vector? Or will validator not allow it?

gcolvin: Right - it's invalid.

gcolvin: Oh - we will need XDUP and XSWAP too.  Either that or implementations will be forced to keep track of the type of every stack slot, and this routine will involve checking the type of the source and destination.  In C++ we can fix that by making u256 a POD, but some other languages can't get that close to the metal.

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,729 @@
...
+	// for every index in the mask copy out the indexed value in one of the sources
+	switch (t)
+	{
+	case 0:
+		for (int i = 0; i < n; ++i)
+		{
+			int j = v8x32(m_SP[0]) [i];
+			v8x32 (m_SPP[0])[i] = j < n ? v8x32(m_SP[2]) [j] : v8x32 (m_SP[2])[j % n];
```
gumb0: This doesn't access source 1 vector, should be `j < n ? v8x32(m_SP[1]) [j] : v8x32 (m_SP[2])[j % n];` as I understand

gcolvin: I think so. thanks.

gcolvin: Yes, this is clearly wrong - the two sides of the condition should be different.  But the spec is unclear as to which is which, and the example in the spec is also clearly wrong.  So both need fixing.  I notice in this spec and others it unclear whether the "first" argument is the first item pushed on the stack or item on the top of the stack - SP[n] versus SP[0].

libevm/VMValidate.cpp
```diff
@@ -36,22 +40,23 @@ void VM::validate(ExtVMFace& _ext)
 	m_ext = &_ext;
 	initEntry();
 	size_t PC;
-	byte OP;
-	for (PC = 0; (OP = m_code[PC]); ++PC)
-		if  (OP == byte(Instruction::BEGINSUB))
+	Instruction OP;
```
gumb0: Are the changes in this file supposed to go into this PR ? They look not really related to SIMD

gcolvin: Changes for SIMD (and some previous ones) broke VMValidate, so I wanted it to compile again.  Changes are pretty much small stuff like changed signatures and casts.

libevm/VMSIMD.cpp
gumb0: They are on in debug build. I build debug all the time

gcolvin: OK.  RelDbg is usually enough for me.  I'll put more in for the next PR.

libevm/VMSIMD.cpp
```diff
@@ -41,28 +41,30 @@ inline a32x8  const& v32x8 (u256 const& _stack_item) { return (a32x8&) *(a32x8*)
 inline a16x16 const& v16x16(u256 const& _stack_item) { return (a16x16&)*(a16x16*)&_stack_item; }
 inline a8x32  const& v8x32 (u256 const& _stack_item) { return (a8x32&) *(a8x32*) &_stack_item; }
 
+class Pow2Bits enum { BITS_8, BITS_16, BITS_32, BITS_64 };
```
gumb0: Wait, this doesn't compile for me at all. Did you intend it to be `enum class`? Then `BITS_8` etc. will be scoped (should be `Pow2Bits::BITS_8` etc. everywhere) and implicit conversion between `uint8_t` and `Pow2Bits` won't work in switches.
Probably easier to use plain enum, although coding standards discourage it

gumb0: Naming: all-uppercase is used only for macros, members should be named something like `Bits8`, `Bits16` etc.

enum itself probably could be named `LaneWidth`

libevm/VMSIMD.cpp
```diff
@@ -169,35 +171,35 @@ u256 VM::vtow(uint8_t _b, const u256& _in)
 }
 
 // out must be by reference because it is really just memory for a vector
-void VM::wtov(uint8_t _b, u256 _in, u256& _o_out)
+void VM::wtov(uint8_t _type, u256 _in, u256& _o_out)
```
gumb0: Naming: `o_out`

libevm/VMSIMD.cpp
```diff
@@ -426,108 +426,107 @@ void VM::xpush(uint8_t _b)
 	}
 }
 
-void VM::xget(uint8_t _b, uint8_t _c)
+void VM::xget(uint8_t _src_type, uint8_t _idx_type)
```
gumb0: Naming: don't use underscores other than for prefix, should be `_srcType` & `_idxType`

libevm/VMSIMD.cpp
```diff
@@ -538,107 +537,106 @@ void VM::xget(uint8_t _b, uint8_t _c)
 	}
 }
 
-void VM::xput(uint8_t _b, uint8_t _c)
+void VM::xput(uint8_t _src_type, uint8_t _dst_type)
```
gumb0: Naming: `_srcType` & `_dstType`

gcolvin: Thought I'd search and destroyed all of these too...

Still - better than _b and _c.

libevm/VMSIMD.cpp
gcolvin: Thought I'd search and replaced all of these...

libevm/VMSIMD.cpp
gcolvin: Thought I'd compiled it, but I did intend enum class, and I was pleasantly surprised not to see the problems you mentioned, so the config macro was probably turned off already.  A simple enum is what I'm sure compiled, and it makes sense in the scope of just one file.  I'll fix up the names.

gcolvin:
Travis is going to fail soon - Homebrew is broken.  No merging until that is fixed.

libevm/VMSIMD.cpp
gcolvin: A grep on `_[a-z]+_[a-z]+` also smoked out some `_o_out` and `_stack_item` hiding in plain sight.

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,726 @@
...
+{
+	// goes onto stack element by element, LSB first
+	uint8_t const* p = m_mem.data() + toInt15(m_SP[0]);
+	uint8_t const count = laneCount(_type);
+	uint8_t const width = laneWidth(_type);
+
+	switch (width)
+	{
+	case Bits8:
+		for (int j = 1,  i = count - 1; 0 <= i; --i)
```
gumb0: Is `j` initialization correct here and in other cases? it will go negative here if count > 2

It was initializeed as `j = n` in earlier commits

libevm/VMSIMD.cpp
```diff
@@ -0,0 +1,726 @@
...
+	// n bytes of type t elements in stack vector
+	// goes onto memory by element, LSB first
+	uint8_t *p = m_mem.data() + toInt15(m_SP[0]);
+	uint8_t const count = laneCount(_type);
+	uint8_t const width = laneWidth(_type);
+
+	switch (width)
+	{
+	case Bits8:
+		for (int j = 1,  i = count - 1; 0 <= i; --i)
```
gumb0: Here too, should it be `j = count` ?

gcolvin:
@gumbo all checks passed as I slept

libevm/VMSIMD.cpp
gcolvin: I need to dig into this after the AllCoreDevs meeting.  This looks like the remains of a merge conflict that I reconciled badly.  It should be `n`, I think I agree, but `n` is gone.

gcolvin:
Thanks, @gumb0 !

PR #4244, opened 2017-07-10
Clear storage when overwriting a contract

### LINKED ISSUE #4237: Contract creation at an existing address should clear the storage
The Yellow Paper specifies that the contract creation should clear the storage, but `cpp-ethereum` seems to preserve the storage entries.

This is one reason of https://github.com/ethereum/go-ethereum/issues/14756

pirapira:
Fixes #4237

### DIFF
--- a/libethereum/Account.h
+++ b/libethereum/Account.h
@@ -133,6 +133,12 @@ public:
 	/// to the trie later.
 	void setStorage(u256 _p, u256 _v) { m_storageOverlay[_p] = _v; changed(); }
 
+	/// Empty the storage.  Used when a contract is overwritten.
+	void clearStorage() { m_storageOverlay.clear(); m_storageRoot = EmptyTrie; changed(); }
+
+	/// Set the storage root.  Used when clearStorage() is reverted.
+	void setStorageRoot(h256 const& _root) { m_storageOverlay.clear(); m_storageRoot = _root; changed(); }
+
 	/// Set a key/value pair in the account's storage to a value that is already present inside the
 	/// database.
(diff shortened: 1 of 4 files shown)


pirapira:
I need to fill some blockchain tests.

pirapira:
Now this seems ready to me.

gumb0:
Hm, this way we don't remove the storage trie from the state DB, but just replace the storage root in the account with empty trie root. And we have an orphaned trie in the DB. (This happens in `commit` in State.h)

Not sure how big of a problem this is, it's after all only handling unrealistic scenario of address collision, as I understand...

To fix we'd probably could add a flag to the Account to show that storage was removed.

pirapira:
This kind of change might be reverted during a transaction.  So we need to "garbage-collect" at the end of a transaction.  Let me read around `State.h`.

gumb0:
All the changes are only in `State`'s hashes until transaction end. Then `commit` makes real changes to the state.

Yeah, with CREATE2 it's not only about address collision anymore, so I think we should fix it.

gumb0:
On the other hand I'm not sure whether we ever really delete the storage values from the database...

pirapira:
Hm.  When the chain-reorg happens, old tries come in useful.

gumb0:
The problem might still be there I think, but it's only wasting some disk space in the rare unrealistic case of a contract writing to storage, then address collision in the same block, so it's fine not to complicate the code with optimizing it.

Looks like we don't do this anyway in the case of contract writing to storage and then suiciding in another transaction of the same block.

pirapira:
I created an issue about that https://github.com/ethereum/cpp-ethereum/issues/4264

PR #4245, opened 2017-07-10
exportLog() now takes a const reference, yielding some speedup

pirapira:
Before this change:
```
ETHEREUM_TEST_PATH="../../tests" test/testeth -t "StateTestsGeneral"  86,34s user 1,35s system 74% cpu 1:58,06 total
```
After this change:
```
ETHEREUM_TEST_PATH="../../tests" test/testeth -t "StateTestsGeneral"  67,13s user 1,12s system 98% cpu 1:09,04 total
```

### DIFF
--- a/test/tools/libtesteth/TestHelper.cpp
+++ b/test/tools/libtesteth/TestHelper.cpp
@@ -150,7 +150,7 @@ std::vector<eth::Network> const& getNetworks()
 	return networks;
 }
 
-json_spirit::mArray exportLog(eth::LogEntries _logs)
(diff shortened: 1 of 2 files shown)


pirapira:
. @fjl was profiling `testeth`, and he told me this function is super slow.

fjl:
I'm pretty sure that this is not why it's slow ;). Still nicer to pass reference though.

pirapira:
@fjl yeah, I'm still looking for something like "opening the same file 100 times as the loop loops".

PR #4247, opened 2017-07-11
libdevcore: speed up toHex, remove _w, _prefix parameters

fjl:
toHex was slow enough to show up in test CPU profiles. Make it disappear
from the profile.

Also remove weird _w parameter. All callers ignore it or use 2 (the
default value) so they can pass _prefix. Remove _prefix too because it's
shorter to say "0x" + toHex(...) if a prefix is needed.

### DIFF
--- a/libdevcore/CommonData.h
+++ b/libdevcore/CommonData.h
@@ -42,44 +42,36 @@ enum class WhenError
 	Throw = 1,
 };
 
-namespace
+template <class Iterator>
+inline std::string toHex(Iterator _it, Iterator _end, std::string _prefix)
 {
+	typedef std::iterator_traits<Iterator> traits;
+	static_assert(sizeof(typename traits::value_type) == 1, "toHex needs byte-sized element type");
 
-const char* hexdigits = "0123456789abcdef";
-
-template <class T>
-inline void writeHex(T const& _data, std::string& o_hex, size_t _off)
-{
-	for (byte i: _data)
+	static char const* hexdigits = "0123456789abcdef";
+	size_t off = _prefix.size();
+	std::string hex(std::distance(_it, _end)*2 + off, '0');
+	hex.replace(0, off, _prefix);
+	for (; _it != _end; _it++)
 	{
-		o_hex[_off++] = hexdigits[i >> 4];
-		o_hex[_off++] = hexdigits[i & 0x0f];
+		hex[off++] = hexdigits[*_it >> 4];
+		hex[off++] = hexdigits[*_it & 0x0f];
 	}
-}
-
+	return hex;
 }
 
 /// Convert a series of bytes to the corresponding hex string.
 /// @example toHex("A\x69") == "4169"
-template <class T>
-std::string toHex(T const& _data)
+template <class T> std::string toHex(T const& _data)
 {
-	static_assert(sizeof(typename T::value_type) == 1, "toHex needs byte-sized element type");
-	std::string hex(_data.size()*2, '0');
-	writeHex(_data, hex, 0);
-	return hex;
+	return toHex(_data.begin(), _data.end(), "");
 }
 
 /// Convert a series of bytes to the corresponding hex string with 0x prefix.
 /// @example toHexPrefix("A\x69") == "0x4169"
-template <class T>
-std::string toHexPrefix(T const& _data)
+template <class T> std::string toHexPrefix(T const& _data)
 {
-	static_assert(sizeof(typename T::value_type) == 1, "toHexPrefix needs byte-sized element type");
-	std::string hex(_data.size()*2 + 2, '0');
-	hex.replace(0, 2, "0x");
-	writeHex(_data, hex, 2);
-	return hex;
+	return toHex(_data.begin(), _data.end(), "0x");
 }
 
 /// Converts a (printable) ASCII hex string into the corresponding byte stream.
--- a/libdevcore/CommonData.h
+++ b/libdevcore/CommonData.h
@@ -68,8 +68,8 @@ template <class T> std::string toHex(T const& _data)
 }
 
 /// Convert a series of bytes to the corresponding hex string with 0x prefix.
-/// @example toHexPrefix("A\x69") == "0x4169"
-template <class T> std::string toHexPrefix(T const& _data)
+/// @example toHexPrefixed("A\x69") == "0x4169"
+template <class T> std::string toHexPrefixed(T const& _data)
 {
 	return toHex(_data.begin(), _data.end(), "0x");
 }
@@ -184,9 +184,9 @@ inline std::string toCompactHex(u256 val, unsigned _min = 0)
 	return toHex(toCompactBigEndian(val, _min));
 }
 
-inline std::string toCompactHexPrefix(u256 val, unsigned _min = 0)
+inline std::string toCompactHexPrefixed(u256 val, unsigned _min = 0)
 {
-	return toHexPrefix(toCompactBigEndian(val, _min));
+	return toHexPrefixed(toCompactBigEndian(val, _min));
 }
 
 // Algorithms for string and string-like collections.
--- a/libdevcore/CommonData.h
+++ b/libdevcore/CommonData.h
@@ -54,7 +54,7 @@ std::string toHex(Iterator _it, Iterator _end, std::string _prefix)
 	hex.replace(0, off, _prefix);
 	for (; _it != _end; _it++)
 	{
-		hex[off++] = hexdigits[*_it >> 4];
+		hex[off++] = hexdigits[(*_it >> 4) & 0x0f];
 		hex[off++] = hexdigits[*_it & 0x0f];
 	}
 	return hex;
--- a/libdevcore/CommonData.h
+++ b/libdevcore/CommonData.h
@@ -44,7 +44,19 @@ enum class WhenError
 
 namespace
 {
-	const char* hexdigits = "0123456789abcdef";
+
+const char* hexdigits = "0123456789abcdef";
+
+template <class T>
+inline void writeHex(T const& _data, std::string& o_hex, size_t _off)
+{
+	for (byte i: _data)
+	{
+		o_hex[_off++] = hexdigits[i>>4];
+		o_hex[_off++] = hexdigits[i&0x0f];
+	}
+}
+
 }
 
 /// Convert a series of bytes to the corresponding string of hex duplets.
@@ -54,12 +66,19 @@ std::string toHex(T const& _data)
 {
 	static_assert(sizeof(typename T::value_type) == 1, "toHex needs byte-sized element type");
 	std::string hex(_data.size()*2, '0');
-	size_t off = 0;
-	for (auto i: _data)
-	{
-		hex[off++] = hexdigits[(byte)(i)>>4];
-		hex[off++] = hexdigits[(byte)(i)&0x0f];
-	}
+	writeHex(_data, hex, 0);
+	return hex;
+}
+
+/// Convert a series of bytes to the corresponding string of hex duplets.
+/// @example toHex("A\x69") == "4169"
+template <class T>
+std::string toHexPrefix(T const& _data)
+{
+	static_assert(sizeof(typename T::value_type) == 1, "toHexPrefix needs byte-sized element type");
+	std::string hex(_data.size()*2 + 2, '0');
+	hex.replace(0, 2, "0x");
+	writeHex(_data, hex, 2);
 	return hex;
 }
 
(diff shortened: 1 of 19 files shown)


libdevcore/CommonData.h
```diff
@@ -42,22 +42,24 @@ enum class WhenError
...
-std::string toHex(T const& _data, int _w = 2, HexPrefix _prefix = HexPrefix::DontAdd)
+std::string toHex(T const& _data)
 {
-	std::ostringstream ret;
-	unsigned ii = 0;
-	for (auto i: _data)
-		ret << std::hex << std::setfill('0') << std::setw(ii++ ? 2 : _w) << (int)(typename std::make_unsigned<decltype(i)>::type)i;
-	return (_prefix == HexPrefix::Add) ? "0x" + ret.str() : ret.str();
+	size_t len = _data.size() * sizeof(typename T::value_type);
+	const byte* data = reinterpret_cast<const byte*>(_data.data());
```
chriseth: If you do that, then please restrict the template to be more specific.

libdevcore/CommonIO.h
```diff
@@ -77,7 +77,7 @@ std::string memDump(bytes const& _bytes, unsigned _width = 8, bool _html = false
 template <class S, class T> struct StreamOut { static S& bypass(S& _out, T const& _t) { _out << _t; return _out; } };
 template <class S> struct StreamOut<S, uint8_t> { static S& bypass(S& _out, uint8_t const& _t) { _out << (int)_t; return _out; } };
 
-inline std::ostream& operator<<(std::ostream& _out, bytes const& _e) { _out << toHex(_e, 2, HexPrefix::Add); return _out; }
+inline std::ostream& operator<<(std::ostream& _out, bytes const& _e) { _out << "0x" << toHex(_e); return _out; }
```
chriseth: Note that this might make a difference depending on the stream. I think some logging output stream we use weirdly inserts spaces between two elements.

libweb3jsonrpc/Debug.cpp
```diff
@@ -92,8 +92,8 @@ Json::Value Debug::debug_traceTransaction(string const& _txHash, Json::Value con
 		Executive e(s, block, t.transactionIndex(), m_eth.blockChain());
 		e.setResultRecipient(er);
 		Json::Value trace = traceTransaction(e, t, _json);
-		ret["gas"] = toHex(t.gas(), HexPrefix::Add);
-		ret["return"] = toHex(er.output, 2, HexPrefix::Add);
+		ret["gas"] = "0x" + toHex(t.gas());
```
chriseth: Perhaps it's better to have a function `toHexWithPrefix` that already allocates the prefix inside the string, saving a copy operation.

libdevcore/CommonData.h
fjl: Can you tell me how to do it?

The problem is that `toHex` is used with many different types: vectors, vector_refs, arrays, strings etc. Ideally we'd have a way of saying that `data()` should return a byte-sized type. We cannot say `data` returns `byte` because it breaks strings. I arrived at this because it felt like the easiest solution and also works with stuff like `std::vector<uint32_t>`.

libdevcore/CommonIO.h
fjl: fixed

libweb3jsonrpc/Debug.cpp
fjl: I tried this earlier, but IMHO it's less elegant for very little gain.

fjl: If you must have it this way, I can commit the earlier version that leaves the _prefix parameter in place.

libdevcore/CommonData.h
fjl: Maybe `static_assert` could be used to ensure that `sizeof(typename T::value_type) == 1`.

fjl: fixed

fjl:
@chriseth PTAL

libdevcore/CommonData.h
```diff
@@ -42,22 +42,44 @@ enum class WhenError
...
-/// @param _w specifies the width of the first of the elements. Defaults to two - enough to represent a byte.
+
+const char* hexdigits = "0123456789abcdef";
+
+template <class T>
+inline void writeHex(T const& _data, std::string& o_hex, size_t _off)
+{
+	for (byte i: _data)
+	{
+		o_hex[_off++] = hexdigits[i>>4];
```
chriseth: Potential buffer overflow if used incorrectly. Can you add an assert at the beginning of the loop?

libdevcore/CommonData.h
```diff
@@ -42,22 +42,44 @@ enum class WhenError
...
+		o_hex[_off++] = hexdigits[i&0x0f];
```
chriseth: Could you please add one space character around each binary operator?

libdevcore/CommonData.h
```diff
@@ -42,22 +42,44 @@ enum class WhenError
...
-enum class HexPrefix
+namespace
 {
-	DontAdd = 0,
-	Add = 1,
-};
-/// Convert a series of bytes to the corresponding string of hex duplets.
-/// @param _w specifies the width of the first of the elements. Defaults to two - enough to represent a byte.
+
+const char* hexdigits = "0123456789abcdef";
```
gumb0: Should be `char const*` according to coding standards

gumb0: Move this and `writeHex` into CommonData.cpp, anonymous namespace stuff shouldn't be in the header

libdevcore/CommonData.h
```diff
@@ -42,22 +42,44 @@ enum class WhenError
...
+}
+
+/// Convert a series of bytes to the corresponding hex string.
 /// @example toHex("A\x69") == "4169"
 template <class T>
-std::string toHex(T const& _data, int _w = 2, HexPrefix _prefix = HexPrefix::DontAdd)
+std::string toHex(T const& _data)
+{
+	static_assert(sizeof(typename T::value_type) == 1, "toHex needs byte-sized element type");
+	std::string hex(_data.size()*2, '0');
```
gumb0: Could you move this string declaration into `writeHex` and return it from there? Instead of having it as a input&output parameter which is really confusing

libdevcore/CommonData.h
fjl: Hmm wait. You didn't complain about this when the loop was inside toHex. Can I just move it there?

libdevcore/CommonData.h
```diff
@@ -42,22 +42,44 @@ enum class WhenError
...
+		o_hex[_off++] = hexdigits[i&0x0f];
+	}
+}
+
+}
+
+/// Convert a series of bytes to the corresponding hex string.
 /// @example toHex("A\x69") == "4169"
 template <class T>
-std::string toHex(T const& _data, int _w = 2, HexPrefix _prefix = HexPrefix::DontAdd)
```
gumb0: I like removing `_w` & having separate function instead of `_prefix` but I remember @chfast resising the similar change in one of the recent PRs

libdevcore/CommonData.h
```diff
@@ -42,22 +42,44 @@ enum class WhenError
...
+	static_assert(sizeof(typename T::value_type) == 1, "toHex needs byte-sized element type");
+	std::string hex(_data.size()*2, '0');
+	writeHex(_data, hex, 0);
+	return hex;
+}
+
+/// Convert a series of bytes to the corresponding hex string with 0x prefix.
+/// @example toHexPrefix("A\x69") == "0x4169"
+template <class T>
+std::string toHexPrefix(T const& _data)
```
gumb0: I would name it `toHexPrefixed` perhaps (it doesn't return a prefix)

gumb0:
So how much faster did it become?

fjl:
Not sure. I made profiles of `testeth ... --filltests` before and after an earlier version of this change and `toHex` didn't appear anymore.

libdevcore/CommonData.h
fjl: The name has to be shorter than `"0x" + toHex`

gumb0:
I merged today some fix using the old version of `toHex`, you should better rebase I think

libdevcore/CommonData.h
gumb0: Why? It should be easier to read, not less to type

libdevcore/CommonData.h
chriseth: If the creation of the string that is inserted into is in the same function (and not too far away inside that function), it is fine. The problem is that this function can be used with `o_hex` strings that are too short.

gumb0: Yeah, better to create this string in this same function and return it, it's not less efficient

fjl:
It's already rebased on that.

libdevcore/CommonData.h
fjl: It's really hard to move this because of the template.

fjl: If you have a solution, please help out ;)

gumb0: Oh I see, didn'n notice the template first, sorry. 
Then  leave it here, but I don't see any sense in putting it into anonymous namespace. I'd leave `writeHex` in `dev` and declare `hexdigits` as a static const inside `writeHex`

fjl:
PTAL

libdevcore/CommonData.h
```diff
@@ -42,22 +42,36 @@ enum class WhenError
...
-};
-/// Convert a series of bytes to the corresponding string of hex duplets.
-/// @param _w specifies the width of the first of the elements. Defaults to two - enough to represent a byte.
+	typedef std::iterator_traits<Iterator> traits;
+	static_assert(sizeof(typename traits::value_type) == 1, "toHex needs byte-sized element type");
+
+	static char const* hexdigits = "0123456789abcdef";
+	size_t off = _prefix.size();
+	std::string hex(std::distance(_it, _end)*2 + off, '0');
+	hex.replace(0, off, _prefix);
```
gumb0: Instead of getting prefix size, allocating additional space for it, then replacing, I'd do just `return _prefix + hex;`

libdevcore/CommonData.h
```diff
@@ -165,17 +179,14 @@ inline std::string toCompactBigEndianString(T _val, unsigned _min = 0)
 	return ret;
 }
 
-/// Convenience function for conversion of a u256 to hex
-inline std::string toHex(u256 val, HexPrefix prefix = HexPrefix::DontAdd)
+inline std::string toCompactHex(u256 val, unsigned _min = 0)
```
gumb0: Rename `val` to `_val` please

libdevcore/CommonData.h
```diff
@@ -165,17 +179,14 @@ inline std::string toCompactBigEndianString(T _val, unsigned _min = 0)
...
 {
-	std::string str = toHex(toBigEndian(val));
-	return (prefix == HexPrefix::Add) ? "0x" + str : str;
+	return toHex(toCompactBigEndian(val, _min));
 }
 
-inline std::string toCompactHex(u256 val, HexPrefix prefix = HexPrefix::DontAdd, unsigned _min = 0)
+inline std::string toCompactHexPrefixed(u256 val, unsigned _min = 0)
```
gumb0: Rename `val` to `_val` please

rlp/main.cpp
```diff
@@ -132,28 +132,28 @@ class RLPStreamer
...
-				m_out << (m_prefs.hexPrefix ? "0x" : "") << toHex(toCompactBigEndian(_d.toInt<bigint>(RLP::LaissezFaire), 1), 1);
-			else
-				m_out << _d.toInt<bigint>(RLP::LaissezFaire);
+			m_out << _d.toInt<bigint>(RLP::LaissezFaire);
 		else if (_d.isData() || (_d.isInt() && m_prefs.stringInts))
 			if (m_prefs.forceString || (!m_prefs.forceHex && isAscii(_d.toString())))
 				m_out << escaped(_d.toString(), m_prefs.escapeAll);
 			else
-				m_out << "\"" << (m_prefs.hexPrefix ? "0x" : "") << toHex(_d.toBytes()) << "\"";
+				m_out << "\"0x" << toHex(_d.toBytes()) << "\"";
```
gumb0: `toHexPrefixed` can be used here?

libdevcore/CommonData.h
fjl: I added `toHexPrefixed` because @chriseth [requested it](https://github.com/ethereum/cpp-ethereum/pull/4247#discussion_r126687705). The [first version](https://github.com/ethereum/cpp-ethereum/pull/4247/commits/22f0532350c508a214f2665bcd0695de3a534e65) did this.

gumb0: Hm, ok, I agree with your comment there, that it's less elegant for little gain

pirapira:
I restarted the Mac build in Travis.

@chriseth is your change request still standing?

libdevcore/CommonData.h
```diff
@@ -42,22 +42,36 @@ enum class WhenError
...
+	typedef std::iterator_traits<Iterator> traits;
+	static_assert(sizeof(typename traits::value_type) == 1, "toHex needs byte-sized element type");
+
+	static char const* hexdigits = "0123456789abcdef";
+	size_t off = _prefix.size();
+	std::string hex(std::distance(_it, _end)*2 + off, '0');
+	hex.replace(0, off, _prefix);
+	for (; _it != _end; _it++)
+	{
+		hex[off++] = hexdigits[*_it >> 4];
```
chriseth: Could you please re-add the conversion to unsigned, just in case we ever use that with a signed type?

`(typename std::make_unsigned<decltype(i)>::type)i`

chriseth: Hm, perhaps `(*_it >> 4) & 0x0f` would already do.

chriseth:
Approving because I won't be available next week.

PR #4254, opened 2017-07-12
[WIP] libtesteth: add transaction RLP to state tests

fjl:
The general state test format changes so the post state includes the RLP
of the (signed) transaction. This makes it easier to run the tests
because no signing is required. It's also slightly more powerful because
any signature scheme can be tested.

The first few commits are refactoring changes.


winsvega:
I think those changes are breaking other test suites.  
Have you checked that  TransactionTests, BlockchainTests, VMTests are not affected ?

winsvega:
here is the way to add transaction RLP 
https://github.com/ethereum/cpp-ethereum/pull/4263

could we split refactoring commits into smaller PR's ? 
it is hard to see that this does not brake existing test suites

fjl:
I can send the refactoring as a separate PR if you want.

fjl:
I will reopen this some other time.

PR #4255, opened 2017-07-12
Accept EIP155-style transactions only after the Spurious Dragon fork

pirapira:


pirapira:
@winsvega weirdly, I didn't have to change any tests.  Maybe there should be a test about using the EIP-155 format before the Spurious Dragon block number.

winsvega:
the tests are already configured to test protection on eip158 block
https://github.com/ethereum/tests/tree/develop/src/TransactionTestsFiller/EIP155

seems like a misunderstanding with fork names. tests says it is 155 but actually it is 158  block number

winsvega:
we might want to add tests for new transactions being rejected on < eip158 fork number

pirapira:
@winsvega I created an issue about adding such a test https://github.com/ethereum/tests/issues/233

PR #4273, opened 2017-07-18
Smaller header files

pirapira:
I'm trying to make compilation units smaller.  Currently I'm removing `#include`s from header files.  Then I had to add some `#include`s in cpp files, but that should be fine.

I also moved one function definition from `State.h` to `State.cpp`.  In this case (a template instantiated in only two different ways), I guess it makes sense to do so.


libdevcore/Common.cpp
```diff
@@ -31,7 +31,9 @@ namespace dev
 
 char const* Version = ETH_PROJECT_VERSION;
 
+const bytes NullBytes;
```
gumb0: While we're here let's change it to `bytes const` according to coding standards.
Nest lines, too.

libdevcore/Common.h
```diff
@@ -69,7 +69,7 @@ namespace dev
 
 extern char const* Version;
 
-static const std::string EmptyString;
+extern const std::string EmptyString;
```
gumb0: `std::string const`

libdevcore/CommonIO.h
```diff
@@ -82,7 +81,7 @@ template <class T> inline std::ostream& operator<<(std::ostream& _out, std::vect
 template <class T, std::size_t Z> inline std::ostream& operator<<(std::ostream& _out, std::array<T, Z> const& _e);
 template <class T, class U> inline std::ostream& operator<<(std::ostream& _out, std::pair<T, U> const& _e);
 template <class T> inline std::ostream& operator<<(std::ostream& _out, std::list<T> const& _e);
-template <class T1, class T2, class T3> inline std::ostream& operator<<(std::ostream& _out, std::tuple<T1, T2, T3> const& _e);
+// template <class T1, class T2, class T3> inline std::ostream& operator<<(std::ostream& _out, std::tuple<T1, T2, T3> const& _e);
```
gumb0: Let's remove this?

libdevcore/Log.h
```diff
@@ -167,17 +156,17 @@ class LogOutputStreamBase
 	void append(u160 const& _t) { m_sstr << EthNavy << _t << EthReset; }
 	void append(double _t) { m_sstr << EthBlue << _t << EthReset; }
 	template <unsigned N> void append(FixedHash<N> const& _t) { m_sstr << EthTeal "#" << _t.abridged() << EthReset; }
-	void append(h160 const& _t) { m_sstr << EthRed "@" << _t.abridged() << EthReset; }
-	void append(h256 const& _t) { m_sstr << EthCyan "#" << _t.abridged() << EthReset; }
-	void append(h512 const& _t) { m_sstr << EthTeal "##" << _t.abridged() << EthReset; }
+	void append(h160 const& _t) { m_sstr << EthRed "@" << _t.h160::abridged() << EthReset; }
```
gumb0: Why is this change needed?

libdevcore/Log.h
```diff
@@ -268,6 +261,8 @@ class LogOutputStream: LogOutputStreamBase
 
 	LogOutputStream& operator<<(LogTag _t) { m_logTag = _t; return *this; }
 
+	LogOutputStream& operator<<(const char* _t) { if (Id::verbosity <= g_logVerbosity) { if (_AutoSpacing && m_sstr.str().size() && m_sstr.str().back() != ' ') m_sstr << " "; append(_t); } return *this; }
```
gumb0: Do we need this? Wouldn't `append` overload for `const char*` above be enough for this?

libdevcore/RangeMask.h
```diff
@@ -94,8 +94,8 @@ class RangeMask
 	/// reference to itself.
 	RangeMask& invert() { return *this = inverted(); }
 
-	template <class S> RangeMask operator-(S const& _m) const { auto ret = *this; return ret -= _m; }
-	template <class S> RangeMask& operator-=(S const& _m) { return invert().unionWith(_m).invert(); }
+	RangeMask operator-(RangeMask const& _m) const { auto ret = *this; return ret -= _m; }
```
gumb0: Looks like we won't be able to do `-` and `-=` with a single number after this... Maybe not a problem, but one convenient usage less

libdevcore/Common.cpp
pirapira: By nesting, you mean adding indentation?

libdevcore/Log.h
pirapira: That was a partial attempt to resolve templates manually, but I don't do this in this PR.

libdevcore/Log.h
pirapira: This one, too, was resolving templates manually.  This one is worse so I revert this.

libdevcore/RangeMask.h
pirapira: Suddenly I discovered this `RangeMask` is never used.  I'll remove it all.

libdevcore/Common.cpp
gumb0: I meant "next lines", sorry

libdevcore/RangeMask.h
gumb0: Please don't remove it yet, I use it in fast-sync branch

libdevcore/RangeMask.h
```diff
@@ -230,8 +230,7 @@ template <class T> inline std::ostream& operator<<(std::ostream& _out, RangeMask
 	return _out;
 }
 
-template <class T>
-RangeMask<T>& RangeMask<T>::unionWith(typename RangeMask<T>::Range const& _m)
+RangeMask& RangeMask::unionWith(typename RangeMask::Range const& _m)
```
gumb0: I think `typename` may be removed here now, since RangeMask is not a template

libdevcore/TrieHash.h
```diff
@@ -21,8 +21,9 @@
 
 #pragma once
 
-#include "Common.h"
-#include "FixedHash.h"
+#include <vector>
```
gumb0: `#include`s for stdlib should be after ones for our libraries, let's do it right it at least for new code

libdevcore/Log.h
```diff
@@ -240,6 +229,10 @@ class LogOutputStreamBase
 		append(_t.second);
 		m_sstr << EthPurple ")" EthReset;
 	}
+	template <class T> void append(const char* _t)
```
gumb0: Not sure if we still need it in this PR, but it should be either non-template or specialization of `append` template below

libdevcore/RangeMask.h
```diff
@@ -1,294 +0,0 @@
-/*
```
gumb0: Retuurn it 😭

pirapira:
The AppVeyor build timeouts.  Maybe I should remove more until it passes.

gumb0:
It's probably the tests that take the majority of AppVeyor time.

(I've heard that we have a setting of 2 hours time for AppVeyor but it sometimes gets reset to 1 hour and then it timeouts. But I don't have access to this setting)

PR #4275, opened 2017-07-19
EIP658: Status code in receipt

pirapira:
Spec: https://github.com/ethereum/EIPs/pull/658

### DIFF
--- a/libethereum/Block.cpp
+++ b/libethereum/Block.cpp
@@ -864,7 +864,14 @@ bool Block::sealBlock(bytesConstRef _header)
 h256 Block::stateRootBeforeTx(unsigned _i) const
 {
 	_i = min<unsigned>(_i, m_transactions.size());
-	return (_i > 0 ? receipt(_i - 1).stateRoot() : m_previousBlock.stateRoot());
+	try
+	{
+		return (_i > 0 ? receipt(_i - 1).stateRoot() : m_previousBlock.stateRoot());
+	}
+	catch(TransactionReceiptVersionError const&)
+	{
+		return {};
+	}
 }
 
 LogBloom Block::logBloom() const
--- a/libethereum/Block.cpp
+++ b/libethereum/Block.cpp
@@ -868,7 +868,7 @@ h256 Block::stateRootBeforeTx(unsigned _i) const
 	{
 		return (_i > 0 ? receipt(_i - 1).stateRoot() : m_previousBlock.stateRoot());
 	}
-	catch(TransactionReceiptVersionError const&)
+	catch (TransactionReceiptVersionError const&)
 	{
 		return {};
 	}
--- a/libethereum/TransactionReceipt.cpp
+++ b/libethereum/TransactionReceipt.cpp
@@ -22,9 +22,7 @@
 #include "TransactionReceipt.h"
 #include <libethcore/Exceptions.h>
 
-#include <boost/variant/static_visitor.hpp>
 #include <boost/variant/get.hpp>
-
 
 using namespace std;
 using namespace dev;
@@ -38,10 +36,13 @@ TransactionReceipt::TransactionReceipt(bytesConstRef _rlp)
 
 	if (!r[0].isData())
 		BOOST_THROW_EXCEPTION(InvalidTransactionReceiptFormat());
+
 	if (r[0].size() == 32)
 		m_statusCodeOrStateRoot = (h256)r[0];
 	else if (r[0].size() == 1)
 		m_statusCodeOrStateRoot = (uint8_t)r[0];
+	else
+		BOOST_THROW_EXCEPTION(InvalidTransactionReceiptFormat());
 
 	m_gasUsed = (u256)r[1];
 	m_bloom = (LogBloom)r[2];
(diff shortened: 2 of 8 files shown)


libethereum/TransactionReceipt.h
```diff
@@ -52,7 +58,12 @@ class TransactionReceipt
 	bytes rlp() const { RLPStream s; streamRLP(s); return s.out(); }
 
 private:
-	h256 m_stateRoot;
+	bool m_hasStatusCode = false;
+	union
```
gumb0: I'd suggest trying to use `boost::variant` for this, it kind of encapsulates union + type variable

gumb0:
Also `toJson` for receipt in libweb3jsonrpc/JsonHelper.cpp needs a fix

gumb0:
More exactly, `dev::eth::createIntermediateState` in State.cpp assumes that intermediate state root from receipt is either valid hash or 0 (then replays transactions) - we could also handle exception in it instead of modifying `Block::stateRootBeforeTx`

pirapira:
I started using `boost::variant` and changed the `stateRoot()` usage.

libethereum/TransactionReceipt.cpp
```diff
@@ -22,54 +22,122 @@
...
 
 void TransactionReceipt::streamRLP(RLPStream& _s) const
 {
-	if (m_stateRoot)
-		_s.appendList(4) << m_stateRoot;
-	else
-		_s.appendList(3);
-		
-	_s << m_gasUsed << m_bloom;
+	struct appenderVisitor : boost::static_visitor<void>
```
gumb0: Wow, couldn't we just do something like
```
    if (m_statusCodeOrStateRoot.which() == 0)
        _s  << boost::get<uint8_t>(m_statusCodeOrStateRoot);
    else
        _s  << boost::get<h256>(m_statusCodeOrStateRoot);
```

libethereum/TransactionReceipt.cpp
```diff
@@ -22,54 +22,122 @@
...
+		}
+	};
+	return boost::apply_visitor(statusCodeVisitor(), m_statusCodeOrStateRoot);
+}
+
+h256 const& TransactionReceipt::stateRoot() const
+{
+	try
+	{
+		return boost::get<h256>(m_statusCodeOrStateRoot);
```
gumb0: In `stateRoot` you use different approach than in `hasStatusCode` & `stateRoot`, choose one for consistency (the one without visitor is easier).

I would actually prefer just
```
    if (m_statusCodeOrStateRoot.which() == 1)
      return boost::get<h256>(m_statusCodeOrStateRoot);
    else
      BOOST_THROW_EXCEPTION(TransactionReceiptVersionError());
```

pirapira:
I made everything into `if (hasStatusCode()) ... else ...`.

libethereum/TransactionReceipt.cpp
```diff
@@ -22,54 +22,88 @@
 #include "TransactionReceipt.h"
 #include <libethcore/Exceptions.h>
 
+#include <boost/variant/static_visitor.hpp>
```
gumb0: This one is not needed now

libethereum/Block.cpp
```diff
@@ -864,7 +864,14 @@ bool Block::sealBlock(bytesConstRef _header)
 h256 Block::stateRootBeforeTx(unsigned _i) const
 {
 	_i = min<unsigned>(_i, m_transactions.size());
-	return (_i > 0 ? receipt(_i - 1).stateRoot() : m_previousBlock.stateRoot());
+	try
+	{
+		return (_i > 0 ? receipt(_i - 1).stateRoot() : m_previousBlock.stateRoot());
+	}
+	catch(TransactionReceiptVersionError const&)
```
gumb0: space after `catch`

libethereum/TransactionReceipt.cpp
```diff
@@ -22,54 +22,88 @@
...
 
-	m_gasUsed = (u256)r[gasUsedIndex];
-	m_bloom = (LogBloom)r[gasUsedIndex + 1];
-	for (auto const& i : r[gasUsedIndex + 2])
+	if (!r[0].isData())
+		BOOST_THROW_EXCEPTION(InvalidTransactionReceiptFormat());
+	if (r[0].size() == 32)
+		m_statusCodeOrStateRoot = (h256)r[0];
+	else if (r[0].size() == 1)
+		m_statusCodeOrStateRoot = (uint8_t)r[0];
```
gumb0: Probable add `else BOOST_THROW_EXCEPTION(InvalidTransactionReceiptFormat());`

libweb3jsonrpc/JsonHelper.cpp
```diff
@@ -185,7 +185,14 @@ Json::Value toJson(dev::eth::TransactionSkeleton const& _t)
 Json::Value toJson(dev::eth::TransactionReceipt const& _t)
 {
 	Json::Value res;
-	res["stateRoot"] = toJS(_t.stateRoot());
+	try
```
gumb0: Maybe somewhat simpler would be to use `hasStatusCode()` instead of try/catch

libethereum/TransactionReceipt.h
```diff
@@ -26,23 +26,32 @@
...
-	TransactionReceipt(u256 _gasUsed, LogEntries const& _log) : TransactionReceipt(h256(), _gasUsed, _log) {}
-
-	h256 const& stateRoot() const { return m_stateRoot; }
+	TransactionReceipt(h256 const& _root, u256 const& _gasUsed, LogEntries const& _log);
+	TransactionReceipt(uint8_t _status, u256 const& _gasUsed, LogEntries const& _log);
+
+	/// @returns true if the receipt has a status code.  Otherwise the receipt has a state root (pre-EIP658).
+	bool hasStatusCode() const;
+	/// @returns the state root.
+	/// @throw TransactionReceiptVersionError when m_hasStatusCode is true.
```
gumb0: Fix comment, there's no `m_hasStatusCode ` anymore

PR #4276, opened 2017-07-19
Trying to cache .ccache directory on macOS Travis run

pirapira:
For this PR, I want to run Travis twice on Mac.  The first run should populate `.ccache` and store it; the second run should use the populated `.ccache`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,6 +103,7 @@ cache:
         - ~/.ethash
         - ~/.local
         - ~/Library/Caches/Homebrew
+        - ~/.ccache
         # Cache whole deps dir hoping you will not need to download and
         # build external dependencies next build.
         - $TRAVIS_BUILD_DIR/deps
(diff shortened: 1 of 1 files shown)


pirapira:
Now I'm seeing some non-empty ccache cache.  This should be carried over to the next run.
```
Cleaned cache
cache directory                     /Users/travis/.ccache
primary config                      /Users/travis/.ccache/ccache.conf
secondary config      (readonly)    /usr/local/Cellar/ccache/3.3.4_1/etc/ccache.conf
cache hit (direct)                     0
cache hit (preprocessed)               0
cache miss                           244
cache hit rate                      0.00 %
called for link                       36
cleanups performed                    23
files in cache                       354
cache size                         687.7 MB
max cache size                       1.0 GB
```

pirapira:
I think I was mitaken about the Travis configuration.

```
- cache
  - ccache: true
```
should apply for all builds.

PR #4277, opened 2017-07-19
[WIP] Cache /usr/local/Cellar on Mac, trying to avoid compilation of llvm

pirapira:
For this PR I want to see Travis run on macOS twice.


pirapira:
First mac build took 63 minutes.  Measuring the second run.

pirapira:
The time does not change, and `brew` is not building llvm anymore.

PR #4278, opened 2017-07-24
New modexp cost

pirapira:
This follows the new cost formula https://github.com/ethereum/EIPs/commit/4d4d8fb for modexp precompiled contract.


gumb0:
No unit tests change was needed - it means they cover only the case for mod & base length <=64. Could we add tests for other two cases?

libethcore/Precompiled.cpp
```diff
@@ -160,6 +160,16 @@ namespace
 			return 8 * (_expLength - 32) + highestBit;
 		}
 	}
+
+	bigint mult_complexity(bigint const& _x)
```
gumb0: I'd name it `multComplexity`

pirapira:
Right, the unit tests are only for very small, or very large `mod | base`.

pirapira:
I added unit tests for the other two cases.

pirapira:
AppVeyor build is failing and somehow I cannot restart AppVeyor.

pirapira:
The tests around `modexp` is passing in AppVeyor.

PR #4282, opened 2017-07-26
Inserting a block to BlockChain without enacting transactions

gumb0:
Essentially this is refactoring to split "behemoth of a method" `BlockChain::import` into two methods: `BlockChain::import` and private `BlockChain::insertBlockAndExtras`, then adding new public method `BlockChain::insertWithoutParent` that reuses `BlockChain::insertBlockAndExtras`.

`BlockChain::insertWithoutParent` will be used to import blocks from snapshot in future PR.

Also there is some removing and moving the code around `BlockChain::import`.

### DIFF
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -45,7 +45,6 @@ using namespace dev;
 using namespace dev::eth;
 namespace fs = boost::filesystem;
 
-#define ETH_CATCH 1
 #define ETH_TIMED_IMPORTS 1
 
 #if defined(_WIN32)
@@ -682,9 +681,7 @@ ImportRoute BlockChain::import(VerifiedBlockRef const& _block, OverlayDB const& 
 
 	BlockReceipts br;
 	u256 td;
-#if ETH_CATCH
 	try
-#endif
 	{
 		// Check transactions are valid and that they result in a state equivalent to our state_root.
 		// Get total difficulty increase and update state, checking it.
@@ -704,7 +701,6 @@ ImportRoute BlockChain::import(VerifiedBlockRef const& _block, OverlayDB const& 
 		checkConsistency();
 #endif // ETH_PARANOIA
 	}
-#if ETH_CATCH
 	catch (BadRoot& ex)
 	{
 		cwarn << "*** BadRoot error! Trying to import" << _block.info.hash() << "needed root" << ex.root;
@@ -717,7 +713,6 @@ ImportRoute BlockChain::import(VerifiedBlockRef const& _block, OverlayDB const& 
 		addBlockInfo(ex, _block.info, _block.block.toBytes());
 		throw;
 	}
-#endif // ETH_CATCH
 
 	// All ok - insert into DB
 	bytes const receipts = br.rlp();
@@ -1089,6 +1084,7 @@ tuple<h256s, h256, unsigned> BlockChain::treeRoute(h256 const& _from, h256 const
 
 	BlockDetails const fromDetails = details(_from);
 	BlockDetails const toDetails = details(_to);
+	// Needed to handle a special case when the parent of inserted block is not present in DB.
 	if (fromDetails.isNull() || toDetails.isNull())
 		return make_tuple(h256s(), h256(), 0);
 
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -158,8 +158,9 @@ private:
 	mutable h256s m_lastHashes;
 };
 
+}
 
-class ImportPerformanceLogger
+class dev::eth::ImportPerformanceLogger
 {
 public:
 	void onStageFinished(std::string const& _name)
@@ -215,8 +216,6 @@ private:
 	std::unordered_map<std::string, double> m_stages;
 	Timer m_stageTimer;
 };
-
-}
 
 #if ETH_DEBUG&&0
 static const chrono::system_clock::duration c_collectionDuration = chrono::seconds(15);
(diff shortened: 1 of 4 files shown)


libethereum/BlockChain.cpp
```diff
@@ -582,59 +532,22 @@ pair<ImportResult, ImportRoute> BlockChain::attemptImport(bytes const& _block, O
...
-	try
-#endif
-	{
-		block = verifyBlock(&_block, m_onBad, ImportRequirements::OutOfOrderChecks);
-	}
-#if ETH_CATCH
-	catch (Exception& ex)
-	{
-//		clog(BlockChainNote) << "   Malformed block: " << diagnostic_information(ex);
-		ex << errinfo_phase(2);
```
gumb0: I haven't found any place where this additional info is used or at least printed out, so I think it's safe to remove this

libethereum/BlockChain.cpp
```diff
@@ -1159,41 +1084,42 @@ void BlockChain::rewind(unsigned _newHead)
 
 tuple<h256s, h256, unsigned> BlockChain::treeRoute(h256 const& _from, h256 const& _to, bool _common, bool _pre, bool _post) const
 {
-//	cdebug << "treeRoute" << _from << "..." << _to;
 	if (!_from || !_to)
 		return make_tuple(h256s(), h256(), 0);
+
+	BlockDetails const fromDetails = details(_from);
+	BlockDetails const toDetails = details(_to);
+	if (fromDetails.isNull() || toDetails.isNull())
```
gumb0: This is needed to handle a special case when the parent of inserted block is not present in DB

libethereum/BlockChain.cpp
```diff
@@ -805,26 +741,85 @@ ImportRoute BlockChain::import(VerifiedBlockRef const& _block, OverlayDB const&
...
 
-#if ETH_TIMED_IMPORTS
 		performanceLogger.onStageFinished("enactment");
-#endif // ETH_TIMED_IMPORTS
 
 #if ETH_PARANOIA
 		checkConsistency();
 #endif // ETH_PARANOIA
+	}
+#if ETH_CATCH
```
chfast: Maybe it is safe to assume ETH_CATCH all the time. Having different implementations depending on ETH_CATCH macro seems to be bad idea (maybe it was useful in time of experimenting with it, but rather not useful now).

libethereum/BlockChain.cpp
chfast: This comment ^ should be placed in the code.

libethereum/BlockChain.h
```diff
@@ -137,6 +138,8 @@ class BlockChain
 	/// block/header and receipts directly into the databases.
 	void insert(bytes const& _block, bytesConstRef _receipts, bool _mustBeNew = true);
 	void insert(VerifiedBlockRef _block, bytesConstRef _receipts, bool _mustBeNew = true);
+	/// insert that doesn't require parent to be imported, useful when we don't have the full blockchain (like restoring from partial snapshot)
```
chfast: Pedantic mode: Write comments as full sentences, starting with upper letter, ending with full stop.

libethereum/ImportPerformanceLogger.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/** @file
```
chfast: The `@file` comment block is only useful if you want to place any documenting comment here.

libethereum/ImportPerformanceLogger.h
```diff
@@ -0,0 +1,74 @@
...
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/** @file 
+ *  Class for logging of importing a Block into BlockChain
```
chfast: Full stop missing.

libethereum/ImportPerformanceLogger.h
```diff
@@ -0,0 +1,74 @@
...
+		double const totalElapsed = m_totalTimer.elapsed();
+		if (totalElapsed > 0.5)
+		{
+			cnote << "SLOW IMPORT: { " << constructReport(totalElapsed, _additionalValues) << " }";
+		}
+	}
+
+private:
+	template <class ValueType>
+	static std::string pairToString(std::pair<std::string const, ValueType> const& _pair)
```
chfast: Because this template is only used in single .cpp file - move it there.

libethereum/ImportPerformanceLogger.h
```diff
@@ -0,0 +1,74 @@
...
+	{
+		return "\"" + _pair.first + "\": " + toString(_pair.second);
+	}
+
+	std::string constructReport(double _totalElapsed, std::unordered_map<std::string, std::string> const& _additionalValues);
+
+	Timer m_totalTimer;
+	std::unordered_map<std::string, double> m_stages;
+	Timer m_stageTimer;
```
chfast: I'd place Timer fields next to each other.

libethereum/BlockChain.cpp
gumb0: Agree

gumb0:
@chfast Thanks for the review!

PR #4290, opened 2017-07-31
EVMJIT: RETURNDATA

chfast:
EVMJIT part: https://github.com/ethereum/evmjit/pull/123.

### DIFF
--- a/libevm/JitVM.cpp
+++ b/libevm/JitVM.cpp
@@ -156,10 +156,9 @@ void create(evm_result* o_result, ExtVMFace& _env, evm_message const* _msg) noex
 	// ExtVM::create takes the sender address from .myAddress.
 	assert(fromEvmC(_msg->sender) == _env.myAddress);
 
-	// TODO: EVMJIT does not support RETURNDATA at the moment, so
-	//       the output is ignored here.
 	h160 addr;
-	std::tie(addr, std::ignore) = _env.create(value, gas, init, Instruction::CREATE, u256(0), {});
+	owning_bytes_ref output;
+	std::tie(addr, output) = _env.create(value, gas, init, Instruction::CREATE, u256(0), {});
 	o_result->gas_left = static_cast<int64_t>(gas);
 	o_result->release = nullptr;
 	if (addr)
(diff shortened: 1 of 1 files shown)


libevm/JitVM.cpp
```diff
@@ -175,9 +174,27 @@ void create(evm_result* o_result, ExtVMFace& _env, evm_message const* _msg) noex
 	}
 	else
 	{
-		o_result->code = EVM_FAILURE;
-		o_result->output_data = nullptr;
-		o_result->output_size = 0;
+		o_result->code = EVM_REVERT;
```
gumb0: It could fail for some other reason, like out of gas or address collision. Will this code affect anything?

chfast: This is a bit messy EVM-C concept what was around before REVERT. The idea is to pass some error messages describing what failed in the EVM. Here we try to reuse the same output buffer to pass consensus-critical REVERT output and debug messages. EVM_REVERT informs the EVM implementation that this output comes from REVERT and should not be discarded.

This long description indicates that this is probably bad idea. What do you think @axic? Maybe long list of error codes should be enough?

chfast: One example of future use: `EVM_STACK_UNSERFLOW`, output: `"Stack overflow in instruction ADD at position 321, instruction counter: 12344"`.

Receiving this EVM can log it, but should not copy it to the RETURNDATA buffer.

chfast: @gumb0 @axic Is my explanation enough at this moment or should I address the issue first?

gumb0: I meant that here `ExtVMFace::create` could not only succeed or revert, but also fail for other reason. But you always return `EVM_REVERT` in case of failure, so EVM will try to use return buffer in case of OOG for example. Will that be a problem?

pirapira:
The branch needs to be updated.

libevm/JitVM.cpp
chfast: No, because the return buffer will be always empty in other cases.

I'm working on addressing these issues, but expect another PR.

gumb0: Ok then

chfast: Still we can revert the merge order :) Let's see how it will look like when I finish.

PR #4291, opened 2017-07-31
Snapshot import

gumb0:
Importing Parity's warp snapshot

The way to try it:
1. Download snapshot using this branch https://github.com/ethereum/cpp-ethereum/pull/4227 or using Parity (it puts it some dir like Parity/Ethereum/chains/ethereum/db/906a34e69aec8c0d/snapshot)
2. `./eth --import-snapshot snapshot_dir -v 9`
(also setting database path with `-d` might be useful)
3. After snapshot import is complete regular sync start automatically from the snapshot block


chfast:
In general snappy external project looks ok, I will try to fix it tomorrow on linux.

However, we have snappy already because leveldb uses it. So on Linux/macOS we might try to using system level snappy (because we are using system level leveldb). I'm not sure if the Windows prebuilt dep has the snappy in it (check the script building it) and if it easy accessible (public API).

gumb0:
@chfast all builds are failing now

gumb0:
@chfast Where's the script building windows leveldb?

chfast:
https://github.com/ethereum/cpp-dependencies
3 .bat scripts with _leveldb suffix.

gumb0:
I think this can be reviewed now. I will work on adding some unit tests.

libethereum/BlockChainImporter.h
```diff
@@ -0,0 +1,51 @@
...
+namespace eth
+{
+
+class BlockChain;
+class BlockHeader;
+
+class BlockChainImporterFace
+{
+public:
+	virtual ~BlockChainImporterFace() {}
```
chfast: I usually do `virtual ~C() = default;`.

libethereum/BlockChainImporter.cpp
```diff
@@ -0,0 +1,61 @@
...
+		m_blockChain.setChainStartBlockNumber(static_cast<unsigned>(_number));
+	}
+
+private:
+	BlockChain& m_blockChain;
+};
+
+}
+
+std::unique_ptr<BlockChainImporterFace> dev::eth::createBlockChainImporter(BlockChain& _blockChain)
```
chfast: Why do we want to hide `BlockChainImporter` behind `BlockChainImporterFace`?

libethereum/SnapshotStorage.cpp
```diff
@@ -0,0 +1,83 @@
...
+{
+
+std::string snappyUncompress(std::string const& _compressed)
+{
+	size_t uncompressedSize = 0;
+	if (!snappy::GetUncompressedLength(_compressed.data(), _compressed.size(), &uncompressedSize))
+		BOOST_THROW_EXCEPTION(FailedToGetUncompressedLength());
+
+	std::vector<char> uncompressed(uncompressedSize);
+	if (!snappy::RawUncompress(_compressed.data(), _compressed.size(), uncompressed.data()))
```
chfast: Because we are using Hunter for Snappy package, you don't need to use RawUncompress any more (Hunter guarantees the same ABI because it uses exactly the same compiler to compile dependencies).

If you want to stick with it, you don't have to use vector. You can do the same with string directly. But still you will init the string with zeros, what is not needed.

libethereum/SnapshotStorage.cpp
```diff
@@ -0,0 +1,83 @@
...
+			BOOST_THROW_EXCEPTION(FailedToReadSnapshotManifestFile());
+
+		return manifestBytes;
+	}
+
+	std::string readChunk(h256 const& _chunkHash) const override
+	{
+		std::string const chunkCompressed = dev::contentsString((m_snapshotDir / toHex(_chunkHash)).string());
+		if (chunkCompressed.empty())
+			BOOST_THROW_EXCEPTION(FailedToReadChunkFile());
```
chfast: I would add information about the chunk id to the exception.

libethereum/SnapshotImporter.cpp
```diff
@@ -0,0 +1,257 @@
...
+using namespace dev;
+using namespace eth;
+
+namespace
+{
+
+struct SnapshotImportLog: public LogChannel
+{
+	static char const* name() { return "SNAP"; }
+	static int const verbosity = 9;
```
chfast: Were you able to use it for debugging? I find our logging system without filters quite useless.

libethereum/BlockChainImporter.cpp
gumb0: Primary reason is to make `SnapshotImporter` better testable by decoupling it better from work with `BlockChain`. In unit tests I create a mock implementation of `BlockChainImporterFace`, see https://github.com/ethereum/cpp-ethereum/pull/4291/files#diff-aa42adc0a686c5dcee440883d97a4080R75 and this way avoid any intercation with the real database.

gumb0: Bonus: implementation details of `BlockChainImporter` are better encapsulated and the we don't need to put `#include "BlockChain.h"` into header file for example

chfast: Please don't do "namespacing" this way. Include the implementation also within `namespace dev { namespace eth {}}`. This caused me a lot of name collisions before.

gumb0: Ok. We should probably put it into CodingStandards

libethereum/BlockChainImporter.h
gumb0: Ok, not much difference, but looks more "modern" :)

libethereum/SnapshotImporter.cpp
gumb0: Yeah. On Windows it somehow looks more clear, at least in release build and I can see snapshot importing progress. On Linux there's more spam from somewhere (probably `BlockChain`)

libethereum/BlockChainImporter.cpp
gumb0: Is it ok to leave `using namespace dev;` & `using namespace eth;` at the top of the file to be used in anon namespace after that?

chfast: I would put the anon namespace inside `dev::eth`, but it's up to you. Truly, I have never seen the convention like here not to put implementations inside the namespace they belong.

libethereum/SnapshotStorage.cpp
gumb0: I'll use `Uncompress` with `string`, it'simpler.
(vector was needed here because string can give only const pointer to it's data)

libethereum/BlockChainImporter.cpp
gumb0: Yeah, anon inside `dev::eth` is better because less repeating of namespace names

PR #4293, opened 2017-08-01
Use MainNetworkTest for ethvm default.

gcolvin:
Gives much faster startup.  Main is still available as command line argument.

### DIFF
--- a/ethvm/main.cpp
+++ b/ethvm/main.cpp
@@ -141,7 +141,7 @@ int main(int argc, char** argv)
 	u256 gasPrice = 0;
 	bool styledJson = true;
 	StandardTrace st;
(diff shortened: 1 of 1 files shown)


gcolvin:
Another big part of the performance regression is that there is no longer a default CMAKE_BUILD_TYPE.  It used to be RelWithDebInfo.  It makes a big difference.  Here are a few times with no default:

pop
exec time: 5.750131
user time: 13.17

div256
exec time: 555.986824
user time: 563.21

mix
exec time: 227.748832
user time: 235.27

And the same with RelWithDebInfo:

pop
exec time: 0.814945
user time: 1.39

div256
exec time: 88.575871
user time: 89.12

mix
exec time: 25.897099
user time: 26.50

PR #4294, opened 2017-08-01
add hive test check to the Blockchain tests

winsvega:
There was an issue when hive got stack with some tests failing even on cpp client. 
That is because we didn't keep hive tests updated to the latest develop. 

This pr adds a test check to the travis for all hive tests.


winsvega:
changed to run hive test check on azure server

PR #4302, opened 2017-08-03
force build type

gcolvin:
default CMAKE_BUILD_TYPE was not getting set for lack FORCE

### DIFF
--- a/cmake/EthOptions.cmake
+++ b/cmake/EthOptions.cmake
@@ -1,6 +1,6 @@
 macro(configure_project)
 	# Default to RelWithDebInfo configuration if no configuration is explicitly specified.
-	set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Build type on single-configuration generators")
+	set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Build type on single-configuration generators" FORCE)
 
 	option(BUILD_SHARED_LIBS "Build project libraries shared" OFF)
 
--- a/cmake/EthOptions.cmake
+++ b/cmake/EthOptions.cmake
@@ -1,7 +1,9 @@
 macro(configure_project)
 	# Default to RelWithDebInfo configuration if no configuration is explicitly specified.
-	set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Build type on single-configuration generators" FORCE)
-
+	if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+	   set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Build type on single-configuration generators" FORCE)
+	endif()
+	
 	option(BUILD_SHARED_LIBS "Build project libraries shared" OFF)
 
 	# Features:


cmake/EthOptions.cmake
```diff
@@ -1,6 +1,6 @@
 macro(configure_project)
 	# Default to RelWithDebInfo configuration if no configuration is explicitly specified.
-	set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Build type on single-configuration generators")
+	set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Build type on single-configuration generators" FORCE)
```
chfast: This way it will be always `RelWithDebInfo`. We need to check for empty value first.

```
if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Build type on single-configuration generators" FORCE)
endif()

The `AND NOT CMAKE_CONFIGURATION_TYPES` added to ignore Visual Studio where CMAKE_BUILD_TYPE make no sense.

Comment (gcolvin) [inline cmake/EthOptions.cmake]: ok

File: cmake/EthOptions.cmake [hunk 2]
```diff
@@ -1,7 +1,9 @@
 macro(configure_project)
 	# Default to RelWithDebInfo configuration if no configuration is explicitly specified.
-	set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Build type on single-configuration generators")
-
+	if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
```
gcolvin: I think this should read
`	if(NOT CMAKE_BUILD_TYPE AND CMAKE_CONFIGURATION_TYPES)`
I noticed the tests were still slow and found the variable still set - because CMAKE_CONFIGURATION_TYPES was set to "Debug;Release;RelWithDebInfo", which seemed no reason to not set the build type. Rather, having no available configuration types would be a reason to not set the build type.  @chfast ?

chfast: You are somehow right. We should not set CMAKE_CONFIGURATION_TYPES at all, because it affects only Visual Studio.
You can find it in the main CMakeLists.txt.

chfast: Sometimes I really hate CMake.

gcolvin: Yeah, CMake sucks.  So does make, but at least it's tried and true.
@chfast How to fix this then?

chfast: Remove `set(CMAKE_CONFIGURATION_TYPES...)` from CMakeLists.txt.

(You may also need to clear cmake cache and configure again -- delete CMakeCache.txt in the build dir).

gcolvin: OK.  Could you stand to do a PR for this?  I have 200 miles of traveling to do today.

chfast: https://github.com/ethereum/cpp-ethereum/pull/4339

gcolvin: @chfast thanks!

PR #4303, opened 2017-08-03
disable Frontier and Homestead tests if --all not set

winsvega:
disable tests on Frontier and Homestead networks if flag --all not set.
Tests will be run on Azure server

### DIFF
--- a/test/tools/libtesteth/TestHelper.cpp
+++ b/test/tools/libtesteth/TestHelper.cpp
@@ -137,6 +137,24 @@ eth::Network stringToNetId(string const& _netname)
 	return eth::Network::FrontierTest;
 }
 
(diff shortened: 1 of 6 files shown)


test/tools/libtesteth/TestHelper.cpp
```diff
@@ -137,6 +137,24 @@ eth::Network stringToNetId(string const& _netname)
 	return eth::Network::FrontierTest;
 }
 
+bool isDisabledNetwork(eth::Network _net)
+{
```
pirapira: I think this should return `false` when `testeth` is filling tests.

winsvega: done

chfast:
This is very bad for local testing because developer will neither run the disabled tests.

PR #4305, opened 2017-08-04
Clear storage before contract initialization

pirapira:
Before this PR, a new contract's storage was cleared after the initialization code runs, but the storage should be cleared before the initialization code runs.


chfast:
Is this a bug fix? This "clear storage" after execution looks super strange.

pirapira:
Yes, this is a bug fix.  I think this should have caused more test failures https://github.com/ethereum/tests/issues/242

PR #4314, opened 2017-08-08
generating and filling a random state test

winsvega:
Usage: 

./testeth -t StateTestsGeneral -- --createRandomTest

### DIFF
--- a/test/tools/fuzzTesting/createRandomTest.cpp
+++ b/test/tools/fuzzTesting/createRandomTest.cpp
@@ -169,6 +169,8 @@ void fillRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, st
 		dev::test::RandomCode::parseTestWithTypes(newTest, nullReplaceMap);
 		json_spirit::read_string(newTest, v);
 		_doTests(v, true);
+		if (!v.get_obj().count("post"))
+			BOOST_ERROR("And error occured when executing random state test!");
 	}
 	catch(...)
 	{
--- a/test/tools/fuzzTesting/createRandomTest.cpp
+++ b/test/tools/fuzzTesting/createRandomTest.cpp
@@ -42,73 +42,29 @@ int checkRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, js
 namespace dev { namespace test {
 int createRandomTest(std::vector<char*> const& _parameters)
 {
-	std::string testSuite;
-	std::string testFillString;
-	json_spirit::mValue testmValue;
-	bool checktest = false;
-	bool filldebug = false;
 	bool debug = false;
-	bool filltest = false;
-
 	TestOutputHelper::initTest(1);
 	dev::test::Options& options = const_cast<dev::test::Options&>(dev::test::Options::get());
 
-	testSuite = options.rCurrentTestSuite;
+	std::string const& testSuite = options.rCurrentTestSuite;
 	if (testSuite != "BlockchainTests" && testSuite != "TransactionTests" && testSuite != "StateTestsGeneral"
-				&& testSuite != "VMTests")
+		&& testSuite != "VMTests")
 	{
 		std::cerr << "Error! Test suite '" + testSuite +"' not supported! (Usage -t TestSuite)" << std::endl;
 		return 1;
 	}
 
+	//parse options for createRandomTest special test suite
 	for (size_t i = 0; i < _parameters.size(); ++i)
 	{
 		auto arg = std::string{_parameters.at(i)};
 
 		if (arg == "--debug")
 			debug = true;
-		else
-		if (arg == "--filldebug")
-			filldebug = true;
 	}
-
-	if (checktest)
-		std::cout << "Testing: " << testSuite.substr(0, testSuite.length() - 1) << "... ";
 
 	if (testSuite == "StateTestsGeneral")
-	{
-		if (checktest)
-			return checkRandomTest(dev::test::doStateTests, testmValue, debug);
-		else
-			fillRandomTest(dev::test::doStateTests, (filltest) ? testFillString : c_testExampleStateTest, filldebug);
-	}
-	else
-	if (testSuite == "BlockchainTests")
-	{
-		if (checktest)
-			return checkRandomTest(dev::test::doBlockchainTests, testmValue, debug);
-		else
-			fillRandomTest(dev::test::doBlockchainTests, (filltest) ? testFillString : c_testExampleBlockchainTest, filldebug);
-	}
-	else
-	if (testSuite == "TransactionTests")
-	{
-		if (checktest)
-			return checkRandomTest(dev::test::doTransactionTests, testmValue, debug);
-		else
-			fillRandomTest(dev::test::doTransactionTests, (filltest) ? testFillString : c_testExampleTransactionTest, filldebug);
-	}
-	else
-	if (testSuite == "VMTests")
-	{
-		if (checktest)
-		{
-			dev::eth::VMFactory::setKind(dev::eth::VMKind::JIT);
-			return checkRandomTest(dev::test::doVMTests, testmValue, debug);
-		}
-		else
-			fillRandomTest(dev::test::doVMTests, (filltest) ? testFillString : c_testExampleVMTest, filldebug);
-	}
+		fillRandomTest(dev::test::doStateTests, c_testExampleStateTest, debug);
 
 	return 0;
 }
@@ -169,12 +125,13 @@ void fillRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, st
 		dev::test::RandomCode::parseTestWithTypes(newTest, nullReplaceMap);
 		json_spirit::read_string(newTest, v);
 		_doTests(v, true);
-		if (!v.get_obj().count("post"))
-			BOOST_ERROR("And error occured when executing random state test!");
+		for (auto& obj: v.get_obj())
+			if (!obj.second.get_obj().count("post"))
+				BOOST_ERROR("An error occured when executing random state test!");
 	}
 	catch(...)
 	{
-		std::cerr << "Test fill exception!";
+		BOOST_ERROR("Error generating random test!");
 	}
 
 	//restroe output
@@ -195,82 +152,78 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
 	for (std::map<std::string, std::string>::const_iterator it = _varMap.begin(); it != _varMap.end(); it++)
 		types.push_back(it->first);
 
-	for (unsigned i = 0; i < types.size(); i++)
+	for (auto const& type: types)
 	{
-		std::size_t pos = _test.find(types.at(i));
+		std::size_t pos = _test.find(type);
 		while (pos != std::string::npos)
 		{
-			if (types.at(i) == "[RLP]")
+			if (type == "[RLP]")
 			{
 				std::string debug;
 				int randomDepth = 1 + (int)dev::test::RandomCode::randomUniInt() % 10;
 				_test.replace(pos, 5, dev::test::RandomCode::rndRLPSequence(randomDepth, debug));
 				cnote << debug;
-				std::string a;
 			}
-			else
-			if (types.at(i) == "[CODE]")
+			else if (type == "[CODE]")
 			{
 				int random = (int)dev::test::RandomCode::randomUniInt() % 100;
 				if (random < 90)
-					_test.replace(pos, types.at(i).length(), "0x"+dev::test::RandomCode::generate(10, options));
+					_test.replace(pos, type.length(), "0x"+dev::test::RandomCode::generate(5, options));
 				else
-					_test.replace(pos, types.at(i).length(), "");
+					_test.replace(pos, type.length(), "");
 			}
-			else
-			if (types.at(i) == "[HEX]")
-				_test.replace(pos, types.at(i).length(), dev::test::RandomCode::randomUniIntHex());
-			else
-			if (types.at(i) == "[HEX32]")
-				_test.replace(pos, types.at(i).length(), dev::test::RandomCode::randomUniIntHex(std::numeric_limits<uint32_t>::max()));
-			else
-			if (types.at(i) == "[GASLIMIT]")
-				_test.replace(pos, types.at(i).length(), dev::test::RandomCode::randomUniIntHex(dev::u256("3000000000")));
-			else
-			if (types.at(i) == "[HASH20]")
-				_test.replace(pos, types.at(i).length(), dev::test::RandomCode::rndByteSequence(20));
-			else
-			if (types.at(i) == "[ADDRESS]")
-				_test.replace(pos, types.at(i).length(), toString(options.getRandomAddress()));
-			else
-			if (types.at(i) == "[0xADDRESS]")
-				_test.replace(pos, types.at(i).length(), "0x" + toString(options.getRandomAddress()));
-			else
-			if (types.at(i) == "[0xDESTADDRESS]")
+			else if (type == "[HEX]")
+				_test.replace(pos, type.length(), dev::test::RandomCode::randomUniIntHex());
+			else if (type == "[HEX32]")
+			{
+				int random = (int)dev::test::RandomCode::randomUniInt() % 100;
+				if (random < 90)
+					_test.replace(pos, type.length(), dev::test::RandomCode::randomUniIntHex(std::numeric_limits<uint32_t>::max()));
+				else
+					_test.replace(pos, type.length(), "0x00");
+			}
+			else if (type == "[GASLIMIT]")
+				_test.replace(pos, type.length(), dev::test::RandomCode::randomUniIntHex(dev::u256("36028797018963967"), dev::u256("18014398509481983")));
+			else if (type == "[GASPRICE]")
+				_test.replace(pos, type.length(), dev::test::RandomCode::randomUniIntHex(dev::u256("5")));
+			else if (type == "[HASH20]")
+				_test.replace(pos, type.length(), dev::test::RandomCode::rndByteSequence(20));
+			else if (type == "[ADDRESS]")
+				_test.replace(pos, type.length(), toString(options.getRandomAddress()));
+			else if (type == "[0xADDRESS]")
+				_test.replace(pos, type.length(), "0x" + toString(options.getRandomAddress()));
+			else if (type == "[0xDESTADDRESS]")
 			{
 				int random = (int)dev::test::RandomCode::randomUniInt() % 100;
 				if (random < 50)
-					_test.replace(pos, types.at(i).length(), "0x" + toString(options.getRandomAddress()));
+					_test.replace(pos, type.length(), "0x" + toString(options.getRandomAddress()));
 				else
-					_test.replace(pos, types.at(i).length(), "");
+					_test.replace(pos, type.length(), "");
 			}
-			else
-			if (types.at(i) == "[0xHASH32]")
-				_test.replace(pos, types.at(i).length(), "0x" + dev::test::RandomCode::rndByteSequence(32));
-			else
-			if (types.at(i) == "[HASH32]")
-				_test.replace(pos, types.at(i).length(), dev::test::RandomCode::rndByteSequence(32));
-			else
-			if (types.at(i) == "[V]")
+			else if (type == "[0xHASH32]")
+				_test.replace(pos, type.length(), "0x" + dev::test::RandomCode::rndByteSequence(32));
+			else if (type == "[HASH32]")
+				_test.replace(pos, type.length(), dev::test::RandomCode::rndByteSequence(32));
+			else if (type == "[V]")
 			{
 				int random = (int)dev::test::RandomCode::randomUniInt() % 100;
 				if (random < 30)
-					_test.replace(pos, types.at(i).length(), "0x1c");
+					_test.replace(pos, type.length(), "0x1c");
 				else
 				if (random < 60)
-					_test.replace(pos, types.at(i).length(), "0x1d");
+					_test.replace(pos, type.length(), "0x1d");
 				else
-					_test.replace(pos, types.at(i).length(), "0x" + dev::test::RandomCode::rndByteSequence(1));
+					_test.replace(pos, type.length(), "0x" + dev::test::RandomCode::rndByteSequence(1));
 			}
 			else
 			{
 				//Replace type from varMap if varMap is set
-				if (_varMap.count(types.at(i)))
-					_test.replace(pos, types.at(i).length(), _varMap.at(types.at(i)));
-				BOOST_ERROR("Skipping undeclared type: " + types.at(i));
+				if (_varMap.count(type))
+					_test.replace(pos, type.length(), _varMap.at(type));
+				BOOST_ERROR("Skipping undeclared type: " + type);
 			}
 
-			pos = _test.find(types.at(i));
+			pos = _test.find(type);
 		}
 	}
 }
@@ -278,7 +231,7 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
 std::vector<std::string> dev::test::RandomCode::getTypes()
 {
 	//declare possible types
-	return {"[RLP]", "[CODE]", "[HEX]", "[HEX32]", "[HASH20]", "[HASH32]", "[0xHASH32]", "[V]", "[GASLIMIT]", "[ADDRESS]", "[0xADDRESS]", "[0xDESTADDRESS]"};
+	return {"[RLP]", "[CODE]", "[HEX]", "[HEX32]", "[HASH20]", "[HASH32]", "[0xHASH32]", "[V]", "[GASLIMIT]", "[GASPRICE]", "[ADDRESS]", "[0xADDRESS]", "[0xDESTADDRESS]"};
 }
 
 std::string const c_testExampleTransactionTest = R"(
@@ -355,7 +308,7 @@ std::string const c_testExampleStateTest = R"(
 		"gasLimit" : [
 			"[HEX32]"
 		],
-		"gasPrice" : "[V]",
+		"gasPrice" : "[GASPRICE]",
 		"nonce" : "0",
 		"secretKey" : "0x45a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065ff2d8",
 		"to" : "[0xDESTADDRESS]",
--- a/test/tools/fuzzTesting/createRandomTest.cpp
+++ b/test/tools/fuzzTesting/createRandomTest.cpp
@@ -50,62 +50,64 @@ int createRandomTest(std::vector<char*> const& _parameters)
 	bool debug = false;
 	bool filltest = false;
 
+	TestOutputHelper::initTest(1);
 	dev::test::Options& options = const_cast<dev::test::Options&>(dev::test::Options::get());
 
 	testSuite = options.rCurrentTestSuite;
-	if (testSuite != "BlockChainTests" && testSuite != "TransactionTests" && testSuite != "StateTests"
+	if (testSuite != "BlockchainTests" && testSuite != "TransactionTests" && testSuite != "StateTestsGeneral"
 				&& testSuite != "VMTests")
-		testSuite = "";
+	{
+		std::cerr << "Error! Test suite '" + testSuite +"' not supported! (Usage -t TestSuite)" << std::endl;
+		return 1;
+	}
 
 	for (size_t i = 0; i < _parameters.size(); ++i)
 	{
 		auto arg = std::string{_parameters.at(i)};
 
-		if (arg == "--fulloutput")
-			options.fulloutput = true;
-		else if (arg == "--debug")
+		if (arg == "--debug")
 			debug = true;
 		else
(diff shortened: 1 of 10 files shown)


test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -50,62 +50,64 @@ int createRandomTest(std::vector<char*> const& _parameters)
...
+		if (arg == "--debug")
 			debug = true;
 		else
 		if (arg == "--filldebug")
 			filldebug = true;
 	}
 
-	if (testSuite == "")
+	if (checktest)
+		std::cout << "Testing: " << testSuite.substr(0, testSuite.length() - 1) << "... ";
```
pirapira: What does this `substr` do?

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -32,7 +32,7 @@ using namespace boost;
 
 Timer TestOutputHelper::m_timer;
 size_t TestOutputHelper::m_currTest = 0;
-size_t TestOutputHelper::m_maxTests = 0;
+size_t TestOutputHelper::m_maxTests = 1;
```
pirapira: What does this change do?

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -44,7 +44,8 @@ void TestOutputHelper::initTest(int _maxTests)
 	NoProof::init();
 	m_timer.restart();
 	m_currentTestCaseName = boost::unit_test::framework::current_test_case().p_name;
-	std::cout << "Test Case \"" + m_currentTestCaseName + "\": " << std::endl;
+	if (!Options::get().createRandomTest)
+		std::cout << "Test Case \"" + m_currentTestCaseName + "\": " << std::endl;
```
pirapira: Should it output a different name when `createRandomTest`?

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -50,62 +50,64 @@ int createRandomTest(std::vector<char*> const& _parameters)
...
 	{
-		std::cerr << "Error! Test suite not supported! (Usage -t TestSuite)" << std::endl;
-		return 1;
+		if (checktest)
+			return checkRandomTest(dev::test::doStateTests, testmValue, debug);
+		else
+			fillRandomTest(dev::test::doStateTests, (filltest) ? testFillString : c_testExampleStateTest, filldebug);
 	}
 	else
+	if (testSuite == "BlockchainTests")
```
pirapira: Maybe `else if` should be on the same line.

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -50,62 +50,64 @@ int createRandomTest(std::vector<char*> const& _parameters)
...
-	if (testSuite == "")
+	if (checktest)
+		std::cout << "Testing: " << testSuite.substr(0, testSuite.length() - 1) << "... ";
+
+	if (testSuite == "StateTestsGeneral")
 	{
-		std::cerr << "Error! Test suite not supported! (Usage -t TestSuite)" << std::endl;
-		return 1;
+		if (checktest)
+			return checkRandomTest(dev::test::doStateTests, testmValue, debug);
```
pirapira: Why is `fillRandomTest` is skipped when `chekcktest`?   Isn't `checktest` similar to `checkstate` option?  I felt the filling should happen even with the `checktest` option.

Also, similarly to `checkstate` option, maybe `checktest` should always be true.

test/tools/libtesteth/TestOutputHelper.cpp
winsvega: there was a division by 0 if TestOutputHelper not being initialized. with init function

test/tools/libtesteth/TestOutputHelper.cpp
winsvega: createRandomTest provide a clean json object to the std output

test/tools/fuzzTesting/createRandomTest.cpp
winsvega: this is a legacy code and actually is not working. 
there was an option to run the test provided by std input. 

this PR enables the support of 
```./testeth -t StateTestsGeneral -- --createRandomTest```
command 
other options like --checkstate are yet not supported. but I could review the code when needed.

Comment (pirapira) [inline test/tools/fuzzTesting/createRandomTest.cpp]: I see.  In that case, let's remove the legacy code before adding new features.

File: test/tools/libtesteth/TestOutputHelper.cpp (hunk 3 above)
Comment (pirapira) [inline test/tools/libtesteth/TestOutputHelper.cpp]: OK.  I didn't know this.

File: test/tools/fuzzTesting/createRandomTest.cpp (hunk 5 above)
Comment (winsvega) [inline test/tools/fuzzTesting/createRandomTest.cpp]: this new features are actually revived old ones.

File: test/tools/fuzzTesting/createRandomTest.cpp [hunk 6]
```diff
@@ -50,62 +50,64 @@ int createRandomTest(std::vector<char*> const& _parameters)
...
+		else
+			fillRandomTest(dev::test::doStateTests, (filltest) ? testFillString : c_testExampleStateTest, filldebug);
```
gumb0: `filltest` is always false, remove it alltogether, it will simplify things

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -218,42 +204,69 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
 				int randomDepth = 1 + (int)dev::test::RandomCode::randomUniInt() % 10;
 				_test.replace(pos, 5, dev::test::RandomCode::rndRLPSequence(randomDepth, debug));
 				cnote << debug;
+				std::string a;
```
gumb0: what's this?

test/tools/fuzzTesting/createRandomTest.cpp
pirapira: Is there a way to make `checktest` true?

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -262,7 +275,8 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
 
 std::vector<std::string> dev::test::RandomCode::getTypes()
 {
-	return {"[RLP]", "[CODE]", "[HEX]", "[HEX32]", "[HASH20]", "[HASH32]", "[0xHASH32]", "[V]", "[GASLIMIT]"};
+	//declare possible types
+	return {"[RLP]", "[CODE]", "[HEX]", "[HEX32]", "[HASH20]", "[HASH32]", "[0xHASH32]", "[V]", "[GASLIMIT]", "[ADDRESS]", "[0xADDRESS]", "[0xDESTADDRESS]"};
```
gumb0: Make it a constant on top of the file instead of `getTypes()` function

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -218,42 +204,69 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
...
 				cnote << debug;
+				std::string a;
 			}
 			else
 			if (types.at(i) == "[CODE]")
-				_test.replace(pos, 6, "0x"+dev::test::RandomCode::generate(10, options));
+			{
+				int random = (int)dev::test::RandomCode::randomUniInt() % 100;
+				if (random < 90)
+					_test.replace(pos, types.at(i).length(), "0x"+dev::test::RandomCode::generate(10, options));
```
gumb0: Don't repeat `types.at(i)` 1000 times, use instead the loop `for (auto const& type: types)`

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -42,71 +42,29 @@ int checkRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, js
...
-			{
-				dev::eth::VMFactory::setKind(dev::eth::VMKind::JIT);
-				return checkRandomTest(dev::test::doVMTests, testmValue, debug);
-			}
-			else
-				fillRandomTest(dev::test::doVMTests, (filltest) ? testFillString : c_testExampleVMTest, filldebug);
-		}
-	}
+	if (testSuite == "StateTestsGeneral")
+		fillRandomTest(dev::test::doStateTests, c_testExampleStateTest, debug);
```
pirapira: Why doesn't VMTest appear here?  Maybe we should say "Test suite VMTest not supported" above?

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -167,10 +125,13 @@ void fillRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, st
 		dev::test::RandomCode::parseTestWithTypes(newTest, nullReplaceMap);
 		json_spirit::read_string(newTest, v);
 		_doTests(v, true);
+		for (auto& obj: v.get_obj())
+			if (!obj.second.get_obj().count("post"))
+				BOOST_ERROR("An error occured when executing random state test!");
```
pirapira: Can we make `_doTests` return a boolean indicating an error?

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -185,84 +146,92 @@ void fillRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, st
...
-		std::size_t pos = _test.find(types.at(i));
+		std::size_t pos = _test.find(type);
 		while (pos != std::string::npos)
 		{
-			if (types.at(i) == "[RLP]")
+			if (type == "[RLP]")
 			{
 				std::string debug;
 				int randomDepth = 1 + (int)dev::test::RandomCode::randomUniInt() % 10;
 				_test.replace(pos, 5, dev::test::RandomCode::rndRLPSequence(randomDepth, debug));
```
pirapira: Replace 5 with type.length().

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -185,84 +146,92 @@ void fillRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, st
...
+					_test.replace(pos, type.length(), "0x" + dev::test::RandomCode::rndByteSequence(1));
 			}
 			else
-				_test.replace(pos, types.at(i).size(), _varMap.at(types.at(i)));
+			{
+				//Replace type from varMap if varMap is set
+				if (_varMap.count(type))
+					_test.replace(pos, type.length(), _varMap.at(type));
+				BOOST_ERROR("Skipping undeclared type: " + type);
+			}
```
pirapira: You can siplify the code by defining `string replacement`.  In each case, you just do `replacement = something;`

After all these cases, a single
```
_test.replace(pos, type.length(), replacement);
```
will do the job.  This way, the many repetitions of `_test.replace(pos, type.length(), ....);` can be removed.

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -303,15 +303,15 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
 	return code;
 }
 
-std::string RandomCode::randomUniIntHex(u256 _maxVal)
+std::string RandomCode::randomUniIntHex(u256 _maxVal, u256 _minVal)
```
pirapira: `_minVal` is never modified.  Consider `u256 const& _minVal`.

test/tools/libtesteth/Options.cpp
```diff
@@ -99,6 +100,11 @@ Options::Options(int argc, char** argv)
 			if (i + 1 >= argc)
 				BOOST_THROW_EXCEPTION(InvalidOption(arg + " option is missing an argument."));
 		};
+		auto throwIfNotRandomTest = [&seenCreateRandomTest, &arg]()
+		{
+			if (!seenCreateRandomTest)
+				BOOST_THROW_EXCEPTION(InvalidOption(arg + " option should follow after --createRandomTest."));
+		};
```
pirapira: I believe `--debug` option can appear before `--createRandomTest`.  So the check should not happen during the parsing.  After the parsing, you can detect it.

test/tools/fuzzTesting/createRandomTest.cpp
winsvega: that would touch too many other unrelated code. like execution of all of the other test suites.

pirapira: OK.  This will be kept track in https://github.com/ethereum/cpp-ethereum/issues/4324

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -303,7 +303,7 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
 	return code;
 }
 
-std::string RandomCode::randomUniIntHex(u256 _maxVal, u256 _minVal)
+std::string RandomCode::randomUniIntHex(u256 _maxVal, u256 const _minVal)
```
pirapira: With `const&`, some memory would be saved.

test/tools/libtesteth/Options.cpp
```diff
@@ -238,12 +232,9 @@ Options::Options(int argc, char** argv)
 				g_logVerbosity = indentLevelInt;
 		}
 		else if (arg == "--createRandomTest")
-		{
 			createRandomTest = true;
-			seenCreateRandomTest = true;
-		}
 		else if (arg == "--debug")
-			throwIfNotRandomTest();
+			randomDebug = true;
```
pirapira: By the way, can we replace this `--debug` with `--verbosity 30` or something?

test/tools/jsontests/StateTests.cpp
```diff
@@ -49,7 +49,7 @@ void doStateTests(json_spirit::mValue& _v, bool _fillin)
 		string testname = i.first;
 		json_spirit::mObject& o = i.second.get_obj();
 
-		if (_fillin)
+		if (_fillin && !Options::get().createRandomTest)
```
pirapira: How many combinations are possible here?
* [ ] _fillin && !createRandom
* [ ] _fillin && createRandom
* [ ] !_fillin && !createRandom
* [ ] !_fillin && createRandom

We should try to remove unnecessary combinations or throw errors on meaningless combinations.

winsvega: what? why? 
if it is a filler then it is being checked that testname in json object match the test filename. 
unless it is a filling of a randomstate test that does not have a file.

pirapira: I don't understand.  Which of the fours possibilities are supposed to happen here?

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -42,71 +42,28 @@ int checkRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, js
...
 {
-	std::string testSuite;
-	std::string testFillString;
-	json_spirit::mValue testmValue;
-	bool checktest = false;
-	bool filldebug = false;
 	bool debug = false;
-	bool filltest = false;
-
+	TestOutputHelper::initTest(1);
```
gumb0: Do you need this? You've made the default to be 1

test/tools/libtesteth/TestOutputHelper.cpp
gumb0: Where was that?  Maybe you shouldn't use the class `TestOutputHelper` that always works with test sequence for the case when `m_maxTests` doesn't make sense

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -560,7 +561,10 @@ void addClientInfo(json_spirit::mValue& _v, std::string const& _testSource)
 		}
 
 		//prepare the relative src path
-		string source = _testSource.substr(_testSource.rfind("/src/"), _testSource.length());
+		string source = _testSource;
+		string::size_type pos = _testSource.rfind("/src/");
```
gumb0: Does this work on Windows?

test/tools/libtesteth/Options.h
```diff
@@ -78,6 +77,12 @@ class Options
 	bool wallet = false;
 	/// @}
 
+	/// CreateRandomTest
+	/// @{
+	bool createRandomTest = false; ///< Generate random test
+	bool randomDebug = false; ///< Debug random test generation
```
gumb0: I don't see this option really used anywhere

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -185,84 +146,91 @@ void fillRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, st
...
-			if (types.at(i) == "[0xHASH32]")
-				_test.replace(pos, 10, "0x" + dev::test::RandomCode::rndByteSequence(32));
-			else
-			if (types.at(i) == "[HASH32]")
-				_test.replace(pos, 8, dev::test::RandomCode::rndByteSequence(32));
-			else
-			if (types.at(i) == "[V]")
+			else if (type == "[CODE]")
+			{
+				int random = (int)dev::test::RandomCode::randomUniInt() % 100;
```
gumb0: `randomUniInt` has a `_maxVal` parameter which can do this `% 100` for you
(which is a terrible way to generate random integers, not uniform at all)

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -42,71 +42,28 @@ int checkRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, js
...
-			if (checktest)
-			{
-				dev::eth::VMFactory::setKind(dev::eth::VMKind::JIT);
-				return checkRandomTest(dev::test::doVMTests, testmValue, debug);
-			}
-			else
-				fillRandomTest(dev::test::doVMTests, (filltest) ? testFillString : c_testExampleVMTest, filldebug);
-		}
-	}
+	if (testSuite == "StateTestsGeneral")
```
gumb0: Please don't duplicate these test suite names all over the place, make string constants, probably in a separate file (to be used by other test code)

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -167,10 +124,14 @@ void fillRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, st
...
 		_doTests(v, true);
+		for (auto& obj: v.get_obj())
+			if (!obj.second.get_obj().count("post"))
+				BOOST_ERROR("An error occured when executing random state test!");
+		dev::test::addClientInfo(v, "random generated test");
 	}
 	catch(...)
 	{
-		std::cerr << "Test fill exception!";
+		BOOST_ERROR("Error generating random test!");
```
gumb0: Which exceptions can happen here? You could display some info from exception object instead of just "unkown error" message

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -185,84 +146,91 @@ void fillRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, st
...
-			else
-			if (types.at(i) == "[HASH32]")
-				_test.replace(pos, 8, dev::test::RandomCode::rndByteSequence(32));
-			else
-			if (types.at(i) == "[V]")
+			else if (type == "[CODE]")
+			{
+				int random = (int)dev::test::RandomCode::randomUniInt() % 100;
+				if (random < 90)
+					replace = "0x"+dev::test::RandomCode::generate(5, options);
```
gumb0: Spaces around `+`

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -303,15 +303,15 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
...
 {
 	if (_maxVal == 0)
 		_maxVal = std::numeric_limits<uint64_t>::max();
 	refreshSeed();
 	int rand = randUniIntGen() % 100;
 	if (rand < 50)
-		return toCompactHexPrefixed((u256)randUniIntGen() % _maxVal);
-	return toCompactHexPrefixed((u256)randUInt64Gen() % _maxVal);
+		return toCompactHexPrefixed(_minVal + (u256)randUniIntGen() % _maxVal, 1);
```
gumb0: This is not correct at all, you'll get numbers between `_minVal` and `_minVal + maxVal - 1`

The optimal way I suggest is to create `boostIntDistrib(_minVal, _maxVal)` here and use it

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -56,7 +57,8 @@ void TestOutputHelper::initTest(json_spirit::mValue& _v)
 	NoProof::init();
 	m_timer.restart();
 	m_currentTestCaseName = boost::unit_test::framework::current_test_case().p_name;
-	std::cout << "Test Case \"" + m_currentTestCaseName + "\": " << std::endl;
+	if (!Options::get().createRandomTest)
```
gumb0: Why doesn't `TestOutputHelper::initTest(json_spirit::mValue& _v)` just call `initTest(_v.get_obj().size())` ?

test/tools/jsontests/StateTests.cpp
gumb0: As I understand all four can happen

(which is terrible, one piece of code has too many responsibilities, it should be 4 different pieces of code/functions)

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -185,84 +146,91 @@ void fillRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, st
...
+				int random = (int)dev::test::RandomCode::randomUniInt() % 100;
+				if (random < 90)
+					replace = dev::test::RandomCode::randomUniIntHex(std::numeric_limits<uint32_t>::max());
+				else
+					replace = "0x00";
+			}
+			else if (type == "[GASLIMIT]")
+				replace = dev::test::RandomCode::randomUniIntHex(dev::u256("36028797018963967"), dev::u256("18014398509481983"));
+			else if (type == "[GASPRICE]")
+				replace = dev::test::RandomCode::randomUniIntHex(dev::u256("5"));
```
gumb0: More efficient to construct it from integer: `u256(5)`

test/tools/libtesteth/TestHelper.cpp
winsvega: test filling does not supported on windows

test/tools/fuzzTesting/createRandomTest.cpp
winsvega: so far this is the only exception.

pirapira: @winsvega The suggestion here was `string const g_StateTestsGeneral = "StateTestsGeneral";` somewhere, and reuse the variable all over the place.

gumb0: Yeah, the name should be `c_` prefix according to Coding Standards

winsvega: done please review other changes so I make all the fixes at once

test/tools/fuzzTesting/createRandomTest.cpp
winsvega: done

test/tools/libtesteth/TestOutputHelper.cpp
winsvega: updated. thanks.

test/tools/jsontests/StateTests.cpp
winsvega: stop being hilarious. we don't need 4 functions here. 
just one condition is being checked here:
```  if (mode is filler  and we are not in random test  )  then  (check that filename is the same as test name) ```

Comment (winsvega) [inline test/tools/jsontests/StateTests.cpp]: the design of the tests was originally to run a function doTests  and  if passed a flag  filling it will generate the test, otherwise it will run a test. because the code for testing and generating a test is pretty much the same.

Comment (pirapira) [inline test/tools/jsontests/StateTests.cpp]: What about removing the `if` and say
```
if (TestOutputHelper::testFillerFileExists())
    BOOST_REQUIRE_MESSAGE(testname + "Filler.json" == TestOutputHelper::testFillerFileName());
```

File: test/tools/fuzzTesting/createRandomTest.cpp [hunk 29]
```diff
@@ -185,84 +145,102 @@ void fillRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, st
...
-				_test.replace(pos, 10, "0x" + dev::test::RandomCode::rndByteSequence(32));
-			else
-			if (types.at(i) == "[HASH32]")
-				_test.replace(pos, 8, dev::test::RandomCode::rndByteSequence(32));
-			else
-			if (types.at(i) == "[V]")
+			else if (type == "[CODE]")
 			{
-				int random = (int)dev::test::RandomCode::randomUniInt() % 100;
+				int random = dev::test::RandomCode::randomPercent();
```
gumb0: A lot of `dev::test::RandomCode::`, maybe it's time for `using namespace dev::test::RandomCode;` ?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -507,18 +512,58 @@ void RandomCodeOptions::setWeight(eth::Instruction _opCode, int _weight)
...
 }
 
-Address RandomCodeOptions::getRandomAddress() const
+Address RandomCodeOptions::getRandomAddress(AddressType _type) const
 {
-	if (addressList.size() > 0)
+	int index;
+	int random;
+	int probability;
+	switch(_type)
```
gumb0: space after keywords

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -303,15 +305,13 @@ std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
...
+std::string RandomCode::randomUniIntHex(u256 const& _minVal, u256 _maxVal)
 {
 	if (_maxVal == 0)
 		_maxVal = std::numeric_limits<uint64_t>::max();
 	refreshSeed();
-	int rand = randUniIntGen() % 100;
-	if (rand < 50)
-		return toCompactHexPrefixed((u256)randUniIntGen() % _maxVal);
-	return toCompactHexPrefixed((u256)randUInt64Gen() % _maxVal);
+	u256 value = _minVal + (u256)randUInt64Gen() % (_maxVal - _minVal);
```
gumb0: Now it's in `[_minVal; _maxVal-1]`, not what I would expect from function signature like this

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -507,18 +512,58 @@ void RandomCodeOptions::setWeight(eth::Instruction _opCode, int _weight)
...
+	int random;
+	int probability;
+	switch(_type)
 	{
-		int index = (int)RandomCode::randomUniInt() % addressList.size();
-		return addressList[index];
+		case AddressType::CALLONLY:
+			index = (int)RandomCode::randomUniInt() % callAddressList.size();
+			return callAddressList[index];
+		break;
```
gumb0: `break` should be on the same level of indentation as the code inside `case`
`case` is usually on the same level of indentation as `{` in our code

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -507,18 +512,58 @@ void RandomCodeOptions::setWeight(eth::Instruction _opCode, int _weight)
...
+	}
 }
 
-Address RandomCodeOptions::getRandomAddress() const
+Address RandomCodeOptions::getRandomAddress(AddressType _type) const
 {
-	if (addressList.size() > 0)
+	int index;
+	int random;
```
gumb0: `random` & `probability` have no reason to be in the outer scope of the function, make them local to the `case` code (by putting code inside `case` into braces `{}`)

gumb0: I'd do the same with `index`, too, actually (not reusing it in different `case`s)

test/tools/libtesteth/Options.cpp
```diff
@@ -279,6 +279,19 @@ Options::Options(int argc, char** argv)
...
 
+	//check restrickted options
+	if (createRandomTest)
+	{
+		string optionlist = "trValueIndex, trGasIndex, trDataIndex, nonetwork, singleTest, performance, quadratic, memory, inputLimits, bigData, wallet, stats, filltests, fillchain";
+		if (trValueIndex >= 0 || trGasIndex >= 0 || trDataIndex >= 0 || nonetwork || singleTest
+			|| 	performance || quadratic || memory || inputLimits || bigData || wallet
+			|| 	stats || filltests || fillchain)
+		{
+			cerr << "--createRandomTest could not be used with one of the options: " + optionlist << endl;
```
gumb0: Also I don't see why you need `optionlist` in a variable, just output it here as a literal

PR #4315, opened 2017-08-08
remove checkstate option

winsvega:
testeth --checkstate option was always used otherwise tests were not checked for correctness when filling


pirapira:
Merge when CI builds go fine.

PR #4317, opened 2017-08-08
Big pack of cleanups

chfast:
In preparation for "unity" builds.
Please review commit by commit.


libdevcore/TrieDB.cpp
```diff
@@ -14,20 +14,15 @@
...
-using namespace std;
-using namespace dev;
 
-h256 const dev::c_shaNull = sha3(rlp(""));
-h256 const dev::EmptyTrie = sha3(rlp(""));
+namespace dev
+{
+
+h256 const c_shaNull = sha3(rlp(""));
+const h256 EmptyTrie = sha3(rlp(""));
```
gumb0: `h256 const`

libdevcore/TrieCommon.h
```diff
@@ -26,6 +26,7 @@
 
 namespace dev
 {
+extern const h256 EmptyTrie;
```
gumb0: `extern h256 const EmptyTrie;`

PR #4318, opened 2017-08-08
Defer element removal after iteration

### LINKED ISSUE #4316: testeth --filltests --fillchain --singletest fails finding an empty file
`ETHEREUM_TEST_PATH="../../tests" ./test/testeth -t 'StateTestsGeneral' -- --filltests --checkstate --fillchain --singletest returndatacopy_following_revert`

causes

```
/home/yh/src/cpp-ethereum/test/tools/libtesteth/TestHelper.cpp(512): error: in "StateTestsGeneral/stReturnDataTest": returndatacopy_following_revert_d0g0v0_Constantinople Failed filling test with Exception: std::bad_alloc
/home/yh/src/cpp-ethereum/test/tools/libtesteth/TestHelper.cpp(522): fatal error: in "StateTestsGeneral/stReturnDataTest": Contents of ../../tests/BlockchainTests/GeneralStateTests/stReturnDataTest/returndatacopy_following_revert_d0g0v0.json is empty. Have you cloned the 'tests' repo branch develop and set ETHEREUM_TEST_PATH to its path?
```

If I remove `--fillchain`, the command works.

pirapira:
A loop in BlockChainTest.cpp removed an elment from a container while iterating through it.
Apparently this caused some memory corruption. Without this fix, I saw a testname.size() more than 100MB.

Fixes #4316

### DIFF
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -144,6 +144,8 @@ void doTransitionTest(json_spirit::mValue& _v, bool _fillin)
 void doBlockchainTestNoLog(json_spirit::mValue& _v, bool _fillin)
 {
 	map<string, json_spirit::mObject> tests;
+	vector<string> erase_list;
+
 	for (auto& i: _v.get_obj())
 	{
 		string testname = i.first;
@@ -205,8 +207,9 @@ void doBlockchainTestNoLog(json_spirit::mValue& _v, bool _fillin)
 				tests[newtestname] = jObj;
 			}
 
-			//delete source test from the json
-			_v.get_obj().erase(_v.get_obj().find(testname));
+			// will be deleted once after the loop.
+			// removing an element while in this loop causes memory corruption.
+			erase_list.push_back(testname);
 		}
 		else
 		{
@@ -220,6 +223,10 @@ void doBlockchainTestNoLog(json_spirit::mValue& _v, bool _fillin)
 			testBCTest(o);
 		}
 	}
+
+	//Delete source test from the json
+	for (auto const& testname: erase_list)
+		_v.get_obj().erase(_v.get_obj().find(testname));
 
 	//Add generated tests to the result file
 	if (_fillin)
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -144,12 +144,13 @@ void doTransitionTest(json_spirit::mValue& _v, bool _fillin)
 void doBlockchainTestNoLog(json_spirit::mValue& _v, bool _fillin)
 {
 	map<string, json_spirit::mObject> tests;
-	vector<string> erase_list;
+	vector<decltype(_v.get_obj().begin())> erase_list;
 
-	for (auto& i: _v.get_obj())
+	// range-for is not used because iterators are necessary for removing elements later.
+	for (auto i = _v.get_obj().begin(); i != _v.get_obj().end(); i++)
 	{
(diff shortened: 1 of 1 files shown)


winsvega:
need to check how this work with --filltests  and --filltests --fillchain

pirapira:
I used those for filling https://github.com/ethereum/tests/pull/255

pirapira:
The Appveyor failure looks like Hunter related.  I've pinged @chfast.

chfast:
Rebase to fix the build.

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -205,8 +207,9 @@ void doBlockchainTestNoLog(json_spirit::mValue& _v, bool _fillin)
 				tests[newtestname] = jObj;
 			}
 
-			//delete source test from the json
-			_v.get_obj().erase(_v.get_obj().find(testname));
```
chfast: Do you remove the current item? If yes, you should keep the iterator somewhere and use the common pattern: https://stackoverflow.com/questions/8234779/how-to-remove-from-a-map-while-iterating-it.

pirapira: That's a good idea.

winsvega: So which method is better. A removeList  or pattern with extra iterator?

pirapira: I'm trying the `it = erase(it)` way.

pirapira: Now I like https://stackoverflow.com/a/39719450/103518

winsvega: this does not work with json. I had to use removeList method.

pirapira: looks like so.  My test filling does not terminate with `it = erase(it)` way somehow.

pirapira: So, I now store iterators in a remove list.

pirapira:
Appveyor is still experiencing some hunter or cmake problem.  Rebasing.

test/tools/jsontests/BlockChainTests.cpp
chfast: I'm only suggesting a pattern I know. Using iterators is always better, because they are the quickest way. Remove list is another data structure and you have to do the find again.

chfast:
Rebased helped. I messed up Hunter cache a bit, but Hunter maintainer promised it will not happen again unless I repeat the mistakes again.

pirapira:
@chfast please review again

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -221,6 +225,10 @@ void doBlockchainTestNoLog(json_spirit::mValue& _v, bool _fillin)
 		}
 	}
 
+	//Delete source test from the json
+	for (auto i: erase_list)
+		_v.get_obj().erase(i);
```
chfast: I realized this is actually wrong. `erase()` invalidates map iterators, so only the first erase is ok, then you have the vector of invalid iterators. If the erase list to be used, it must contain map keys, not iterators.

pirapira: Not in a map, I read https://stackoverflow.com/a/263958/103518.  Maybe I need to check better references.

chfast: Yes, you are right. I was reading it wrong. Only the erased iterator is invalidated.

PR #4319, opened 2017-08-08
remove fields with '//' prefix from .json tests when filling

winsvega:
refactoring makeAllFieldsHex function

### DIFF
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -260,6 +260,7 @@ std::tuple<eth::State, ImportTest::ExecOutput, eth::ChangeLog> ImportTest::execu
 	return std::make_tuple(initialState, out, initialState.changeLog());
 }
 
+//convert json object of "key" : "stringdeciamal"  into "key" : "stringhex"
 json_spirit::mObject& ImportTest::makeAllFieldsHex(json_spirit::mObject& _o, bool _isHeader)
 {
 	static const set<string> hashes {"bloom" , "coinbase", "hash", "mixHash", "parentHash", "receiptTrie",
@@ -268,17 +269,21 @@ json_spirit::mObject& ImportTest::makeAllFieldsHex(json_spirit::mObject& _o, boo
 	for (auto const& i: _o)
 	{
 		bool isHash = false;
-		std::string key = i.first;
+		std::string const& key = i.first;
 
+		//do not touch the bytecode
 		if (key == "extraData" || key == "code")
 			continue;
 
+		//convert data only if it is an integer field
 		if (key == "data" && i.second.type() != json_spirit::int_type)
 			continue;
 
+		//do not apply 0x prefix to the hashes
 		if (hashes.count(key))
 			isHash = true;
 
+		//nonce is an string decimal field if it is not in the header
 		if (_isHeader && key == "nonce")
 			isHash = true;
 
@@ -291,22 +296,20 @@ json_spirit::mObject& ImportTest::makeAllFieldsHex(json_spirit::mObject& _o, boo
 			str = value.get_str();
 		else if (value.type() == json_spirit::array_type)
 		{
-			bool isArrayOfString = true;
+			//convert transaction value, gaslimit arrays
 			json_spirit::mArray arr;
-			for (auto const& j: value.get_array())
+			for (auto& j: value.get_array())
 			{
-				if (j.type() != json_spirit::str_type)
+				if (j.type() == json_spirit::str_type)
 				{
-					isArrayOfString = false;
-					break;
+					str = j.get_str();
+					str = (str.substr(0, 2) == "0x") ? str : toCompactHexPrefixed(toInt(str), 1);
+					arr.push_back(str);
 				}
-
-				str = j.get_str();
-				arr.push_back((str.substr(0, 2) == "0x") ? str : toCompactHexPrefixed(toInt(str), 1));
+				else
+					arr.push_back(j);
 			}
-
-			if (isArrayOfString)
-				_o[key] = arr;
+			_o[key] = arr;
 			continue;
 		}
 		else continue;
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -264,14 +264,16 @@ json_spirit::mObject& ImportTest::makeAllFieldsHex(json_spirit::mObject& _o, boo
 {
 	static const set<string> hashes {"bloom" , "coinbase", "hash", "mixHash", "parentHash", "receiptTrie",
 									 "stateRoot", "transactionsTrie", "uncleHash", "currentCoinbase",
-									 "previousHash", "to", "address", "caller", "origin", "secretKey", "data", "extraData"};
-
+									 "previousHash", "to", "address", "caller", "origin", "secretKey", "extraData"};
 	for (auto const& i: _o)
 	{
 		bool isHash = false;
 		std::string key = i.first;
 
-		if (key == "data" || key == "extraData")
+		if (key == "extraData" || key == "code")
+			continue;
+
+		if (key == "data" && i.second.type() != json_spirit::int_type)
 			continue;
 
 		if (hashes.count(key))
@@ -289,13 +291,22 @@ json_spirit::mObject& ImportTest::makeAllFieldsHex(json_spirit::mObject& _o, boo
 			str = value.get_str();
 		else if (value.type() == json_spirit::array_type)
 		{
+			bool isArrayOfString = true;
 			json_spirit::mArray arr;
 			for (auto const& j: value.get_array())
 			{
+				if (j.type() != json_spirit::str_type)
+				{
+					isArrayOfString = false;
+					break;
+				}
+
 				str = j.get_str();
 				arr.push_back((str.substr(0, 2) == "0x") ? str : toCompactHexPrefixed(toInt(str), 1));
 			}
-			_o[key] = arr;
+
+			if (isArrayOfString)
+				_o[key] = arr;
 			continue;
 		}
 		else continue;
@@ -532,7 +543,12 @@ void parseJsonIntValueIntoVector(json_spirit::mValue const& _json, vector<int>& 
 			_out.push_back(val.get_int());
 	}
 	else
-		_out.push_back(_json.get_int());
+	{
+		if (_json.type() == json_spirit::int_type)
+			_out.push_back(_json.get_int());
+		else if (_json.type() == json_spirit::str_type)
+			_out.push_back((int)toInt(_json));
+	}
 }
 
 template<class T>
@@ -768,7 +784,7 @@ int ImportTest::exportTest(bytes const& _output)
 
 	// export pre state
 	m_testObject["pre"] = fillJsonWithState(m_statePre);
-	m_testObject["env"] = makeAllFieldsHex(m_testObject["env"].get_obj());
-	m_testObject["transaction"] = makeAllFieldsHex(m_testObject["transaction"].get_obj());
+	m_testObject["env"] = m_testObject["env"].get_obj();
+	m_testObject["transaction"] = m_testObject["transaction"].get_obj();
 	return err;
 }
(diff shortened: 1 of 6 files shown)


test/tools/libtesteth/ImportTest.cpp
```diff
@@ -289,13 +291,22 @@ json_spirit::mObject& ImportTest::makeAllFieldsHex(json_spirit::mObject& _o, boo
...
+					break;
+				}
+
 				str = j.get_str();
 				arr.push_back((str.substr(0, 2) == "0x") ? str : toCompactHexPrefixed(toInt(str), 1));
 			}
-			_o[key] = arr;
+
+			if (isArrayOfString)
+				_o[key] = arr;
```
pirapira: Why do you sometimes give up the format conversion and leave the original `_o[key]`?

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -534,6 +535,33 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
...
 
+void postParseJson(json_spirit::mValue& _obj)
+{
+	if(_obj.type() == json_spirit::obj_type)
+	{
+		for (auto& i: _obj.get_obj())
+		{
+			if (i.first.substr(0, 2) == "//")
+			{
+				_obj.get_obj().erase(_obj.get_obj().find(i.first));
```
pirapira: Do not erase an element of `_obj.get_obj()` while going through `_obj.get_obj()`.

This caused the problem https://github.com/ethereum/cpp-ethereum/issues/4316

test/tools/libtesteth/ImportTest.cpp
winsvega: no way to convert array object

pirapira: Why?  Can't you convert each element of the array?

winsvega: the function makeAllfieldsHex takes a json object 
{
  "key" : "decvalue",
  "key" : "decvalue",
  "key" : intvalue
}

winsvega: which could be an element of some array. thus makeallfield hex being executed on postParseJson for each element of the array

winsvega: you see logic adds up and make a mess. maybe I don't have to parse an array in makeallfieldhex if it is being executed from postparseJson. need to test that.

pirapira: The function `makeAllfiledsHex` should not care if the passed object is part of an array or not.  That concern should be addressed somewhere out of the `makeAllfieldsHex` function.

pirapira: If you never expect an array in `makeAllfieldsHex` function, then you should just raise an error.

pirapira: If you expect an array in `makeAllfieldsHex`, why don't you perform something on each element of the array?

winsvega: that is exactly what this function does. 
updated the code with comments

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -289,11 +296,18 @@ json_spirit::mObject& ImportTest::makeAllFieldsHex(json_spirit::mObject& _o, boo
...
-				str = j.get_str();
-				arr.push_back((str.substr(0, 2) == "0x") ? str : toCompactHexPrefixed(toInt(str), 1));
+				if (j.type() == json_spirit::str_type)
+				{
+					str = j.get_str();
+					str = (str.substr(0, 2) == "0x") ? str : toCompactHexPrefixed(toInt(str), 1);
+					arr.push_back(str);
+				}
+				else
+					arr.push_back(j);
```
pirapira: If `j.type()` is `object_type`, the elements of `j` should be modified too.

winsvega: that should be done by postParseJson() that works recursive

winsvega: makeAllFieldsHex is not designed to work recursively

pirapira: Now I'm confused.  Naturally I would make `postParseJson()` call `makeAllFieldsHex()` and `makeAllFieldsHex()` call `postParseJson()`.  But it seems like something else is happening.

pirapira: No, `postParseJson()` does not modify what's left from `makeAllFieldsHex`.

winsvega: check the line 553

pirapira: @winsvega OK. The code is scanning the same object twice.  I think you should rename the functions as `postParseJsonObject` and `postParseJsonArray` and make them call each other.  That will be much simpler.

winsvega: one code is scanning the object. the makeAllFieldsHex do an operation. 
I mean in the future we might want to add more post processing operations

pirapira: I'll change this code as I like it after this PR.  I hope you'll like the result.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -501,6 +501,7 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
 			json_spirit::read_string(s, v);
 			doTests(v, true);
 			addClientInfo(v, testfilename);
+			postParseJson(v);
```
pirapira: Shouldn't this happen just after the parsing, 4 lines above?

winsvega: doTests adds new sections to json. addClinetInfo should really be inside postParseJson

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -534,6 +535,36 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
 	}
 }
 
+void postParseJson(json_spirit::mValue& _obj)
+{
+	if(_obj.type() == json_spirit::obj_type)
```
pirapira: A space is missing.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -534,6 +535,36 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
...
+			}
+
+			if (i.second.type() != json_spirit::str_type)
+			{
+				postParseJson(i.second);
+				if (i.first != "_info" && i.second.type() == json_spirit::obj_type)
+					i.second = ImportTest::makeAllFieldsHex(i.second.get_obj(), i.first == "env" ? true : false);
+			}
+		}
+		for(auto& i: removeList)
```
pirapira: A space is missing as well.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -534,6 +535,36 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
 	}
 }
 
+void postParseJson(json_spirit::mValue& _obj)
+{
+	if(_obj.type() == json_spirit::obj_type)
+	{
+		std::list<string> removeList;
+		for (auto& i: _obj.get_obj())
```
pirapira: I suggest making an independent function for the body of this `for` loop.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -534,6 +535,36 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
...
+		{
+			if (i.first.substr(0, 2) == "//")
+			{
+				removeList.push_back(i.first);
+				continue;
+			}
+
+			if (i.second.type() != json_spirit::str_type)
```
pirapira: You should remove this `if`, and execute the following two sentences always.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -260,23 +260,30 @@ std::tuple<eth::State, ImportTest::ExecOutput, eth::ChangeLog> ImportTest::execu
...
 }
 
+//convert json object of "key" : "stringdeciamal"  into "key" : "stringhex"
 json_spirit::mObject& ImportTest::makeAllFieldsHex(json_spirit::mObject& _o, bool _isHeader)
 {
 	static const set<string> hashes {"bloom" , "coinbase", "hash", "mixHash", "parentHash", "receiptTrie",
 									 "stateRoot", "transactionsTrie", "uncleHash", "currentCoinbase",
-									 "previousHash", "to", "address", "caller", "origin", "secretKey", "data", "extraData"};
-
+									 "previousHash", "to", "address", "caller", "origin", "secretKey", "extraData"};
```
pirapira: Can you replace the string literals with constants?

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -289,11 +296,18 @@ json_spirit::mObject& ImportTest::makeAllFieldsHex(json_spirit::mObject& _o, boo
 			str = value.get_str();
 		else if (value.type() == json_spirit::array_type)
```
pirapira: In the current design, this function should not look inside arrays because `postParseJson()` does look into the array.

winsvega: it does. but to remove '//' prefixed fields. 
make all fields hex takes an object to convert. an object like 
"key" : "value" 

this case describes a situation 
"key" : [
     "value" 
     "value"
]

test/tools/libtesteth/ImportTest.cpp
winsvega: this string is already constant

test/tools/libtesteth/ImportTest.cpp
pirapira: I don't understand.  `postParseJson()` does recursion.  So `makeAllFieldsHex` should not.  Otherwise two different functions traverse the same branch of JSON.  That's confusing.

test/tools/libtesteth/ImportTest.cpp
pirapira: They are literals.  They are not constants.

test/tools/libtesteth/TestHelper.cpp
pirapira: To be honest, I don't think `doTests` should modify the input document, but return an output JSON file when it's doing filling.  Moreover, `doTests` should be responsible for the correct formatting of the output.   That's for a separate PR.

test/tools/libtesteth/ImportTest.cpp
winsvega: thats what I am trying to make a point here. 
make allFields just convert the values. there should be no recursion.
it is just that there is a case when value could be an array of values

test/tools/libtesteth/ImportTest.cpp
winsvega: I see no point in making this const string a const string of const string

pirapira:
@winsvega will you make a smaller PR just about removing `//` fields?  That would be quicker to review.

test/tools/libtesteth/ImportTest.cpp
winsvega: yes it is a hardcode but do we have a choise here?

pirapira: The point is to make the C++ compiler detect misspellings.

pirapira: Yes, we have a choice to define `string const g_bloom = "bloom"` and use `g_bloom` here.  I'm suggesting to do that.

pirapira:
I'm suggesting this because
* I support the feature of removing fields whose names start with `//`.
* I do not like the direction taken about `makeAllFieldsHex`.  I prefer generating the correct format from the begnning, rather than creating something misformed first and fixing the format later.  The "refactoring" about `makeAllFieldsHex` is moving in the opposite direction.  It's trying to fix the formats after generating JSON objects.

test/tools/libtesteth/ImportTest.cpp
pirapira: @winsvega let's leave `makeAllFieldsHex` aside, and work on a smaller PR just about `//`.

winsvega:
hm. if I remove makeAllFieldsHex. there would be like 10ths of lines with 
"0x" + all over the test suites. It would be hard to catch every place. so I did it in one place that is common for all tests.

pirapira:
@winsvega Let's separate `//` and `makeAllFieldsHex`.

PR #4320, opened 2017-08-08
standardized vmtrace logs

cdetrio:
This fixes the VM trace logs to be aligned with the other clients, following the work-in-progress standard format https://github.com/ethereum/tests/issues/249.

Some edge cases still remain, which might cause the cpp trace to differ under some circumstances, but this is a good start (traces are equivalent for all EIP158 state tests).

### DIFF
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -130,6 +130,7 @@ void StandardTrace::operator()(uint64_t _steps, uint64_t PC, Instruction inst, b
 	r["pc"] = toString(PC);
 	r["gas"] = toString(gas);
 	r["gasCost"] = toString(gasCost);
+	r["depth"] = toString(ext.depth);
 	if (!!newMemSize)
 		r["memexpand"] = toString(newMemSize);
 
--- a/libevm/VM.cpp
+++ b/libevm/VM.cpp
@@ -196,7 +196,7 @@ owning_bytes_ref VM::exec(u256& _io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
 	m_ext = &_ext;
 	m_schedule = &m_ext->evmSchedule();
 	m_onOp = _onOp;
-	m_onFail = &VM::onOperation;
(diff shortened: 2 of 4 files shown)


gcolvin:
One thing worries me - the ON_OP() macro ultimately calls the _onOp() funtion object passed to VM::exec(), which is the hook for more than logging.  Moving the place it is called will cause different values to be passed in, which apparently is what we want for logging, but might break other tools.  @yann300 ?

gumb0:
@gcolvin I haven't seen it used for anything besides tracing

gumb0:
I would remove all mentions on `m_onFail` and use `ON_OP()` in  `VM::throwBadStack`

yann300:
That might break some stuff in remix if we change where on_op is called (is on_op called after or before the actual execution of the step). 
Just ping us when all the implementations agree on smt and we'll adapt.

gcolvin:
@yann300 @cdetrio  I'm wondering if this is premature, given that the standard format isn't done?  It seems as soon we merge this remix will break.
@gumb0 If m_onFail isn't used elsewhere I agree it should go.  I'm on the road right now and can't check the code.

gcolvin:
@chfast I have some worries I've already stated.
Right now injected opcodes are just passed to m_onOp like any other opcode, which doesn't seem to be causing trouble.  I think geth does a similar transformation, but I'm not sure how it handles tracing it, @obscuren.
I think the current C++ VM could reconstruct the original opcodes, but in principle, a VM could transform the codes into something completely different, even machine code, so there is a bigger problem of how and whether to trace and debug optimized code.

pirapira:
Trace is an artificial layer and when we go down the optimization we will need to fake it.

@cdetrio says clients produce the same trace with this, I think it should be added.

If we try to keep remix working, one way is to emit the new format with a special command line option, and keep the old behavior without that option.  Is that preferable?

yann300:
I'm totally fine to update remix to the new trace as long as all the clients have the same format.

pirapira:
I'm buying this
> traces are equivalent for all EIP158 state tests

PR #4327, opened 2017-08-10
Support syncing without the full previous chain

gumb0:
Methods for saving to Extras DB block hash of the block where our part of the chain starts (will be used after importing snapshot)

BlockChainSync uses chain start block number to avoid downloading earlier blocks.


gumb0:
@chfast yes, I could do that

PR #4330, opened 2017-08-11
remove "//" from json before filling the test

winsvega:
remove comments from test filler so it wont displayed in the final test

### DIFF
--- a/test/tools/libtesteth/TestHelper.cpp
+++ b/test/tools/libtesteth/TestHelper.cpp
@@ -499,6 +499,7 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
 			BOOST_REQUIRE_MESSAGE(s.length() > 0, "Contents of " + testfilename + " is empty.");
 
 			json_spirit::read_string(s, v);
+			removeComments(v);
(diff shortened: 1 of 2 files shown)


test/tools/libtesteth/TestHelper.cpp
```diff
@@ -534,6 +534,36 @@ void executeTests(const string& _name, const string& _testPathAppendix, const st
...
+			_obj.get_obj().erase(_obj.get_obj().find(i));
+	}
+	else if (_obj.type() == json_spirit::array_type)
+	{
+		for (auto& i: _obj.get_array())
+			postParseJson(i);
+	}
+
+	//Add client info to the top level of the test
+	if (!_testfilename.empty())
```
pirapira: I think we should add the client info always.

winsvega: this functions is recursive. testfilename is not empty only on a top level

pirapira: That's wicked.  Maybe put it back out of this function.  I think that keeps this function a simple syntactic transformation, not adding information, just removing comments.

PR #4332, opened 2017-08-11
randomPercent to the RandomCode

winsvega:
randomPercent() in RandomCode

### DIFF
--- a/test/tools/fuzzTesting/fuzzHelper.cpp
+++ b/test/tools/fuzzTesting/fuzzHelper.cpp
@@ -83,14 +83,14 @@ namespace test
 {
 
 boost::random::mt19937 RandomCode::gen;
-boostIntDistrib RandomCode::prcentDist = boostIntDistrib (0, 100);
+boostIntDistrib RandomCode::percentDist = boostIntDistrib (0, 100);
 boostIntDistrib RandomCode::opCodeDist = boostIntDistrib (0, 255);
 boostIntDistrib RandomCode::opLengDist = boostIntDistrib (1, 32);
 boostIntDistrib RandomCode::opMemrDist = boostIntDistrib (0, 10485760);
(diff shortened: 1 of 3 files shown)


axic:
Why not just add a `max` parameter to `randomUintGen`?

winsvega:
that is what we are trying to avoid. before it was working with randomUintGen(maxValue)

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -83,12 +83,14 @@ namespace test
 {
 
 boost::random::mt19937 RandomCode::gen;
+boostIntDistrib RandomCode::prcentDist = boostIntDistrib (0, 100);
```
pirapira: `percentDist`.

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -83,12 +83,14 @@ namespace test
...
 boostIntDistrib RandomCode::opCodeDist = boostIntDistrib (0, 255);
 boostIntDistrib RandomCode::opLengDist = boostIntDistrib (1, 32);
 boostIntDistrib RandomCode::opMemrDist = boostIntDistrib (0, 10485760);
 boostIntDistrib RandomCode::uniIntDist = boostIntDistrib (0, 0x7fffffff);
 boostUint64 RandomCode::uInt64Dist = boostUint64 (0, std::numeric_limits<uint64_t>::max());
 
+boostIntGenerator RandomCode::randPrcentGen = boostIntGenerator(gen, prcentDist);
```
pirapira: `randPercentGen`.  Somehow I sense this was done to keep the widths uniform, but it's anyway not always possible to keep the same number of characters.

PR #4334, opened 2017-08-13
Fix for leaking unix sockets in IPC RPC

dirtyfilthy:
Currently sockets in libweb3jsonrpc/UnixSocketServer.cpp are not closed after a request is completed. After performing many IPC requests, the sockets hang around and eventually exhaust available file descriptors. This is easily verified by lsof-ing the process after performing IPC requests.

This pull request closes sockets after the response is generated, and also adds shutdown(sock. O_RDRW) to CloseConnection to ensure sockets are closed in a timely manner.

### DIFF
--- a/libweb3jsonrpc/UnixSocketServer.cpp
+++ b/libweb3jsonrpc/UnixSocketServer.cpp
@@ -111,7 +111,7 @@ void UnixDomainSocketServer::Listen()
 			DEV_GUARDED(x_sockets)
 				m_sockets.insert(connection);
 
-			std::thread handler([this, connection](){ GenerateResponse(connection); });
+			std::thread handler([this, connection](){ GenerateResponse(connection); CloseConnection(connection);});
 			handler.detach();
 		}
 	}
@@ -119,6 +119,7 @@ void UnixDomainSocketServer::Listen()
 
 void UnixDomainSocketServer::CloseConnection(int _socket)
 {
+	shutdown(_socket, SHUT_RDWR);
 	close(_socket);
 }
 


libweb3jsonrpc/UnixSocketServer.cpp
```diff
@@ -111,14 +111,15 @@ void UnixDomainSocketServer::Listen()
 			DEV_GUARDED(x_sockets)
 				m_sockets.insert(connection);
 
-			std::thread handler([this, connection](){ GenerateResponse(connection); });
+			std::thread handler([this, connection](){ GenerateResponse(connection); CloseConnection(connection);});
```
chfast: Please reformat it. Should look like:
```
std::thread handler([this, connection]()
{
    ...
}
```

libweb3jsonrpc/UnixSocketServer.cpp
```diff
@@ -111,14 +111,15 @@ void UnixDomainSocketServer::Listen()
...
 			handler.detach();
 		}
 	}
 }
 
 void UnixDomainSocketServer::CloseConnection(int _socket)
 {
+	shutdown(_socket, SHUT_RDWR);
```
chfast: I don't understand why this is needed. Seems to be redundant to `close()`.

In any case, can you also handle error codes?

chfast:
@dirtyfilthy Are you still around?

libweb3jsonrpc/UnixSocketServer.cpp
axic: `shutdown` should immediately send the signal to the other side, while `close` would wait for regular timeouts. Not sure how this applies to unix sockets though.

chfast: According to docs, close() does shutdown(). But docs are not very detailed.

axic: The top two answers are give some insight: https://stackoverflow.com/questions/4160347/close-vs-shutdown-socket

pirapira:
@dirtyfilthy ?

chfast:
I'm for keeping the `shutdown()`. Only the formatting left. I will take care of that myself.

chfast:
Moved to #4421

PR #4335, opened 2017-08-14
refactor fuzzTest type replacement

winsvega:
slowly bring up --createRandomTest  option back

### DIFF
--- a/test/tools/fuzzTesting/createRandomTest.cpp
+++ b/test/tools/fuzzTesting/createRandomTest.cpp
@@ -207,55 +207,54 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
 	for (std::map<std::string, std::string>::const_iterator it = _varMap.begin(); it != _varMap.end(); it++)
 		types.push_back(it->first);
 
-	for (unsigned i = 0; i < types.size(); i++)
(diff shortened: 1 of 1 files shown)


test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -207,55 +207,54 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
...
 			}
 			else
-				_test.replace(pos, types.at(i).size(), _varMap.at(types.at(i)));
+			{
+				//Replace type from varMap if varMap is set
+				if (_varMap.count(type))
+					replace = _varMap.at(type);
+				else
+					BOOST_ERROR("Skipping undeclared type: " + type);
+			}
```
pirapira: Does this make sesne?
```
if (string.empty())
    BOOST_ERROR("Trying to replace a tag with the empty string.  Something is wrong.");
```

winsvega: only if the type is not "[CODE]" or "[DESTADDR]" that will be introduced next

pirapira: OK.

PR #4338, opened 2017-08-15
EIP649 reward reduction and difficulty bomb delay

pirapira:
The reward-reduction and difficulty bomb delay have been both implemented.

Tests are updated so that `testeth` with no options passes.

### DIFF
--- a/libethcore/ChainOperationParams.cpp
+++ b/libethcore/ChainOperationParams.cpp
@@ -67,10 +67,10 @@ u256 ChainOperationParams::u256Param(string const& _name) const
 	return u256(fromBigEndian<u256>(fromHex(at)));
 }
 
-u256 ChainOperationParams::blockReward(bool isByzantiumOrLater) const
+u256 ChainOperationParams::blockReward(EVMSchedule const& _schedule) const
 {
-	if (isByzantiumOrLater)
-		return 3 * ether;
+	if (_schedule.blockRewardOverwrite)
+		return *_schedule.blockRewardOverwrite;
 	else
 		return m_blockReward;
 }
--- a/libethcore/ChainOperationParams.cpp
+++ b/libethcore/ChainOperationParams.cpp
@@ -75,7 +75,7 @@ u256 ChainOperationParams::blockReward(bool isByzantiumOrLater) const
 		return m_blockReward;
 }
 
-void ChainOperationParams::setBlockReward(u256 const& newBlockReward)
+void ChainOperationParams::setBlockReward(u256 const& _newBlockReward)
 {
-	m_blockReward = newBlockReward;
+	m_blockReward = _newBlockReward;
 }
--- a/libethcore/ChainOperationParams.cpp
+++ b/libethcore/ChainOperationParams.cpp
@@ -53,7 +53,7 @@ ChainOperationParams::ChainOperationParams()
 		{"registrar", "5e70c0bbcd5636e0f9f9316e9f8633feb64d4050"},
 		{"networkID", "0x0"}
 	};
-	blockReward = u256("0x4563918244F40000");
+	m_blockReward = u256("0x4563918244F40000");
 }
 
 u256 ChainOperationParams::u256Param(string const& _name) const
@@ -65,4 +65,17 @@ u256 ChainOperationParams::u256Param(string const& _name) const
 		at = it->second;
 
 	return u256(fromBigEndian<u256>(fromHex(at)));
+}
+
+u256 ChainOperationParams::blockReward(bool isByzantiumOrLater) const
+{
+	if (isByzantiumOrLater)
+		return 3 * ether;
+	else
+		return m_blockReward;
+}
+
+void ChainOperationParams::setBlockReward(u256 const& newBlockReward)
+{
+	m_blockReward = newBlockReward;
 }
--- a/libethcore/ChainOperationParams.h
+++ b/libethcore/ChainOperationParams.h
@@ -73,7 +73,11 @@ struct ChainOperationParams
 	std::string sealEngineName = "NoProof";
 
 	/// General chain params.
-	u256 blockReward = 0;
+private:
+	u256 m_blockReward;
+public:
+	u256 blockReward(bool isByzantiumOrLater) const;
+	void setBlockReward(u256 const& blockReward);
 	u256 maximumExtraDataSize = 1024;
 	u256 accountStartNonce = 0;
 	bool tieBreakingGas = true;
--- a/libethcore/ChainOperationParams.h
+++ b/libethcore/ChainOperationParams.h
@@ -76,8 +76,8 @@ struct ChainOperationParams
 private:
 	u256 m_blockReward;
 public:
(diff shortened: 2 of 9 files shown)


test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -531,8 +552,9 @@ void testBCTest(json_spirit::mObject& _o)
 		//check the balance before and after the block according to mining rules
 		if (blockFromFields.blockHeader().parentHash() == preHash)
 		{
+			bool const isByzantiumOrLater = ChainBranch::networkIsByzantiumOrLater(blockChainName);
```
pirapira: @winsvega what's the best way to know if it's already Byzantium here?

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -563,10 +585,10 @@ void testBCTest(json_spirit::mObject& _o)
 	ImportTest::compareStates(postState, blockchain.topBlock().state());
 }
 
-bigint calculateMiningReward(u256 const& _blNumber, u256 const& _unNumber1, u256 const& _unNumber2)
+bigint calculateMiningReward(u256 const& _blNumber, u256 const& _unNumber1, u256 const& _unNumber2, bool isByzantium)
 {
 	unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(test::TestBlockChain::s_sealEngineNetwork)).createSealEngine());
-	bigint baseReward = se->chainParams().blockReward;
+	bigint baseReward = se->chainParams().blockReward(isByzantium);
```
winsvega: this should be done inside chain params. the thing is that blockReward in this case will depend on blockNumber and evm schedule like other fork dependent params.

isByzantium = number  >=  chainParams().u256Param("ByzantiumBlockNumber")

libethcore/ChainOperationParams.cpp
```diff
@@ -66,3 +66,16 @@ u256 ChainOperationParams::u256Param(string const& _name) const
 
 	return u256(fromBigEndian<u256>(fromHex(at)));
 }
+
+u256 ChainOperationParams::blockReward(bool isByzantiumOrLater) const
+{
+	if (isByzantiumOrLater)
+		return u256(3 * (bigint(10) ^ 18)); // 3 ETH
```
chriseth: Is `^` implemented as exp or as xor?

chriseth: I think there should also be a global constant called `ether` somewhere with the right value.

pirapira: I wondered these too.

test/tools/jsontests/BlockChainTests.cpp
pirapira: Thank you.

winsvega: I mean it feels like isByzantium should be calculated inside chainParams. 
Cause it knows the schedule from s_sealEngineNetwork. if should be able to return the reward if you give it the blocknumber

pirapira: I'll try.

pirapira: No, `ChainParams` does not know the schedule.  It does not have access to a `sealEngineNetwork`.

winsvega: it has the genesis transition block number for sure. 
how is it done for schedule variables?

winsvega: I mean it is strange that we calculate isBizantium manually outside of getBlockReward

pirapira: What is the genesis transition block?  What are schedule variables?

Is `ChainParams` supposed to know when Byzantium starts?  In that case, I will need to fix `ChainParams` class first.

winsvega: ChainParams  constructed from genesisInfo so it does know all the transition blocks. 
Schedule variables are: EVMSchedule.h

Ah, and here there is a function 
EVMSchedule const& SealEngineBase::evmSchedule(u256 const& _blockNumber) const

pirapira: OK.  I'll file another PR first to fix this first.

pirapira: @winsvega so, the SealEngine knows when Byzantium starts.  ChainParams should not know that.

winsvega: why? chainParams initialized from genesisInfo. it should know that. cause seal engine knows that from chainParams.

``` chainParams().u256Param("byzantiumForkBlock")  ```  
indicates when Byzanium starts.  

do you think we should look at this issue more carefully and perhaps do a refactoring PR for libethereum instead?

Comment (pirapira) [inline test/tools/jsontests/BlockChainTests.cpp]: No, I came up with one solution.

**Comment by pirapira** [comment]
@winsvega shall these cases appear in the test case spreadsheet?

**Comment by winsvega** [comment]
I mean we could add the tests with this PR. that was the idea of a submodule. 
to cover the PR changes with relevant tests

**Comment by pirapira** [comment]
@winsvega I prefer a separate PR for adding new tests.  We should deliver the fixed tests earlier, even without the new tests.

**Comment by winsvega** [comment]
if we have the tests at least we know the change on this PR is covered. that what is the point

**Comment by pirapira** [comment]
Let me see if I need to fix any fillers.  If yes, we have the changes covered somehow.

**Comment by pirapira** [comment]
I had an `ethereum/tests` commit on my local machine, pushed that and used that in the submodule.

**Comment by chfast** [comment]
At first look, I don't like how we manage chain configuration. Do we need to use JSON-like structure and constructs like "get u256 param" all the way to the bottom?

**Comment by pirapira** [comment]
@chfast I believe we should do
```
_bc.chainParams().EIP158ForkBlock()
```
instead of
```
_bc.chainParams().u256Param("EIP158ForkBlock")
```
And, remove `u256Param` from the public interface of `ChainParams`.

I want to do it in a separate PR.  Do you want it before this one?

**Comment by chfast** [comment]
After.

I'd like to review the chain config design later on...
For EVM-C single enum is enough. Also we can consider configs for chains + separated config for chain switches (hard forks).

File: libethcore/EVMSchedule.h [hunk 4]
```diff
@@ -71,6 +72,8 @@ struct EVMSchedule
 	unsigned blockhashGas = 20;
 	unsigned maxCodeSize = unsigned(-1);
 
+	boost::optional<u256> blockRewardOverwrite{};
```
chfast: Explicit initialization not needed.

libethcore/ChainOperationParams.h
```diff
@@ -73,7 +73,11 @@ struct ChainOperationParams
 	std::string sealEngineName = "NoProof";
 
 	/// General chain params.
-	u256 blockReward = 0;
+private:
+	u256 m_blockReward = 0;
```
chfast: Explicit initialization not needed.

pirapira:
@chfast removed the unnecessary initializations.

PR #4344, opened 2017-08-16
[WIP] Increase TransientDirectory performance on Linux

chfast:


chfast:
This is only PoC, probably not for merging.

winsvega:
so have you tested it? 
current setup leave tmp directories which sometimes might collide and lead to errors on travis

chfast:
I think this can be closed safely because it does not provide any value any more. @halfalicious @gumb0

halfalicious:
I agree that this can be closed since we run all tests using MemoryDB by default

PR #4340, opened 2017-08-16
add --createRandomTest option

winsvega:
remove legacy code

Usage:

``` ./testeth -t StateTestsGeneral -- --createRandomTest ```

PR #4346, opened 2017-08-17
Fuzz Helper changes

chfast:


chfast:
I don't get what changes do you request here @winsvega.

winsvega:
I need more time. first merge this PR 
https://github.com/ethereum/cpp-ethereum/pull/4340
then test the randomCode option. 

then review your changes

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -437,13 +432,11 @@ std::string RandomCode::fillArguments(eth::Instruction _opcode, RandomCodeOption
 			code += getPushCode(randOpMemrGen());	//memlen1
 			break;
 		default:
-			smart = false;
```
winsvega: I think we should throw an exception here. if smart is set but there is no instruction for an opcode here should be an exception

PR #4341, opened 2017-08-16
Separate input and output in doTest functions

pirapira:
Currently, various `doTest(json_spirit::mValue&)` functions modify the JSON value.  Sometimes a filler is given but gradually morphed into a finished test.  I discussed this structure with @winsvega and @gumb0.  The conclusion was we should separate the input and the output.

This PR just changes the type of `doTest` so that it takes a const reference and returns a JSON value.  The internal workings of each function are not changed yet.

PR #4348, opened 2017-08-17
Change the type of doRlpTests

pirapira:
_fillin = true was never used, so _fillin option was removed.

This is split from #4341


winsvega:
I feel like you removing here something important even if it is not used anymore. 
I remember there was an idea to test RLP format more carefully. there is even a random rlp generation function in random test. this was used one day. and for some reason it was frozen.

pirapira:
Code without tests is bad.  Dead code is evil.

PR #4352, opened 2017-08-18
Replace many std::endl with "\n" sometimes with std::flush

pirapira:
Spin off from this discussion https://github.com/ethereum/cpp-ethereum/pull/4340#discussion_r133702673

### DIFF
--- a/test/tools/fuzzTesting/createRandomTest.cpp
+++ b/test/tools/fuzzTesting/createRandomTest.cpp
@@ -45,7 +45,7 @@ int createRandomTest()
 	dev::test::Options& options = const_cast<dev::test::Options&>(dev::test::Options::get());
 	if (options.rCurrentTestSuite != test::c_StateTestsGeneral)
 	{
-		std::cerr << "Error! Test suite '" + options.rCurrentTestSuite + "' not supported! (Usage -t <TestSuite>)" << std::endl;
+		std::cerr << "Error! Test suite '" + options.rCurrentTestSuite + "' not supported! (Usage -t <TestSuite>)\n";
 		return 1;
 	}
 	else
@@ -81,12 +81,12 @@ int checkRandomTest(std::function<void(json_spirit::mValue&, bool)> _doTests, js
 	}
 	catch (dev::Exception const& _e)
 	{
-		std::cout << " Failed test with Exception: " << diagnostic_information(_e) << std::endl;
+		std::cout << " Failed test with Exception: " << diagnostic_information(_e) << "\n";
 		ret = 1;
 	}
 	catch (std::exception const& _e)
 	{
-		std::cout << " Failed test with Exception: " << _e.what() << std::endl;
+		std::cout << " Failed test with Exception: " << _e.what() << "\n";
 		ret = 1;
 	}
 	return ret;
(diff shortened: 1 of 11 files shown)


winsvega:
please paste the output of testeth to see how it affects the output

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -72,7 +72,7 @@ int createRandomTest(std::vector<char*> const& _parameters)
 
 	if (testSuite == "")
 	{
-		std::cerr << "Error! Test suite not supported! (Usage -t TestSuite)" << std::endl;
+		std::cerr << "Error! Test suite not supported! (Usage -t TestSuite)\n";
```
winsvega: maybe replace std::endl with  some marco or function/ custom operator?

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -642,7 +642,7 @@ void ImportTest::checkGeneralTestSectionSearch(json_spirit::mObject const& _expe
 				int errcode = compareStates(expectState, postState, stateMap, WhenError::Throw);
 				if (errcode > 0)
 				{
-					cerr << trInfo << std::endl;
+					cerr << trInfo << "\n" << std::flush;
```
chfast: Why flush? cerr is not buffered so flush is never needed (am I wrong?). Flush is only needed in some rare cases in cout.

test/unittests/libdevcrypto/MemTrie.cpp
```diff
@@ -49,7 +49,7 @@ class MemTrieNode
 	void putRLP(RLPStream& _parentStream) const;
 
 #if ENABLE_DEBUG_PRINT
-	void debugPrint(std::string const& _indent = "") const { std::cerr << std::hex << hash256() << ":" << std::dec << std::endl; debugPrintBody(_indent); }
+	void debugPrint(std::string const& _indent = "") const { std::cerr << std::hex << hash256() << ":" << std::dec << "\n" << std::flusyh; debugPrintBody(_indent); }
```
chfast: I'm guessing this is not compiled.

pirapira: Oh, me too.

test/tools/libtesteth/ImportTest.cpp
pirapira: Didn't know about that `cerr` / `cout` difference.

pirapira: Fixed this and something similar in the same file.

test/unittests/libdevcrypto/MemTrie.cpp
pirapira: Fixed the spelling.

pirapira: Ah, but this one is on cerr too.

pirapira: Removed the flush altogether.

chfast:
I meant to remove all "flushes". I think the only case when explicit flush is needed **on std::cout** is when we need to use `std::system`.

pirapira:
OK, removed `flush` except one.  The one is a spinner.  I want to see the `.` every ten seconds.

PR #4358, opened 2017-08-18
hash logs in tests to reduce it's filesize

winsvega:
log field is optional

Tests
https://github.com/ethereum/tests/pull/263


pirapira:
This doesn't solve the problem with many log entries.

pirapira:
I think `testeth` needs to read the hash when it's not filling.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -645,9 +645,18 @@ void ImportTest::checkGeneralTestSectionSearch(json_spirit::mObject const& _expe
...
-				BOOST_CHECK_MESSAGE(_expects.at("hash").get_str() == toHexPrefixed(tr.postState.rootHash().asBytes()), TestOutputHelper::testName() + " Expected another postState hash! expected: " + _expects.at("hash").get_str() + " actual: " + toHex(tr.postState.rootHash().asBytes()) + " in " + trInfo);
+			{
+				//checking filled state test against client
+				BOOST_CHECK_MESSAGE(_expects.at("hash").get_str() == toHexPrefixed(tr.postState.rootHash().asBytes()),
+									TestOutputHelper::testName() + " on " + test::netIdToString(tr.netId) + ": Expected another postState hash! expected: " + _expects.at("hash").get_str() + " actual: " + toHex(tr.postState.rootHash().asBytes()) + " in " + trInfo);
+				if (_expects.count("logs"))
+					BOOST_CHECK_MESSAGE(_expects.at("logs").get_str() == exportLog(tr.output.second.log()),
+									TestOutputHelper::testName() + " on " + test::netIdToString(tr.netId) + " Transaction log mismatch! expected: " + _expects.at("logs").get_str() + " actual: " + exportLog(tr.output.second.log()) + " in " + trInfo);
+				else
+					BOOST_ERROR(TestOutputHelper::testName() + " Expect section or postState missing logs field!");
```
pirapira: This case is hit only when the expect section is missing.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -645,9 +645,18 @@ void ImportTest::checkGeneralTestSectionSearch(json_spirit::mObject const& _expe
...
 				}
-			}			
+			}
 			else if (_expects.count("hash"))
-				BOOST_CHECK_MESSAGE(_expects.at("hash").get_str() == toHexPrefixed(tr.postState.rootHash().asBytes()), TestOutputHelper::testName() + " Expected another postState hash! expected: " + _expects.at("hash").get_str() + " actual: " + toHex(tr.postState.rootHash().asBytes()) + " in " + trInfo);
+			{
+				//checking filled state test against client
+				BOOST_CHECK_MESSAGE(_expects.at("hash").get_str() == toHexPrefixed(tr.postState.rootHash().asBytes()),
+									TestOutputHelper::testName() + " on " + test::netIdToString(tr.netId) + ": Expected another postState hash! expected: " + _expects.at("hash").get_str() + " actual: " + toHex(tr.postState.rootHash().asBytes()) + " in " + trInfo);
+				if (_expects.count("logs"))
```
pirapira: Why is this checked only when `"hash"` field exists?

fjl:
I meant that it'd be OK to include a single hash of all logs. With that, the size would not depend on the number of logs.

pirapira:
@fjl yep, now we've figured out how to do that.

pirapira:
@winsvega please rebase.

test/tools/libtesteth/ImportTest.cpp
winsvega: if there is a hash then it is a post state of a filled test with logs
otherwise it is an expect section of a testFiller

pirapira: I see, there should be an error message when there is `"logs"` but there is no `"hash"`.  Otherwise, the `"logs"` section in fillers will be silently ignored.

winsvega: well, actually we do not really care about the fillers anymore. there could be any additional fields now. after you split input and output, the output json will be filled by strict json format without trash fields

pirapira: OK.  I think extra fields in the filler should cause an error (to detect typos, to detect something the filler author wanted but coudn't), but that's something separate from this PR.

PR #4359, opened 2017-08-18
separate input and output of fillBCTest() function

pirapira:
This is the first step for separating input and output of doBlockchainTest() function.

This PR changes `fillBCTest()` function so that it takes an input and returns an output.  Before this PR, the function modified a filler into a completed test in place.


test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -244,46 +244,65 @@ json_spirit::mValue doBlockchainTestNoLog(json_spirit::mValue const& _input, boo
 	return v;
 }
 
-void fillBCTest(json_spirit::mObject& _o)
+json_spirit::mObject fillBCTest(json_spirit::mObject const& _input)
 {
+	json_spirit::mObject output;
```
winsvega: could you name it outputJson?

pirapira:
I'll try and then merge.

test/tools/jsontests/BlockChainTests.cpp
pirapira: yes.

pirapira:
#4366 should fix "the missing 0x problem".  This one should be merged after that.

pirapira:
or not.

PR #4364, opened 2017-08-21
EVM-C: Renames

chfast:


libevm/JitVM.cpp
```diff
@@ -398,25 +397,25 @@ owning_bytes_ref JitVM::exec(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onO
...
 		return EVM_CONSTANTINOPLE;
 	if (_schedule.haveRevert)
 		return EVM_BYZANTIUM;
 	if (_schedule.eip158Mode)
-		return EVM_CLEARING;
+		return EVM_SPURIOUS_DRAGON;
 	if (_schedule.eip150Mode)
-		return EVM_ANTI_DOS;
+		return EVM_TANGERINE_WHISTLE;
 	return _schedule.haveDelegateCall ? EVM_HOMESTEAD : EVM_FRONTIER;
```
axic: Would it be easier to read making this thing an if-elseif-else waterfall?

libevm/JitVM.cpp
```diff
@@ -398,25 +397,25 @@ owning_bytes_ref JitVM::exec(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onO
 	return output;
 }
 
-evm_mode JitVM::scheduleToMode(EVMSchedule const& _schedule)
+evm_revision JitVM::scheduleToMode(EVMSchedule const& _schedule)
```
axic: Also renaming to `scheduleToRevision`?

libevm/JitVM.cpp
chfast: "No else after return". But I can replace the last `? :`.

libevm/JitVM.cpp
chfast: Yep.

libevm/JitVM.cpp
axic: I do not necessarily agree with the "no else after return" rule, in many cases it makes it more readable.

At least replace the ternary, should help readability.

chfast:
Now it compiles. What do you think @axic?

PR #4370, opened 2017-08-22
Copy "network" field in fillBCTest()

pirapira:
Before this commit
```
test/testeth -t 'TransitionTests/bcEIP158ToByzantium' -- --filltests --singletest  ByzantiumTransition
```
yielded an error
```
BlockChainTests.cpp(127): fatal error: in "TransitionTests/bcEIP158ToByzantium": critical check o.count("network") has failed
```

After this commit, the same command succeeds.

### DIFF
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -264,9 +264,6 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input)
 	string chainnetwork = "default";
 	std::map<string, ChainBranch*> chainMap = { {chainname , new ChainBranch(genesisBlock)}};
 
-	if (_input.count("noBlockChainHistory") > 0)
-		output["noBlockChainHistory"] = _input.at("noBlockChainHistory");
-
 	if (_input.count("network") > 0)
 		output["network"] = _input.at("network");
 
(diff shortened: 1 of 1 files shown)


pirapira:
@jwasinger here is another type of commands you can try during Travis runs.

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -267,6 +267,9 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input)
 	if (_input.count("noBlockChainHistory") > 0)
```
winsvega: this field is not needed in filled test. this is an instruction for test filling

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -267,6 +267,9 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input)
...
 		output["noBlockChainHistory"] = _input.at("noBlockChainHistory");
 
+	if (_input.count("network") > 0)
```
winsvega: input must require network field to be set

test/tools/jsontests/BlockChainTests.cpp
pirapira: I didn't know that.

pirapira: Removed.

test/tools/jsontests/BlockChainTests.cpp
pirapira: I tried.  Please have a look.

pirapira:
@winsvega it is wrong to require "network" field.

For example `tests/src/BlockchainTestsFiller/bcRandomBlockhashTest/randomStatetest392BCFiller.json` does not contain the "network" field.

pirapira:
The Travis failure is a timeout on macos.

PR #4372, opened 2017-08-22
Remove two macros that were never used

pirapira:
@winsvega first started to wonder if they are ever used.


pirapira:
AppVeyor timed out.

PR #4375, opened 2017-08-22
Display more informative messages by not rethrowing an exception

pirapira:
This PR makes debugging easier when the fillers are malformed.

**Before this PR:**

```
75%...
100%
/home/yh/src/cpp-ethereum/test/tools/libtesteth/TestHelper.cpp(525): error: in "VMTests/vmtests": suicide Failed test with Exception: value type is 2 not 1
Test Case "vmArithmeticTest":
24%...
```
TestHelper.cpp(535) is not informative because it's a line that catches all errors.

**After this PR:**
```
50%...
75%...
100%
unknown location(0): fatal error: in "VMTests/vmtests": std::runtime_error: value type is 2 not 1
/home/yh/src/cpp-ethereum/test/tools/jsontests/vm.cpp(441): last checkpoint
Test Case "vmArithmeticTest":
24%...
```
vm.cpp(441), aha, it's a problem in "logs" field!

Both outputs are taken from
```
test/testeth -t 'VMTests' -- --filltests
```

### DIFF
--- a/test/tools/jsontests/StateTests.cpp
+++ b/test/tools/jsontests/StateTests.cpp
@@ -41,6 +41,8 @@ namespace dev {  namespace test {
 
 json_spirit::mValue doStateTests(json_spirit::mValue const& _input, bool _fillin)
 {
+	BOOST_REQUIRE_MESSAGE(_input.type() == obj_type,
+		TestOutputHelper::testFileName() + " A GeneralStateTest file should contain an object.");
 	BOOST_REQUIRE_MESSAGE(!_fillin || _input.get_obj().size() == 1,
 		TestOutputHelper::testFileName() + " A GeneralStateTest filler should contain only one test.");
 	json_spirit::mValue v = json_spirit::mObject();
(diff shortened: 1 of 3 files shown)


pirapira:
The new output does not contain the name of the test case, which is bad.

winsvega:
you could use TestOutputHelper to access the testcase name from different scopes.

pirapira:
@winsvega where shall I apply your suggestion?  On top of this PR or on `develop`?  Where am I supposed to access TestOutputHelper?

winsvega:
you said you wanted more precise error information that shows the line of the error exception.
but you said thus if you remove exception it wont show you the test name. 
so if you want to display a testname out of testsuite scope. use TestOutputHelper which stores the test name.

pirapira:
No, since I stopped catching the exception, I don't have an "out of testsuite" scope.

winsvega:
what is this comment about? 
`The new output does not contain the name of the test case, which is bad.`

pirapira:
This line
```
unknown location(0): fatal error: in "VMTests/vmtests": std::runtime_error: value type is 2 not 1
/home/yh/src/cpp-ethereum/test/tools/jsontests/vm.cpp(441): last checkpoint
```
does not contain the name of the test case.  That's what I meant.

I don't have control over this message.  I cannot use TestOutputHelper to improve this message, I think.

winsvega:
well you could change boost require to bost_check_message  on those lines.

pirapira:
Good idea.  Maybe before `v.get_obj()`, I can insert a boost_check_message.

PR #4378, opened 2017-08-23
Remove TestHelper.h includes from unit tests

pirapira:
Many unit tests included TestHelper.h, but most of them just needed TestOutputHelper.h,
which is included from TestHelper.h.  So this commit replaces many TestHelper.in includes
with TestOutputHelper.h includes.

Sometimes, additional includes were necessary, but all compilation units got simpler.

### DIFF
--- a/test/unittests/libdevcore/rlp.cpp
+++ b/test/unittests/libdevcore/rlp.cpp
@@ -20,17 +20,19 @@
  * RLP test functions.
  */
 
-#include <fstream>
-#include <sstream>
-
 #include <libdevcore/Log.h>
 #include <libdevcore/RLP.h>
 #include <libdevcore/Common.h>
 #include <libdevcore/CommonIO.h>
-#include <algorithm>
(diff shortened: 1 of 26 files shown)


test/unittests/libdevcore/rlp.cpp
```diff
@@ -20,17 +20,19 @@
...
-#include <sstream>
-
 #include <libdevcore/Log.h>
 #include <libdevcore/RLP.h>
 #include <libdevcore/Common.h>
 #include <libdevcore/CommonIO.h>
-#include <algorithm>
 #include <json_spirit/JsonSpiritHeaders.h>
+#include <test/tools/libtesteth/TestOutputHelper.h>
 #include <test/tools/libtesteth/TestHelper.h>
```
winsvega: does this line needed?

pirapira: Yes, if you remove this, the build fails.

PR #4380, opened 2017-08-23
Remove ICAP and ABI

chfast:


pirapira:
I'm seeing some Hunter related error in Appveyor.

chfast:
I think this is GitHub connection issue...

PR #4382, opened 2017-08-23
Further refactoring on doBlockchainTest

pirapira:
This PR further separates input and output in blockchaintest filling.  The code became a bit shorter.

### DIFF
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -146,15 +146,16 @@ json_spirit::mValue doTransitionTest(json_spirit::mValue const& _input, bool _fi
 
 json_spirit::mValue doBlockchainTestNoLog(json_spirit::mValue const& _input, bool _fillin)
 {
-	json_spirit::mValue v = _input; // TODO: avoid copying and add only valid fields to the new object.
+	json_spirit::mValue v = json_spirit::mObject();
 	map<string, json_spirit::mObject> tests;
-	vector<decltype(v.get_obj().begin())> erase_list;
+	vector<string> erase_list;
 
 	// range-for is not used because iterators are necessary for removing elements later.
-	for (auto i = v.get_obj().begin(); i != v.get_obj().end(); i++)
+	for (auto i = _input.get_obj().begin(); i != _input.get_obj().end(); i++)
 	{
 		string testname = i->first;
-		json_spirit::mObject& o = i->second.get_obj();
+		v.get_obj()[testname] = i->second.get_obj();
+		json_spirit::mObject& o = v.get_obj()[testname].get_obj();
 
 		//Select test by name if --singletest is set and not filling state tests as blockchain
 		if (!Options::get().fillchain && !TestOutputHelper::passTest(testname))
@@ -214,7 +215,7 @@ json_spirit::mValue doBlockchainTestNoLog(json_spirit::mValue const& _input, boo
 
 			// will be deleted once after the loop.
 			// removing an element while in this loop causes memory corruption.
-			erase_list.push_back(i);
+			erase_list.push_back(testname);
 		}
 		else
 		{
@@ -230,8 +231,8 @@ json_spirit::mValue doBlockchainTestNoLog(json_spirit::mValue const& _input, boo
 	}
 
 	//Delete source test from the json
-	for (auto i: erase_list)
-		v.get_obj().erase(i);
+	for (auto testname: erase_list)
+		v.get_obj().erase(v.get_obj().find(testname));
 
 	//Add generated tests to the result file
 	if (_fillin)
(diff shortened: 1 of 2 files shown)


test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -183,64 +179,48 @@ json_spirit::mValue doBlockchainTestNoLog(json_spirit::mValue const& _input, boo
 				dev::test::TestBlockChain::s_sealEngineNetwork = network;
 				string newtestname = testname + "_" + test::netIdToString(network);
 
-				json_spirit::mObject jObj = o;
-				if (o.count("expect"))
+				json_spirit::mObject jObjOutput = inputTest;
+				if (inputTest.count("expect"))
 				{
 					//prepare the corresponding expect section for the test
```
winsvega: need to add a comment here that from one test with general expect section we write  'forknumber' tests with individual expect section to the output test which will again be used as a filler and overwriten.

pirapira: What is a "forknumber" test?

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -146,28 +146,24 @@ json_spirit::mValue doTransitionTest(json_spirit::mValue const& _input, bool _fi
 
 json_spirit::mValue doBlockchainTestNoLog(json_spirit::mValue const& _input, bool _fillin)
 {
-	json_spirit::mValue v = _input; // TODO: avoid copying and add only valid fields to the new object.
-	map<string, json_spirit::mObject> tests;
-	vector<decltype(v.get_obj().begin())> erase_list;
+	json_spirit::mValue v = json_spirit::mObject();
```
winsvega: you could convert mValue to mObject

pirapira: No, I'm converting an `mObject` to `mValue`.  This function returns an `mValue`.

winsvega: so when return you could do (mValue) obj;

test/tools/jsontests/BlockChainTests.cpp
winsvega: from one test in a filler we produce  a test for Frontier, Homestead,  and so on. total number 'forknumber'

test/tools/jsontests/BlockChainTests.cpp
pirapira: OK, that's a bit cleaner.

test/tools/jsontests/BlockChainTests.cpp
pirapira: I see, when there are seven forks, "forknumber" is equial to seven, so "forknumber" tests mean 7 tests.

pirapira:
@winsvega do you want it as a unit test or a `testeth` command invocation?

winsvega:
a unit test would be nice. we could run doTest function for a particular test.  
however need to disable mining  use sealEngine NoProof  so it to run faster

pirapira:
OK, will do.

pirapira:
@winsvega I added such unit tests.

test/unittests/libtesteth/blockchainTest.cpp
```diff
@@ -0,0 +1,28 @@
...
+/** @file blockchainTest.cpp
+ * Unit tests for blockchain test filling/execution.
+ */
+
+#include <boost/test/unit_test.hpp>
+#include <test/tools/libtesteth/TestOutputHelper.h>
+
+using namespace dev::test;
+
+BOOST_FIXTURE_TEST_SUITE(BlockChainTestSuite, TestOutputHelper)
```
winsvega: maybe add a certain indicator. that this suite is coverage for testeth
like CoverBlockchainTest

test/unittests/libtesteth/blockchainTest.cpp
```diff
@@ -18,11 +18,133 @@
...
+			}
+		}
+	)";
+	json_spirit::mValue input;
+	json_spirit::read_string(s, input);
+	json_spirit::mValue output = doBlockchainTestNoLog(input, true);
+	BOOST_CHECK_MESSAGE(output.get_obj().size() == getNetworks().size(), "A wrong number of tests were generated.");
+}
+
+BOOST_AUTO_TEST_CASE_EXPECTED_FAILURES(fillingWithWrongExpectation, 2)
```
winsvega: what about BOOST_CHEK_THROW ?

winsvega:
very nice!!

test/unittests/libtesteth/blockchainTest.cpp
```diff
@@ -18,11 +18,133 @@
...
 
 using namespace dev::test;
 
 BOOST_FIXTURE_TEST_SUITE(BlockChainTestSuite, TestOutputHelper)
 
+BOOST_AUTO_TEST_CASE(fillingExpectationOnMultipleNetworks)
+{
+	std::string const s = R"(
+		{
+			"BLOCKHASH_Bounds" : {
```
winsvega: I think we could define this template once and then use a replace function from fuzzTest.cpp
you could define your own map for replacement. even something like 
[EXPECT] -> "{ "network" : "Frontier", ...  }"

test/unittests/libtesteth/blockchainTest.cpp
pirapira: It didn't work.  `doBlockchainTest()` already registered failures even inside `BOOST_CHECK_THROW`.

test/unittests/libtesteth/blockchainTest.cpp
pirapira: I don't like these templates.  I'll visit there some day.

PR #4383, opened 2017-08-23
[WIP] generalize test execution logic

winsvega:
this is a first step to replace common logic of executing test files from different suites and folders. 
eventually this should remove original test::executeTests function and replace it with refactored version introduced in this PR. 

@pirapira  I could this step by step by merging small PRs. so it would be easy to review

### DIFF
--- a/test/tools/jsontests/StateTests.cpp
+++ b/test/tools/jsontests/StateTests.cpp
@@ -118,28 +118,7 @@ public:
 		string casename = boost::unit_test::framework::current_test_case().p_name;
 		if (casename == "stQuadraticComplexityTest" && !test::Options::get().quadratic)
 			return;
-		fillAllFilesInFolder(casename);
-	}
-
-	void fillAllFilesInFolder(string const& _folder)
-	{
-		std::string fillersPath = test::getTestPath() + "/src/GeneralStateTestsFiller/" + _folder;
-
-		string filter = test::Options::get().singleTestName.empty() ? string() : test::Options::get().singleTestName + "Filler";
-		std::vector<boost::filesystem::path> files = test::getJsonFiles(fillersPath, filter);
-		int fileCount = files.size();
-
-		if (test::Options::get().filltests)
-			fileCount *= 2; //tests are checked when filled and after they been filled
-		test::TestOutputHelper::initTest(fileCount);
-
-		for (auto const& file: files)
-		{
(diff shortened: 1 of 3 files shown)


test/tools/libtesteth/TestHelper.cpp
```diff
@@ -460,6 +460,34 @@ void userDefinedTest(std::function<json_spirit::mValue(json_spirit::mValue const
 	}
 }
 
+void executeTests2(test::testType _type, const std::string& _testFolder)
```
winsvega: execute test files of a given type in a given folder. 
we have boost_test_case(stInitCode)  for example where stInitCode is a folder name with tests for that test suite.

test/tools/jsontests/StateTests.cpp
```diff
@@ -118,28 +118,7 @@ class generaltestfixture
...
-			fileCount *= 2; //tests are checked when filled and after they been filled
-		test::TestOutputHelper::initTest(fileCount);
-
-		for (auto const& file: files)
-		{
-			test::TestOutputHelper::setCurrentTestFileName(file.filename().string());
-			test::executeTests(file.filename().string(), "/GeneralStateTests/"+_folder, "/GeneralStateTestsFiller/"+_folder, dev::test::doStateTests);
-		}
-
-		test::TestOutputHelper::finishTest();
```
pirapira: What about creating a class that
* calls `initTest()` in the constructor and
* `finishTest()` in the destructor?

winsvega: so the problem now is that we have 2 different logic. 
1) the test cases = number of files. 
2) the test cases = number of tests in the file
3) the test cases = number of tests in all files

winsvega: what I want to do is to make a single place in the code which looks up for tests files in the given folder. so to check if there are tests without fillers

winsvega: I guess to resolve this we need to move to a format. where a single .json file contains a single test. 
currently VMTests and Transaction tests have many tests in one file.

winsvega: then we could use  number of tests = number of files in the folder.

pirapira: @winsvega Let's discuss the format first.

pirapira:
I talked with @gumb0 before separating input and output.  I want to hear his opinion about this, too.

PR #4384, opened 2017-08-23
EIP 684: contract creation on nonempty code or nonzero nonce fails.

pirapira:
Note this change is applied for all blocks from block number zero.


winsvega:
So once and for all. address collision result in OOG, on all forks including future updates. 

For transactions that will run parallel FCFS rules are applied

pirapira:
FCFS?

pirapira:
I see.  Well, transactions have nonces.

winsvega:
there was an EIP about different transaction proccessing

pirapira:
Rebased.

cdetrio:
EIP 684: https://github.com/ethereum/EIPs/issues/684

PR #4402, opened 2017-08-25
Uncomment a unit test file whose tests are completely coommented out

pirapira:


winsvega:
those tests are commented out because it was failing on travis. or sometimes failing. 
the goal of this tests is to cover code lines with those exceptions being thrown out of blockchain

winsvega:
so maybe we could re enable it and see how it is working at the moment. cause its been a year since this test commented

pirapira:
@winsvega I'll try enabling it.

gumb0:
@winsvega In general, this should be done not by commenting out the test, but by disabling using boost.test methods (probably something like `BOOST_AUTO_TEST_SUITE(BlockQueueSuite, *unit_test::disabled())`
Otherwise test code easily becomes uncompilable and useless.

pirapira:
I uncommented the test and it compiles and works.

Issue #4401, opened 2017-08-25
 test/testeth -t StateTestsGeneral -- --filltests --fillchain fails on stExample


pirapira:
On cpp-ethereum ea4710de5846070aca6db3f74b0adabc0ba15b2c and tests 35777d7779302fb028becd2794004c03743ae297,

```
test/testeth -t StateTestsGeneral -- --filltests --fillchain &> fillchain.out
```
prints
```
error: in "StateTestsGeneral/stExample": add11_d0g0v0_Byzantium Check State: 2adc25665018aa1fe0e6bc666dac8fc2697ff9ba: incorrect balance 3000000000000041012, expected 5000000000000041012
add11_d0g0v0_Byzantium
```

pirapira:
`test/testeth -t StateTestsGeneral/stExample -- --filltests --fillchain` reproduces the issue in a smaller amount of time.

pirapira:
The problem still exists in cpp 8d20c1763a5be70b662f210529843d103d523479 and tests 1d30b4795664f64b1b157971754e14a10cfd9115

pirapira:
I found that
```
src/GenStateTestAsBcTemp/stExample/add11_d0g0v0Filler.json
```
contains the wrong balance.

pirapira:
Moreover, it contains many expectations for `network "all"`.
```
                "network" : [
                    "ALL"
                ],
                "result" : {
                    "0x095e7baea6a6c7c4c2dfeb977efac326af552d87" : {
                        "balance" : "0x0de0b6b3a76586a0",
                        "code" : "0x6001600101600055",
                        "storage" : {
                            "0x00" : "0x02"
                        }
                    },
                    "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba" : {
                        "balance" : "0x4563918244f4a034",
                        "code" : "",
                        "nonce" : "0x00",
                        "storage" : {
                        }
                    },
                    "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
                        "balance" : "0x0de0b6b3a761d92c",
                        "code" : "",
                        "nonce" : "0x01",
                        "storage" : {
                        }
                    }
                }
            },
            {
                "network" : [
                    "ALL"
                ],
                "result" : {
                    "0x095e7baea6a6c7c4c2dfeb977efac326af552d87" : {
                        "balance" : "0x0de0b6b3a76586a0",
                        "code" : "0x6001600101600055",
[...]

PR #4403, opened 2017-08-25
Fix the network id in --fillchain

pirapira:
Fixes https://github.com/ethereum/cpp-ethereum/issues/4401

The problem happened when the expectation section in the filler contained

```
network : ["ALL"]
```

In that case, many expectations of network = "ALL" was created, instead of
many expectations of different networks ("Frontier", "Homestead", ...).

### DIFF
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -117,13 +117,9 @@ bytes ImportTest::executeTest()
 				vector<size_t> stateIndexesToPrint; //not used
 				json_spirit::mArray expetSectionArray;
 
-				//gather the expect sections for such a transaction on all networks (if defined)
-				std::vector<string> checkedNetworks;
 				for (auto const& net : networks)
 				{
 					tr.netId = net;
-					if (std::find(checkedNetworks.begin(), checkedNetworks.end(), test::netIdToString(net)) != checkedNetworks.end())
-						continue;
 
 					// Calculate the block reward
 					ChainParams const chainParams{genesisInfo(net)};
@@ -154,10 +150,6 @@ bytes ImportTest::executeTest()
 
 							json_spirit::mObject expetSectionObj;
 							expetSectionObj["network"] = test::netIdToString(net);
-							std::vector<string> networks;
-							ImportTest::parseJsonStrValueIntoVector(exp.get_obj().at("network"), networks);
-							for(auto const& netname : networks)
-								checkedNetworks.push_back(netname);
 							expetSectionObj["result"] = obj;
 							expetSectionArray.push_back(expetSectionObj);
 							break;
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -153,7 +153,7 @@ bytes ImportTest::executeTest()
 							}
 
 							json_spirit::mObject expetSectionObj;
-							expetSectionObj["network"] = exp.get_obj().at("network");
+							expetSectionObj["network"] = test::netIdToString(net);
 							std::vector<string> networks;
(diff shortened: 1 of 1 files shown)


test/tools/libtesteth/ImportTest.cpp
```diff
@@ -153,7 +153,7 @@ bytes ImportTest::executeTest()
 							}
 
 							json_spirit::mObject expetSectionObj;
-							expetSectionObj["network"] = exp.get_obj().at("network");
```
winsvega: what was the issue again?  hive tests were working before that.

winsvega:
have you checked how other expect sections translated now? 
like 
{ 
["Homestead", "EIP158"]
["Frontier", "EIP150"] 
}

test/tools/libtesteth/ImportTest.cpp
winsvega: this is a crucial change. we have to check that everything is correct

winsvega: I mean double check

pirapira: The issue was this
```
                "network" : [
                    "ALL"
                ],
                "result" : {
                    "0x095e7baea6a6c7c4c2dfeb977efac326af552d87" : {
                        "balance" : "0x0de0b6b3a76586a0",
                        "code" : "0x6001600101600055",
                        "storage" : {
                            "0x00" : "0x02"
                        }
                    },
                    "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba" : {
                        "balance" : "0x4563918244f4a034",
                        "code" : "",
                        "nonce" : "0x00",
                        "storage" : {
                        }
                    },
                    "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
                        "balance" : "0x0de0b6b3a761d92c",
                        "code" : "",
                        "nonce" : "0x01",
                        "storage" : {
                        }
                    }
                }
            },
            {
                "network" : [
                    "ALL"
                ],
                "result" : {
                    "0x095e7baea6a6c7c4c2dfeb977efac326af552d87" : {
                        "balance" : "0x0de0b6b3a76586a0",
                        "code" : "0x6001600101600055",
                        "storage" : {
                            "0x00" : "0x02"
                        }
                    },
                    "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba" : {
                        "balance" : "0x4563918244f4a034",
                        "code" : "",
                        "nonce" : "0x00",
                        "storage" : {
                        }
                    },
                    "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b" : {
                        "balance" : "0x0de0b6b3a761d92c",
                        "code" : "",
                        "nonce" : "0x01",
                        "storage" : {
                        }
                    }
                }
            },
```

pirapira: ^ you see `"network" : [ "ALL" ] ` is repeated?

pirapira:
@winsvega I'm trying --filltests --fillchain on this branch.  `Test Case "stCallCodes":` completely succeeds.

`stCallCreateCallCodeTest` `stExample` `stInitCodeTest` all succeed.

pirapira:
@winsvega this was not an issue because the block reward was the same on all networks.  `network : [All]` was repeated five times but the first expectation was used and it was fine.  If the results change depending on networks, `network : [All]` didn't appear in the GeneralStateTest filler anyway.

Now, even when the GeneralStateTest results are the same for all networks, and `network : [All]` appears in the GeneralStateTest filler, the blockchaintest might result in different balances.  So this bug became visible.

winsvega:
right. also not that changing network might result in no errors because errors are checked with expect sections. if the expect section will be missed or wrong you might miss the initial errors. ok let me check.

pirapira:
Updated the code to fix the problem where some networks could be skipped.

pirapira:
AppVeyor timed out.

PR #4408, opened 2017-08-25
Fix the requirement-actual order in RequirementError macro

pirapira:
In RequirementError() macros, the first argument should be the requirement and the second should be the actual value.

The ordering is evident in
`libdevcore/Exceptions.h`
where
```
using RequirementError = boost::tuple<errinfo_required, errinfo_got>;
```
can be seen.


winsvega:
ask Pawel btw

PR #4423, opened 2017-08-29
TestOutputHelper as RAII

pirapira:
TestOutputHelper class has a constructor that calls `initTest()` and a destructor that calls `finishTest()`.  This PR, instead of calling `initTest()` and `finishTest()` manually, creates a `TestOutputHelper` object and let its constructor and destructor do the job.

I tried to keep the original behavior as much as possible, but there were some `initTest()` without `finishTest()`. These missing `finishTest()` calls had to be added.

### DIFF
--- a/test/tools/jsontests/vm.cpp
+++ b/test/tools/jsontests/vm.cpp
@@ -296,8 +296,9 @@ namespace dev { namespace test {
 
 json_spirit::mValue doVMTests(json_spirit::mValue const& _input, bool _fillin)
 {
-	if (string(boost::unit_test::framework::current_test_case().p_name) != "vmRandom")
-		TestOutputHelper::initTest(_input.get_obj().size());
+	unique_ptr<TestOutputHelper> testOutputHelper;
+	if (string(boost::unit_test::framework::current_test_case().p_name) != "vmRandom") // TODO: remove this hard-coding
+		testOutputHelper.reset(new TestOutputHelper(_input.get_obj().size()));
 
 	json_spirit::mValue v = json_spirit::mObject();
 	json_spirit::mObject& output = v.get_obj();
@@ -472,7 +473,6 @@ json_spirit::mValue doVMTests(json_spirit::mValue const& _input, bool _fillin)
 		}
 	}
 
-	TestOutputHelper::finishTest();
 	return v;
 }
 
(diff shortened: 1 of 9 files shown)


winsvega:
I really would prefere to see TestOutputHelper as a static object rather then const& on interfaces

pirapira:
@winsvega the `TestOutputHelper const*` arguments will disappear when we decide whether or not to call `initTests()` before `doTests()` .  In the current codebase, `initTests()` might and might not be called before `doTests()`, so `doTests()` needs to be informed about that.

pirapira:
@winsvega  now Travis detects when we forget to call `initTest()`:
```
TestHelper.cpp(406): fatal error: in "BlockchainTests/bcForgedTest": executeTests() expects a non-null pointer to a TestOutputHelper object.
```

This is great, isn't it?  This PR makes such error detection possible.

test/tools/jsontests/vm.cpp
```diff
@@ -296,8 +296,9 @@ namespace dev { namespace test {
 
 json_spirit::mValue doVMTests(json_spirit::mValue const& _input, bool _fillin)
 {
-	if (string(boost::unit_test::framework::current_test_case().p_name) != "vmRandom")
-		TestOutputHelper::initTest(_input.get_obj().size());
+	unique_ptr<TestOutputHelper> testOutputHelper;
+	if (string(boost::unit_test::framework::current_test_case().p_name) != "vmRandom") // TODO: remove this hard-coding
+		testOutputHelper.reset(new TestOutputHelper(_input.get_obj().size()));
```
pirapira: All those `TestOutputHelper const*` arguments appeared when I tried to remove this hard-coded "vmRandom", but I now leave this.

gumb0: Maybe to get rid of it we could make `doVMTests` and separate `doVMTestsWithOutput` that creates `TestOutputHelper` then calls `doVMTests`

Then call internal `doVMTests` only from `vmRandom`

pirapira: I see.  I'll name the new function `doVMTestsNoLog()` following the similar function `doBlockchainTestsNoLog()`.

gumb0:
Can we make `TestOutputHelper::initTest()` and `TestOutputHelper::finishTest()` private now ?

winsvega:
things are getting over complicated now. 
I am working on a refactoring of the tests execution. so before testoutputhelper was used to indicate the test that was run from a file where in one file there were like dozens of tests. now when all tests are in it's own file we could remove that logic. and thus test output helper init and finish will need to be called before the execution of a test suite and after. (basically in a test fixture class)

pirapira:
@winsvega let's talk this over.

pirapira:
@gumb0 yes, I was able to make them private.

PR #4430, opened 2017-08-31
Remove ChainOperationParams::u256Param that looks up parameters by strings

pirapira:
Instead, introduced a number of u256 fields in ChainOperationParams.

I found one place that still looked up "MetropolisForkBlock".  I replaced it with `.constantinopleForkBlock`.

PR #4434, opened 2017-08-31
Fix typo homstead

pirapira:
There were many occurrences of "homstead", but it should be "homestead"


pirapira:
I have #4430 after this one.

PR #4436, opened 2017-08-31
BlockChain unit tests improvements

gumb0:
1. New test for new method `BlockChain::insertWithoutParent()`

2. Make BlockChain unit tests use chain parameters with `NoProof` seal engine, which makes them faster and more isolated by skipping full mining


gumb0:
@winsvega Yes, I think it should be reasonably fast

chfast:
Holy f*ck, how fast is it now? Thanks for finding this...

PR #4437, opened 2017-08-31
Cleanups again

chfast:


test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -390,7 +390,7 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input)
 				importedBlocks.back().clearState(); //close the state as it wont be needed. too many open states would lead to exception.
 			}
 		}
-		catch (Exception const& _e)
+		catch (dev::Exception const& _e)
```
gumb0: Why do you do this?

chfast: I want to speed up the build time by using "unity builds" - the whole library is built as a single compilation unity. Because of the extensive use of `using namespace` I get name conflicts. I don't remember now where is the other `Exception`.

chfast: Hm.. maybe it was `CryptoPP::Exception`. I will revert this then.

gumb0: There's also `Json::Exception` in jsoncpp

PR #4438, opened 2017-08-31
split vm tests

winsvega:
tests https://github.com/ethereum/tests/pull/296

### DIFF
--- a/test/tools/jsontests/vm.cpp
+++ b/test/tools/jsontests/vm.cpp
@@ -296,12 +296,6 @@ namespace dev { namespace test {
 
 json_spirit::mValue doVMTests(json_spirit::mValue const& _input, bool _fillin)
 {
-	TestOutputHelper testOutputHelper(_input.get_obj().size());
-	return doVMTestsNoLog(_input, _fillin);
-}
-
-json_spirit::mValue doVMTestsNoLog(json_spirit::mValue const& _input, bool _fillin)
-{
 	json_spirit::mValue v = json_spirit::mObject();
 	json_spirit::mObject& output = v.get_obj();
 	for (auto& i: _input.get_obj())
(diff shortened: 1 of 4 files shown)


chfast:
Can you explain what this is about?

test/tools/jsontests/vm.cpp
```diff
@@ -480,114 +474,55 @@ json_spirit::mValue doVMTestsNoLog(json_spirit::mValue const& _input, bool _fill
...
-	dev::test::executeTests("vmBitwiseLogicOperationTest", "/VMTests", "/VMTestsFiller", dev::test::doVMTests);
-}
-
-BOOST_AUTO_TEST_CASE(vmSha3Test)
-{
-	dev::test::executeTests("vmSha3Test", "/VMTests", "/VMTestsFiller", dev::test::doVMTests);
-}
 
-BOOST_AUTO_TEST_CASE(vmEnvironmentalInfoTest)
+class vmtestfixture
```
chfast: I have no idea, why out of nowhere you named this class this way.

test/tools/jsontests/vm.cpp
```diff
@@ -480,114 +474,55 @@ json_spirit::mValue doVMTestsNoLog(json_spirit::mValue const& _input, bool _fill
...
 		{
-			BOOST_ERROR(" Failed test with Exception: " << diagnostic_information(_e));
+			cnote << "Skipping " << casename << " because --all option is not specified.\n";
+			return;
 		}
-		catch (std::exception const& _e)
+		fillAllFilesInFolder(casename);
+	}
+
+	void fillAllFilesInFolder(string const& _folder)
```
pirapira: Maybe you can make this `private`.

test/tools/jsontests/vm.cpp
winsvega: following the same scheme as for the other tests.

test/tools/jsontests/vm.cpp
winsvega: I am going to remove this function in the next PR

test/tools/jsontests/vm.cpp
chfast: Like `ConstantinopleTransitionTestFixture` or `TestOutputFixture`?

winsvega: like transactiontestfixture, generaltestfixture

pirapira: Let's do CamelCase in class names.

winsvega:
this PR is about refactoring the testeth. it moves towards creation of a function
runAllTestsInFolder that will unify the tests execution for all suites

winsvega:
https://github.com/ethereum/cpp-ethereum/pull/4383

chfast:
I don't think  a single test case "run all tests in a folder" is a good idea if you register the test names first (to be available to Boost.Test filtering).

winsvega:
the test cases will be as it is right now

pirapira:
I want to understand what @winsvega is trying to do.  I might like it, I might not like it, but it's better to know before an actual implementation happens.

Will `runAllTestsInFolder()` be used in within a test case or out of a test case?

Will `runAllTestsInFolder()` invoke multiple test cases?

Will `runAllTestsInFolder()` called from test cases, from test fixtures, or from somewhere else?

winsvega:
@pirapira we discussed this on skype. 
runAllTestsInFolder will run from fixtures to basically generalize the logic that we have in all suites. 
and to make check of all files in folders at one place not repeating the code in different fixtures
runAllTestsInFolder invoce 1 test case but used in many test cases

pirapira:
@winsvega sounds good.

PR #4439, opened 2017-09-01
More boost::filesystem::path

pirapira:
This PR replaces many strings with boost::filesystem::path.


chfast:
I would much more preferred switching to C++17 first and use `std::filesystem`.

pirapira:
@chfast after this PR, that will be one replacement per file of `namespace fs = boost::filesystem` with `namespace fs = std::filesystem`.  I have nothing against doing C++17 first, though.

chfast:
No, C++17 is a big step.

PR #4443, opened 2017-09-01
fix --singlenet option

winsvega:
1 line change.


winsvega:
well I thought you could use --singletest on a disabled tests and receive nothing. because test is disabled. 
the issue is that singletest  should imply Frontier and Homestead on enabled tests.

pirapira:
@winsvega I believe https://github.com/ethereum/cpp-ethereum/pull/4445 is cleaner. Does that work for your problem?

pirapira:
Suddenly I realize this PR is about `--singlenet` option, not `--singletest` option.

PR #4445, opened 2017-09-01
Make testeth --singletest imply --all

pirapira:
An alternative to https://github.com/ethereum/cpp-ethereum/pull/4443


pirapira:
I think I saw something wrong.

PR #4451, opened 2017-09-04
Add 0x in some error messages

pirapira:
In test fillers and test files, state roots are expected to contain the 0x prefix.
This commit adds such prefixes in error messages so that it's easier to edit the test files using the error messages.

For example,
```
test/tools/libtesteth/ImportTest.cpp(640): error: in "StateTestsGeneral/stExample": add11 on Byzantium: Expected another postState hash! expected: 0x07454a767e5f04461256f3812ffca930443c04a47d05ce3f38940c4a14b8c479 actual: 17454a767e5f04461256f3812ffca930443c04a47d05ce3f38940c4a14b8c479 in Byzantium data: 0 gas: 0 val: 0
```
becomes
```
test/tools/libtesteth/ImportTest.cpp(640): error: in "StateTestsGeneral/stExample": add11 on Byzantium: Expected another postState hash! expected: 0x07454a767e5f04461256f3812ffca930443c04a47d05ce3f38940c4a14b8c479 actual: 0x17454a767e5f04461256f3812ffca930443c04a47d05ce3f38940c4a14b8c479 in Byzantium data: 0 gas: 0 val: 0
```

### DIFF
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -817,7 +817,7 @@ void compareBlocks(TestBlock const& _a, TestBlock const& _b)
 	if (sha3(RLP(_a.bytes())[0].data()) != sha3(RLP(_b.bytes())[0].data()))
 	{
 		cnote << "block header mismatch\n";
-		cnote << toHex(RLP(_a.bytes())[0].data()) << "vs" << toHex(RLP(_b.bytes())[0].data());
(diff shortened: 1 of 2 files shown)


winsvega:
I thought we do not add 0x to hashes?

pirapira:
Currently we do in the filled tests.

winsvega:
I see so we stick with a scheme to keep all fields 0x prefixed. we actually were thinking of rolling back to hashes without 0x  
I guess the clients are ok with the current scheme

PR #4452, opened 2017-09-04
Allow "ALL" as a network name in the "expect" section

pirapira:
This fixes a bug I introduced in #4435 .  "ALL" should be allowed as a network name.

### DIFF
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -151,7 +151,8 @@ void spellCheckNetworkNamesInExpectField(json_spirit::mArray const& _expects)
 		json_spirit::mObject const& expectObj = expect.get_obj();
 			ImportTest::parseJsonStrValueIntoVector(expectObj.at("network"), netlist);
 			for (string const& networkName: netlist)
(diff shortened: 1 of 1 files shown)


test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -151,7 +151,8 @@ void spellCheckNetworkNamesInExpectField(json_spirit::mArray const& _expects)
 		json_spirit::mObject const& expectObj = expect.get_obj();
 			ImportTest::parseJsonStrValueIntoVector(expectObj.at("network"), netlist);
 			for (string const& networkName: netlist)
-				(void)stringToNetId(networkName);
+				if (networkName != "ALL") // "ALL" is allowed as a wildcard.
```
winsvega: should we use a c_ALL  string constant for this?

pirapira: Yes.

pirapira: Will be done https://github.com/ethereum/cpp-ethereum/issues/4453

pirapira:
@winsvega ready to merge?

PR #4458, opened 2017-09-05
Remove the blockhash contract when it is empty

### LINKED ISSUE #4457: testeth sometimes produces a poststate containing empty accounts for Constantinople network
For example https://github.com/ethereum/tests/pull/303#discussion_r136947505

pirapira:
During Constantinople Blockchaintests, `constantinopleForkBlock` is set to zero, but the block zero is never processed, so the blockhash contract stays empty.  After that, in block number one, the blockhash contract is called, so it is created in the trie as an empty account.  And the empty blockhash contract stays because the commit behavior was not `RemoveEmptyAccounts`.

This PR changes the commit behavior.  I'm discussing with @winsvega what to do about the blockhash contract deployment.

Fixes #4457.

Work in progress: filling blockchaintests on this PR.

### DIFF
--- a/libethereum/Block.cpp
+++ b/libethereum/Block.cpp
@@ -720,7 +720,7 @@ void Block::updateBlockhashContract()
 			e.go();
 		e.finalize();
 
-		m_state.commit(State::CommitBehaviour::KeepEmptyAccounts);
+		m_state.commit(State::CommitBehaviour::RemoveEmptyAccounts);
 	}
 }
 


gumb0:
Yeah, I think if `constantinopleForkBlock  == 0`, blockhash contract should be part of genesis block somehow.
If you want to test blockhash contract's coming into existence, you need network with `constantinopleForkBlock  > 0`

winsvega:
we could put a contract to the genesis pre state on a stage of test filling.
and we have transition tests for constantinople at block 5

pirapira:
@winsvega an issue about adding the blockhash contract in the pre state of filled tests https://github.com/ethereum/cpp-ethereum/issues/4459

pirapira:
Wow, the PR makes all Constantinople blockchain tests failing.

winsvega:
I guess it is failing because of the empty contracts?

pirapira:
@winsvega it seems like so.

pirapira:
Depends on https://github.com/ethereum/tests/issues/305

PR #4460, opened 2017-09-05
generalize test execution logic

winsvega:
the goal of this change is to make 
```void executeTests```
private and eventually simplify the logic of executeTests itself.

in this PR executeTests removed from VMTests

### DIFF
--- a/test/tools/jsontests/vm.cpp
+++ b/test/tools/jsontests/vm.cpp
@@ -28,7 +28,6 @@
 #include <libevm/ExtVMFace.h>
 #include <boost/filesystem.hpp>
 
-
 using namespace std;
 using namespace json_spirit;
 using namespace dev;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: test/tools/libtesteth/TestHelper.cpp [hunk 1]
```diff
@@ -404,6 +404,39 @@ void checkCallCreates(eth::Transactions const& _resultCallCreates, eth::Transact
...
+	}
+
+	path const fillersPath = path(test::getTestPath()) / "src" / path(testSuiteFolder + "Filler") / path(_folder);
+	string const filter = test::Options::get().singleTestName.empty() ? string() : test::Options::get().singleTestName + "Filler";
+	std::vector<boost::filesystem::path> const files = test::getJsonFiles(fillersPath.string(), filter);
+	size_t fileCount = files.size();
+	if (test::Options::get().filltests)
+		fileCount *= 2; //tests are checked when filled and after they been filled
+
+	path const destTestFolder = path(testSuiteFolder) / path(_folder);
```
winsvega: once executeTests is simplified. this wont be needed.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -499,8 +532,9 @@ void addClientInfo(json_spirit::mValue& _v, fs::path const& _testSource)
 				comment = existingInfo["comment"].get_str();
 		}
 
+		string const& source = _testSource.string();
 		clientinfo["filledwith"] = prepareVersionString();
-		clientinfo["source"] = _testSource.string();
+		clientinfo["source"] = source.substr(source.rfind("/src/"));
```
winsvega: get a relative path of the source filler to put it to the comment section

pirapira: I oppose parsing and decomposing strings here.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -404,6 +404,39 @@ void checkCallCreates(eth::Transactions const& _resultCallCreates, eth::Transact
...
+void runAllTestsInFolder(test::testType _type, const std::string& _folder)
+{
+	using path = boost::filesystem::path;
+	string testSuiteFolder;
+	std::function<json_spirit::mValue(json_spirit::mValue const&, bool)> doTests;
+	switch (_type)
+	{
+		case test::testType::VMTests:
+			testSuiteFolder = test::c_VMTests;
+			doTests = test::doVMTests;
```
pirapira: Declare a `struct` containing fields like `testSuiteFolder` and `doTests`.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -404,6 +404,39 @@ void checkCallCreates(eth::Transactions const& _resultCallCreates, eth::Transact
 	}
 }
 
+void runAllTestsInFolder(test::testType _type, const std::string& _folder)
+{
+	using path = boost::filesystem::path;
```
pirapira: In other places of the codebase, `using fs = boost::filesystem` is used.  Using the same name here would make the codebase more readable.

winsvega: like using fs = boost::filesystem 
ad fs::path()  instead path() ?

test/tools/libtesteth/TestHelper.cpp
winsvega: what your suggestion then?

test/tools/libtesteth/TestHelper.cpp
pirapira: yes.

test/tools/libtesteth/TestHelper.cpp
pirapira: `addClientInfo()` should not receive the absolute path.  It should receive the path to be printed.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -404,6 +404,48 @@ void checkCallCreates(eth::Transactions const& _resultCallCreates, eth::Transact
 	}
 }
 
+struct doTestInfo
```
pirapira: `testTypeParams`?

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -434,7 +476,7 @@ void executeTests(const string& _name, fs::path const& _testPathAppendix, fs::pa
 		json_spirit::read_string(s, v);
 		removeComments(v);
 		json_spirit::mValue output = doTests(v, true);
-		addClientInfo(output, testfilename);
+		addClientInfo(output, testfileUnderTestPath);
```
pirapira: Looks like I forgot this change in a merge conflict.  Thanks.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -404,6 +404,48 @@ void checkCallCreates(eth::Transactions const& _resultCallCreates, eth::Transact
...
+doTestInfo getDoTestsInfo(test::testType _type)
+{
+	doTestInfo info;
+	switch (_type)
+	{
+		case test::testType::VMTests:
+			info.testSuiteFolder = test::c_VMTests;
+			info.doTests = test::doVMTests;
+			break;
+		default:
```
pirapira: I think it should fail, saying "not implemented yet".

test/tools/libtesteth/TestHelper.cpp
winsvega: thats why conflicts on rebase PR are bad

test/tools/libtesteth/TestHelper.h
```diff
@@ -105,6 +111,7 @@ dev::eth::BlockHeader constructHeader(
 	u256 const& _timestamp,
 	bytes const& _extraData);
 void updateEthashSeal(dev::eth::BlockHeader& _header, h256 const& _mixHash, dev::eth::Nonce const& _nonce);
+void runAllTestsInFolder(test::testType _type, const std::string& _folder);
```
gumb0: `std::string const&`

test/tools/libtesteth/TestHelper.h
```diff
@@ -60,6 +60,12 @@ struct ValueTooLarge: virtual Exception {};
 struct MissingFields : virtual Exception {};
 bigint const c_max256plus1 = bigint(1) << 256;
 extern std::string const c_StateTestsGeneral;
+extern std::string const c_VMTests;
+enum class testType
```
gumb0: Type name should start with upper case letter

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -404,6 +404,49 @@ void checkCallCreates(eth::Transactions const& _resultCallCreates, eth::Transact
...
+			params.doTests = test::doVMTests;
+			break;
+		default:
+			BOOST_ERROR("getTestTypeParams: undefined test type!");
+		break;
+	}
+	return params;
+}
+
+void runAllTestsInFolder(test::testType _type, const std::string& _folder)
```
gumb0: `std::string const&`

test/tools/jsontests/vm.cpp
```diff
@@ -477,30 +477,11 @@ class VmTestFixture
...
-		{
-			test::TestOutputHelper::setCurrentTestFileName(file.filename().string());
-			test::executeTests(file.filename().string(), "/VMTests/"+_folder, "/VMTestsFiller/"+_folder, dev::test::doVMTests);
-		}
+		test::runAllTestsInFolder(test::testType::VMTests, casename);
 	}
 };
 
-
+std::string const test::c_VMTests = "VMTests";
```
gumb0: Why wouldn't you put this into `TestHelper.cpp` ? It's kind of hard to find this definition when declaration is in `TestHelper.h` and definition in some unpredictable file elsewhere

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -404,6 +404,49 @@ void checkCallCreates(eth::Transactions const& _resultCallCreates, eth::Transact
 	}
 }
 
+struct testTypeParams
```
gumb0: Just a note: more object-oriented way to do this would be to have an interface like
```
class TestSute
{
public:
  virtual ~TestSute() {}
  virtual json_spirit::mValue doTests(json_spirit::mValue const&, bool) = 0;
  virtual std::string folder() const = 0;
};
```

have different implemntations for different suites and instead of `getTestTypeParams` something like
```
std::unique_ptr<TestSute> createSuite(test::testType _type);
```

gumb0: type name should star with upper case (`TestTypeParams`)

test/tools/jsontests/vm.cpp
winsvega: because those strings should match the name of the boost test cases

pirapira: @gumb0 @winsvega the fundamental weird thing: the definitions in `vm.cpp` implement some declarations in `TestHelper.h`.  `doVMTests()` is declared in `TestHelper.h` and is defined in `vm.cpp`.

winsvega:
@gumb0, I followed your suggestion and replaced the struct with abstract class.
I had to make a copy of executetests function to accept new param.
please review if this implements the idea that you proposed. I could continue to work in this direction.

test/tools/libtesteth/TestHelper.h
```diff
@@ -60,7 +60,13 @@ struct ValueTooLarge: virtual Exception {};
 struct MissingFields : virtual Exception {};
 bigint const c_max256plus1 = bigint(1) << 256;
 extern std::string const c_StateTestsGeneral;
-
+class TestSute
```
gumb0: I recommend putting it into separate file like `TestSuite.h`

test/tools/jsontests/vm.cpp
```diff
@@ -293,178 +292,181 @@ eth::OnOpFunc FakeExtVM::simpleTrace() const
 	};
 }
 
+
+#include <test/tools/libtesteth/TestHelper.h>
 namespace dev { namespace test {
 
-json_spirit::mValue doVMTests(json_spirit::mValue const& _input, bool _fillin)
+class VmTestSuite : public TestSute
```
gumb0: Should be no space between `VmTestSuite` and colon

test/tools/jsontests/vm.cpp
```diff
@@ -293,178 +292,181 @@ eth::OnOpFunc FakeExtVM::simpleTrace() const
...
 {
-	json_spirit::mValue v = json_spirit::mObject();
-	json_spirit::mObject& output = v.get_obj();
-	for (auto& i: _input.get_obj())
+	json_spirit::mValue doTests(json_spirit::mValue const& _input, bool _fillin) const
```
gumb0: Use `override` specifier

test/tools/jsontests/vm.cpp
```diff
@@ -293,178 +292,181 @@ eth::OnOpFunc FakeExtVM::simpleTrace() const
...
-	}
-
-	return v;
-}
-
-} } // namespace close
 
+		return v;
+	}
+	std::string folder() const	{ return "VMTests"; }
```
gumb0: Use `override`

test/tools/libtesteth/TestHelper.h
```diff
@@ -105,7 +111,8 @@ dev::eth::BlockHeader constructHeader(
 	u256 const& _timestamp,
 	bytes const& _extraData);
 void updateEthashSeal(dev::eth::BlockHeader& _header, h256 const& _mixHash, dev::eth::Nonce const& _nonce);
-void executeTests(const std::string& _name, boost::filesystem::path const& _testPathAppendix, boost::filesystem::path const& _fillerPathAppendix, std::function<json_spirit::mValue(json_spirit::mValue const&, bool)> doTests, bool _addFillerSuffix = true);
+void runAllTestsInFolder(test::TestSute const& _testSuite, std::string const& _folder);
```
gumb0: I'm confused, what does `_folder` refer to and what does `_testSuite.folder()` refer to?

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -404,7 +404,75 @@ void checkCallCreates(eth::Transactions const& _resultCallCreates, eth::Transact
...
+
+	auto testOutput = dev::test::TestOutputHelper(fileCount);
+	for (auto const& file: files)
+	{
+		test::TestOutputHelper::setCurrentTestFileName(file.filename().string());
+		executeTests2(file.filename().string(), destTestFolder.string(), srcTestFolder.string(), _suite);
+	}
+}
+
+void executeTests2(const string& _name, fs::path const& _testPathAppendix, fs::path const& _fillerPathAppendix, test::TestSute const& _suite)
```
gumb0: Well, we definitely should find a way to avoid this copy-paste. 
How about `runAllTestsInFolder` calling `executeTests` with lambda for `std::function` parameter
```
executeTests(...,
  [&_suite](json_spirit::mValue const& _input, bool _fillin) { _suite.doTests(_input, _fillin); } )
```
(this should be on line 425)

gumb0:
This is good, but `executeTest2` is not good. I think we should be able to get rid of it.

test/tools/libtesteth/TestHelper.cpp
gumb0: I suppose this should be a temporary measure and in the end all test suites should implemet `TestSuite` and `executeTest` will have `TestSuite` parameter instead of `std::function` and lambda will be not needed

test/tools/libtesteth/TestHelper.h
winsvega: _folder in runAllTestsInFolder is the folder containing the test files
the folder() of testSuite class is the folder containing all folders with tests for that test suite. 
perhaps testSuite should have  like srcFolder,  destFolder

test/tools/libtesteth/TestHelper.cpp
winsvega: I tried that it did not compile. 
this lambda function would have also to return json_spirit::mValue

winsvega:
```
json_spirit::mValue suiteTestDo = [&_suite](json_spirit::mValue const& _input, bool _fillin)
	{
		return _suite.doTests(_input, _fillin);
	};

/home/wins/Ethereum/cpp-ethereum/test/tools/libtesteth/TestHelper.cpp:424: error: conversion from ‘dev::test::runAllTestsInFolder(const dev::test::TestSute&, const string&)::__lambda20’ to non-scalar type ‘json_spirit::mValue {aka json_spirit::Value_impl<json_spirit::Config_map<std::basic_string<char> > >}’ requested
  };
  ^
```

winsvega: sorry just checked out.  auto  lambda works!

test/tools/libtesteth/TestHelper.h
winsvega: tha would be a pretty small header like 3 lines

winsvega:
@gumb0  updated the PR

test/tools/libtesteth/TestHelper.h
gumb0: So is `_folder` always a subfolder of `_testSuite.folder()` ? Maybe call the parameter then `_subfolder`  and the function `runTestsInSubfolder` ?
Or at least add a comment about this.

gumb0: It would be nice to see the relation between `runAllTestsInFolder` and `executeTests` from their names.
Like for example naming them `runTestsInSubfolder` & `runTestsInFile` (also you could have `runTestsInSuite` somewhere above)

winsvega:
hm the way it is right now we could actually init TestSuite to know which subfolder to use for the test.
and then runAllTests would just accept the TestSuite class reference with all the info. 
is this a way to go?

gumb0:
^ Well, no, I think I wouldn't do it like that. If it's called `TestSuite` let it represent the whole suite, then use different subfolder names to run different parts of the suite

test/tools/libtesteth/TestHelper.h
gumb0: That's fine, it would make the code easier to navigate I think

winsvega:
I think runAllTestsInFolder could be a part of TestSuite class. that solves the issue

gumb0:
^ Yes, having it as a method of `TestSuite` seems natural, we can do it like that...

But I personally don't like mixing abstract methods with concrete methods in one class, I think it leads to confusion, I would then have still two separate things - `TestSuite` concrete class with `runTestsInSubfolder` method and separate abstract interface with `doTests` method (also would call it better `runTestsInJson`)

```
class TestSuite
{
  void runTestsInSubfolder(std::string const& _subfolder, TestRunner& _runner);
};

class TestRunner
{
  virtual ~TestRunner() {}
  virtual json_spirit::mValue runTestsInJson(json_spirit::mValue const&, bool) = 0;
};
```

winsvega:
hm but it run files in folder. not tests in json

gumb0:
`executeTests` now reads one file, parses json, then calls `doTests` to run the tests in this json, right?

winsvega:
not yet. but right that is what we heading to. 
executeTests should run 1 json file.

gumb0:
Resolve conflict in TestHelper.h

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/** @file
```
gumb0: This makes sense only if you provide some file description.
You can remove it altogether

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -0,0 +1,59 @@
...
+	string const filter = test::Options::get().singleTestName.empty() ? string() : test::Options::get().singleTestName + "Filler";
+	std::vector<boost::filesystem::path> const files = test::getJsonFiles(fillersPath.string(), filter);
+	size_t fileCount = files.size();
+	if (test::Options::get().filltests)
+		fileCount *= 2; //tests are checked when filled and after they been filled
+
+	fs::path const destTestFolder = fs::path(folder()) / fs::path(_folder);
+	fs::path const srcTestFolder = fs::path(folder() + "Filler") / fs::path(_folder);
+
+	auto suiteTestDo = [&](json_spirit::mValue const& _input, bool _fillin)
```
gumb0: I think you can have a capture here like `[]`, instead of `[&]`, then it will not capture all of the local variable and will be more efficient.

test/tools/libtesteth/TestHelper.h
```diff
@@ -37,6 +37,7 @@
 #include <test/tools/libtesteth/Options.h>
 #include <test/tools/libtesteth/ImportTest.h>
 #include <test/tools/libtesteth/TestOutputHelper.h>
+#include <test/tools/libtesteth/TestSuite.h>
```
gumb0: You don't need this `#include` here, put it in `vm.cpp` instead

test/tools/libtesteth/TestSuite.cpp
winsvega: it does not work because this  is const

gumb0: Either make `TestSuite::runAllTestsInFolder` & `TestSuite::doTests` both const, or both non-const

winsvega: it is both const

test/tools/libtesteth/TestSuite.cpp
gumb0: Put description here, too (you wrote it in header only)

PR #4469, opened 2017-09-06
MaxAddress should be relevant only in Constantinople and later

### LINKED ISSUE #4461: MaxAddress should be checked only on transaction and only on Constantinople
Check for sender is here
https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/Executive.cpp#L332

and here
https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/Executive.cpp#L256

those rules should be applied only for transaction creation. and only if a constantinople rules are enabled.
and we have a case here when   MaxAddress is an account that does a create. what YP says about that situation?

pirapira:
Fixes #4461


pirapira:
@winsvega It should be just about the transaction sender.  Your idea is worth a test case.  Will you add it in the test case spreadsheet?

winsvega:
added

PR #4473, opened 2017-09-06
Add Circle CI

chfast:


gumb0:
It doesn't run unit tests?

gumb0:
Ah I see, one of the testeths runs all the rest

chfast:
Yes. We should use ctest for this at some point, but I need to talk about it with @winsvega.

winsvega:
have you checked that ctest works as intended? it's been a while since we used it
the build on Travis still goes

chfast:
No, they do not.

PR #4474, opened 2017-09-06
Travis CI: Upgrade XCode to 9

chfast:
This removes the need for custom clang compiler installation.


pirapira:
Please rebase.

chfast:
This should also fix the issue with current Travis/macOS builds where clang cannot be installed.

PR #4489, opened 2017-09-12
transaction tests through TestSuite class

winsvega:
replace old doTransaction tests function with generic  doTests function

### DIFF
--- a/test/tools/libtesteth/TestSuite.cpp
+++ b/test/tools/libtesteth/TestSuite.cpp
@@ -31,17 +31,22 @@ namespace dev
 namespace test
 {
 
-void TestSute::runAllTestsInFolder(const std::string& _folder) const
+fs::path TestSuite::getFullPath(string const& _testFolder, bool _isFiller) const
 {
-	fs::path const fillersPath = fs::path(test::getTestPath()) / "src" / fs::path(folder() + "Filler") / fs::path(_folder);
+	string isFiller = _isFiller ? "Filler" : string();
+	return fs::path(test::getTestPath()) / "src" / fs::path(suiteFolder() + isFiller) / fs::path(_testFolder);
+}
+
+void TestSuite::runAllTestsInFolder(string const& _testFolder) const
+{
 	string const filter = test::Options::get().singleTestName.empty() ? string() : test::Options::get().singleTestName + "Filler";
-	std::vector<boost::filesystem::path> const files = test::getJsonFiles(fillersPath.string(), filter);
+	std::vector<boost::filesystem::path> const files = test::getJsonFiles(getFullPath(_testFolder, true).string(), filter);
 	size_t fileCount = files.size();
 	if (test::Options::get().filltests)
 		fileCount *= 2; //tests are checked when filled and after they been filled
 
-	fs::path const destTestFolder = fs::path(folder()) / fs::path(_folder);
-	fs::path const srcTestFolder = fs::path(folder() + "Filler") / fs::path(_folder);
+	fs::path const destTestFolder = fs::path(suiteFolder()) / fs::path(_testFolder);
+	fs::path const srcTestFolder = fs::path(suiteFolder() + "Filler") / fs::path(_testFolder);
 
 	auto suiteTestDo = [this](json_spirit::mValue const& _input, bool _fillin)
 	{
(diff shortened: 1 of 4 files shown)


test/tools/libtesteth/TestSuite.cpp
```diff
@@ -56,5 +61,20 @@ void TestSute::runAllTestsInFolder(const std::string& _folder) const
 	}
 }
 
+void TestSuite::copyAllTestsFromFolder(string const& _testFolder) const
```
gumb0: Will it be used for any tests besides Transaction tests? If not, maybe leave it in `TransactionTestSuite`

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -56,5 +61,20 @@ void TestSute::runAllTestsInFolder(const std::string& _folder) const
...
+{
+	std::vector<fs::path> const files = test::getJsonFiles(getFullPath(_testFolder, true).string());
+	for (auto const& file : files)
```
gumb0: No space before colon

test/tools/libtesteth/TestSuite.h
```diff
@@ -20,19 +20,32 @@
...
+	virtual std::string suiteFolder() const = 0;
+
+	//Structure  <suiteFolder>/<testFolder>/<test>.json
+	//Return full path to folder for tests from _testFolder
+	boost::filesystem::path getFullPath(std::string const& _testFolder, bool _isFiller) const;
+
+	//Execute all tests from _folder
+	void runAllTestsInFolder(std::string const& _testFolder) const;
+
+	void copyAllTestsFromFolder(std::string const& _testFolder) const;
```
gumb0: Add comment here, too, it's not quite clear where it copies and why it's needed

test/tools/libtesteth/TestSuite.h
```diff
@@ -20,19 +20,32 @@
...
 {
 
-class TestSute
+class TestSuite
 {
 public:
-	virtual ~TestSute() {}
+	virtual ~TestSuite() {}
+
+	//Main test executive function. should be declared for each test suite. it fills and run the test .json file
```
gumb0: "declared" is not quite correct (declaration is only the signature of the function), I'd say "implemented". (Also it's kind of clear from the fact that its pure virtual function, so maybe don't mention it at all)

gumb0: typo: "run" => "runs"

gumb0: Space after `//` would make it look nicer
Also, better start sentences with capital letter.

test/tools/libtesteth/TestSuite.h
```diff
@@ -20,19 +20,32 @@
...
 	virtual json_spirit::mValue doTests(json_spirit::mValue const&, bool) const = 0;
-	virtual std::string folder() const = 0;
-	void runAllTestsInFolder(std::string const& _folder) const;
+
+	//A folder of the test suite. like "VMTests". should be declared for each test suite.
+	virtual std::string suiteFolder() const = 0;
+
+	//Structure  <suiteFolder>/<testFolder>/<test>.json
+	//Return full path to folder for tests from _testFolder
+	boost::filesystem::path getFullPath(std::string const& _testFolder, bool _isFiller) const;
```
gumb0: Avoid bool parameters at all costs, it's a clear sign your method does two distinct things and should be really two separate methods. Here you should have for example `getFullPath` and `getFullFillerPath`

gumb0: Also, looks like you can make it private or even just a helper (two helpers) in anonymous namespace in TestSuite.cpp

test/tools/libtesteth/TestSuite.h
```diff
@@ -20,19 +20,32 @@
 
 #pragma once
 #include <test/tools/libtesteth/JsonSpiritHeaders.h>
+#include <boost/filesystem/path.hpp>
```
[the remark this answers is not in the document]
winsvega: what about 
`boost::filesystem::path getFullPath(std::string const& _testFolder, bool _isFiller) const;`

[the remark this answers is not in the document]
gumb0: Oh yeah, sorry, didn't notice it, let's keep it if you don't move `getFullPath` to cpp

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -31,17 +31,28 @@ namespace dev
...
 	string const filter = test::Options::get().singleTestName.empty() ? string() : test::Options::get().singleTestName + "Filler";
-	std::vector<boost::filesystem::path> const files = test::getJsonFiles(fillersPath.string(), filter);
+	std::vector<boost::filesystem::path> const files = test::getJsonFiles(getFullPathFiller(suiteFolder(), _testFolder).string(), filter);
 	size_t fileCount = files.size();
 	if (test::Options::get().filltests)
 		fileCount *= 2; //tests are checked when filled and after they been filled
 
-	fs::path const destTestFolder = fs::path(folder()) / fs::path(_folder);
-	fs::path const srcTestFolder = fs::path(folder() + "Filler") / fs::path(_folder);
+	fs::path const destTestFolder = fs::path(suiteFolder()) / fs::path(_testFolder);
```
gumb0: you don't need to convert `string` to `fs::path` on the right side of `/`

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -56,5 +67,22 @@ void TestSute::runAllTestsInFolder(const std::string& _folder) const
 	}
 }
 
+// Copy .json tests from the src folder to the dest folder because such test is crafted manually and could not be filled.
```
gumb0: I'd prefer this comment to be in header file

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -31,17 +31,28 @@ namespace dev
 namespace test
 {
 
-void TestSute::runAllTestsInFolder(const std::string& _folder) const
+// Structure  <suiteFolder>/<testFolder>/<test>.json
+// Return full path to folder for tests from _testFolder
+fs::path getFullPathFiller(string const& _suiteFolder, string const& _testFolder)
```
gumb0: It's better to define these inside anonymous namespace (`namespace {   } `) instead of `test`, because they're internal for this file.

If they're in `test` it can lead to name conflicts during link-time.

PR #4491, opened 2017-09-12
ecpairing final gas costs 100k base + 80k per point

cdetrio:
See the updated https://github.com/ethereum/EIPs/pull/212 ([permalink](https://github.com/ethereum/EIPs/blob/9961da4c77659c6217eae41130c3a40d3f526ac4/EIPS/pairings.md)).


holiman:
:+1:

PR #4494, opened 2017-09-13
CMake: Map imported configs with only single entry

chfast:
This should fix CMake 3.9 generator for Visual Studio.


gumb0:
AppVeyor build still fails

chfast:
Yes, but much further. Can you take a look? Something about packaging executables. Maybe CMake 3.9 also changes artifacts location?

gumb0:
Maybe. I'll try to install new cmake

PR #4496, opened 2017-09-13
Refactor ExtVMFace

chfast:
This moves utility structs like `LogEntry` from ExtVMFace.h (libevm) to libethcore. Previously some files from libethcore have to include ExtVMFace.h.

Also `Address` is moved from libdevcrypto to libdevcore as it is used not only in crypto context.


winsvega:
@pirapira

libdevcore/FixedHash.cpp
```diff
@@ -14,20 +14,25 @@
...
 #include <boost/algorithm/string.hpp>
 
 using namespace std;
 using namespace dev;
 
-std::random_device dev::s_fixedHashEngine;
+namespace dev
+{
+
+const Address ZeroAddress = Address();
```
gumb0: `Address const`

libdevcore/FixedHash.cpp
```diff
@@ -14,20 +14,25 @@
...
-/** @file FixedHash.cpp
- * @author Gav Wood <i@gavwood.com>
- * @date 2014
- */
 
 #include "FixedHash.h"
 #include <boost/algorithm/string.hpp>
 
 using namespace std;
 using namespace dev;
```
gumb0: I think we can remove this now

libdevcore/FixedHash.h
```diff
@@ -351,6 +351,25 @@ using h160Set = std::set<h160>;
 using h256Hash = std::unordered_set<h256>;
 using h160Hash = std::unordered_set<h160>;
 
+/// An Ethereum address: 20 bytes.
+/// @NOTE This is not endian-specific; it's just a bunch of bytes.
+using Address = h160;
```
gumb0: I'd prefer to put these into separate Address.h/cpp

libethcore/Transaction.h
```diff
@@ -14,19 +14,16 @@
...
 	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
 */
-/** @file TransactionBase.h
- * @author Gav Wood <i@gavwood.com>
- * @date 2014
- */
 
 #pragma once
 
+#include <boost/optional.hpp>
```
gumb0: `#include`s should be in increasing order of generality, so this goes after our own headers

libethereum/TransactionReceipt.h
```diff
@@ -22,11 +22,9 @@
 #pragma once
 
 #include <array>
-#include <libdevcore/Common.h>
-#include <libdevcore/RLP.h>
-#include <libevm/ExtVMFace.h>
-
 #include <boost/variant/variant.hpp>
+#include <libdevcore/RLP.h>
```
gumb0: should be before boost & `<array>`

libethcore/Common.h
```diff
@@ -90,6 +94,81 @@ enum class RelativeBlock: BlockNumber
...
 };
 
+enum class BlockPolarity
+{
+	Unknown,
+	Dead,
+	Live
+};
+
+struct LogEntry
```
gumb0: These could go into separate `LogEntry.h`, but it's up to you

libethcore/Transaction.h
chfast: I think we have it the other way around in most files. Personally, I don't care. Why is this better?

libdevcore/FixedHash.h
chfast: Two cons for this: increased build time (?) and additional maintenance cost. I'm not sure...

libethcore/Transaction.h
gumb0: Link from coding standards: http://stackoverflow.com/questions/614302/c-header-order/614333#614333

It prevents the headers from not including all necessary headers but still working because of weird inter-header dependency

libdevcore/FixedHash.h
gumb0: I think it's your way it could be theoretically increased build time, because everyone who needs only `FixedHash` has to compile `Address`-related stuff, too

gumb0: It would make navigating the code easier, putting `Address` into Address.h instead of suprisingly `FixedHash.h`

libethcore/BasicAuthority.h
```diff
@@ -23,6 +23,7 @@
 
 #pragma once
 
+#include <libdevcrypto/Common.h>
```
pirapira: Why is this added now?

libethcore/Common.cpp
```diff
@@ -191,5 +189,26 @@ string TransactionSkeleton::userReadable(bool _toProxy, function<pair<bool, stri
 		formatBalance(gas * gasPrice) + ".");
 }
 
+LogEntry::LogEntry(RLP const& _r)
+{
```
pirapira: Should there be an assertion against
- `_r` not being a list and
- `_r` being too short?

libethcore/Common.cpp
```diff
@@ -191,5 +189,26 @@ string TransactionSkeleton::userReadable(bool _toProxy, function<pair<bool, stri
...
+	address = (Address)_r[0];
```
pirapira: Rename it to `m_address`.

pirapira: Ahh, it's a struct. Nevermind.

libethcore/Common.h
```diff
@@ -90,6 +94,81 @@ enum class RelativeBlock: BlockNumber
...
+{
+	LogEntry() = default;
+	explicit LogEntry(RLP const& _r);
+	LogEntry(Address const& _address, h256s _ts, bytes&& _d): address(_address), topics(std::move(_ts)), data(std::move(_d)) {}
```
pirapira: I wonder why `_ts` is not an rvalue reference and `_d` is an rvalue reference...

chfast: Generic way is to pass by value and move in the initialization list (like with `h256s` case). This way you handle both copy and move, depending what argument you have.

The version with `&&` prevents you from passing argument by copy -- you can only pass an argument being rvalue. This is restrictive, but enforces some optimizations. I don't like it because it is more confusing than useful.

libethcore/BasicAuthority.h
chfast: Previously this header was implicitly included for Address. Now we still need it for crypto::Signature.

chfast:
I think I addressed all issues. Can you review before builds?

libethcore/Common.h
```diff
@@ -90,6 +96,13 @@ enum class RelativeBlock: BlockNumber
 	Pending = PendingBlock
 };
 
+enum class BlockPolarity
```
gumb0: This is used only in `LocalisedLogEntry`, maybe move it to `LogEntry.h`?

chfast: Unfortunately, this is also used in Client.cpp and ClientBase.cpp.

PR #4501, opened 2017-09-15
fix ctest cmake to parse all suites with correct depth

winsvega:


chfast:
Does it work now? Does it execute `*disabled()` test cases (probably yes)?

winsvega:
it works as if to run ./testeth without flags.

PR #4509, opened 2017-09-18
improve and package performance suite for public use

gcolvin:
Improve and package performance tests for public use. Includes
* better equalized timings
* a more realistic rng test
* python scripts for analysis
* a README.


gcolvin:
@chfast  I agree this belongs in a more general place - there is no C++ in it, but pushing to it forces a rebuild of cpp-ethereum.  Where to put it?
@tcsiwula This is where you can find it for now.

chfast:
I don't know. It can be https://github.com/ethereum/tests or https://github.com/ethereum/benchmarking or new repo.

If decision is made let me know. I know some git magic tricks to move it with the history.

chfast:
@holiman would you like to have such performance tests in "benchmarking" repo?

PR #4511, opened 2017-09-18
Activate Byzantium on Ropsten

gumb0:


libethashseal/genesis/ropsten.cpp
```diff
@@ -25,7 +25,7 @@ R"E(
 		"homesteadForkBlock": "0x00",
 		"EIP150ForkBlock": "0x00",
 		"EIP158ForkBlock": "0x0a",
-		"byzantiumForkBlock": "0xfffffffffffffff",
+		"byzantiumForkBlock": "0x19f0a0",
```
winsvega: I think we should also explicitly set daoHardforkBlock  here.

libethashseal/genesis/ropsten.cpp
```diff
@@ -50,10 +50,14 @@ R"E(
...
-		"0000000000000000000000000000000000000004": { "wei": "1", "precompiled": { "name": "identity", "linear": { "base": 15, "word": 3 } } },
+		"0000000000000000000000000000000000000001": { "precompiled": { "name": "ecrecover", "linear": { "base": 3000, "word": 0 } } },
+		"0000000000000000000000000000000000000002": { "precompiled": { "name": "sha256", "linear": { "base": 60, "word": 12 } } },
+		"0000000000000000000000000000000000000003": { "precompiled": { "name": "ripemd160", "linear": { "base": 600, "word": 120 } } },
+		"0000000000000000000000000000000000000004": { "precompiled": { "name": "identity", "linear": { "base": 15, "word": 3 } } },
+		"0000000000000000000000000000000000000005": { "precompiled": { "name": "modexp", "startBlock": "0x19f0a0" } },
+		"0000000000000000000000000000000000000006": { "precompiled": { "name": "alt_bn128_G1_add", "startBlock": "0x19f0a0", "linear": { "base": 500, "word": 0 } } },
+		"0000000000000000000000000000000000000007": { "precompiled": { "name": "alt_bn128_G1_mul", "startBlock": "0x19f0a0", "linear": { "base": 40000, "word": 0 } } },
```
winsvega: would it be better to hardcode this configuration as in the other precompiles? 
`"linear": { "base": 40000, "word": 0 }`

gumb0: do you mean for `alt_bn128_pairing_product` ?

winsvega: I mean hardcode `"linear": { "base": 40000, "word": 0 }`
for `alt_bn128_G1_mul`

so we won't have genesis config issues.

pirapira: Are you sure about the names of the precompiled?

libethashseal/genesis/ropsten.cpp
```diff
@@ -50,10 +50,14 @@ R"E(
...
-		"0000000000000000000000000000000000000001": { "wei": "1", "precompiled": { "name": "ecrecover", "linear": { "base": 3000, "word": 0 } } },
-		"0000000000000000000000000000000000000002": { "wei": "1", "precompiled": { "name": "sha256", "linear": { "base": 60, "word": 12 } } },
-		"0000000000000000000000000000000000000003": { "wei": "1", "precompiled": { "name": "ripemd160", "linear": { "base": 600, "word": 120 } } },
-		"0000000000000000000000000000000000000004": { "wei": "1", "precompiled": { "name": "identity", "linear": { "base": 15, "word": 3 } } },
+		"0000000000000000000000000000000000000001": { "precompiled": { "name": "ecrecover", "linear": { "base": 3000, "word": 0 } } },
+		"0000000000000000000000000000000000000002": { "precompiled": { "name": "sha256", "linear": { "base": 60, "word": 12 } } },
+		"0000000000000000000000000000000000000003": { "precompiled": { "name": "ripemd160", "linear": { "base": 600, "word": 120 } } },
+		"0000000000000000000000000000000000000004": { "precompiled": { "name": "identity", "linear": { "base": 15, "word": 3 } } },
+		"0000000000000000000000000000000000000005": { "precompiled": { "name": "modexp", "startBlock": "0x19f0a0" } },
```
gumb0: This is not right, the code parses `startingBlock`, not `startBlock`

`byzantiumTransitionTest` config is incorrect @winsvega @pirapira

pirapira: I'll fix that.

libethashseal/genesis/ropsten.cpp
```diff
@@ -25,7 +25,8 @@ R"E(
 		"homesteadForkBlock": "0x00",
 		"EIP150ForkBlock": "0x00",
 		"EIP158ForkBlock": "0x0a",
-		"byzantiumForkBlock": "0xfffffffffffffff",
+		"daoHardforkBlock": "0x00",
```
gumb0: it's 0 by default, probably has no effect, since dao fork happens only in transition

PR #4516, opened 2017-09-18
separate bench tests under --all option

winsvega:


test/unittests/libethcore/PrecompiledTest.cpp
```diff
@@ -656,6 +656,12 @@ namespace
 {
 void benchmarkPrecompiled(char const name[], vector_ref<const PrecompiledTest> tests, int n)
 {
+	if (!Options::get().all)
+	{
+		std::cout << "Skipping benchmart test because --all option is not specified.\n";
```
chfast: Typo: bechmart.

test/unittests/libethcore/PrecompiledTest.cpp
```diff
@@ -656,6 +656,12 @@ namespace
 {
 void benchmarkPrecompiled(char const name[], vector_ref<const PrecompiledTest> tests, int n)
 {
+	if (!Options::get().all)
```
chfast: There are also 2 SHA256 benchmarks. Search for `label("bench")` or `label("perf")`. Change the label to "bench" if "perf" is used.

winsvega: those are already work under --all option. anyway pushed the fix

PR #4517, opened 2017-09-19
Hunter Crypto++

chfast:
This uses cryptopp package from Hunter instead of building it with our CMake scripts.


libdevcrypto/CryptoPP.cpp
```diff
@@ -27,7 +27,7 @@
 #include <libdevcore/Assertions.h>
 #include <libdevcore/SHA3.h>
 
-static_assert(CRYPTOPP_VERSION == 570, "Wrong Crypto++ version");
+static_assert(CRYPTOPP_VERSION == 565, "Wrong Crypto++ version");
```
gumb0: Do we downgrade?

chfast: Yes. 5.6.5 is the last released version. Previously we used 5.6.5 + my commit fixing some compiler warnings. But we don't have care about warnings because Hunter is hiding them.

PR #4520, opened 2017-09-19
Resync pending state if needed before adding new transactions

gumb0:
This should fix the assert happening during sync:
https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/Block.cpp#L326

It happened when `m_working` in `Client` is not up-to-date with the latest modifications to `BlockChain` and then we tried to import new transactions from `TransactionQueue`

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -398,6 +398,8 @@ void Client::syncBlockQueue()
 
 void Client::syncTransactionQueue()
 {
+	resyncStateFromChain();
+
 	Timer timer;
 
 	h256Hash changeds;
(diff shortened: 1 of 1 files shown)


libethereum/Client.cpp
```diff
@@ -659,7 +658,7 @@ void Client::doWork(bool _doWait)
 	bool isSealed = false;
 	DEV_READ_GUARDED(x_working)
 		isSealed = m_working.isSealed();
-	if (!isSealed && !isSyncing() && !m_remoteWorking && m_syncTransactionQueue.compare_exchange_strong(t, false))
```
gumb0: This will make processing of pending transactions to start closer to sync finish.
(`isSyncing()` returns `false` when there are no more blocks to download, but there still may be many blocks to import in `BlockQueue`)

libethereum/Client.cpp
```diff
@@ -659,7 +658,7 @@ void Client::doWork(bool _doWait)
 	bool isSealed = false;
 	DEV_READ_GUARDED(x_working)
 		isSealed = m_working.isSealed();
```
gumb0: `

gumb0:
Wait, this leads to deadlock

PR #4523, opened 2017-09-19
Exception on invalid JSON or unknown field in genesis configuration

### LINKED ISSUE #4514: genesis configuration parsing should fail when there is an unknown field
There was a misspelled option "startBlock" in many genesis configuration files: https://github.com/ethereum/cpp-ethereum/pull/4513

In order to detect these at least at runtime, the genesis configuration parsing should fail when it sees an unknown field.

pirapira:
Fixes #4514.

### DIFF
--- a/libdevcore/JsonUtils.cpp
+++ b/libdevcore/JsonUtils.cpp
@@ -1,0 +1,33 @@
+/*
+	This file is part of cpp-ethereum.
+
+	cpp-ethereum is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <libdevcore/JsonUtils.h>
+#include <json_spirit/JsonSpiritHeaders.h>
+#include <set>
+#include <string>
+#include <ostream>
+
+void dev::validateFieldNames(json_spirit::mObject const& _obj, std::set<std::string> const& _allowedFields)
+{
+	for (auto const& elm: _obj)
+		if (_allowedFields.find(elm.first) == _allowedFields.end())
+		{
+			std::string const comment = "Unknown field in config: " + elm.first;
+			std::cerr << comment << "\n";
+			BOOST_THROW_EXCEPTION(UnknownField() << errinfo_comment(comment));
+		}
+}
--- a/libethereum/Account.cpp
+++ b/libethereum/Account.cpp
@@ -81,6 +81,50 @@ PrecompiledContract createPrecompiledContract(js::mObject& _precompiled)
 }
 
 }
+namespace
+{
+	string const c_str_wei = "wei";
+	string const c_str_finney = "finney";
+	string const c_str_balance = "balance";
+	string const c_str_nonce = "nonce";
+	string const c_str_code = "code";
+	string const c_str_storage = "storage";
+	string const c_str_shouldnotexist = "shouldnotexist";
+	string const c_str_precompiled = "precompiled";
+	std::set<string> const c_known_account_fields = {
+		c_str_wei, c_str_finney, c_str_balance, c_str_nonce, c_str_code, c_str_storage, c_str_shouldnotexist,
+		c_str_code, c_str_precompiled
+	};
+	void validateAccount(js::mObject const& o)
+	{
+		for (auto const& field: o)
+			if (c_known_account_fields.find(field.first) == c_known_account_fields.end())
+			{
+				string const comment = "Unrecognized account field: " + field.first;
+				cerr << comment << "\n";
+				BOOST_THROW_EXCEPTION(UnknownField() << errinfo_comment(comment));
+			}
+	}
+	void validateAccounts(js::mObject const& o)
+	{
+		for (auto const& account: o)
+			validateAccount(account.second.get_obj());
+	}
+	string const c_str_alloc = "alloc";
+	string const c_str_accounts = "accounts";
+	void validateAccountMapObj(js::mObject const& o)
+	{
+		for (auto const& field: o)
+		{
+			if (field.first == c_str_alloc)
+				validateAccounts(field.second.get_obj());
+			else if (field.first == c_str_accounts)
+				validateAccounts(field.second.get_obj());
+			else
+				validateAccount(field.second.get_obj());
+		}
+	}
+}
 AccountMap dev::eth::jsonToAccountMap(std::string const& _json, u256 const& _defaultNonce, AccountMaskMap* o_mask, PrecompiledContractMap* o_precompiled)
 {
 	auto u256Safe = [](std::string const& s) -> u256 {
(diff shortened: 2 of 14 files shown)


libethereum/ChainParams.cpp
```diff
@@ -115,19 +205,22 @@ ChainParams ChainParams::loadGenesis(string const& _json, h256 const& _stateRoot
...
-	cp.parentHash = h256(genesis["parentHash"].get_str());
-	cp.author = genesis.count("coinbase") ? h160(genesis["coinbase"].get_str()) : h160(genesis["author"].get_str());
-	cp.difficulty = genesis.count("difficulty") ? u256(fromBigEndian<u256>(fromHex(genesis["difficulty"].get_str()))) : 0;
-	cp.gasLimit = u256(fromBigEndian<u256>(fromHex(genesis["gasLimit"].get_str())));
-	cp.gasUsed = genesis.count("gasUsed") ? u256(fromBigEndian<u256>(fromHex(genesis["gasUsed"].get_str()))) : 0;
-	cp.timestamp = u256(fromBigEndian<u256>(fromHex(genesis["timestamp"].get_str())));
-	cp.extraData = bytes(fromHex(genesis["extraData"].get_str()));
+	validateGenesis(genesis);
+
+	// TODO: check for unknown fields
```
winsvega: so this comment could be removed

pirapira:
The failure in AppVeyor looks real.

test/unittests/libethereum/BlockChain.cpp
```diff
@@ -389,4 +390,18 @@ BOOST_AUTO_TEST_CASE(updateStats)
...
+{
+	h256 emptyStateRoot;
+	/* Below, a comma is missing between fields. */
+	BOOST_CHECK_THROW(ChainParams("{ \"sealEngine\" : \"unknown\" \"accountStartNonce\" : \"3\" }", emptyStateRoot), json_spirit::Error_position);
+}
+
+BOOST_AUTO_TEST_CASE(unknownFieldThrows)
+{
+	h256 emptyStateRoot;
+	/* Below, a comma is missing between fields. */
```
gumb0: this comment is not true

libethereum/ChainParams.cpp
```diff
@@ -51,50 +51,113 @@ ChainParams::ChainParams(string const& _json, h256 const& _stateRoot)
 	*this = loadConfig(_json, _stateRoot);
 }
 
+string const c_str_sealEngine = "sealEngine";
```
gumb0: Put all these globals and functions into anonymous namespace

libethereum/Account.cpp
```diff
@@ -81,6 +81,50 @@ PrecompiledContract createPrecompiledContract(js::mObject& _precompiled)
 }
 
 }
+namespace
+{
+	string const c_str_wei = "wei";
```
gumb0: I don't using underscores here other than for `c_` prefix. I'd prefer these to be `c_strWei` or just `c_wei` etc.
Also `c_knownAccountFields `

winsvega:
@pirapira  please rebase both this PR an the test repo on develop

pirapira:
Addressed all comments so far.

libethereum/ChainParams.cpp
```diff
@@ -107,6 +173,36 @@ ChainParams ChainParams::loadGenesisState(string const& _json, h256 const& _stat
...
+		c_extraData, c_mixHash, c_nonce
+	};
+
+	void validateGenesis(js::mObject const& _genesis)
+	{
+		for (auto const& elm: _genesis)
+			if (c_knownGenesisFields.find(elm.first) == c_knownGenesisFields.end())
+			{
+				string const comment = "Unknown field in genesis config[genesis]: " + elm.first;
+				cerr << comment << "\n";
```
gumb0: Do we need this? Wouldn't exception handler output it, too, since we add it to exception?

libethereum/ChainParams.cpp
```diff
@@ -51,50 +51,116 @@ ChainParams::ChainParams(string const& _json, h256 const& _stateRoot)
 	*this = loadConfig(_json, _stateRoot);
 }
 
+namespace
+{
+	string const c_sealEngine = "sealEngine";
```
gumb0: Usually we don't indent what's inside the namespace

libethereum/ChainParams.cpp
```diff
@@ -51,50 +51,116 @@ ChainParams::ChainParams(string const& _json, h256 const& _stateRoot)
...
+
+	set<string> const c_knownParamNames = {
+		c_minGasLimit, c_maxGasLimit, c_gasLimitBoundDivisor, c_homesteadForkBlock,
+		c_EIP150ForkBlock, c_EIP158ForkBlock, c_accountStartNonce, c_maximumExtraDataSize,
+		c_tieBreakingGas, c_blockReward, c_byzantiumForkBlock, c_constantinopleForkBlock,
+		c_daoHardforkBlock, c_minimumDifficulty, c_difficultyBoundDivisor, c_durationLimit,
+		c_chainID, c_networkID, c_allowFutureBlocks, c_registrar
+	};
+
+	void validateParams(js::mObject const& _params)
```
gumb0: Well, this is repetition. How about making one function like `validateFields(jsObject const& _obj, std<string> const& _allowedFields)`?

libethereum/Account.cpp
```diff
@@ -81,6 +81,50 @@ PrecompiledContract createPrecompiledContract(js::mObject& _precompiled)
...
+	string const c_nonce = "nonce";
+	string const c_code = "code";
+	string const c_storage = "storage";
+	string const c_shouldnotexist = "shouldnotexist";
+	string const c_precompiled = "precompiled";
+	std::set<string> const c_knownAccountFields = {
+		c_wei, c_finney, c_balance, c_nonce, c_code, c_storage, c_shouldnotexist,
+		c_code, c_precompiled
+	};
+	void validateAccount(js::mObject const& o)
```
gumb0: `_o`

libethereum/ChainParams.cpp
pirapira: Good idea.

libethereum/Account.cpp
pirapira: OK.

libethereum/ChainParams.cpp
pirapira: The exception handler currently does not output it. The handler is somewhere in boost_test.

libethereum/Account.cpp
```diff
@@ -81,6 +81,50 @@ PrecompiledContract createPrecompiledContract(js::mObject& _precompiled)
...
+		for (auto const& account: o)
+			validateAccount(account.second.get_obj());
+	}
+	string const c_alloc = "alloc";
+	string const c_accounts = "accounts";
+	void validateAccountMapObj(js::mObject const& o)
+	{
+		for (auto const& field: o)
+		{
+			if (field.first == c_alloc)
```
gumb0: This I guess can be simplified to `if (field.first == c_alloc || field.first == c_accounts)`

gumb0: Also, does anyone still uses these `alloc` and `account` options?

gumb0:
Some improvement suggestions but in general looks good.

libethereum/Account.cpp
pirapira: I tried removing these.

libdevcore/JsonUtils.cpp
```diff
@@ -0,0 +1,33 @@
...
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <libdevcore/JsonUtils.h>
+#include <json_spirit/JsonSpiritHeaders.h>
+#include <set>
+#include <string>
+#include <ostream>
+
+void dev::validateFieldNames(json_spirit::mObject const& _obj, std::set<std::string> const& _allowedFields)
```
winsvega: could we do a function like this to validate test fillers for unknown fields?
or do you think that json scheme on tests repo is enough?

pirapira: I think the Json scheme on test repo is enough.

pirapira:
The AppVeyor failure should be fixed after rebased on https://github.com/ethereum/cpp-ethereum/pull/4532

pirapira:
Rebased.

PR #4526, opened 2017-09-20
Use Python codecov

chfast:
Although the coverage is 6% lower, I think the results are more accurate. If you have time, you can check some of the changes.

Or maybe not...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,10 +75,8 @@ before_script:
 script:
     - cd $TRAVIS_BUILD_DIR/build && ../scripts/tests.sh $TRAVIS_TESTS
 after_success:
-    - cd $TRAVIS_BUILD_DIR/build
-    - make coverage.data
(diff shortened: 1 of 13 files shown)


winsvega:
should we enable this on Azure runalltests script?

gumb0:
This needs rebase

chfast:
Sorry, guys, this Python script is fast but produces invalid reports. If you check the report for this PR, you will find duplicated files there. So of them are incorrectly placed in the project root. E.g. AES.cpp is in project root and libdevcrypto. 

I though it happens due using relative include paths, so I fixed all of them, but it has not changed anything.

I reported the issue, but no answer so far: https://github.com/codecov/codecov-python/issues/110.

Because of this inconsistency in reports (and probably no support for Visual Studio), I'm thinking about changing coverage service provider. Any suggestions welcome.

PR #4529, opened 2017-09-21
Move EVM-C context support to ExtVMFace

chfast:
For incoming EVM-C changes it is the best when ExtVMFace is also an implementation of EVM-C context.


libevm/ExtVMFace.cpp
```diff
@@ -21,6 +21,263 @@ namespace dev
 {
 namespace eth
 {
+namespace
+{
+
+static_assert(sizeof(Address) == sizeof(evm_uint160be),
+			  "Address types size mismatch");
+static_assert(alignof(Address) == alignof(evm_uint160be),
+			  "Address types alignment mismatch");
```
axic: I think these would fit a single line, though that is down to taste.

axic:
@chfast not that familiar with this code to be provide a reasonable review. The wrappers looked alright though.

gumb0:
Is this mostly making `evm_host` variable a public static in ExtVMFace instead of the global inside `JitVM.cpp`? Why is it needed?

`ExtVMFace` is already 1) an interface 2) having some methods implemented 3) and public data members 4) and inheriting `evm_context` 😆 Isn't this too much adding statics to it?

chfast:
I agree the ExtVMFace is messy. But this change is temporary, will be removed soon: https://github.com/ethereum/cpp-ethereum/commit/249e5d1700ad2c3c1f2c66d34201989964aafb60.

I don't think ExtVMFace should be an interface. This is only needed for VMTests. But if we want to keep it this way (for unit testing EVM), we can move data members and some method bodies. The `evm_context` probably should stay - this is also just an interface but in C, so it is very verbose.

chfast:
Thank you for your understanding :)

PR #4532, opened 2017-09-21
Avoid holding multiple locks for updating stats

pirapira:
This might fix the frequent deadlocks seen in AppVeyor.

It didn't look important to update all fields at once.

### DIFF
--- a/libdevcore/vector_ref.h
+++ b/libdevcore/vector_ref.h
@@ -7,6 +7,7 @@
 #include <type_traits>
 #include <vector>
 #include <string>
+#include <atomic>
 
 #ifdef __INTEL_COMPILER
 #pragma warning(disable:597) //will not be called for implicit or explicit conversions
@@ -75,7 +76,7 @@ public:
 	/// @note adapted from OpenSSL's implementation.
 	void cleanse()
 	{
-		static unsigned char s_cleanseCounter = 0;
+		static std::atomic<unsigned char> s_cleanseCounter{0u};
 		uint8_t* p = (uint8_t*)begin();
 		size_t const len = (uint8_t*)end() - p;
 		size_t loop = len;
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -1165,9 +1165,13 @@ void BlockChain::updateStats() const
 			m_lastStats.memBlocks += i.second.size() + 64;
 	DEV_READ_GUARDED(x_details)
 		m_lastStats.memDetails = getHashSize(m_details);
+	size_t logBloomsSize = 0;
+	size_t blocksBloomsSize = 0;
 	DEV_READ_GUARDED(x_logBlooms)
-		DEV_READ_GUARDED(x_blocksBlooms)
-			m_lastStats.memLogBlooms = getHashSize(m_logBlooms) + getHashSize(m_blocksBlooms);
+		logBloomsSize = getHashSize(m_logBlooms);
+	DEV_READ_GUARDED(x_blocksBlooms)
+		blocksBloomsSize = getHashSize(m_blocksBlooms);
+	m_lastStats.memLogBlooms = logBloomsSize + blocksBloomsSize;
 	DEV_READ_GUARDED(x_receipts)
 		m_lastStats.memReceipts = getHashSize(m_receipts);
 	DEV_READ_GUARDED(x_blockHashes)
@@ -1188,13 +1192,6 @@ void BlockChain::garbageCollect(bool _force)
 	m_lastCollection = chrono::system_clock::now();
 
 	Guard l(x_cacheUsage);
-	WriteGuard l1(x_blocks);
-	WriteGuard l2(x_details);
-	WriteGuard l3(x_blockHashes);
-	WriteGuard l4(x_receipts);
-	WriteGuard l5(x_logBlooms);
-	WriteGuard l6(x_transactionAddresses);
-	WriteGuard l7(x_blocksBlooms);
 	for (CacheID const& id: m_cacheUsage.back())
 	{
 		m_inUse.erase(id);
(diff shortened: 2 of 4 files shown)


libethereum/BlockChain.cpp
```diff
@@ -1160,20 +1160,38 @@ template <class K, class T> static unsigned getHashSize(unordered_map<K, T> cons
 void BlockChain::updateStats() const
 {
 	m_lastStats.memBlocks = 0;
-	DEV_READ_GUARDED(x_blocks)
-		for (auto const& i: m_blocks)
-			m_lastStats.memBlocks += i.second.size() + 64;
-	DEV_READ_GUARDED(x_details)
-		m_lastStats.memDetails = getHashSize(m_details);
-	DEV_READ_GUARDED(x_logBlooms)
+	{
```
gumb0: braces are not needed (`DEV_READ_GUARDED` makes the following statement a body of the loop with 1 iteration)

gumb0: So yeah, changes in `updateStats` don't seem to do anything

libethereum/BlockChain.cpp
```diff
@@ -1187,42 +1205,56 @@ void BlockChain::garbageCollect(bool _force)
...
 
-	Guard l(x_cacheUsage);
-	WriteGuard l1(x_blocks);
-	WriteGuard l2(x_details);
-	WriteGuard l3(x_blockHashes);
-	WriteGuard l4(x_receipts);
-	WriteGuard l5(x_logBlooms);
-	WriteGuard l6(x_transactionAddresses);
-	WriteGuard l7(x_blocksBlooms);
 	for (CacheID const& id: m_cacheUsage.back())
```
gumb0: But you need `x_cacheUsage` be read-locked for this

libethereum/BlockChain.cpp
pirapira: Ah, OK.  It was not holding seven locks then.

libethereum/BlockChain.cpp
pirapira: OK.  I put back the guard (some lines above, might not be visible).

pirapira:
Ready to merge?

libethereum/BlockChain.cpp
```diff
@@ -1188,37 +1192,49 @@ void BlockChain::garbageCollect(bool _force)
...
+			}
 		case ExtraDetails:
-			m_details.erase(id.first);
-			break;
+			{
+				WriteGuard l2(x_details);
+				m_details.erase(id.first);
+				break;
+			}
+		// TODO: ExtraBlockHash seems missing.
```
gumb0: Yes, let's add it, with `m_blockHashes` & `x_blockHashes`

chfast: @pirapira addressed?

libethereum/BlockChain.cpp
```diff
@@ -1188,37 +1192,49 @@ void BlockChain::garbageCollect(bool _force)
...
 		case ExtraReceipts:
-			m_receipts.erase(id.first);
-			break;
+			{
+				WriteGuard l4(x_receipts);
```
chfast: I don't think you have to enumerate guards. They have different scopes.

chfast:
Do we have any unit tests for this? Have you checked with thread sanitizer?

libethereum/BlockChain.cpp
pirapira: Right, I should drop the numbers.

pirapira:
@chfast  there is a unit test about it https://github.com/ethereum/cpp-ethereum/blob/0dfac1e2da1e9069bd963d2a9ca94a8c34739b9b/test/unittests/libethereum/BlockChain.cpp#L358

I'm trying tsan.

pirapira:
ThreadSanitizer has no new complaints on the unit test.  When I ran the usual run of `testeth` with ThreadSanitizer, it ate gigabytes of memory and my mouse pointer froze, so I killed that.

chfast:
Yes, tsan can triple the memory usage.

chfast:
Weirdly, it look like a hand in some other test on Windows.

pirapira:
AppVeyor has a new feature: detects excessive locking.  So I try to fix that too.

pirapira:
tsan says "as if synchronized via sleep"

pirapira:
I removed all tsan warnings during ETHEREUM_TEST_PATH='../../tests' test/testeth -t '*/capability'

pirapira:
@gumb0 @chfast please review the additional changes.

libdevcore/vector_ref.h
```diff
@@ -76,10 +77,15 @@ class vector_ref
 	void cleanse()
 	{
 		static unsigned char s_cleanseCounter = 0;
+		size_t count = 0;
+		static std::mutex x_cleanseCounter;
```
gumb0: I think we can make it `std::atomic<unsigned char>` instead of using mutex

libethereum/BlockChain.cpp
```diff
@@ -1188,38 +1193,55 @@ void BlockChain::garbageCollect(bool _force)
...
 		case ExtraDetails:
+		{
+			WriteGuard l(x_details);
 			m_details.erase(id.first);
 			break;
+		}
+		case ExtraBlockHash:
+		{
+			WriteGuard l(x_blockHashes);
+			m_blockHashes.erase(static_cast<uint64_t>(FixedHash<32>::Arith(id.first)));
```
gumb0: Sorry, now I see that we actually don't add block hashes into `m_cacheUsage` because of this: https://github.com/ethereum/cpp-ethereum/blob/develop/libethereum/BlockChain.h#L392

So not handling `ExtraBlockHash` was not en error.

Let's remove this code, but leave `case ExtraBlockHash` with a comment that block hashes don't go into `m_cacheUsage`

pirapira: OK.  I'll add a comment.

libp2p/Host.cpp
```diff
@@ -942,6 +948,16 @@ void Host::restoreNetwork(bytesConstRef _b)
 	}
 }
 
+bool Host::peerSlotsAvailable(Host::PeerSlotType _type)
```
gumb0: How does this change help?

libdevcore/vector_ref.h
pirapira: I'll do this.

libp2p/Host.cpp
pirapira: I think I separated two lock-holdings.  And I saw a deadlock disappear.

libp2p/Host.h
```diff
@@ -296,6 +296,7 @@ class Host: public Worker
 	/// Interface addresses (private, public)
 	std::set<bi::address> m_ifAddresses;								///< Interface addresses.
 
+	mutable Mutex x_listenPort;
```
gumb0: I think we can make it `std::atomic<int>` instead of using mutex

libp2p/Host.cpp
pirapira: `peerCount()` takes a lock.

gumb0: Ah, I see, fine

pirapira:
Done changing. I see no tsan errors from `ETHEREUM_TEST_PATH='../../tests' test/testeth -t '*/capability'`.

winsvega:
Andrey did a comment about using atomic variables. We tried that one and tha issue still happend.

chfast:
Awesome work @pirapira and @gumb0.

PR #4533, opened 2017-09-21
[WIP] Sync improvements

gumb0:
Trying to avoid being confused by nodes of another chain


gumb0:
The better approach was implemented in https://github.com/ethereum/cpp-ethereum/pull/4545

PR #4534, opened 2017-09-21
fuzzTests optimization

winsvega:
add boost test
pass random code options as a reference

### DIFF
--- a/test/tools/fuzzTesting/createRandomTest.cpp
+++ b/test/tools/fuzzTesting/createRandomTest.cpp
@@ -29,36 +29,30 @@
 #include <libdevcore/Common.h>
 #include <test/tools/jsontests/StateTests.h>
 
-//String Variables
-extern std::string const c_testExampleStateTest;
-extern std::string const c_testExampleTransactionTest;
-extern std::string const c_testExampleVMTest;
-extern std::string const c_testExampleBlockchainTest;
-extern std::string const c_testExampleRLPTest;
-
-//Main Test functinos
-int fillRandomTest(dev::test::TestSuite const& _testSuite, std::string const& _testString);
-
 namespace dev { namespace test {
-int createRandomTest()
+
+bool createRandomTest()
 {
 	StateTestSuite suite;
 	dev::test::Options& options = const_cast<dev::test::Options&>(dev::test::Options::get());
 	if (options.rCurrentTestSuite != suite.suiteFolder())
 	{
 		std::cerr << "Error! Test suite '" + options.rCurrentTestSuite + "' not supported! (Usage -t <TestSuite>)\n";
-		return 1;
+		return false;
 	}
 	else
 	{
+		RandomCodeOptions options;
 		TestOutputHelper testOutputHelper;
-		return fillRandomTest(suite, c_testExampleStateTest);
+		std::string test = test::RandomCode::fillRandomTest(suite, c_testExampleStateTest, options);
+		return test.empty() ? false : true;
 	}
 }
+
 }} //namespaces
 
 //Prints a generated test Json into std::out
-int fillRandomTest(dev::test::TestSuite const& _testSuite, std::string const& _testString)
+std::string dev::test::RandomCode::fillRandomTest(dev::test::TestSuite const& _testSuite, std::string const& _testString, dev::test::RandomCodeOptions const& _options)
 {
 	bool wasError = false;
 	json_spirit::mValue v;
@@ -66,7 +60,7 @@ int fillRandomTest(dev::test::TestSuite const& _testSuite, std::string const& _t
 	{
 		std::string newTest = _testString;
 		std::map<std::string, std::string> nullReplaceMap;
-		dev::test::RandomCode::parseTestWithTypes(newTest, nullReplaceMap);
+		dev::test::RandomCode::parseTestWithTypes(newTest, nullReplaceMap, _options);
 		json_spirit::read_string(newTest, v);
 		v = _testSuite.doTests(v, true); //filltests
 		_testSuite.doTests(v, false); //checktest
@@ -85,17 +79,17 @@ int fillRandomTest(dev::test::TestSuite const& _testSuite, std::string const& _t
 	if (wasError)
 	{
 		std::cerr << json_spirit::write_string(v, true);
-		return 1;
+		return std::string();
 	}
+	else
+		std::cout << json_spirit::write_string(v, true);
 
-	std::cout << json_spirit::write_string(v, true);
-	return 0;
+	return json_spirit::write_string(v, true);
 }
 
 /// Parse Test string replacing keywords to fuzzed values
-void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std::string, std::string> const& _varMap)
+void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std::string, std::string> const& _varMap, RandomCodeOptions const& _options)
 {
-	dev::test::RandomCodeOptions options; //use default options
 	std::vector<std::string> types = getTypes();
 
 	for (std::map<std::string, std::string>::const_iterator it = _varMap.begin(); it != _varMap.end(); it++)
@@ -115,7 +109,7 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
 				cnote << debug;
 			}
 			else if (type == "[CODE]")
-				replace = dev::test::RandomCode::generate(10, options);
+				replace = dev::test::RandomCode::generate(10, _options);
 			else if (type == "[HEX]")
 				replace = dev::test::RandomCode::randomUniIntHex();
 			else if (type == "[HEX32]")
@@ -140,14 +134,14 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
 				replace = test::RandomCode::randomUniIntHex(dev::u256("100000"), dev::u256("36028797018963967"));
 			else if (type == "[DESTADDRESS]")
 			{
-				Address address = options.getRandomAddress(RandomCodeOptions::AddressType::PrecompiledOrStateOrCreate);
+				Address address = _options.getRandomAddress(RandomCodeOptions::AddressType::PrecompiledOrStateOrCreate);
 				if (address != ZeroAddress) //else transaction creation
 					replace = "0x" + toString(address);
 			}
 			else if (type == "[ADDRESS]")
-				replace = toString(options.getRandomAddress(RandomCodeOptions::AddressType::StateAccount));
+				replace = toString(_options.getRandomAddress(RandomCodeOptions::AddressType::StateAccount));
 			else if (type == "[0xADDRESS]")
-				replace = "0x" + toString(options.getRandomAddress(RandomCodeOptions::AddressType::StateAccount));
+				replace = "0x" + toString(_options.getRandomAddress(RandomCodeOptions::AddressType::StateAccount));
 			else if (type == "[TRANSACTIONGASLIMIT]")
 				replace = test::RandomCode::randomUniIntHex(dev::u256("5000"), dev::u256("10000000"));
 			else if (type == "[GASPRICE]")
--- a/test/tools/fuzzTesting/fuzzHelper.cpp
+++ b/test/tools/fuzzTesting/fuzzHelper.cpp
@@ -568,6 +568,7 @@ BOOST_AUTO_TEST_CASE(rndCode)
 	try
 	{
 		test::RandomCodeOptions options;
+		options.emptyCodeProbability = 0;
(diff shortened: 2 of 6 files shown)


test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -85,17 +79,18 @@ int fillRandomTest(dev::test::TestSuite const& _testSuite, std::string const& _t
...
+		std::cout << json_spirit::write_string(v, true);
 
-	std::cout << json_spirit::write_string(v, true);
-	return 0;
+	return json_spirit::write_string(v, true);
 }
 
 /// Parse Test string replacing keywords to fuzzed values
-void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std::string, std::string> const& _varMap)
+void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std::string, std::string> const& _varMap, RandomCodeOptions& _options)
```
pirapira: Can this be `RandomCodeOptions const&`?

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -85,17 +79,18 @@ int fillRandomTest(dev::test::TestSuite const& _testSuite, std::string const& _t
...
 {
-	dev::test::RandomCodeOptions options; //use default options
+	dev::test::RandomCodeOptions& options = _options;
```
pirapira: You can remove this line, and use `_options` below.

test/tools/fuzzTesting/createRandomTest.cpp
winsvega: I guess no. if it use random generators it can't be const. and we want it to use both static (from input) and random code generation.

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -253,7 +254,7 @@ std::string RandomCode::rndByteSequence(int _length, SizeStrictness _sizeType)
 }
 
 //generate smart random code
-std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions _options)
+std::string RandomCode::generate(int _maxOpNumber, RandomCodeOptions& _options)
```
pirapira: Can this be `RandomCodeOptions const&`?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -564,9 +563,24 @@ BOOST_AUTO_TEST_CASE(rndCode)
...
-		std::string code = test::RandomCode::generate(1000);
-		cnote << code;
+		test::RandomCodeOptions options;
+		std::string code = test::RandomCode::generate(1000, options);
+		BOOST_REQUIRE(!code.empty());
+	}
+	catch(...)
+	{
+		BOOST_ERROR("Exception thrown when generating random code!");
+	}
```
pirapira: Can't there be a way to show what kind of exception it was.  At least show `e.what()`.

test/tools/fuzzTesting/createRandomTest.cpp
pirapira: Which field has to be modified?

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -29,17 +29,8 @@
...
-extern std::string const c_testExampleVMTest;
-extern std::string const c_testExampleBlockchainTest;
-extern std::string const c_testExampleRLPTest;
-
-//Main Test functinos
-int fillRandomTest(dev::test::TestSuite const& _testSuite, std::string const& _testString);
-
 namespace dev { namespace test {
+
 int createRandomTest()
```
pirapira: If the return value is either `true` or `false`, make it a `bool`.

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -341,7 +340,7 @@ std::string RandomCode::getPushCode(int _value)
 	return getPushCode(hexString);
 }
 
-std::string RandomCode::fillArguments(eth::Instruction _opcode, RandomCodeOptions const& _options)
+std::string RandomCode::fillArguments(eth::Instruction _opcode, RandomCodeOptions& _options)
```
pirapira: Why did you need to remove `const`?

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: hmm. although if I move opCodeProbability out of options I guess it might become const

test/tools/fuzzTesting/fuzzHelper.h
```diff
@@ -108,7 +128,7 @@ class RandomCode
 	static int randomPercent() { refreshSeed(); return percentDist(gen); }
 
 private:
-	static std::string fillArguments(dev::eth::Instruction _opcode, RandomCodeOptions const& _options);
+	static std::string fillArguments(dev::eth::Instruction _opcode, RandomCodeOptions& _options);
```
pirapira: Why did you need to remove `const`?

test/tools/fuzzTesting/createRandomTest.cpp
pirapira: Why do you think it can't be const when it uses random generators?

test/tools/fuzzTesting/fuzzHelper.cpp
pirapira: Which line changes `opCodeProbability`?

test/tools/libtesteth/TestHelper.h
```diff
@@ -112,7 +112,7 @@ json_spirit::mObject fillJsonWithState(eth::State const& _state, eth::AccountMas
 json_spirit::mObject fillJsonWithTransaction(eth::Transaction const& _txn);
 
 //Fill Test Functions
-int createRandomTest();	//returns 0 if succeed, 1 if there was an error;
+bool createRandomTest();	//returns 0 if succeed, 1 if there was an error;
```
pirapira: Update the comment, and everywhere that uses the return value.

winsvega: what should be updated everywhere that uses the return value?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -550,12 +547,19 @@ Address RandomCodeOptions::getRandomAddress(AddressType _type) const
 	}
 }
 
-void RandomCodeOptions::setWeights()
+int RandomCode::weightedOpcode(std::vector<int>& _weights)
```
pirapira: Can this function receive `opCodeProbability` instead of `_weights`?  Then, you don't need the static variable.  Static variables are like global variables, so let's try not to introduce them.

pirapira: Alternatively, maybe you can make `opCodeProbability` a local variable in this function.  That seems cleaner.

test/tools/fuzzTesting/fuzzHelper.h
```diff
@@ -121,6 +138,7 @@ class RandomCode
 	static IntDistrib opLengDist;			///< 1..32  byte string
 	static IntDistrib opMemrDist;			///< 1..10MB  byte string
 	static IntDistrib uniIntDist;			///< 0..0x7fffffff
+	static DescreteDistrib opCodeProbability; ///< An Opcode distribution with weights
```
pirapira: Let's try not to introduce a new static variable.

test/tools/libtesteth/TestHelper.h
pirapira: This function is used somewhere, no?

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: from where should it receve opCodeProbability ?

test/tools/fuzzTesting/fuzzHelper.h
winsvega: would it be better if I declare Random code as single tone?

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: the idea is that random generators are managed by random code class.

test/tools/libtesteth/TestHelper.h
pirapira: Before, `0` was success and `1` was failure.  Now, `false` is failure and `true` is success.

Remember in C++, `if (false)` and `if(0)` is very similar.  So, something is probably broken.

test/tools/fuzzTesting/fuzzHelper.cpp
pirapira: The caller.  Read my second comment as well, and choose one of my two suggestions (or something else of course).

test/tools/fuzzTesting/fuzzHelper.h
pirapira: tone?

winsvega: so like all of those fields would be private fields of a class.
and the only static instance would be an instance of that class

test/tools/libtesteth/TestHelper.h
winsvega: this one is only used in create random code. I always check that --createRandomCode option is working

pirapira: Does it check the return value of `createRandomTest()`?

test/tools/fuzzTesting/fuzzHelper.h
pirapira: First, try to remove this line.

test/tools/libtesteth/TestHelper.h
winsvega: hm. that changes are in my local repo. can't link yet.

test/tools/libtesteth/boostTest.cpp
```diff
@@ -90,37 +86,32 @@ void setDefaultOrCLocale()
...
-				if (arg == "--checktest")
 				{
-					argc = i + 1;
+					argv[i+1] = (char*)dynamicTestSuiteName.c_str();
 					break;
 				}
 			}
 
 			//add random tests suite
-			// FIXME:
```
pirapira: Do you remember what was the idea behind this `FIXME`?

test/tools/libtesteth/boostTest.cpp
```diff
@@ -90,37 +86,32 @@ void setDefaultOrCLocale()
...
 				//replace test suite to random tests
+				std::string arg = std::string{argv[i]};
 				if (arg == "-t" && i+1 < argc)
-					argv[i+1] = (char*)std::string("RandomTestCreationSuite").c_str();
-
-				//don't pass long raw test input to boost
-				if (arg == "--checktest")
 				{
-					argc = i + 1;
+					argv[i+1] = (char*)dynamicTestSuiteName.c_str();
```
pirapira: `i+1` should be `i + 1` with spaces.

pirapira: Before doing this assignment, make sure `argc > i + 1` holds.

winsvega: this is done one line up

test/tools/libtesteth/boostTest.cpp
winsvega: No. I don't think that is was my comment either.

test/tools/libtesteth/boostTest.cpp
pirapira: OK.  The next question: why is it necessary to edit the user input?

test/tools/libtesteth/boostTest.cpp
pirapira: Then leave it.

pirapira: Please recover the `FIXME`.

winsvega:
why? does it says anything about whats needed to be fixed? what is the point of returning it?

test/tools/libtesteth/boostTest.cpp
winsvega: this is a trick to overwrite boost option -t GeneralStateTests     with   -t randomTestCreationSuite which is dynamicly generated.

winsvega: so you type  -t GeneralStateTests -- --createRandomTest  
but instead it runs boost with -t randomTestCreationSuite 
which then reads what particular test to generate. in this case it is GeneralStateTest

pirapira:
I can read it like "random test suite should not be added here, this needs to be fixed.".

test/tools/libtesteth/boostTest.cpp
pirapira: Will you add a comment explaining these?

Issue #4301, opened 2017-08-02
Rename Guru Meditation into Bad Block?


pirapira:
Sometimes `cpp-ethereum` shows a highlighted `Guru Meditation`.  Maybe we should rewrite the message and just say `Bad Block`.

axic:
Done in #4540.

PR #4540, opened 2017-09-22
Rename bad block error message.

jacksonh:
Discussed in Issue:  #4301. This is a more explicit error message for bad blocks.


chfast:
Do you agree @pirapira ?

pirapira:
Yes!

PR #4543, opened 2017-09-22
additional mainnet difficulty tests fix

winsvega:


test/unittests/libethcore/difficulty.cpp
```diff
@@ -214,7 +214,7 @@ BOOST_AUTO_TEST_CASE(difficultyTestsCustomMainNetwork)
 
 	if (dev::test::Options::get().filltests)
 	{
-		u256 homesteadBlockNumber = 3500000;
+		u256 homesteadBlockNumber = 4300000;
```
pirapira: Why?

winsvega: ok i haven't mention that. but this blocknumber is byzantine transition number

winsvega: it used to be the test cases around homestead number. now this is around byzantine

pirapira: Please change the name of the variable.

test/unittests/libethcore/difficulty.cpp
```diff
@@ -42,14 +42,14 @@ std::string const c_testDifficulty = R"(
 		"currentDifficulty" : "[CDIFF]"
 	},
 )";
+h256 const someHash = sha3("whatever nonempty string");
```
pirapira: `nonzeroHash` is a better name, I think.

PR #4545, opened 2017-09-25
Request DAO fork block before syncing from the node

gumb0:
This helps to avoid being confused by Ethereum Classic nodes during sync with the main net.

Before starting to request block data from the node, we request DAO fork block header and check its extraData.

### DIFF
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -221,7 +221,24 @@ void BlockChainSync::onPeerStatus(std::shared_ptr<EthereumPeer> _peer)
 	else if (_peer->m_asking != Asking::State && _peer->m_asking != Asking::Nothing)
 		_peer->disable("Peer banned for unexpected status message.");
 	else
-		syncPeer(_peer, false);
+	{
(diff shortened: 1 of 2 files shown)


libethereum/BlockChainSync.cpp
```diff
@@ -522,6 +559,16 @@ void BlockChainSync::onPeerBlockHeaders(std::shared_ptr<EthereumPeer> _peer, RLP
...
 }
 
+bool BlockChainSync::verifyDaoChallengeResponse(RLP const& _r)
+{
+	if (_r.itemCount() != 1)
+		return false;
+
+	BlockHeader info(_r[0].data(), HeaderData);
+	return info.number() == host().chain().sealEngine()->chainParams().daoHardforkBlock &&
+		info.extraData() == fromHex("0x64616f2d686172642d666f726b");
```
pirapira: I think the block hash is harder to imitate.

gumb0: We don't know block hash here, having only header data...

extraData checking of DAO fork block is done also by geth, and also already by our code in `Ethash::verify()`

pirapira: Oh, I see.

PR #4549, opened 2017-09-25
Travis CI: Fix Hunter cache upload

chfast:
On Linux we have to upgrade requests Python package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -126,9 +126,13 @@ after_success:
     - make coverage.data
     # Get CodeCov script (fixed commit) and upload coverage report.
     - bash <(curl -s https://raw.githubusercontent.com/codecov/codecov-bash/d8016a07d67d813f98afdbaf98dc4f46d9806118/codecov) -X fix -f coverage.data
-after_script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -126,9 +126,13 @@ after_success:
     - make coverage.data
     # Get CodeCov script (fixed commit) and upload coverage report.
     - bash <(curl -s https://raw.githubusercontent.com/codecov/codecov-bash/d8016a07d67d813f98afdbaf98dc4f46d9806118/codecov) -X fix -f coverage.data
-after_script:
+after_script: |
```
pirapira: I don't know what this `|` means and I don't know how to search for this `|` in Google.

chfast: This is YAML literal style http://www.yaml.org/spec/1.2/spec.html#id2795688.
The way of having a multi-line string literal. All indented lines below are going to be a single literal.

There is also `>` variant, but I never remember what's the difference.

pirapira: Thank you for the enlightenment.

PR #4550, opened 2017-09-26
Update EVM-C API

chfast:
This also brings back the concurrent_queue needed for SmartVM build.


gumb0:
What is SmartVM?

chfast:
SmartVM is a dispatch class that executes EVM code in the interpreter or in the JIT.

PR #4551, opened 2017-09-26
CMake: Pull some changes from solidity

chfast:
In contains https://github.com/ethereum/solidity/pull/2946 changes.

### DIFF
--- a/cmake/EthCcache.cmake
+++ b/cmake/EthCcache.cmake
@@ -3,13 +3,13 @@
 # The ccache is auto-enabled if the tool is found.
 # To disable set -DCCACHE=OFF option.
 if(NOT DEFINED CMAKE_CXX_COMPILER_LAUNCHER)
-    find_program(CCACHE ccache DOC "ccache tool path; set to OFF to disable")
-    if(CCACHE)
-        set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE})
-        if(COMMAND cotire)
-            # Change ccache config to meet cotire requirements.
-            set(ENV{CCACHE_SLOPPINESS} pch_defines,time_macros)
-        endif()
-        message(STATUS "[ccache] Enabled: ${CCACHE}")
+  find_program(CCACHE ccache DOC "ccache tool path; set to OFF to disable")
+  if(CCACHE)
+    set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE})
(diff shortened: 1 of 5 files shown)


cmake/EthCcache.cmake
```diff
@@ -3,13 +3,13 @@
...
 if(NOT DEFINED CMAKE_CXX_COMPILER_LAUNCHER)
-    find_program(CCACHE ccache DOC "ccache tool path; set to OFF to disable")
-    if(CCACHE)
-        set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE})
-        if(COMMAND cotire)
-            # Change ccache config to meet cotire requirements.
-            set(ENV{CCACHE_SLOPPINESS} pch_defines,time_macros)
-        endif()
-        message(STATUS "[ccache] Enabled: ${CCACHE}")
+  find_program(CCACHE ccache DOC "ccache tool path; set to OFF to disable")
```
axic: There whitespace only changes.

cmake/EthCompilerSettings.cmake
```diff
@@ -41,75 +50,23 @@ if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MA
...
-		endif()
-
-		# Some Linux-specific Clang settings.  We don't want these for OS X.
-		if ("${CMAKE_SYSTEM_NAME}" MATCHES "Linux")
-
-			# Tell Boost that we're using Clang's libc++.   Not sure exactly why we need to do.
-			add_definitions(-DBOOST_ASIO_HAS_CLANG_LIBCXX)
-
-			# Use fancy colors in the compiler diagnostics
-			add_compile_options(-fcolor-diagnostics)
```
axic: Color output and `BOOST_ASIO_HAS_CLANG_LIBCXX` is removed for the linux case.

chfast: None of them are relevant any more.
1. We actually are using libstdc++ not libc++ on Linux.
2. Clang is using colors in terminal by default. That was for some old version.

cmake/EthCcache.cmake
chfast: Yes. This was accepted this way in solidity. I want this file to be identical to the one from solidity.

Using indenting of 2 spaces (as suggested by Hunter) was probably a mistake...

PR #4553, opened 2017-09-26
Get x_work when changing m_state

### LINKED ISSUE #4552: A deadlock was observed in "kademlia" test
https://travis-ci.org/ethereum/cpp-ethereum/jobs/279673618 timeouts with

```
Test Case "kademlia": 
.
.
.
.
.
```

pirapira:
Fixes #4552 

This fixes the deadlock in "kademlia" test.  The problem happned when the worker thread runs so quickly that when the main thread starts waiting for `m_state` change, the thread has already finished working.  This resulted in the main thread waiting forever for `m_state` to become `Started`.

### DIFF
--- a/libdevcore/Worker.cpp
+++ b/libdevcore/Worker.cpp
@@ -48,9 +48,13 @@ void Worker::startWorking()
 			while (m_state != WorkerState::Killing)
 			{
 				WorkerState ex = WorkerState::Starting;
-				bool ok = m_state.compare_exchange_strong(ex, WorkerState::Started);
+				{
+					// When somebody waits for this change on m_state, they'll give me the lock saying 'm_state_notifier.wait(l)'
+					Guard l(x_work);
+					bool ok = m_state.compare_exchange_strong(ex, WorkerState::Started);
+					(void)ok;
+				}
 //				cnote << "Trying to set Started: Thread was" << (unsigned)ex << "; " << ok;
-				(void)ok;
 				m_state_notifier.notify_all();
 
 				try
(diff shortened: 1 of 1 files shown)


pirapira:
I was confused about assignees and reviewers on this PR, it seems.

gumb0:
I've got the notification anyway

libdevcore/Worker.cpp
```diff
@@ -48,9 +48,13 @@ void Worker::startWorking()
 			while (m_state != WorkerState::Killing)
 			{
 				WorkerState ex = WorkerState::Starting;
-				bool ok = m_state.compare_exchange_strong(ex, WorkerState::Started);
+				{
+					// When somebody waits for this change on m_state, they'll give me the lock saying 'm_state_notifier.wait(l)'
+					Guard l(x_work);
```
gumb0: Let's use `std::unique_lock`, because it's used in other places of this class

libdevcore/Worker.cpp
```diff
@@ -48,9 +48,13 @@ void Worker::startWorking()
...
+					bool ok = m_state.compare_exchange_strong(ex, WorkerState::Started);
```
gumb0: I think we can just ignore the returned value instead of `(void)ok;`, no?

libdevcore/Worker.cpp
```diff
@@ -48,9 +48,13 @@ void Worker::startWorking()
 			while (m_state != WorkerState::Killing)
 			{
 				WorkerState ex = WorkerState::Starting;
-				bool ok = m_state.compare_exchange_strong(ex, WorkerState::Started);
+				{
+					// When somebody waits for this change on m_state, they'll give me the lock saying 'm_state_notifier.wait(l)'
```
gumb0: I'd just say that it's condition_variable-related mutex, maybe it's clearer...

libdevcore/Worker.cpp
pirapira: Should be the same thing.  Probably it's just fine to `m_state = Started;`.

libdevcore/Worker.cpp
pirapira: OK.

pirapira:
I rebased and took @gumb0's suggestions.

PR #4557, opened 2017-09-27
thread safe metric initialization

gcolvin:
Attempted fix of potential race condition initializing metrics structure.  Since no known failure has been observed it's probably not testable, so please review the two lines carefully.

### DIFF
--- a/libevm/VMOpt.cpp
+++ b/libevm/VMOpt.cpp
@@ -35,9 +35,8 @@ void VM::reportStackUse()
 std::array<InstructionMetric, 256> VM::c_metrics;
 void VM::initMetrics()
 {
-	// initialization of statics is thread safe, so exchange() will return false only once
-	static std::atomic<bool> done(false);
-	if (!done.exchange(true))
+	static bool done =
+	[]()
 	{
 		for (unsigned i = 0; i < 256; ++i)
 		{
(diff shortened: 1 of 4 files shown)


libevm/VMOpt.cpp
```diff
@@ -35,9 +35,9 @@ void VM::reportStackUse()
 std::array<InstructionMetric, 256> VM::c_metrics;
 void VM::initMetrics()
 {
-	// FIXME: This is not thread safe.
-	static bool done=false;
-	if (!done)
+	// initialization of statics is thread safe, so exchange() will return false only once
+	static std::atomic<bool> done(false);
+	if (!done.exchange(true))
```
chfast: I think this is worse now. If the first still writes to c_metrics the second will see that the flag is true, will skip the initialization part and use c_metrics partly initialized.

To do it in "easy" C++11 way, you can put all initialization into a lambada returning true, and assign this lambda invocation to a `static bool initialized` variable. C++11 guarantees thread-safe static initialization.

See this example: https://github.com/ethereum/cpp-ethereum/blob/develop/libdevcrypto/LibSnark.cpp#L40-L47

Looks hackish, but this allows the compiler to choose the best synchronization option available.

gcolvin: You are right, thanks. I wrote it with the lambda first (which is really cool and requires no library help) but then worried whether everything going into the initialization was itself statically initialized.  So I did something worse.  Chasing back the instructionInfo() function I see no problem, so I'll go back to my first approach.

gcolvin: Digging a little deeper I'm trying to clean up the unnecessarily cluttered initialization code.

gcolvin:
@chfast I think this ready.

PR #4564, opened 2017-09-27
Lighter headers

pirapira:
Mostly removing unused functions.

In `libethereum/BlockQueue.h`, a template variable `T` was used for closures, but I replaced it with some `std::function`.


libdevcore/concurrent_queue.h
```diff
@@ -1,60 +0,0 @@
...
-
-namespace dev
-{
-
-/// Concurrent queue.
-/// You can push and pop elements to/from the queue. Pop will block until the queue is not empty.
-/// The default backend (_QueueT) is std::queue. It can be changed to any type that has
-/// proper push(), pop(), empty() and front() methods.
-template<typename _T, typename _QueueT = std::queue<_T>>
-class concurrent_queue
```
chfast: This is not used anywhere?

pirapira: Never even included.

chfast: Sad to see it going away :(

chfast: This is needed by SmartVM.

PR #4566, opened 2017-09-28
[wip] Move libethcore/Transaction to libethcore/TransactionBase

pirapira:
because it's about class TransactionBase, not Transaction. There is also libethereum/Transaction.h


pirapira:
I also noticed a local variable that is never read.

libethcore/TransactionBase.cpp
```diff
@@ -46,28 +46,27 @@ TransactionBase::TransactionBase(TransactionSkeleton const& _ts, Secret const& _
 
 TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _checkSig)
 {
-	int field = 0;
-	RLP rlp(_rlpData);
+	RLP const rlp(_rlpData);
```
gumb0: `RLP` class is immutable, so usually I don't bother declaring it const

pirapira: I see.  I didn't know that.

chfast:
I'm not sure I didn't also want to merge Transaction from libethereum with this one. Or other way around.

pirapira:
I'll merge to `libethereum/Transaction.h`.

pirapira:
Merging them involves a big operation.  Either
* move `libethereum/Interaface` to `libethcore` or
* move `libethcore/SealEngine` to `libethereum` is necessary.

PR #4568, opened 2017-09-29
[WIP] tests on threads

winsvega:
an attempt to run test execution on threads

looks like json spirit is not thread safe.
getting weird errors from it.

https://bugs.launchpad.net/ubuntu/+source/json-spirit/+bug/1416679

### DIFF
--- a/eth/MinerAux.h
+++ b/eth/MinerAux.h
@@ -111,7 +111,7 @@ public:
 			catch (...)
 			{
 				cerr << "Bad " << arg << " option: " << argv[i] << endl;
-				BOOST_THROW_EXCEPTION(BadArgument());
+				ETH_THROW_EXCEPTION(BadArgument());
 			}
 		else if (arg == "--benchmark-trial" && i + 1 < argc)
 			try {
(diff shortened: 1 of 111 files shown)


winsvega:
ok. json spirit could be compiled with thread safe.
but it seems like boost is the issue:

http://www.boost.org/doc/libs/1_57_0/libs/test/doc/html/open-issues.html
https://stackoverflow.com/questions/26488051/can-boost-unit-testing-be-multithreaded

chfast:
@winsvega Wrapping all `throw`s with a mutex lock is no-go. We better replace Boost.Test with something else that is thread-safe than trying to workaround it.

winsvega:
I was just curious to see if it solves the problem. it didn't

however we have some tests which rely on ETH_CHECK
those ctest  Client.Base tests are using ETH_CHECK and working.

Maybe it is good to wrap CHECKS afterall? we could then switch from boost check to something changing only one file.

chfast:
@winsvega yes, I don't see a problem with wrapping BOOST_CHECK if we need it.
Personally, I'd use name `CHECK` instead of `ETH_CHECK`. I don't think we need to prefix macros. We are not building general purpose library.

winsvega:
lets keep this branch in case we ever decide to switch from boost. 
for now I guess I will use ctest to run tests on threads.

winsvega:
implemented in retesteth

PR #4569, opened 2017-10-02
Shift opcodes

gcolvin:
SHL and SHR are well tested by test/unittests/performance/rc5.sol.  SAR is only lightly tested.
The `#define EIP_145 true` in VMConfig.h will need to be turned off once all tests pass here.


axic:
Thanks @gcolvin! Would it be possible to merge in the changes into #4054 ?

gcolvin:
@axic @chfast  It appears we have been working independently on these opcodes here and in https://github.com/ethereum/cpp-ethereum/pull/4054, and that we have come up with different implementations.  So one or the other of these should be merged, but not both.  If I don't here from you today perhaps we can discuss it after the staff meeting tomorrow.

gcolvin:
Mostly redundant with #4054, differences to be kept should be merged there.

libevm/VM.cpp
```diff
@@ -621,6 +621,70 @@ void VM::interpretCases()
...
+				if (shiftee & hibit)
+					m_SPP[0] = allbits;
+				else
+					m_SPP[0] = 0;
+			}
+			else
+			{
+				uint64_t amount = uint64_t(m_SP[0]);
+			    m_SPP[0] = shiftee >> amount;
+			    m_SPP[0] |= allbits << (256 - amount);
```
chfast: I think you have to check `shiftee & hibit` here as well.

PR #4579, opened 2017-10-05
Remove ThreadSanitizer warnings 

### LINKED ISSUE #4544: A ThreadSanitizer warning in the `eth` command
On version 3d6eb1fb26ea5c93c66bb380070f5347ec78974f, `eth/eth` shows the following warning

```
==================
WARNING: ThreadSanitizer: data race on vptr (ctor/dtor vs virtual call) (pid=11157)
  Write of size 8 at 0x7ba400000000 by main thread:
    #0 EthashClient /home/yh/src/cpp-ethereum/libethashseal/EthashClient.cpp:57 (discriminator 1) (eth+0x774982)
    #1 WebThreeDirect /home/yh/src/cpp-ethereum/libwebthree/WebThree.cpp:59 (discriminator 1) (eth+0x7ed672)
    #2 main /home/yh/src/cpp-ethereum/eth/main.cpp:929 (discriminator 1) (eth+0x502e7f)

  Previous read of size 8 at 0x7ba400000000 by thread T5:
    #0 dev::eth::Client::rejigSealing() /home/yh/src/cpp-ethereum/libethereum/Client.cpp:580 (eth+0x66b592)
    #1 dev::eth::Client::doWork(bool) /home/yh/src/cpp-ethereum/libethereum/Client.cpp:669 (eth+0x667c56)
    #2 dev::eth::Client::doWork() /home/yh/src/cpp-ethereum/libethereum/Client.h:237 (eth+0x67537e)
    #3 non-virtual thunk to dev::eth::Client::doWork() /home/yh/src/cpp-ethereum/libdevcore/../libethereum/Client.h:? (eth+0x6753c0)
    #4 dev::Worker::workLoop() /home/yh/src/cpp-ethereum/libdevcore/Worker.cpp:131 (eth+0x80ff30)
    #5 operator() /home/yh/src/cpp-ethereum/libdevcore/Worker.cpp:59 (eth+0x81045f)
    #6 void std::_Bind_simple<dev::Worker::startWorking()::$_0 ()>::_M_invoke<>(std::_Index_tuple<>) /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/functional:1390 (discriminator 2) (eth+0x8102d9)
    #7 std::_Bind_simple<dev::Worker::startWorking()::$_0 ()>::operator()() /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/functional:1380 (eth+0x810289)
    #8 std::thread::_State_impl<std::_Bind_simple<dev::Worker::startWorking()::$_0 ()> >::_M_run() /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/thread:197 (eth+0x81014d)
    #9 std::error_code::default_error_condition() const ??:? (libstdc++.so.6+0xbb83e)

  Location is heap block of size 20448 at 0x7ba400000000 allocated by main thread:
    #0 operator new(unsigned long) ??:? (eth+0x4edd73)
    #1 WebThreeDirect /home/yh/src/cpp-ethereum/libwebthree/WebThree.cpp:59 (eth+0x7ed609)
    #2 main /home/yh/src/cpp-ethereum/eth/main.cpp:929 (discriminator 1) (eth+0x502e7f)

  Thread T5 'eth' (tid=11440, running) created by main thread at:
    #0 pthread_create ??:? (eth+0x462346)
    #1 std::thread::_M_start_thread(std::unique_ptr<std::thread::_State, std::default_delete<std::thread::_State> >, void (*)()) ??:? (libstdc++.so.6+0xbbb54)
    #2 dev::Worker::startWorking() /home/yh/src/cpp-ethereum/libdevcore/Worker.cpp:44 (discriminator 1) (eth+0x80f900)
    #3 dev::eth::Client::init(dev::p2p::Host*, boost::filesystem::path const&, dev::WithExisting, boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<256u, 256u, (boost::multiprecision::cpp_integer_type)0, (boost::multiprecision::cpp_int_check_type)0, void>, (boost::multiprecision::expression_template_option)0>) /home/yh/src/cpp-ethereum/libethereum/Client.cpp:126 (eth+0x6674bb)
    #4 Client /home/yh/src/cpp-ethereum/libethereum/Client.cpp:78 (eth+0x666e49)
    #5 EthashClient /home/yh/src/cpp-ethereum/libethashseal/EthashClient.cpp:56 (eth+0x774968)
    #6 WebThreeDirect /home/yh/src/cpp-ethereum/libwebthree/WebThree.cpp:59 (discriminator 1) (eth+0x7ed672)
    #7 main /home/yh/src/cpp-ethereum/eth/main.cpp:929 (discriminator 1) (eth+0x502e7f)

SUMMARY: ThreadSanitizer: data race on vptr (ctor/dtor vs virtual call) /home/yh/src/cpp-ethereum/libethashseal/EthashClient.cpp:57 (discriminator 1) in EthashClient
==================
```

pirapira:
This PR removes various ThreadSanitizer warnings.  Each commit deals with one or two warnings.

Fixes #4544.

### DIFF
--- a/libethereum/Client.h
+++ b/libethereum/Client.h
@@ -320,6 +320,9 @@ protected:
 
 	std::weak_ptr<EthereumHost> m_host;		///< Our Ethereum Host. Don't do anything if we can't lock.
 
+	std::condition_variable m_signalled;
+	Mutex x_signalled;
+
 	Handler<> m_tqReady;
 	Handler<h256 const&> m_tqReplaced;
 	Handler<> m_bqReady;
@@ -339,8 +342,6 @@ protected:
 	SharedMutex x_functionQueue;
 	std::queue<std::function<void()>> m_functionQueue;	///< Functions waiting to be executed in the main thread.
 
-	std::condition_variable m_signalled;
-	Mutex x_signalled;
 	std::atomic<bool> m_syncTransactionQueue = {false};
 	std::atomic<bool> m_syncBlockQueue = {false};
 
--- a/libethereum/EthereumHost.cpp
+++ b/libethereum/EthereumHost.cpp
@@ -54,19 +54,19 @@ namespace
 class EthereumPeerObserver: public EthereumPeerObserverFace
 {
 public:
-	EthereumPeerObserver(BlockChainSync& _sync, TransactionQueue& _tq): m_sync(_sync), m_tq(_tq) {}
+	EthereumPeerObserver(shared_ptr<BlockChainSync> _sync, TransactionQueue& _tq): m_sync(_sync), m_tq(_tq) {}
 
 	void onPeerStatus(std::shared_ptr<EthereumPeer> _peer) override
 	{
 		try
 		{
-			m_sync.onPeerStatus(_peer);
+			m_sync->onPeerStatus(_peer);
 		}
 		catch (FailedInvariant const&)
 		{
 			// "fix" for https://github.com/ethereum/webthree-umbrella/issues/300
 			clog(NetWarn) << "Failed invariant during sync, restarting sync";
-			m_sync.restartSync();
+			m_sync->restartSync();
 		}
 	}
 
@@ -81,7 +81,7 @@ public:
 	{
 		try
 		{
-			m_sync.onPeerAborting();
+			m_sync->onPeerAborting();
 		}
 		catch (Exception&)
 		{
@@ -93,13 +93,13 @@ public:
 	{
 		try
 		{
-			m_sync.onPeerBlockHeaders(_peer, _headers);
+			m_sync->onPeerBlockHeaders(_peer, _headers);
 		}
 		catch (FailedInvariant const&)
 		{
 			// "fix" for https://github.com/ethereum/webthree-umbrella/issues/300
 			clog(NetWarn) << "Failed invariant during sync, restarting sync";
-			m_sync.restartSync();
+			m_sync->restartSync();
 		}
 	}
 
@@ -107,13 +107,13 @@ public:
 	{
 		try
 		{
-			m_sync.onPeerBlockBodies(_peer, _r);
+			m_sync->onPeerBlockBodies(_peer, _r);
 		}
 		catch (FailedInvariant const&)
 		{
 			// "fix" for https://github.com/ethereum/webthree-umbrella/issues/300
 			clog(NetWarn) << "Failed invariant during sync, restarting sync";
-			m_sync.restartSync();
+			m_sync->restartSync();
 		}
 	}
 
@@ -121,13 +121,13 @@ public:
 	{
 		try
 		{
-			m_sync.onPeerNewHashes(_peer, _hashes);
+			m_sync->onPeerNewHashes(_peer, _hashes);
 		}
 		catch (FailedInvariant const&)
 		{
 			// "fix" for https://github.com/ethereum/webthree-umbrella/issues/300
 			clog(NetWarn) << "Failed invariant during sync, restarting sync";
-			m_sync.restartSync();
+			m_sync->restartSync();
 		}
 	}
 
@@ -135,13 +135,13 @@ public:
 	{
 		try
 		{
-			m_sync.onPeerNewBlock(_peer, _r);
+			m_sync->onPeerNewBlock(_peer, _r);
 		}
 		catch (FailedInvariant const&)
 		{
 			// "fix" for https://github.com/ethereum/webthree-umbrella/issues/300
 			clog(NetWarn) << "Failed invariant during sync, restarting sync";
-			m_sync.restartSync();
+			m_sync->restartSync();
 		}
 	}
 
@@ -158,7 +158,7 @@ public:
 	}
 
 private:
-	BlockChainSync& m_sync;
+	shared_ptr<BlockChainSync> m_sync;
 	TransactionQueue& m_tq;
 };
 
@@ -380,7 +380,7 @@ EthereumHost::EthereumHost(BlockChain const& _ch, OverlayDB const& _db, Transact
 	// TODO: Composition would be better. Left like that to avoid initialization
 	//       issues as BlockChainSync accesses other EthereumHost members.
 	m_sync.reset(new BlockChainSync(*this));
-	m_peerObserver = make_shared<EthereumPeerObserver>(*m_sync, m_tq);
+	m_peerObserver = make_shared<EthereumPeerObserver>(m_sync, m_tq);
 	m_latestBlockSent = _ch.currentHash();
 	m_tq.onImport([this](ImportResult _ir, h256 const& _h, h512 const& _nodeId) { onTransactionImported(_ir, _h, _nodeId); });
 }
--- a/libethereum/EthereumPeer.cpp
+++ b/libethereum/EthereumPeer.cpp
@@ -109,8 +109,8 @@ void EthereumPeer::setRude()
 
 void EthereumPeer::abortSync()
 {
-	if (m_observer)
-		m_observer->onPeerAborting();
+	if (auto observer = m_observer.lock())
+		observer->onPeerAborting();
 }
 
 
@@ -236,7 +236,10 @@ bool EthereumPeer::isCriticalSyncing() const
 
 bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 {
-	assert(m_observer);
+	auto observer = m_observer.lock();
+	auto hostData = m_hostData.lock();
+	if (!observer || !hostData)
+		return false;
 
 	m_lastAsk = std::chrono::system_clock::to_time_t(chrono::system_clock::now());
 	try
@@ -255,12 +258,12 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 
 		clog(NetMessageSummary) << "Status:" << m_protocolVersion << "/" << m_networkId << "/" << m_genesisHash << ", TD:" << m_totalDifficulty << "=" << m_latestHash;
 		setIdle();
-		m_observer->onPeerStatus(dynamic_pointer_cast<EthereumPeer>(dynamic_pointer_cast<EthereumPeer>(shared_from_this())));
+		observer->onPeerStatus(dynamic_pointer_cast<EthereumPeer>(dynamic_pointer_cast<EthereumPeer>(shared_from_this())));
 		break;
 	}
 	case TransactionsPacket:
 	{
-		m_observer->onPeerTransactions(dynamic_pointer_cast<EthereumPeer>(dynamic_pointer_cast<EthereumPeer>(shared_from_this())), _r);
+		observer->onPeerTransactions(dynamic_pointer_cast<EthereumPeer>(dynamic_pointer_cast<EthereumPeer>(shared_from_this())), _r);
 		break;
 	}
 	case GetBlockHeadersPacket:
@@ -280,7 +283,7 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 			break;
 		}
 
-		pair<bytes, unsigned> const rlpAndItemCount = m_hostData->blockHeaders(blockId, numHeadersToSend, skip, reverse);
+		pair<bytes, unsigned> const rlpAndItemCount = hostData->blockHeaders(blockId, numHeadersToSend, skip, reverse);
 
 		RLPStream s;
 		prep(s, BlockHeadersPacket, rlpAndItemCount.second).appendRaw(rlpAndItemCount.first, rlpAndItemCount.second);
@@ -295,7 +298,7 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 		else
 		{
 			setIdle();
-			m_observer->onPeerBlockHeaders(dynamic_pointer_cast<EthereumPeer>(shared_from_this()), _r);
+			observer->onPeerBlockHeaders(dynamic_pointer_cast<EthereumPeer>(shared_from_this()), _r);
 		}
 		break;
 	}
@@ -311,7 +314,7 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 			break;
 		}
 
-		pair<bytes, unsigned> const rlpAndItemCount = m_hostData->blockBodies(_r);
+		pair<bytes, unsigned> const rlpAndItemCount = hostData->blockBodies(_r);
 
 		addRating(0);
 		RLPStream s;
@@ -326,13 +329,13 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 		else
 		{
 			setIdle();
-			m_observer->onPeerBlockBodies(dynamic_pointer_cast<EthereumPeer>(shared_from_this()), _r);
+			observer->onPeerBlockBodies(dynamic_pointer_cast<EthereumPeer>(shared_from_this()), _r);
 		}
 		break;
 	}
 	case NewBlockPacket:
 	{
-		m_observer->onPeerNewBlock(dynamic_pointer_cast<EthereumPeer>(shared_from_this()), _r);
+		observer->onPeerNewBlock(dynamic_pointer_cast<EthereumPeer>(shared_from_this()), _r);
 		break;
 	}
 	case NewBlockHashesPacket:
@@ -351,7 +354,7 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 		for (unsigned i = 0; i < itemCount; ++i)
 			hashes[i] = std::make_pair(_r[i][0].toHash<h256>(), _r[i][1].toInt<u256>());
 
-		m_observer->onPeerNewHashes(dynamic_pointer_cast<EthereumPeer>(shared_from_this()), hashes);
+		observer->onPeerNewHashes(dynamic_pointer_cast<EthereumPeer>(shared_from_this()), hashes);
 		break;
 	}
 	case GetNodeDataPacket:
@@ -365,7 +368,7 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 		}
 		clog(NetMessageSummary) << "GetNodeData (" << dec << count << " entries)";
 
-		strings const data = m_hostData->nodeData(_r);
+		strings const data = hostData->nodeData(_r);
 
 		addRating(0);
 		RLPStream s;
@@ -386,7 +389,7 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 		}
 		clog(NetMessageSummary) << "GetReceipts (" << dec << count << " entries)";
 
-		pair<bytes, unsigned> const rlpAndItemCount = m_hostData->receipts(_r);
+		pair<bytes, unsigned> const rlpAndItemCount = hostData->receipts(_r);
 
 		addRating(0);
 		RLPStream s;
@@ -401,7 +404,7 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 		else
 		{
 			setIdle();
-			m_observer->onPeerNodeData(dynamic_pointer_cast<EthereumPeer>(shared_from_this()), _r);
+			observer->onPeerNodeData(dynamic_pointer_cast<EthereumPeer>(shared_from_this()), _r);
 		}
 		break;
 	}
@@ -412,7 +415,7 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 		else
 		{
 			setIdle();
-			m_observer->onPeerReceipts(dynamic_pointer_cast<EthereumPeer>(shared_from_this()), _r);
+			observer->onPeerReceipts(dynamic_pointer_cast<EthereumPeer>(shared_from_this()), _r);
 		}
 		break;
 	}
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -149,8 +149,15 @@ void Host::stop()
 		return;
 
 	// wait for m_timer to reset (indicating network scheduler has stopped)
-	while (!!m_timer)
-		this_thread::sleep_for(chrono::milliseconds(50));
+	{
+		std::unique_lock<std::mutex> l{x_runTimer};
+		while (!!m_timer)
+		{
+			l.unlock();
+			this_thread::sleep_for(chrono::milliseconds(50));
+			l.lock();
+		}
+	}
 
 	// stop worker thread
 	if (isWorking())
@@ -659,7 +666,8 @@ void Host::run(boost::system::error_code const&)
 		m_ioService.stop();
 
 		// resetting timer signals network that nothing else can be scheduled to run
-		m_timer.reset();
+		DEV_GUARDED(x_runTimer)
+			m_timer.reset();
 		return;
 	}
 
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -835,7 +835,7 @@ bytes Host::saveNetwork() const
 			network << p.id << (p.peerType == PeerType::Required ? true : false)
 				<< chrono::duration_cast<chrono::seconds>(p.m_lastConnected.time_since_epoch()).count()
 				<< chrono::duration_cast<chrono::seconds>(p.m_lastAttempted.time_since_epoch()).count()
-				<< p.m_failedAttempts << (unsigned)p.m_lastDisconnect << p.m_score << p.m_rating;
+				<< p.m_failedAttempts << (unsigned)p.m_lastDisconnect << p.m_score.load() << p.m_rating.load();
 			count++;
 		}
 	}
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -249,8 +249,8 @@ void Host::startPeerSession(Public const& _id, RLP const& _rlp, unique_ptr<RLPXF
 		else
 		{
 			// peer doesn't exist, try to get port info from node table
-			if (m_nodeTable)
-				if (Node n = m_nodeTable->node(_id))
+			if (auto nodeTable = this->nodeTable())
+				if (Node n = nodeTable->node(_id))
 					p = make_shared<Peer>(n);
 
 			if (!p)
@@ -350,8 +350,10 @@ void Host::onNodeTableEvent(NodeID const& _n, NodeTableEventType const& _e)
 	if (_e == NodeEntryAdded)
 	{
 		clog(NetP2PNote) << "p2p.host.nodeTable.events.nodeEntryAdded " << _n;
-		// only add iff node is in node table
-		if (Node n = m_nodeTable->node(_n))
+		auto nodeTable = this->nodeTable();
+		if (!nodeTable)
+			return;
+		if (Node n = nodeTable->node(_n))
 		{
 			shared_ptr<Peer> p;
 			DEV_RECURSIVE_GUARDED(x_sessions)
@@ -514,8 +516,8 @@ void Host::addNode(NodeID const& _node, NodeIPEndpoint const& _endpoint)
 	if (_endpoint.tcpPort < 30300 || _endpoint.tcpPort > 30305)
 		clog(NetConnect) << "Non-standard port being recorded: " << _endpoint.tcpPort;
 
-	if (m_nodeTable)
-		m_nodeTable->addNode(Node(_node, _endpoint));
+	if (auto nodeTable = this->nodeTable())
+		nodeTable->addNode(Node(_node, _endpoint));
 }
 
 void Host::requirePeer(NodeID const& _n, NodeIPEndpoint const& _endpoint)
@@ -546,19 +548,19 @@ void Host::requirePeer(NodeID const& _n, NodeIPEndpoint const& _endpoint)
 				m_peers[_n] = p;
 			}
 		// required for discovery
-		if (m_nodeTable)
-			m_nodeTable->addNode(*p, NodeTable::NodeRelation::Unknown);
+		if (auto nodeTable = this->nodeTable())
+			nodeTable->addNode(*p, NodeTable::NodeRelation::Unknown);
 	}
-	else if (m_nodeTable)
+	else if (auto nodeTable = this->nodeTable())
 	{
-		m_nodeTable->addNode(node);
+		nodeTable->addNode(node);
 		auto t = make_shared<boost::asio::deadline_timer>(m_ioService);
 		t->expires_from_now(boost::posix_time::milliseconds(600));
 		t->async_wait([this, _n](boost::system::error_code const& _ec)
 		{
 			if (!_ec)
-				if (m_nodeTable)
-					if (auto n = m_nodeTable->node(_n))
+				if (auto nodeTable = this->nodeTable())
+					if (auto n = nodeTable->node(_n))
 						requirePeer(n.id, n.endpoint);
 		});
 		DEV_GUARDED(x_timers)
@@ -584,8 +586,9 @@ void Host::connect(std::shared_ptr<Peer> const& _p)
 		return;
 	}
 
-	if (!!m_nodeTable && !m_nodeTable->haveNode(_p->id) && _p->peerType == PeerType::Optional)
-		return;
+	if (auto nodeTable = this->nodeTable())
+		if (!nodeTable->haveNode(_p->id) && _p->peerType == PeerType::Optional)
+			return;
 
 	// prevent concurrently connecting to a node
 	Peer *nptr = _p.get();
@@ -659,7 +662,8 @@ void Host::run(boost::system::error_code const&)
 	if (!m_run)
 	{
 		// reset NodeTable
-		m_nodeTable.reset();
+		DEV_GUARDED(x_nodeTable)
+			m_nodeTable.reset();
 
 		// stopping io service allows running manual network operations for shutdown
 		// and also stops blocking worker thread, allowing worker thread to exit
@@ -671,7 +675,8 @@ void Host::run(boost::system::error_code const&)
 		return;
 	}
 
-	m_nodeTable->processEvents();
+	if (auto nodeTable = this->nodeTable()) // This again requires x_nodeTable, which is why an additional variable nodeTable is used.
+		nodeTable->processEvents();
 
 	// cleanup zombies
 	DEV_GUARDED(x_connecting)
@@ -763,7 +768,8 @@ void Host::startedWorking()
 		m_netPrefs.discovery
 	);
 	nodeTable->setEventHandler(new HostNodeTableHandler(*this));
-	m_nodeTable = nodeTable;
+	DEV_GUARDED(x_nodeTable)
+		m_nodeTable = nodeTable;
 	restoreNetwork(&m_restoreNetwork);
 
 	clog(NetP2PNote) << "p2p.started id:" << id();
@@ -848,9 +854,9 @@ bytes Host::saveNetwork() const
 		}
 	}
 
-	if (!!m_nodeTable)
+	if (auto nodeTable = this->nodeTable())
 	{
-		auto state = m_nodeTable->snapshot();
+		auto state = nodeTable->snapshot();
 		state.sort();
 		for (auto const& entry: state)
 		{
@@ -901,7 +907,10 @@ void Host::restoreNetwork(bytesConstRef _b)
 			{
 				Node n((NodeID)i[3], NodeIPEndpoint(i));
 				if (i.itemCount() == 4 && n.endpoint.isAllowed())
-					m_nodeTable->addNode(n);
+				{
+					if (auto nodeTable = this->nodeTable())
+						nodeTable->addNode(n);
+				}
 				else if (i.itemCount() == 11)
 				{
 					n.peerType = i[4].toInt<bool>() ? PeerType::Required : PeerType::Optional;
@@ -917,15 +926,16 @@ void Host::restoreNetwork(bytesConstRef _b)
 					m_peers[p->id] = p;
 					if (p->peerType == PeerType::Required)
 						requirePeer(p->id, n.endpoint);
-					else
-						m_nodeTable->addNode(*p.get(), NodeTable::NodeRelation::Known);
+					else if (auto nodeTable = this->nodeTable())
+						nodeTable->addNode(*p.get(), NodeTable::NodeRelation::Known);
 				}
 			}
 			else if (i.itemCount() == 3 || i.itemCount() == 10)
 			{
 				Node n((NodeID)i[2], NodeIPEndpoint(bi::address_v4(i[0].toArray<byte, 4>()), i[1].toInt<uint16_t>(), i[1].toInt<uint16_t>()));
 				if (i.itemCount() == 3 && n.endpoint.isAllowed())
-					m_nodeTable->addNode(n);
+					DEV_GUARDED(x_nodeTable)
+						m_nodeTable->addNode(n);
 				else if (i.itemCount() == 10)
 				{
 					n.peerType = i[3].toInt<bool>() ? PeerType::Required : PeerType::Optional;
@@ -941,8 +951,8 @@ void Host::restoreNetwork(bytesConstRef _b)
 					m_peers[p->id] = p;
 					if (p->peerType == PeerType::Required)
 						requirePeer(p->id, n.endpoint);
-					else
-						m_nodeTable->addNode(*p.get(), NodeTable::NodeRelation::Known);
+					else if (auto nodeTable = this->nodeTable())
+						nodeTable->addNode(*p.get(), NodeTable::NodeRelation::Known);
 				}
 			}
 		}
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -142,21 +142,16 @@ void Host::stop()
 	// called to force io_service to kill any remaining tasks it might have -
 	// such tasks may involve socket reads from Capabilities that maintain references
 	// to resources we're about to free.
-
-	// ignore if already stopped/stopping, at the same time,
-	// signal run() to prepare for shutdown and reset m_timer
-	if (!m_run.exchange(false))
-		return;
-
-	// wait for m_timer to reset (indicating network scheduler has stopped)
+	
 	{
-		std::unique_lock<std::mutex> l{x_runTimer};
-		while (!!m_timer)
-		{
-			l.unlock();
-			this_thread::sleep_for(chrono::milliseconds(50));
-			l.lock();
-		}
+		unique_lock<mutex> l(x_runTimer);
+		// ignore if already stopped/stopping, at the same time,
+		// signal run() to prepare for shutdown and reset m_timer
+		if (!m_run.exchange(false))
+			return;
+
+		while (m_timer)
+			m_timerSemaphore.wait(l);
 	}
 
 	// stop worker thread
@@ -672,6 +667,8 @@ void Host::run(boost::system::error_code const&)
 		// resetting timer signals network that nothing else can be scheduled to run
 		DEV_GUARDED(x_runTimer)
 			m_timer.reset();
+
+		m_timerSemaphore.notify_all();
 		return;
 	}
 
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -846,7 +846,7 @@ bytes Host::saveNetwork() const
 			network << p.id << (p.peerType == PeerType::Required ? true : false)
 				<< chrono::duration_cast<chrono::seconds>(p.m_lastConnected.time_since_epoch()).count()
 				<< chrono::duration_cast<chrono::seconds>(p.m_lastAttempted.time_since_epoch()).count()
-				<< p.m_failedAttempts << (unsigned)p.m_lastDisconnect << p.m_score.load() << p.m_rating.load();
+				<< p.m_failedAttempts.load() << (unsigned)p.m_lastDisconnect << p.m_score.load() << p.m_rating.load();
 			count++;
 		}
 	}
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -244,9 +244,8 @@ void Host::startPeerSession(Public const& _id, RLP const& _rlp, unique_ptr<RLPXF
 		else
 		{
 			// peer doesn't exist, try to get port info from node table
-			if (auto nodeTable = this->nodeTable())
-				if (Node n = nodeTable->node(_id))
-					p = make_shared<Peer>(n);
+			if (Node n = nodeFromNodeTable(_id))
+				p = make_shared<Peer>(n);
 
 			if (!p)
 				p = make_shared<Peer>(Node(_id, UnspecifiedNodeIPEndpoint));
@@ -345,10 +344,7 @@ void Host::onNodeTableEvent(NodeID const& _n, NodeTableEventType const& _e)
 	if (_e == NodeEntryAdded)
 	{
 		clog(NetP2PNote) << "p2p.host.nodeTable.events.nodeEntryAdded " << _n;
-		auto nodeTable = this->nodeTable();
-		if (!nodeTable)
-			return;
-		if (Node n = nodeTable->node(_n))
+		if (Node n = nodeFromNodeTable(_n))
 		{
 			shared_ptr<Peer> p;
 			DEV_RECURSIVE_GUARDED(x_sessions)
@@ -511,8 +507,7 @@ void Host::addNode(NodeID const& _node, NodeIPEndpoint const& _endpoint)
 	if (_endpoint.tcpPort < 30300 || _endpoint.tcpPort > 30305)
 		clog(NetConnect) << "Non-standard port being recorded: " << _endpoint.tcpPort;
 
-	if (auto nodeTable = this->nodeTable())
-		nodeTable->addNode(Node(_node, _endpoint));
+	addNodeToNodeTable(Node(_node, _endpoint));
 }
 
 void Host::requirePeer(NodeID const& _n, NodeIPEndpoint const& _endpoint)
@@ -543,20 +538,19 @@ void Host::requirePeer(NodeID const& _n, NodeIPEndpoint const& _endpoint)
 				m_peers[_n] = p;
 			}
 		// required for discovery
-		if (auto nodeTable = this->nodeTable())
-			nodeTable->addNode(*p, NodeTable::NodeRelation::Unknown);
+		addNodeToNodeTable(*p, NodeTable::NodeRelation::Unknown);
 	}
-	else if (auto nodeTable = this->nodeTable())
+	else
 	{
-		nodeTable->addNode(node);
+		if (!addNodeToNodeTable(node))
+			return;
 		auto t = make_shared<boost::asio::deadline_timer>(m_ioService);
 		t->expires_from_now(boost::posix_time::milliseconds(600));
 		t->async_wait([this, _n](boost::system::error_code const& _ec)
 		{
 			if (!_ec)
-				if (auto nodeTable = this->nodeTable())
-					if (auto n = nodeTable->node(_n))
-						requirePeer(n.id, n.endpoint);
+				if (auto n = nodeFromNodeTable(_n))
+					requirePeer(n.id, n.endpoint);
 		});
 		DEV_GUARDED(x_timers)
 			m_timers.push_back(t);
@@ -581,9 +575,8 @@ void Host::connect(std::shared_ptr<Peer> const& _p)
 		return;
 	}
 
-	if (auto nodeTable = this->nodeTable())
-		if (!nodeTable->haveNode(_p->id) && _p->peerType == PeerType::Optional)
-			return;
+	if (!nodeTableHasNode(_p->id) && _p->peerType == PeerType::Optional)
+		return;
 
 	// prevent concurrently connecting to a node
 	Peer *nptr = _p.get();
@@ -905,8 +898,7 @@ void Host::restoreNetwork(bytesConstRef _b)
 				Node n((NodeID)i[3], NodeIPEndpoint(i));
 				if (i.itemCount() == 4 && n.endpoint.isAllowed())
 				{
-					if (auto nodeTable = this->nodeTable())
-						nodeTable->addNode(n);
+					addNodeToNodeTable(n);
 				}
 				else if (i.itemCount() == 11)
 				{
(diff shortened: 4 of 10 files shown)


libdevcore/Worker.cpp
```diff
@@ -118,7 +118,8 @@ void Worker::terminate()
 	std::unique_lock<Mutex> l(x_work);
 	if (m_work)
 	{
-		m_state.exchange(WorkerState::Killing);
+		if (m_state.exchange(WorkerState::Killing) == WorkerState::Killing)
```
chfast: Check also stopWorking() - it has similar pattern.

pirapira: I added a similar change there.

pirapira:
AppVeyor is stuck at "updateStats" test.  I'll try it with ThreadSanitizer.

pirapira:
I cannot reproduce the AppVeyor issue.

libethashseal/EthashClient.cpp
```diff
@@ -59,6 +59,11 @@ EthashClient::EthashClient(
 	asEthashClient(*this);
 }
 
+EthashClient::~EthashClient()
+{
+	terminate();
```
gumb0: So it will be called twice - once in `~EthashClient`, then in `~Client` ?

libethereum/Client.h
```diff
@@ -198,6 +198,9 @@ class Client: public ClientBase, protected Worker
 	virtual Block block(h256 const& _block) const override;
 	using ClientBase::block;
 
+	/// should be called after the constructor of the most derived class finishes.
+	void afterConstructor() { startWorking(); };
```
gumb0: I would call this method `startWorking()`

libethereum/EthereumHost.cpp
```diff
@@ -395,6 +412,8 @@ EthereumHost::EthereumHost(BlockChain const& _ch, OverlayDB const& _db, Transact
 EthereumHost::~EthereumHost()
 {
 	terminate();
+	m_peerObserver->disableSync(); // m_peerObserver contains a reference to BlockchainSync, so that should be disabled first.
```
gumb0: Can we instead just do `m_peerObserver.reset()` ?

libethereum/ClientBase.h
```diff
@@ -188,6 +188,8 @@ class ClientBase: public Interface
 	std::unordered_map<h256, h256s> m_specialFilters = std::unordered_map<h256, std::vector<h256>>{{PendingChangedFilter, {}}, {ChainChangedFilter, {}}};
 															///< The dictionary of special filters and their additional data
 	std::map<unsigned, ClientWatch> m_watches;				///< Each and every watch - these reference a filter.
+	std::condition_variable m_signalled;
```
gumb0: How does this move help?

libethereum/Client.h
pirapira: OK, I'll try that.

libethashseal/EthashClient.cpp
pirapira: Yes, which is not a problem.  I'll comment about that.

libethereum/EthereumHost.cpp
gumb0: Ah I see, probably not, because peers still have shared_ptr to observer.

But then - can we make `BlockChainSync` member of `EthereumPeerObserver` to be shared_ptr ? This should solve the problem

pirapira: I'll try that.

libethereum/ClientBase.h
pirapira: `ClientBase::m_tq` uses this condition variable.

pirapira: This was the warning:

```
  Write of size 8 at 0x7ba400004df8 by main thread:
    #0 pthread_cond_destroy ??:? (eth+0x48d790) // cond destroy.
    #1 ~Client /home/yh/src/cpp-ethereum2/libethereum/Client.cpp:85 (discriminator 3) (eth+0x695d80)
    #2 ~EthashClient /home/yh/src/cpp-ethereum2/libethashseal/EthashClient.cpp:65 (discriminator 4) (eth+0x7b112d)
    #3 ~EthashClient /home/yh/src/cpp-ethereum2/libethashseal/EthashClient.cpp:63 (eth+0x7b11b9)
    #4 std::default_delete<dev::eth::Client>::operator()(dev::eth::Client*) const /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/bits/unique_ptr.h:76 (discriminator 1) (eth+0x82e62f)
    #5 std::unique_ptr<dev::eth::Client, std::default_delete<dev::eth::Client> >::reset(dev::eth::Client*) /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/bits/unique_ptr.h:347 (eth+0x82c3c9)
    #6 ~WebThreeDirect /home/yh/src/cpp-ethereum2/libwebthree/WebThree.cpp:91 (eth+0x82b7e6)
    #7 main /home/yh/src/cpp-ethereum2/eth/main.cpp:1262 (discriminator 15) (eth+0x512b49)

  Previous read of size 8 at 0x7ba400004df8 by thread T6:
    #0 pthread_cond_broadcast ??:? (eth+0x4869fc)
    #1 std::condition_variable::notify_all() ??:? (libstdc++.so.6+0xb5848)
    #2 operator() /home/yh/src/cpp-ethereum2/libethereum/Client.cpp:102 (eth+0x69bb21)
    #3 std::_Function_handler<void (), dev::eth::Client::init(dev::p2p::Host*, boost::filesystem::path const&, dev::WithExisting, boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<256u, 256u, (boost::multiprecision::cpp_integer_type)0, (boost::multiprecision::cpp_int_check_type)0, void>, (boost::multiprecision::expression_template_option)0>)::$_5>::_M_invoke(std::_Any_data const&) /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/functional:1731 (discriminator 1) (eth+0x69b9f1)
    #4 std::function<void ()>::operator()() const /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/functional:2127 (eth+0x63dc03)
    #5 dev::eth::Signal<>::HandlerAux::fire() /home/yh/src/cpp-ethereum2/libdevcore/../libethcore/Common.h:164 (eth+0x63c78d)
    #6 dev::eth::Signal<>::operator()() /home/yh/src/cpp-ethereum2/libdevcore/../libethcore/Common.h:193 (eth+0x63096e)
    #7 dev::eth::TransactionQueue::manageImport_WITH_LOCK(dev::FixedHash<32u> const&, dev::eth::Transaction const&) /home/yh/src/cpp-ethereum2/libethereum/TransactionQueue.cpp:171 (eth+0x74dc9a)
    #8 dev::eth::TransactionQueue::import(dev::eth::Transaction const&, dev::eth::IfDropped) /home/yh/src/cpp-ethereum2/libethereum/TransactionQueue.cpp:99 (eth+0x74d734)
    #9 dev::eth::TransactionQueue::verifierBody() /home/yh/src/cpp-ethereum2/libethereum/TransactionQueue.cpp:396 (eth+0x74f733)
    #10 operator() /home/yh/src/cpp-ethereum2/libethereum/TransactionQueue.cpp:45 (eth+0x750181)
    #11 void std::_Bind_simple<dev::eth::TransactionQueue::TransactionQueue(unsigned int, unsigned int)::$_4 ()>::_M_invoke<>(std::_Index_tuple<>) /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/functional:1390 (discriminator 2) (eth+0x7500b9)
    #12 std::_Bind_simple<dev::eth::TransactionQueue::TransactionQueue(unsigned int, unsigned int)::$_4 ()>::operator()() /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/functional:1380 (eth+0x750069)
    #13 std::thread::_State_impl<std::_Bind_simple<dev::eth::TransactionQueue::TransactionQueue(unsigned int, unsigned int)::$_4 ()> >::_M_run() /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/thread:197 (eth+0x74ff2d)
    #14 std::error_code::default_error_condition() const ??:? (libstdc++.so.6+0xbb83e)

  As if synchronized via sleep:
    #0 nanosleep ??:? (eth+0x488457)
    #1 void std::this_thread::sleep_for<long, std::ratio<1l, 1000l> >(std::chrono::duration<long, std::ratio<1l, 1000l> > const&) /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/thread:323 (discriminator 1) (eth+0x51926a)
    #2 dev::p2p::Host::stop() /home/yh/src/cpp-ethereum2/libp2p/Host.cpp:153 (discriminator 1) (eth+0x8cdf31)
    #3 ~WebThreeDirect /home/yh/src/cpp-ethereum2/libwebthree/WebThree.cpp:90 (eth+0x82b7d5)
    #4 main /home/yh/src/cpp-ethereum2/eth/main.cpp:1262 (discriminator 15) (eth+0x512b49)
```

libethashseal/EthashClient.cpp
pirapira: Done.

libethereum/EthereumHost.cpp
pirapira: Done.

pirapira:
Took @gumb0's suggestions and still saw no ThreadSanitizer warnings.

libethashseal/EthashClient.cpp
gumb0: I think not calling `terminate` here shouldn't cause any problems, because no virtual functions are overriden in `EthashClient`
But if it fixes some TSAN warnings, then fine.

Also note that there's `ClientTest` class without this call, too.

libethereum/ClientBase.h
gumb0: I see, I would actually prefer to move `m_tq` into `Client` (so that it's close to `m_bq` & `m_bc` and all transaction queue handling logic is in one class)

You'd need something like `virtual TransactionQueue& tq() = 0;` in the `ClientBase` to make it work

libethashseal/EthashClient.cpp
pirapira: Without this change I see a warning:
```
WARNING: ThreadSanitizer: data race on vptr (ctor/dtor vs virtual call) (pid=21629)
  Write of size 8 at 0x7ba400000550 by main thread:
    #0 ~Client /home/yh/src/cpp-ethereum2/libethereum/Client.cpp:82 (eth+0x697c0d)
    #1 ~EthashClient /home/yh/src/cpp-ethereum2/libethashseal/EthashClient.h:37 (discriminator 2) (eth+0x7b4021)
    #2 ~EthashClient /home/yh/src/cpp-ethereum2/libethashseal/EthashClient.h:37 (eth+0x7b4049)
    #3 std::default_delete<dev::eth::Client>::operator()(dev::eth::Client*) const /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/bits/unique_ptr.h:76 (discriminator 1) (eth+0x83039f)
    #4 std::unique_ptr<dev::eth::Client, std::default_delete<dev::eth::Client> >::reset(dev::eth::Client*) /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/bits/unique_ptr.h:347 (eth+0x82e179)
    #5 ~WebThreeDirect /home/yh/src/cpp-ethereum2/libwebthree/WebThree.cpp:91 (eth+0x82d596)
    #6 main /home/yh/src/cpp-ethereum2/eth/main.cpp:1262 (discriminator 15) (eth+0x514a19)

  Previous read of size 8 at 0x7ba400000550 by thread T13:
    #0 dev::Worker::workLoop() /home/yh/src/cpp-ethereum2/libdevcore/Worker.cpp:140 (eth+0x852597)
    #1 operator() /home/yh/src/cpp-ethereum2/libdevcore/Worker.cpp:62 (eth+0x852afd)
    #2 void std::_Bind_simple<dev::Worker::startWorking()::$_0 ()>::_M_invoke<>(std::_Index_tuple<>) /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/functional:1390 (discriminator 2) (eth+0x852989)
    #3 std::_Bind_simple<dev::Worker::startWorking()::$_0 ()>::operator()() /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/functional:1380 (eth+0x852939)
    #4 std::thread::_State_impl<std::_Bind_simple<dev::Worker::startWorking()::$_0 ()> >::_M_run() /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/thread:197 (eth+0x8527fd)
    #5 std::error_code::default_error_condition() const ??:? (libstdc++.so.6+0xbb83e)
```

Myaybe the destructor was forwarded or something.

gumb0: Ok I see, vptr is probaly being modified somewhere between `~EthashClient` & `~Client` body and worker thread can access it at this moment, that's the problem. 
Let's leave `terminate()` here. And add it to `ClientTest`

libethereum/ClientBase.h
pirapira: By @gumb0 's suggestion, I'll try to move all these into `Client`.

pirapira: When I remove this change, I don't see the error anymore.  Maybe the other fixes already killed this warning.  So I just remove this commit.

libethashseal/EthashClient.cpp
pirapira: Done.

pirapira:
```
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
```

The problem persists after a restart.

libweb3jsonrpc/IpcServerBase.h
```diff
@@ -46,7 +47,7 @@ template <class S> class IpcServerBase: public jsonrpc::AbstractServerConnector
 	void GenerateResponse(S _connection);
 
 protected:
-	bool m_running = false;
+	std::atomic<bool> m_running{false};
```
gumb0: Let's use then `std::atomic::exchange()` in `IpcServerBase::StartListening` & `IpcServerBase::StopListening` instead of if+assignment

pirapira: That makes sense.  Will do.

libp2p/Host.cpp
```diff
@@ -694,8 +706,15 @@ void Host::run(boost::system::error_code const&)
 	}
 
 	for (auto p: toConnect)
-		if (p->peerType == PeerType::Required && reqConn++ < m_idealPeerCount)
+	{
+		bool peerTypeIsRequired{};
+		{
+			RecursiveGuard l(x_sessions);
```
gumb0: this looks weird... `toConnect` is local list of shared_ptrs, do you try to protect Peer with a mutex for sessions?

Maybe let's try to make `Node::peerType` atomic, will it help?

libp2p/Host.h
```diff
@@ -318,6 +318,8 @@ class Host: public Worker
 	/// Peers we try to connect regardless of p2p network.
 	std::set<NodeID> m_requiredPeers;
 	Mutex x_requiredPeers;
+	/// returns true if a member of m_requiredPeers
+	bool isRequiredPeer(NodeID const&); // cannot be const because of Mutex
```
gumb0: the usual thing to do in this case is to make mutex `mutable` and leave the method `const`

libweb3jsonrpc/IpcServerBase.h
pirapira: Done.

libp2p/Host.cpp
pirapira: Yes, that worked.

libp2p/Host.h
pirapira: Done.

libethereum/EthereumHost.cpp
```diff
@@ -164,7 +163,7 @@ class EthereumPeerObserver: public EthereumPeerObserverFace
 	}
 
 private:
-	BlockChainSync& m_sync;
+	shared_ptr<BlockChainSync> m_sync;
```
gumb0: I'm afraid we have the same problem with `m_syncMutex` below. If `EthereumHost` gets destroyed before `EthereumPeerObserver` mutex is destroyed with it... Not sure yet what to do with it

`m_tq` is destroyed with the `Client`, also not sure if `EthereumPeers` can outlive `Client`

libp2p/Host.cpp
```diff
@@ -155,6 +155,9 @@ void Host::stop()
 	// stop worker thread
 	if (isWorking())
 		stopWorking();
+
+	// This destroyes EthereumHost objects.  EthereumHost objects contain a const reference to Client's, so they need to disappear first.
```
gumb0: What did you mean by this? I think `EtherumHost` has references to member of `Client`, but not to `Client` itself?

pirapira:
On a 8 core machine, I see more TSAN errors.

libethereum/EthereumHost.cpp
pirapira: (for myself) try to remove `m_syncMutex`.

pirapira: Somehow I'm making other classes hold `m_tq` as a weak pointer.

libp2p/Host.cpp
pirapira: Updated the comment.

libethereum/EthereumHost.cpp
pirapira: And I was able to remove `m_syncMutex` and still seeing no warnings from ThreadSanitizer.

pirapira: About `m_tq`, I decided to use shared/weak pointers.

pirapira:
Done about removing warnings.

libethereum/EthereumHost.cpp
```diff
@@ -54,36 +54,41 @@ namespace
 class EthereumPeerObserver: public EthereumPeerObserverFace
 {
 public:
-	EthereumPeerObserver(BlockChainSync& _sync, RecursiveMutex& _syncMutex, TransactionQueue& _tq): m_sync(_sync), m_syncMutex(_syncMutex), m_tq(_tq) {}
+	EthereumPeerObserver(shared_ptr<BlockChainSync> _sync, weak_ptr<TransactionQueue> _tq): m_sync(_sync), m_tq(_tq) {}
 
 	void onPeerStatus(std::shared_ptr<EthereumPeer> _peer) override
 	{
-		RecursiveGuard l(m_syncMutex);
+		if (!m_sync)
```
gumb0: Why do you need these checks? Isn't it always non-null?

libethereum/BlockChainSync.h
```diff
@@ -139,11 +138,11 @@ class BlockChainSync: public HasInvariants
 		}
 	};
 
-	EthereumHost& m_host;
+	std::weak_ptr<EthereumHost> m_host;
```
gumb0: `EthereumHost` has a `shared_ptr<BlockChainSync>`, so the first one always lives longer than the second one, so I'm not sure why exactly do you need this change?

libethereum/EthereumHost.cpp
pirapira: Without these checks, I see a segmentation fault.

pirapira: `AddressSanitizer` reports an access to zero-page, but cannot show the source line.

With `gdb` I get more clue
```
$ gdb --args eth/eth --ropsten --kill --pin --peerset "required:20c9ad97c081d63397d7b685a412227a40e23c8bdc6688c6f37e97cfbc22d2b4d1db1510d8f61e6a8866ad7f0e17c02b14182d37ea7c3c8b9c2683aeb6b733a1@52.169.14.227:30303 required:6ce05930c72abc632c58e2e4324f7c7ea478cec0ed4fa2528982cf34483094e9cbc9216e7aa349691242576d552a2a56aaeae426c5303ded677ce455ba1acd9d@13.84.180.240:30303" --port 30304

(gdb) run
Thread 31 "p2p" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffe5ee3700 (LWP 6139)]
__GI___pthread_mutex_lock (mutex=0x28) at ../nptl/pthread_mutex_lock.c:67
67      ../nptl/pthread_mutex_lock.c: No such file or directory.
#1  0x00005555557707ef in __gthread_mutex_lock (__mutex=0x28)
    at /usr/include/x86_64-linux-gnu/c++/7/bits/gthr-default.h:748
#2  __gthread_recursive_mutex_lock (__mutex=0x28)
    at /usr/include/x86_64-linux-gnu/c++/7/bits/gthr-default.h:810
#3  std::recursive_mutex::lock (this=0x28) at /usr/include/c++/7/mutex:107
#4  std::lock_guard<std::recursive_mutex>::lock_guard (__m=..., this=<synthetic pointer>)
    at /usr/include/c++/7/bits/std_mutex.h:162
#5  dev::eth::BlockChainSync::onPeerAborting (this=0x0)
    at /home/yh/src/cpp-ethereum/libethereum/BlockChainSync.cpp:947
#6  0x00005555556f11ac in (anonymous namespace)::EthereumPeerObserver::onPeerAborting (
    this=<optimized out>) at /home/yh/src/cpp-ethereum/libethereum/EthereumHost.cpp:87
#7  0x00005555557003be in dev::eth::EthereumPeer::abortSync (this=this@entry=0x7fffd80682e0)
    at /home/yh/src/cpp-ethereum/libethereum/EthereumPeer.cpp:113
#6  0x00005555556f11ac in (anonymous namespace)::EthereumPeerObserver::onPeerAborting (
    this=<optimized out>) at /home/yh/src/cpp-ethereum/libethereum/EthereumHost.cpp:87
87                              m_sync->onPeerAborting();
```

libethereum/BlockChainSync.h
pirapira: Without this change, I get this warning
```
==================
WARNING: ThreadSanitizer: heap-use-after-free (pid=10192)
  Read of size 8 at 0x7b540000b4b8 by main thread (mutexes: write M1070):
    #0 dev::eth::EthereumHost::chain() const /home/yh/src/cpp-ethereum/libethereum/EthereumHost.h:85 (eth+0x77cc7d)
    #1 dev::eth::BlockChainSync::invariants() const /home/yh/src/cpp-ethereum/libethereum/BlockChainSync.cpp:918 (eth+0x77b956)
```

This is because `BlockChainSync.cpp` looks up `m_host->chain()` after `m_host` is freed.

> EthereumHost has a shared_ptr<BlockChainSync>, so the first one always lives longer than the second one

No.  The first one always lives shorter than the second one.

pirapira:
I can reproduce the failures locally but not on develop.

pirapira:
The error in `circleci: macOS-XCode9` looks interesting:
```
  1/398 Test   #2: RandomCodeTests/rndStateTest ..........................................................***Exception: SegFault  1.62 sec
Running 1 test case...
Test Case "rndStateTest": 
```

I'll try to reproduce this, maybe next week.

pirapira:
I cannot reproduce the segmentation fault in `RandomCodeTests/rndStateTest` on a Linux environment.  Maybe I should try this on a mac machine at home.

libethereum/EthereumHost.cpp
pirapira: @gumb0 told me that I left `EthereumPeerObserver` unaware of the `BlockChainSync` object.  If I fix this, I shouldn't need `if(m_sync)` checks.

libethereum/BlockChainSync.h
gumb0: Right, sorry, I meant `BlockChainSync` lives at least as long as `EthereumHost`...

I see, `BlockChainSync` is now shared between `EthereumHost` and `EthereumPeer`s (through `EthereumPeerObserver`)

gumb0: Actually I think that `p2p::Host` takes care (in `Host::doneWorking()`) to stop any communication and clear all peers when it is finishing. And `p2p::Host::m_capabilities` holding `EthereumHost` should be still alive after that.

There might be something not right in either `p2p::Host` or `Worker` which should be fixed instead of complicating the code here...

Would it be possible to get more detailed stack of this problem with `EthereumHost` not living long enough?

pirapira: I'll try to get a better picture after fixing yet another TSAN warning (which I didn't see last week).

pirapira: Here are the stacks:
```
==================
WARNING: ThreadSanitizer: heap-use-after-free (pid=8057)
  Read of size 8 at 0x7b540000a338 by main thread (mutexes: write M1083):
    #0 dev::eth::EthereumHost::chain() const /home/yh/src/cpp-ethereum-vptr/libethereum/EthereumHost.h:85 (eth+0x00000033c86e)
    #1 dev::eth::BlockChainSync::invariants() const /home/yh/src/cpp-ethereum-vptr/libethereum/BlockChainSync.cpp:918 (eth+0x00000033c86e)
    #2 dev::InvariantChecker::checkInvariants(dev::HasInvariants const*, char const*, char const*, int, bool) /home/yh/src/cpp-ethereum-vptr/libdevcore/Common.cpp:39 (eth+0x00000045ab15)
    #3 dev::eth::BlockChainSync::onPeerAborting() /home/yh/src/cpp-ethereum-vptr/libethereum/BlockChainSync.cpp:909 (eth+0x000000341077)
    #4 onPeerAborting /home/yh/src/cpp-ethereum-vptr/libethereum/EthereumHost.cpp:87 (eth+0x00000025b0ae)
    #5 dev::eth::EthereumPeer::abortSync() /home/yh/src/cpp-ethereum-vptr/libethereum/EthereumPeer.cpp:113 (eth+0x00000027586f)
    #6 dev::eth::EthereumPeer::~EthereumPeer() /home/yh/src/cpp-ethereum-vptr/libethereum/EthereumPeer.cpp:67 (eth+0x000000277268)
    #7 void __gnu_cxx::new_allocator<dev::eth::EthereumPeer>::destroy<dev::eth::EthereumPeer>(dev::eth::EthereumPeer*) /usr/include/c++/7/ext/new_allocator.h:140 (eth+0x00000026b8bf)
    #8 void std::allocator_traits<std::allocator<dev::eth::EthereumPeer> >::destroy<dev::eth::EthereumPeer>(std::allocator<dev::eth::EthereumPeer>&, dev::eth::EthereumPeer*) /usr/include/c++/7/bits/alloc_traits.h:487 (eth+0x00000026b8bf)
    #9 std::_Sp_counted_ptr_inplace<dev::eth::EthereumPeer, std::allocator<dev::eth::EthereumPeer>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() /usr/include/c++/7/bits/shared_ptr_base.h:535 (eth+0x00000026b8bf)
    #10 std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() /usr/include/c++/7/bits/shared_ptr_base.h:154 (eth+0x0000000dcf01)
    #11 std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count() /usr/include/c++/7/bits/shared_ptr_base.h:684 (eth+0x0000005f6f7f)
    #12 std::__shared_ptr<dev::p2p::Capability, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr() /usr/include/c++/7/bits/shared_ptr_base.h:1123 (eth+0x0000005f6f7f)
    #13 std::__shared_ptr<dev::p2p::Capability, (__gnu_cxx::_Lock_policy)2>::reset() /usr/include/c++/7/bits/shared_ptr_base.h:1235 (eth+0x0000005f6f7f)
    #14 dev::p2p::Session::~Session() /home/yh/src/cpp-ethereum-vptr/libp2p/Session.cpp:58 (eth+0x0000005f6f7f)
    #15 void __gnu_cxx::new_allocator<dev::p2p::Session>::destroy<dev::p2p::Session>(dev::p2p::Session*) /usr/include/c++/7/ext/new_allocator.h:140 (eth+0x0000005874f9)
    #16 void std::allocator_traits<std::allocator<dev::p2p::Session> >::destroy<dev::p2p::Session>(std::allocator<dev::p2p::Session>&, dev::p2p::Session*) /usr/include/c++/7/bits/alloc_traits.h:487 (eth+0x0000005874f9)
    #17 std::_Sp_counted_ptr_inplace<dev::p2p::Session, std::allocator<dev::p2p::Session>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() /usr/include/c++/7/bits/shared_ptr_base.h:535 (eth+0x0000005874f9)
    #18 std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() /usr/include/c++/7/bits/shared_ptr_base.h:154 (eth+0x0000000dcf01)
    #19 std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count() /usr/include/c++/7/bits/shared_ptr_base.h:684 (eth+0x000000600a38)
    #20 std::__shared_ptr<dev::p2p::SessionFace, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr() /usr/include/c++/7/bits/shared_ptr_base.h:1123 (eth+0x000000600a38)
    #21 std::shared_ptr<dev::p2p::SessionFace>::~shared_ptr() /usr/include/c++/7/bits/shared_ptr.h:93 (eth+0x000000600a38)
    #22 ~<lambda> /home/yh/src/cpp-ethereum-vptr/libp2p/Session.cpp:311 (eth+0x000000600a38)
    #23 ~read_op /home/yh/.hunter/_Base/b96750b/a5f7296/9d3cb74/Install/include/boost/asio/impl/read.hpp:221 (eth+0x000000600a38)
    #24 ~binder2 /home/yh/.hunter/_Base/b96750b/a5f7296/9d3cb74/Install/include/boost/asio/detail/bind_handler.hpp:108 (eth+0x000000600a38)
    #25 do_complete /home/yh/.hunter/_Base/b96750b/a5f7296/9d3cb74/Install/include/boost/asio/detail/reactive_socket_recv_op.hpp:101 (eth+0x000000600a38)
    #26 boost::asio::detail::epoll_reactor::shutdown_service() <null> (eth+0x0000005687bb)
    #27 boost::asio::detail::service_registry::~service_registry() /home/yh/.hunter/_Base/b96750b/a5f7296/9d3cb74/Install/include/boost/asio/detail/impl/service_registry.ipp:37 (eth+0x000000583385)
    #28 boost::asio::io_service::~io_service() /home/yh/.hunter/_Base/b96750b/a5f7296/9d3cb74/Install/include/boost/asio/impl/io_service.ipp:53 (eth+0x000000583385)
    #29 dev::p2p::Host::~Host() /home/yh/src/cpp-ethereum-vptr/libp2p/Host.cpp:119 (eth+0x000000583385)
    #30 dev::WebThreeDirect::~WebThreeDirect() /home/yh/src/cpp-ethereum-vptr/libwebthree/WebThree.cpp:78 (eth+0x000000456d9e)
    #31 main /home/yh/src/cpp-ethereum-vptr/eth/main.cpp:937 (eth+0x0000000c0f32)

  Previous write of size 8 at 0x7b540000a338 by main thread:
    #0 operator delete(void*) <null> (libtsan.so.0+0x00000006fd54)
    #1 __gnu_cxx::new_allocator<std::_Sp_counted_ptr_inplace<dev::eth::EthereumHost, std::allocator<dev::eth::EthereumHost>, (__gnu_cxx::_Lock_policy)2> >::deallocate(std::_Sp_counted_ptr_inplace<dev::eth::EthereumHost, std::allocator<dev::eth::EthereumHost>, (__gnu_cxx::_Lock_policy)2>*, unsigned long) /usr/include/c++/7/ext/new_allocator.h:125 (eth+0x0000002327de)
    #2 std::allocator_traits<std::allocator<std::_Sp_counted_ptr_inplace<dev::eth::EthereumHost, std::allocator<dev::eth::EthereumHost>, (__gnu_cxx::_Lock_policy)2> > >::deallocate(std::allocator<std::_Sp_counted_ptr_inplace<dev::eth::EthereumHost, std::allocator<dev::eth::EthereumHost>, (__gnu_cxx::_Lock_policy)2> >&, std::_Sp_counted_ptr_inplace<dev::eth::EthereumHost, std::allocator<dev::eth::EthereumHost>, (__gnu_cxx::_Lock_policy)2>*, unsigned long) /usr/include/c++/7/bits/alloc_traits.h:462 (eth+0x0000002327de)
    #3 std::__allocated_ptr<std::allocator<std::_Sp_counted_ptr_inplace<dev::eth::EthereumHost, std::allocator<dev::eth::EthereumHost>, (__gnu_cxx::_Lock_policy)2> > >::~__allocated_ptr() /usr/include/c++/7/bits/allocated_ptr.h:73 (eth+0x0000002327de)
    #4 std::_Sp_counted_ptr_inplace<dev::eth::EthereumHost, std::allocator<dev::eth::EthereumHost>, (__gnu_cxx::_Lock_policy)2>::_M_destroy() /usr/include/c++/7/bits/shared_ptr_base.h:543 (eth+0x0000002327de)
    #5 std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_weak_release() /usr/include/c++/7/bits/shared_ptr_base.h:193 (eth+0x000000235645)
    #6 std::__weak_count<(__gnu_cxx::_Lock_policy)2>::~__weak_count() /usr/include/c++/7/bits/shared_ptr_base.h:777 (eth+0x00000022232b)
    #7 std::__weak_ptr<dev::eth::EthereumHost, (__gnu_cxx::_Lock_policy)2>::~__weak_ptr() /usr/include/c++/7/bits/shared_ptr_base.h:1609 (eth+0x00000022232b)
    #8 std::weak_ptr<dev::eth::EthereumHost>::~weak_ptr() /usr/include/c++/7/bits/shared_ptr.h:520 (eth+0x00000022232b)
    #9 dev::eth::Client::~Client() /home/yh/src/cpp-ethereum-vptr/libethereum/Client.cpp:81 (eth+0x00000022232b)
    #10 dev::eth::EthashClient::~EthashClient() /home/yh/src/cpp-ethereum-vptr/libethashseal/EthashClient.cpp:62 (eth+0x00000038b317)
    #11 dev::eth::EthashClient::~EthashClient() /home/yh/src/cpp-ethereum-vptr/libethashseal/EthashClient.cpp:65 (eth+0x00000038b36a)
    #12 std::default_delete<dev::eth::Client>::operator()(dev::eth::Client*) const /usr/include/c++/7/bits/unique_ptr.h:78 (eth+0x000000456d45)
    #13 std::unique_ptr<dev::eth::Client, std::default_delete<dev::eth::Client> >::reset(dev::eth::Client*) /usr/include/c++/7/bits/unique_ptr.h:376 (eth+0x000000456d45)
    #14 dev::WebThreeDirect::~WebThreeDirect() /home/yh/src/cpp-ethereum-vptr/libwebthree/WebThree.cpp:91 (eth+0x000000456d45)
    #15 main /home/yh/src/cpp-ethereum-vptr/eth/main.cpp:937 (eth+0x0000000c0f32)
```

gumb0: I see, this is complicated...
I think what's happening is shared_ptr's of `EthereumPeer` are being captured in lambda-handlers of network messages and saved inside `m_ioService` of `Host` and these handlers are not deleted as I thought when `Host` calls `m_ioService.stop()` in `Host::doneWorking()`. And I don't see an easy way to clear these handlers...

Anyway how about this fix: make `EthereumPeer::m_observer` weak_ptr (probably `EthereumPeer::m_hostData`, too)
Then network message handlers won't keep `BlockChainSync` alive and it won't try to access `EthereumHost` after it's deleted. Conceptually observer object should be allowed to be destroyed for it's own reasons, the object being observed just won't notify anyone then.
I think the code could be much simpler then.

pirapira: Sounds good.  I'll try.

pirapira: That worked.

libethereum/ClientBase.h
```diff
@@ -180,7 +180,7 @@ class ClientBase: public Interface
 	virtual void prepareForTransaction() = 0;
 	/// }
 
-	TransactionQueue m_tq;							///< Maintains a list of incoming transactions not yet in a block on the blockchain.
+	std::shared_ptr<TransactionQueue> m_tq;	///< Maintains a list of incoming transactions not yet in a block on the blockchain. This is never null.  Only passed on as a weak_ptr.
```
gumb0: Not sure if "Only passed as a weak_ptr" is true, `BlockChain::sync` receives it as shared_ptr

libethereum/EthereumPeer.cpp
```diff
@@ -236,7 +236,10 @@ bool EthereumPeer::isCriticalSyncing() const
 
 bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
 {
-	assert(m_observer);
+	auto observer = m_observer.lock();
+	assert(observer);
```
gumb0: I think we should better return false here instead of assert, for the case when network message handler is called at the time observer doesn't exist anymore

libp2p/Host.cpp
```diff
@@ -149,12 +149,22 @@ void Host::stop()
...
+			l.lock();
+		}
+	}
 
 	// stop worker thread
 	if (isWorking())
 		stopWorking();
+
+	// This destroyes EthereumHost objects.  EthereumHost objects contain a references to Clients' members, so EthereumHost objects need to disappear first.
+	m_capabilities.clear();
```
gumb0: I think we can't put this into `Host::stop()`. Network can be stopped and then restarted with admin RPC. (`Host::stop()` & `Host::start()` are called in this case and restart won't work)

libp2p/Host.cpp
```diff
@@ -149,12 +149,22 @@ void Host::stop()
 		return;
 
 	// wait for m_timer to reset (indicating network scheduler has stopped)
-	while (!!m_timer)
-		this_thread::sleep_for(chrono::milliseconds(50));
+	{
```
gumb0: Maybe we could call `m_timer.wait()` or `m_timer.cancel()` instead of all this piece? I'm not sure

libp2p/Host.cpp
pirapira: This change is not necessary anymore.  I don't see TSAN warnings without it.  I'll remove it.

libethereum/ClientBase.h
pirapira: I'm changing the sentence into "Only stored in this object."

libethereum/EthereumPeer.cpp
pirapira: OK.

libp2p/Host.cpp
pirapira: `m_timer` is a unique_ptr and it's reset in another thread.  I started using a condition variable.

pirapira:
Travis, Linux:
```
/home/travis/build/ethereum/cpp-ethereum/libp2p/Host.h:224:92: note:   ‘std::shared_ptr<dev::p2p::NodeTable>’ is not derived from ‘const std::atomic<_ITp>’

  bool haveNetwork() const { Guard l(x_runTimer); return m_run && !!atomic_load(&m_nodeTable); }
```

pirapira:
So I'm hitting this issue https://stackoverflow.com/questions/34205880/why-does-atomic-load-with-a-shared-ptr-not-compile-with-gcc-4-9

chfast:
There was a proposal (with some support) to raise GCC minimal requirement to version 5.

libethereum/Client.cpp
```diff
@@ -82,6 +82,8 @@ Client::~Client()
 {
 	stopWorking();
 	terminate();
+	m_tqReady = {}; // m_tqReady holds a closure that uses m_signalled.
```
gumb0: Not sure, how can this help anything? It is a `shared_ptr<function>`, and you reset it here in `Client`, but `TransactionQueue` still will hold a non-null shared_ptr to a function calling `Client::onTransactionQueueReady()`

pirapira: `m_tqReady` is a member of `ClientBase`.  After `Client` is destroyed but `ClientBase` is alive, when `m_tqReady()` is called, it tries to access a member of `Client` that has been destroyed.

pirapira: And I don't see any TSAN warnings without these.  This was cargo cult.

libp2p/Host.h
```diff
@@ -304,20 +303,25 @@ class Host: public Worker
...
 	static const unsigned c_timerInterval = 100;							///< Interval which m_timer is run when network is connected.
+	std::condition_variable m_timerSemaphore;
 
 	std::set<Peer*> m_pendingPeerConns;									/// Used only by connect(Peer&) to limit concurrently connecting to same node. See connect(shared_ptr<Peer>const&).
 	Mutex x_pendingNodeConns;
 
 	bi::tcp::endpoint m_tcpPublic;											///< Our public listening endpoint.
 	KeyPair m_alias;															///< Alias for network communication. Network address is k*G. k is key material. TODO: Replace KeyPair.
 	std::shared_ptr<NodeTable> m_nodeTable;									///< Node table (uses kademlia-like discovery).
+	mutable std::mutex x_nodeTable;
```
gumb0: I doubt that this makes sense, because operations on a `shared_ptr` (like reset, copy, assignment, operator bool) are thread-safe.
So what does it fix?

pirapira: It fixes a warning:
```
WARNING: ThreadSanitizer: data race (pid=68621)
  Write of size 8 at 0x7ffc9f6af3b8 by thread T30:
    #0 std::__shared_ptr<dev::p2p::NodeTable, (__gnu_cxx::_Lock_policy)2>::operator=(std::__shared_ptr<dev::p2p::NodeTable, (__gnu_cxx::_Lock_policy)2> const&) /usr/include/c++/7/bits/shared_ptr_base.h:1034 (eth+0x0000005805bb)   
    #1 std::shared_ptr<dev::p2p::NodeTable>::operator=(std::shared_ptr<dev::p2p::NodeTable> const&) /usr/include/c++/7/bits/shared_ptr.h:93 (eth+0x0000005805bb)
    #2 dev::p2p::Host::startedWorking() /home/yh/src/cpp-ethereum-vptr/libp2p/Host.cpp:761 (eth+0x0000005805bb)
    #3 operator() /home/yh/src/cpp-ethereum-vptr/libdevcore/Worker.cpp:61 (eth+0x0000004ab343)
    #4 __invoke_impl<void, dev::Worker::startWorking()::<lambda()> > /usr/include/c++/7/bits/invoke.h:60 (eth+0x0000004ab6a7)
    #5 __invoke<dev::Worker::startWorking()::<lambda()> > /usr/include/c++/7/bits/invoke.h:95 (eth+0x0000004ab6a7)
    #6 _M_invoke<0> /usr/include/c++/7/thread:234 (eth+0x0000004ab6a7)
    #7 operator() /usr/include/c++/7/thread:243 (eth+0x0000004ab6a7)
    #8 _M_run /usr/include/c++/7/thread:186 (eth+0x0000004ab6a7)
    #9 <null> <null> (libstdc++.so.6+0x0000000bc0fe)

  Previous read of size 8 at 0x7ffc9f6af3b8 by main thread (mutexes: write M1111):
    #0 std::__shared_ptr<dev::p2p::NodeTable, (__gnu_cxx::_Lock_policy)2>::operator bool() const /usr/include/c++/7/bits/shared_ptr_base.h:1261 (eth+0x0000005716f8)
    #1 dev::p2p::Host::haveNetwork() const /home/yh/src/cpp-ethereum-vptr/libp2p/Host.h:224 (eth+0x0000005716f8)
    #2 dev::p2p::Host::start() /home/yh/src/cpp-ethereum-vptr/libp2p/Host.cpp:129 (eth+0x0000005716f8)
    #3 dev::WebThreeDirect::startNetwork() /home/yh/src/cpp-ethereum-vptr/libdevcore/../libwebthree/WebThree.h:204 (eth+0x0000000bf0d9)
    #4 main /home/yh/src/cpp-ethereum-vptr/eth/main.cpp:1117 (eth+0x0000000bf0d9)
```

Some change is necessary because in the original code, I see
```
if (m_nodeTable)
   m_nodeTable->whatever();
```
which is not thread safe.

pirapira: I think I saw this

> If multiple threads of execution access the same shared_ptr without synchronization and any of those accesses uses a non-const member function of shared_ptr then a data race will occur;

http://en.cppreference.com/w/cpp/memory/shared_ptr

And I added `atomic_load`.  Then gcc 4.8.3 complained.  So I used mutex.

gumb0: I see, then accessing it through just `std::shared_ptr<NodeTable> nodeTable() const { return m_nodeTable; }` without mutex would be as safe for this case, because it will return another `shared_ptr` and

> All member functions (including copy constructor and copy assignment) can be called by multiple threads on different instances of shared_ptr without additional synchronization even if these instances are copies and share ownership of the same object.

But this won't fix data races around `NodeTable` itself.

gumb0: No, I think I'm not right, copying from `m_nodeTable` still can happen simultaneously with writing to it, so my idea is not safe

libethereum/Client.cpp
pirapira: About the `TransactionQueue` problem I filed an issue https://github.com/ethereum/cpp-ethereum/issues/4608

pirapira:
Review #4609 first.  This one is rebased on top of that.

libethereum/BlockChainSync.h
gumb0: So did you push this change? I don't see any changes in `EthereumPeer`

pirapira: Wow, I have it only in one local branch.

libp2p/Host.cpp
gumb0: So it's still here, are we removing it?

libethereum/ClientBase.h
gumb0: This was not changed yet

libp2p/Host.cpp
pirapira: Yes.  Done.

pirapira:
The bad news is that I'm seeing a warning from TSAN again.  I think I lost some changes.

libp2p/Session.cpp
```diff
@@ -275,8 +275,22 @@ void Session::drop(DisconnectReason _reason)
 	m_peer->m_lastDisconnect = _reason;
 	if (_reason == BadProtocol)
 	{
-		m_peer->m_rating /= 2;
-		m_peer->m_score /= 2;
+		int old_rate = 0;
```
gumb0: Could we extract this to a separate helper function e.g. `halveAtomicInt(std::atomic<int>&)` ?

Also naming convention prescribes `oldRate` & `newRate`

pirapira:
I recovered enough commits to make TSAN warnings go away.

libp2p/Host.cpp
pirapira: I think I have a commit containing a condition variable somewhere.  I have to recover that.

pirapira:
I'm again seeing some new TSAN warnings.

pirapira:
Again segfault in rndStateTest.

gumb0:
I tried to simplify a little the code accessing `m_nodeTable` by extracting most of the operations with it to separate methods. @pirapira please verify that TSAN is still ok with it

pirapira:
@gumb0 I'm on it.

pirapira:
Yes, TSAN seems happy.

For the record, this is the command I run
```
eth/eth --ropsten --kill --pin --peerset "required:20c9ad97c081d63397d7b685a412227a40e23c8bdc6688c6f37e97cfbc22d2b4d1db1510d8f61e6a8866ad7f0e17c02b14182d37ea7c3c8b9c2683aeb6b733a1@52.169.14.227:30303 required:6ce05930c72abc632c58e2e4324f7c7ea478cec0ed4fa2528982cf34483094e9cbc9216e7aa349691242576d552a2a56aaeae426c5303ded677ce455ba1acd9d@13.84.180.240:30303" --listen 20000
```
and then send a SIGINT.

pirapira:
@gumb0 do you want to make more changes?

gumb0:
Not sure yet, I will try to finish review today.

libp2p/Host.cpp
```diff
@@ -142,15 +142,17 @@ void Host::stop()
 	// called to force io_service to kill any remaining tasks it might have -
 	// such tasks may involve socket reads from Capabilities that maintain references
 	// to resources we're about to free.
+	
+	{
+		unique_lock<mutex> l(x_runTimer);
+		// ignore if already stopped/stopping, at the same time,
+		// signal run() to prepare for shutdown and reset m_timer
+		if (!m_run.exchange(false))
```
gumb0: Not sure why do you do this under `x_runTimer` lock?

pirapira: When I move this out above the block, I see
```
==================
WARNING: ThreadSanitizer: data race (pid=65004)
  Write of size 8 at 0x7ba400004df8 by main thread:
    #0 pthread_cond_destroy <null> (libtsan.so.0+0x000000029892)
    #1 dev::eth::Client::~Client() /home/yh/src/cpp-ethereum-vptr/libethereum/Client.cpp:82 (eth+0x00000022094c)
    #2 dev::eth::EthashClient::~EthashClient() /home/yh/src/cpp-ethereum-vptr/libethashseal/EthashClient.cpp:62 (eth+0x000000389795)
    #3 dev::eth::EthashClient::~EthashClient() /home/yh/src/cpp-ethereum-vptr/libethashseal/EthashClient.cpp:65 (eth+0x0000003897e8)
    #4 std::default_delete<dev::eth::Client>::operator()(dev::eth::Client*) const /usr/include/c++/7/bits/unique_ptr.h:78 (eth+0x0000004551c3)
    #5 std::unique_ptr<dev::eth::Client, std::default_delete<dev::eth::Client> >::reset(dev::eth::Client*) /usr/include/c++/7/bits/unique_ptr.h:376 (eth+0x0000004551c3)
    #6 dev::WebThreeDirect::~WebThreeDirect() /home/yh/src/cpp-ethereum-vptr/libwebthree/WebThree.cpp:91 (eth+0x0000004551c3)
    #7 main /home/yh/src/cpp-ethereum-vptr/eth/main.cpp:937 (eth+0x0000000bf8f2)

  Previous read of size 8 at 0x7ba400004df8 by thread T28:
    #0 pthread_cond_broadcast <null> (libtsan.so.0+0x00000002975f)
    #1 std::condition_variable::notify_all() <null> (libstdc++.so.6+0x0000000b5df8)
    #2 std::function<void ()>::operator()() const /usr/include/c++/7/bits/std_function.h:706 (eth+0x0000001cf4f3)
    #3 dev::eth::Signal<>::operator()() <null> (eth+0x0000001d6612)
    #4 dev::eth::TransactionQueue::manageImport_WITH_LOCK(dev::FixedHash<32u> const&, dev::eth::Transaction const&) /home/yh/src/cpp-ethereum-vptr/libethereum/TransactionQueue.cpp:171 (eth+0x00000031b951)
    #5 dev::eth::TransactionQueue::import(dev::eth::Transaction const&, dev::eth::IfDropped) /home/yh/src/cpp-ethereum-vptr/libethereum/TransactionQueue.cpp:99 (eth+0x00000031bee0)
    #6 dev::eth::TransactionQueue::verifierBody() /home/yh/src/cpp-ethereum-vptr/libethereum/TransactionQueue.cpp:396 (eth+0x00000031c49c)
    #7 operator() /home/yh/src/cpp-ethereum-vptr/libethereum/TransactionQueue.cpp:45 (eth+0x00000031ccac)
    #8 __invoke_impl<void, dev::eth::TransactionQueue::TransactionQueue(unsigned int, unsigned int)::<lambda()> > /usr/include/c++/7/bits/invoke.h:60 (eth+0x00000031cd57)
    #9 __invoke<dev::eth::TransactionQueue::TransactionQueue(unsigned int, unsigned int)::<lambda()> > /usr/include/c++/7/bits/invoke.h:95 (eth+0x00000031cd57)
    #10 _M_invoke<0> /usr/include/c++/7/thread:234 (eth+0x00000031cd57)
    #11 operator() /usr/include/c++/7/thread:243 (eth+0x00000031cd57)
    #12 _M_run /usr/include/c++/7/thread:186 (eth+0x00000031cd57)
    #13 <null> <null> (libstdc++.so.6+0x0000000bc0fe)

  As if synchronized via sleep:
    #0 nanosleep <null> (libtsan.so.0+0x00000004c5a1)
    #1 void std::this_thread::sleep_for<long, std::ratio<1l, 1000l> >(std::chrono::duration<long, std::ratio<1l, 1000l> > const&) /usr/include/c++/7/thread:376 (eth+0x0000000b723c)
    #2 stopSealingAfterXBlocks(dev::eth::Client*, unsigned int, unsigned int&) /home/yh/src/cpp-ethereum-vptr/eth/main.cpp:274 (eth+0x0000000b723c)
    #3 main /home/yh/src/cpp-ethereum-vptr/eth/main.cpp:1247 (eth+0x0000000bf614)
```

I'm confused now because `Host.cpp` doesn't appear in the stacks.

pirapira: @gumb0 forgetting the TSAN warning, I can somehow explain.  If I set `m_run` outside without the lock, the worker termination can immediately happen, `m_timerSemaphore` might be a notification, even before this thread gets the lock.  And, this thread waits on `m_timerSemaphore` forever.

gumb0: But no, `while (m_timer)` will be `false` in this case

gumb0: These stacks I think show that in `Client` condition variable `m_signalled` is being destroyed while still being used by the `m_tqReady` handler. 

Can you try moving `m_signalled` definition above `m_tqReady` in `Client.h` to change destruction order?

gumb0: Also I would name `m_timerSemaphore` something like `m_timerReset` (it's not a semaphore, it's a condition variable anyway)

pirapira: Yes, moving `m_signalled` definition worked.

pirapira: Pushing....

pirapira: Done an done.  I'm not seeing TSAN warnings.

pirapira:
Turning invisible complications into visible complications...

PR #4581, opened 2017-10-05
do not print json test in createRandomTest unit test

winsvega:


test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -45,6 +45,7 @@ bool createRandomTest()
 		RandomCodeOptions options;
 		TestOutputHelper testOutputHelper;
 		std::string test = test::RandomCode::fillRandomTest(suite, c_testExampleStateTest, options);
+		std::cout << test << "\n";
```
pirapira: Why is an output being added in a PR "do not print ..."?  I'm confused.

winsvega: do not print in unit test. but print in --createRandomTest  option. I just moved the print statement.

Issue #4515, opened 2017-09-18
libfuzzer-friendly GeneralStateTest generation


pirapira:
Currently, `test/tools/fuzzTesting` can randomly generate a GeneralStateTest.

This issue keeps track of replacing the randomly generated numbers with some input byte string (given at runtime) so that `libfuzzer` can explore the VM implementation.

Concretely, the task involves implementing a function
```
extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
```
that generates a GeneralStateTest and then run it.  The input `Data` should be used in a straightforward way (not going through compression/encryption/decryption).  The whole function should be deterministic, without relying on pseudo random numbers or timestamps.

When we connect `libfuzzer` with this function, `libfuzzer` will try to find inputs that cover different execution paths in the VM implementation.

winsvega:
so we need to timplement LLVMFuzzerTestOneInput function in testeth which would take  array of uint8_t as data and parse it into general state test which would then be executed. 
what is the return int value means?

pirapira:
@winsvega the return value is not useful.  Currently only zero is supported, or something like that.  You can throw an exception to indicate an error.  The most interesting data is the code coverage, which is collected by the libfuzzer instrumentation `cmake -DSANITIZE=fuzzer` .  After a nights execution, the `CORPUS` directory will contain hundreds of bytestrings that translate to interesting GeneralStateTests.

winsvega:
what would be the entry point with libfuzzer?

pirapira:
`libfuzzer` contains the `main()` function.  When you link something with `libfuzzer`, `libfuzzer`'s `main()` will be the entry point.

winsvega:
I wonder how boost would work if the entry point is different

pirapira:
No idea.  In the worst case, we introduce another macro.

winsvega:
Guido does not want to use testeth randomtest generator for smart random state tests

pirapira:
OK.

PR #4587, opened 2017-10-06
tweak createRandomTest params

cdetrio:
A bit rough, but these params work well to generate interesting traces, with many CALLs succeeding so that the call depth increases. Also the Byzantium opcodes and precompiles are weighted heavier (sampled more frequently).

### DIFF
--- a/test/tools/fuzzTesting/createRandomTest.cpp
+++ b/test/tools/fuzzTesting/createRandomTest.cpp
@@ -34,7 +34,8 @@ namespace dev { namespace test {
 bool createRandomTest()
 {
 	StateTestSuite suite;
-	dev::test::Options& options = const_cast<dev::test::Options&>(dev::test::Options::get());
+	dev::test::Options& optionsConst = const_cast<dev::test::Options&>(dev::test::Options::get());
+	dev::test::Options& options = optionsConst;
 	if (options.rCurrentTestSuite != suite.suiteFolder())
 	{
 		std::cerr << "Error! Test suite '" + options.rCurrentTestSuite + "' not supported! (Usage -t <TestSuite>)\n";
@@ -52,7 +53,7 @@ bool createRandomTest()
 }} //namespaces
 
 //Prints a generated test Json into std::out
-std::string dev::test::RandomCode::fillRandomTest(dev::test::TestSuite const& _testSuite, std::string const& _testString, dev::test::RandomCodeOptions const& _options)
+std::string dev::test::RandomCode::fillRandomTest(dev::test::TestSuite const& _testSuite, std::string const& _testString, dev::test::RandomCodeOptions& _options)
 {
 	bool wasError = false;
 	json_spirit::mValue v;
@@ -88,7 +89,7 @@ std::string dev::test::RandomCode::fillRandomTest(dev::test::TestSuite const& _t
 }
 
 /// Parse Test string replacing keywords to fuzzed values
-void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std::string, std::string> const& _varMap, RandomCodeOptions const& _options)
+void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std::string, std::string> const& _varMap, RandomCodeOptions& _options)
 {
 	std::vector<std::string> types = getTypes();
 
@@ -109,7 +110,7 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
 				cnote << debug;
 			}
 			else if (type == "[CODE]")
-				replace = dev::test::RandomCode::generate(10, _options);
+				replace = dev::test::RandomCode::generate(50, _options);
 			else if (type == "[HEX]")
 				replace = dev::test::RandomCode::randomUniIntHex();
 			else if (type == "[HEX32]")
@@ -134,16 +135,20 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
 				replace = test::RandomCode::randomUniIntHex(dev::u256("100000"), dev::u256("36028797018963967"));
 			else if (type == "[DESTADDRESS]")
 			{
-				Address address = _options.getRandomAddress(RandomCodeOptions::AddressType::PrecompiledOrStateOrCreate);
+				Address address = _options.getRandomAddress(RandomCodeOptions::AddressType::DestinationAccount);
 				if (address != ZeroAddress) //else transaction creation
 					replace = "0x" + toString(address);
 			}
-			else if (type == "[ADDRESS]")
-				replace = toString(_options.getRandomAddress(RandomCodeOptions::AddressType::StateAccount));
-			else if (type == "[0xADDRESS]")
+			else if (type == "[ADDRESS]") {
+				Address destAddress = _options.getRandomAddress(RandomCodeOptions::AddressType::StateAccount);
+				_options.addAddress(destAddress, RandomCodeOptions::AddressType::DestinationAccount);
+				replace = toString(destAddress);
+			}
+			else if (type == "[0xADDRESS]") {
 				replace = "0x" + toString(_options.getRandomAddress(RandomCodeOptions::AddressType::StateAccount));
+			}
 			else if (type == "[TRANSACTIONGASLIMIT]")
-				replace = test::RandomCode::randomUniIntHex(dev::u256("5000"), dev::u256("10000000"));
+				replace = test::RandomCode::randomUniIntHex(dev::u256("25000"), dev::u256("10000000"));
 			else if (type == "[GASPRICE]")
 				replace = test::RandomCode::randomUniIntHex(0, dev::u256("10"));
 			else
@@ -176,8 +181,8 @@ std::vector<std::string> dev::test::RandomCode::getTypes()
 		"[0xHASH32]",			//Random hash string 0x...  32 byte length
 		"[V]",					//Random V value for transaction sig. could be invalid.
 		"[BLOCKGASLIMIT]",		//Random block gas limit with max of 2**55-1
-		"[DESTADDRESS]",		//Random destination address for transaction (could be empty string)
 		"[ADDRESS]",			//Random account address
+		"[DESTADDRESS]",		//Random destination address for transaction (could be empty string)
 		"[0xADDRESS]",			//Random account address
 		"[TRANSACTIONGASLIMIT]", //Random reasonable gas limit for a transaction
 		"[GASPRICE]"			//Random reasonable gas price for transaction (could be 0)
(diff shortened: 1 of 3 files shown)


test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -34,7 +34,8 @@ namespace dev { namespace test {
 bool createRandomTest()
 {
 	StateTestSuite suite;
-	dev::test::Options& options = const_cast<dev::test::Options&>(dev::test::Options::get());
+	dev::test::Options& optionsConst = const_cast<dev::test::Options&>(dev::test::Options::get());
```
winsvega: no need for cast here at all. please use dev::test::Options const&   here

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -134,16 +135,20 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
 				replace = test::RandomCode::randomUniIntHex(dev::u256("100000"), dev::u256("36028797018963967"));
 			else if (type == "[DESTADDRESS]")
 			{
-				Address address = _options.getRandomAddress(RandomCodeOptions::AddressType::PrecompiledOrStateOrCreate);
+				Address address = _options.getRandomAddress(RandomCodeOptions::AddressType::DestinationAccount);
```
winsvega: by doing this you exclude contract creation transaction and transactions directly to precompiled contracts

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -134,16 +135,20 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
...
 				if (address != ZeroAddress) //else transaction creation
 					replace = "0x" + toString(address);
 			}
-			else if (type == "[ADDRESS]")
-				replace = toString(_options.getRandomAddress(RandomCodeOptions::AddressType::StateAccount));
-			else if (type == "[0xADDRESS]")
+			else if (type == "[ADDRESS]") {
+				Address destAddress = _options.getRandomAddress(RandomCodeOptions::AddressType::StateAccount);
```
winsvega: so instead of making Options not const again. we better find a way to make test with accounts that are defined in options already set, rather then changing the options once we generate the account.

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -176,8 +181,8 @@ std::vector<std::string> dev::test::RandomCode::getTypes()
 		"[0xHASH32]",			//Random hash string 0x...  32 byte length
 		"[V]",					//Random V value for transaction sig. could be invalid.
 		"[BLOCKGASLIMIT]",		//Random block gas limit with max of 2**55-1
-		"[DESTADDRESS]",		//Random destination address for transaction (could be empty string)
 		"[ADDRESS]",			//Random account address
+		"[DESTADDRESS]",		//Random destination address for transaction (could be empty string)
```
winsvega: what is the change here?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -391,40 +399,40 @@ std::string RandomCode::fillArguments(eth::Instruction _opcode, RandomCodeOption
...
 			code += getPushCode(randUniIntGen());	//gaslimit
 			return code;
 		case eth::Instruction::SUICIDE: //(SUICIDE address)
 			code += getPushCode(toString(_options.getRandomAddress()));
 			return code;
 		case eth::Instruction::RETURN:  //(RETURN memlen1 memlen2)
 		case eth::Instruction::REVERT:  //(REVERT memlen1 memlen2)
-			code += getPushCode(randOpMemrGen());	//memlen1
-			code += getPushCode(randOpMemrGen());	//memlen1
+			code += getPushCode(randoOpSmallMemrGen());	//memlen1
```
winsvega: would it be better to change randOpMemrGen rather then creating a new dist?
is randOpMemrGen used anywhere now?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -491,6 +505,10 @@ RandomCodeOptions::RandomCodeOptions() :
 	addAddress(Address("0x0000000000000000000000000000000000000006"), AddressType::Precompiled);
 	addAddress(Address("0x0000000000000000000000000000000000000007"), AddressType::Precompiled);
 	addAddress(Address("0x0000000000000000000000000000000000000008"), AddressType::Precompiled);
+	addAddress(Address("0x0000000000000000000000000000000000000005"), AddressType::ByzantiumPrecompiled);
```
winsvega: do we really need to distinguish precompiled from byzantium and other forks ?  
you could just initialize options with only byzantium precompiles

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -520,15 +547,42 @@ Address RandomCodeOptions::getRandomAddress(AddressType _type) const
...
 			return precompiledAddressList[(int)RandomCode::randomUniInt(0, precompiledAddressList.size())];
+		case AddressType::DestinationAccount:
+			if (RandomCode::randomPercent() < emptyAddressProbability)
+				return ZeroAddress;
+			if (test::RandomCode::randomPercent() < precompiledDestProbability)
+				return precompiledAddressList[(int)RandomCode::randomUniInt(0, precompiledAddressList.size())];
+			else {
+				if (destinationAddressList.size() == 0) {
+					return ZeroAddress;
+				} else {
```
winsvega: no need for  {}  here

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -520,15 +547,42 @@ Address RandomCodeOptions::getRandomAddress(AddressType _type) const
 	{
 		case AddressType::Precompiled:
 			return precompiledAddressList[(int)RandomCode::randomUniInt(0, precompiledAddressList.size())];
+		case AddressType::DestinationAccount:
+			if (RandomCode::randomPercent() < emptyAddressProbability)
+				return ZeroAddress;
```
winsvega: I think we should add a comment that zeroAddress means contract creation if it is a transaction

test/tools/fuzzTesting/createRandomTest.cpp
cdetrio: it swaps the order that `[ADDRESS]` and `[DESTADDRESS]` are parsed in the test case template.

test/tools/fuzzTesting/fuzzHelper.cpp
cdetrio: We have to distinguish them to give Byzantium precompiles higher weights (sampling probability).

test/tools/fuzzTesting/createRandomTest.cpp
cdetrio: The "constant options" restriction results in many useless/uninteresting random tests (e.g. send a tx with data to an empty address). Using this extra `destAddress` list makes test cases where the tx destination is an account with random code.

cdetrio:
If the reason for constant options is for libfuzzer inputs https://github.com/ethereum/cpp-ethereum/issues/4515, then the `destinationAddressList` is actually fine because its deterministic (the goal of constant options for libfuzzer is deterministic test case generation, not constant options per se). Maybe we can store the `destinationAddressList` somewhere else outside of options? Then the options can stay constant.

cc @pirapira

test/tools/fuzzTesting/createRandomTest.cpp
pirapira: @cdetrio which member of the `options` class do you want to change?  You can probably use the `mutable` keyword in the declaration of `options` class to achieve your goal.  Or,you can also separate the mutable thing from the options.

pirapira:
@cdetrio constant option itself is a desired goal to keep.

test/tools/fuzzTesting/createRandomTest.cpp
cdetrio: Sorry, you'll have to ELI5 (and don't know c++). The `const_cast` was already there https://github.com/ethereum/cpp-ethereum/blob/develop/test/tools/fuzzTesting/createRandomTest.cpp#L37

test/tools/fuzzTesting/createRandomTest.cpp
cdetrio: @pirapira  its the `destinationAddressList` that needs to change, for the line `_options.addAddress(destAddress, RandomCodeOptions::AddressType::DestinationAccount);` to add an address to that list.

pirapira: @cdetrio you cannot change the command line options from inside the program.  You should copy the `destinationAddressList` to something mutable.

cdetrio: These address lists are not command line options, despite being members of `_options`.  I had to change `_options` to not-const in order to call [addAddress](https://github.com/ethereum/cpp-ethereum/blob/develop/test/tools/fuzzTesting/fuzzHelper.h#L61-L62) from createRandomTest.cpp. The other calls to addAddress work because they are in [fuzzHelper.cpp](https://github.com/ethereum/cpp-ethereum/blob/develop/test/tools/fuzzTesting/fuzzHelper.cpp#L480-L493), but they're not read from the command line.

Maybe all of the address lists can be moved into something mutable? Let's figure out what exactly needs to be constant, and what doesn't.

pirapira: Then the lists should move out of `Options` class.

pirapira: Hold on, it's not `Options` but `RandomCodeOptions`.  Then just say `std::vector<dev::Address> mutable destinationAddressList;`

test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -134,16 +135,20 @@ void dev::test::RandomCode::parseTestWithTypes(std::string& _test, std::map<std:
...
-				replace = toString(_options.getRandomAddress(RandomCodeOptions::AddressType::StateAccount));
-			else if (type == "[0xADDRESS]")
+			else if (type == "[ADDRESS]") {
+				Address destAddress = _options.getRandomAddress(RandomCodeOptions::AddressType::StateAccount);
+				_options.addAddress(destAddress, RandomCodeOptions::AddressType::DestinationAccount);
+				replace = toString(destAddress);
+			}
+			else if (type == "[0xADDRESS]") {
 				replace = "0x" + toString(_options.getRandomAddress(RandomCodeOptions::AddressType::StateAccount));
+			}
```
pirapira: Spurious `{` and `}`.  Read https://github.com/ethereum/cpp-ethereum/blob/develop/CodingStandards.txt

test/tools/fuzzTesting/createRandomTest.cpp
winsvega: why not just create a template with defined accounts. then initialize options with those accounts? 
this way options would be const and there would be no calls to non existent accounts

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: you could define options with precompiled that are exist only in Byzantine. 
the goal is to make options configurable.

PR #4588, opened 2017-10-08
add more required libs

sizeofvoid:


chfast:
They are not needed to be installed in the system.

sizeofvoid:
I'm confused, why do you difine the following in your `CMakeLists.txt`?
```
find_package(Boost REQUIRED filesystem system thread)
find_package(jsoncpp CONFIG REQUIRED)
...
```

chfast:
They are downloaded and built by Hunter package manager. You don't have to install them in the system.

PR #4592, opened 2017-10-10
copy comment filed from _info section

winsvega:
from test source to generated test

### DIFF
--- a/test/tools/libtesteth/TestSuite.cpp
+++ b/test/tools/libtesteth/TestSuite.cpp
@@ -55,24 +55,36 @@ void removeComments(json_spirit::mValue& _obj)
 	}
 }
 
-void addClientInfo(json_spirit::mValue& _v, fs::path const& _testSource, h256 const& _testSourceHash)
+void addClientInfo(json_spirit::mValue& _testDestination, json_spirit::mValue const& _testSource, fs::path const& _testSourcePath, h256 const& _testSourceHash)
 {
-	for (auto& i: _v.get_obj())
+	for (auto& i: _testDestination.get_obj())
 	{
 		json_spirit::mObject& o = i.second.get_obj();
 		json_spirit::mObject clientinfo;
 
 		string comment;
-		if (o.count("_info"))
+		if (o.count("_info")) // take the comment from generated test file
 		{
-			json_spirit::mObject& existingInfo = o["_info"].get_obj();
-			if (existingInfo.count("comment"))
-				comment = existingInfo["comment"].get_str();
+			json_spirit::mObject const& info = o["_info"].get_obj();
+			comment = info.count("comment") ? info.at("comment").get_str() : string();
+		}
+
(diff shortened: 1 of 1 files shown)


test/tools/libtesteth/TestSuite.cpp
```diff
@@ -55,24 +55,36 @@ void removeComments(json_spirit::mValue& _obj)
...
 		string comment;
-		if (o.count("_info"))
+		if (o.count("_info")) // take the comment from generated test file
 		{
-			json_spirit::mObject& existingInfo = o["_info"].get_obj();
-			if (existingInfo.count("comment"))
-				comment = existingInfo["comment"].get_str();
+			json_spirit::mObject const& info = o["_info"].get_obj();
+			comment = info.count("comment") ? info.at("comment").get_str() : string();
+		}
```
pirapira: Let's just take comments from the sources only.

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -55,24 +55,36 @@ void removeComments(json_spirit::mValue& _obj)
...
+		// copy comment from test source if there is one
+		for (auto& j: _testSource.get_obj())
+		{
+			json_spirit::mObject const& o2 = j.second.get_obj();
+			if (o2.count("_info") && j.first == i.first)
+			{
+				json_spirit::mObject const& existingInfo = o2.at("_info").get_obj();
+				if (existingInfo.count("comment"))
+					comment = existingInfo.at("comment").get_str();
+				break;
```
pirapira: Why is the first comment taken, and the rest is ignored?

winsvega: the comment from exact tetst is taken.   
the loop is for if  (info exist and test name is the same) 
searching for the test in filler file

pirapira: I think we don't need to find the test again for this purpose.  When we are filling the test, we have access to the original filler and the generated test.  Let's copy comments there.

pirapira: OK, maybe the current form is the easiest for now.

winsvega: we dont really have an access 
json_spirit::mValue output = doTests(v, true);
addClientInfo(output, v, boostRelativeTestPath, sha3(byteContents));

first all tests are being executed then we add comments to all tests

winsvega: to do this as you say we better copy the info field when filling the test 
then restore the original logic as it was

pirapira:
I will try to remove the funciton `addClientInfo` and make that a part of `doTests()` in a separate PR.

winsvega:
no need to do that. we just have to properly copy _info  section when filling the test in doTest function in all test suites. 
otherwise you will have a copy of addClientInfo in all test suite doTest function

PR #4594, opened 2017-10-11
Disable framing in devp2p for all protocol versions

gumb0:
This was preventing from succesfully communicating with devp2p v5 geth nodes

### DIFF
--- a/libp2p/Session.h
+++ b/libp2p/Session.h
@@ -90,7 +90,7 @@ public:
 class Session: public SessionFace, public std::enable_shared_from_this<SessionFace>
 {
 public:
-	static bool isFramingAllowedForVersion(unsigned _version) { return _version > 4; }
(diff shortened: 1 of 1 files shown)


libp2p/Session.h
```diff
@@ -90,7 +90,7 @@ class SessionFace
 class Session: public SessionFace, public std::enable_shared_from_this<SessionFace>
 {
 public:
-	static bool isFramingAllowedForVersion(unsigned _version) { return _version > 4; }
+	static bool isFramingAllowedForVersion(unsigned _version) { return false; }
```
chfast: You have to ignore the argument. `(void)_version`.

chfast:
Let's ship this change. Later we should probably drop this feature.

chfast:
CC @gluk256.

PR #4605, opened 2017-10-16
Upgrade libff

chfast:


pirapira:
I restarted the failing Travis build (macos).

axic:
For the record, this has been used to build a new binary that is being used for 2-3 days now in the Solidity tests and the illegal instruction failure has not occurred since.

chfast:
I rebased to fix macOS build.

PR #4609, opened 2017-10-18
Remove some ThreadSanitizer warnings

pirapira:
This should be the easier part of #4579.  Adding mutex or atomic, avoiding deadlocks, or stopping threads before vptr table is destroyed.

### DIFF
--- a/libethereum/EthereumHost.h
+++ b/libethereum/EthereumHost.h
@@ -132,7 +132,6 @@ private:
 	bool m_newTransactions = false;
 	bool m_newBlocks = false;
 
-	mutable RecursiveMutex x_sync;
 	mutable Mutex x_transactions;
 	std::unique_ptr<BlockChainSync> m_sync;
 	std::atomic<time_t> m_lastTick = { 0 };
--- a/libp2p/Host.h
+++ b/libp2p/Host.h
@@ -285,6 +285,9 @@ private:
 	/// Get or create host identifier (KeyPair).
 	static KeyPair networkAlias(bytesConstRef _b);
 
+	/// returns true if a member of m_requiredPeers
+	bool isRequiredPeer(NodeID const&) const;
+
 	bytes m_restoreNetwork;										///< Set by constructor and used to set Host key and restore network peers & nodes.
 
 	std::atomic<bool> m_run{false};													///< Whether network is running.
@@ -317,7 +320,7 @@ private:
 	
 	/// Peers we try to connect regardless of p2p network.
 	std::set<NodeID> m_requiredPeers;
-	Mutex x_requiredPeers;
+	mutable Mutex x_requiredPeers;
 
 	/// The nodes to which we are currently connected. Used by host to service peer requests and keepAlivePeers and for shutdown. (see run())
 	/// Mutable because we flush zombie entries (null-weakptrs) as regular maintenance from a const method.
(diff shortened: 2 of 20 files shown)


pirapira:
I don't know why circleCI->macOS-XCode9 says failed.  It's still compiling the codebase.

libp2p/Host.h
```diff
@@ -317,7 +317,9 @@ class Host: public Worker
 	
 	/// Peers we try to connect regardless of p2p network.
 	std::set<NodeID> m_requiredPeers;
-	Mutex x_requiredPeers;
+	Mutex mutable x_requiredPeers;
+	/// returns true if a member of m_requiredPeers
+	bool isRequiredPeer(NodeID const&) const;
```
gumb0: I'd prefer not to intermix data member declarations with method declarations (so I'd move this line above all data members)

Also `mutable` is usually before the type name.

pirapira: OK.

libethereum/EthereumHost.h
```diff
@@ -132,7 +132,6 @@ class EthereumHost: public p2p::HostCapability<EthereumPeer>, Worker
 	bool m_newTransactions = false;
 	bool m_newBlocks = false;
 
-	mutable RecursiveMutex x_sync;
```
gumb0: I have some doubts about removing this mutes. Some operations that were atomic now are not atomic...

At least let's add `RecursiveGuard l(x_sync);` inside `BlockChainSync::competeSync()` and `BlockChainSync::abortSync()`, otherwise they now stay unprotected

pirapira: Yes, I think I should do the "at least" plan.

pirapira: Oh, those `RecursiveGuard`s are already added on `develop`.

pirapira:
Ready to merge?

pirapira:
Rebased to fix the Travis macos build.

pirapira:
Now CircleCI's macos fails.

pirapira:
```
  1/398 Test   #1: RandomCodeTests/rndCode ...............................................................***Exception: SegFault  1.81 sec
```

pirapira:
I started
```
$ while true; do ./testeth -t 'RandomCodeTests/rndCode' &>> rndCode.log; done
```
locally.

pirapira:
@winsvega does 'testeth -t RandomCodeTests/rndCode' generate a different test every time it is executed?

pirapira:
I cannot reproduce the problem.  I need to try on my mac when I can reach it.

chfast:
I noticed the same failure in other branch.

winsvega:
the test is different every time thats right. so it potentially could have generated a test which failed on mac. 
the test case is just a unit test for this function. because this function should produce state test output.

chfast:
No. This is failing all the time.

pirapira:
I cannot reproduce this issue with
```
Apple LLVM version 9.0.0 (clang-900.0.38)
```

On CircleCI, I see
```
Apple LLVM version 9.0.0 (clang-900.0.37)
```

Maybe I should have a look at Apple LLVM's release notes.

chfast:
On Circle CI there is an option to "Rebuild with SSH". It will add your GitHub public key to the container and you will be able to login there with SSH.
Maybe after the build you will be able to inspect the crash output or use gdb/lldb.

Issue #4602, opened 2017-10-15
CMake: Build errors if project path contains spaces


JohnnyGlobe:
hi everyone

I am trying to build ethereum on elementry OS following this guide: https://github.com/ethereum/cpp-ethereum/, when i get to the step "cmake --build ." I keep running into this error:

[ 5%] Completed 'jsonrpccpp'
[ 6%] Built target jsonrpccpp
[ 9%] Built target mpir
[ 12%] Built target snark
/bin/sh: 1: Syntax error: Unterminated quoted string
CMakeFiles/BuildInfo.h.dir/build.make:57: recipe for target 'CMakeFiles/BuildInfo.h' failed
make[2]: [CMakeFiles/BuildInfo.h] Error 2
CMakeFiles/Makefile2:247: recipe for target 'CMakeFiles/BuildInfo.h.dir/all' failed
make[1]: [CMakeFiles/BuildInfo.h.dir/all] Error 2
Makefile:138: recipe for target 'all' failed
make: * [all] Error 2

I cant seem to find online what the cause is, any ideas?
thanks!

chfast:
I have no idea what's going on there, but can you do `ls -l /bin/sh`?

JohnnyGlobe:
Thanks for the reply chfast!
I get this from doing ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Jul 14 22:24 /bin/sh -> dash

chfast:
Please try following commands to give us more information:

1. In build dir do `make BuildInfo.h VERBOSE=1`
2. `cat CMakeCache.txt`
3. Do you by chance have you spaces in the project dir path?

JohnnyGlobe:
It looks like BuildInfo.h was not generated... it is not in the build directory

No I do not have spaces

By doing cat CMakeCache.txt i get:

# This is the CMakeCache file.
# For build in directory: /home/john/John's Linux Folder/cpp-ethereum/build
# It was generated by CMake: /usr/local/bin/cmake
# You can edit this file to change values found and used by cmake.
# If you do not want to change any of the values, simply exit the editor.
# If you do want to change a value, simply edit, save, and exit the editor.
# The syntax for the file is as follows:
# KEY:TYPE=VALUE
# KEY is the name of a variable in the cache.
# TYPE is a hint to GUIs for the type of VALUE, DO NOT EDIT TYPE!.
# VALUE is the current value for the KEY.

########################
# EXTERNAL cache entries
########################

[...]

PR #4611, opened 2017-10-19
Fix BuildInfo CMake custom target

chfast:
I don't fully understand why, but it's better to pass strings to external commands without "".

Fixes https://github.com/ethereum/cpp-ethereum/issues/4602.


chfast:
If you see carefully, this is actually the other way around :) I had to remove `""` because CMake was escaping spaces with `\`. When you are using shell, arguments are not separated by spaces, so single argument with spaces is not confusing.

PR #4612, opened 2017-10-19
make RandomCode a singleton

winsvega:
make RandomCode a single tone. 
Plan is to make two different implementations of RandomCode class 
1) random code generation using boost random distribs
2) deterministic code generation using seed.


test/tools/fuzzTesting/randomCode.h
```diff
@@ -0,0 +1,170 @@
+/*
```
pirapira: The filename should be `RandomCode.h`.

PR #4614, opened 2017-10-20
separate boost random code logic

winsvega:
with this we are ready to implement deterministic random code which will produce a state test from given byte sequence seed

### DIFF
--- a/test/tools/fuzzTesting/BoostRandomCode.cpp
+++ b/test/tools/fuzzTesting/BoostRandomCode.cpp
@@ -1,0 +1,54 @@
+/*
+	This file is part of cpp-ethereum.
+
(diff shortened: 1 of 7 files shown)


pirapira:
I scheduled a "rebuild with SSH" to debug the failing test.

winsvega:
which test ?

pirapira:
`ci/circleci: macOS-XCode9`.  When I log in to the instance with SSH, I cannot reproduce the issue.  It's weird.  Maybe we should turn on ASAN for this instance and see what's printed.

pirapira:
I give up debugging this.  I could not reproduce the issue with
* "rebuild with SSH" or
* adding address sanitizer in the cmake options

PR #4615, opened 2017-10-20
print state root after json trace

cdetrio:
This might break change logs, but it demonstrates the state root needed for [standard traces](https://github.com/ethereum/tests/issues/249#issuecomment-331272388).


holiman:
:+1:

winsvega:
Have to check how --jsontrace  is functioning after this check

pirapira:
@winsvega have you checked what you wanted to check?

winsvega:
looks like --statediff option requre Permanence::Uncommitted
this change affects only --jsontrace ''  which seems to work just fine.

gumb0:
What's the status of this, are we going to merge it? @winsvega

PR #4617, opened 2017-10-23
make TestOutputHelper a singletone

winsvega:
first make sure that this bug is happening by default.

I have an assumption that the reason of the bug might be due to the TestOutputHelper static fields initialization. Will check in this PR.


pirapira:
Why is it called [wip]?

winsvega:
because I was waiting for macOS build to become green.
It looks like it has resolved itself independent of this PR

PR #4621, opened 2017-10-24
Correct gas schedule for ECMUL

pirapira:
I haven't run tests.  So we should wait for the CIs at least.


axic:
For reference: as discovered in https://github.com/ethereum/solidity/pull/3114

PR #4597, opened 2017-10-12
Better command line argument handling  #3628

demon1999:
Resolves #3628

PR #4622, opened 2017-10-24
Download snapshot through Parity's warp protocol

gumb0:
This is the solution using boost.fiber

### Motivation for using boost.fiber

The approach that `BlockChainSync` class takes to deal with asynchronous p2p nature of the sync looks like an anti-pattern that led to it becoming utterly unmaintainable. That approach could be described as having several network message handlers in one class, saving info about current stage of the sync in a bunch of member variables in the class, branching a lot in each handler depending on the current state of member variables. This way the essence of the sync algorithm itself gets spread out over a lot of different parts of the class, it's hard to keep track of what's currently going on and what should happen next. 

I am looking for a better approach that can elegantly express what we want to achieve. Fibers allow us here to have a "main algorithm" in one fiber that gets interrupted each time we need to get some data asynchronously, then gets resumed from the place we left off when we have data already. So the main algorithm looks like it's synchronous. 
The data itself is sent to fiber through `buffered_channel` - producers/consumers-kind of data structure, where the fiber gets "blocked" if it tries to get the data not pushed yet to the channel. 
To some extent I think this is similar to goroutines and channels of golang.

Also switching context between the fibers is very cheap unlike with threads, and it all happens in the single thread, therefore no need to deal with thread-safety/possible races/mutexes etc.

This is still kind of experiment to see where it goes, I myself am not fully convinced that fibers are the best solution for this. 

### Current status:
- It is able to download the snapshot from from several peers at once (this is the improvement over the version in https://github.com/ethereum/cpp-ethereum/pull/4227)
- It starts downloading from the first peer found that supports warp protocol (this probably should be changed to better strategy in later PRs) It's better currently to run it giving the peer address from the needed network with `--peerset required:`
- Peer not responding to request should be handled correctly, but I haven't seen this in practice and still need to test if this works.

### Still to be done:
- [x] Probably peer disconnecting should be handled better - if it happens after we reqested a chunk, I think chunk will stay marked as being downloaded forever.
- [x] Without `--peerset` we connect to Ethereum Classic nodes very often and happily download the snapshot for classic chain. To deal with this probably we need to do DAO challenge first similar to full sync.
- [x] Need to add boost.fiber to hunter's boost to be able to link at all (no idea yet how to do it)
- [x] ~~fix macOS build failure because of unused `SnapshotLog::debug` - this doesn't make sense to me yet~~ looks like I've already found a workaround doing import https://github.com/ethereum/cpp-ethereum/blob/04e8ca5afe785cafe18905d6f0f032cb1d869c7a/libethereum/SnapshotImporter.cpp#L45
- [ ] tests

### DIFF
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -856,17 +856,9 @@ int main(int argc, char** argv)
         chainParams.allowFutureBlocks = true;
     }
 
-    dev::WebThreeDirect web3(
-        WebThreeDirect::composeClientVersion("eth"),
-        getDataDir(),
-        snapshotPath,
-        chainParams,
-        withExisting,
-        nodeMode == NodeMode::Full ? caps : set<string>(),
-        netPrefs,
-        &nodesState,
-        testingMode
-    );
+    dev::WebThreeDirect web3(WebThreeDirect::composeClientVersion("eth"), getDataDir(),
+        snapshotPath, chainParams, withExisting, nodeMode == NodeMode::Full ? caps : set<string>(),
+        netPrefs, &nodesState, testingMode);
 
     if (!extraData.empty())
         web3.ethereum()->setExtraData(extraData);
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -342,6 +342,7 @@ int main(int argc, char** argv)
         ("only", po::value<string>()->value_name("<n>"), "Equivalent to --export-from n --export-to n.")
         ("format", po::value<string>()->value_name("<binary/hex/human>"), "Set export format.")
         ("dont-check", "Prevent checking some block aspects. Faster importing, but to apply only when the data is known to be valid.")
+        ("download-snapshot", po::value<string>()->value_name("<path>"), "Download Parity Warp Sync snapshot data to the specified path.")
         ("import-snapshot", po::value<string>()->value_name("<path>"), "Import blockchain and state data from the Parity Warp Sync snapshot.\n");
 
     po::options_description generalOptions("General Options", c_lineWidth);
@@ -382,6 +383,9 @@ int main(int argc, char** argv)
             return -1;
         }
 
+    std::string snapshotPath;
+    if (vm.count("download-snapshot"))
+        snapshotPath = vm["download-snapshot"].as<string>();
     if (vm.count("import-snapshot"))
     {
         mode = OperationMode::ImportSnapshot;
@@ -855,6 +859,7 @@ int main(int argc, char** argv)
     dev::WebThreeDirect web3(
         WebThreeDirect::composeClientVersion("eth"),
         getDataDir(),
+        snapshotPath,
         chainParams,
         withExisting,
         nodeMode == NodeMode::Full ? caps : set<string>(),
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -333,6 +333,7 @@ int main(int argc, char** argv)
         ("no-discovery",  "Disable node discovery, implies --no-bootstrap.")
         ("pin",  "Only accept or connect to trusted peers.");
 
+    std::string snapshotPath;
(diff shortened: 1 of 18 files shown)


axic:
Is there an EIP about the warp protocol? Wouldn't it make sense creating one?

cc @arkpar @karalabe

gumb0:
Now there's only parity's doc https://github.com/paritytech/parity/wiki/Warp-Sync

eth/main.cpp
```diff
@@ -148,6 +148,7 @@ void help()
 		<< "    --to <n>  Export only to block n (inclusive); n may be a decimal, a '0x' prefixed hash, or 'latest'.\n"
 		<< "    --only <n>  Equivalent to --export-from n --export-to n.\n"
 		<< "    --dont-check  Prevent checking some block aspects. Faster importing, but to apply only when the data is known to be valid.\n\n"
+		<< "    --download-snapshot <path> Download Parity Warp snapshot data to the specified path." << endl
```
pirapira: I spent so much time removing `endl`s.

gumb0: Sorry I'll fix this 😆 
Anyway I hope this to go away after https://github.com/ethereum/cpp-ethereum/pull/4597

libethereum/WarpHostCapability.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include <libdevcore/Worker.h>
+
+namespace dev
+{
+
+namespace eth
+{
+
+class WarpHostCapability: public p2p::HostCapability<WarpPeerCapability>, Worker
```
pirapira: Since this inherits `Worker`, there should be a destructor that calls `terminate()` (bad design it is...).

gumb0: Oh right, thanks

chfast:
Using Fiber sounds like interesting experiment. They seems to work nice together with boost.asio.
This video have an idea how to manage lifetime of connection handlers: https://www.youtube.com/watch?v=rwOv_tw2eA4.

gumb0:
Another relevant talk: https://www.youtube.com/watch?v=mCD6VLVS_y4
(he starts to talk about fibers only on minute 44 though)

gumb0:
Looks like boost.fiber from hunter now fails to build on GCC 4.8 (including Travis) but builds on GCC 5

gumb0:
Might be related: https://github.com/boostorg/fiber/issues/121 but should be fixed in boost 1.66 used here

gumb0:
Looks like boost.fiber starting with boost 1.66 requires `std::regex` support and GCC 4.8 doesn't support it

gumb0:
I pinned boost version to 1.65.1 for now, it seems to work fine on GCC 4.8
It might be resolved in boost.fiber, see https://github.com/boostorg/fiber/pull/152
Otherwise we would need to give up GCC 4 support

chfast:
I don't see problem with going with 1.65.1 for now and then skip 1.66.

gumb0:
I might add some tests/minor tweaks, but overall it's ready to be reviewed I think.

gumb0:
A couple of disadvantages of using fibers that I see, to be fair:
- Data from network messages might be copied more often. That is, when all the logic of network algorithm is in network handlers (like in `BlockChainSync`) we often can accesss the data through `RLP` class only, that is using only with the view of the data without additional copying. Here we immediately copy this data (into futures or channels) to be handled by the different fiber later.
- Debugging might be difficult, because debuggers don't have native support for the fibers, they can't show which fibers exist at the moment and at which point is each fiber suspended.
Theoretically we should be able to see current fibers list if we build debug boost.fibers and look up the stack, but I didn't try this.

gumb0:
Issues fixed, I'll do the changes around reformatting `program_options`-setup code in a separate PR

chfast:
Let's ship it.

chfast:
:shipit:

PR #4624, opened 2017-10-24
Fix an assertion that sometimes fails

pirapira:
Since `randomPercent()` can be 100, the computed `size` can be identical to `_maxOpNumber`.


chfast:
Can this be a reason for macOS random failures?

pirapira:
This one causes SIGABRT, but those random failures were SIGSEGV.

winsvega:
when did this caused SIGABRT ?

pirapira:
When I ran
```
while true; do ./testeth -t 'RandomCodeTests/rndCode' || break ; done
```

PR #4625, opened 2017-10-24
Support high CHAIN_ID encoded in Transaction's v value

gumb0:
EIP155 is not clear about the possible ranges of `v` 
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md

But other clients support high values, so we should, too.

### DIFF
--- a/libethcore/TransactionBase.cpp
+++ b/libethcore/TransactionBase.cpp
@@ -70,13 +70,13 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
 
 		if (isZeroSignature(r, s))
 		{
-			m_chainId = static_cast<int>(v);
+			m_chainId = v;
 			m_vrs = SignatureStruct{r, s, 0};
 		}
 		else
 		{
 			if (v > 36)
-				m_chainId = static_cast<int>((v - 35) / 2); 
+				m_chainId = (v - 35) / 2; 
 			else if (v == 27 || v == 28)
 				m_chainId = -4;
 			else
--- a/libethcore/TransactionBase.cpp
+++ b/libethcore/TransactionBase.cpp
@@ -76,7 +76,7 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
 		else
 		{
 			if (v > 36)
-				m_chainId = static_cast<int>((v - 35) / 2);
+				m_chainId = static_cast<int>((v - 35) / 2); 
 			else if (v == 27 || v == 28)
 				m_chainId = -4;
 			else
@@ -167,8 +167,14 @@ void TransactionBase::streamRLP(RLPStream& _s, IncludeSignature _sig, bool _forE
 		if (!m_vrs)
 			BOOST_THROW_EXCEPTION(TransactionIsUnsigned());
 
-		int vOffset = m_chainId*2 + 35;
-		_s << (m_vrs->v + vOffset) << (u256)m_vrs->r << (u256)m_vrs->s;
+		if (hasZeroSignature())
+			_s << m_chainId;
+		else
+		{
+			int const vOffset = m_chainId * 2 + 35;
+			_s << (m_vrs->v + vOffset);
+		}
+		_s << (u256)m_vrs->r << (u256)m_vrs->s;
 	}
 	else if (_forEip155hash)
 		_s << m_chainId << 0 << 0;
--- a/libethcore/TransactionBase.cpp
+++ b/libethcore/TransactionBase.cpp
@@ -64,23 +64,25 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
 
 		m_data = rlp[5].toBytes();
 
-		byte v = rlp[6].toInt<byte>();
-		h256 r = rlp[7].toInt<u256>();
-		h256 s = rlp[8].toInt<u256>();
+		int const v = rlp[6].toInt<int>();
+		h256 const r = rlp[7].toInt<u256>();
+		h256 const s = rlp[8].toInt<u256>();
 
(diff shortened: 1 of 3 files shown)


libethcore/TransactionBase.cpp
```diff
@@ -64,23 +64,25 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
 
 		m_data = rlp[5].toBytes();
 
-		byte v = rlp[6].toInt<byte>();
-		h256 r = rlp[7].toInt<u256>();
-		h256 s = rlp[8].toInt<u256>();
+		u256 const v = rlp[6].toInt<u256>();
```
chfast: If we use `int` later on, can we use `toInt<int>()` here?

gumb0: Yes, I think it makes sense

That way we'll fail when parsing RLP if it's bigger than `int` instead of silently overflowing

gumb0:
I don't want to deal with updating the tests right now
@winsvega or @pirapira It would be great if you could help me here

winsvega:
does it fail on current tests?

gumb0:
It does but looks like only on ZeroSig transactions, looks like I broke them and I see where, let me try to fix it...

There were more failures when I parsed `v` as `u256` and not like `int`, so I thought maybe more changes in the tests needed...

winsvega:
if zero sig are supported with constantinople flag enabled then those tests should not fail.

chfast:
Still broken. You should first try locally.

gumb0:
Ok I think we have a bug in the code for creating RLP of ZeroSig Transaction
https://github.com/ethereum/cpp-ethereum/blob/70d23560a51af2240a445370cd27640db621f20e/libethcore/TransactionBase.cpp#L163-L170

For ZeroSig `m_vrs->v == m_chainId` (before my modifications in this PR) 
So instead of writing `CHAIN_ID` to RLP  in this case we write `CHAIN_ID + CHAIN_ID * 2 + 35` I think.

If I'm right I'll create a fix here and we will need to update Constantinople Transaction tests.

gumb0:
(My modifications here assigned `m_vrs->v = 0` for ZeroSig and this affected RLP and this affected the transaction hash wich the tests are complaining about)

winsvega:
we have to check that with other client implementations.

gumb0:
Anyway parsing Transaction RLP should be symmetrical to creating RLP

gumb0:
(locally running `testeth.exe -t TransactionTests` iin this branch spits out a lot of `Filler hash is different!` errors, no idea why. `git submodule update` doesn't help)

winsvega:
that should not happen actually. filler hash is different means you have different version of tests sources and compiled tests. let me see.

winsvega:
ok. the tests are failing because you changed the way transaction hash is calculated for zeroSig transactions. 

Vitalik said zero signature transaction EIP is not yet confirmed. thus we could either update the tests or delete those.

winsvega:
run your PR on this test commit
https://github.com/ethereum/tests/pull/368

pirapira:
Let's just update the tests.  It's a waste of time to revert decisions.

winsvega:
@pirapira 
I want to place zeroSigTransaction tests into a pending PR untill we clarify the EIP realisation

pirapira:
@winsvega OK.

libethcore/TransactionBase.cpp
```diff
@@ -64,23 +64,25 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
...
+		if (isZeroSignature(r, s))
+		{
+			m_chainId = static_cast<int>(v);
+			m_vrs = SignatureStruct{r, s, 0};
+		}
 		else
 		{
 			if (v > 36)
-				m_chainId = (v - 35) / 2;
+				m_chainId = static_cast<int>((v - 35) / 2);
```
pirapira: Why do you need a cast here?  `v` is `int`, `35` and `2` are both `int`, I think.

winsvega: because of the division?

gumb0: @pirapira You are right, it's not needed, I'll remove it

libethcore/TransactionBase.cpp
```diff
@@ -64,23 +64,25 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
...
+		h256 const r = rlp[7].toInt<u256>();
+		h256 const s = rlp[8].toInt<u256>();
 
-		m_vrs = SignatureStruct{ r, s, v };
-
-		if (hasZeroSignature())
-			m_chainId = m_vrs->v;
+		if (isZeroSignature(r, s))
+		{
+			m_chainId = static_cast<int>(v);
```
axic: `v` is already `int`, no need to cast here?

gumb0: Right, thanks, this is a leftover from the version when `v` was `u256`

gumb0:
I've updated jsontests submodule to point to https://github.com/ethereum/tests/pull/368, that is tests version without ZeroSig tests.
It required to remove the test case definition from TransactionTests.cpp.
Please review.

libethcore/TransactionBase.cpp
```diff
@@ -64,23 +64,25 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
...
+		if (isZeroSignature(r, s))
+		{
+			m_chainId = v;
+			m_vrs = SignatureStruct{r, s, 0};
+		}
 		else
 		{
 			if (v > 36)
-				m_chainId = (v - 35) / 2;
+				m_chainId = (v - 35) / 2;
```
pirapira: nitpicking: a space at the end.

PR #4627, opened 2017-10-24
add --seed option to --createRandomTest

winsvega:
usage

```
./testeth -t GeneralStateTests -- --createRandomTest --seed 123457
```

### DIFF
--- a/test/tools/fuzzTesting/BoostRandomCode.cpp
+++ b/test/tools/fuzzTesting/BoostRandomCode.cpp
@@ -38,9 +38,14 @@ u256 BoostRandomCode::randomUniInt(u256 const& _minVal, u256 const& _maxVal)
 
 void BoostRandomCode::refreshSeed()
 {
-	auto now = std::chrono::steady_clock::now().time_since_epoch();
-	auto timeSinceEpoch = std::chrono::duration_cast<std::chrono::nanoseconds>(now).count();
-	gen.seed(static_cast<unsigned int>(timeSinceEpoch));
+	if (Options::get().randomTestSeed == 0)
+	{
+		auto now = std::chrono::steady_clock::now().time_since_epoch();
+		auto timeSinceEpoch = std::chrono::duration_cast<std::chrono::nanoseconds>(now).count();
+		gen.seed(static_cast<unsigned int>(timeSinceEpoch));
+	}
+	else
+		gen.seed(Options::get().randomTestSeed);
 }
 
 uint8_t BoostRandomCode::weightedOpcode(std::vector<int>& _weights)
--- a/test/tools/fuzzTesting/BoostRandomCode.cpp
+++ b/test/tools/fuzzTesting/BoostRandomCode.cpp
@@ -38,14 +38,14 @@ u256 BoostRandomCode::randomUniInt(u256 const& _minVal, u256 const& _maxVal)
 
 void BoostRandomCode::refreshSeed()
 {
-	if (Options::get().randomTestSeed == 0)
+	if (!Options::get().randomTestSeed.is_initialized())
 	{
 		auto now = std::chrono::steady_clock::now().time_since_epoch();
 		auto timeSinceEpoch = std::chrono::duration_cast<std::chrono::nanoseconds>(now).count();
 		gen.seed(static_cast<unsigned int>(timeSinceEpoch));
 	}
 	else
-		gen.seed(Options::get().randomTestSeed);
+		gen.seed(Options::get().randomTestSeed.get());
 }
 
 uint8_t BoostRandomCode::weightedOpcode(std::vector<int>& _weights)
--- a/test/tools/fuzzTesting/fuzzHelper.cpp
+++ b/test/tools/fuzzTesting/fuzzHelper.cpp
@@ -257,32 +257,36 @@ std::string RandomCodeBase::generate(int _maxOpNumber, RandomCodeOptions const& 
 		eth::Instruction inst = (eth::Instruction) opcode;
 		eth::InstructionInfo info = eth::instructionInfo(inst);
 
-		if (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
+		while (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
 			|| std::find(invalidOpcodes.begin(), invalidOpcodes.end(), inst) != invalidOpcodes.end())
 		{
 			if (_options.useUndefinedOpCodes)
+			{
 				code += toCompactHex(opcode, 1);
+				return "0x" + code;
+			}
 			else
 			{
 				//Byte code is yet not implemented. do not count it.
-				i--;
+				opcode++;
+				inst = (eth::Instruction) opcode;
+				info = eth::instructionInfo(inst);
 				continue;
 			}
 		}
+
+		if (info.name.find("PUSH") != std::string::npos)
+		{
+			code += toCompactHex(opcode);
+			code += fillArguments(inst, _options);
+		}
 		else
 		{
-			if (info.name.find("PUSH") != std::string::npos)
-			{
-				code += toCompactHex(opcode);
-				code += fillArguments(inst, _options);
-			}
-			else
-			{
-				code += fillArguments(inst, _options);
-				code += toCompactHex(opcode, 1);
-			}
+			code += fillArguments(inst, _options);
+			code += toCompactHex(opcode, 1);
 		}
 	}
+
 	return "0x" + code;
 }
 
--- a/test/tools/fuzzTesting/fuzzHelper.cpp
+++ b/test/tools/fuzzTesting/fuzzHelper.cpp
@@ -239,51 +239,57 @@ std::string RandomCodeBase::rndByteSequence(int _length, SizeStrictness _sizeTyp
 	return hash;
 }
 
+bool isOpcodeDefined(uint8_t _opcode)
+{
+	eth::Instruction inst = (eth::Instruction) _opcode;
+	eth::InstructionInfo info = eth::instructionInfo(inst);
+	if (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
+		|| std::find(invalidOpcodes.begin(), invalidOpcodes.end(), inst) != invalidOpcodes.end())
+		return false;
+	return true;
+}
+
+uint8_t makeOpcodeDefined(uint8_t _opcode)
+{
+	while (!isOpcodeDefined(_opcode))
+		_opcode++; //Byte code is yet not implemented. Try next one.
+	return _opcode;
+}
+
 //generate smart random code
 std::string RandomCodeBase::generate(int _maxOpNumber, RandomCodeOptions const& _options)
 {
 	std::string code;
-
 	if (test::RandomCode::get().randomPercent() < _options.emptyCodeProbability)
 		return code;
 
-	//random opCode amount
+	//generate [0 ... _maxOpNumber] opcodes.
 	int size = (int)(test::RandomCode::get().randomPercent() * _maxOpNumber / 100);
 	assert(size <= _maxOpNumber);
 
 	for (auto i = 0; i < size; i++)
 	{
 		uint8_t opcode = _options.getWeightedRandomOpcode();
-		eth::Instruction inst = (eth::Instruction) opcode;
-		eth::InstructionInfo info = eth::instructionInfo(inst);
-
-		while (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
-			|| std::find(invalidOpcodes.begin(), invalidOpcodes.end(), inst) != invalidOpcodes.end())
+		if (!isOpcodeDefined(opcode) && _options.useUndefinedOpCodes)
 		{
-			if (_options.useUndefinedOpCodes)
-			{
-				code += toCompactHex(opcode, 1);
-				return "0x" + code;
-			}
-			else
-			{
-				//Byte code is yet not implemented. do not count it.
-				opcode++;
-				inst = (eth::Instruction) opcode;
-				info = eth::instructionInfo(inst);
-				continue;
-			}
-		}
-
-		if (info.name.find("PUSH") != std::string::npos)
-		{
-			code += toCompactHex(opcode);
-			code += fillArguments(inst, _options);
+			code += toCompactHex(opcode, 1);
+			continue;
 		}
 		else
 		{
-			code += fillArguments(inst, _options);
-			code += toCompactHex(opcode, 1);
+			opcode = makeOpcodeDefined(opcode);
+			eth::Instruction inst = (eth::Instruction) opcode;
+			eth::InstructionInfo info = eth::instructionInfo(inst);
+			if (info.name.find("PUSH") != std::string::npos)
+			{
+				code += toCompactHex(opcode);
+				code += fillArguments(inst, _options);
+			}
+			else
+			{
+				code += fillArguments(inst, _options);
+				code += toCompactHex(opcode, 1);
+			}
 		}
 	}
 
--- a/test/tools/fuzzTesting/fuzzHelper.cpp
+++ b/test/tools/fuzzTesting/fuzzHelper.cpp
@@ -243,10 +243,9 @@ bool isOpcodeDefined(uint8_t _opcode)
 {
 	eth::Instruction inst = (eth::Instruction) _opcode;
 	eth::InstructionInfo info = eth::instructionInfo(inst);
-	if (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
-		|| std::find(invalidOpcodes.begin(), invalidOpcodes.end(), inst) != invalidOpcodes.end())
-		return false;
-	return true;
+	bool isUndefined = (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
+		|| std::find(invalidOpcodes.begin(), invalidOpcodes.end(), inst) != invalidOpcodes.end());
+	return !isUndefined;
 }
 
 uint8_t makeOpcodeDefined(uint8_t _opcode)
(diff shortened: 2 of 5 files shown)


test/tools/fuzzTesting/fuzzHelper.h
```diff
@@ -47,14 +48,20 @@ class RandomCode
...
 	RandomCode(RandomCode const&) = delete;
 	void operator=(RandomCode const&) = delete;
+	~RandomCode() { delete generator; }
 
 private:
-	RandomCode(){}
-	BoostRandomCode generator;
+	RandomCode(){
+		if (Options::get().randomTestSeed.empty())
+			generator = new BoostRandomCode();
```
chfast: This abstraction is stupid. Only use DeterministicRandomCode (with better name).
If seed is empty just pick a random one (you can use random_device for it).

test/tools/fuzzTesting/fuzzHelper.h
```diff
@@ -47,14 +48,20 @@ class RandomCode
 	static RandomCodeBase& get()
 	{
 		static RandomCode instance;
-		return instance.generator;
+		return *instance.generator;
 	}
 	RandomCode(RandomCode const&) = delete;
 	void operator=(RandomCode const&) = delete;
+	~RandomCode() { delete generator; }
```
chfast: No kidding. This is no go. Use `std::unique_ptr`. But probably not needed at all according to my other comments.

test/tools/libtesteth/Options.cpp
```diff
@@ -205,6 +206,11 @@ Options::Options(int argc, char** argv)
 		}
 		else if (arg == "--createRandomTest")
 			createRandomTest = true;
+		else if (arg == "--seed")
+		{
+			throwIfNoArgumentFollows();
+			randomTestSeed = argv[++i];
```
chfast: I think the seed should be a number, not a string. Preferably `uint64_t` as [mt19937_64](http://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine) is good choice for random engine.

test/tools/fuzzTesting/fuzzHelper.h
winsvega: this is just an example. there will be a different implementation of DeterministicRandomCode

test/tools/libtesteth/Options.h
```diff
@@ -54,6 +54,7 @@ class Options
 	bool statediff = false;///< Fill full post state in General tests
 	bool fulloutput = false;///< Replace large output to just it's length
 	bool createRandomTest = false; ///< Generate random test
+	uint64_t randomTestSeed; ///< Define a seed for random test
```
chfast: Init it.

test/tools/libtesteth/Options.cpp
```diff
@@ -205,6 +207,11 @@ Options::Options(int argc, char** argv)
 		}
 		else if (arg == "--createRandomTest")
 			createRandomTest = true;
+		else if (arg == "--seed")
+		{
+			throwIfNoArgumentFollows();
+			randomTestSeed = static_cast<uint64_t>(toInt(argv[++i]));
```
chfast: I think you should fails also if the value is bigger than uint64_t.

test/tools/fuzzTesting/BoostRandomCode.cpp
```diff
@@ -38,9 +38,14 @@ u256 BoostRandomCode::randomUniInt(u256 const& _minVal, u256 const& _maxVal)
 
 void BoostRandomCode::refreshSeed()
 {
-	auto now = std::chrono::steady_clock::now().time_since_epoch();
-	auto timeSinceEpoch = std::chrono::duration_cast<std::chrono::nanoseconds>(now).count();
-	gen.seed(static_cast<unsigned int>(timeSinceEpoch));
+	if (Options::get().randomTestSeed == 0)
+	{
+		auto now = std::chrono::steady_clock::now().time_since_epoch();
```
chfast: Why have you used a clock instead of `std::random_device`?

test/tools/fuzzTesting/BoostRandomCode.cpp
```diff
@@ -38,9 +38,14 @@ u256 BoostRandomCode::randomUniInt(u256 const& _minVal, u256 const& _maxVal)
...
+		auto timeSinceEpoch = std::chrono::duration_cast<std::chrono::nanoseconds>(now).count();
+		gen.seed(static_cast<unsigned int>(timeSinceEpoch));
```
chfast: This should be cast to `gen::result_type`.

test/tools/fuzzTesting/fuzzHelper.h
```diff
@@ -47,14 +47,16 @@ class RandomCode
...
+		return *instance.generator.get();
 	}
 	RandomCode(RandomCode const&) = delete;
 	void operator=(RandomCode const&) = delete;
 
 private:
-	RandomCode(){}
-	BoostRandomCode generator;
+	RandomCode(){
+		generator.reset(new BoostRandomCode());
```
chfast: I don't see why this changes is needed now. Can we keep the old composition?

test/tools/fuzzTesting/fuzzHelper.h
```diff
@@ -47,14 +47,16 @@ class RandomCode
 	static RandomCodeBase& get()
 	{
 		static RandomCode instance;
-		return instance.generator;
+		return *instance.generator.get();
```
chfast: You don't need `.get()` here.

test/tools/libtesteth/Options.h
pirapira: Can you turn this into `boost::optional<uint64_t>`?  When this option is not given it shouldn't be zero, or any number.

test/tools/fuzzTesting/fuzzHelper.h
```diff
@@ -47,14 +47,16 @@ class RandomCode
...
 	}
 	RandomCode(RandomCode const&) = delete;
 	void operator=(RandomCode const&) = delete;
 
 private:
-	RandomCode(){}
-	BoostRandomCode generator;
+	RandomCode(){
```
pirapira: The `{` should be on the next line.

test/tools/fuzzTesting/fuzzHelper.h
```diff
@@ -47,14 +47,16 @@ class RandomCode
...
+		generator.reset(new BoostRandomCode());
+	}
+	std::unique_ptr<RandomCodeBase> generator;
```
pirapira: Why not just have
```
RandomCodeBase generator;
```
?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -257,32 +257,36 @@ std::string RandomCodeBase::generate(int _maxOpNumber, RandomCodeOptions const&
...
 		eth::InstructionInfo info = eth::instructionInfo(inst);
 
-		if (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
+		while (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
 			|| std::find(invalidOpcodes.begin(), invalidOpcodes.end(), inst) != invalidOpcodes.end())
 		{
 			if (_options.useUndefinedOpCodes)
+			{
 				code += toCompactHex(opcode, 1);
+				return "0x" + code;
```
pirapira: Why not `continue` and fill more opcodes?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -257,32 +257,36 @@ std::string RandomCodeBase::generate(int _maxOpNumber, RandomCodeOptions const&
...
+			}
 			else
 			{
 				//Byte code is yet not implemented. do not count it.
-				i--;
+				opcode++;
+				inst = (eth::Instruction) opcode;
+				info = eth::instructionInfo(inst);
 				continue;
```
pirapira: Why is `code` changed in all other cases but not here?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -257,32 +257,36 @@ std::string RandomCodeBase::generate(int _maxOpNumber, RandomCodeOptions const&
 		eth::Instruction inst = (eth::Instruction) opcode;
 		eth::InstructionInfo info = eth::instructionInfo(inst);
 
-		if (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
+		while (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
```
pirapira: Why is this a loop now, what is it iterating over?  What changes between the first run and the second run?

test/tools/fuzzTesting/fuzzHelper.h
winsvega: I plan to add one more implementation

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: random codes are genereated in for sycle. 
if a random code is undefined  then we change it to a defined opcode in a while loop untill it is defined. (if we dont accept undefined opcodes)

there was an issue with static seed. a random undefined opcode would lead to endless loop.

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: it is continued in a for loop

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: random codes are genereated in for sycle.
if a random code is undefined then we change it to a defined opcode in a while loop untill it is defined. (if we dont accept undefined opcodes)

there was an issue with static seed. a random undefined opcode would lead to endless loop.

test/tools/fuzzTesting/BoostRandomCode.cpp
winsvega: this page says it has decreasing perfomance 
http://en.cppreference.com/w/cpp/numeric/random/random_device

test/tools/fuzzTesting/BoostRandomCode.cpp
winsvega: this page says it has decreasing perfomance 
http://en.cppreference.com/w/cpp/numeric/random/random_device

test/tools/fuzzTesting/BoostRandomCode.cpp
winsvega: there is no such thing

test/tools/fuzzTesting/fuzzHelper.cpp
pirapira: No, when you say `return`, the execution returns from the function.

You should continue in the for-loop.  Do you know `break` in loops?

test/tools/fuzzTesting/fuzzHelper.cpp
pirapira: I suggest separating this `findingDefinedOpcode` feature as a function.

test/tools/fuzzTesting/fuzzHelper.h
```diff
@@ -47,14 +47,17 @@ class RandomCode
...
 	void operator=(RandomCode const&) = delete;
 
 private:
-	RandomCode(){}
-	BoostRandomCode generator;
+	RandomCode()
+	{
+		generator.reset(new BoostRandomCode());
+	}
+	std::unique_ptr<RandomCodeBase> generator;
```
pirapira: Why is this not `RandomCodeBase generator;`?

winsvega: Because there will be another implementation other than BoostRandomCode

test/tools/fuzzTesting/fuzzHelper.cpp
winsvega: hm your right. but it should exit the while loop and continue the for loop then

test/tools/fuzzTesting/fuzzHelper.h
pirapira: OK.

test/tools/fuzzTesting/fuzzHelper.cpp
pirapira: Check what `break` does in loops.

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -239,52 +239,58 @@ std::string RandomCodeBase::rndByteSequence(int _length, SizeStrictness _sizeTyp
...
-		eth::Instruction inst = (eth::Instruction) opcode;
-		eth::InstructionInfo info = eth::instructionInfo(inst);
-
-		while (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
-			|| std::find(invalidOpcodes.begin(), invalidOpcodes.end(), inst) != invalidOpcodes.end())
+		if (!isOpcodeDefined(opcode) && _options.useUndefinedOpCodes)
 		{
-			if (_options.useUndefinedOpCodes)
+			code += toCompactHex(opcode, 1);
+			continue;
```
pirapira: Here, the execution enters an infinite loop.

I think the execution should go out of the while loop.

winsvega: this continue goes to the for loop. it wont be infinite

test/tools/fuzzTesting/fuzzHelper.h
chfast: I still don't agree on this. Introduce this change when introducing the other implementation of RandomCodeBase.

winsvega: there was. during the refactoring of this PR we decided to remove it.

test/tools/fuzzTesting/fuzzHelper.cpp
pirapira: I think you are mistaken what `continue` does in nested loops.

winsvega: I have removed the nested loop. there is makeOpcodeDefined func now

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -239,52 +239,58 @@ std::string RandomCodeBase::rndByteSequence(int _length, SizeStrictness _sizeTyp
 	return hash;
 }
 
+bool isOpcodeDefined(uint8_t _opcode)
+{
+	eth::Instruction inst = (eth::Instruction) _opcode;
+	eth::InstructionInfo info = eth::instructionInfo(inst);
+	if (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
```
chfast: `if (...) return true; else return false;` is not very smart. Just return check condition value.

test/tools/libtesteth/Options.cpp
```diff
@@ -210,7 +209,7 @@ Options::Options(int argc, char** argv)
 		else if (arg == "--seed")
 		{
 			throwIfNoArgumentFollows();
-			randomTestSeed = static_cast<uint64_t>(toInt(argv[++i]));
+			randomTestSeed = static_cast<uint64_t>(min<u256>(UINT64_MAX, toInt(argv[++i])));
```
chfast: What's the difference? You sill hide the invalid value error.

test/tools/libtesteth/Options.cpp
```diff
@@ -278,7 +277,7 @@ Options::Options(int argc, char** argv)
 	}
 	else
 	{
-		if (randomTestSeed != 0)
+		if (randomTestSeed.is_initialized())
```
chfast: It can be `if (randomTestSeed)`.

test/tools/fuzzTesting/fuzzHelper.h
chfast: So also remove this.

test/tools/libtesteth/Options.cpp
winsvega: m_vrs in transaction signature also using is_initialized. 
I just want it to be clear that it is not the int value being checked but boost::optional

test/tools/libtesteth/Options.cpp
pirapira: What is hidden? Invalid value errors about negative `int`s?

test/tools/fuzzTesting/fuzzHelper.h
```diff
@@ -44,20 +44,17 @@ struct RlpDebug
...
 	RandomCode(RandomCode const&) = delete;
 	void operator=(RandomCode const&) = delete;
 
 private:
-	RandomCode()
-	{
-		generator.reset(new BoostRandomCode());
-	}
-	std::unique_ptr<RandomCodeBase> generator;
+	RandomCode(){}
```
chfast: This either should be `RandomCode() = default` or removed.

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -243,10 +243,9 @@ bool isOpcodeDefined(uint8_t _opcode)
 {
 	eth::Instruction inst = (eth::Instruction) _opcode;
 	eth::InstructionInfo info = eth::instructionInfo(inst);
-	if (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
-		|| std::find(invalidOpcodes.begin(), invalidOpcodes.end(), inst) != invalidOpcodes.end())
-		return false;
-	return true;
+	bool isUndefined = (info.name.find("INVALID_INSTRUCTION") != std::string::npos || info.name.empty()
```
chfast: Why not `isDefined = ...; return isDefined`? Double negation is confusing.

test/tools/libtesteth/Options.cpp
```diff
@@ -209,7 +209,10 @@ Options::Options(int argc, char** argv)
 		else if (arg == "--seed")
 		{
 			throwIfNoArgumentFollows();
-			randomTestSeed = static_cast<uint64_t>(min<u256>(UINT64_MAX, toInt(argv[++i])));
+			u256 input = toInt(argv[++i]);
+			if (input > UINT64_MAX)
```
chfast: I would stick to `std::numeric_limits<uint64_t>` although longer.

PR #4632, opened 2017-10-27
Set weights in a correct way

pirapira:
When I saw
```
for (int i = (int)(eth::Instruction::PUSH1); i < 32; i++)
```
I noticed the loop never loops.

### DIFF
--- a/test/tools/fuzzTesting/fuzzHelper.cpp
+++ b/test/tools/fuzzTesting/fuzzHelper.cpp
@@ -426,16 +426,16 @@ RandomCodeOptions::RandomCodeOptions() :
 	sendingAddressProbability(3)	// probability of calling to the tx sending account
 {
 	//each op code with same weight-probability
-	for (auto i = 0; i < 255; i++)
-		mapWeights.insert(std::pair<int, int>(i, 40));
+	for (uint8_t i = 0; i < 255; i++)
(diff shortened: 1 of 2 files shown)


test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -426,16 +426,16 @@ RandomCodeOptions::RandomCodeOptions() :
...
 	//Probability of instructions
 	setWeight(eth::Instruction::STOP, 1);
-	for (int i = (int)(eth::Instruction::PUSH1); i < 32; i++)
+	for (uint8_t i = (int)(eth::Instruction::PUSH1); i <= (uint8_t)(eth::Instruction::PUSH32); i++)
 		setWeight((eth::Instruction) i, 1);
-	for (int i = (int)(eth::Instruction::SWAP1); i < 16; i++)
+	for (uint8_t i = (int)(eth::Instruction::SWAP1); i <= (uint8_t)(eth::Instruction::SWAP16); i++)
 		setWeight((eth::Instruction) i, 10);
-	for (int i = (int)(eth::Instruction::DUP1); i < 16; i++)
+	for (uint8_t i = (int)(eth::Instruction::DUP1); i <= (uint8_t)(eth::Instruction::DUP16); i++)
```
winsvega: probably no need to use (int) then

PR #4633, opened 2017-10-27
Avoid static atomic flag

pirapira:
Trying to prevent the CircleCI MacOS issue

### DIFF
--- a/test/tools/libtesteth/boostTest.cpp
+++ b/test/tools/libtesteth/boostTest.cpp
@@ -27,6 +27,7 @@
 #include <clocale>
 #include <cstdlib>
 #include <iostream>
+#include <thread>
 #include <test/tools/libtesteth/TestHelper.h>
 
 using namespace boost::unit_test;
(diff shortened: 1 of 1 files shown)


test/tools/libtesteth/boostTest.cpp
```diff
@@ -47,11 +48,10 @@ void createRandomTestWrapper()
 	exit(0);
 }
 
-static std::atomic_bool stopTravisOut;
-void travisOut()
+void travisOut(std::atomic_bool *_stopTravisOut)
```
pirapira: When I make this a reference, the compiler errors type checking the constructor of `std::thread` below.

winsvega: are you talking about createRandomTest test issue ? 
if so I dont think that this is the cause. because it was failing only on this test and other tests were just fine.

pirapira: I have no idea, but will keep removing static variables until the issue disappears.

pirapira: @winsvega however, the issue disappeared when I added some debug outputs in the `maing()` function in `boostTest.cpp`.   Also, the issue disappeared when I removed the `travisOut` thread.  https://circleci.com/gh/ethereum/cpp-ethereum/1078

pirapira:
I'm less and less sure because all builds got fixed around the same time.

PR #4636, opened 2017-10-29
Fix delete-non-virtual-dtor warnings for more recent compilers

jacquelinekay:
This pull request adds virtual destructors with default implementations for several base classes to fix warnings due to deleting objects with non-virtual destructors.

I was unable to build cpp-ethereum without warnings using Clang 6 (nightly) or GCC 8 because the [delete-non-virtual-destructor](https://gcc.gnu.org/wiki/VerboseDiagnostics#delete-non-virtual-dtor) warning is emitted more aggressively than GCC 4.8, the compiler used on Travis. To fix it, I added virtual destructors to Node, UDPSocketEvents, and AccountHolder. I made BlockChainSync final instead of adding a virtual destructor to it, since it doesn't have any virtual functions or derived classes within the project, and I don't know enough about this project to know if it is intended to be derived from in the future.

As a side note, if the maintainers are interested in testing with higher releases of Clang than the default in Ubuntu 14.04, I recommend the [LLVM nightly packages](http://apt.llvm.org/) for deployment.

### DIFF
--- a/libethereum/BlockChainSync.h
+++ b/libethereum/BlockChainSync.h
@@ -46,7 +46,7 @@ class EthereumPeer;
  * @brief Base BlockChain synchronization strategy class.
  * Syncs to peers and keeps up to date. Base class handles blocks downloading but does not contain any details on state transfer logic.
  */
-class BlockChainSync: public HasInvariants
+class BlockChainSync final: public HasInvariants
 {
 public:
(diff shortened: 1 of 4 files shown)


chfast:
We are testing with GCC6 and clang5. Nightly compilers can be added when we have nightly builds in place.

PR #4639, opened 2017-10-31
Better command line in rlp

demon1999:


rlp/main.cpp
```diff
@@ -215,61 +185,117 @@ int main(int argc, char** argv)
...
-			encoding = Encoding::Binary;
-		else if (arg == "-e" || arg == "--encrypt")
-			encrypt = true;
 		else if (inputFile.empty())
 			inputFile = arg;
 		else
 			otherInputs.push_back(arg);
 	}
+	if (vm.count("help")) {
+		cout << "Usage rlp <mode> [OPTIONS]" << endl << modesOptions
```
chfast: I'm not expert here, but shouldn't the program_options be able to output that for us?

demon1999: fixed

rlp/main.cpp
```diff
@@ -216,60 +186,119 @@ int main(int argc, char** argv)
...
+		prefs.forceString = prefs.forceHex = false;
+	if (vm.count("force-string"))
+		prefs.forceString = true;
+	if (vm.count("force-hex"))
+		prefs.forceHex = true, prefs.forceString = false;
+	if (vm.count("force-escape"))
+		prefs.escapeAll = true;
+	if (vm.count("nice"))
+		prefs.forceString = true, prefs.stringInts = false, prefs.forceHex = false, prefs.indent = "  ";
+	if (vm.count("lenience"))
```
gumb0: Here's the duplicated piece of code from `lenience` till `encrypt`

gumb0:
What if we change it to `rlp --mode=<mode>`, would it be easier to make help output look better and could we then get rid of handling unrecognized options?

The challenge might be in different modes having different number of arguments following them (like `assemble` has arbitrary number of files)

Check out also "positional options". maybe it helps somehow http://www.boost.org/doc/libs/1_58_0/doc/html/program_options/tutorial.html#idp337627504

gumb0:
What happened to the indentation? Looks like it all changed to spaces

demon1999:
It's my fault. I'll fix it

rlp/main.cpp
```diff
@@ -189,18 +155,18 @@ void putOut(bytes _out, Encoding _encoding, bool _encrypt, bool _quiet)
...
-	case Encoding::Base64:
-		cout << toBase64(&_out) << endl;
-		break;
-	case Encoding::Binary:
-		cout.write((char const*)_out.data(), _out.size());
-		break;
-	case Encoding::Keccak:
-		cout << sha3(_out).hex() << endl;
-		break;
+		case Encoding::Hex: case Encoding::Auto:
```
gumb0: Don't change the indentation of `cases`s inside switches, I just fixed it before in my commit here

rlp/main.cpp
```diff
@@ -1,16 +1,13 @@
 /*
 	This file is part of cpp-ethereum.
-
```
gumb0: Don't remove these empty lines in comments

CMakeLists.txt
```diff
@@ -98,8 +98,8 @@ include(EthExecutableHelper)
 include(EthDependencies)
 include(EthUtils)
 
-hunter_add_package(Boost COMPONENTS filesystem system thread)
-find_package(Boost CONFIG REQUIRED filesystem system thread)
+hunter_add_package(Boost COMPONENTS program_options filesystem system thread)
```
chfast: Hasn't this been merged to develop already?

rlp/main.cpp
```diff
@@ -216,60 +186,108 @@ int main(int argc, char** argv)
...
+		cout << "Usage rlp <mode> [OPTIONS]\nModes:\n"
+			 << "    create   <json>  Given a simplified JSON string, output the RLP." << endl
+			 << "    render   [ <file> | -- ]  Render the given RLP." << endl
+			 << "    list     [ <file> | -- ]  List the items in the RLP list by hash and size." << endl
+			 << "    extract  [ <file> | -- ]  Extract all items in the RLP list, named by hash." << endl
+			 << "    assemble [ <manifest> | <base path> ] <file> ...  Given a manifest & files, output the RLP." << endl
+			 << renderOptions << generalOptions;
+		exit(0);
+	}
+	if (vm.count("lenience"))
```
chfast: This seems to be a weird way of doing this. Can't we inform boost we would like to store this option in `bool lenience` variable and use the variable directly?

chfast:
Rebase and merge?

gumb0:
I'll rebase it

rlp/main.cpp
gumb0: This is the way suggested by the official tutorials http://www.boost.org/doc/libs/1_65_1/doc/html/program_options/tutorial.html#idp437555840
But there's an option to store it local variable, too, I think.

This way is actually more compact (no need to define a local variable for each option), but duplicating the option name is a downside

PR #4640, opened 2017-10-31
Update the snapcraft metadata

elopio:
With this, the snap build will succeed in build.snapcraft.io
- Add the git source-type, to get the git submodules.
- For now remove the binaries that are not installed in DESTDIR.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -9,16 +9,13 @@ confinement: devmode
 apps:
   eth:
     command: eth
-  ethkey:
-    command: ethkey
   ethvm:
     command: ethvm
-  rlp:
-    command: rlp
 
 parts:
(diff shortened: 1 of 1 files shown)


chfast:
Hey! I think it works: https://build.snapcraft.io/user/ethereum/cpp-ethereum/101013.
How can we disabled i386 and ARM?

elopio:
Wooohooo! 
@chfast we are adding some more semantics to the snapcraft.yaml file to declare the architectures, but that's not yet ready: https://github.com/canonical-websites/build.snapcraft.io/issues/556
If you have any comments or requests there, now it's great time to send them :)

Does cpp-ethereum not support those architectures? Or is it that we are missing something for the build to succeed?

Issue #4641, opened 2017-11-01
"std::string" parameter could be improved with "const std::string&"


orpillar:
To avoid copying strings around, the following "std::string" parameter could be improved with "const std::string&":

- eth/MinerAux.h: void doBenchmark(std::string _m, unsigned _warmupDuration = 15, unsigned _trialDuration = 3, unsigned _trials = 5)
- libdevcore/CommonData.h:std::string toHex(Iterator _it, Iterator _end, std::string _prefix)
- libdevcore/CommonIO.h:template <class T> inline std::string toString(std::chrono::time_point<T> const& _e, std::string _format = "%Y-%m-%d %H:%M:%S")
- libdevcore/CommonIO.h:template <class T> inline std::string toString(std::chrono::time_point<T> const& _e, std::string _format = "%F %T")
- libdevcore/Exceptions.h:struct InterfaceNotSupported: virtual Exception { public: InterfaceNotSupported(std::string _f): Exception("Interface " + _f + " not supported.") {} };
- libdevcore/Exceptions.h:struct ExternalFunctionFailure: virtual Exception { public: ExternalFunctionFailure(std::string _f): Exception("Function " + _f + "() failed.") {} };
- libdevcore/FileSystem.cpp:fs::path dev::getDataDir(string _prefix)
- libdevcore/FileSystem.cpp:fs::path dev::getDefaultDataDir(string _prefix)
- libdevcore/FileSystem.h:boost::filesystem::path getDataDir(std::string _prefix = "ethereum");
- libdevcore/FileSystem.h:boost::filesystem::path getDefaultDataDir(std::string _prefix = "ethereum");
- libweb3jsonrpc/Web3.h:  Web3(std::string _clientVersion = "C++ (ethereum-cpp)"): m_clientVersion(_clientVersion) {}

chfast:
I mostly agree, but that requires a Pull Request.

pushkarmoi:
If no one has already done this,, should I implement these changes and create a pull request? 
I'm new here :D

chfast:
Fixed in #4644.

orpillar:
Without handy working environment, browsing code may be better fit for me for now.
Thanks for looking into the issue.

PR #4644, opened 2017-11-07
Replace "std::string" with "const std::string&" Issue #4641

nedodn:
Replaced instances of std::string with const std::string& to fix Issue #4641

### DIFF
--- a/libdevcore/CommonData.h
+++ b/libdevcore/CommonData.h
@@ -43,7 +43,7 @@ enum class WhenError
 };
 
 template <class Iterator>
-std::string toHex(Iterator _it, Iterator _end, std::string _prefix)
+std::string toHex(Iterator _it, Iterator _end, const std::string& _prefix)
 {
 	typedef std::iterator_traits<Iterator> traits;
 	static_assert(sizeof(typename traits::value_type) == 1, "toHex needs byte-sized element type");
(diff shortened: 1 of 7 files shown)


libdevcore/CommonData.h
```diff
@@ -43,7 +43,7 @@ enum class WhenError
 };
 
 template <class Iterator>
-std::string toHex(Iterator _it, Iterator _end, std::string _prefix)
+std::string toHex(Iterator _it, Iterator _end, const std::string& _prefix)
```
chfast: We write `std::string const&` not `const std::string&`. Can you change it this way?

libweb3jsonrpc/Web3.h
```diff
@@ -9,7 +9,7 @@ namespace rpc
 class Web3: public Web3Face
 {
 public:
-	Web3(std::string _clientVersion = "C++ (ethereum-cpp)"): m_clientVersion(_clientVersion) {}
+	Web3(const std::string& _clientVersion = "C++ (ethereum-cpp)"): m_clientVersion(_clientVersion) {}
```
chfast: Here pass by value should stay, but `std::move` may be used on the initialization list. Can you also change the default value to `cpp-ethereum`?

PR #4652, opened 2017-11-19
fix test name check when using --singletest

winsvega:
do not check all the filenames of filled tests when using --singletest
allows to run a single test even if there is a test with a broken name.


pirapira:
Before this change, what happened if there was a test with a broken name?

winsvega:
testeth was returning an error when running a test suite with such test. even if you used it with --singletest ''
which lead to a check of all file names in the test suite,

PR #4654, opened 2017-11-21
revive --singletest <file testname> option

winsvega:
usage 

```
./testeth -t <TestSuiteName> -- --singletest <PathToTheTestFile.json> <TestNameInsideFile>
```

executes a specific test in a provided .json file for a specified -t TestSuite. 

father update: 
https://github.com/ethereum/cpp-ethereum/issues/4667

### DIFF
--- a/test/tools/libtesteth/Options.h
+++ b/test/tools/libtesteth/Options.h
@@ -63,7 +63,7 @@ public:
 	/// Test selection
 	/// @{
 	bool singleTest = false;
-	std::string singleTestFile;
+	boost::optional<std::string> singleTestFile;
 	std::string singleTestName;
 	std::string singleTestNet;
 	int trDataIndex;	///< GeneralState data
--- a/test/tools/libtesteth/TestHelper.cpp
+++ b/test/tools/libtesteth/TestHelper.cpp
@@ -95,6 +95,15 @@ void mine(BlockHeader& _bi, SealEngineFace* _sealer, bool _verify)
 namespace test
 {
 
+void tryRunSingleTestFile(dev::test::TestSuite const& _suite)
+{
+	if (test::Options::get().singleTestFile.is_initialized())
+	{
+		boost::filesystem::path file(test::Options::get().singleTestFile.get());
+		_suite.runTestWithoutFiller(file);
+		exit(0);
+	}
+}
 
 string netIdToString(eth::Network _netId)
 {
--- a/test/tools/libtesteth/TestSuite.cpp
+++ b/test/tools/libtesteth/TestSuite.cpp
@@ -111,6 +111,16 @@ string const c_copierPostf = "Copier";
 
 void TestSuite::runAllTestsInFolder(string const& _testFolder) const
 {
+	// Allow to execute a custom test .json file on any test suite
+	if (!test::Options::get().singleTestFile.empty())
+	{
+		auto& testOutput = test::TestOutputHelper::get();
+		testOutput.initTest(1);
+		executeFile(fs::path(test::Options::get().singleTestFile));
+		testOutput.finishTest();
+		exit(0);
+	}
+
 	// check that destination folder test files has according Filler file in src folder
 	string const filter = test::Options::get().singleTestName.empty() ? string() : test::Options::get().singleTestName;
 	vector<fs::path> const compiledFiles = test::getJsonFiles(getFullPath(_testFolder), filter);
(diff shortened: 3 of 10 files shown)


pirapira:
AppVeyor seems to have timed out.

pirapira:
Please update the `--help` output and the documentation.

winsvega:
@pirapira  help documentation already has a note about this option. it was not working due to refactoring for some time.

pirapira:
@winsvega I'll have a look.

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -111,6 +111,16 @@ string const c_copierPostf = "Copier";
...
+	// Allow to execute a custom test .json file on any test suite
+	if (!test::Options::get().singleTestFile.empty())
+	{
+		auto& testOutput = test::TestOutputHelper::get();
+		testOutput.initTest(1);
+		executeFile(fs::path(test::Options::get().singleTestFile));
+		testOutput.finishTest();
+		exit(0);
+	}
+
```
pirapira: This code should not be in `runAllTestsInFolder()`.  The caller of `runAllTestsInFolder()` already knows that the execution is not about any folder, so it can avoid calling `runAllTestsInFolder()`.

winsvega: I know. I just want it to be in one place. If you move this check outside then it will have to be added in all .cpp for all test suites.

pirapira: The move it further up, closer to `main()`.

test/tools/libtesteth/Options.h
```diff
@@ -53,7 +53,7 @@ class Options
 	std::string rCurrentTestSuite; ///< Remember test suite before boost overwrite (for random tests)
 	bool statediff = false;///< Fill full post state in General tests
 	bool fulloutput = false;///< Replace large output to just it's length
-	bool createRandomTest = false; ///< Generate random test
+	boost::optional<bool> createRandomTest; ///< Generate random test
```
pirapira: What is better with this optional?

test/tools/libtesteth/boostTest.cpp
```diff
@@ -36,15 +36,17 @@ static std::ostringstream strCout;
...
 	std::cerr.rdbuf(oldCerrStreamBuf);
 
-	if (!dev::test::createRandomTest())
-		throw framework::internal_error("Create random test error! See std::error for more details.");
-
+	if (dev::test::Options::get().createRandomTest.is_initialized())
+	{
+		if (!dev::test::createRandomTest())
+			throw framework::internal_error("Create random test error! See std::error for more details.");
+	}
```
pirapira: Why is it ok when `createRandomTest` is not initialized?

test/tools/libtesteth/boostTest.cpp
```diff
@@ -36,15 +36,17 @@ static std::ostringstream strCout;
 std::streambuf* oldCoutStreamBuf;
 std::streambuf* oldCerrStreamBuf;
 
-void createRandomTestWrapper()
+void customTestSuiteWrapper()
```
pirapira: Is this still being used only for creating random test?  In that case, I think the old name is better.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -95,6 +95,15 @@ void mine(BlockHeader& _bi, SealEngineFace* _sealer, bool _verify)
 namespace test
 {
 
+void tryRunSingleTestFile(dev::test::TestSuite const& _suite)
+{
+	if (test::Options::get().singleTestFile.is_initialized())
+	{
+		boost::filesystem::path file(test::Options::get().singleTestFile.get());
+		_suite.runTestWithoutFiller(file);
+		exit(0);
```
pirapira: Why do you `exit` suddenly?  The function should usually just return when it finishes the task.

Maybe this function can return a boolean.

test/tools/libtesteth/Options.h
winsvega: dunno. Pawel or you suggested to use boost::optional

test/tools/libtesteth/boostTest.cpp
winsvega: I see. I was hoping to put --singletest into this function. but then I realized that it is not clear how to link -t argument with the test suite

test/tools/libtesteth/TestHelper.cpp
winsvega: because it should stop execution of other test cases in the suite

pirapira: Can you do something as if `--singletest 1234543212345` so that no test cases match?

winsvega: yes. it will scan the tests and if there are no test with such name - nothing will be executed.

gumb0: Can you instead of this have a function that calls either `_suite.runTestWithoutFiller(file);`  or `_suite.runAllTestsInFolder(casename);` depending on the result of `test::Options::get().singleTestFile.is_initialized()` ?
Will it help get rid of `exit` ?

winsvega: no. because we have many test cases called. but if singleTest is defined we have to stop the execution of all of them

winsvega: no. because we have many test cases called. but if singleTest is defined we have to stop the execution of all of them

pirapira:
@winsvega please paste here a link to the issue about removing `exit()`.

winsvega:
https://github.com/ethereum/cpp-ethereum/issues/4667

PR #4655, opened 2017-11-21
Added clang-format from (evmjit)

TomzOk:
clang-format is a popular auto-formatting tool for c family. Auto-formatting makes it much easier for contributors to adhere to the projects style guidelines.

This PR simply copied the clang-format used in the evm-jit submodule into the parent module to better unify the style rules across Ethereum c-family projects.  This format file does not adhere to current style rules so it probably would not be merged without consensus on a style format!  However I think having the same style guide across all Ethereum projects is a good idea.

Thanks!

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,89 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Chromium
+AccessModifierOffset: -1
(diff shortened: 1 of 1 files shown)


chfast:
The .clang-format merged in #4706. Thank you.

PR #4657, opened 2017-11-22
rpc: Drop HTTP server

chfast:
Depends on #4673.


chriseth:
While I welcome the general direction of removing the HTTP server, it is important for debugging solidity tests that remix has debugging access to eth. Is there a simple ipc-http-bridge we can use?

chfast:
No there is no `ipc-http-bridge`. 
Also @gumb0 reported that HTTP is needed for ethstats at the moment, so I will have to address this too before removing HTTP server.

How solidity is using HTTP transport? Can't it use Unix sockets?

chriseth:
Solidity does use IPC, I only use HTTP to inspect test failures from remix.

PR #4663, opened 2017-11-27
Use more recent version of Hunter

dmakarov:
This update enables using the patched version of cryptopp library,
which is identical to the currently used version 5.6.5, except it can
be built with cmake Xcode generator on MacOS, whereas unpatched
version doesn't work for Xcode generator.


chfast:
Let me rerun Windows builds.

PR #4664, opened 2017-11-28
general transaction tests

winsvega:
A general transaction tests scheme https://github.com/ethereum/tests/issues/373

The cpp pr is ready.
Tests are ready https://github.com/ethereum/tests/pull/374


test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,164 @@ namespace fs = boost::filesystem;
 
 namespace dev {  namespace test {
 
+mObject getExpectSection(mValue const& _expect, eth::Network _network)
+{
+	mObject obj;
```
pirapira: `BOOST_REQUIRE(_expect.type() == json_spirit::array_type)` maybe?

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,164 @@ namespace fs = boost::filesystem;
...
+		BOOST_REQUIRE(value.type() == json_spirit::obj_type);
+		obj = value.get_obj();
+		BOOST_REQUIRE_MESSAGE(obj.count("network"), "network section not set in expect section!");
+		vector<string> networks;
+		ImportTest::parseJsonStrValueIntoVector(obj["network"], networks);
+		BOOST_CHECK_MESSAGE(networks.size() > 0, TestOutputHelper::get().testName() + " Network array not set!");
+		ImportTest::checkAllowedNetwork(networks);
+
+		if (test::inArray(networks, test::netIdToString(_network)) || test::inArray(networks, string("ALL")))
+			return obj;
```
pirapira: Maybe just push back `obj` into a vector, and after the loop, check that the vector contains only one element.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,164 @@ namespace fs = boost::filesystem;
...
+	RLPStream rlpStream = createRLPStreamFromTransactionFields(tObj);
+	out["rlp"] = toHexPrefixed(rlpStream.out());
+
+	// Theoretical block for transaction check
+	BlockHeader bh;
+	bh.setNumber(1);	//Seal engine below enables network rules from block 0
+	bh.setGasLimit(u256("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"));
+
+	mValue expectObj = _o.at("expect");
+	for (auto const network : test::getNetworks())
```
pirapira: No space before `:`.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,164 @@ namespace fs = boost::filesystem;
 
 namespace dev {  namespace test {
 
+mObject getExpectSection(mValue const& _expect, eth::Network _network)
+{
+	mObject obj;
+	for (auto const& value : _expect.get_array())
```
pirapira: There should be no space before `:`.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,164 @@ namespace fs = boost::filesystem;
...
+	// Theoretical block for transaction check
+	BlockHeader bh;
+	bh.setNumber(1);	//Seal engine below enables network rules from block 0
+	bh.setGasLimit(u256("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"));
+
+	mValue expectObj = _o.at("expect");
+	for (auto const network : test::getNetworks())
+	{
+		unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(network)).createSealEngine());
+		bool onConstantinople = (network == eth::Network::ConstantinopleTest);
```
pirapira: Maybe `>=` instead of `==`.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,164 @@ namespace fs = boost::filesystem;
...
+	bytes stream = importByteArray(_o.at("rlp").get_str());
+	RLP rlp(stream);
+	string const& testname = TestOutputHelper::get().testName();
+
+	// Theoretical block for transaction check
+	BlockHeader bh;
+	bh.setNumber(1);	//Seal engine below enables network rules from block 0
+	bh.setGasLimit(u256("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"));
+
+	for (auto const network : test::getNetworks())
```
pirapira: No space before `:`.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,164 @@ namespace fs = boost::filesystem;
...
+	{
+		string networkname = test::netIdToString(network);
+		BOOST_REQUIRE_MESSAGE(_o.count(networkname) > 0, testname + " Transaction test missing network results! (" + networkname + ")");
+		BOOST_REQUIRE(_o.at(networkname).type() == json_spirit::obj_type);
+		mObject obj = _o.at(networkname).get_obj();
+		try
+		{
+			txFromRlp = Transaction(rlp.data(), CheckTransaction::Everything);
+			bool onConstantinople = (network == eth::Network::ConstantinopleTest);
```
pirapira: Maybe `>=` instead of `==`.

test/tools/libtesteth/ImportTest.h
```diff
@@ -89,5 +89,14 @@ class ImportTest
 	json_spirit::mObject& m_testOutputObject;
 };
 
+template<class T>
+bool inArray(std::vector<T> const& _array, const T _val)
+{
+	for (auto const& obj : _array)
```
pirapira: No space before `:`.

test/tools/jsontests/TransactionTests.cpp
winsvega: hm. you right. but we dont have a fork larger thatn Constantinople yet. 
and we will have to make sure that enum of any other fork is larger

test/tools/jsontests/TransactionTests.cpp
winsvega: is it bad to have return in a loop ?

test/tools/jsontests/TransactionTests.cpp
winsvega: fixed

test/tools/jsontests/TransactionTests.cpp
pirapira: Just in case many fields match.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,167 @@ namespace fs = boost::filesystem;
...
+		}
+		catch(...)
+		{
+			BOOST_CHECK_MESSAGE(obj.count("hash") == 0,
+			 testname + "A transaction object should not be defined because the RLP is invalid! (" + networkname + ")");
+			continue;
+		}
+
+		BOOST_REQUIRE(obj.count("sender") > 0);
+		Address addressReaded = Address(obj["sender"].get_str());
```
pirapira: `Readed` is not an English word.  `addressExpected`?

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,167 @@ namespace fs = boost::filesystem;
...
+			cnote << testname;
+			cnote << "Transaction Exception: " << diagnostic_information(_e);
+			BOOST_CHECK_MESSAGE(obj.count("hash") == 0,
+			 testname + "A transaction object should not be defined because the RLP is invalid! (" + networkname + ")");
+			continue;
+		}
+		catch(...)
+		{
+			BOOST_CHECK_MESSAGE(obj.count("hash") == 0,
+			 testname + "A transaction object should not be defined because the RLP is invalid! (" + networkname + ")");
```
pirapira: The indentation is a bit off here.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,167 @@ namespace fs = boost::filesystem;
...
+			se->verifyTransaction(ImportRequirements::Everything, txFromRlp, bh, 0);
+			if (!(txFromRlp.signature().isValid() || onConstantinopleAndZeroSig))
+				BOOST_THROW_EXCEPTION(Exception() << errinfo_comment(testname + "transaction from RLP signature is invalid (" + networkname + ")") );
+		}
+		catch(Exception const& _e)
+		{
+			cnote << testname;
+			cnote << "Transaction Exception: " << diagnostic_information(_e);
+			BOOST_CHECK_MESSAGE(obj.count("hash") == 0,
+			 testname + "A transaction object should not be defined because the RLP is invalid! (" + networkname + ")");
```
pirapira: The indentation here is also a bit off.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,167 @@ namespace fs = boost::filesystem;
...
+		string networkname = test::netIdToString(network);
+		BOOST_REQUIRE_MESSAGE(_o.count(networkname) > 0, testname + " Transaction test missing network results! (" + networkname + ")");
+		BOOST_REQUIRE(_o.at(networkname).type() == json_spirit::obj_type);
+		mObject obj = _o.at(networkname).get_obj();
+		try
+		{
+			bytes stream = importByteArray(_o.at("rlp").get_str());
+			RLP rlp(stream);
+
+			txFromRlp = Transaction(rlp.data(), CheckTransaction::Everything);
```
pirapira: It's better to make this `Transaction txFromRlp = ...` and remove the definition above.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,167 @@ namespace fs = boost::filesystem;
...
+			bool onConstantinople = (network == eth::Network::ConstantinopleTest);
+			bool onConstantinopleAndZeroSig = onConstantinople && txFromRlp.hasZeroSignature();
+			unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(network)).createSealEngine());
+			se->verifyTransaction(ImportRequirements::Everything, txFromRlp, bh, 0);
+			if (!(txFromRlp.signature().isValid() || onConstantinopleAndZeroSig))
+				BOOST_THROW_EXCEPTION(Exception() << errinfo_comment(testname + "transaction from RLP signature is invalid (" + networkname + ")") );
+		}
+		catch(Exception const& _e)
```
pirapira: A space is missing after `catch`.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,167 @@ namespace fs = boost::filesystem;
...
+		{
+			cnote << testname;
+			cnote << "Transaction Exception: " << diagnostic_information(_e);
+			BOOST_CHECK_MESSAGE(obj.count("hash") == 0,
+			 testname + "A transaction object should not be defined because the RLP is invalid! (" + networkname + ")");
+			continue;
+		}
+		catch(...)
```
pirapira: A space is missing after `catch` here as well.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,167 @@ namespace fs = boost::filesystem;
...
+				BOOST_CHECK_MESSAGE(toString(txFromFields.sender()) == expectSender, "Error filling transaction test " + TestOutputHelper::get().testName() + ": expected another sender address! (got: " + toString(txFromFields.sender()) + "), expected: (" + expectSender + ")");
+			}
+
+			mObject resultObject;
+			resultObject["sender"] = toString(txFromFields.sender());
+			//resultObject["transaction"] = ImportTest::makeAllFieldsHex(tObj);
+			resultObject["hash"] = toString(txFromFields.sha3());
+			out[test::netIdToString(network)] = resultObject;
+		}
+		catch(Exception const& _e)
```
pirapira: A space is missing after `catch`.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,167 @@ namespace fs = boost::filesystem;
...
+	// Theoretical block for transaction check
+	BlockHeader bh;
+	bh.setNumber(1);	//Seal engine below enables network rules from block 0
+	bh.setGasLimit(u256("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"));
+
+	mValue expectObj = _o.at("expect");
+	for (auto const network: test::getNetworks())
+	{
+		unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(network)).createSealEngine());
+		bool onConstantinople = (network == eth::Network::ConstantinopleTest);
```
pirapira: This `==` should be replaced with `>=` as well.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,167 @@ namespace fs = boost::filesystem;
...
+		BOOST_REQUIRE_MESSAGE(obj.count("network"), "network section not set in expect section!");
+		vector<string> networks;
+		ImportTest::parseJsonStrValueIntoVector(obj["network"], networks);
+		BOOST_CHECK_MESSAGE(networks.size() > 0, TestOutputHelper::get().testName() + " Network array not set!");
+		ImportTest::checkAllowedNetwork(networks);
+
+		if (test::inArray(networks, test::netIdToString(_network)) || test::inArray(networks, string("ALL")))
+			objVector.push_back(obj);
+	}
+	BOOST_REQUIRE_MESSAGE(objVector.size() == 1,"Expect network should occur once in expect section of transaction test filler! (" + test::netIdToString(_network) + ")");
```
pirapira: A space is missing after `,`.

test/tools/jsontests/TransactionTests.cpp
winsvega: this field is used after try, catch block.

pirapira: Then move the definition into the body of this `for` loop.

test/tools/jsontests/TransactionTests.cpp
winsvega: this wont work. there are other networks (configurations) larger than this. it does not mean what you expect it to mean. I think it is better to select networks more explicitly here

pirapira: Then, use a function like `hasZeroSignature(network)`.

winsvega: this function is already used.

test/tools/jsontests/TransactionTests.cpp
winsvega: done

test/tools/jsontests/TransactionTests.cpp
pirapira: Why not here?

winsvega:
how about this solution? see the last commit.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -79,8 +79,9 @@ json_spirit::mObject FillTransactionTest(json_spirit::mObject const& _o)
 	mValue expectObj = _o.at("expect");
 	for (auto const network: test::getNetworks())
 	{
-		unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(network)).createSealEngine());
-		bool onConstantinople = (network == eth::Network::ConstantinopleTest);
+		ChainParams params(genesisInfo(network));
+		unique_ptr<SealEngineFace> se(params.createSealEngine());
+		bool onConstantinople = (params.constantinopleForkBlock == 0); //Test networks has forkblocks set to 0 if rules are enabled
```
pirapira: Why is this zero for true?

winsvega: Test networks has forkblocks set to 0 if rules are enabled. because it indicates the blocknumber from which it start to work

winsvega: transaction tests does not have blocknumber anymore

pirapira: Ah, OK.  The current block number is 1, so let's compare it against 1.

If ConstantinopleForkBlock = 1, we should be already on Constantinople.

winsvega: it is set to 0. do you mean 
` if 1 >= ConstantinopleForkBlock  ` ?

pirapira: Yeah, the theoretical block number is 1, not zero.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,170 @@ namespace fs = boost::filesystem;
...
+	BOOST_REQUIRE_MESSAGE(_o.count("transaction") > 0, "transaction section not found! " + TestOutputHelper::get().testFileName());
+	BOOST_REQUIRE_MESSAGE(_o.count("expect") > 0, "expect section not found! " + TestOutputHelper::get().testFileName());
+
+	mObject tObj = _o.at("transaction").get_obj();
+	RLPStream rlpStream = createRLPStreamFromTransactionFields(tObj);
+	out["rlp"] = toHexPrefixed(rlpStream.out());
+
+	// Theoretical block for transaction check
+	BlockHeader bh;
+	bh.setNumber(1);	//Seal engine below enables network rules from block 0
```
pirapira: The block number is set to 1 here.

winsvega: done

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,174 @@ namespace fs = boost::filesystem;
...
+				string expectSender = toString(expectSection["sender"].get_str());
+				BOOST_CHECK_MESSAGE(toString(txFromFields.sender()) == expectSender, "Error filling transaction test " + TestOutputHelper::get().testName() + ": expected another sender address! (got: " + toString(txFromFields.sender()) + "), expected: (" + expectSender + ")");
+			}
+
+			mObject resultObject;
+			resultObject["sender"] = toString(txFromFields.sender());
+			//resultObject["transaction"] = ImportTest::makeAllFieldsHex(tObj);
+			resultObject["hash"] = toString(txFromFields.sha3());
+			out[test::netIdToString(network)] = resultObject;
+		}
```
pirapira: Are we really expecting an exception from this big range of code?

Maybe only `se->verifyTransaction()` needs to be in the `try` block, I suspect.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,174 @@ namespace fs = boost::filesystem;
...
+			se->verifyTransaction(ImportRequirements::Everything, txFromFields, bh, 0);
+			if (expectSection.count("sender") > 0)
+			{
+				string expectSender = toString(expectSection["sender"].get_str());
+				BOOST_CHECK_MESSAGE(toString(txFromFields.sender()) == expectSender, "Error filling transaction test " + TestOutputHelper::get().testName() + ": expected another sender address! (got: " + toString(txFromFields.sender()) + "), expected: (" + expectSender + ")");
+			}
+
+			mObject resultObject;
+			resultObject["sender"] = toString(txFromFields.sender());
+			//resultObject["transaction"] = ImportTest::makeAllFieldsHex(tObj);
```
pirapira: Why is this commented out?

test/tools/jsontests/TransactionTests.cpp
winsvega: it is.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,174 @@ namespace fs = boost::filesystem;
...
+		}
+		catch (Exception const& _e)
+		{
+			cnote << testname;
+			cnote << "Transaction Exception: " << diagnostic_information(_e);
+			BOOST_CHECK_MESSAGE(obj.count("hash") == 0,
+				testname + "A transaction object should not be defined because the RLP is invalid! (" + networkname + ")");
+			continue;
+		}
+		catch (...)
```
pirapira: Are we really expecting this arbitrary exceptions happening from invalid RLP?

test/tools/jsontests/TransactionTests.cpp
winsvega: I decided that it is not neccessary to print and check rlp decoded from string for each fork. 
this check should be done by hash.

test/tools/jsontests/TransactionTests.cpp
winsvega: this one is to be safe.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,174 @@ namespace fs = boost::filesystem;
...
-
-				//Transaction is Valid
-				if (o.count("expect") > 0)
-				{
-					bool expectValid = (o["expect"].get_str() == "valid");
-					BOOST_CHECK_MESSAGE(expectValid, testname + " Check state: Transaction '" << i.first << "' is expected to be invalid!");
-					o.erase(o.find("expect"));
-				}
-			}
+				v[testname] = FillTransactionTest(o);
```
pirapira: Maybe also call `TestTransactionTest` on `v[testname]`?

test/tools/jsontests/TransactionTests.cpp
pirapira: OK.

test/tools/jsontests/TransactionTests.cpp
winsvega: it is done from the TestSuite

test/tools/jsontests/TransactionTests.cpp
pirapira: Still the `try` block can be much shorter here.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,174 @@ namespace fs = boost::filesystem;
...
+		}
+
+		//Transaction is Valid
+		bool expectValid = (expectSection["result"].get_str() == "valid");
+		BOOST_CHECK_MESSAGE(expectValid, TestOutputHelper::get().testName() + " Check state: Transaction is expected to be invalid! (" + test::netIdToString(network) + ")");
+	}
+	return out;
+}
+
+void TestTransactionTest(json_spirit::mObject const& _o)
```
pirapira: Splitting out this function is good.

test/tools/libtesteth/ImportTest.h
```diff
@@ -89,5 +89,14 @@ class ImportTest
 	json_spirit::mObject& m_testOutputObject;
 };
 
+template<class T>
+bool inArray(std::vector<T> const& _array, const T _val)
```
pirapira: This can take `T const& _val` as a const reference as well.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,174 @@ namespace fs = boost::filesystem;
...
+
 class TransactionTestSuite: public TestSuite
 {
 	json_spirit::mValue doTests(json_spirit::mValue const& _input, bool _fillin) const override
 	{
-		json_spirit::mValue v = _input; // TODO: avoid copying and only add valid fields into the new object.
-		unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(eth::Network::MainNetworkTest)).createSealEngine());
-		for (auto& i: v.get_obj())
+		BOOST_REQUIRE_MESSAGE(_input.type() == obj_type,
+		TestOutputHelper::get().get().testFileName() + " TransactionTest file should contain an object.");
```
pirapira: The second line should be indented deeper than the first line.

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -38,134 +39,174 @@ namespace fs = boost::filesystem;
...
+		BOOST_REQUIRE_MESSAGE(!_fillin || _input.get_obj().size() == 1,
+		TestOutputHelper::get().testFileName() + " TransactionTest filler should contain only one test.");
```
pirapira: Here too, the second line should be indented deeper than the first line.

test/tools/jsontests/TransactionTests.cpp
pirapira: Let's remove and see what happens.

test/tools/jsontests/TransactionTests.cpp
winsvega: should I remove the comment for good?

pirapira: Yes, I think so.

pirapira:
CircleCI fails with
```
-- Could NOT find mhd (missing:  MHD_LIBRARY MHD_INCLUDE_DIR) 
```

PR #4670, opened 2017-11-30
Clang format

chfast:


libethereum/State.cpp
```diff
@@ -44,40 +44,44 @@ const char* StateChat::name() { return EthViolet "⚙" EthWhite " ◌"; }
 
 namespace
 {
-
 /// @returns true when normally halted; false when exceptionally halted.
-bool executeTransaction(Executive& _e, Transaction const& _t, OnOpFunc const& _onOp)
+bool executeTransaction(Executive& _e,
+						Transaction const& _t,
```
gumb0: Hm, why did it do this, looks weird

gumb0: Probably `AlignAfterOpenBracket: Align` affects it, and `TabWidth:        4` while github uses  8

gumb0: I think I'd prefer `AlignAfterOpenBracket: DontAlign` and probably we should set `TabWidth` to 8 to make it look more normal on github

libethereum/State.cpp
```diff
@@ -186,11 +192,13 @@ Account* State::account(Address const& _addr)
...
 	RLP state(stateBack);
-	auto i = m_cache.emplace(
-		std::piecewise_construct,
-		std::forward_as_tuple(_addr),
-		std::forward_as_tuple(state[0].toInt<u256>(), state[1].toInt<u256>(), state[2].toHash<h256>(), state[3].toHash<h256>(), Account::Unchanged)
-	);
+	auto i = m_cache.emplace(std::piecewise_construct,
+							 std::forward_as_tuple(_addr),
+							 std::forward_as_tuple(state[0].toInt<u256>(),
+												   state[1].toInt<u256>(),
```
gumb0: This one is also weird because of `TabWidth: 4` I think

libethereum/State.cpp
chfast: Here `AlwaysBreak` matches our current style the most.

GitHub, code reviews, code copy & paste are reasons I'm not a fun of tabs. Tabs seem to work in your own editor only. Do you know anyone using tabs of size 2 in cpp-ethereum & solidity.

libethereum/State.cpp
chfast: Yes.

libethereum/State.h
```diff
@@ -326,29 +395,45 @@ class State
...
-	mutable std::unordered_map<Address, Account> m_cache;	///< Our address cache. This stores the states of each address that has (or at least might have) been changed.
-	mutable std::vector<Address> m_unchangedCacheEntries;	///< Tracks entries in m_cache that can potentially be purged if it grows too large.
-	mutable std::set<Address> m_nonExistingAccountsCache;	///< Tracks addresses that are known to not exist.
-	AddressHash m_touched;						///< Tracks all addresses touched so far.
+	OverlayDB m_db;  ///< Our overlay for the state tree.
+	SecureTrieDB<Address, OverlayDB>
+		m_state;  ///< Our state tree, as an OverlayDB DB.
+	mutable std::unordered_map<Address, Account> m_cache;  ///< Our address
+														   ///< cache. This
+														   ///< stores the
```
chfast: This is caused by "reflow comments" and max column 80. We can turn it off for now, but we can also fix the comments by moving them before the variables instead of having them after with `///<`.

libethereum/State.cpp
```diff
@@ -21,13 +21,13 @@
 
 #include "State.h"
 
-#include <boost/filesystem.hpp>
-#include <boost/timer.hpp>
 #include <libdevcore/Assertions.h>
 #include <libdevcore/TrieHash.h>
 #include <libevm/VMFactory.h>
-#include "BlockChain.h"
+#include <boost/filesystem.hpp>
```
chfast: Includes sorting can also be turned off for now, be I believe we can config what we want with set of regex and priorities. I just didn't have time to do it.

libethereum/State.h
```diff
@@ -123,33 +149,38 @@ struct Change
...
+			   Code);  // For this the special constructor needs to be used.
 	}
 
 	/// Helper constructor especially for storage change log.
-	Change(Address const& _addr, u256 const& _key, u256 const& _value):
-			kind(Storage), address(_addr), value(_value), key(_key)
-	{}
+	Change(Address const& _addr, u256 const& _key, u256 const& _value)
+	  : kind(Storage), address(_addr), value(_value), key(_key)
+	{
```
chfast: I don't know how to keep it like `{}`. Ideas?

libethereum/State.cpp
```diff
@@ -502,10 +516,7 @@ size_t State::codeSize(Address const& _a) const
 		return 0;
 }
 
-size_t State::savepoint() const
-{
-	return m_changeLog.size();
-}
+size_t State::savepoint() const { return m_changeLog.size(); }
```
chfast: In this file some short non-inline function were in single line, some were not. But seeing the changes, I think the opposite option should be better.

libethereum/State.h
isaachier: Working on this now. See [clang reference](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) regarding `SplitEmptyFunction`.

chfast: I fixed it.

chfast:
Do you have any preferences how to format constructor initialization list?

libethereum/State.h
chfast: Also fixed.

libethereum/State.cpp
```diff
@@ -228,10 +238,10 @@ unordered_map<Address, u256> State::addresses() const
 {
 #if ETH_FATDB
 	unordered_map<Address, u256> ret;
-	for (auto& i: m_cache)
+	for (auto& i : m_cache)
```
winsvega: so should here be a space or not? Youichi is always fixing this in my  PRs

libethereum/State.cpp
```diff
@@ -521,21 +531,11 @@ void State::rollback(size_t _savepoint)
...
-			m_cache.erase(change.address);
-			break;
-		case Change::Code:
-			account.setCode(std::move(change.oldCode));
-			break;
+		case Change::StorageRoot: account.setStorageRoot(change.value); break;
+		case Change::Balance: account.addBalance(0 - change.value); break;
+		case Change::Nonce: account.setNonce(change.value); break;
+		case Change::Create: m_cache.erase(change.address); break;
+		case Change::Code: account.setCode(std::move(change.oldCode)); break;
```
winsvega: this one is shorter but harder to read

libethereum/State.cpp
pirapira: @winsvega I'm following the current CodingStandards.txt.  After this PR is merged, I'll stop doing that.

pirapira:
I think I can get used to the new style somehow.  The benefit seems to overweigh tiny bits of strangeness here and there.

libethereum/State.cpp
chfast: There is not way to do the version without space in clang-format, so we will use space from now on.
Not having this option in clang-format usually means: nobody in the world of C++ cares and formats the code like that.

libethereum/State.cpp
chfast: Yes, I agree. I will check if I can bring it back. But this is a question if we allow single-lines after `case:` in general. Sometimes it is useful, like

```
case A: return "A";
case B: return "B";
```

libethereum/State.cpp
gumb0: Yeah I never saw anything like that before cpp-etherem

chfast:
@pirapira This is not final config, it's discussion. Please report "tiny bits of strangeness" so we can fix or discuss them.

libethereum/State.cpp
gumb0: I don't like these with return on the same line ^

libethereum/State.cpp
gumb0: Allowing one-line only for inline functions inside the class is probably not possible?

libethereum/State.cpp
chfast: So fixed. I think we are consistent there with config:
```
AllowShortBlocksOnASingleLine: false
AllowShortCaseLabelsOnASingleLine: false
AllowShortIfStatementsOnASingleLine: false
```

libethereum/State.cpp
chfast: It is possible. I changed the config to `Inline`. This affects the logging helper function at the beginning of this file.

chfast:
Now we have spaces!

To ignore whitespace in review, add `?w=1` in URL: https://github.com/ethereum/cpp-ethereum/pull/4670/files?w=1

libethereum/State.cpp
```diff
@@ -21,720 +21,792 @@
...
+    return EthViolet "⚙" EthWhite " ◌";
+}
 
 namespace
 {
-
 /// @returns true when normally halted; false when exceptionally halted.
-bool executeTransaction(Executive& _e, Transaction const& _t, OnOpFunc const& _onOp)
+bool executeTransaction(Executive& _e,
+                        Transaction const& _t,
```
gumb0: I don't like this when each parameter is on the separate line when all three could fit into one line on any screen.
Other opinions?

libethereum/State.h
```diff
@@ -96,60 +122,62 @@ class Executive;
...
+    u256 value;       ///< Change value, e.g. balance, storage and nonce.
+    u256 key;         ///< Storage key. Last because used only in one case.
+    bytes oldCode;    ///< Code overwritten by CREATE, empty except in case of
+                      ///< address collision.
+
+    /// Helper constructor to make change log update more readable.
+    Change(Kind _kind, Address const& _addr, u256 const& _value = 0)
+      : kind(_kind), address(_addr), value(_value)
+    {
+        assert(_kind !=
```
chfast: I cannot make it to break the comment instead of the expression...

gumb0: Hm, is there a way to enforce comments being on separate lines only?

isaachier: I'm guessing you can set it to the `ColumnLimit` (in this case 80 IIRC) so the formatter never appends it to the same line.

isaachier:
Just a suggestion as I have done this on projects in the past: do not format all the files at once after this pull request is approved. Rather, enforce a new rule so that new changes must be formatted. This way active developments can avoid painful merge conflicts.

chfast:
Thanks for suggestion @isaachier. This is exactly what we plan to do.

libethereum/State.cpp
chfast: Assuming we want to leave the column limit of 80, the options are:

```cpp
/// @returns true when normally halted; false when exceptionally halted.
bool executeTransaction(Executive& _e,
                        Transaction const& _t,
                        OnOpFunc const& _onOp)
{
    _e.initialize(_t);

    if (!_e.execute())
        _e.go(_onOp);
    return _e.finalize();
}
```

```cpp
/// @returns true when normally halted; false when exceptionally halted.
bool executeTransaction(Executive& _e, Transaction const& _t,
                        OnOpFunc const& _onOp)
{
    _e.initialize(_t);

    if (!_e.execute())
        _e.go(_onOp);
    return _e.finalize();
}
```

```cpp
/// @returns true when normally halted; false when exceptionally halted.
bool executeTransaction(Executive& _e,
    Transaction const& _t,
    OnOpFunc const& _onOp)
{
    _e.initialize(_t);

    if (!_e.execute())
        _e.go(_onOp);
    return _e.finalize();
}
```

```cpp
/// @returns true when normally halted; false when exceptionally halted.
bool executeTransaction(
    Executive& _e, Transaction const& _t, OnOpFunc const& _onOp)
{
    _e.initialize(_t);

    if (!_e.execute())
        _e.go(_onOp);
    return _e.finalize();
}
```

This seems to be a combination of 3 options, but somehow I cannot get all of them. E.g I cannot get
```cpp
/// @returns true when normally halted; false when exceptionally halted.
bool executeTransaction(Executive& _e, Transaction const& _t,
    OnOpFunc const& _onOp)
{
    _e.initialize(_t);

    if (!_e.execute())
        _e.go(_onOp);
    return _e.finalize();
}
```

Also I'm not sure we can get `)` on a new line.

libethereum/State.h
```diff
@@ -169,200 +197,250 @@ using ChangeLog = std::vector<Change>;
...
+    void setCode(Address const& _address, bytes&& _code);
+
+    /// Delete an account (used for processing suicides).
+    void kill(Address _a);
+
+    /// Get the storage of an account.
+    /// @note This is expensive. Don't use it unless you need to.
+    /// @returns map of hashed keys to key-value pairs or empty map if no
+    /// account exists at that address.
+    std::map<h256, std::pair<u256, u256>> storage(
```
pirapira: I find this line ending with `(` a bit scary.  I think I can get used to it.

chfast: The only tradeoff we can have here is to allow putting the return type on its own line... let me try this.

chfast:
Let's discuss argument & parametr breaking options. The following example does not show "bin packing" precisely, but maybe we are able to choose top candidates. I'm not sure either all of the options are implementable in clang-format.

```cpp

// Option 0.
unsigned long int computeCosmologicalConstant(int einsteinAge, double jupiterCollapseProb, bool unknownFactor)
{
    return einsteinAge * jupiterCollapseProb + unknownFactor;
}

// Option 1.
unsigned long int computeCosmologicalConstant(int einsteinAge,
    double jupiterCollapseProb, bool unknownFactor)
{
    return einsteinAge * jupiterCollapseProb + unknownFactor;
}

// Option 2.
unsigned long int computeCosmologicalConstant(int einsteinAge,
                                              double jupiterCollapseProb,
                                              bool unknownFactor)
{
    return einsteinAge * jupiterCollapseProb + unknownFactor;
}

// Option 3.
unsigned long int computeCosmologicalConstant(
    int einsteinAge, double jupiterCollapseProb, bool unknownFactor)
{
    return einsteinAge * jupiterCollapseProb + unknownFactor;
}

// Option 4.
unsigned long int computeCosmologicalConstant(
    int einsteinAge,
    double jupiterCollapseProb,
    bool unknownFactor)
{
    return einsteinAge * jupiterCollapseProb + unknownFactor;
}

// Option 5.
unsigned long int computeCosmologicalConstant(int einsteinAge,
    double jupiterCollapseProb, bool unknownFactor
)
{
    return einsteinAge * jupiterCollapseProb + unknownFactor;
}

// Option 6.
unsigned long int computeCosmologicalConstant(int einsteinAge,
                                              double jupiterCollapseProb,
                                              bool unknownFactor
)
{
    return einsteinAge * jupiterCollapseProb + unknownFactor;
}

// Option 7.
unsigned long int computeCosmologicalConstant(
    int einsteinAge, double jupiterCollapseProb, bool unknownFactor
)
{
    return einsteinAge * jupiterCollapseProb + unknownFactor;
}

// Option 8.
unsigned long int computeCosmologicalConstant(
    int einsteinAge,
    double jupiterCollapseProb,
    bool unknownFactor
)
{
    return einsteinAge * jupiterCollapseProb + unknownFactor;
}
```

gumb0:
For me option 0 is the best and if the line gets too ling, then it gets wrapped as in option 1.

Closing parenthesis on a separate line (options 5-8) is super weird

chfast:
My discovery about clang-format: when it has to break function parameters it prefers to put all params in the next line if they fit instead of leaving some in the first line. See example:

```
unsigned long int computeCosmologicalConstant(
        int einsteinAge, double jupiterCollapseProb, bool unknownFactor)
{
    return einsteinAge * jupiterCollapseProb + unknownFactor;
}

bool executeTransaction(Executive& _e, Transaction const& _t,
        OnOpFunc const& _onOp, int hello_world)
{
    _e.initialize(_t);

    if (!_e.execute())
        _e.go(_onOp);
    return _e.finalize();
}
``` 

There is a config option `AllowAllParametersOfDeclarationOnNextLine: false` but it only considered for format without bin-packing (every param on its own line).

Is this ok for you?

pirapira:
@chfast everything on the second line is better than bin-packing for me.

chfast:
This is my final proposal.

I also reformatted 2 old files. None of the code changes are going to be committed, only the clang-format config.

chfast:
@pirapira that's the case I believe, isn't it?

pirapira:
@chfast yes, it was a response to your question above.

gumb0:
I think `ColumnLimit: 80` is too narrow, too many line breaks

libdevcore/RLP.h
```diff
@@ -63,415 +86,780 @@ template <class T> struct Converter { static T convert(RLP const&, int) { BOOST_
...
+    explicit operator bool() const { return !isNull(); }
+
+    /// No value.
+    bool isNull() const { return m_data.size() == 0; }
+
+    /// Contains a zero-length string or zero-length list.
+    bool isEmpty() const
+    {
+        return !isNull() && (m_data[0] == c_rlpDataImmLenStart ||
+                                m_data[0] == c_rlpListStart);
```
gumb0: I don't understand this alignment. What did it align with?

pirapira: four more indentations than `(`.

pirapira:
Up to 92, I'm completely happy.  My life changes a bit (vertical split / horizontal split) around 120.

libdevcore/RLP.h
```diff
@@ -63,415 +86,780 @@ template <class T> struct Converter { static T convert(RLP const&, int) { BOOST_
...
+    size_t size() const { return isData() ? length() : 0; }
+    size_t sizeStrict() const
+    {
+        if (!isData())
+            BOOST_THROW_EXCEPTION(BadCast());
+        return length();
+    }
+
+    /// Equality operators; does best-effort conversion and checks for equality.
+    bool operator==(char const* _s) const
```
gumb0: I though there was an option to allow one-liners for inline function implementations? Or was this wider that 80 symbols?

libethereum/State.cpp
```diff
@@ -21,720 +21,777 @@
...
+                std::set<u256> cached;
+                if (r)
+                {
+                    SecureTrieDB<h256, OverlayDB> memdb(
+                        const_cast<OverlayDB*>(&_s.m_db),
+                        r[2].toHash<h256>());  // promise we won't alter the
+                                               // overlay! :)
+                    for (auto const& j : memdb)
+                        mem[j.first] = RLP(j.second).toInt<u256>(),
+                        back.insert(j.first);
```
gumb0: hm, weird, I would expect this line to be indented one level more

gumb0:
I'd try width 120

libethereum/State.cpp
gumb0: ah it's two expressions separated by comma in one loop-body-statement, ok, it's the code itself that is weird

chfast:
The column limit is not about how much you can fit on the screen (I checked I can fit ~160 in my current layout). I think there are 2 reasons for it.

1. Shorter lines are easier to read, line columns of text in newspaper. Here we can argue that a lot of breaks also make it less readable.
2. There are use cases when you 2 files on the screen side by side. One example is diff review (see PRs on GitHub).

We can argue long days what number would be the best. To avoid that I picked the one that is commonly used.

Personal note: after using the 80 limit in Python I get used to it and started write code in the way that it fit nicely.

libdevcore/RLP.h
chfast: Weird, but I think it's continuation (4) from `(`.

libdevcore/RLP.h
chfast: It does not fit (82 chars).

libethereum/State.cpp
chfast: Wow. This is comma operator. It actually means:
```cpp
for (auto const& j : memdb)
{
    mem[j.first] = RLP(j.second).toInt<u256>();
    back.insert(j.first);
}
```

chfast:
I can propose 2 solutions:

1. 80, because (see arguments above). Then we review this decision after some time (in a month). If there are complains we increase it to 100.
2. We go with 100 as a compromise straight away.

I prefer option 1.

gumb0:
Some examples I find repelling because of 80:
```
        _itemCount = 1;  // for all following iterations, we've effectively
                         // appended a single item only since we completed a
                         // list.
```

```
            if (c_rlpDataIndLenZero + brbr > 0xff)
                BOOST_THROW_EXCEPTION(RLPException() << errinfo_comment(
                                          "Number too large for RLP"));
```

```
    TransactionReceipt const receipt =
        _envInfo.number() >= _sealEngine.chainParams().byzantiumForkBlock ?
            TransactionReceipt(
                statusCode, startGasUsed + e.gasUsed(), e.logs()) :
            TransactionReceipt(
                rootHash(), startGasUsed + e.gasUsed(), e.logs());
```

These were one-liners on 11 lines, now all of them take 49 lines
```
    bool operator==(char const* _s) const
    {
        return isData() && toString() == _s;
    }
    bool operator!=(char const* _s) const
    {
        return isData() && toString() != _s;
    }
    bool operator==(std::string const& _s) const
    {
        return isData() && toString() == _s;
    }
...
```

Also, have you seen testeth code? 😆  Some lines start at column further than 80 https://github.com/ethereum/cpp-ethereum/blob/develop/test/tools/libtesteth/ImportTest.cpp#L148

(and you can't really compare python, it's much more compact)

I would be ok with trying 100.

chfast:
clang-format config: #4706.

PR #4672, opened 2017-12-01
cpp-ethereum#4598 1)Switch to C++17 style byte. 2)Fixed all compiler …

lazyGoose:
This PR: 
1) Switch to C++17 style byte.
2) Fixed all compiler errors in `libdevcore` directory.

According to a `C++ 17` doc, 
1) `std::byte` is not an integer and not a character.
2) `std::byte` is storage of bits.

So, I have to define all `operators` for the new `byte`. It looks a little messy, but I can gradually change all functions so that those crypto functions don’t require `int`, `char` and `bigint` to interact with `byte`. In this way, the code should be much nicer~

### DIFF
--- a/libdevcore/Base64.cpp
+++ b/libdevcore/Base64.cpp
@@ -33,17 +33,17 @@ using namespace dev;
 
 static inline bool is_base64(byte c)
 {
-	return (isalnum(c) || (c == '+') || (c == '/'));
+	return (isalnum(as_unsigned_char(c)) || (c == '+') || (c == '/'));
 }
 
 static inline byte find_base64_char_index(byte c)
 {
 	if ('A' <= c && c <= 'Z') return c - 'A';
-	else if ('a' <= c && c <= 'z') return c - 'a' + 1 + find_base64_char_index('Z');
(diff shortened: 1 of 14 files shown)


gumb0:
Checkout CI build failures, for example
```
In file included from /home/travis/build/ethereum/cpp-ethereum/libdevcore/OverlayDB.cpp:23:0:
/home/travis/build/ethereum/cpp-ethereum/libdevcore/TrieDB.h: In member function ‘void dev::GenericTrieDB<_DB>::iterator::Node::incrementChild()’:
/home/travis/build/ethereum/cpp-ethereum/libdevcore/TrieDB.h:144:83: error: operands to ?: have different types ‘int’ and ‘dev::byte’
    void incrementChild() { child = child == 16 ? 0 : child == 15 ? 17 : (child + 1); }
                                                                                   ^
```

```
C:\projects\cpp-ethereum\libdevcore\CommonJS.cpp(73): error C2593: 'operator ==' is ambiguous [C:\projects\cpp-ethereum\build\libdevcore\devcore.vcxproj]
```

PR #4673, opened 2017-12-01
Add jsonrpcproxy.py

chfast:
Hey @chriseth @gumb0,
I'm shipping the JSON-RPC Proxy (HTTP -> Unix Socket).

Usage is simple `./jsonrpcproxy.py` or  `./jsonrpcproxy.py --help`.

I'm not sure a documentation in eth is needed, but I may add some information to `eth -j` later on.

### DIFF
--- a/scripts/jsonrpcproxy.py
+++ b/scripts/jsonrpcproxy.py
@@ -1,0 +1,113 @@
+#!/usr/bin/env python3
+
+"""
+JSON-RPC Proxy
+
+This Python script provides HTTP proxy to Unix Socket based JSON-RPC servers.
+Check out --help option for more information.
+
+Build with cython:
+
+cython rpcproxy.py --embed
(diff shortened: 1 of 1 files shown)


scripts/jsonrpcproxy.py
```diff
@@ -0,0 +1,151 @@
...
+class Proxy(HTTPServer):
+
+    def __init__(self, proxy_url, backend_url):
+
+        proxy_url = urlparse(proxy_url)
+        assert proxy_url.scheme == 'http'
+        proxy_address = proxy_url.hostname, proxy_url.port
+
+        backend_url = urlparse(backend_url)
+        assert backend_url.scheme == 'unix'
```
gumb0: this now fails for me on windows

chfast: It is not ready yet. I'm moving to Windows to finish it.

PR #4674, opened 2017-12-02
Remove Whisper

chfast:


isaachier:
I like the idea of removing deprecated API.

winsvega:
I have not worked with whisper. But I guess this one is outdated anyways.

PR #4675, opened 2017-12-03
Clang format

isaachier:
Sorry to duplicate #4670 but thought you might want to see my work here as well. Not sure which you prefer.


chfast:
Yes, thanks.

Can you reformat some files including `libethereum/State.*` with it?

.clang-format
```diff
@@ -70,14 +80,15 @@ ObjCBlockIndentWidth: 2
 ObjCSpaceAfterProperty: false
 ObjCSpaceBeforeProtocolList: false
 PenaltyBreakBeforeFirstCallParameter: 1
-PenaltyBreakComment: 300
-PenaltyBreakFirstLessLess: 120
-PenaltyBreakString: 1000
-PenaltyExcessCharacter: 1000000
-PenaltyReturnTypeOnItsOwnLine: 200
+PenaltyBreakComment: 100
```
chfast: Can you explain what you get from adjusting penalties?

isaachier: Sorry didn't see your comments before. I think the number itself is mostly arbitrary so just a matter of how much you want the formatter to avoid a strategy. In my version, I believe the formatter will try splitting a string before it exceeds the column limit.

chfast:
The .clang-format merged in #4706. Thank you.

PR #4678, opened 2017-12-04
running singletest without filler from custom test suite

winsvega:
move singletest without filler logic into custom test suite


test/tools/libtesteth/Options.cpp.orig
```diff
@@ -0,0 +1,347 @@
...
+			else if (indentLevel == "1")
+				logVerbosity = Verbosity::NiceReport;
+			else
+				logVerbosity = Verbosity::Full;
+
+			int indentLevelInt = atoi(argv[i]);
+			if (indentLevelInt > g_logVerbosity)
+				g_logVerbosity = indentLevelInt;
+		}
+<<<<<<< 6d548d541c87e3a7fb93b70aa2db3449aaea36c4
```
pirapira: Some git garbage is being committed.

pirapira: I think this whole file can be removed.

winsvega:
wtf with github. I have already fixed this issue and it pops up again.

PR #4683, opened 2017-12-05
disable rndStateTest on macOs9

winsvega:
this should disable the test which sometimes fails on macos9

this test is about random code function which is used on linux. for some reason it sometimes fails on macos. it is not reproducable and hard to trace. but it makes os9 builds red.
so as long as it works on travis its fine

### DIFF
--- a/test/tools/fuzzTesting/fuzzHelper.cpp
+++ b/test/tools/fuzzTesting/fuzzHelper.cpp
@@ -640,12 +640,14 @@ BOOST_AUTO_TEST_CASE(rndStateTest)
 	try
 	{
 		//Mac os bug is here
-		//Analyze
+		int currVerbosity = g_logVerbosity;
+		g_logVerbosity = 10;
 		test::StateTestSuite suite;
 		test::RandomCodeOptions options;
 		test::TestOutputHelper::get().setCurrentTestFileName(std::string());
 		std::string test = test::RandomCode::get().fillRandomTest(suite, c_testExampleStateTest, options);
 		BOOST_REQUIRE(!test.empty());
+		g_logVerbosity = currVerbosity;
 	}
 	catch(dev::Exception const& _e)
 	{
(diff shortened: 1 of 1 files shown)


chfast:
For me it is not fine that "for some reason it sometimes fails". I would like to know why it fails.

> so as long as it works on travis its fine

No, because Travis does not run any tests on macOS.

pirapira:
I'm sure this failure is pointing to some useful fix.

winsvega:
hm. it does not disable the test. ok maybe we could investigate in this PR what happens on macOs with this test.

chfast:
Maybe this is something about "out-of-memory" because of the other processes running in parallel.

You can login to Circle CI build via SSH and try running this single tests 100 times in parallel.

You can also enabled core dumps and/or print stack trace of the crash...

winsvega:
does it have my key? if you manage the CI here is my key
```
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDGGDUpmaOveRb2oRLmp77k7H/v0W2aagnrkEHAvtMarElBSlzEVfN9wRiPerio1wTGsieVW+urmxy7u26+rplExFVP28U5oxuZIse4sJ1EV28wD43cIHKqvCbfdJbi+3EQIjfwtO73xGoQ7xivgiTAUPwvP1bA1eqaZBU7tsmxik1O+gQVnGqswR23qz4WWTa7Dl3/g9M6qgfKW/dTN1Dc0LfUC07Zu3x6vk5bV+VWFqnu72R4DArZR6k8zUepkuuivTuy9la/R+p1ZOxy16X4EcSQNUcpKsgpi75MlmGANNb7eQPjJ0BFcOKzWnaCbWyzWVKsBGaeBXTugVvFsCnd wins@wins-Aspire-7750G
````

chfast:
@winsvega It will have your key from GitHub. On CircleCI press button "rebuild with SSH" - you will get instructions how to connect.

winsvega:
this test is not failing nor in testeth execution nor in ctest run. 
I've tried to run it in a loop. it just does not want to fail. 
maybe we could add ctest -V flag to the builds just to get more info?

chfast:
The `ctest` command is here: https://github.com/ethereum/cpp-ethereum/blob/develop/circle.yml#L57.

chfast:
Please see #4724.

PR #4687, opened 2017-12-06
testeth throw an error if -d -g -v --singlenet select no transactions

winsvega:


test/tools/jsontests/StateTests.cpp
```diff
@@ -104,10 +105,17 @@ json_spirit::mValue StateTestSuite::doTests(json_spirit::mValue const& _input, b
...
-					importer.checkGeneralTestSection(exp.get_obj(), wrongTransactionsIndexes, i->first);
+					if (importer.checkGeneralTestSection(exp.get_obj(), wrongTransactionsIndexes, i->first))
+						foundResults = true;
 				}
 			}
 
+			if (!foundResults)
+			{
+				Options const& opt = Options::get();
+				BOOST_ERROR("Transaction not found! (Network: " + (opt.singleTestNet.empty() ? "Any" : opt.singleTestNet) + ", dataInd: " + toString(opt.trDataIndex) + ", gasInd: " + toString(opt.trGasIndex) + ", valInd: " + toString(opt.trValueIndex) + ")");
```
pirapira: Do we want to cause this error even when `--singletest` is not specified?

winsvega: especially if it is not specified. that would mean that execution logic is broken (no tests executed)

pirapira: Really? `testeth -- -d 3` executes some tests.  Those tests with less than three data are skipped.  Do you want an error for this command?

winsvega: ah, common. it is about an error in a single test suite. 
in yout example above all test suites are executed and some of them would be skipped. 
then I think we should allow -d -v -g --singlenet only if a single test suite (single doTest function)

winsvega: well if this would be a static variable then I could check it after execution all tests. but that would be not elegant.

pirapira: > I think we should allow -d -v -g --singlenet only if a single test suite (single doTest function)

Maybe.  For now, let's error only when `--singletest` is set.

winsvega: hm. so wait. it continues the execution. so you right it might be a warning. just throw a warning if no tests from the test suites are executed.

pirapira: That works for me.

winsvega: hm. boost WARN_MESSAGE  does not print to console.

pirapira: OK.  An error is good then.

PR #4689, opened 2017-12-07
Fix high memory consumption during snapshot import

gumb0:
The accounts with huge storage are splitted between several chunks of snapshot. Previous approach was to first collect the full storage of such account in `std::map` in memory and then construct a trie out of it and save to database. This consumed a lot of RAM.

The new approach is to commit to the database the parts of the storage of each chunk (so there's one commit per each chunk).

### DIFF
--- a/libethereum/StateImporter.cpp
+++ b/libethereum/StateImporter.cpp
@@ -72,6 +72,7 @@ public:
 	std::string lookupCode(h256 const& _hash) const override { return m_trie.db()->lookup(_hash); }
 
 private:
+	// can be used only with already imported accounts
 	h256 accountStorageRoot(h256 const& _addressHash) const
 	{
 		std::string const account = m_trie.at(_addressHash);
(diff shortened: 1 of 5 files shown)


libethereum/StateImporter.cpp
```diff
@@ -68,12 +65,24 @@ class StateImporter: public StateImporterFace
...
 	h256 stateRoot() const override { return m_trie.root(); }
 
 	std::string lookupCode(h256 const& _hash) const override { return m_trie.db()->lookup(_hash); }
 
 private:
-	bool containsAccount(h256 const& _addressHash) const { return m_trie.contains(_addressHash); }
+	h256 accountStorageRoot(h256 const& _addressHash) const
+	{
+		std::string const account = m_trie.at(_addressHash);
+		assert(!account.empty());
```
chfast: Please add a comment that this function can only be used for existing accounts.

PR #4692, opened 2017-12-08
evm: Fix CREATE2 implementation

chfast:
The memory requirements for CREATE2 were not correctly checked.


chfast:
No regressions on my machine, but I will wait for all builds...

pirapira:
That means tests are missing.

chfast:
@winsvega said we don't have any tests for Constantinople.

winsvega:
we have some transaction tests.
for constantinople there was zeroSig tests. but its removed from the repo for now.
new tests are not developed. because no specs (EIPs) and roadmap is defined.

PR #4695, opened 2017-12-10
Add .ycm_conf_extra.py

isaachier:
For those using "YouCompleteMe," this is pretty helpful. Mostly same as template but adds Hunter directories too.


chfast:
Does it have to be in the main dir?

isaachier:
Yes. See https://github.com/valloric/YouCompleteMe

PR #4697, opened 2017-12-12
Database improvements

isaachier:
Allow user to choose from LevelDB, RocksDB, and LMDB using `cmake -DDB=` (valid values: `LEVELDB`, `ROCKSDB`, `LMDB`). Improve interface for database interaction. Avoid storing read/write options across the application code.


cmake/FindLMDB.cmake
```diff
@@ -0,0 +1,28 @@
...
+find_library(LMDB_LIBRARIES NAMES lmdb   PATHS "$ENV{LMDB_DIR}/lib" )
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(LMDB DEFAULT_MSG LMDB_INCLUDE_DIR LMDB_LIBRARIES)
+
+if(LMDB_FOUND)
+  message(STATUS "Found lmdb    (include: ${LMDB_INCLUDE_DIR}, library: ${LMDB_LIBRARIES})")
+  mark_as_advanced(LMDB_INCLUDE_DIR LMDB_LIBRARIES)
+
+  caffe_parse_header(${LMDB_INCLUDE_DIR}/lmdb.h
```
isaachier: Need to fix this. Also need to add option for locally installed lmdb++ and/or copy the header to the source tree.

cmake/FindLMDB.cmake
```diff
@@ -0,0 +1,28 @@
+# Try to find the LMBD libraries and headers
```
chfast: Why do you need this file if using `find_package(LMDB CONFIG)`?

chfast:
Thanks for this PR, but it does many changes at once, so reviewing this will take some time.

isaachier:
No worries it took me a while to do this as well.

cmake/FindLMDB.cmake
isaachier: I assumed you wanted compatibility for non-Hunter builds. I wouldn't do it if I hadn't seen the other find modules there.

chfast: No needed. We keep Find modules only for packages not in Hunter.

chfast:
Is there an option to split this into smaller parts?

isaachier:
I'm not sure if it would help. The major change is hiding the database implementation behind a DB interface so that is a single cohesive change. The addition of a rocksdb and lmdb accounts for some new files that could be added in another review but idk how much that would help you.

gumb0:
So now it gets LevelDB from Hunter and LevelDB from Hunter does not support Windows as I understand...

isaachier:
@gumb0 LevelDB overall does not support Windows. @chfast had suggested using the fork from bitcoin-core, but I think they do a major rewrite. Not sure how Windows is supported today, but it is not through standard LevelDB with MSVC. If there is a way to build LevelDB on a Windows host, Hunter should have no problem there either.

cmake/EthOptions.cmake
```diff
@@ -11,14 +11,59 @@ macro(configure_project)
...
 	option(PARANOID "Enable additional checks when validating transactions (deprecated)" OFF)
 	option(MINIUPNPC "Build with UPnP support" OFF)
 	option(FASTCTEST "Enable fast ctest" OFF)
 
+	set(supported_dbs LEVELDB ROCKSDB LMDB)
+	string(REPLACE ";" ", " supported_dbs_str "${supported_dbs}")
+	set(DB LEVELDB CACHE STRING "Database implementation to use (${supported_dbs_str})")
+	set_property(CACHE DB PROPERTY STRINGS ${supported_dbs})
+	set(valid_db OFF)
+	foreach(supported_db ${supported_dbs})
```
chfast: Don't do this validation. If CMake does not support it out of the box, forget it.

cmake/EthOptions.cmake
```diff
@@ -11,14 +11,59 @@ macro(configure_project)
...
+			string(TOLOWER "${DB}" hunter_name)
+		endif()
+		hunter_add_package(${hunter_name})
+
+		if(DB STREQUAL LMDB)
+			find_package(lmdbxx CONFIG REQUIRED)
+		elseif(DB STREQUAL ROCKSDB)
+			find_package(RocksDB CONFIG REQUIRED)
+		else()
+			find_package(leveldb CONFIG REQUIRED)
```
chfast: Do not take leveldb from Hunter until it supports Windows. So: lmdbxx and rocksdb from Hunter only, leveldb from system only.

chfast:
It would be easier if we did DB interface changes first for leveldb only, then add rocksdb and lmdb in a separate PR.

gumb0:
Also I strongly dislike the practice of `#ifdef`-ing the whole contents of the file. It should be handled with conditional compilation in cmake files instead.

libethereum/BlockChain.cpp
```diff
@@ -612,31 +599,33 @@ void BlockChain::insert(VerifiedBlockRef _block, bytesConstRef _receipts, bool _
...
-	extrasBatch.Put(toSlice(_block.info.hash(), ExtraReceipts), (ldb::Slice)_receipts);
+	extrasTransaction->insert(toSlice(_block.info.hash(), ExtraDetails), (db::Slice)dev::ref(bd.rlp()));
+	extrasTransaction->insert(toSlice(_block.info.hash(), ExtraLogBlooms), (db::Slice)dev::ref(blb.rlp()));
+	extrasTransaction->insert(toSlice(_block.info.hash(), ExtraReceipts), (db::Slice)_receipts);
 
-	ldb::Status o = m_blocksDB->Write(m_writeOptions, &blocksBatch);
-	if (!o.ok())
+	try
+	{
+		blocksTransaction->commit();
```
chfast: I don't like this design where you have to wrap every DB call with a try-catch block. If using exceptions in the API does not allow you to have more generic exception handler this does not make much sense.

libdevcore/db.h
```diff
@@ -21,16 +21,80 @@
...
+	}
+};
+
+inline std::ostream& operator<<(std::ostream& out, const DB& database)
+{
+	database.print(out);
+	return out;
+}
+
+struct FailedToOpenDB: virtual Exception { using Exception::Exception; };
```
chfast: I like you didn't use Boost here. I believe using Boost Exceptions does not give you a lot, but obscures code, especially `BOOST_THROW_EXCEPTION`.

But, I'm not sure it is good to have so much exception types. Can we try to have single exception type for database this time?

@gumb0 ?

gumb0: What do you mean they didn't use boost, `Exception` is derived from `boost::exception`?

In general I think having a class for each kind of exception was not a very good idea, but I'm not sure about changing this practice now.
Different errors still should be distinguishable, we'd need to create some `enum` for error code if we use one exception class for db errors...

gumb0: `BOOST_THROW_EXCEPTION` gives us output of source code location of throw in debug build, this is valuable

gumb0: Also not using `DEV_SIMPLE_EXCEPTION` macro to define exceptions here goes against the practice of surrounding code and this way name of exception class won't be output

libethereum/BlockChain.cpp
isaachier: Yes I agree with you to an extent although I followed the perhaps outdated style guide that suggested using exceptions instead of error values where possible.

libdevcore/db.h
isaachier: Actually I took the errors here from the since-removed WhisperDB files.

isaachier:
Closing in favor of smaller pull requests.

libethereum/BlockChain.cpp
gumb0: That's a valid suggestion, but don't add try-catch around every call, make a more general try-catch somewhere above for all of the throwing calls (catching `Exception` probably)

Here to distinguish between writing to blocks and extras you could have two different exception classes

cmake/EthOptions.cmake
isaachier: OK I will update other diff with the advice from here. Still not sure how any version of leveldb is working on Windows unless it is the forked version from bitcoin.

libethereum/BlockChain.cpp
isaachier: Honestly, not sure how this is much different than the `if`/`else` that was here originally. The idea of exiting on commit failure still has to be explicit.

gumb0: The whole `BlockChain::insert` method can have probably a single try/catch, a single concise error handler probably. That's very different from if/else for each call.

isaachier: The original implementation was a bunch of if/else statements which are now try/catch statements. I like your suggestion for a large unified try/catch block. Just wondering if the if/else was acceptable, and if so, why multiple try/catch is any worse.

gumb0: The original I guess was the old code not really in compliance with current coding standards

tiendq:
> LevelDB overall does not support Windows.

@isaachier Sorry to ask this question too late :), regarding your comment quoted above how can `aleth` run on Windows? As far as I know, there is still only `leveldb` used in current codebase.

@gumb0 what is status of `RocksDB` now? #4844 is likely suspended for awhile.

isaachier:
@tiendq Bitcoin maintains a custom branch of LevelDB to support Windows. Personally, I don't view this as standard LevelDB, because the API can diverge significantly from the standard LevelDB.

tiendq:
@isaachier I was asking about this project, not Bitcoin :). `ethereum-aleth` uses standard LevelDB, doesn't it? Is it running on Windows? (I'm using macOS).

I installed LevelDB globally with `brew` on macOS.

isaachier:
See @gumbo's comment above: https://github.com/ethereum/aleth/pull/4697#issuecomment-351416635. Hunter is not used here because of that limitation.

gumb0:
@tiendq Status of RocksDB integration: we were waiting for release 5.14. now when it's out someone has to continue work in https://github.com/ethereum/aleth/pull/4844

chfast:
The rocksdb 5.14 is already in Hunter and it builds on Windows (I added both this version and Hunter CI tests for it on Windows). We are ready to continue...

PR #4704, opened 2017-12-13
Add CORS headers.

chriseth:


chriseth:
@chfast it completely disables a certain client-side browser protection, so you should rather be aware of the consequences :)

chfast:
It looked to me like that.

PR #4706, opened 2017-12-14
Add clang-format config file

chfast:


CONTRIBUTING.md
```diff
@@ -1,5 +1,12 @@
...
+# cpp-ethereum Contributing and Review Guidelines
 
-1. All submitted Pull Requests are assumed to be ready for review 
-   unless they are labeled with `[in progress]` 
-   or have "[WIP]" in title.
+1. All submitted **Pull Requests** are assumed to be ready for review 
+   unless they are labeled with `[in progress]` or have "[WIP]" in title.
+
+2. **Code formatting** rules are described by the [Clang-Format Style Options] file [.clang-format].
+   Please use the [clang-format] tool to format your code _changes_ accordingly.
```
gumb0: Maybe add the exact command to do this?

chfast: The example below is ok?

JoeLoser: I think the example below is fine. Perhaps we should create a commit hook to apply clang-format? We can install the hook at CMake time.

chfast: Do you mean this hook? https://github.com/andrewseidl/githook-clang-format
Both the hook and installation with CMake seems quite invasive. Let's see what happens without it first.

JoeLoser: That particular one works, yep. While invasive certainly, it forces `clang-format` to always be run. Otherwise, if a PR gets merged without adhering to the style specified in `.clang-format`, then the next person's commit that touches one of those files and runs `clang-format` will see diffs on lines they did not originally touch, due to the previous person not running `clang-format`. So, having it run automatically for everyone eliminates this issue.

But, I do hear you that perhaps this is premature to worry about this before we have the issue of someone not running `clang-format` on their changed files.

chfast: It's hard to force automation on developers' side.

Let us figure out the process by practice. We have just found the [git-clang-format](https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/git-clang-format) tool. It seems to work great for our needs. We will document the workflow in CONTRIBUTING.md soon.

PR #4707, opened 2017-12-14
jsonrpcproxy: Improvements

### LINKED ISSUE #4705: jsonrpcproxy: Connect lazily
Do not fail if the Unix Socket is not there yet. Instead print a warning. Also keep the information about last successful connection established.

chfast:
This includes a change where jsonrpcproxy.py can be started before a backend client (at least on Linux and macOS).

Resolves #4705.

### DIFF
--- a/scripts/jsonrpcproxy.py
+++ b/scripts/jsonrpcproxy.py
@@ -32,9 +32,9 @@ BUFSIZE = 32
 DELIMITER = ord('\n')
 INFO = """JSON-RPC Proxy
 
-Version: {}
-Proxy: {}
-Backend: {}
+Version:  {version}
+Proxy:    {proxy_url}
+Backend:  {backend_url} (connected: {connected})
 """
 
 
@@ -56,6 +56,9 @@ class UnixSocketConnector(object):
         if os_error_number == errno.ECONNREFUSED:
             return "Connection to '{}' refused"
         return "Unknown error when connecting to '{}'"
+
+    def is_connected(self):
+        return self.socket is not None
 
     def _connect(self):
         if self.socket is None:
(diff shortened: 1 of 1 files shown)


scripts/jsonrpcproxy.py
```diff
@@ -31,13 +32,71 @@
...
+                raise err from ex
+
+    def _reconnect(self):
+        self.socket.shutdown(socket.SHUT_RDWR)
+        self.socket.close()
+        self.socket = None
+        self._connect()
+
+    def recv(self, max_length):
+        self._connect()
```
gumb0: is it possible in python instead of explicitly calling this to have `get_socket()` method that will return `self.socket` intitializing it if needed ?

chfast: You mean `self.get_socket().recv(...)`? Yes. You can even have a property like this `self.socket.recv(...)`, but I'd like to avoid Pytonish features.

I will try this, let's see how it looks.

scripts/jsonrpcproxy.py
```diff
@@ -31,13 +32,70 @@
...
+                # Assign last, to keep it None in case of exception.
+                self._socket = s
+            except OSError as ex:
+                msg = self._get_error_message(ex.errno)
+                err = BackendError(msg.format(self._socket_path))
+                raise err from ex
+        return self._socket
+
+    def close(self):
+        self._socket.shutdown(socket.SHUT_RDWR)
```
gumb0: would be better to add
```
if self._socket is None:
  return
```

chfast: Right.

chfast: Fixed.

PR #4708, opened 2017-12-15
LevelDB behind generic DB interface

isaachier:


circle.yml
```diff
@@ -56,14 +56,19 @@ defaults:
 
         ctest --output-on-failure -j $TEST_PARALLEL_JOBS
 
-  store-eth: &store-artifacts
+  store-eth: &store-eth
```
isaachier: Not sure where these changes came from. Let me know if this looks right or if I reverted something.

isaachier:
Diff looks weird. Closing for now.

PR #4709, opened 2017-12-15
Database refactoring - leveldb changes

isaachier:


gumb0:
I haven't yet looked into your changes much, but think we're fine now with selecting only one at compile time (say LevelDB which we currently use), but the interfaces should be designed in such a way that it shouldn't be very difficult to implement selecting/switching databases in runtime.

isaachier:
I agree that makes sense. The switch from `ifdef` to CMake selection is straightforward. Just means getting rid of globs that currently select source files and making it a list. Hope that isn't too much of a hassle to maintain.

gumb0:
@chfast consideres the file list to be superior approach, I tend to agree

chfast:
CMake do not recommend using globs. List files manually and include db specific implementations depending on `-DDB` flag.

gumb0:
Still need to get LevelDB from system instead of hunter https://github.com/ethereum/cpp-ethereum/pull/4697#discussion_r157155477

isaachier:
Ya just been lazy... Getting to it

libdevcore/LevelDB.cpp
```diff
@@ -0,0 +1,197 @@
...
+	cpp-ethereum is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/** @file LevelDB.cpp
+ * @author Isaac Hier <isaachier@gmail.com>
```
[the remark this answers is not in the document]
isaachier: OK makes sense.

libdevcore/LevelDB.cpp
```diff
@@ -0,0 +1,197 @@
...
+{
+	if (m_db) {
+		DEV_IGNORE_EXCEPTIONS(rollback());
+	}
+}
+
+void LevelDBTransaction::insert(Slice const& _key, Slice const& _value)
+{
+	m_writeBatch.Put(
+		leveldb::Slice(_key.data(), _key.size()),
```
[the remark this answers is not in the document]
isaachier: Good idea

libdevcore/LevelDB.h
```diff
@@ -0,0 +1,62 @@
...
+
+namespace dev
+{
+namespace db
+{
+
+class LevelDB: public DB
+{
+public:
+	static leveldb::ReadOptions defaultReadOptions();
```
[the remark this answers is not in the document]
isaachier: I think they are used in the transaction helper function. Also, they are useful for creating a sane default argument that the caller can modify before constructing the database object.

libdevcore/LevelDB.h
```diff
@@ -0,0 +1,62 @@
...
+	std::string lookup(Slice const& _key) const override;
+	bool exists(Slice const& _key) const override;
+	void insert(Slice const& _key, Slice const& _value) override;
+	void kill(Slice const& _key) override;
+	std::unique_ptr<Transaction> begin() override;
+	void forEach(std::function<bool(Slice const&, Slice const&)> f) const override;
+
+private:
+	std::unique_ptr<leveldb::DB> m_db;
+	leveldb::ReadOptions m_readOptions;
```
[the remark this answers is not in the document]
isaachier: Yes but `const` members are annoying for movable/copyable objects so I rarely ever make `const` members.

libdevcore/OverlayDB.cpp
```diff
@@ -96,9 +89,14 @@ bytes OverlayDB::lookupAux(h256 const& _h) const
 	std::string v;
 	bytes b = _h.asBytes();
 	b.push_back(255);	// for aux
-	m_db->Get(m_readOptions, bytesConstRef(&b), &v);
-	if (v.empty())
+	try
+	{
+		v = m_db->lookup(db::Slice(reinterpret_cast<char const*>(&b[0]), b.size()));
+	}
+	catch (const db::FailedLookupInDB& ex)
```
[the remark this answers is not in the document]
isaachier: Will remove

libdevcore/OverlayDB.cpp
```diff
@@ -135,18 +130,24 @@ void OverlayDB::kill(h256 const& _h)
...
-		if (ret.empty() && _h != EmptyTrie)
-			cnote << "Decreasing DB node ref count below zero with no DB node. Probably have a corrupt Trie." << _h;
-
-		// TODO: for 1.1: ref-counted triedb.
+		{
+			try
+			{
+				ret = m_db->lookup(db::Slice(reinterpret_cast<char const*>(_h.data()), 32));
+			}
+			catch (const db::FailedLookupInDB& ex)
```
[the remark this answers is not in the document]
isaachier: Same

libdevcore/db.h
```diff
@@ -21,16 +21,80 @@
...
+namespace dev
+{
+namespace db
+{
+
+// Transaction implements database transaction for a specific concrete database
+// implementation. The transaction implementation must call rollback in its
+// destructor if commit has not yet been called (making sure to catch any
+// exceptions).
+class Transaction
```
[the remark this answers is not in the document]
isaachier: Actually, `transaction` has somewhat broader implications than `batch`. A `transaction` is more correct in the case of `LMDB` (to be submitted as separate PR) where the transactions surround read operations as well as write operations.

libdevcore/db.h
```diff
@@ -21,16 +21,80 @@
...
+{
+public:
+	virtual ~Transaction() = default;
+
+	virtual void insert(Slice const& _key, Slice const& _value) = 0;
```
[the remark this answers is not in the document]
isaachier: Interesting, will change.

libdevcore/db.h
```diff
@@ -21,16 +21,80 @@
...
+	virtual void kill(Slice const& _key) = 0;
+
+	virtual void commit() = 0;
```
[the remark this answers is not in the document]
isaachier: My issue with that is that the batch essentially loses all transactional elements when you treat it like a dumb container. lmdb has a rich transaction API compared to leveldb and rocksdb. So I can do it, but it feels like the wrong choice for an object-oriented design.

libdevcore/db.h
```diff
@@ -21,16 +21,80 @@
...
+	Transaction() = default;
+	// Noncopyable
+	Transaction(const Transaction&) = delete;
+	Transaction& operator=(const Transaction&) = delete;
+	// Movable
+	Transaction(Transaction&&) = default;
+	Transaction& operator=(Transaction&&) = default;
+};
+
+class DB
```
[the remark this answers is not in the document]
isaachier: OK

libdevcore/db.h
```diff
@@ -21,16 +21,80 @@
...
+{
+public:
+	virtual ~DB() = default;
+	virtual std::string lookup(Slice const& _key) const = 0;
+	virtual bool exists(Slice const& _key) const = 0;
+	virtual void insert(Slice const& _key, Slice const& _value) = 0;
+	virtual void kill(Slice const& _key) = 0;
+	virtual std::unique_ptr<Transaction> begin() = 0;
```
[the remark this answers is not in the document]
isaachier: I understand, will do

libdevcore/db.h
```diff
@@ -21,16 +21,80 @@
...
+
+	// A database must implement the `forEach` method that allows the caller
+	// to pass in a function `f`, which will be called with the key and value
+	// of each record in the database. If `f` returns false, the `forEach`
+	// method must return immediately.
+	virtual void forEach(std::function<bool(Slice const&, Slice const&)> f) const = 0;
+
+	virtual void print(std::ostream& out) const
```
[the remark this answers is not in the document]
isaachier: Left this open to alternative implementations for new databases down the line. Nothing overrides this yet.

libdevcore/db.h
```diff
@@ -21,16 +21,80 @@
...
+	{
+		forEach([&out](Slice const& key, Slice const& value) {
+			out << toHex(key) << " => " << toHex(value) << '\n';
+			return static_cast<bool>(out);
+		});
+	}
+};
+
+inline std::ostream& operator<<(std::ostream& out, const DB& database)
```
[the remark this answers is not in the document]
isaachier: Right, forgot about that in a few places.

libdevcore/LevelDB.h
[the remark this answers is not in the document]
gumb0: `const` members (as everything else `const`) make reasoning about the code easier: when you see it's const, you know for sure that nothing can modify it during the lifetime of the object

And I'd say classes that implement virtual interfaces very rarely or never should be movable/copyable

libdevcore/db.h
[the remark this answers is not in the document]
gumb0: Well this interface has methods only for modification, will it change in LMDB PR ?

libdevcore/LevelDB.h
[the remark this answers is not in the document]
gumb0: Ok makes sense

libdevcore/db.h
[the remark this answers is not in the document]
gumb0: I'm not a fan of mixing pure virtual interfaces with implementation like this. I'd leave it a a free function in this file until a real need to override it arises
(or maybe just move implementation into `operator<<`)

libdevcore/db.h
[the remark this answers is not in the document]
gumb0: Your variant is easier to use incorrectly and harder to use correctly. Looking at the interface the user needs to guess some implicit things: what happens to `Transaction` after `commit` or `rollback`, can I reuse it for another transaction? Is it allowed to use `Transaction` If `DB` is destroyed earlier? 
I'd say it's worse kind of object-oriented design

@chfast what do you think?

For reference, here's my suggestion once again
```
class WriteBatchFace
{
public:
	virtual ~WriteBatchFace() = default;

	virtual void insert(Slice _key, Slice _value) = 0;
	virtual void kill(Slice _key) = 0;
};

class DatabaseFace
{
public:
	virtual ~DatabaseFace() = default;
	virtual std::string lookup(Slice _key) const = 0;
	virtual bool exists(Slice _key) const = 0;
	virtual void insert(Slice  _key, Slice _value) = 0;
	virtual void kill(Slice _key) = 0;
	
	virtual std::unique_ptr<Transaction> createWriteBatch() const = 0;
	virtual void commit(std::unique_ptr<Transaction>&& _batch) = 0;

	virtual void forEach(std::function<bool(Slice const&, Slice const&)> f) const = 0;
};
```

libdevcore/db.h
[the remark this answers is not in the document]
gumb0: Also, can this be const method?

libdevcore/db.h
```diff
@@ -21,16 +21,80 @@
...
+	virtual void commit() = 0;
+	virtual void rollback() = 0;
+
+protected:
+	Transaction() = default;
+	// Noncopyable
+	Transaction(const Transaction&) = delete;
+	Transaction& operator=(const Transaction&) = delete;
+	// Movable
+	Transaction(Transaction&&) = default;
```
gumb0: Does it need to be movable? I don't see move constructor defined in `LevelDBTransaction`, I think I'd leave it nonmovable

gumb0:
Hi @isaachier, do you still plan to work on this?

isaachier:
Ya sorry. Been a bit busy with other things. Will try to get on this at some point this week.

libdevcore/db.h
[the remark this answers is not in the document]
isaachier: I guess you are correct that I cannot expose the broader notion because of limitations in the other implementations that only support transactional writes.

libdevcore/db.h
isaachier: Nonmovable doesn't really make sense to me. If anything, I would say the default should be movable until proven otherwise.

isaachier: Some evidence to back this up: https://stackoverflow.com/a/14304618/1930331.

libdevcore/db.h
```diff
@@ -32,62 +32,45 @@ namespace dev
...
+	WriteBatchFace() = default;
 	// Noncopyable
-	Transaction(const Transaction&) = delete;
-	Transaction& operator=(const Transaction&) = delete;
+	WriteBatchFace(const WriteBatchFace&) = delete;
+	WriteBatchFace& operator=(const WriteBatchFace&) = delete;
 	// Movable
-	Transaction(Transaction&&) = default;
-	Transaction& operator=(Transaction&&) = default;
+	WriteBatchFace(WriteBatchFace&&) = default;
```
gumb0: See the answer by Herb Sutter in the link you gave https://stackoverflow.com/a/14303116/1636308
What we are dealing with here is type 2, `WriteBatchFace` is going to be used only through pointers and references and it doesn't need to be movable. It's an interface anyway, making an interface movable doesn't  really make sense to me.

It doesn't even make `LevelDBWriteBatch` movable, it's still nonmovable now I think (has const memeber)
But doesn't need to anyway.

isaachier: That answer actually says nothing about making a type non-movable, only non-copyable. I changed it now regardless, but the question still stands. Also, I agree `const` members make move-assignment impossible, but not move-construction.

gumb0: _These do not need move construction or assignment for two reasons: They're not copyable, and they already have an even more efficient natural "move" operation -- you just copy/move the pointer to the object and the object itself doesn't have to move to a new memory location at all._

gumb0:
The tests are failing now with this error
```
Test Case "opendb": 
/home/travis/build/ethereum/cpp-ethereum/libdevcore/LevelDB.cpp(93): fatal error: in "dev::db::LevelDB::LevelDB(const string&, leveldb::ReadOptions, leveldb::WriteOptions, leveldb::Options)": boost::exception_detail::clone_impl<dev::db::FailedToOpenDB>: std::exception
/home/travis/build/ethereum/cpp-ethereum/test/unittests/libethereum/BlockChain.cpp(65): last checkpoint
```

You can check it locally with a command `testeth.exe -t BlockChainFrontierSuite/opendb`

isaachier:
Brief investigation here indicates that this is an issue with new BlockChain trying to open DB directory that already exists. It tries to delete old files first, but I suspect (from a few similar questions online) that the LOCK file is hard to delete, and BlockChain ignores that using `DEV_IGNORE_EXCEPTIONS`. The alternative would be to avoid naming databases with a static name like "/tmp/extras" and "/tmp/extras.old" and just generate new suffixes each time, like "/tmp/extrasX057" and "/tmp/extrasABC4" or something. Not sure if this change introduced anything, because that code hasn't changed at all.

gumb0:
The test checks that opening the same database twice results in `DatabaseAlreadyOpen` exception being thrown from `BlockChain` constructor. I think what now happens is another exception being thrown earlier from `LevelDB` constructor

isaachier:
@gumb0 thanks for the help. The bug should no longer be an issue.

winsvega:
oneUncleGeneration2_EIP158 is a blockchain test with a forth block containing an uncle header of block number 2

libdevcore/LevelDB.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+inline leveldb::Slice toLDBSlice(Slice _slice)
+{
+    return leveldb::Slice(_slice.data(), _slice.size());
+}
+
+class LevelDBWriteBatch: public WriteBatchFace
+{
+public:
+	explicit LevelDBWriteBatch(leveldb::WriteOptions _writeOptions): m_writeOptions(std::move(_writeOptions))
```
gumb0: `WriteOptions` are not needed for this class

libdevcore/LevelDB.cpp
```diff
@@ -0,0 +1,169 @@
...
+{
+	m_writeBatch.Delete(toLDBSlice(_key));
+}
+
+}
+
+leveldb::ReadOptions LevelDB::defaultReadOptions()
+{
+	leveldb::ReadOptions readOptions;
+	readOptions.verify_checksums = true;
```
gumb0: Did we have it on previously?

libdevcore/LevelDB.cpp
```diff
@@ -0,0 +1,169 @@
...
+	}
+	if (!db) {
+		BOOST_THROW_EXCEPTION(FailedToOpenDB("null database pointer"));
+	}
+	m_db.reset(db);
+}
+
+std::string LevelDB::lookup(Slice _key) const
+{
+	const leveldb::Slice key(_key.data(), _key.size());
```
gumb0: `const` should be after type name everywhere

libdevcore/LevelDB.cpp
```diff
@@ -0,0 +1,169 @@
...
+}
+
+std::unique_ptr<WriteBatchFace> LevelDB::createWriteBatch() const
+{
+	return std::unique_ptr<WriteBatchFace>(new LevelDBWriteBatch(m_writeOptions));
+}
+
+void LevelDB::commit(std::unique_ptr<WriteBatchFace>&& _batch)
+{
+	const auto status = m_db->Write(m_writeOptions, &static_cast<LevelDBWriteBatch*>(_batch.get())->writeBatch());
```
gumb0: I would prefer here `dynamic_cast` + check result is not null

libdevcore/OverlayDB.h
```diff
@@ -50,10 +55,7 @@ class OverlayDB: public MemoryDB
 private:
 	using MemoryDB::clear;
 
-	std::shared_ptr<ldb::DB> m_db;
-
-	ldb::ReadOptions m_readOptions;
-	ldb::WriteOptions m_writeOptions;
+	std::shared_ptr<db::DatabaseFace> m_db;
```
gumb0: I don't see the point of using `shared_ptr` here, can we change it to `unique_ptr` ?

gumb0: Oh, the point is in having copy assignment I guess... Could you then add copy constructor & assignment with `=default` to make it more explicit

libdevcore/OverlayDB.cpp
```diff
@@ -50,33 +40,36 @@ void OverlayDB::commit()
 			for (auto const& i: m_main)
 			{
 				if (i.second.second)
-					batch.Put(ldb::Slice((char const*)i.first.data(), i.first.size), ldb::Slice(i.second.first.data(), i.second.first.size()));
+					writeBatch->insert(db::Slice(reinterpret_cast<char const*>(i.first.data()), i.first.size), db::Slice(reinterpret_cast<char const*>(i.second.first.data()), i.second.first.size()));
```
gumb0: Too complicated expression, could you please make a helper function for converting `h256` to `Slice`?
Second `reinterpet_cast` here is not needed

libdevcore/OverlayDB.cpp
```diff
@@ -135,18 +130,24 @@ void OverlayDB::kill(h256 const& _h)
...
-		// No point node ref decreasing for EmptyTrie since we never bother incrementing it in the first place for
-		// empty storage tries.
-		if (ret.empty() && _h != EmptyTrie)
-			cnote << "Decreasing DB node ref count below zero with no DB node. Probably have a corrupt Trie." << _h;
-
-		// TODO: for 1.1: ref-counted triedb.
+		{
+			try
+			{
+				ret = m_db->lookup(db::Slice(reinterpret_cast<char const*>(_h.data()), 32));
```
gumb0: I think can just use `m_db->exists(...)` here instead of `lookup` in `try-catch`

libdevcore/LevelDB.cpp
isaachier: Yes. I copied these options from the callers to leveldb, just wrapped them here.

libdevcore/LevelDB.cpp
isaachier: Right. I wish clang-format could do this transformation.

libdevcore/LevelDB.cpp
gumb0: Actuallly I can't find where it was set to true, I think it was default `false`

libethereum/BlockChain.cpp
```diff
@@ -299,10 +296,17 @@ unsigned BlockChain::open(fs::path const& _path, WithExisting _we)
...
 	// TODO: Implement ability to rebuild details map from DB.
-	std::string l;
-	m_extrasDB->Get(m_readOptions, ldb::Slice("best"), &l);
-	m_lastBlockHash = l.empty() ? m_genesisHash : *(h256*)l.data();
-	m_lastBlockNumber = number(m_lastBlockHash);
+	try
+	{
+		auto const l = m_extrasDB->lookup(db::Slice("best"));
+		m_lastBlockHash = *(h256*)l.data();
+		m_lastBlockNumber = number(m_lastBlockHash);
```
gumb0: Can this be done once after the try-catch block instead of duplicating in try and in catch?

libethereum/BlockChain.cpp
```diff
@@ -427,19 +427,13 @@ void BlockChain::rebuild(fs::path const& _path, std::function<void(unsigned, uns
...
 string BlockChain::dumpDatabase() const
 {
-	stringstream ss;
-
-	ss << m_lastBlockHash << endl;
-	ldb::Iterator* i = m_extrasDB->NewIterator(m_readOptions);
-	for (i->SeekToFirst(); i->Valid(); i->Next())
-		ss << toHex(bytesConstRef(i->key())) << "/" << toHex(bytesConstRef(i->value())) << endl;
-	return ss.str();
+	// TODO
```
gumb0: This shouldn't be difficult to implement, right?

libethereum/BlockChain.cpp
```diff
@@ -1375,18 +1379,26 @@ bool BlockChain::isKnown(h256 const& _hash, bool _isCurrent) const
 	DEV_READ_GUARDED(x_blocks)
 		if (!m_blocks.count(_hash))
 		{
-			string d;
-			m_blocksDB->Get(m_readOptions, toSlice(_hash), &d);
-			if (d.empty())
+			try
+			{
+				m_blocksDB->lookup(toSlice(_hash));
```
gumb0: Better to replace this with a call to `exists`

libethereum/BlockChain.cpp
```diff
@@ -1375,18 +1379,26 @@ bool BlockChain::isKnown(h256 const& _hash, bool _isCurrent) const
...
 		}
 	DEV_READ_GUARDED(x_details)
 		if (!m_details.count(_hash))
 		{
-			string d;
-			m_extrasDB->Get(m_readOptions, toSlice(_hash, ExtraDetails), &d);
-			if (d.empty())
+			try
+			{
+				m_extrasDB->lookup(toSlice(_hash, ExtraDetails));
```
gumb0: Replace with a call to `exists`

libethereum/BlockChain.cpp
isaachier: Overall, `exists` followed by `lookup` effectively does 2 lookups, whereas `lookup` that might throw will only do one lookup.

isaachier: But I agree in this case seeing as the value is discarded.

cmake/EthOptions.cmake
```diff
@@ -11,14 +11,20 @@ macro(configure_project)
...
 	option(FATDB "Enable fat state database" ON)
-	option(ROCKSDB "Build with RocksDB instead of LevelDB" OFF)
 	option(PARANOID "Enable additional checks when validating transactions (deprecated)" OFF)
 	option(MINIUPNPC "Build with UPnP support" OFF)
 	option(FASTCTEST "Enable fast ctest" OFF)
 
+	if(WIN32 OR NOT HUNTER_ENABLED)
+		find_package(LevelDB REQUIRED)
+	else()
+	  hunter_add_package(leveldb)
```
chfast: We can have this change added, but not in this PR. Please remove it for now.

chfast: Also, format CMake with indentation of 4 spaces.

libdevcore/LevelDB.cpp
chfast: `clang-format` can't do it at the moment. But you should format the code with `clang-format` anyway.

About the placement of the `const` itself, I suggest we rather follow C++ Core Guidelines: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rl-const.

gumb0: Yeah, but again, I wouldn't like to have two styles mixed in the codebase

gumb0:
Merge from `develop` now makes difficult reformatting only the changes. I'm going to rebase instead and then reformat

gumb0:
Rebased and reformatted.

cmake/EthOptions.cmake
gumb0: @chfast I rolled back this change.

chfast:
GCC is complaining about some formatting...
Anyway, I think we can land it when fixed.

gumb0:
clang-format behaves a bit funny around our macros like `DEV_WRITE_GUARDED`
https://github.com/ethereum/cpp-ethereum/pull/4709/commits/c0a6f0ef96995c58c0ba0a85066d4645c41b8a80#diff-4bd28474d53dacc284d9432a1a785c10L1277

On the left side it indented `m_blocksDB->forEach(...)` to the level of previous line, although after macro expansion it's really one-line for-loop body. Then GCC complains about this indentation.

On the right side I worked around it with additional curly braces

gumb0:
Thanks a lot for your work @isaachier! And sorry it took so long.
Looking forward for your other changes with other DB support.

chfast:
So having the attention of both of you, what do you think about shipping with rocksdb by default?
The main advantage of this for me is the fact rocksdb has the Windows support in Hunter.
Also, parity seems to have tweaked it nicely to get good performance boost.

gumb0:
I'm ok with trying RocksDB and switching to it if it works fine.

(I don't know whether Parity's tweaks went upstream or whether they were deeper than in rust wrappers at all)

isaachier:
@gumb0 thanks for taking the time to review it!

I think RocksDB is widely considered an improvement over LevelDB. Idk all the issues parity had. One potential advantage of RocksDB inspired me to add an interface method here: `exists`. Not sure if implemented yet, but `exists` allows RocksDB to use a Bloom filter for efficient lookups without going to disk.

gumb0:
@isaachier Should we expect further PRs from you? (With other DBs)

isaachier:
I hope so but I'm not sure how soon. Have a number of other projects to work on right now.

PR #4710, opened 2017-12-18
Giving out snapshot through Warp protocol

gumb0:
1. Copy snapshot into data dir after importing it.
2. Support server-side of Warp protocol, so that cpp nodes could download snapshots from other cpp nodes.

Snapshot files won't ever change and the node will give out the same snapshot all the time.
Further improvements could support periodic generation of our own snapshots.

### DIFF
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -1025,7 +1025,7 @@ int main(int argc, char** argv)
 			SnapshotImporter importer(*stateImporter, *blockChainImporter);
 
 			auto snapshotStorage(createSnapshotStorage(filename));
-			importer.import(*snapshotStorage);
+			importer.import(*snapshotStorage, web3.ethereum()->blockChain().genesisHash());
 			// continue with regular sync from the snapshot block
 		}
 		catch (...)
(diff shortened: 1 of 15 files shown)


gumb0:
Applied clang-format in the separate commit

eth/main.cpp
```diff
@@ -1025,8 +1025,8 @@ int main(int argc, char** argv)
 			SnapshotImporter importer(*stateImporter, *blockChainImporter);
 
 			auto snapshotStorage(createSnapshotStorage(filename));
-			importer.import(*snapshotStorage);
-			// continue with regular sync from the snapshot block
+            importer.import(*snapshotStorage, web3.ethereum()->blockChain().genesisHash());
```
rgolovanov: Here should be tabs according to CS

libethereum/WarpPeerCapability.cpp
```diff
@@ -0,0 +1,138 @@
...
+		case WarpStatusPacket:
+		{
+			if (_r.itemCount() < 7)
+				BOOST_THROW_EXCEPTION(InvalidWarpStatusPacket());
+
+			auto const protocolVersion = _r[0].toInt<unsigned>(RLP::VeryStrict);
+			auto const networkId = _r[1].toInt<u256>(RLP::VeryStrict);
+			auto const totalDifficulty = _r[2].toInt<u256>(RLP::VeryStrict);
+			auto const latestHash = _r[3].toHash<h256>(RLP::VeryStrict);
+			auto const genesisHash = _r[4].toHash<h256>(RLP::VeryStrict);
```
pirapira: Is it a good idea to check `genesisHash` against `_chainGenesisHash` given to `init()`?  (Likewise for `networkId`?)

gumb0: Yes, I do that in the snapshot downloading PRs.

In this PR it doesn't make much difference, because it's a server part, and downloading peer checks it anyway.
I will add it later, anyway

pirapira: OK.

PR #4713, opened 2017-12-19
Add .editorconfig file

chfast:


.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.{cpp,c,hpp,h}]
+indent_style = tab
```
gumb0: should be space now?

chfast: We rather want GitHub to display tabs as 4 spaces, so I'm not sure it will work with `indent_style = spaces`.

gumb0: ok, let's try

PR #4716, opened 2017-12-21
convert TestHelper.cpp with clang

winsvega:
this is what I got by executing clang-format 
```
YAML:36:23: error: unknown key 'SplitEmptyFunction'
  SplitEmptyFunction: false

```

### DIFF
--- a/test/tools/libtesteth/TestHelper.cpp
+++ b/test/tools/libtesteth/TestHelper.cpp
@@ -1,26 +1,26 @@
 /*
-	This file is part of cpp-ethereum.
+        This file is part of cpp-ethereum.
 
-	cpp-ethereum is free software: you can redistribute it and/or modify
-	it under the terms of the GNU General Public License as published by
-	the Free Software Foundation, either version 3 of the License, or
-	(at your option) any later version.
+        cpp-ethereum is free software: you can redistribute it and/or modify
+        it under the terms of the GNU General Public License as published by
+        the Free Software Foundation, either version 3 of the License, or
+        (at your option) any later version.
 
-	cpp-ethereum is distributed in the hope that it will be useful,
-	but WITHOUT ANY WARRANTY; without even the implied warranty of
-	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


gumb0:
The result doesn't look right at all.
What's your clang-format version?

chfast:
@winsvega You need to upgrade clang-format, because the version you have does not understand the config. If you tell me the version I can try to make it compatible (it is always backward compatible). But from the error I can tell you rather have to upgrade, because `SplitEmptyFunction` is a setting we really want to have.

winsvega:
updated with clang-format 5

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -67,531 +66,540 @@ void connectClients(Client& c1, Client& c2)
...
-			using fsIterator = fs::directory_iterator;
-			for (fsIterator it(_dirPath); it != fsIterator(); ++it)
-			{
-				if (fs::is_regular_file(it->path()) && it->path().extension() == ext)
-					files.push_back(it->path());
-			}
-		}
-	}
-	return files;
+    for (auto& account : _o.count("alloc") ? _o["alloc"].get_obj() :
```
chfast: This looks bad. We have 3 `:` here. Can you move the "accounts" initialization outside of the loop?

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -67,531 +66,540 @@ void connectClients(Client& c1, Client& c2)
...
 #else
-	boost::filesystem::path path(boost::filesystem::temp_directory_path() / boost::filesystem::unique_path());
-	string cmd = string("lllc ") + path.string();
-	writeFile(path.string(), _code);
-	string result = executeCmd(cmd);
-	boost::filesystem::remove(path);
-	result = "0x" + result;
-	checkHexHasEvenLength(result);
-	return result;
+    boost::filesystem::path path(
```
chfast: Here you can save some space by `using fs =`.

winsvega:
I have rebased the PR. looks like everything is at it's place.
but a quick review would be nice

PR #4718, opened 2017-12-22
Enable Warp server only if we have snapshot

gumb0:
Previously it always created Warp capability and reported snapshot number 0, snapshot hash 0.

Also imported snapshot path construction was duplicated in `SnapshotImporter` and `WarpHostCapability`, now it's in a single place in `importedSnapshotPath()`


libethereum/WarpPeerCapability.cpp
```diff
@@ -33,29 +33,24 @@ void WarpPeerCapability::init(unsigned _hostProtocolVersion, u256 _hostNetworkId
     u256 _chainTotalDifficulty, h256 _chainCurrentHash, h256 _chainGenesisHash,
     std::shared_ptr<SnapshotStorageFace const> _snapshot)
 {
+    assert(_snapshot);
     m_snapshot = _snapshot;
```
chfast: `std::move` the shared pointer.

PR #4719, opened 2017-12-22
testeth throws an error if code is missing after parsing

winsvega:


test/tools/libtesteth/TestHelper.cpp
```diff
@@ -239,7 +239,12 @@ void replaceLLLinState(json_spirit::mObject& _o)
 		{
 			string code = obj["code"].get_str();
 			obj["code"] = compileLLL(code);
-		}
+                        if (code.size() > 0)
```
chfast: This is ok, but I think you have clang-format wrongly configured. Do you have version 5.0?

Can you try this command:

```
clang-format -i -lines=242:247 test/tools/libtesteth/TestHelper.cpp
```

winsvega: my version is 3.8.
how did you managed to install 5 ? 
I followed this https://askubuntu.com/questions/905205/installing-clang-5-0-and-using-c17
does not install clang-5

winsvega:
runned 
`clang-format -i -lines=242:247 test/tools/libtesteth/TestHelper.cpp`

with clang-format 5

PR #4723, opened 2017-12-27
Update boost shift test expected results

chfast:
The bug was fixed somewhere around boost 1.65.


chfast:
Ping?

chfast:
Boost is fixed, yes.

PR #4724, opened 2017-12-27
Refactor random test

chfast:
The contains a lot of code cleanups (using `override`, code formatting, etc), but the main goal to try to fix rndCode test on macOS.

Now the seed for BoostRandomCode is set once (it gets if from `test::Options` what is another dependency hiding anti-pattern). Not generating code should not reach for system clock for every random number.

### DIFF
--- a/test/tools/fuzzTesting/BoostRandomCode.cpp
+++ b/test/tools/fuzzTesting/BoostRandomCode.cpp
@@ -1,59 +1,71 @@
 /*
-	This file is part of cpp-ethereum.
+    This file is part of cpp-ethereum.
 
-	cpp-ethereum is free software: you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


winsvega:
I want to run some manual checks to make sure it does not break the functionality.

chfast:
I noticed that `test/testeth -- --createRandomTest` does not work at all, before and after changes.

winsvega:
hmm. the boost test is passing. why so?

chfast:
There is no error if you do not specify test filter.

Anyway, this works:
```
test/testeth -t GeneralStateTests -- --createRandomTest --seed 2
```
The result is always the same.

winsvega:
because you specify the same seed

Issue #4681, opened 2017-12-04
Unhanded exception in JSON-RPC


chfast:
2017-MM-DD (published patch date)
TALOS-2017-0471
CVE-2017-12119

CPP-Ethereum JSON-RPC Denial Of Service Vulnerabilities

### Summary
An exploitable unhandled exception vulnerability exists in multiple APIs of CPP-Ethereum's JSON-RPC.
Specially crafted JSON requests can cause a unhandled exception resulting in denial of service. 
An attacker can send malicious JSON to trigger this vulnerability. 

### Tested Versions

Ethereum commit 4e1015743b95821849d001618a7ce82c7c073768
  
### Product URLs

[http://cpp-ethereum.org](http://cpp-ethereum.org)

### CVSSv3 Score

7.5 - CVSS:3.0/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H

### CWE

CWE-248: Uncaught Exception

### Details

	CPP-Ethereum is a C++ ethereum client, one of the 3 most popular clients for the ethereum platform.
One of the components that is part of cpp-ethereum is a JSON-RPC server which exposes various APIs to manage 
client/node functionality. A lack of proper exception handling in the implementation of some APIs allows a 
remote attacker to send malformed JSON requests and crash the client/node. The following list of APIs are vulnerable:

	List of APIs available by default when JSON-RPC server is turned on:
		- debug_storageRangeAt
		- debug_traceBlockByNumber

	List of APIs available when the "--admin-via-http" switch is used:
		- miner_start
		- admin_eth_vmTrace
		- personal_newAccount
		- admin_eth_getReceiptByHashAndIndex
		- admin_setVerbosity
		- admin_verbosity
		
To handle JSON objects, the `JsonCpp` project (https://github.com/open-source-parsers/jsoncpp) is used and the definition of the `asInt` method looks as follows:

	json_value.cpp

	Line 732	Value::Int Value::asInt() const {
	Line 733	  switch (type_) {
	Line 734	  case intValue:
	Line 735		JSON_ASSERT_MESSAGE(isInt(), "LargestInt out of Int range");
	Line 736		return Int(value_.int_);
	Line 737	  case uintValue:
	Line 738		JSON_ASSERT_MESSAGE(isInt(), "LargestUInt out of Int range");
	Line 739		return Int(value_.uint_);
	Line 740	  case realValue:
[...]

PR #4725, opened 2017-12-27
Catch exceptions from parsing parameters in JSON RPC stubs

gumb0:
Resolves https://github.com/ethereum/cpp-ethereum/issues/4681


gumb0:
Response returned from the requests described in the issue looks now like
```
{"error":{"code":-32602,"data":null,"message":"INVALID_PARAMS: Invalid method parameters (invalid name and/or type) recognised: LargestUInt out of Int range"},"id":1,"jsonrpc":"2.0"}
```

chfast:
Ok, but does not compile on anything except Windows.

gumb0:
fixed and reformatted

PR #4733, opened 2018-01-02
Update IpcServerBase.cpp

mesin:
Related issues:

Error installing cpp-ethereum on OSX https://github.com/ethereum/homebrew-ethereum/issues/126
brew install cpp-ethereum error https://github.com/ethereum/homebrew-ethereum/issues/121
brew install cpp-ethereum CMake Error https://github.com/ethereum/homebrew-ethereum/issues/124


chfast:
Sorry for late response, but this is wrong fix. I will check CMake config later on.

PR #4744, opened 2018-01-08
evmJit tests

winsvega:
add evmjit state tests


pirapira:
Why are they called evmjit tests?  They can be run only with jit?

PR #4747, opened 2018-01-09
fix vm cli option

jwasinger:
### Issue
Can't select vm type:
```
$ eth/eth --vm jit

Invalid argument: jit
```

Solved by this PR

### DIFF
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -349,7 +349,7 @@ int main(int argc, char** argv)
 	generalOptions.add_options()
 		("db-path,d", po::value<string>()->value_name("<path>"), ("Load database from path\n(default: " + getDataDir().string() + ").\n").c_str())
 #if ETH_EVMJIT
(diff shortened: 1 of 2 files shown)


eth/main.cpp
```diff
@@ -349,7 +349,7 @@ int main(int argc, char** argv)
 	generalOptions.add_options()
 		("db-path,d", po::value<string>()->value_name("<path>"), ("Load database from path\n(default: " + getDataDir().string() + ").\n").c_str())
 #if ETH_EVMJIT
-		("vm", "<vm-kind>  Select VM; options are: interpreter, jit or smart (default: interpreter)")
+		("vm", po::value<string>()->value_name("<n>"), "<vm-kind>  Select VM; options are: interpreter, jit or smart (default: interpreter)")
```
axic: I think it should be `value_name("<vm-kind>")` and the `<vm-kind>` text be removed from the help text.

PR #4751, opened 2018-01-10
Circleci debug macos

chfast:


codecov.yml
```diff
@@ -26,7 +26,7 @@ coverage:
       default:
         enabled: no
 comment:
-  layout: "reach, diff, flags, files, footer"
+  layout: "reach, diff, flags"
```
axic: This codecov change seems unrelated? Perhaps you want to open a new PR.

chfast:
I think that's not for merge. I wanted to debug weird failure on macOS.

codecov.yml
chfast: This is done.

PR #4753, opened 2018-01-10
Move VM options to evm library

chfast:
I'm kinda proud of this idea. The VM program options definition was moved to libevm, easy accessible by eth and ethvm. 

It also includes custom verifier for VMKind option. This part might not be perfect. boost::program_options does not handle sets of values by default. I have to list the available options manually...

### DIFF
--- a/libevm/VMFactory.cpp
+++ b/libevm/VMFactory.cpp
@@ -79,7 +79,7 @@ void validate(boost::any& v, const std::vector<std::string>& values, VMKind* /* 
     throw po::validation_error(po::validation_error::invalid_option_value);
 }
 
-po::options_description getVMOptions(unsigned _lineLength)
+po::options_description vmProgramOptions(unsigned _lineLength)
 {
     // It must be a static object because boost expects const char*.
     static const std::string description = [] {
--- a/libevm/VMFactory.cpp
+++ b/libevm/VMFactory.cpp
@@ -24,43 +24,102 @@
 #include "SmartVM.h"
 #endif
 
+namespace po = boost::program_options;
+
 namespace dev
 {
 namespace eth
 {
 namespace
 {
-	auto g_kind = VMKind::Interpreter;
+auto g_kind = VMKind::Interpreter;
+
+/// A helper type to build the tabled of VM implementations.
+///
+/// More readable than std::tuple.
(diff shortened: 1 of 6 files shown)


libevm/VMFactory.cpp
```diff
@@ -24,43 +24,102 @@
...
 
+po::options_description getVMOptions(unsigned _lineLength)
+{
+    po::options_description opts("VM Options", _lineLength);
+    opts.add_options()("vm",
+        po::value<VMKind>()
+            ->value_name("<name>")
+            ->default_value(VMKind::Interpreter, "interpreter")
+            ->notifier(VMFactory::setKind),
+        "Select VM implementation; options are: interpreter, jit or smart");
```
axic: The `jit or smart` shouldn't only be here in `ETH_EVMJIT`?

chfast: Right. Fixing...

libevm/VMFactory.cpp
```diff
@@ -24,43 +24,115 @@
...
+VMKindTableEntry vmKindsTable[] = {
+    {VMKind::Interpreter, "interpreter"},
+#if ETH_EVMJIT
+    {VMKind::JIT, "jit"},
+    {VMKind::Smart, "smart"},
+#endif
+};
+}
+
+void validate(boost::any& v, const std::vector<std::string>& values, VMKind* /* target_type */, int)
```
gumb0: Where is this used?
And where is now getting `VMKind` from option string value?

libevm/VMFactory.h
```diff
@@ -30,6 +31,13 @@ enum class VMKind
 	Smart
 };
 
+/// Provide a set of program options related to VMs.
+///
+/// @param _lineLength  The line length for description text wrapping, the same as in
+///                     boost::program_options::options_description::options_description().
+boost::program_options::options_description getVMOptions(
```
gumb0: We don't use `get` prefix, I'd name it `vmProgramOptions()`

libevm/VMFactory.cpp
chfast: This overload is required by `po::value<VMKind>`. This function also coverts the string to `VMKind`.

PR #4735, opened 2018-01-02
replace LLL in transacton data

winsvega:
replace LLL code in transaction data field

PR #4754, opened 2018-01-11
Support compiling wast to wasm in testeth

axic:
Depends on #4779.

### DIFF
--- a/test/tools/libtesteth/TestHelper.cpp
+++ b/test/tools/libtesteth/TestHelper.cpp
@@ -21,6 +21,7 @@
 #include "TestHelper.h"
 #include "Options.h"
 #include "TestOutputHelper.h"
+#include "wast2wasm.h"
 
 #include <BuildInfo.h>
 #include <libethashseal/EthashCPUMiner.h>
@@ -263,6 +264,8 @@ string replaceCode(string const& _code)
         checkHexHasEvenLength(_code);
         return _code;
     }
+    if (_code.find("(module") == 0)
+        return wast2wasm(_code);
 
     string compiledCode = compileLLL(_code);
     if (_code.size() > 0)
--- a/test/tools/libtesteth/wast2wasm.cpp
+++ b/test/tools/libtesteth/wast2wasm.cpp
@@ -4,9 +4,13 @@
 
 #include <boost/test/unit_test.hpp>
 
+// Disable MSVC warning "unary minus operator applied to unsigned type".
+#pragma warning(push)
+#pragma warning(disable: 4146)
 #include <wasm-binary.h>
 #include <wasm-s-parser.h>
 #include <wasm-validator.h>
+#pragma warning(pop)
 
 using namespace std;
 
--- a/test/tools/libtesteth/wast2wasm.cpp
+++ b/test/tools/libtesteth/wast2wasm.cpp
@@ -1,0 +1,48 @@
+#include "wast2wasm.h"
+
+#include <libdevcore/CommonData.h>
+
+#include <boost/test/unit_test.hpp>
+
+#include <wasm-binary.h>
+#include <wasm-s-parser.h>
+#include <wasm-validator.h>
+
+using namespace std;
+
+namespace dev
+{
+
+namespace test
+{
+
+/* from https://github.com/ewasm/evm2wasm */
+string wast2wasm(string const& input, bool debug)
+{
+    wasm::Module module;
+
+    try {
+        wasm::SExpressionParser parser(const_cast<char*>(input.c_str()));
+        wasm::Element& root = *parser.root;
+        wasm::SExpressionWasmBuilder builder(module, *root[0]);
+    } catch (wasm::ParseException& p) {
+        ostringstream err;
+        p.dump(err);
+        BOOST_ERROR("Error parsing wasm: " + err.str());
+    }
+
+    BOOST_REQUIRE_MESSAGE(wasm::WasmValidator().validate(module), "Compiled WASM module is not valid.");
+
+    wasm::BufferWithRandomAccess buffer(debug);
+    wasm::WasmBinaryWriter writer(&module, buffer, debug);
+    writer.write();
+
+    ostringstream output;
+    buffer.writeTo(output);
+
+    return toHexPrefixed(output.str());
+}
+
+}
+
+}
--- a/test/tools/libtesteth/wast2wasm.cpp
+++ b/test/tools/libtesteth/wast2wasm.cpp
@@ -26,7 +26,10 @@ string wast2wasm(string const& input, bool debug)
     wasm::Module module;
 
     try {
-        wasm::SExpressionParser parser(const_cast<char*>(input.c_str()));
+        // FIXME: binaryen 1.37.28 actually modifies the input...
+        //        as a workaround make a copy here
+        string tmp = input;
+        wasm::SExpressionParser parser(const_cast<char*>(tmp.c_str()));
         wasm::Element& root = *parser.root;
         wasm::SExpressionWasmBuilder builder(module, *root[0]);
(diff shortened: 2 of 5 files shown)


axic:
@chfast this need binaryen support in cpp-eth :)

cc @jwasinger @hugo-dc @holgerd77

@winsvega this should not be merged soon, it is only used by ewasm testing, can be merged once all of that is stable

axic:
Should follow the instructions from here: https://gitter.im/ethereum/tests?at=5a577e71d0514c785b1a2384

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -382,6 +382,35 @@ string compileLLL(string const& _code)
...
+  wasm::WasmBinaryWriter writer(&wasm, buffer, debug);
+  writer.write();
+
+  ostringstream output;
+  buffer.writeTo(output);
+
+  return bytes(output.begin(), output.end());
+}
+
 void checkHexHasEvenLength(string const& _str)
```
winsvega: do not change importCode function. 
the changes must be done to replaceLLL function 
but after this PR got merged please https://github.com/ethereum/cpp-ethereum/pull/4735/files

just rename replaceLLL function to compileSrcCode or something
and parse wasm and LLL from there

winsvega:
is it because of a compiler version I use. or something is wrong with the dependency source files?

axic:
Will rebase this on top off #4779 and I'll need to only change `replaceBytecode` to support wasm.

axic:
@chfast this also needs to work when hera is turned off

test/tools/libtesteth/TestHelper.cpp
axic: Has been updated now.

axic:
@winsvega do you have more context for this error?

```
/home/wins/Ethereum/cpp-ethereum/build/deps/src/binaryen/src/literal.h:158: error: specialization of ‘template<class _Tp> struct std::hash’ in different namespace [-fpermissive]
 template<> struct hash<wasm::Literal> {
                   ^
```

Like the preceding/following 40 lines.

winsvega:
I think its the first one. because of this the following errors are due to the missing defenition.

axic:
@winsvega without more context it is really hard to figure this out. Can you upload logs somewhere?

winsvega:
selection_001

axic:
@winsvega fixed all the build issues - please retry

winsvega:
looks like it works now, but
```
/home/wins/Ethereum/cpp-ethereum/test/tools/libtesteth/TestHelper.cpp:268: error: undefined reference to 'dev::test::wast2wasm(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool)'
collect2: error: ld returned 1 exit status
test/CMakeFiles/testeth.dir/build.make:1798: recipe for target 'test/testeth' failed
make[2]: *** [test/testeth] Error 1
CMakeFiles/Makefile2:1269: recipe for target 'test/CMakeFiles/testeth.dir/all' failed
make[1]: *** [test/CMakeFiles/testeth.dir/all] Error 2
Makefile:140: recipe for target 'all' failed
make: *** [all] Error 2
```

what was the issue with binaryen?

axic:
@winsvega you should rerun cmake to regenerate dependency lists.

winsvega:
that didn't help

axic:
@winsvega fixed! Please try with the latest version :)

winsvega:
worked. 
how did you fixed the -fpermissive issue?

axic:
Moved it into its own file so it won't collide with the include mess we have.

winsvega:
so its like we have our own version of this dependency now?

axic:
There is no new build dependency (binaryen is included via hera). What is the mac build issue?

winsvega:
```
-- Build files have been written to: /Users/distiller/build/deps/src/binaryen-build
ninja: error: mkdir(/lib): Permission denied
ninja: build stopped: .
Exited with code 1
```

winsvega:
could you comment what was the issue with binaryen ?
https://github.com/WebAssembly/binaryen/issues/1362#issuecomment-358098858

axic:
I have no idea. Moved it to a file on its own where the other 40+ testeth headers are not included. Anything down the dependency chain could have caused it.

winsvega:
windows build fails
```
wast2wasm.cpp
c:\projects\cpp-ethereum\build\deps\src\binaryen\src\parsing.h(175): error C2220: warning treated as error - no 'object' file generated (compiling source file C:\projects\cpp-ethereum\test\tools\libtesteth\wast2wasm.cpp) [C:\projects\cpp-ethereum\build\test\testeth.vcxproj]
c:\projects\cpp-ethereum\build\deps\src\binaryen\src\parsing.h(175): warning C4146: unary minus operator applied to unsigned type, result still unsigned (compiling source file C:\projects\cpp-ethereum\test\tools\libtesteth\wast2wasm.cpp) [C:\projects\cpp-ethereum\build\test\testeth.vcxproj]
```

axic:
@chfast do not forget to apply the same MSVC change to hera (and evm2wasm).

winsvega:
also need to check that wasm test fillers are compiled correctly

axic:
@winsvega can you test it? Would be nice merging this.

winsvega:
yaml compiler works but the code starts with 0x00 byte. which means stop execution.

axic:
> yaml compiler works but the code starts with 0x00 byte. which means stop execution.

It starts with a wasm magic (`\x00asm`).

winsvega:
yaml tests from test repo fails on this bytecode. looks like it does not recognize wasm bytecode.

axic:
> yaml tests from test repo fails on this. looks like it does not recognize wasm bytecode.

I am not sure what you mean. What test case are you compiling with this? The ones from @jwasinger ? In any case, this patch should have zero effect on the current EVM test suite (is this the case?) and it will only do conversion on eWASM test cases and the output of those cannot be interpreted by EVM (hence we'll not merge the eWASM test suite into the main tests repo anytime soon).

winsvega:
`the output of those cannot be interpreted by EVM`
ok.

test/tools/libtesteth/wast2wasm.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+/* from https://github.com/ewasm/evm2wasm */
+string wast2wasm(string const& input, bool debug)
+{
+    wasm::Module module;
+
+    try {
+        // FIXME: binaryen 1.37.28 actually modifies the input...
+        //        as a workaround make a copy here
+        string tmp = input;
```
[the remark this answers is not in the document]
chfast: I confirm `tmp` is a copy of `input`.

test/tools/libtesteth/wast2wasm.cpp
```diff
@@ -0,0 +1,55 @@
...
+        wasm::SExpressionParser parser(const_cast<char*>(tmp.c_str()));
```
chfast: You don't need the cast now.

axic: Updated.

axic: A cast is still needed as `c_str()` returns `const char*` and `SExpressionParser` wants `char*`. Do you suggest any other cast / solution?

axic:
I was trigger happy - merging this now.

chfast:
You probably could have used `.data()` instead of `.c_str()`.

axic:
For `.data()` the spec says it is still `const char*`, but it is not guaranteed to be zero terminated. That binaryen class needs zero terminated c-strings.

axic:
@chfast can you please push the MSVC workaround to evm2wasm and hera?

PR #4755, opened 2018-01-11
Generalize the EVM-C VM wrapper

chfast:


axic:
Looks good to me with my limited C++.

libevm/EVMC.h
```diff
@@ -0,0 +1,33 @@
...
+namespace dev
+{
+namespace eth
+{
+class EVM;
+
+/// Translate the EVMSchedule to EVM-C revision.
+evm_revision toRevision(EVMSchedule const& _schedule);
+
+template<evm_create_fn createFn>
```
gumb0: So factory function has always the same signature, what's exactly the purpose of templatizing `EVMC` ?

You could just have it a simple subclass of `VMFace` and have various
```
std::uniqie_ptr<EVMC> createEvmJit();
std::uniqie_ptr<EVMC> createEvmWasm();
...
```
(then you'd need `#include <evmjit.h>` only in implementation, not in header)
or
```
std::uniqie_ptr<EVMC> createEvmC(evm_create_fn createFn);
```

chfast: I wanted to do it without a template, but we have an issue with singletons at the moment. The template is a temporary workaround because it has `instance<hera>()` and `instance<evmjit>()`.

We have to rethink the lifetime of an VM instance. cpp-ethereum creates a new VM instance for every call.

chfast:
I will try different approach to this, probably as a new PR...

libevm/EVMC.h
```diff
@@ -0,0 +1,33 @@
...
+class EVMC : public VMFace
+{
+public:
+    owning_bytes_ref exec(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp) final;
+
+    bool isCodeReady(evm_revision _mode, uint32_t _flags, h256 _codeHash);
+    void compile(evm_revision _mode, uint32_t _flags, bytesConstRef _code, h256 _codeHash);
+
+private:
+    static EVM& instance();
```
gumb0: This shouldn't be static then

gumb0: So it would be something like
```
std::unique_ptr<EVMC> createEvmJit()
{
  return return std::unique_ptr<VMFace>(new EVMC{evmjit_create});
}

EVMC(evm_create_fn _createFn) : m_createFn(_createFn) {}

EVM& EVMC::instance()
{
  static EVM evm{m_createFn};
  return evm;
}
```
It's not a lot of changes and will be simpler I think

axic:
@chfast please add support for `--vm jit` in `testeth`

libevm/EVMC.h
chfast: Why not?

gumb0: I mean if `EVMC` would be not a template, that different instances of `EVMC` would need different `EVM` instance

gumb0: And if we want to create only once singleton-style, then we could make `EVMC` itself a singleton I guess

gumb0: Wait no, my example below won't work. How about something like this.
```
std::unique_ptr<EVMC> createEvmJit()
{
  static EVM evm{evmjit_create};
  return std::unique_ptr<VMFace>(new EVMC(evm));
}

EVMC(EVM& _evm) : m_instance(_evm) {}

EVM& EVMC::instance()
{
  return m_instance;
}
```

chfast:
No singletons in: #4761.

PR #4761, opened 2018-01-12
Generalize the EVM-C VM wrapper v2

chfast:


libevm/VMFactory.cpp
```diff
@@ -125,7 +125,7 @@ std::unique_ptr<VMFace> VMFactory::create(VMKind _kind)
     case VMKind::Interpreter:
         return std::unique_ptr<VMFace>(new VM);
     case VMKind::JIT:
-        return std::unique_ptr<VMFace>(new JitVM);
+        return std::unique_ptr<VMFace>(new JitVM{evmjit_create()});
     case VMKind::Smart:
         return std::unique_ptr<VMFace>(new SmartVM);
```
axic: Don't you want to pass `evmjit_create()` to `SmartVM` too?

chfast: No because SmartVM only works with EVMJIT, so it will know what to use.

axic: Why not call it `SmartJITVM` in that case?

libevm/EVMC.h
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+private:
+    /// The VM instance created with EVM-C <prefix>_create() function.
+    evm_instance* m_instance = nullptr;
+};
+
+
+/// The wrapper implementing the VMFace interface with a EVM-C VM as a backend.
+class EVMC : public EVM, public VMFace
```
gumb0: I think you could privately inherit `EVM` ?

chfast: Yes, I did it at first and I made only `isCodeReady()` and `compile()` public. But that's 2 out of 3 methods... so I don't see a big pro for it.

PR #4766, opened 2018-01-14
Hera+

chfast:


libevm/VMFactory.cpp
```diff
@@ -54,6 +55,7 @@ VMKindTableEntry vmKindsTable[] = {
 #endif
 #if ETH_HERA
     {VMKind::Hera, "hera"},
+    {VMKind::HeraPlus, "hera+"},
```
axic: Should this be `heraplus` on the command instead?

axic:
Probably the most generic version would be "ChainedVM" which has a list of `VMKind`s which are executed in order (until one responds with "non-UNSUPPORTED-BYTCODE")

axic:
@chfast you didnt update the evmjit repo to the right version, it doesn't have the kind value

libevm/VMFactory.cpp
chfast: Can be  :)

PR #4771, opened 2018-01-15
Add --evmc program option

chfast:
This option expects value like this `--evmc name=value`. The option is passed then to an EVM-C instance.

### DIFF
--- a/libevm/VMFactory.cpp
+++ b/libevm/VMFactory.cpp
@@ -82,6 +82,33 @@ void validate(boost::any& v, const std::vector<std::string>& values, VMKind* /* 
     throw po::validation_error(po::validation_error::invalid_option_value);
 }
 
+namespace
+{
+/// The name of the program option --evmc. The boost will trim the tailing
+/// space and we can reuse this variable in exception message.
+const char c_evmcPrefix[] = "evmc ";
+
+/// The list of EVM-C options stored as pairs of (name, value).
+std::vector<std::pair<std::string, std::string>> s_evmcOptions;
+
+/// The additional parser for EVM-C options. The options should look like
+/// `--evmc name=value` or `--evmc=name=value`. The boost pass the strings
+/// of `name=value` here. This function splits the name and value or reports
+/// the syntax error if the `=` character is missing.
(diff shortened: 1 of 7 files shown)


test/tools/libtesteth/Options.cpp
```diff
@@ -72,8 +73,29 @@ void printVersion()
...
+
+        // For some reason boost is confused by -- separator. This extra parser "skips" the --.
+        auto skipDoubleDash = [](const std::string& s) -> std::pair<std::string, std::string> {
+            if (s == "--")
+                return {"--", {}};
+            return {};
+        };
+
+        auto vmOpts = vmProgramOptions();
+        po::parsed_options parsed = po::command_line_parser(argc, argv)
```
winsvega: how do I get back parsed vm option from Options::get() ?

libevm/VMFactory.cpp
```diff
@@ -82,6 +82,38 @@ void validate(boost::any& v, const std::vector<std::string>& values, VMKind* /*
...
+/// space and we can reuse this variable in exception message.
+const char c_evmcPrefix[] = "evmc ";
+
+/// The list of EVM-C options stored as pairs of (name, value).
+std::vector<std::pair<std::string, std::string>> s_evmcOptions;
+
+/// The additional parser for EVM-C options. The options should look like
+/// `--evmc name=value` or `--evmc=name=value`. The boost pass the strings
+/// of `name=value` here. This function splits the name and value or reports
+/// the syntax error if the `=` character is missing.
```
axic: How to pass multiple envs? `--evmc debug=true --evmc tracing=true` ?

chfast: Yes.

test/tools/libtesteth/Options.cpp
chfast: You don't. This is only to handle the VM options in single place (VMFactory.cpp). I don't care about the rest of the options. There was a PR introducing program_options in testeth, why is it not merged yet?

winsvega: I believe it was an issue

PR #4772, opened 2018-01-15
Add eWASM schedule

chfast:
It's Byzantium + unlimited code size


axic:
If we could I would also remove the ~~dumb~~ weird idea of not charging for 0 bytes in contract code in `eWASMSchedule` :)

axic:
So apparently it can be done:
```
 	unsigned txDataZeroGas = 4;
 	unsigned txDataNonZeroGas = 68;
```

I'd suggest going with `txDataZeroGas = txDataNonZeroGas` in `EWASMSchedule`.

libethcore/EVMSchedule.h
```diff
@@ -116,6 +116,13 @@ static const EVMSchedule ByzantiumSchedule = []
 	return schedule;
 }();
 
+static const EVMSchedule EWASMSchedule = []
+{
+    EVMSchedule schedule = ByzantiumSchedule;
+    schedule.maxCodeSize = std::numeric_limits<unsigned>::max();
```
winsvega: which EIP is this?

chfast: The Lisbon spring EIP.

libethereum/ChainParams.cpp
```diff
@@ -70,6 +70,7 @@ string const c_daoHardforkBlock = "daoHardforkBlock";
 string const c_EIP150ForkBlock = "EIP150ForkBlock";
 string const c_EIP158ForkBlock = "EIP158ForkBlock";
 string const c_byzantiumForkBlock = "byzantiumForkBlock";
+string const c_eWASMForkBlock = "eWASMForkBlock";
```
winsvega: with another fork I think it would be reasonable to make markers in tests. like 
execute this test on  Frontier, Homestead  but not Metropolis. 
and so on. this way we will reduce the amount of tests.

libethcore/EVMSchedule.h
winsvega: link?

libethereum/ChainParams.cpp
chfast: I don't fully understand, but some improvements in configuration would be nice. I have some ideas now. We should also discuss this with geth team. They should have more experience.

Also, we can use YAML now for better readability.

chfast:
Looks like every test fails now :D Nice.

winsvega:
what is the default forkblock number value?

chfast:
I have not set any, so probably 0. Where do set the default value?

winsvega:
the value is defined in genesis config files. 
https://github.com/ethereum/cpp-ethereum/tree/develop/libethashseal/genesis

axic:
Just copy constanintople:
```
		"constantinopleForkBlock": "0xffffffffffffffffff",
```

chfast:
Pushed one more commit. It sets default value for fork block numbers in infinity (instead of 0), so we don't have to put these magic `0xffffff...` for every new fork in every config file.

test/unittests/libethashseal/EthashTest.cpp
```diff
@@ -48,12 +52,13 @@ BOOST_AUTO_TEST_CASE(calculateDifficultyByzantiumWithoutUncles)
 	header.setNumber(0x2001);
 	header.setTimestamp(130);
 
-	BOOST_REQUIRE_EQUAL(ethash.calculateDifficulty(header, parentHeader), 999024);
+	BOOST_CHECK_EQUAL(ethash.calculateDifficulty(header, parentHeader), 999024);
```
gumb0: Why did you change this?

chfast: Just pedantic mode. I prefer weaker check if available.

PR #4777, opened 2018-01-16
makeBlockchainTestFromStateTest

winsvega:
move dynamic blockchain test generation to a separate func

### DIFF
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -56,6 +56,114 @@ ImportTest::ImportTest(json_spirit::mObject const& _input, json_spirit::mObject&
 	importState(_input.at("pre").get_obj(), m_statePre);
 }
 
+void ImportTest::makeBlockchainTestFromStateTest(vector<eth::Network> const& _networks) const
+{
+    // Generate blockchain test filler
+    string testnameOrig = TestOutputHelper::get().testName();
+    for (auto& tr : m_transactions)
+    {
+        json_spirit::mObject json;
+        json_spirit::mObject testObj;
+
+        // generate the test name with transaction detail
+        string postfix = "_d" + toString(tr.dataInd);
+        postfix += "g" + toString(tr.gasInd);
+        postfix += "v" + toString(tr.valInd);
+        string testname = testnameOrig + postfix;
+
+        // basic genesis
+        json_spirit::mObject genesisObj = TestBlockChain::defaultGenesisBlockJson();
+        genesisObj["coinbase"] = toString(m_envInfo->author());
+        genesisObj["gasLimit"] = toCompactHexPrefixed(m_envInfo->gasLimit());
+        genesisObj["timestamp"] = toCompactHexPrefixed(m_envInfo->timestamp() - 50);
+        testObj["genesisBlockHeader"] = genesisObj;
+        testObj["pre"] = fillJsonWithState(m_statePre);
+
+        // generate expect sections for this transaction
+        if (m_testInputObject.count("expect"))
+        {
+            State s = State(0, OverlayDB(), eth::BaseState::Empty);
+            AccountMaskMap m = std::unordered_map<Address, AccountMask>();
+            StateAndMap smap{s, m};
+            vector<size_t> stateIndexesToPrint;  // not used
+            json_spirit::mArray expetSectionArray;
+
+            for (auto const& net : _networks)
+            {
+                // Calculate the block reward
+                ChainParams const chainParams{genesisInfo(net)};
+                EVMSchedule const schedule = chainParams.scheduleForBlockNumber(1);
+                u256 const blockReward = chainParams.blockReward(schedule);
+
+                TrExpectSection search{tr, smap};
+                for (auto const& exp : m_testInputObject.at("expect").get_array())
+                {
+                    TrExpectSection* search2 = &search;
+                    checkGeneralTestSectionSearch(exp.get_obj(), stateIndexesToPrint, "", search2);
+                    if (search.second.first.addresses().size() !=
+                        0)  // if match in the expect sections for this tr found
+                    {
+                        // replace expected mining reward (in state tests it is 0)
+                        json_spirit::mObject obj =
+                            fillJsonWithState(search2->second.first, search2->second.second);
+                        for (auto& adr : obj)
+                        {
+                            if (adr.first == toHexPrefixed(m_envInfo->author()))
+                            {
+                                if (adr.second.get_obj().count("balance"))
+                                {
+                                    u256 expectCoinbaseBalance =
+                                        toInt(adr.second.get_obj()["balance"]);
+                                    expectCoinbaseBalance += blockReward;
+                                    adr.second.get_obj()["balance"] =
+                                        toCompactHexPrefixed(expectCoinbaseBalance);
+                                }
+                            }
+                        }
+
+                        json_spirit::mObject expetSectionObj;
+                        expetSectionObj["network"] = test::netIdToString(net);
+                        expetSectionObj["result"] = obj;
+                        expetSectionArray.push_back(expetSectionObj);
+                        break;
+                    }
+                }  // for exp
+            }      // for net
+
+            testObj["expect"] = expetSectionArray;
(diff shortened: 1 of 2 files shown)


test/tools/libtesteth/ImportTest.cpp
```diff
@@ -89,113 +197,12 @@ bytes ImportTest::executeTest()
...
-			genSuite.executeTest(TestOutputHelper::get().caseName(), testFillerFile);
-
-		} //transactions
-	}//fillchain
-
-	m_transactions.clear();
-	m_transactions = transactionResults;
-	return bytes();
+    m_transactions.clear();
+    m_transactions = transactionResults;  // update transactions with execution results.
```
chfast: You don't need `.clear()` before  assignment.

test/tools/libtesteth/ImportTest.h
```diff
@@ -84,8 +84,9 @@ class ImportTest
 	using StateAndMap = std::pair<eth::State, eth::AccountMaskMap>;
 	using TrExpectSection = std::pair<transactionToExecute, StateAndMap>;
 	bool checkGeneralTestSectionSearch(json_spirit::mObject const& _expects, std::vector<size_t>& _errorTransactions, std::string const& _network = "", TrExpectSection* _search = NULL) const;
+    void makeBlockchainTestFromStateTest(std::vector<eth::Network> const& _networks) const;
```
pirapira: Can you change the signature so that the input goes as an argument and the output comes as a return value?

Or, will you document where the input is taken from, and where the output goes?

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -56,6 +56,114 @@ ImportTest::ImportTest(json_spirit::mObject const& _input, json_spirit::mObject&
...
+    for (auto& tr : m_transactions)
+    {
+        json_spirit::mObject json;
+        json_spirit::mObject testObj;
+
+        // generate the test name with transaction detail
+        string postfix = "_d" + toString(tr.dataInd);
+        postfix += "g" + toString(tr.gasInd);
+        postfix += "v" + toString(tr.valInd);
+        string testname = testnameOrig + postfix;
```
pirapira: Can this be `const`?

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -56,6 +56,114 @@ ImportTest::ImportTest(json_spirit::mObject const& _input, json_spirit::mObject&
...
+        testObj["genesisBlockHeader"] = genesisObj;
+        testObj["pre"] = fillJsonWithState(m_statePre);
+
+        // generate expect sections for this transaction
+        if (m_testInputObject.count("expect"))
+        {
+            State s = State(0, OverlayDB(), eth::BaseState::Empty);
+            AccountMaskMap m = std::unordered_map<Address, AccountMask>();
+            StateAndMap smap{s, m};
+            vector<size_t> stateIndexesToPrint;  // not used
```
pirapira: If it's really not used, remove the variable.  If it's used somewhere, remove the comment.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -56,6 +56,114 @@ ImportTest::ImportTest(json_spirit::mObject const& _input, json_spirit::mObject&
...
+        genesisObj["gasLimit"] = toCompactHexPrefixed(m_envInfo->gasLimit());
+        genesisObj["timestamp"] = toCompactHexPrefixed(m_envInfo->timestamp() - 50);
+        testObj["genesisBlockHeader"] = genesisObj;
+        testObj["pre"] = fillJsonWithState(m_statePre);
+
+        // generate expect sections for this transaction
+        if (m_testInputObject.count("expect"))
+        {
+            State s = State(0, OverlayDB(), eth::BaseState::Empty);
+            AccountMaskMap m = std::unordered_map<Address, AccountMask>();
```
pirapira: Are you sure you need ` = ...;`?

I guess `AccountMaskMap m;` would initialize `m` to be empty.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -56,6 +56,114 @@ ImportTest::ImportTest(json_spirit::mObject const& _input, json_spirit::mObject&
 	importState(_input.at("pre").get_obj(), m_statePre);
 }
 
+void ImportTest::makeBlockchainTestFromStateTest(vector<eth::Network> const& _networks) const
+{
+    // Generate blockchain test filler
+    string testnameOrig = TestOutputHelper::get().testName();
+    for (auto& tr : m_transactions)
```
pirapira: This can be `auto const`.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -56,6 +56,114 @@ ImportTest::ImportTest(json_spirit::mObject const& _input, json_spirit::mObject&
...
+                        0)  // if match in the expect sections for this tr found
+                    {
+                        // replace expected mining reward (in state tests it is 0)
+                        json_spirit::mObject obj =
+                            fillJsonWithState(search2->second.first, search2->second.second);
+                        for (auto& adr : obj)
+                        {
+                            if (adr.first == toHexPrefixed(m_envInfo->author()))
+                            {
+                                if (adr.second.get_obj().count("balance"))
```
pirapira: Why not `if (A && B) {...}` instead of `if (A) { if (B) {...}}`?

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -56,6 +56,114 @@ ImportTest::ImportTest(json_spirit::mObject const& _input, json_spirit::mObject&
...
+                        json_spirit::mObject expetSectionObj;
+                        expetSectionObj["network"] = test::netIdToString(net);
+                        expetSectionObj["result"] = obj;
+                        expetSectionArray.push_back(expetSectionObj);
+                        break;
+                    }
+                }  // for exp
+            }      // for net
+
+            testObj["expect"] = expetSectionArray;
```
pirapira: This function is too big, so conisider something like
`testObj["expect"] = translateExpectSection(m_testInput.at("expect"));`

test/tools/libtesteth/ImportTest.h
winsvega: well I wanted it to be out of header. but it is too much linked with private fields

pirapira: Inside or outside of the header, a comment is needed about what this function does.

test/tools/libtesteth/ImportTest.cpp
winsvega: I don't want to split this function anymore. it does it's purpose - generating blockchain tests out of state tests.

test/tools/libtesteth/ImportTest.h
winsvega: what do you want? just a comment?

winsvega: this is a small PR to split a large function. I am planing to make a new future to testeth soon.

winsvega: if we spend another week to just split one function into two I wont be able to make new changes in a month.

test/tools/libtesteth/ImportTest.cpp
pirapira: We need a third-party opinion.

test/tools/libtesteth/ImportTest.h
pirapira: @winsvega yes, I want a comment.

You can say "I'll do xyz later".  I'll make a note.

winsvega: I guess after Pwaels review ther would have to be done like lots of changes. here.

test/tools/libtesteth/ImportTest.cpp
pirapira: OK, I'd merge any PR that's an improvement.

test/tools/libtesteth/ImportTest.h
pirapira: A comment here is the only remaining issue for me.

PR #4779, opened 2018-01-16
testeth: clean up LLL helpers (a tiny bit, perhaps, hopefully)

axic:


winsvega:
let mer fun testfiller test

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -231,27 +231,47 @@ bytes importByteArray(string const& _str)
...
-        data = importByteArray(replaceLLL(_o.at("data").get_str()));
-    else
-        for (auto const& j : _o.at("data").get_array())
-            data.push_back(toByte(j));
-    return data;
+    return processDataOrCode(_o, "data");
 }
 
-string replaceLLL(string const& _code)
+string replaceBytecode(string const& _code)
```
winsvega: perhaps name this function compileCode()  
as it will be used to compile LLL, wasm and whatever.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -231,27 +231,47 @@ bytes importByteArray(string const& _str)
...
 {
+    if (_code == "")
+        return "0x";
+    if (_code.substr(0, 2) == "0x" && _code.size() >= 2)
+    {
+        checkHexHasEvenLength(_code);
```
winsvega: this check should also be done after compileLLL 
so to make sure that after compileLLL we return something like 0x.... 

but. we should not engage compileLLL if the code is already hash 0x...
I guess it should be a sub function (259-265)

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -231,27 +231,47 @@ bytes importByteArray(string const& _str)
...
+
     string compiledCode = compileLLL(_code);
     if (_code.size() > 0)
         BOOST_REQUIRE_MESSAGE(compiledCode.size() > 0,
             "Bytecode is missing! '" + _code + "' " + TestOutputHelper::get().testName());
     return compiledCode;
 }
 
-void replaceLLLinState(json_spirit::mObject& _o)
+void replaceBytecodeinState(json_spirit::mObject& _o)
```
winsvega: compileCodeInState

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -394,21 +406,7 @@ void checkHexHasEvenLength(string const& _str)
...
-        else
-            code = importByteArray(_o.at("code").get_str());
-    else if (_o.at("code").type() == json_spirit::array_type)
-    {
-        code.clear();
-        for (auto const& j : _o.at("code").get_array())
-            code.push_back(toByte(j));
-    }
-    return code;
+    return processDataOrCode(_o, "code");
```
winsvega: now why do we need importCode function?

axic: Feel free to do it - this PR tried to keep changes at minimum.

test/tools/libtesteth/TestHelper.cpp
axic: Okay will rename `bytecode` to `code` in both functions.

test/tools/libtesteth/TestHelper.cpp
axic: `compileLLL` runs `checkHexHasEvenLength` and inserts the `0x` prefix on its own

axic:
@winsvega fixed. Happy to merge?

axic:
@winsvega so you have tested this with the fillers and works properly?

winsvega:
yes. although it changes empty data to have 0x  which I recall had an effect on test generation somewhere. transaction tests maybe. but all the tests are green. LLL compilation works.

PR #4787, opened 2018-01-18
debug macos [WIP]

winsvega:
save random state test before running the filler.

whenever all tests are running on macos 
this change will save the test filler of randomStateTest so we could see if macOS really fails because of some random code execution issue.

I suggest to merge this PR so we have better chance to hit this bug
this change then will be removed after we solve the issue.


chfast:
How do you want to access this file?

winsvega:
you right. hmm. then I guess I will have to print out the test filler into cerr.

chfast:
You can upload it as an artifact. Check CircleCI config, it stores `eth`.

winsvega:
how to access the artifact?

winsvega:
gived up. testeth random code support on macos is not neccessary.

chfast:
Please delete branches you don't need any more!!!

PR #4788, opened 2018-01-18
EVM-C: Rejected

chfast:


libevm/EVMC.cpp
```diff
@@ -23,33 +23,38 @@ EVM::EVM(evm_instance* _instance) noexcept : m_instance(_instance)
...
-	{
-		cwarn << "Execution rejected by EVM JIT (gas limit: " << io_gas << "), executing with interpreter";
-		return VMFactory::create(VMKind::Interpreter)->exec(io_gas, _ext, _onOp);
-	}
+    bool rejected = false;
+    // Do not accept requests with gas > 2^63 (int64 max)
+    rejected |= io_gas > std::numeric_limits<int64_t>::max();
+    rejected |= _ext.envInfo().number() > std::numeric_limits<int64_t>::max();
+    rejected |= _ext.envInfo().timestamp() > std::numeric_limits<int64_t>::max();
+    rejected |= _ext.envInfo().gasLimit() > std::numeric_limits<int64_t>::max();
```
axic: Would be nice having an option which disables the fallback to interpreter if these conditions fail.

axic:
Please add the `set option` to Hera and update the submodule pointer before merging this.

libevm/EVMC.cpp
chfast: They are quite legacy one, affected only by some exotic tests.

axic: That is fair enough, but I'd like to be sure that it is possible running a pure eWASM client without silently falling back on the interpreter.

chfast: You will not be able to run eWASM anyway, because these values will not fit into 64-bit types in EVM-C.
I can change them into failures, let's see what tests are there.

axic: Yep, the option would mean to throw a failure here.

I guess you want this configurable in order to successfully run state tests using evmjit.

libevm/EVMC.cpp
```diff
@@ -23,33 +23,38 @@ EVM::EVM(evm_instance* _instance) noexcept : m_instance(_instance)
...
-	rejected |= _ext.envInfo().timestamp() > std::numeric_limits<int64_t>::max();
-	rejected |= _ext.envInfo().gasLimit() > std::numeric_limits<int64_t>::max();
-	if (rejected)
-	{
-		cwarn << "Execution rejected by EVM JIT (gas limit: " << io_gas << "), executing with interpreter";
-		return VMFactory::create(VMKind::Interpreter)->exec(io_gas, _ext, _onOp);
-	}
+    bool rejected = false;
+    // Do not accept requests with gas > 2^63 (int64 max)
+    rejected |= io_gas > std::numeric_limits<int64_t>::max();
```
axic: I think this should also test for `_ext.depth`. I know it is unlikely due to other restrictions, but `execute` does a typecast `static_cast<int32_t>(_ext.depth)` and it feels better to do this correctly.

libevm/EVMC.cpp
```diff
@@ -23,33 +23,38 @@ EVM::EVM(evm_instance* _instance) noexcept : m_instance(_instance)
...
+    rejected |= _ext.envInfo().number() > std::numeric_limits<int64_t>::max();
+    rejected |= _ext.envInfo().timestamp() > std::numeric_limits<int64_t>::max();
+    rejected |= _ext.envInfo().gasLimit() > std::numeric_limits<int64_t>::max();
 
-	auto gas = static_cast<int64_t>(io_gas);
-	auto r = execute(_ext, gas);
+    if (!rejected)
+    {
+        auto gas = static_cast<int64_t>(io_gas);
+        auto r = execute(_ext, gas);
```
axic: Why these two have to be auto?

chfast: My linter suggest using `auto` if the type would be repeated (case: `auto gas = static_cast<int64_t>`).
In the second case, I probably should write the type.

axic:
Must be rebased after #4796.

chfast:
Rebased.

libevm/VMFactory.cpp
```diff
@@ -51,11 +49,9 @@ VMKindTableEntry vmKindsTable[] = {
     {VMKind::Interpreter, "interpreter"},
 #if ETH_EVMJIT
     {VMKind::JIT, "jit"},
-    {VMKind::Smart, "smart"},
 #endif
 #if ETH_HERA
```
pirapira: Does Travis enable `ETH_HERA`?

chfast: No.

axic: We should consider enabling that.

PR #4789, opened 2018-01-19
store randomcode seed on macOS

winsvega:
store the seed of random code generator on macOs as artifact
so if it fails we could possibly reproduce the random code test

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -69,6 +69,11 @@ defaults:
       path: ~/build/test/testeth
       destination: testeth
 
+  store-randomcode-seed: &store-randomcode-seed
+    store_artifacts:
+      path: ~/build/test/randomCodeSeed.txt
+      destination: randomCodeSeed.txt
+
   save-deps-cache: &save-deps-cache
     cache-save:
       name: "Save dependencies cache"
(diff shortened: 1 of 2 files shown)


chfast:
Just print it out. Why do you need a file to dump a single number?

winsvega:
How? Ctest skip the output unless you set a specific flag. But this flag will apply to all of our tests.

test/tools/fuzzTesting/BoostRandomCode.cpp
```diff
@@ -46,7 +46,12 @@ BoostRandomCode::BoostRandomCode()
     {
         auto now = std::chrono::steady_clock::now().time_since_epoch();
         auto timeSinceEpoch = std::chrono::duration_cast<std::chrono::nanoseconds>(now).count();
-        gen.seed(static_cast<unsigned int>(timeSinceEpoch));
+		unsigned int time = static_cast<unsigned int>(timeSinceEpoch);
```
chfast: Tabs used here.

circle.yml
```diff
@@ -69,6 +69,11 @@ defaults:
       path: ~/build/test/testeth
       destination: testeth
 
+  store-randomcode: &store-randomcode
```
chfast: How about naming it `store-randomcode-seed`?

PR #4796, opened 2018-01-22
EVMC: Do not fall back to Interpreter

chfast:
Before EVMC was falling back to execution in the Interpreter if any of values: block gas limit, timestamp, block number was enormous. It seems we don't have tests like those any more.

I tested `testeth -t BlockchainTests,GeneralStateTests,VMTests -- --vm jit --all`.

Related to #4788.

### DIFF
--- a/libevm/EVMC.cpp
+++ b/libevm/EVMC.cpp
@@ -21,35 +21,34 @@ EVM::EVM(evm_instance* _instance) noexcept : m_instance(_instance)
         m_instance->set_option(m_instance, pair.first.c_str(), pair.second.c_str());
 }
 
-owning_bytes_ref EVMC::exec(u256& io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp)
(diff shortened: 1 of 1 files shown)


libevm/EVMC.cpp
```diff
@@ -21,35 +21,33 @@ EVM::EVM(evm_instance* _instance) noexcept : m_instance(_instance)
...
 
-	auto gas = static_cast<int64_t>(io_gas);
-	auto r = execute(_ext, gas);
+    // TODO: The following checks should be removed by changing the types
+    //       used for gas, block number and timestamp.
+    (void)int64max;
+    assert(io_gas <= int64max);
+    assert(_ext.envInfo().number() <= int64max);
+    assert(_ext.envInfo().timestamp() <= int64max);
+    assert(_ext.envInfo().gasLimit() <= int64max);
```
axic: Please also assert for "static_cast<int32_t>(_ext.depth)"

winsvega:
@pirapira, are those limits defined by yellowpaper now?

pirapira:
@winsvega your question will be tracked at https://github.com/ethereum/yellowpaper/issues/474

PR #4797, opened 2018-01-22
Timestamp as int64_t

chfast:


chfast:
@winsvega Can you change the timestamps in difficulty tests to something _before_ the Sun dies.

pirapira:
All this seems to come from the yellow paper (37) `Hₛ ∈ P₂₅₆`.  The time stamp is supposed to be a 256-bit number.

chfast:
I guess the values in tests are up to 64-bits long. I need 63-bit, because we are using `int64_t`.

chfast:
But a YP change would be nice to have.

pirapira:
I didn't know we are currently testing at most 64-bit.  Then doing it 63-bit is also fine.

chfast:
All timestamps in JSON files are 8 bytes long. @winsvega said they are random.

pirapira:
The test generation procedure has to be documented https://github.com/ethereum/tests/issues/404 .

winsvega:
it is 
`u256 pStamp = test::RandomCode::get().randomUniInt();`

chfast:
How to regenerate them?

chfast:
The `randomUniInt()` returns UINT64_MAX max. Funny :)

winsvega:
if you change the funtion then it is not randomUniInt anymore. (unsigned)

chfast:
Values from 0 to 9223372036854775807. Looks unsigned to me.

winsvega:
int64_t and uint64_t  have the same max() ?

chfast:
No.

chfast:
Now, by default, you get a random int from uniform distribution between 0 and 9223372036854775807. Before it was between 0 and 18446744073709551615. If you want different max value, you can pass is as an argument.

I want int64 max to be default "big" value, not uint64 as before. Because we don't want to relay on any Ethereum client to have unsigned types.

winsvega:
this change will also affect random code generation. but it should be fine

chfast:
Review now?

axic:
@chfast should the evm-c interface be updated (`EnvInfo`) and the static assert be removed?

chfast:
You are right @axic, I missed this one. Will remove it in the next change.

PR #4804, opened 2018-01-23
testeth expand expect section network syntax

winsvega:
add >Net, <Net, >=Net, <=Net

### DIFF
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -596,6 +596,9 @@ bool ImportTest::checkGeneralTestSectionSearch(json_spirit::mObject const& _expe
     else
         network.emplace(_network);
 
+    // replace ">Homestead" with "Homestead, EIP150, ..."
+    network = test::translateNetworks(network);
+
     BOOST_CHECK_MESSAGE(network.size() > 0, TestOutputHelper::get().testName() + " Network array not set!");
 	checkAllowedNetwork(network);
 
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -547,23 +547,34 @@ void parseJsonIntValueIntoVector(json_spirit::mValue const& _json, vector<int>& 
 		_out.push_back(_json.get_int());
 }
 
+set<string> const& getAllowedNetworks()
+{
+    static set<string> allowedNetowks;
+    if (allowedNetowks.empty())
+    {
+        allowedNetowks.emplace("ALL");
+        for (auto const& net : test::getNetworks())
+            allowedNetowks.emplace(test::netIdToString(net));
+    }
+    return allowedNetowks;
+}
+
+void ImportTest::checkAllowedNetwork(string const& _network)
+{
+    set<string> const& allowedNetowks = getAllowedNetworks();
+    if (!allowedNetowks.count(_network))
+    {
+        // Can't use boost at this point
+        std::cerr << TestOutputHelper::get().testName() + " Specified Network not found: "
+                  << _network << "\n";
+        exit(1);
+    }
+}
+
 void ImportTest::checkAllowedNetwork(std::set<std::string> const& _networks)
 {
-    set<eth::Network> const& allnetworks = test::getNetworks();
-    set<string> allowedNetowks;
-    allowedNetowks.emplace("ALL");
-    for (auto const& net : allnetworks)
-        allowedNetowks.emplace(test::netIdToString(net));
-
     for (auto const& net: _networks)
-	{
-        if (!allowedNetowks.count(net))
-        {
-			//Can't use boost at this point
-			std::cerr << TestOutputHelper::get().testName() + " Specified Network not found: " << net << "\n";
-			exit(1);
-		}
-	}
+        ImportTest::checkAllowedNetwork(net);
 }
 
 bool ImportTest::checkGeneralTestSection(json_spirit::mObject const& _expects, vector<size_t>& _errorTransactions, string const& _network) const
@@ -596,11 +607,9 @@ bool ImportTest::checkGeneralTestSectionSearch(json_spirit::mObject const& _expe
     else
         network.emplace(_network);
 
-    // replace ">Homestead" with "Homestead, EIP150, ..."
+    // replace ">=Homestead" with "Homestead, EIP150, ..."
     network = test::translateNetworks(network);
-
     BOOST_CHECK_MESSAGE(network.size() > 0, TestOutputHelper::get().testName() + " Network array not set!");
-	checkAllowedNetwork(network);
 
 	if (!Options::get().singleTestNet.empty())
 	{
--- a/test/tools/libtesteth/TestHelper.cpp
+++ b/test/tools/libtesteth/TestHelper.cpp
@@ -198,35 +198,46 @@ set<string> translateNetworks(set<string> const& _networks)
         size_t sizeStart = out.size();
         string possibleNet = net.substr(1, net.length() - 1);
         vector<string>::iterator it = std::find(forks.begin(), forks.end(), possibleNet);
-        if (net[0] == '>' && it != forks.end())
+
+        if (it != forks.end() && net.size() > 1)
         {
-            while (++it != forks.end())
-                out.emplace(*it);
-        }
-        else if (net[0] == '<' && it != forks.end())
-        {
-            while (it != forks.begin())
-                out.emplace(*(--it));
+            if (net[0] == '>')
+            {
+                while (++it != forks.end())
+                    out.emplace(*it);
+            }
+            else if (net[0] == '<')
+            {
+                while (it != forks.begin())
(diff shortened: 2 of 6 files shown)


test/tools/libtesteth/TestHelper.cpp
```diff
@@ -181,6 +181,55 @@ set<eth::Network> const& getNetworks()
...
+    for (auto const& net : getNetworks())
+        forks.push_back(test::netIdToString(net));
+
+    set<string> out;
+    for (auto const& net : _networks)
+    {
+        size_t sizeStart = out.size();
+        string possibleNet = net.substr(1, net.length() - 1);
+        vector<string>::iterator it = std::find(forks.begin(), forks.end(), possibleNet);
+        if (net[0] == '>' && it != forks.end())
```
pirapira: Check emptiness of `net` before accessing `net[0]`.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -181,6 +181,55 @@ set<eth::Network> const& getNetworks()
...
+        else if (net[0] == '<' && net[1] == '=' && it != forks.end())
+        {
+            out.emplace(*it);
+            while (it != forks.begin())
+                out.emplace(*(--it));
+        }
+
+        // if nothing has been inserted, just push the untranslated network as is
+        if (out.size() == sizeStart)
+            out.emplace(net);
```
pirapira: This is a good place to catch a typo in the network name.

test/unittests/libtesteth/testHelperTest.cpp
```diff
@@ -0,0 +1,77 @@
...
+    BOOST_REQUIRE(networks.count("Frontier") > 0);
+    BOOST_REQUIRE(networks.count("Homestead") > 0);
+    for (auto const& net : test::getNetworks())
+    {
+        if (net != eth::Network::FrontierTest && net != eth::Network::HomesteadTest)
+            BOOST_REQUIRE(networks.count(test::netIdToString(net)) == 0);
+    }
+}
+
+BOOST_AUTO_TEST_SUITE_END()
```
pirapira: :+1:

test/tools/libtesteth/TestHelper.h
```diff
@@ -77,6 +77,7 @@ std::string netIdToString(eth::Network _netId);
 eth::Network stringToNetId(std::string const& _netname);
 bool isDisabledNetwork(eth::Network _net);
 std::set<eth::Network> const& getNetworks();
+std::set<std::string> translateNetworks(std::set<std::string> const& _networks);
```
pirapira: Please add a comment above this line, and explain what the function does.

test/tools/libtesteth/TestHelper.cpp
winsvega: this function not being callled on every path that define networks

pirapira: If the network name is checked already, please add an assertion.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -181,6 +181,66 @@ set<eth::Network> const& getNetworks()
...
+                    out.emplace(*(it++));
+            }
+            else if (net[0] == '<' && net[1] == '=')
+            {
+                out.emplace(*it);
+                while (it != forks.begin())
+                    out.emplace(*(--it));
+            }
+        }
+
```
pirapira: Perhaps, if `net` is `"ALL"`, that should be translated into all networks.  "ALL" is very similar to ">=Frontier"

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -547,23 +547,34 @@ void parseJsonIntValueIntoVector(json_spirit::mValue const& _json, vector<int>&
...
 {
-    set<eth::Network> const& allnetworks = test::getNetworks();
-    set<string> allowedNetowks;
-    allowedNetowks.emplace("ALL");
-    for (auto const& net : allnetworks)
-        allowedNetowks.emplace(test::netIdToString(net));
+    static set<string> allowedNetowks;
+    if (allowedNetowks.empty())
+    {
+        allowedNetowks.emplace("ALL");
```
pirapira: Should this allow `ALL`?

test/tools/libtesteth/TestHelper.cpp
winsvega: this would require to refactor a bit more of the other code which rely on ALL

test/tools/libtesteth/ImportTest.cpp
winsvega: this is a list of strings that are allowed at the expect section network field

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -181,6 +181,66 @@ set<eth::Network> const& getNetworks()
...
+{
+    // construct vector with test network names in a right order (from Frontier to Homestead ... to
+    // Constantinople)
+    vector<string> forks;
+    for (auto const& net : getNetworks())
+        forks.push_back(test::netIdToString(net));
+
+    set<string> out;
+    for (auto const& net : _networks)
+    {
```
pirapira: Optional: you can factor out
```
set<string> translateNetwork(string const& _network)
```
which does only one iteration of this loop.

test/tools/libtesteth/TestHelper.cpp
winsvega: also will hawe to change tests that are using ALL

test/tools/libtesteth/TestHelper.cpp
winsvega: what you mean by "factor out"?

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -181,6 +181,66 @@ set<eth::Network> const& getNetworks()
...
+            else if (net[0] == '<' && net[1] == '=')
+            {
+                out.emplace(*it);
+                while (it != forks.begin())
+                    out.emplace(*(--it));
+            }
+        }
+
+        // if nothing has been inserted, just push the untranslated network as is
+        if (out.size() == sizeStart)
```
pirapira: Please add a test case on
`{Frontier, <Homestead}`
That might break this code.

test/tools/libtesteth/TestHelper.cpp
pirapira: To factor out a piece of code is to move it into the body of a new function.

test/tools/libtesteth/ImportTest.cpp
pirapira: Is `<Homestead` not allowed, but `ALL` is allowed?

test/tools/libtesteth/TestHelper.cpp
pirapira: I see.  Maybe that's not for now.

test/tools/libtesteth/ImportTest.cpp
pirapira: Same as above.  Consideration for the future.

winsvega: <Homestead not allowed. only Homestead, Frontier, EIP150, EIP158, Byzantium and Constantinople. 
ALL is allowed and means all of that.

test/tools/libtesteth/TestHelper.cpp
winsvega: hm. you right.

test/tools/libtesteth/TestHelper.cpp
winsvega: if the subfunction (in this case cycle body) is not used anywhere else. I prefere to keep it at one place. unless it is too big.

pirapira: OK.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -181,6 +181,77 @@ set<eth::Network> const& getNetworks()
...
+            {
+                while (it != forks.end())
+                {
+                    out.emplace(*(it++));
+                    isNetworkTranslated = true;
+                }
+            }
+            else if (net[0] == '<' && net[1] == '=')
+            {
+                out.emplace(*it);
```
pirapira: `isNetworkTranslated` should be already true here.

pirapira: Please add a unit test for `<=Frontier`.

winsvega: why should it be true?

pirapira: Because a network name is translated.

pirapira: And the execution might not enter `while`.

winsvega: I see

winsvega: fixed

PR #4805, opened 2018-01-24
Remove executable flag on files in libevm

axic:


pirapira:
Waiting for CI builds, though a failure would be surprising.

PR #4809, opened 2018-01-26
filltests only on expected networks

winsvega:
fixed the bug: --filltests --fillchain  leaded to blockchain tests being generated twice

testeth optimize: fill tests only on those networks that are specified in expect section of StateTests/BlockchainTests 

expect section is now required. 

TODO:
optimize and refill the test fillers. 
*this PR breaks --filltests command on existing test cases

### DIFF
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -205,7 +205,7 @@ bytes ImportTest::executeTest(bool _isFilling)
     }
 
     vector<transactionToExecute> transactionResults;
-	for (auto const& net : networks)
+    for (auto const& net : networks)
 	{
 		if (isDisabledNetwork(net))
 			continue;
@@ -308,7 +308,7 @@ json_spirit::mObject ImportTest::makeAllFieldsHex(json_spirit::mObject const& _i
         bool isData = (key == "data");
 
         if (_isHeader && key == "nonce")
-			isHash = true;
+            isHash = true;
 
 		std::string str;
 		json_spirit::mValue value = i.second;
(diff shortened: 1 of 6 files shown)


pirapira:
I don't understand.  If a test was generated twice, you can cancel the second.  You don't need to change the requirements, like "expect is now needed".

winsvega:
The bug was fixed by chance. The actual PR is about expect section

pirapira:
@winsvega now I don't know if it's a good change.  Many clients only see the filled tests, so they will run fewer tests after this PR.  What's the benefit?

winsvega:
the whole goal of this PR was to optimize the test run.
like we do not need to run staticcall tests on forks before the staticcall was introduced. 
this will remove hundreds of tests with no actuall coverage.

pirapira:
The Travis failure seems relevant.

winsvega:
yes. either I fix 2 specific tests now (the build will become green) and then eventually fix all other tests. (refill) 
or we would keep this PR opef for a little while longer

pirapira:
Fixing two specific tests will be enough, provided that the review passes.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -164,18 +163,49 @@ void ImportTest::makeBlockchainTestFromStateTest(set<eth::Network> const& _netwo
...
+            BOOST_REQUIRE(m_testInputObject.count("expect") > 0);
+            networks = getAllNetworksFromExpectSections(m_testInputObject.at("expect").get_array());
+        }
+        else
+        {
+            // Run tests only on networks that are in post state of the filled test
+            for (auto const& post : m_testInputObject.at("post").get_obj())
+                networks.emplace(test::stringToNetId(post.first));
+        }
+    }
```
pirapira: Why not
```
else if (_isFilling)
{
   ...
}
else
{
  ...
}
```

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -164,18 +163,49 @@ void ImportTest::makeBlockchainTestFromStateTest(set<eth::Network> const& _netwo
...
 
-bytes ImportTest::executeTest()
+/// returns all networks that are defined in all expect sections
+set<eth::Network> ImportTest::getAllNetworksFromExpectSections(json_spirit::mArray const& _expects)
 {
-	assert(m_envInfo);
+    set<string> allNetworks;
+    for (auto const& exp : _expects)
+    {
+        requireJsonFields(exp.get_obj(), "expect",
```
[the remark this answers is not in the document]
winsvega: no. thats why the check is here

test/tools/libtesteth/ImportTest.cpp
winsvega: change after change after change. 
would you like to refactor this?

pirapira: That's how we work on code, no?

pirapira:
Waiting for AppVeyor.

PR #4811, opened 2018-01-26
Update Dockerfile for eth

chfast:


scripts/docker/eth.dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# Build stage
+FROM alpine:latest as builder
+RUN apk add --no-cache \
+        git \
+        cmake \
+        g++ \
+        make \
+        linux-headers \
+        leveldb-dev --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing/
+RUN git clone --recursive https://github.com/ethereum/cpp-ethereum --branch develop --single-branch --depth 1
```
chriseth: I'm also not an expert, but I think if you use multiple run-statements, the intermediate snapshots will be stored, so your image will be larger.

scripts/docker/eth.dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN mkdir /build
+RUN cd /build && cmake /cpp-ethereum -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=RelWithDebInfo -DTOOLS=OFF -DTESTS=OFF
+RUN cd /build && make -j8 && make install
+
+# Install stage
+FROM alpine:latest
```
chriseth: I have never seen multiple from statements - which effect does that have?

chfast: I've just found it today by inspecting `--squash` option: https://docs.docker.com/engine/userguide/eng-image/multistage-build/.

In this case we have to images: builder and eth. The builder one is discarded, so multiple layers does not matter. In the "eth" stage we just copy the binary from the "builder" stage. The final image is 125 Mb.

holiman:
Lgtm!

scripts/docker/eth.dockerfile
chriseth: Oh that's nice! Still, 125 MB sounds quite big to me. Hm, ok, that's probably the debugging info:

My local eth binary is 140MB big, after stripping, it has 8.3 MB.

chfast: Yes, it includes debug symbols.

PR #4822, opened 2018-01-31
 Change type for block number to int64

chfast:


test/tools/jsontests/vm.cpp
```diff
@@ -106,7 +106,7 @@ EnvInfo FakeExtVM::importEnv(mObject const& _o, LastBlockHashesFace const& _last
 	blockHeader.setDifficulty(toInt(_o.at("currentDifficulty")));
 	blockHeader.setTimestamp(toInt64(_o.at("currentTimestamp")));
 	blockHeader.setAuthor(Address(_o.at("currentCoinbase").get_str()));
-	blockHeader.setNumber(toInt(_o.at("currentNumber")));
+	blockHeader.setNumber(toInt64(_o.at("currentNumber")));
```
pirapira: Have you thought about using `toPositiveInt64` here?

libevm/EVMC.cpp
```diff
@@ -29,8 +29,6 @@ owning_bytes_ref EVMC::exec(u256& io_gas, ExtVMFace& _ext, const OnOpFunc& _onOp
     //       used for gas, block number and timestamp.
     (void)int64max;
     assert(io_gas <= int64max);
-    assert(_ext.envInfo().number() <= int64max);
-    assert(_ext.envInfo().timestamp() <= int64max);
```
pirapira: Maybe check for `>= 0`?

test/tools/jsontests/vm.cpp
chfast: Yes, I can rework `toInt64` to `toPositiveInt64`.

chfast: Is the "positive" good name? We want to include 0. But "nonnegative" looks weird.

pirapira: I don't have a good alternative.  We can pretend French for a moment:

> In typical French mathematical usage, zero is both positive and negative
https://math.stackexchange.com/questions/26705/is-zero-positive-or-negative

chfast:
Changes applied.

chfast:
I'm debugging it.

winsvega:
the filler hash issue is recent commit pushed to the tests repo. but you could get it green on a commit before that

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -277,17 +277,24 @@ u256 toInt(json_spirit::mValue const& _v)
...
+        n = _v.get_int64();
+        break;
     default:
         cwarn << "Bad type for scalar: " << _v.type();
     }
+
+    if (n < 0)
+        throw std::out_of_range{"unexpected negative value: " + std::to_string(n)};
+
     return 0;
```
chfast: Ooops! Should have been `n`, not `0`.

pirapira: wow.

PR #4823, opened 2018-01-31
fix --filltests

winsvega:


pirapira:
What kind of problem does it fix?

pirapira:
Are test failures in CircleCI intended?

winsvega:
failures are happening because we have to refill the tests. its in progress. 

it fixes expect section fields check for the blockchain test filler (your comment on a previous PR)

pirapira:
@winsvega so, you are saying the tests should fail on `cpp-ethereum:develop` as well?

pirapira:
@winsvega why is the disabling part of a fix PR?

winsvega:
as I mentioned in skype. we have broken the tests with last commit to the test repo. 
if there was PR with that commit to the cpp develop it wont work

pirapira:
@winsvega https://circleci.com/gh/ethereum/workflows/cpp-ethereum/tree/develop the test is working until we update it.

pirapira:
@winsvega so, this PR breaks CircleCI tests, I think.

winsvega:
not the PR. the tests are not fixed yet

pirapira:
@winsvega CircleCI is not seeing the broken tests yet.  It's seeing some old revision of `ethereum/tests` when everything was fine.  That's why the CircleCI test passes on `develop`.

winsvega:
wait the PR has actually an issue. bc test filler expect section network field could be both string and array...

winsvega:
there are already tests for array. but no tests for string. string are generated with --fillchain option. technically we could generate array. but that would easier to be done when updating all the tests. cause now we have lots of tests with both array and string type.

pirapira:
So, Travis should fill blockchain tests with network section as array and as string type.

PR #4837, opened 2018-02-07
Fix static initialization fiasco around Invalid256

gumb0:
This came up suddenly - `Invalid256` was a global defined in `livedevcore/Common.cpp` and it was used in initialization of `NullBlockDetails` defined in `BlockDetails.cpp`

### DIFF
--- a/libdevcore/Common.cpp
+++ b/libdevcore/Common.cpp
@@ -1,18 +1,18 @@
 /*
-	This file is part of cpp-ethereum.
+    This file is part of cpp-ethereum.
 
-	cpp-ethereum is free software: you can redistribute it and/or modify
-	it under the terms of the GNU General Public License as published by
-	the Free Software Foundation, either version 3 of the License, or
-	(at your option) any later version.
+    cpp-ethereum is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
 
-	cpp-ethereum is distributed in the hope that it will be useful,
-	but WITHOUT ANY WARRANTY; without even the implied warranty of
-	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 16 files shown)


gumb0:
@chfast Yes, I tried, `u256` can't be `constexpr`, `boost::multiprecision` doesn't have `constexpr` constructor (I guess it does heap allocation?)

chfast:
It does not in the "fixed size" form, but maybe it's to complex to be added or they didn't have time to do it yet.

chfast:
Please check out this: http://www.boost.org/doc/libs/1_66_0/libs/multiprecision/doc/html/boost_multiprecision/tut/lits.html.

gumb0:
@chfast ah right, I remember now, constructor is `constexpr`, but the problem is that it's defined as
```
u256 const Invalid256 = ~(u256)0;
```
And `operator~` is definitely not `constexpr`
>However, you will not be able to perform compile-time arithmetic on such types.

gumb0:
`std::numeric_limits<u256>::max()`,  `std::numeric_limits<u256>::infinity()` are also not `constexpr`

chfast:
Try `constexpr static uint256 Invalid256{"0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"};`

gumb0:
^ This doesn't work, I guess constructor from string is not `constexpr`.
But this works, at least on windows:
```
    using namespace boost::multiprecision::literals;
    constexpr u256 Invalid256 = 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff_cppui256;
```

I'll make another PR then.

chfast:
I wanted to suggest literal, but though it will fallback to the constructor anyway. A bit weird :)

gumb0:
I see some metaprogramming magic there in the implemenation of literals, parsing the string through C++ templates etc. https://github.com/boostorg/multiprecision/blob/develop/include/boost/multiprecision/cpp_int/literals.hpp
So it's not a regular constructor magically working at compile time

PR #4838, opened 2018-02-08
Conditionally enable bitshift on ConstantinopleForkBlock

tcsiwula:
Conditionally enable bitshift on ConstantinopleForkBlock

### DIFF
--- a/libethcore/EVMSchedule.h
+++ b/libethcore/EVMSchedule.h
@@ -33,6 +33,7 @@ struct EVMSchedule
 	bool haveDelegateCall = true;
 	bool eip150Mode = false;
 	bool eip158Mode = false;
+	bool haveBitwiseShifting = false;
 	bool haveRevert = false;
 	bool haveReturnData = false;
 	bool haveStaticCall = false;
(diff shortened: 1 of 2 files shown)


libethcore/EVMSchedule.h
```diff
@@ -33,6 +33,7 @@ struct EVMSchedule
 	bool haveDelegateCall = true;
 	bool eip150Mode = false;
 	bool eip158Mode = false;
+	bool eip145Mode = false;
```
gumb0: We should better give it more descriptive name like `haveBitwiseShifting`

gumb0:
Hey Tim, one more thing that needs to be done here is actually using the flag that you've added.

So the goal is to get rid of compile-time checks `#if EIP_145` in https://github.com/ethereum/cpp-ethereum/pull/4054 and insted check this flag in every shift instruction 
```
			if (!m_schedule->haveBitwiseShifting)
				throwBadInstruction();
```

libevm/VM.cpp
```diff
@@ -653,6 +660,10 @@ void VM::interpretCases()
 
 		CASE(SAR)
 		{
+			// Pre-constantinople
+			if (!m_schedule->haveBitwiseShifting)
+				throwBadInstruction();
+
 			ON_OP();
 			updateIOGas();
```
axic: There is a missed else-endif block below.

tcsiwula: Thanks Alex :)

axic:
Can you please squash some of these commits? :)

tcsiwula:
Hoping this works ^_^

axic:
Thanks!

PR #4846, opened 2018-02-14
Script to plot sync performance

gumb0:
Finally I've created a script to somehow analyze "SLOW IMPORT" records in the log


gumb0:
For example here are some results for the recent ~24 hour run of sync after snapshot import
image
image
image

gumb0:
image

PR #4848, opened 2018-02-14
[WIP] Move JSON code to jsoncpp

axic:


axic:
I think this can be closed, because in the last half a year it seems everything moved over to `json_spirit`, so it requires a lot of dedication to rewrite everything.

PR #4850, opened 2018-02-16
Refactor logging

gumb0:
Rewriting logging system using boost.log

- [x] Create basic logging primitives in `Log.h`
- [ ] Change all logging statements to use new logging
- [ ] Separate global logger for critical errors, also probably remove `cdebug`, `clog`, `cnote`, `ctrace` or leave only one of them
- [ ] `setupLogging()` in all executables
- [ ] Get rid of global verbosity variable (make it argument to `setupLogging()` instead)
- [ ] new command line parameter for filtering channels (probably create program options group to handle it easily in all tools)
- [ ] Handle exceptions from logging system

### Summary of changes noticeable to user
- Names of many channels will be changed, all channels will have readable names to allow filtering by channel (e.g. there will be `warn` instead of `  ✘`)
- Channel names won't be colored, so log will be less colorful overall
- "AutoSpacing" feature is removed - it allowed to output to log stream not caring about spaces between values, like `cwarn << "Transaction hash" << hash << "already imported";` Now spaces need to be explicit like `cwarn << "Transaction hash " << hash << " already imported";` I'll try to change this in all log messages, but I might miss some.
- Some channels were supposed to be visible only in Debug build (the ones with `LogChannel::debug == true`) - but this was working only for Windows, where `NDEBUG` macro is defined in debug, for other platforms debug channels were visible in Release, too. This difference is removed, channel visibility will depend only on filtering, but not on build type.

Log message line will look like
```
2018-04-17 12:15:59 eth timer Block import 5123675 3184 ms
```
(`eth` is the name of the thread, `timer` is the name of the channel)

### DIFF
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -829,6 +829,7 @@ int main(int argc, char** argv)
         }
     }
 
+    setupLogging(g_logVerbosity);
 
     if (!privateChain.empty())
     {
--- a/libdevcore/Log.cpp
+++ b/libdevcore/Log.cpp
@@ -28,13 +28,25 @@
 #include <pthread.h>
 #endif
 #include "Guards.h"
+
+#include <boost/core/null_deleter.hpp>
+#include <boost/log/attributes/clock.hpp>
+#include <boost/log/attributes/function.hpp>
+#include <boost/log/expressions.hpp>
+#include <boost/log/sinks/async_frontend.hpp>
+#include <boost/log/sinks/text_ostream_backend.hpp>
+#include <boost/log/sources/global_logger_storage.hpp>
+#include <boost/log/sources/severity_channel_logger.hpp>
+#include <boost/log/support/date_time.hpp>
+
 using namespace std;
-using namespace dev;
 
 //⊳⊲◀▶■▣▢□▷◁▧▨▩▲◆◉◈◇◎●◍◌○◼☑☒☎☢☣☰☀♽♥♠✩✭❓✔✓✖✕✘✓✔✅⚒⚡⦸⬌∅⁕«««»»»⚙
 
+namespace dev
+{
 // Logging
-int dev::g_logVerbosity = 5;
+int g_logVerbosity = 5;
 mutex x_logOverride;
 
 /// Map of Log Channel types to bool, false forces the channel to be disabled, true forces it to be enabled.
@@ -42,7 +54,7 @@ mutex x_logOverride;
 /// or equal to the currently output verbosity (g_logVerbosity).
 static map<type_info const*, bool> s_logOverride;
 
-bool dev::isChannelVisible(std::type_info const* _ch, bool _default)
+bool isChannelVisible(std::type_info const* _ch, bool _default)
 {
     Guard l(x_logOverride);
     if (s_logOverride.count(_ch))
@@ -74,7 +86,7 @@ const char* RightChannel::name() { return EthGreen "-->"; }
 const char* WarnChannel::name() { return EthOnRed EthBlackBold "  X"; }
 const char* NoteChannel::name() { return EthBlue "  i"; }
 const char* DebugChannel::name() { return EthWhite "  D"; }
-const char* TraceChannel::name() { return EthGray "..."; }
+
 #else
 const char* LogChannel::name() { return EthGray "···"; }
 const char* LeftChannel::name() { return EthNavy "◀▬▬"; }
@@ -82,7 +94,6 @@ const char* RightChannel::name() { return EthGreen "▬▬▶"; }
 const char* WarnChannel::name() { return EthOnRed EthBlackBold "  ✘"; }
 const char* NoteChannel::name() { return EthBlue "  ℹ"; }
 const char* DebugChannel::name() { return EthWhite "  ◇"; }
-const char* TraceChannel::name() { return EthGray "..."; }
 #endif
 
 LogOutputStreamBase::LogOutputStreamBase(char const* _id, std::type_info const* _info, unsigned _v, bool _autospacing):
@@ -147,22 +158,22 @@ ThreadLocalLogContext g_logThreadContext;
 
 ThreadLocalLogName g_logThreadName("main");
 
-void dev::ThreadContext::push(string const& _n)
+void ThreadContext::push(string const& _n)
 {
     g_logThreadContext.push(_n);
 }
 
-void dev::ThreadContext::pop()
+void ThreadContext::pop()
 {
     g_logThreadContext.pop();
 }
 
-string dev::ThreadContext::join(string const& _prior)
+string ThreadContext::join(string const& _prior)
 {
     return g_logThreadContext.join(_prior);
 }
 
-string dev::getThreadName()
+string getThreadName()
 {
 #if defined(__GLIBC__) || defined(__APPLE__)
     char buffer[128];
@@ -174,7 +185,7 @@ string dev::getThreadName()
 #endif
 }
 
-void dev::setThreadName(string const& _n)
+void setThreadName(string const& _n)
 {
 #if defined(__GLIBC__)
     pthread_setname_np(pthread_self(), _n.c_str());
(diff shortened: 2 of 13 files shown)


chfast:
This tabs to spaces conversion here is quite annoying. Possible to submit it separately first?

gumb0:
@chfast Ok, I'll create PR with tabs conversion of files touched here

libethereum/State.cpp
```diff
@@ -546,18 +546,17 @@ void State::rollback(size_t _savepoint)
 
 std::pair<ExecutionResult, TransactionReceipt> State::execute(EnvInfo const& _envInfo, SealEngineFace const& _sealEngine, Transaction const& _t, Permanence _p, OnOpFunc const& _onOp)
 {
-    auto onOp = _onOp;
-#if ETH_VMTRACE
-    if (isChannelVisible<VMTraceChannel>())
```
gumb0: This check is not needed anymore, because in the logging statement like
```
        BOOST_LOG(traceLogger) << dumpStackMemoryStorage(ext, vm);
```
It doesn't call `dumpStackMemoryStorage(ext, vm)` at all when current log filters block all `traceLogger` messages.

gumb0:
@chfast rebased it, now should be better

eth/main.cpp
```diff
@@ -829,6 +829,7 @@ int main(int argc, char** argv)
         }
     }
 
+    setupLogging(g_logVerbosity);
```
chfast: Idea: We can create a common logging program options group and handle this with the notification.

libdevcore/Log.cpp
```diff
@@ -28,61 +28,55 @@
...
 
 /// Map of Log Channel types to bool, false forces the channel to be disabled, true forces it to be enabled.
 /// If a channel has no entry, then it will output as long as its verbosity (LogChannel::verbosity) is less than
 /// or equal to the currently output verbosity (g_logVerbosity).
 static map<type_info const*, bool> s_logOverride;
 
-bool dev::isChannelVisible(std::type_info const* _ch, bool _default)
+bool isChannelVisible(std::type_info const* _ch, bool _default)
 {
     Guard l(x_logOverride);
```
chfast: We had hard times in ethminer to figure out what this is about.

libethereum/Executive.cpp
```diff
@@ -369,21 +390,19 @@ bool Executive::executeCreate(Address const& _sender, u256 const& _endowment, u2
...
-        o << endl << "    STACK" << endl;
-        for (auto i: vm.stack())
-            o << (h256)i << endl;
-        o << "    MEMORY" << endl << ((vm.memory().size() > 1000) ? " mem size greater than 1000 bytes " : memDump(vm.memory()));
-        o << "    STORAGE" << endl;
-        for (auto const& i: ext.state().storage(ext.myAddress))
-            o << showbase << hex << i.second.first << ": " << i.second.second << endl;
-        dev::LogOutputStream<VMTraceChannel, false>() << o.str();
-        dev::LogOutputStream<VMTraceChannel, false>() << " < " << dec << ext.depth << " : " << ext.myAddress << " : #" << steps << " : " << hex << setw(4) << setfill('0') << PC << " : " << instructionInfo(inst).name << " : " << dec << gas << " : -" << dec << gasCost << " : " << newMemSize << "x32" << " >";
+        BOOST_LOG(traceLogger) << dumpStackMemoryStorage(ext, vm);
```
chfast: Can we shorten this to `LOG(trace)`?

libdevcore/Log.cpp
gumb0: That's about the feature to temporarily disable output from some channel.
This disabling was used in `Block::enact()` (maybe nowhere else), I removed it for now, don't see any difference so far.

libethereum/Executive.cpp
gumb0: Shortening `BOOST_LOG` to `LOG` sure, but logger name... loggers might be declared in various scopes (and therefore have different or no prefixes).

In general I think recommendation is have a logger defined in a scope of a task performed, usually it's a member of a class doing some job, like `m_execLogger`

Some loggers may be global, like the ones I defined at the bottom in Log.h, then the expression to log is like
```
BOOST_LOG(g_traceLogger::get()) << ...;
```

And here we have a local alias for a member logger (to capture it in lambda)

gumb0: I'm actually not sure if we are going to need more global loggers. Maybe it might be needed if we want to conveniently log into the same channel from different places... But then we could also defined several local loggers writing into the same channel

So yeah, maybe it could be a good idea to 
```
#define LOG(x) BOOST_LOG(m_##x##Logger)
```
and then have a convention to name loggers with `Logger` suffix

chfast: It might be more convenient to have a logger per module (1-4 files) not per class.

Also, we should probably have a global loggers for errors and warnings, right? Or it should be handled by module loggers with high severity?

eth/main.cpp
gumb0: Program options group is a good idea, though I imagine we still would explicitly call `setupLogging` in each executable, it will have more parameters for other filtering (at least by channel)

libethereum/Executive.cpp
gumb0: Logger is just an object used to write logs, and there's a concept of "channels" which represent log topics. And there can be for example several loggers writing to the same channel, or one logger writing to various channels etc.

So I think it will be usually more convenient to have a logger as a member inside one class, even if several classes write to the same channel
Sharing a logger is like sharing other data - either it's global (and might need thread-safety and proper fisco-free initialization), or there's a hassle of passing it around)
Maybe having a global logger inside unnamed namespace in cpp is fine. It's all flexible and we can try different things.

gumb0: As for errors and warnings - I think it's simpler to start with each channel having the same severity in all messages (like it's now) and separate channels for errors and warnings.
There are already `global cwarn`, `cnote`, `cdebug`, `clog` defined in `Log.h`, probably we'll remove some of these and add `cerror`

circle.yml
```diff
@@ -145,8 +145,8 @@ jobs:
       - CC:  gcc-6
       - GCOV: gcov-6
       - GENERATOR: Ninja
-      - BUILD_PARALLEL_JOBS: 4
-      - TEST_PARALLEL_JOBS: 4
```
chfast: This should not affect running tests in parallel

gumb0: Ok, I'll change this back

gumb0:
Two logging systems are not aware of each other so merging it now will result in some messages breaking each other in the middle

gumb0:
Rebased.

gumb0:
Added to the description the summary of how changes look like so far

gumb0:
@chfast Now log sometimes can look like
```
N X  12:39:25.586|p2p|f51ffdb5ΓÇª|Parity/v1.9.5-stable-ff821daf1-20180321/x86_64-linux-gnu/rustc1.24.1  Error reading - Abrupt peer disconnect: End of file
N X  12:39:26.220|p2p|14c55d31ΓÇª|Parity/v1.9.5-unstable-ff821da-20180321/x86_64-linux2018-04-17 12:39:27 e-th timer txExec 2820 ms
2018-04-17 12:39:27 eth timer class boost::multiprecision::number<struct boost::multiprecision::backends::cpp_int_backend<256,256,0,0,void>,0> __cdecl dev::eth::Block::enact(const struct dev::eth::VerifiedBlockRef &g,cnuo/rnusstt cc1l.a2s4s.1   Edrevr:o:re rteahd:i:nBgl o-c kACbhrauipnt  &p)e e2r8 3d1i smcso
n2n0e1c8t-:0 4E-n1d7  o1f2 :f3i9l:e27 eth note SLOW IMPORT: { "blockHash": "c138dc52ΓÇª", "blockNumber": 5123696, "gasPerSecond": 641220, "transactions": 51, "gasUsed": 1823571, "preliminaryChecks": 0.012457, "enactment": 2.84391, "collation": 0.00092, "writing": 0.001377, "checkBest": 0.006189, "total": 2.86506 }
```

That is, messages from two logging systems intermingled.

Maybe better to create separate PRs to this PR? (and rebase this one when needed)

chfast:
If there are some fixes needed here then go ahead and include them. But I would limit the migration to the new API.

gumb0:
I mean I don't want to fix this problem above, I want just to get rid of the old logging system, then it's fine.

gumb0:
@chfast So I've created a branch `log-refactor-all` to accumulate all changes
I'm ready to merge this PR into `log-refactor-all` (then I'll create cumulative PR from it)
Then the next PR will be https://github.com/ethereum/cpp-ethereum/pull/4962

gumb0:
Actually no, testeth output looks funny on AppVeyor, I'll try to fix it here first

gumb0:
Ok, I'll add it in one of the next PRs

PR #4860, opened 2018-02-26
AKA Aleth, Eth

jamesray1:


README.md
```diff
@@ -1,6 +1,6 @@
 # cpp-ethereum - Ethereum C++ client
 
-This repository contains [cpp-ethereum](http://cpp-ethereum.org), the [Ethereum](https://ethereum.org) C++ client.
+This repository contains [cpp-ethereum](http://cpp-ethereum.org), the [Ethereum](https://ethereum.org) C++ client, also known as Eth or Ceth.
```
chfast: I have never heard about Ceth.

jamesray1: No, it is more of a proposal to use Ceth to more readily distinguish it from ETH, the ticker for Ether, although I guess that in most contexts the two are readily distinguishable.

gumb0:
I don't like "Ceth" because of difficulties in pronunciation (is it "seth"? "keth"? "seeth", "keeth"?)

gumb0:
And "eth" is yeah, kind of too general, ambiguous with ETH

jamesray1:
OK, maybe let's brainstorm another name then.

How about Satya? It means truth in Sanskrit.

gumb0:
Ethane

jamesray1:
I'd prefer not to name it after a petrochemical. Ethallium is OK.

jamesray1:
Although technically ethane isn't a petrochemical, it is just produced from fossil fuels. https://en.wikipedia.org/wiki/Ethane

axic:
> A free pokemon from https://cheeaun.github.io/repokemon/

Haha, that's cool! There's none starting with `eth` or `et` though.

gumb0:
Also from chemistry
Ethanium
Ethylene
Ethione

Also Anisole (one of compounds known as Ethers)

Physics-inspired:
Apeiron (some unperceptible pre-matter ancient greeks beleived, somehow reminiscent of ether)
Nucleus
Kinesis (I like this one, but it's already taken by many people including Amazon and some token)

Astronomy-inspired:
Calypso (moon of Saturn)
Amalthea (moon of Jupiter)

Some pokemon names I liked:
Entei
Makuhita
Ampharos
Wurmple
Swirlix

Spanish-inspired:
Tenedor
Nodo
Acuerdo

Konrad Zuse-inspired:
Z1

ninabreznik:
If this client was really the first implementation, you should use this advantage and express this in the name. 
@serapath suggested Trueth, but something in this sense would be cool imho.

jamesray1:
Also you can change names around, e.g. Ethspring -> Springeth, which rolls off the tongue a bit more easily. I like Springeth, reminds one of vitality, and goes well with newness and being the first and oldest implementation.

chfast:
Consider these two options:

### Change the source package name, keep the prefix "eth"

For this the source package name should begin (or at least end) with "eth". E.g. "ethquake", "planeth".

- `eth` stays `eth` or becomes `ethquake` / `planeth`.
- `ethvm` -> `eth-vm`
- `ethkey` -> `eth-key`

### Change the source package name and the all tool names

Example: "entei"

- `eth` -> `entei` or `entei-client`
- `ethvm` -> entei-vm`
- `ethkey` -> entei-key`

### End notes

The `testeth` probably should keep its name.

The nice thing about having naming convention of `namespace-command` is you can use git-style CLI
sub-command. If a git command X is not built it, it executes `git-X` external executable. This is what makes `git clang-format` and `git submodule` work, both implemented as shell scripts.

Just for convenience, we could have e.g. `entei rpcproxy start` implemented in Python and `entei console` implemented in JS using ethereum-console project.

jamesray1:
For a name for the first client, other options are alpha-eth, eth-alpha and primeth ([prim- or prime means first in Latin](https://www.wordhippo.com/what-is/the/latin-word-for-e0996a37c13d44c3b06074939d43fa3759bd32c1.html) and consequently in many Romance languages). For primeth we could use the Primeape Pokémon as a mascot.

chfast:
> other options are alpha-eth, eth-alpha

Hey, maybe we reuse the old name "aleth"?

gumb0:
Maybe just Prime.

Aleth also not bad.

chfast:
The "prime" stands also for a prime number is not very unique. You can find quite some software project with that name.

jamesray1:
I edited the PR to re-add Aleth with a short explanation.

gumb0:
The change to the name was merged in other PRs

Issue #4773, opened 2018-01-16
Node discovery problem: "timestamp in the past"


gumb0:
After running sync with the main net for some time, it looses all peers, can't get any new ones and all I see in the log is a sequence of messages like
```
!P!  09:42:40 AM.868|p2p  Invalid packet (timestamp in the past) from 77.245.213.14 : 1056
!P!  09:42:41 AM.152|p2p  Invalid packet (timestamp in the past) from 115.231.148.29 : 15135
!P!  09:42:42 AM.659|p2p  Invalid packet (timestamp in the past) from 115.231.148.29 : 15135
!P!  09:42:42 AM.802|p2p  Invalid packet (timestamp in the past) from 185.178.192.139 : 30303
```

Workaround that helps: kill `~\.ethereum\network.rlp` which stores the list of previously connected peers and restart `eth`

gumb0:
Could it be because of invalid system time on the machine?
But syncing time from ntp with these commands doesn't help:
```
sudo service ntp stop
sudo ntpd -gq
sudo service ntp start
```

PR #4861, opened 2018-02-27
Fix discovery issues

gumb0:
Fixes https://github.com/ethereum/cpp-ethereum/issues/4773

It was not really a discovery problem, more like a problem of managing peers and in `Host` class, see comment below.

Also unnecessary mutex `x_pendingPeerConns` removed.

### DIFF
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -1,18 +1,18 @@
 /*
-	This file is part of cpp-ethereum.
+    This file is part of cpp-ethereum.
 
-	cpp-ethereum is free software: you can redistribute it and/or modify
-	it under the terms of the GNU General Public License as published by
-	the Free Software Foundation, either version 3 of the License, or
-	(at your option) any later version.
+    cpp-ethereum is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
 
-	cpp-ethereum is distributed in the hope that it will be useful,
-	but WITHOUT ANY WARRANTY; without even the implied warranty of
-	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-	GNU General Public License for more details.
+    cpp-ethereum is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


libp2p/Host.cpp
```diff
@@ -60,930 +60,938 @@ ReputationManager::ReputationManager()
...
-	
-	if (!m_netPrefs.pin)
-	{
-		unsigned pendingCount = 0;
-		DEV_GUARDED(x_pendingNodeConns)
-			pendingCount = m_pendingPeerConns.size();
-		int openSlots = m_idealPeerCount - peerCount() - pendingCount + reqConn;
-		if (openSlots > 0)
-			for (auto p: toConnect)
-				if (p->peerType == PeerType::Optional && openSlots--)
```
gumb0: Here I guess the intention was to break the loop when we run out of `openSlots`. What was really happening is of course `openSlots` going negative when there are a lot of peers in `toConnect` available and it tried to connect to all of them at once.

Then `connect()` added all these peers to `m_pendingPeerConns` (the set of peers we are trying to connect to) and this variable affects `peerSlotsAvailable()`, so when the handshake was complete with any peer it checked `peerSlotsAvailablr()` (in `startPeerSession()`) and it always looked as no available slots and so it disconnected.

gumb0:
Could also affect `Too many open files error` as sometimes it tried to open a lot of sockets at once.

libp2p/Host.cpp
```diff
@@ -60,930 +60,926 @@ ReputationManager::ReputationManager()
...
+        }
+    }
+
+    for (auto p: toConnect)
+        if (p->peerType == PeerType::Required && reqConn++ < m_idealPeerCount)
+            connect(p);
+    
+    if (!m_netPrefs.pin)
+    {
+        int openSlots = m_idealPeerCount - peerCount() - m_pendingPeerConns.size() + reqConn;
```
chfast: Are there any type conversions happening here? What are the chances of overflow?

gumb0: Hmm, all the values on the right of assignment are `unsigned`
Not sure now if it's  safe...
realistically none of these values go anywhere near `max<int>()`

gumb0: I will better rewrite it without cast to `int`

gumb0:
Thanks for the link 👌

PR #4863, opened 2018-02-28
Add tests for ethash light cache

chfast:
I need some reference tests for ethash library. The expected hashes are taken from this implementation, so not very useful here.


chfast:
Why?

pirapira:
I had forgotten what `operator==(std::string, char*)` does.

Issue #4845, opened 2018-02-13
Improve database error reporting


gumb0:
After DB layer refactoring when exception happens while working with the database, all I can see in the log is usually something like
```
  ✘  04:32:07 PM.455|eth  Error writing to state database: std::exception
```

PR #4864, opened 2018-02-28
Improve exceptions

gumb0:
Fixes https://github.com/ethereum/cpp-ethereum/issues/4845

Having a `std::string` member in an exception class goes against boost recommendations (see http://www.boost.org/doc/libs/1_39_0/libs/exception/doc/exception_types_as_simple_semantic_tags.html and http://www.boost.org/community/error_handling.html)
and leads to weird behaviour.
(`boost::enable_current_exception()` copies the exception object provided to `BOOST_THROW_EXCEPTION` macro, but this string member is not copied; `boost::enable_current_exception()` actually requires copy constructor to never throw)

Also there's no point in storing additional error string as a `std::string` member because `error_info` mechanism of boost exceptions can and should be used to store any additional info about the error.

First commit here was changing `Exception` constructor to store the string in `error_info` instead of a member variable.
Then I decided to get rid of all exception constructors taking a parameter string and use the approach more natural to boost execption with using `operator<<` at the point of throw.

### DIFF
--- a/libdevcore/Exceptions.h
+++ b/libdevcore/Exceptions.h
@@ -33,15 +33,26 @@
 
 namespace dev
 {
+// error information to be added to exceptions
+using errinfo_invalidSymbol = boost::error_info<struct tag_invalidSymbol, char>;
+using errinfo_wrongAddress = boost::error_info<struct tag_address, std::string>;
+using errinfo_comment = boost::error_info<struct tag_comment, std::string>;
+using errinfo_required = boost::error_info<struct tag_required, bigint>;
+using errinfo_got = boost::error_info<struct tag_got, bigint>;
+using errinfo_min = boost::error_info<struct tag_min, bigint>;
+using errinfo_max = boost::error_info<struct tag_max, bigint>;
+using RequirementError = boost::tuple<errinfo_required, errinfo_got>;
+using errinfo_hash256 = boost::error_info<struct tag_hash, h256>;
+using errinfo_required_h256 = boost::error_info<struct tag_required_h256, h256>;
+using errinfo_got_h256 = boost::error_info<struct tag_get_h256, h256>;
+using Hash256RequirementError = boost::tuple<errinfo_required_h256, errinfo_got_h256>;
+using errinfo_extraData = boost::error_info<struct tag_extraData, bytes>;
 
 /// Base class for all exceptions.
 struct Exception: virtual std::exception, virtual boost::exception
 {
-	Exception(std::string _message = std::string()): m_message(std::move(_message)) {}
-	const char* what() const noexcept override { return m_message.empty() ? std::exception::what() : m_message.c_str(); }
-
-private:
-	std::string m_message;
+    Exception() {}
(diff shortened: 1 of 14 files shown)


chfast:
I finally realized  what I don't like about the Boost approach -- it's too "dynamic".

1. Throwing an exception is quite complex exception now, quite a lot of boilerplate.
2. It's hard to know what an exception contains. I cannot jump to the exception declaration and inspect it fields. At runtime in can contain multiple error_infos but it's hard to tell if any and where they come from.

If we need to add additional data on the fly, can we do it like this?

```
catch(Exception& ex)
{
    ex.addComment("comment");
    throw;
}
```

gumb0:
What you suggest is even more dynamic, you couldn't then easily check whether the data of specific type is present in `Exception`. And what if I want to analyze an integer error code in `catch` ?

gumb0:
Boilerplate - on the one hand yes, on the other hand something like
```
        BOOST_THROW_EXCEPTION(FailedToOpenDB() << errinfo_db(status.ToString()) << errinfo_path(_path));
```
is less bolierplate comparing to declaring special exception class with error code and path...

gumb0:
In general this dynamic approach seems unusual to me, too

Once I worked with a more static approach to this, something like
```
class Exception
{
public:
  Exception() {}
  Exception(Exception const& _suberror) {}
};
...
catch(LowerLevelException& ex)
{
  throw HigherLevelException(ex);
}
```
`LowerLevelException` and `HigherLevelException` could have some additional data members specific to them.
And if you throw with a macro you could even save the file position of each throw along the stack.

gumb0:
Anyway in this PR I'm fixing the bug in the current approach.

If it's too ugly, look at the first commit, it's very simple, it keeps `std::string` as a constructor argument and it works, too.

gumb0:
> Maybe some error messages will become worse after this change, because outputting ex.what() will show only the exception name now (it's better to output boost::diagnostic_information(ex) in this case)

Fixed this by returning `boost::diagnostic_information` from `what`

libdevcore/Exceptions.h
```diff
@@ -21,34 +21,38 @@
...
-	Exception(std::string _message = std::string()): m_message(std::move(_message)) {}
-	const char* what() const noexcept override { return m_message.empty() ? std::exception::what() : m_message.c_str(); }
-
-private:
-	std::string m_message;
+    const char* what() const noexcept override { return boost::diagnostic_information_what(*this); }
 };
 
-#define DEV_SIMPLE_EXCEPTION(X) struct X: virtual Exception { const char* what() const noexcept override { return #X; } }
+#define DEV_SIMPLE_EXCEPTION(X)  \
```
gumb0: We could live without these macros now, actually

gumb0:
Boost actually advertises this dynamic nature as the strength of boost::exception
> Deriving from boost::exception effectively decouples the semantics of a failure from the information that is relevant to each individual instance of reporting a failure with a given semantic.
>
> In other words: with boost::exception, what data a given exception object transports depends primarily on the context in which failures are reported (not on its type.)

Issue #4815, opened 2018-01-27
Support libscrypt upstream


axic:
Upstream is: https://github.com/technion/libscrypt

Seems like it was kind of abandoned in 2015, but recently accepting commits. cpp-ethereum's import is from September 2015, there were only 3 commits since.

If upstream is revived, it could be built in CMake by downloading a version.

axic:
Seems like @chfast already started this: https://github.com/technion/libscrypt/pull/43

PR #4873, opened 2018-03-06
Use libscrypt from Hunter

chfast:
I fixed 2 tests, disabled 1.

The disabled test tests keyfile v2. I'm not sure what the expected result should be. I've found this commit adding the compatibility with v2: https://github.com/ethereum/cpp-ethereum/commit/8399161a8e9b4f0cf3a38e14f5c83e566aad21c2 but it does not include the test. I cannot find where the test was added.

Resolves https://github.com/ethereum/cpp-ethereum/issues/4815.


axic:
```

The following tests FAILED:
	  2 - RandomCodeTests/rndStateTest (SEGFAULT)
```

CMakeLists.txt
```diff
@@ -155,6 +158,3 @@ if (TESTS)
     enable_testing()
     add_subdirectory(test)
 endif()
-
-# TODO: Split out json_spirit, libscrypt and sec256k1.
-add_subdirectory(utils)
```
axic: So its not even needed for `json_spirit`?

chfast: We use json_spirit by loading headers only. This is an old CMake script that was not used.

chfast:
This test is failing from time to time only on CircleCI / macOS. To be disabled, because we cannot find the reason.

Issue #4585, opened 2017-10-06
Assertion failure in TrieDB


chfast:
```
 ⚡   17:24:10.187|eth  txExec 5865 ms
 ⚡   17:24:10.187|eth  dev::u256 dev::eth::Block::enact(const dev::eth::VerifiedBlockRef&, const dev::eth::BlockChain&) 5865 ms
  ℹ  17:24:10.190|eth  SLOW IMPORT: { "transactions": 87, "blockNumber": 4337888, "gasUsed": 6656766, "gasPerSecond": 1134498, "blockHash": "61861b16…", "writing": 0.000142, "total": 5.86897, "collation": 6.2e-05, "preliminaryChecks": 0.000354, "checkBest": 0.00082, "enactment": 5.86759  }
 ⚡   17:24:10.190|eth  Block import 4337888 5869 ms
eth: /home/chfast/Projects/ethereum/cpp-ethereum/libdevcore/../libdevcore/TrieDB.h:923: void dev::GenericTrieDB<_DB>::mergeAtAux(dev::RLPStream&, const dev::RLP&, dev::NibbleSlice, dev::bytesConstRef) [with _DB = dev::OverlayDB; dev::bytesConstRef = dev::vector_ref<const unsigned char>]: Assertion `!r.isNull()' failed.

Thread 14 "eth" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffe91b5700 (LWP 11753)]
0x00007ffff6907428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
54	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff6907428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff690902a in __GI_abort () at abort.c:89
#2  0x00007ffff68ffbd7 in __assert_fail_base (fmt=<optimized out>, assertion=assertion@entry=0x152f54a "!r.isNull()", 
    file=file@entry=0x1423c60 "/home/chfast/Projects/ethereum/cpp-ethereum/libdevcore/../libdevcore/TrieDB.h", line=line@entry=923, 
[...]

Issue #4584, opened 2017-10-06
eth crashed during mainnet sync


chfast:
```
eth: /home/chfast/Projects/ethereum/cpp-ethereum/libethereum/Account.h:160: void dev::eth::Account::noteCode(dev::bytesConstRef): Assertion `sha3(_code) == m_codeHash' failed.
[1]    10789 abort (core dumped)  eth/eth
```

gumb0:
I got this, too
```
#0  0x00007ffff6ba2c37 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff6ba6028 in __GI_abort () at abort.c:89
#2  0x00007ffff6b9bbf6 in __assert_fail_base (fmt=0x7ffff6cec3b8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x940c04 "sha3(_code) == m_codeHash",
    file=file@entry=0x85f3d0 "/home/ethnode/cpp-ethereum/libethereum/Account.h", line=line@entry=160,
    function=function@entry=0x98ad40 <dev::eth::Account::noteCode(dev::vector_ref<unsigned char const>)::__PRETTY_FUNCTION__> "void dev::eth::Account::noteCode(dev::bytesConstRef)") at assert.c:92
#3  0x00007ffff6b9bca2 in __GI___assert_fail (assertion=0x940c04 "sha3(_code) == m_codeHash", file=0x85f3d0 "/home/ethnode/cpp-ethereum/libethereum/Account.h", line=160,
    function=0x98ad40 <dev::eth::Account::noteCode(dev::vector_ref<unsigned char const>)::__PRETTY_FUNCTION__> "void dev::eth::Account::noteCode(dev::bytesConstRef)") at assert.c:101
#4  0x00000000005624fb in noteCode (_code=..., this=0x7fffe3cce480) at /home/ethnode/cpp-ethereum/libethereum/Account.h:160
#5  dev::eth::State::code (this=<optimized out>, _addr=...) at /home/ethnode/cpp-ethereum/libethereum/State.cpp:463
#6  0x00000000005397bb in dev::eth::Executive::call (this=this@entry=0x7fffee692060, _p=..., _gasPrice=..., _origin=...) at /home/ethnode/cpp-ethereum/libethereum/Executive.cpp:300
[...]

Issue #4578, opened 2017-10-05
Asserts failures in TrieDB.h during sync


gumb0:
Sometimes I see asserts like 
```
eth: /home/ethnode/cpp-ethereum/libdevcore/../libdevcore/TrieDB.h:971: dev::bytes dev::GenericTrieDB<_DB>::deleteAt(const dev::RLP&, dev::NibbleSlice) [with _DB = dev::OverlayDB; dev::bytes = std::vector<unsigned char>]: Assertion `_orig.isList() && (_orig.itemCount() == 2 || _orig.itemCount() == 17)' failed.
```

chfast:
See also #4585

sungjik:
I'm also getting the TrieDB.h:971 assert failure.

PR #4876, opened 2018-03-07
Further improve DB exceptions

gumb0:
This should improve the situation with lower-level DB errors being ignored during lookup/insert, leading to confusing assert failures like https://github.com/ethereum/cpp-ethereum/issues/4578 https://github.com/ethereum/cpp-ethereum/issues/4585 https://github.com/ethereum/cpp-ethereum/issues/4584

### DIFF
--- a/libdevcore/LevelDB.cpp
+++ b/libdevcore/LevelDB.cpp
@@ -29,6 +29,32 @@ inline leveldb::Slice toLDBSlice(Slice _slice)
     return leveldb::Slice(_slice.data(), _slice.size());
 }
 
+template <class _T>
+void throwDatabaseError(leveldb::Status const& _status, boost::filesystem::path const& _path)
+{
+    _T ex;
+    ex << errinfo_db(_status.ToString());
+    if (!_path.empty())
+        ex << errinfo_path(_path.string());
+
+    BOOST_THROW_EXCEPTION(ex);
+}
+
+void checkStatus(leveldb::Status const& _status, boost::filesystem::path const& _path = {})
+{
+    if (_status.ok())
+        return;
+
+    if (_status.IsCorruption())
+        throwDatabaseError<DBCorruption>(_status, _path);
+    else if (_status.IsIOError())
+        throwDatabaseError<IOError>(_status, _path);
+    else if (_status.IsNotFound())
+        throwDatabaseError<NotFound>(_status, _path);
+    else
+        throwDatabaseError<DatabaseError>(_status, _path);
+}
+
 class LevelDBWriteBatch : public WriteBatchFace
 {
 public:
@@ -76,20 +102,10 @@ LevelDB::LevelDB(boost::filesystem::path const& _path, leveldb::ReadOptions _rea
     leveldb::WriteOptions _writeOptions, leveldb::Options _dbOptions)
   : m_db(nullptr), m_readOptions(std::move(_readOptions)), m_writeOptions(std::move(_writeOptions))
 {
-    std::string const path = _path.string();
     auto db = static_cast<leveldb::DB*>(nullptr);
-    auto const status = leveldb::DB::Open(_dbOptions, path, &db);
-    if (!status.ok())
-    {
-        std::string const statusStr = status.ToString();
+    auto const status = leveldb::DB::Open(_dbOptions, _path.string(), &db);
+    checkStatus(status, _path);
 
-        if (status.IsCorruption())
-            BOOST_THROW_EXCEPTION(DBCorruption() << errinfo_db(statusStr) << errinfo_path(path));
-        else if (status.IsIOError())
-            BOOST_THROW_EXCEPTION(DBIOError() << errinfo_db(statusStr) << errinfo_path(path));
-        else
-            BOOST_THROW_EXCEPTION(FailedToOpenDB() << errinfo_db(statusStr) << errinfo_path(path));
-    }
     assert(db);
     m_db.reset(db);
 }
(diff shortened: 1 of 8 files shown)


libdevcore/db.h
```diff
@@ -70,14 +70,16 @@ class DatabaseFace
     virtual void forEach(std::function<bool(Slice, Slice)> f) const = 0;
 };
 
-DEV_SIMPLE_EXCEPTION(FailedToOpenDB);
```
gumb0: I think the purpose of having different exception classes is being able to handle them differently on the calling side. 

What we're really interested in here is what kind of error was returned from DB - corruption or IO error or not found or other, and not which method was called.

libethereum/BlockChain.cpp
```diff
@@ -298,7 +299,7 @@ unsigned BlockChain::open(fs::path const& _path, WithExisting _we)
         auto const l = m_extrasDB->lookup(db::Slice("best"));
         m_lastBlockHash = h256(l, h256::FromBinary);
     }
-    catch (db::FailedLookupInDB const& /* ex */)
```
gumb0: Changes like these might lead to more exceptions being unhandled, but if it's low-level IO error or DB corruption, we can't do much but exit anyway I guess

libdevcore/LevelDB.cpp
```diff
@@ -29,6 +29,32 @@ inline leveldb::Slice toLDBSlice(Slice _slice)
     return leveldb::Slice(_slice.data(), _slice.size());
 }
 
+template <class _T>
+void throwDatabaseError(leveldb::Status const& _status, boost::filesystem::path const& _path)
```
chfast: I don't see reason for this helper function. Do this preprocessing in checkStatus().

libdevcore/OverlayDB.cpp
```diff
@@ -129,8 +129,16 @@ void OverlayDB::rollback()
 std::string OverlayDB::lookup(h256 const& _h) const
 {
     std::string ret = MemoryDB::lookup(_h);
-    if (ret.empty() && m_db)
-        DEV_IGNORE_EXCEPTIONS(ret = m_db->lookup(toSlice(_h)));
+    if (!ret.empty() || !m_db)
```
chfast: Suspicious: when the `m_db` can be null? Maybe it's too defensive.

libdevcore/OverlayDB.cpp
```diff
@@ -129,8 +129,16 @@ void OverlayDB::rollback()
...
+        return ret;
+
+    try
+    {
+        ret = m_db->lookup(toSlice(_h));
+    }
+    catch (db::NotFound const&)
```
chfast: Is there any place where this exception is useful? Maybe lookup should return just an empty output (or optional).
Also, not finding an entry in a database is not an exceptional case (especially when the method is called "lookup", not "get").

libdevcore/db.h
chfast: I agree here. This is good improvement. However, I'd also like to see how the code would look like if we move the "status" from type to error_info, i.e. having `DatabaseError() << errinfo_status(status)`.

libdevcore/db.h
```diff
@@ -70,14 +70,16 @@ class DatabaseFace
...
-DEV_SIMPLE_EXCEPTION(DBCorruption);
-DEV_SIMPLE_EXCEPTION(DBIOError);
+DEV_SIMPLE_EXCEPTION(DatabaseError);
+struct DBCorruption : virtual DatabaseError
+{
+};
+struct IOError : virtual DatabaseError
+{
+};
+struct NotFound : virtual DatabaseError
```
chfast: By following Java convention: Errors are fatal, Exceptions not. So I'd suggest renaming it to:

- DatabaseError -> DatabaseException (because NotFound is part of it)
- DBCorruption -> (DB)CorruptionError

libethereum/BlockChain.cpp
chfast: See comment above: Errors should be treated as fatal, reported nicely and exit.

libethereum/BlockChain.cpp
```diff
@@ -298,7 +299,7 @@ unsigned BlockChain::open(fs::path const& _path, WithExisting _we)
...
+    catch (db::NotFound const& /* ex */)
```
chfast: The comment `/* ex */` rather unnecessary.

libdevcore/db.h
gumb0: Hm, this would break "is-a" relationship between `CorruptionError` and `DatabaseException`

libdevcore/OverlayDB.cpp
gumb0: I think I'll try to get rid of `NotFound` exception

libdevcore/db.h
gumb0: After getting rid of `NotFound` exception, I see that we have only one place now, where we need to distinguish DB error type.
So I think I will change the code now the way you suggest here.

chfast: Let me know when the next review is needed.

gumb0:
There were test failures on AppVeyor, they all seem expected + `TransientDirectory` error that happens sometimes
I've pushed a commit that removes some unnecessary error output from test log (the error `what()` is now more verbose and looked more scary) and added messages to the tests that are marked as expected failure.

chfast:
I don't understand what I've changed. I can only find additional `cout` line besides tabs to spaces conversions.

gumb0:
@chfast Just two additional `cout`s to make log clearer and this removal which output confusing messages https://github.com/ethereum/cpp-ethereum/pull/4876/files#diff-c2418ef721397320904b1e59230a1a16L241

chfast:
ok

gumb0:
AppVeyor timed out now, I will merge anyway

PR #4883, opened 2018-03-12
CMake: Add option to disable VM optimizations

### LINKED ISSUE #4475: feature request: cmake flag to set EVM_OPTIMIZE=false
This is needed to support standardized tracing.  Without a cmake flag, the code needs to be patched as done [here](https://github.com/cdetrio/cpp-ethereum/commit/7154c17542034031bcb56c9fb90089c2d4139428#diff-11510de5169187b42a8e4bb8ae298582R64).

See also  https://github.com/ethereum/tests/issues/249

chfast:
Fixes #4475.


chfast:
@pirapira No, because it is enabled by default. People from Hive and fuzzing are asking for it to have correct VM traces.

I pushed additional commit removing some PARANOIA related dead code if you mind to take a last look.

pirapira:
@chfast now I see it's `ON` by default.

pirapira:
`PARANOIA` removals look good too.

PR #4885, opened 2018-03-12
Remove rndCode test

chfast:


chfast:
It's not gone, just hidden.

pirapira:
I'm not even sure if the problem was in the test or in the tested code.

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -620,21 +620,6 @@ int RandomCodeOptions::getWeightedRandomOpcode() const
...
-		std::string code = test::RandomCode::get().generate(1000, options);
-		BOOST_REQUIRE(!code.empty());
-	}
-	catch(dev::Exception const& _e)
-	{
-		BOOST_ERROR("Exception thrown when generating random code! " + diagnostic_information(_e));
-	}
-}
-
 BOOST_AUTO_TEST_CASE(rndStateTest)
```
winsvega: this test should be removed too

PR #4886, opened 2018-03-13
Delete destructor to avoid aggregate initialization

esteve:
Delete destructor so that VMFactory can't be instantiated via aggregate initialization (e.g. `VMFactory f{}` is valid)


esteve:
@chfast do you mind having a look at this? Thanks!

PR #4888, opened 2018-03-13
 evm: Pass the "is CREATE" information to EVM-C VM

chfast:
Main feature: pass information whatever a call is a CREATE to the EVM-C VMs.
This PR also contains a lot of code formatting and cleanups. Sorry for that, but that was not avoidable. You can review commit by commit.

@axic @hugo-dc

### DIFF
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -299,7 +299,9 @@ bool Executive::call(CallParameters const& _p, u256 const& _gasPrice, Address co
         {
             bytes const& c = m_s.code(_p.codeAddress);
             h256 codeHash = m_s.codeHash(_p.codeAddress);
-            m_ext = make_shared<ExtVM>(m_s, m_envInfo, m_sealEngine, _p.receiveAddress, _p.senderAddress, _origin, _p.apparentValue, _gasPrice, _p.data, &c, codeHash, m_depth, _p.staticCall);
+            m_ext = make_shared<ExtVM>(m_s, m_envInfo, m_sealEngine, _p.receiveAddress,
+                _p.senderAddress, _origin, _p.apparentValue, _gasPrice, _p.data, &c, codeHash,
+                m_depth, false, _p.staticCall);
         }
     }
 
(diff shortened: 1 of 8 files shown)


libethereum/Executive.cpp
```diff
@@ -362,7 +364,8 @@ bool Executive::executeCreate(Address const& _sender, u256 const& _endowment, u2
 
     // Schedule _init execution if not empty.
     if (!_init.empty())
-        m_ext = make_shared<ExtVM>(m_s, m_envInfo, m_sealEngine, m_newAddress, _sender, _origin, _endowment, _gasPrice, bytesConstRef(), _init, sha3(_init), m_depth);
+        m_ext = make_shared<ExtVM>(m_s, m_envInfo, m_sealEngine, m_newAddress, _sender, _origin,
```
axic: Maybe format it the same way as above? Line break after `newAddress`.

libevm/EVMC.h
```diff
@@ -70,14 +70,13 @@ class EVM
     Result execute(ExtVMFace& _ext, int64_t gas)
     {
         auto mode = toRevision(_ext.evmSchedule());
+        evm_call_kind kind = _ext.isCreate ? EVM_CREATE : EVM_CALL;
         uint32_t flags = _ext.staticCall ? EVM_STATIC : 0;
```
axic: Should we make any assumptions for evmc here that `isCreate` and `staticCall` is mutually exclusive?

libethereum/Executive.cpp
chfast: I can't. There is only single correct way of formatting code :) At least that's how clang-format works.

libevm/EVMC.h
chfast: I'd rather not. If we want that, the good way of doing it, would be to extend the "call kind" with some additional kinds...

axic: Either new kinds or an assert here.

libevm/EVMC.h
```diff
@@ -70,14 +70,14 @@ class EVM
...
-        evm_message msg = {toEvmC(_ext.myAddress), toEvmC(_ext.caller),
-                           toEvmC(_ext.value), _ext.data.data(),
-                           _ext.data.size(), toEvmC(_ext.codeHash), gas,
-                           static_cast<int32_t>(_ext.depth), EVM_CALL, flags};
-        return Result{m_instance->execute(
-            m_instance, &_ext, mode, &msg, _ext.code.data(), _ext.code.size()
-        )};
+        assert(flags != EVM_STATIC || kind == EVM_CALL);  // STATIC implies a CALL.
```
chfast: Assert added here.

PR #4891, opened 2018-03-14
Hera dynamic loading prototype

esteve:
This is a change in anticipation for implementing dynamic loading of loading VMs.


esteve:
@axic @chfast I noticed there's a mix of tabs and spaces in `VMFactory.h`, what's the consensus?

chfast:
Spaces. You can use clang-format to format changes. You can also convert tabs to spaces in whole file if that will not mess up the PR a lot.

esteve:
> I actually prefer to keep it as the array unless there are bigger reasons for it.
> 
> The array does not require dynamic initialization and is only needed once when processing command line options.

My goal is to be able to add the options dynamically based on the available VMs, instead of it being compile-time, see:

https://github.com/esteve/cpp-ethereum/compare/vmfactory-initializer...esteve:dl-evm#diff-7c3f9da96ed44d254ef72736165afc91

should all the VM still be available on the command line and fail if they don't exist as shared libraries instead?

chfast:
I would start slow. Let's have a fixed list for now and check if hey are available at build time or dynamically (not sure how to do this exactly). For dynamic loading we can have a simple mapping name => libname.so.

chfast:
@esteve Maybe it will be easier if you publish full set of changes. Then we can decide if we want to split into smaller patches.

esteve:
@chfast done, I've pushed the rest of the commits from `dl-evm` to this branch.

cmake/toolchain.cmake
```diff
@@ -1,4 +1,5 @@
 # Require C++11.
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
-set(CMAKE_CXX_EXTENSIONS Off)
\ No newline at end of file
+set(CMAKE_CXX_EXTENSIONS Off)
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
chfast: Why I don't understand why it is needed here. Shouldn't it be only enabled in Hera?

esteve: The dependencies are built statically without `-fPIC`, so when hera is built as a shared library and is linked against for example `libboost_system-mt.a`, the linker fails with:

`relocation R_X86_64_32S against '.rodata' can not be used when making a shared object; `

esteve: I've submitted https://github.com/ewasm/hera/pull/162 instead which adds `CMAKE_POSITION_INDEPENDENT_CODE` to Hera only.

chfast:
Yes. Sorry @esteve I did not used any of your commits. I tried but a lot of tweaks were needed here and then and it was much easier to do it in small steps in bottom-up manner.

esteve:
@chfast no worries :-) I hope the changes here were useful. Thanks for taking care of it, #4988 looks great!

PR #4895, opened 2018-03-16
set DefaultGasPrice to 0

winsvega:
small PRs to cpp-ethereum from testrpc branch

### DIFF
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -239,7 +239,7 @@ int main(int argc, char** argv)
 //  TransactionPriority priority = TransactionPriority::Medium;
 //  double etherPrice = 30.679;
 //  double blockFees = 15.0;
-    u256 askPrice = DefaultGasPrice;
+    u256 askPrice = 0;
     u256 bidPrice = DefaultGasPrice;
     bool alwaysConfirm = true;
 
@@ -1062,8 +1062,8 @@ int main(int argc, char** argv)
 
     web3.setIdealPeerCount(peers);
     web3.setPeerStretch(peerStretch);
-//  std::shared_ptr<eth::BasicGasPricer> gasPricer = make_shared<eth::BasicGasPricer>(u256(double(ether / 1000) / etherPrice), u256(blockFees * 1000));
-    std::shared_ptr<eth::TrivialGasPricer> gasPricer = make_shared<eth::TrivialGasPricer>(askPrice, bidPrice);
+    std::shared_ptr<eth::TrivialGasPricer> gasPricer =
+        make_shared<eth::TrivialGasPricer>(askPrice, bidPrice);
     eth::Client* c = nodeMode == NodeMode::Full ? web3.ethereum() : nullptr;
     if (c)
     {


eth/main.cpp
```diff
@@ -1062,7 +1062,16 @@ int main(int argc, char** argv)
...
+
+    // allow any gasPrice in testing mode
+    if (testingMode)
+    {
+        askPrice = 0;
+        bidPrice = 0;
+    }
+    //  std::shared_ptr<eth::BasicGasPricer> gasPricer =
+    //  make_shared<eth::BasicGasPricer>(u256(double(ether / 1000) / etherPrice), u256(blockFees *
+    //  1000));
```
matteougolotti: Why the commented out code ? It may be better to remove it, if it's no longer useful.

winsvega: this commented line was there for ages. @pirapira @chfast  do you approve removal?

pirapira: Sure.

chfast: Yes, remove.

winsvega:
I think it was meant to reject transactions with low gasPrice, so the client auto select the most profitable transactions.

chfast:
So one price is default price for your own transactions? That should not affect testing at all.

The second price is the minimum tx price that is accepted? The set it to 1 by default. I would avoid tx with 0 gas price unless this is needed by retesteth. If needed, set to 0 by default then.

geth stopped using prices for this. They just have a tx queue limited in size, but they can accept tx with 0 price if there is space for it in the queue.

winsvega:
the current value is 
`static const u256 DefaultGasPrice = 20 * shannon;`
we could set it to 0, then no special if's for testing is needed.

chfast:
I'm for setting it to 0 by default.

@gumb0 Does it mean that eth will accept any transaction to its tx pool?

gumb0:
Looking at this https://github.com/ethereum/cpp-ethereum/blob/a24880812b7be8ceb96fa51cfad050c09fd95e97/libethereum/Block.cpp#L337

I think the logic is to mine the transactions not cheaper than ask price and to relay only transactions with price >= 0.9 * ask price.

If it's zero, we'll mine and relay all transactions.

chfast:
@gumb0 thanks for looking into it.

How about we rename the option `--ask-price` to whatever geth / parity has. And set the default value to 0.

So the "big" price is probably the default price for transactions sent by the user... I'm not sure what to do with that, but I believe you can always override it by setting the gas price for a transaction explicitly. So this param should not affect RPC testing.

libethereum/GasPricer.h
```diff
@@ -40,7 +40,7 @@ enum class TransactionPriority
 	Highest = 8
 };
 
-static const u256 DefaultGasPrice = 20 * shannon;
+static const u256 DefaultGasPrice = 0;
```
chfast: I'm not sure it make sense to have it now. It's only used in 2 places.

winsvega:
is this `ClientTests/Personal` failure related to the change? @gumb0

libethereum/GasPricer.h
winsvega: so should we set the default value to 0 ?

chfast: Nevermind. It's used in more places.

gumb0:
It could be related, yes.
I guess we should set the gas price of transaction in the test now. Or I would prefer actually to leave the default bid price 20 shannon as earlier.

chfast:
Is 20 shannon == 20 Gwei?

gumb0:
@chfast Right

chfast:
I'm for keeping it for the bid price as well.

winsvega:
"I'm for keeping it for the bid price as well."
what that supposed to mean? 

So you want askPrice = 0, BidPrice = 20*shannon;  
?

pirapira:
(maybe irrelevant: usually in financial markets, `bid` is smaller than `ask` because `bid` is a price one is willing to buy at, and `ask` is a price one is willing to sell at.)

eth/main.cpp
```diff
@@ -1062,8 +1062,8 @@ int main(int argc, char** argv)
 
     web3.setIdealPeerCount(peers);
     web3.setPeerStretch(peerStretch);
-//  std::shared_ptr<eth::BasicGasPricer> gasPricer = make_shared<eth::BasicGasPricer>(u256(double(ether / 1000) / etherPrice), u256(blockFees * 1000));
-    std::shared_ptr<eth::TrivialGasPricer> gasPricer = make_shared<eth::TrivialGasPricer>(askPrice, bidPrice);
+    std::shared_ptr<eth::TrivialGasPricer> gasPricer =
+        make_shared<eth::TrivialGasPricer>(askPrice, testingMode ? 0 : bidPrice);
```
chfast: Why big price must be 0 in testing mode?

winsvega: checking it now. looks like it does not affect the tests.

chfast: It should not affect test transactions unless the transactions does not have gas price specified. In other words this is tx gas price in case user does not specify any.

winsvega:
windows stoped before the test finish
libp2p/p2pPeer/requirePeer fails on macOS

PR #4901, opened 2018-03-20
Add comments on selfdestruct(this)

pirapira:
I said I'd do this in a recent conversation with @chfast.  The code looks weird, and the behavior is also weird, but that seems to be the current consensus.  So it's worth some comments.

### DIFF
--- a/libethereum/ExtVM.cpp
+++ b/libethereum/ExtVM.cpp
@@ -141,8 +141,7 @@ void ExtVM::suicide(Address _a)
     // witnessing the current consensus
     // 'GeneralStateTests/stSystemOperationsTest/suicideSendEtherPostDeath.json'.
     m_s.addBalance(_a, m_s.balance(myAddress));
-    // Below sets the balance to zero even when _a == myAddress.
-    m_s.subBalance(myAddress, m_s.balance(myAddress));
+    m_s.setBalance(myAddress, 0);
     ExtVMFace::suicide(_a);
 }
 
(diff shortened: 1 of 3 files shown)


libethereum/ExtVM.cpp
```diff
@@ -136,10 +136,14 @@ std::pair<h160, owning_bytes_ref> ExtVM::create(u256 _endowment, u256& io_gas, b
...
-	m_s.addBalance(_a, m_s.balance(myAddress));
-	m_s.subBalance(myAddress, m_s.balance(myAddress));
-	ExtVMFace::suicide(_a);
+    // Why transfer is no used here? That caused a consensus issue before (see Quirk #2 in
+    // http://martin.swende.se/blog/Ethereum_quirks_and_vulns.html). There is one test case
+    // witnessing the current consensus
+    // 'GeneralStateTests/stSystemOperationsTest/suicideSendEtherPostDeath.json'.
+    m_s.addBalance(_a, m_s.balance(myAddress));
+    // Below sets the balance to zero even when _a == myAddress.
+    m_s.subBalance(myAddress, m_s.balance(myAddress));
```
chfast: Can we change this to `m_s.setBalance(myAddress, 0)` if such function exists?

pirapira: I think that's better.  The code speaks for itself and I don't need a comment.  I'll write `setBalance`.

chfast: Hm... if there is not setBalance() at the moment, it mean we will add another state change journal case. I'm not sure it's worth it. Please take a look there.

pirapira: `setBalance` will be a wrapper around `addBalance` so there won't be a new changelog journal (`subBalance` is already a wrapper around `addBalance`).

pirapira:
Oh, I wasted one push...

Issue #4899, opened 2018-03-19
'--jsontrace' should not cause the client to fallback on the interpreter when EVMC is enabled


jwasinger:
The behavior is confusing.  EVMC should take precedence over `--jsontrace`.

chfast:
This is about ethvm?

axic:
I think so.

Issue #4795, opened 2018-01-22
ethvm: Do not silently fall back to interpreter


chfast:
By default, ethvm ignores `--vm` flag and falls back to interpreter. This is super confusing. It's much better to terminate with an error.

PR #4903, opened 2018-03-20
 ethvm: Do not chenge VM selection depending on tracing

chfast:
Mostly: do not force Interpreter if VM tracing is required. Also cleanups some parts of ethvm.

Fixes https://github.com/ethereum/cpp-ethereum/issues/4899.
Fixes https://github.com/ethereum/cpp-ethereum/issues/4795.

### DIFF
--- a/ethvm/main.cpp
+++ b/ethvm/main.cpp
@@ -110,7 +110,6 @@ int main(int argc, char** argv)
     setDefaultOrCLocale();
     string inputFile;
     Mode mode = Mode::Statistics;
-    VMKind vmKind = VMKind::Legacy;
     State state(0);
     Address sender = Address(69);
     Address origin = Address(69);
@@ -316,19 +315,6 @@ int main(int argc, char** argv)
     unordered_map<byte, pair<unsigned, bigint>> counts;
     unsigned total = 0;
     bigint memTotal;
-    auto onOp = [&](uint64_t step, uint64_t PC, Instruction inst, bigint m, bigint gasCost,
-                    bigint gas, VMFace const* _vm, ExtVMFace const* extVM) {
-        if (mode == Mode::Statistics)
-        {
-            counts[(byte)inst].first++;
-            counts[(byte)inst].second += gasCost;
-            total++;
-            if (m > 0)
-                memTotal = m;
-        }
-        else if (mode == Mode::Trace)
-            st(step, PC, inst, m, gasCost, gas, _vm, extVM);
-    };
 
     executive.initialize(t);
     if (!code.empty())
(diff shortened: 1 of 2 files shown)


ethvm/main.cpp
```diff
@@ -345,23 +350,28 @@ int main(int argc, char** argv)
...
         cout << "Expensive operations:\n";
-        for (auto const& c: {Instruction::SSTORE, Instruction::SLOAD, Instruction::CALL, Instruction::CREATE, Instruction::CALLCODE, Instruction::DELEGATECALL, Instruction::MSTORE8, Instruction::MSTORE, Instruction::MLOAD, Instruction::SHA3})
-            if (!!counts[(byte)c].first)
-                cout << "  " << instructionInfo(c).name << " x " << counts[(byte)c].first << " (" << counts[(byte)c].second << " gas)\n";
+        for (auto const& c : {Instruction::SSTORE, Instruction::SLOAD, Instruction::CALL,
+                 Instruction::CREATE, Instruction::CALLCODE, Instruction::DELEGATECALL,
+                 Instruction::MSTORE8, Instruction::MSTORE, Instruction::MLOAD, Instruction::SHA3})
+            if (counts[(byte)c].first != 0)
+                cout << "  " << instructionInfo(c).name << " x " << counts[(byte)c].first << " ("
+                     << counts[(byte)c].second << " gas)\n";
```
axic: `static_cast` for `c`, and perhaps a local variable instead of 3 casts?

axic:
Should a warning be shown if `tracing` is selected but the VM ignores it?

ethvm/main.cpp
```diff
@@ -316,32 +315,30 @@ int main(int argc, char** argv)
...
 
+    OnOpFunc onOp;
+    if (mode == Mode::Statistics)
+    {
+        onOp = [&](uint64_t, uint64_t, Instruction inst, bigint m, bigint gasCost, bigint,
+                   VMFace const*, ExtVMFace const*) {
+            counts[(byte)inst].first++;
+            counts[(byte)inst].second += gasCost;
+            total++;
+            if (m > memTotal)
```
axic: This line was probably a bugfix worth a separate commit.

chfast:
> Should a warning be shown if tracing is selected but the VM ignores it?

This is hard to do at this point because this is constantly changing. E.g. Hera supports tracing but not through EVM-C.

axic:
> E.g. Hera supports tracing but not through EVM-C.

It actually doesn't support tracing, only pseudo tracing for evm2wasm compiled EVM code :)

Perhaps the future evm-c tracing API should be more flexible and/or have a kind/flavour option.

chfast:
Long term: we can report back if a VM supports tracing.
Short term: we can add some ifs and warnings, but I wouldn't do it until someone complains.

chfast:
Thanks for checking...

Issue #4882, opened 2018-03-12
Migrate Interpreter to EVM-C


chfast:
- [ ] Create a backup copy of the interpreter called "tracer". For some time, it will be the only VM implementation that supports tracing so that's the reason for the name.
- [ ] Migrate the interpreter to EVM-C. The migration is to be applied to the current code (not the copy) to keep the git history of this part.
- [ ] Implement tracing in EVM-C.
- [ ] Bring back support for tracing in the interpreter.
- [ ] Remove the tracer implementation.

axic:
How about writing it from scratch on EVM-C in a new repo?

chfast:
@axic That's different project and will take much more time. This here I need to better test EVM-C in geth and to simplify code in cpp-ethereum.

axic:
@chfast do you want to rename the files to match the header (`interpreter.h`) and possible move into its own directory?

PR #4905, opened 2018-03-21
Interpreter using EVM-C

chfast:
Finally we have the Interpreter using EVM-C interface. The old interface is available as `--vm legacy` and is the default one. The new is available as `--vm interpreter`.

I can split the last commit that does the essential changes into another PR if needed.

Part of #4882.

### DIFF
--- a/libevm/CMakeLists.txt
+++ b/libevm/CMakeLists.txt
@@ -1,7 +1,9 @@
 
 set(sources
+    EVMC.cpp EVMC.h
     ExtVMFace.cpp ExtVMFace.h
     Instruction.cpp Instruction.h
+    interpreter.h
     LegacyVM.cpp LegacyVM.h
     LegacyVMCalls.cpp
     LegacyVMOpt.cpp
@@ -14,12 +16,6 @@ set(sources
     VMValidate.cpp
     VMFactory.cpp VMFactory.h
 )
-
-if(EVMJIT OR HERA)
-    list(APPEND sources
-        EVMC.cpp EVMC.h
-    )
-endif()
 
 add_library(evm ${sources})
 
--- a/libevm/VM.cpp
+++ b/libevm/VM.cpp
@@ -157,16 +157,15 @@ void VM::updateSSGas()
     evm_uint256be rawValue;
     m_context->fn_table->get_storage(&rawValue, m_context, &m_message->destination, &key);
     u256 value = fromEvmC(rawValue);
-    if (!value && m_SP[1])
-        m_runGas = VMSchedule::sstoreSetGas;
-    else
-        m_runGas = toInt63(VMSchedule::sstoreResetGas);
+    m_runGas = (!value && m_SP[1]) ? VMSchedule::sstoreSetGas : VMSchedule::sstoreResetGas;
 }
 
 uint64_t VM::gasForMem(u512 _size)
 {
+    constexpr int64_t memoryGas = VMSchedule::memoryGas;
+    constexpr int64_t quadCoeffDiv = VMSchedule::quadCoeffDiv;
     u512 s = _size / 32;
-    return toInt63((u512) VMSchedule::memoryGas * s + s * s / VMSchedule::quadCoeffDiv);
+    return toInt63(memoryGas * s + s * s / quadCoeffDiv);
 }
 
 void VM::updateIOGas()
@@ -196,8 +195,9 @@ void VM::updateMem(uint64_t _newMem)
 void VM::logGasMem()
 {
     unsigned n = (unsigned) m_OP - (unsigned) Instruction::LOG0;
+    constexpr int64_t logDataGas = VMSchedule::logDataGas;
     m_runGas = toInt63(
-        VMSchedule::logGas + VMSchedule::logTopicGas * n + u512(VMSchedule::logDataGas) * m_SP[1]);
+        VMSchedule::logGas + VMSchedule::logTopicGas * n + logDataGas * u512(m_SP[1]));
     updateMem(memNeed(m_SP[0], m_SP[1]));
 }
 
@@ -208,7 +208,9 @@ void VM::fetchInstruction()
     adjustStack(metric.args, metric.ret);
 
     // FEES...
-    std::array<int64_t, 9> tierStepGas{{0, 2, 3, 5, 8, 10, 20, 0, 0}};
+    std::array<int64_t, 9> tierStepGas{
+        {VMSchedule::stepGas0, VMSchedule::stepGas1, VMSchedule::stepGas2, VMSchedule::stepGas3,
+            VMSchedule::stepGas4, VMSchedule::stepGas5, VMSchedule::stepGas6, 0, 0}};
     m_runGas = tierStepGas[static_cast<unsigned>(metric.gasPriceTier)];
     m_newMemSize = m_mem.size();
     m_copyMemSize = 0;
@@ -404,7 +406,9 @@ void VM::interpretCases()
         CASE(SHA3)
         {
             ON_OP();
-            m_runGas = toInt63(VMSchedule::sha3Gas + (u512(m_SP[1]) + 31) / 32 * VMSchedule::sha3WordGas);
+            constexpr int64_t sha3Gas = VMSchedule::sha3Gas;
+            constexpr int64_t sha3WordGas = VMSchedule::sha3WordGas;
+            m_runGas = toInt63(sha3Gas + (u512(m_SP[1]) + 31) / 32 * sha3WordGas);
             updateMem(memNeed(m_SP[0], m_SP[1]));
             updateIOGas();
 
@@ -1738,7 +1742,7 @@ void VM::interpretCases()
 
         CASE(JUMPDEST)
         {
-            m_runGas = 1;
+            m_runGas = VMSchedule::jumpdestGas;
             ON_OP();
             updateIOGas();
         }
--- a/libevm/VM.cpp
+++ b/libevm/VM.cpp
@@ -18,20 +18,16 @@
 #include "interpreter.h"
 #include "VM.h"
 
-using namespace std;
-using namespace dev;
-using namespace dev::eth;
-
 extern "C" evm_instance* interpreter_create() noexcept
 {
-    return new (std::nothrow) VM;
+    return new (std::nothrow) dev::eth::VM;
 }
 
 namespace
 {
 void destroy(evm_instance* _instance)
 {
-    delete static_cast<VM*>(_instance);
+    delete static_cast<dev::eth::VM*>(_instance);
 }
 
 void delete_output(const evm_result* result)
@@ -42,7 +38,7 @@ void delete_output(const evm_result* result)
 evm_result execute(evm_instance* _instance, evm_context* _context, evm_revision _rev,
     const evm_message* _msg, uint8_t const* _code, size_t _codeSize) noexcept
 {
-    auto vm = static_cast<VM*>(_instance);
+    auto vm = static_cast<dev::eth::VM*>(_instance);
     evm_result result = {};
     dev::eth::owning_bytes_ref output;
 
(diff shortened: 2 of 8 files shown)


libevm/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@
 set(sources
     ExtVMFace.cpp ExtVMFace.h
     Instruction.cpp Instruction.h
+    interpreter.h
```
axic: Lowercase?

libevm/VM.cpp
```diff
@@ -15,12 +15,34 @@
...
 */
 
+#include "interpreter.h"
 #include "VM.h"
 
 using namespace std;
 using namespace dev;
 using namespace dev::eth;
 
+evm_instance* interpreter_create()
```
axic: Should this be in `extern "C"` if the header says so?

libevm/VMFactory.cpp
```diff
@@ -133,7 +133,7 @@ po::options_description vmProgramOptions(unsigned _lineLength)
     add("vm",
         po::value<VMKind>()
             ->value_name("<name>")
-            ->default_value(VMKind::Interpreter, "interpreter")
+            ->default_value(VMKind::Legacy, "legacy")
```
axic: What is the legacy vm vs. the interpreter?

axic:
Please rebase after #4906.

libevm/CMakeLists.txt
chfast: Because there is no Interpreter class there and it follows the convention of EVM-C implementation headers: evmjit.h, hera.h, interpreter.h.

libevm/VM.cpp
chfast: Seems to be working on my machine :D

libevm/VMFactory.cpp
chfast: The same implementation but different interfaces. I keep the "legacy" one always have a working one and keep the vm tracing support.

chfast:
Two commits moved here: https://github.com/ethereum/cpp-ethereum/pull/4908.

libevm/VM.cpp
axic: I think the header may force it into that.

axic:
I'll review this once the others are merged and this one rebased.

libevm/CMakeLists.txt
axic: Ok, how about also creating a directory for it in that case?

chfast: I was thinking about this, but that requires another PR. Changing and moving files in the same time makes reviews horrible.

libevm/VM.cpp
```diff
@@ -119,22 +152,20 @@ void VM::adjustStack(unsigned _removed, unsigned _added)
...
-    if (!m_ext->store(m_SP[0]) && m_SP[1])
-        m_runGas = toInt63(m_schedule->sstoreSetGas);
-    else if (m_ext->store(m_SP[0]) && !m_SP[1])
-    {
-        m_runGas = toInt63(m_schedule->sstoreResetGas);
-        m_ext->sub.refunds += m_schedule->sstoreRefundGas;
-    }
+    evm_uint256be key = toEvmC(m_SP[0]);
+    evm_uint256be rawValue;
+    m_context->fn_table->get_storage(&rawValue, m_context, &m_message->destination, &key);
```
axic: When is `updateSSGas` called? Before applying the actual change? How was this correct before? And how is it correct now (it doesn't handle the refund case)?

libevm/VM.cpp
```diff
@@ -228,8 +263,8 @@ void VM::interpretCases()
         CASE(CREATE2)
         {
             ON_OP();
-            if (!m_schedule->haveCreate2)
-                throwBadInstruction();
+            // TODO: Bring back support for CREATE2.
```
axic: How about `m_rev < EVM_CONSTANTINOPLE` (if it is still planned for that)?

libevm/VM.cpp
```diff
@@ -296,22 +331,29 @@ void VM::interpretCases()
...
+            evm_uint256be rawBalance;
+            m_context->fn_table->get_balance(&rawBalance, m_context, &m_message->destination);
+            u256 balance = fromEvmC(rawBalance);
+            if (balance > 0 || m_rev < EVM_SPURIOUS_DRAGON)
+            {
                 // After EIP150 hard fork charge additional cost of sending
                 // ethers to non-existing account.
-                if (m_schedule->suicideChargesNewAccountGas() && !m_ext->exists(dest))
-                    m_runGas += m_schedule->callNewAccountGas;
+                int destinationExists =
```
axic: Isn't this a bool?

libevm/VM.cpp
```diff
@@ -374,72 +416,106 @@ void VM::interpretCases()
...
                 throwDisallowedStateChange();
 
             logGasMem();
             updateIOGas();
 
-            m_ext->log({m_SP[2]}, bytesConstRef(m_mem.data() + (uint64_t)m_SP[0], (uint64_t)m_SP[1]));
+            uint8_t const* data = m_mem.data() + size_t(m_SP[0]);
+            size_t dataSize = size_t(m_SP[1]);
+
+            evm_uint256be topics[] = {toEvmC(m_SP[2])};
```
axic: Is this correct?

libevm/VM.cpp
```diff
@@ -374,72 +416,106 @@ void VM::interpretCases()
...
 
             logGasMem();
             updateIOGas();
 
-            m_ext->log({m_SP[2], m_SP[3], m_SP[4], m_SP[5]}, bytesConstRef(m_mem.data() + (uint64_t)m_SP[0], (uint64_t)m_SP[1]));
+            uint8_t const* data = m_mem.data() + size_t(m_SP[0]);
+            size_t dataSize = size_t(m_SP[1]);
+
+            evm_uint256be topics[] = {
+                toEvmC(m_SP[2]), toEvmC(m_SP[3]), toEvmC(m_SP[4]), toEvmC(m_SP[5])};
```
axic: I wonder if the `LOG`s could be simplified now into a single switch case?

libevm/VM.cpp
```diff
@@ -1312,20 +1397,22 @@ void VM::interpretCases()
...
+            m_runGas = m_rev >= EVM_TANGERINE_WHISTLE ? 700 : 20;
             m_copyMemSize = toInt63(m_SP[3]);
             updateMem(memNeed(m_SP[1], m_SP[3]));
             updateIOGas();
 
-            Address a = asAddress(m_SP[0]);
-            copyDataToMemory(&m_ext->codeAt(a), m_SP + 1);
+            evm_address address = toEvmC(asAddress(m_SP[0]));
+            uint8_t const* pCode = nullptr;
+            size_t codeSize = m_context->fn_table->get_code(&pCode, m_context, &address);
```
axic: Does this makes sure that `pCode` is not dereferenced if `codeSize` is 0?

libevm/VM.cpp
chfast: It's called at the beginning of SSTORE.
In EVM-C refunds are handled by the Client (in this case by ExtVMFace.cpp). Mostly because they are applied at the end of transaction and VM has only the "call" scope. Maybe it would not be so difficult to count refunds in VM and return it with evm_result.

libevm/VM.cpp
```diff
@@ -97,22 +152,20 @@ void VM::adjustStack(unsigned _removed, unsigned _added)
...
-    }
+    evm_uint256be key = toEvmC(m_SP[0]);
+    evm_uint256be rawValue;
+    m_context->fn_table->get_storage(&rawValue, m_context, &m_message->destination, &key);
+    u256 value = fromEvmC(rawValue);
+    if (!value && m_SP[1])
+        m_runGas = VMSchedule::sstoreSetGas;
     else
-        m_runGas = toInt63(m_schedule->sstoreResetGas);
+        m_runGas = toInt63(VMSchedule::sstoreResetGas);
```
chfast: toInt63 not needed.

libevm/VM.cpp
chfast: I'm not sure it's planned. We will enable this when the decision is made.

libevm/VM.cpp
chfast: No. I use `int` for bools in EVM-C. Bool was added in C99 and looks like `_Bool`.

libevm/VM.cpp
chfast: Probably yes, but I don't want to spend time on this VM implementation more than to make it work via EVM-C. A lot of cleanups and optimization can be done here, but I'd rather implement it from scratch with a different license.

libevm/VM.cpp
chfast: Here the `pCode` is just an output argument. But in the next line, copyDataToMemory will not do anything with it if the size is 0.

axic: This comment was meant for the `copyDataToMemory` line.

libevm/VM.cpp
```diff
@@ -15,12 +15,78 @@
...
+    {
+        auto output = vm->exec(_context, _rev, _msg, _code, _codeSize);
+        result.status_code = EVM_SUCCESS;
+        result.gas_left = vm->m_io_gas;
+        auto output_data = new uint8_t[output.size()];
+        std::memcpy(output_data, output.data(), output.size());
+        result.output_data = output_data;
+        result.output_size = output.size();
+        result.release = delete_output;
+        return result;
```
gumb0: You can have one `return result;` in the end of the function

libevm/VM.cpp
```diff
@@ -15,12 +15,78 @@
...
+    const evm_message* _msg, uint8_t const* _code, size_t _codeSize) noexcept
+{
+    auto vm = static_cast<VM*>(_instance);
+    evm_result result = {};
+    try
+    {
+        auto output = vm->exec(_context, _rev, _msg, _code, _codeSize);
+        result.status_code = EVM_SUCCESS;
+        result.gas_left = vm->m_io_gas;
+        auto output_data = new uint8_t[output.size()];
```
gumb0: Maybe a helper like `void setOutput(owning_bytes_ref _output, evm_result& o_result)` is possible not to repeat yourself

libevm/VM.cpp
```diff
@@ -15,12 +15,78 @@
     along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+#include "interpreter.h"
 #include "VM.h"
 
 using namespace std;
 using namespace dev;
```
gumb0: These can be deleted since you explicitly define everything in the `namespase dev { namespace eth {` below?

libevm/VM.h
```diff
@@ -21,15 +21,22 @@
...
+
+#include <boost/optional.hpp>
+
 namespace dev
 {
 namespace eth
 {
 
-class VM: public VMFace
+class VM: public evm_instance
```
gumb0: This looks like you didn't apply clang-format yet

libevm/VM.cpp
chfast: Is this addressed in your opinion @axic ?

libevm/VM.h
```diff
@@ -21,15 +21,22 @@
...
 {
 public:
-    owning_bytes_ref exec(u256& _io_gas, ExtVMFace& _ext, OnOpFunc const& _onOp) final;
+    VM();
+
+    owning_bytes_ref exec(evm_context* _context, evm_revision _rev, const evm_message* _msg,
```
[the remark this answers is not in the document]
chfast: It's not a virtual method any more.

libevm/VM.cpp
axic: Yes

chfast:
I believe I address all the reported issues (2 new commits, I don't want to delete GH commits attached to existing commits). If I missed something or you have found new issues let me know.

axic:
Can you add `--vm interpreter` and `--vm legacy` to the CI?

Also probably worth rebasing after recent changes to logic.

chfast:
Done.

libevm/VM.cpp
```diff
@@ -1614,7 +1738,7 @@ void VM::interpretCases()
 
         CASE(JUMPDEST)
         {
-            m_runGas = 1;
```
axic: Hm, this could be also improved in the legacyvm :)

PR #4907, opened 2018-03-21
Map EVM-C status codes to VM exceptions

chfast:
Fixes https://github.com/ewasm/hera/issues/129.

### DIFF
--- a/libevm/EVMC.cpp
+++ b/libevm/EVMC.cpp
@@ -38,25 +38,45 @@ owning_bytes_ref EVMC::exec(u256& io_gas, ExtVMFace& _ext, const OnOpFunc& _onOp
     auto gas = static_cast<int64_t>(io_gas);
     EVM::Result r = execute(_ext, gas);
 
-    if (r.status() == EVM_REJECTED)
+    switch (r.status())
     {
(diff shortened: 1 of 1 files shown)


libevm/EVMC.cpp
```diff
@@ -44,19 +44,47 @@ owning_bytes_ref EVMC::exec(u256& io_gas, ExtVMFace& _ext, const OnOpFunc& _onOp
         return VMFactory::create(VMKind::Legacy)->exec(io_gas, _ext, _onOp);
     }
 
-    // TODO: Add EVM-C result codes mapping with exception types.
-    if (r.status() == EVM_FAILURE)
+    const evm_status_code status = r.status();
+    assert(status >= 0);  // Do not allow internal errors to pass through.
+
+    if (status == EVM_SUCCESS || status == EVM_REVERT)
```
axic: Why not move the above `EVM_REJECTED` case here too? `if rejected else if success|revert else`

axic: Or perhaps move the entire thing into the switch statement.

chfast: Let's see...

PR #4909, opened 2018-03-22
Always resort to EVM_FAILURE in cross-contract call/create in EVM-C

axic:
Fixes https://github.com/ewasm/hera/issues/122 and https://github.com/ewasm/hera/issues/129


axic:
> You can just replace EVM_REVERT with EVM_FAILURE. For failures other than EVM_REVERT the output buffer is empty anyway.

EVM-C says that output buffer must be empty if `FAILURE` is set. If we just set it to `FAILURE` even if it was a `REVERT`, that may not be the case.

The reason for the way the change is done is the anticipation that `REVERT` will be supported soon 😉

axic:
Actually currently EVM-C is in favour of the original version of this PR: https://github.com/ethereum/evmjit/blob/develop/include/evm.h#L201

I don't mind which version it is going for, just the header must be updated.

chfast:
EVMJIT works fine. :shipit:

PR #4914, opened 2018-03-27
Circle CI: Lower GCC6 memory requirements

chfast:
By switching to RelWithDebInfo build type.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -136,12 +136,12 @@ jobs:
 
   Linux-GCC6-Debug:
     environment:
-      - BUILD_TYPE: Debug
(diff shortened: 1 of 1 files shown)


chfast:
@gumb0 I remember you had also problems with GCC6 on CircleCI. Should be also lower the number or parallel jobs?

gumb0:
Yes, I limited it to 1 in https://github.com/ethereum/cpp-ethereum/pull/4850

chfast:
With -j3 it is ok. Can I merge it? CircleCI has some issues, so codecov failed.

Issue #4911, opened 2018-03-26
Invalid Pong packet in discovery protocol


gumb0:
I think we have [this](https://github.com/paritytech/parity/issues/8038) problem, too.
Looks like we put into Pong packet a hash of original Ping hash
https://github.com/ethereum/cpp-ethereum/blob/3888dedc3bacc38348df58215a51b8a4d73f7be0/libp2p/NodeTable.cpp#L520

It should be just the same Ping hash, as I understand.

PR #4915, opened 2018-03-27
Fix Pong packet of discovery protocol

gumb0:
Fixes https://github.com/ethereum/cpp-ethereum/issues/4911

We seem to have connectivity problems lately, this possibly can be the reason (my understanding is that  Geth started to validate this hash in recent releases)


gumb0:
Asked @fjl to review this, too

gumb0:
Actually while I'm here, I'd try to implement the validation of Pong hashes, too

PR #4916, opened 2018-03-27
EVM-C: copycode()

chfast:
Continuation of https://github.com/ethereum/evmjit/pull/172.

### DIFF
--- a/libevm/ExtVMFace.cpp
+++ b/libevm/ExtVMFace.cpp
@@ -84,14 +84,21 @@ size_t getCodeSize(evm_context* _context, evm_address const* _addr)
     return env.codeSizeAt(fromEvmC(*_addr));
 }
 
-size_t getCode(byte const** o_code, evm_context* _context, evm_address const* _addr)
+size_t copyCode(evm_context* _context, evm_address const* _addr, size_t _codeOffset,
+    byte* _bufferData, size_t _bufferSize)
 {
-    assert(o_code != nullptr);  // Use get_code_size() to get code size.
     auto& env = static_cast<ExtVMFace&>(*_context);
     Address addr = fromEvmC(*_addr);
-    auto& code = env.codeAt(addr);
-    *o_code = code.data();
-    return code.size();
+    bytes const& code = env.codeAt(addr);
+
+    // Handle "big offset" edge case.
+    if (_codeOffset >= code.size())
+        return 0;
+
+    size_t maxToCopy = code.size() - _codeOffset;
+    size_t numToCopy = std::min(maxToCopy, _bufferSize);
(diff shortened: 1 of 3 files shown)


libevm/ExtVMFace.cpp
```diff
@@ -84,14 +84,21 @@ size_t getCodeSize(evm_context* _context, evm_address const* _addr)
...
+    bytes const& code = env.codeAt(addr);
+
+    // Handle "big offset" edge case.
+    if (_codeOffset >= code.size())
+        return 0;
+
+    size_t maxToCopy = code.size() - _codeOffset;
+    size_t numToCopy = std::min(maxToCopy, _bufferSize);
+    byte* newEnd = std::copy_n(&code[_codeOffset], numToCopy, _bufferData);
+    return newEnd - _bufferData;
```
axic: Shouldn't `copy_n` always copy `numToCopy`?

libevm/ExtVMFace.cpp
```diff
@@ -84,14 +84,21 @@ size_t getCodeSize(evm_context* _context, evm_address const* _addr)
     return env.codeSizeAt(fromEvmC(*_addr));
 }
 
-size_t getCode(byte const** o_code, evm_context* _context, evm_address const* _addr)
+size_t copyCode(evm_context* _context, evm_address const* _addr, size_t _codeOffset,
```
axic: Oh, I think we made a mistake. This limits code size to 64k.

libevm/ExtVMFace.cpp
chfast: It does?

axic: So why not `assert(newEnd - _bufferData == numToCopy)` and just return `numToCopy`?

chfast: Oh, this one. I missed that when was reworking the code here.

libevm/ExtVMFace.cpp
chfast: No clue what you are talking about.

chfast: Still no clue.

axic: Never mind, had an embedded system in mind where `size_t` is 16 bits.

I think it would make sense setting an expectation (`static_assert`) that `size_t` is at least 32 bits wide.

axic: Where is the assert? For ewasm it can certainly be bigger than 65k (think about "system contracts").

chfast: There is no assert, but size_t is 32-bit at worse if you target 32-bit ARM architectures. In practice `size_t` is the size of CPU registers.

Issue #4900, opened 2018-03-20
issue of checking duplicate node in bucket within "NodeTable::noteActiveNode"


bean88:
https://github.com/ethereum/cpp-ethereum/blob/a24880812b7be8ceb96fa51cfad050c09fd95e97/libp2p/NodeTable.cpp#L342-L348

removed is passed by reference, so once we got an duplicate element to node, all the items after it will also be removed from the bucket

gumb0:
Wow yeah, I think you're right, thanks for the report

bean88:
you are welcome😁

PR #4920, opened 2018-03-29
Discovery protocol implementation improvements

gumb0:
Fixes https://github.com/ethereum/cpp-ethereum/issues/4900

I started with fixing the issue reported there ^ and ended up rewriting most of that function.

### DIFF
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -21,8 +21,16 @@
 
 #include "NodeTable.h"
 using namespace std;
-using namespace dev;
-using namespace dev::p2p;
+
+namespace dev
+{
+namespace p2p
+{
+inline bool operator==(
+    std::weak_ptr<NodeEntry> const& _weak, std::shared_ptr<NodeEntry> const& _shared)
+{
+    return !_weak.owner_before(_shared) && !_shared.owner_before(_weak);
+}
 
 const char* NodeTableWarn::name() { return "!P!"; }
 const char* NodeTableNote::name() { return "*P*"; }
@@ -222,62 +230,41 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID _target)
     unsigned tail = head == 0 ? lastBin : (head - 1) % s_bins;
     
     map<unsigned, list<shared_ptr<NodeEntry>>> found;
-    unsigned count = 0;
     
     // if d is 0, then we roll look forward, if last, we reverse, else, spread from d
     if (head > 1 && tail != lastBin)
-        while (head != tail && head < s_bins && count < s_bucketSize)
+        while (head != tail && head < s_bins)
         {
             Guard l(x_state);
             for (auto const& n: m_state[head].nodes)
                 if (auto p = n.lock())
-                {
-                    if (count < s_bucketSize)
-                        found[distance(_target, p->id)].push_back(p);
-                    else
-                        break;
-                }
-            
-            if (count < s_bucketSize && tail)
+                    found[distance(_target, p->id)].push_back(p);
+
+            if (tail)
                 for (auto const& n: m_state[tail].nodes)
                     if (auto p = n.lock())
-                    {
-                        if (count < s_bucketSize)
-                            found[distance(_target, p->id)].push_back(p);
-                        else
-                            break;
-                    }
+                        found[distance(_target, p->id)].push_back(p);
 
             head++;
             if (tail)
                 tail--;
         }
     else if (head < 2)
-        while (head < s_bins && count < s_bucketSize)
+        while (head < s_bins)
         {
             Guard l(x_state);
             for (auto const& n: m_state[head].nodes)
                 if (auto p = n.lock())
-                {
-                    if (count < s_bucketSize)
-                        found[distance(_target, p->id)].push_back(p);
-                    else
-                        break;
-                }
+                    found[distance(_target, p->id)].push_back(p);
             head++;
         }
     else
-        while (tail > 0 && count < s_bucketSize)
+        while (tail > 0)
         {
             Guard l(x_state);
             for (auto const& n: m_state[tail].nodes)
                 if (auto p = n.lock())
-                {
-                    if (count < s_bucketSize)
-                        found[distance(_target, p->id)].push_back(p);
-                    else
-                        break;
-                }
+                    found[distance(_target, p->id)].push_back(p);
             tail--;
         }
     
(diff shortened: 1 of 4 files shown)


gumb0:
I will try now to add unit tests to check this, more changes in the code might be needed.

libp2p/NodeTable.cpp
```diff
@@ -328,50 +315,46 @@ void NodeTable::noteActiveNode(Public const& _pubk, bi::udp::endpoint const& _en
...
+            NodeBucket& s = bucket_UNSAFE(newNode.get());
+
+            if (s.nodes.size() < s_bucketSize)
             {
-                if (removed)
-                    clog(NodeTableWarn) << "DANGER: Bucket overflow when swapping node position.";
-                
-                // It's only contested iff nodeentry exists
-                contested = s.nodes.front().lock();
-                if (!contested)
```
gumb0: This condition was always `false` I think, because all `shared_ptr<NodeEntry>` are never destroyed, so `lock()` above was always successful. 
Sp I remove this part

libp2p/NodeTable.cpp
```diff
@@ -328,50 +315,46 @@ void NodeTable::noteActiveNode(Public const& _pubk, bi::udp::endpoint const& _en
...
+        newNode->endpoint.address = _endpoint.address();
+        newNode->endpoint.udpPort = _endpoint.port();
+
+        shared_ptr<NodeEntry> nodeToEvict;
         {
+            // Find a bucket to put a node to
             Guard l(x_state);
-            NodeBucket& s = bucket_UNSAFE(node.get());
-            bool removed = false;
-            s.nodes.remove_if([&node, &removed](weak_ptr<NodeEntry> const& n)
```
gumb0: Iterating over the whole list with `remove_if` was actually not needed, because each node is unique in the bucket.
Now I don't remove at all, but find this node and move it to the back of the list with `splice`

libp2p/NodeTable.cpp
```diff
@@ -328,50 +315,46 @@ void NodeTable::noteActiveNode(Public const& _pubk, bi::udp::endpoint const& _en
...
-        node->endpoint.address = _endpoint.address();
-        node->endpoint.udpPort = _endpoint.port();
-        
-        shared_ptr<NodeEntry> contested;
+        newNode->endpoint.address = _endpoint.address();
+        newNode->endpoint.udpPort = _endpoint.port();
+
+        shared_ptr<NodeEntry> nodeToEvict;
         {
+            // Find a bucket to put a node to
```
gumb0: See also @fjl's description of what is happening here at https://github.com/ethereum/devp2p/blob/master/discv4.md#node-table

libp2p/NodeTable.cpp
```diff
@@ -21,8 +21,16 @@
...
 #include "NodeTable.h"
 using namespace std;
-using namespace dev;
-using namespace dev::p2p;
+
+namespace dev
+{
+namespace p2p
+{
+inline bool operator==(
```
gumb0: (this is needed to find a `weak_ptr` corresponding to a `shared_ptr`)
The whole idea of storing `shared_ptr<NodeEntry>` in a separate map and then storing corresponding `weak_ptr`s in the actual buckets of the Node Table seems unnecessary complicated to me, I will file a separate issue about this.

gumb0:
I will try now to add unit tests to check this, more changes in the code might be needed.

libp2p/NodeTable.cpp
gumb0: Writing the tests I found that I was not right here, `shared_ptr<NodeEntry>` can be destroyed when overwritten with `m_nodes[_node.id] = ret;` in `NodeTable::addNode()`

test/unittests/libp2p/net.cpp
```diff
@@ -52,292 +52,387 @@ BOOST_FIXTURE_TEST_SUITE(net, NetFixture)
...
+    void onReceived(UDPSocketFace*, bi::udp::endpoint const&, bytesConstRef _packet) { if (_packet.toString() == "AAAA") success = true; }
 
-	shared_ptr<UDPSocket<TestUDPSocket, 1024>> m_socket;
+    shared_ptr<UDPSocket<TestUDPSocketHost, 1024>> m_socket;
 
-	bool success = false;
+    bool success = false;
 };
 
-BOOST_AUTO_TEST_CASE(requestTimeout)
```
gumb0: Removing because it doesn't look like testing any of our code, just checking how `std::list::remove_if` works

test/unittests/libp2p/net.cpp
```diff
@@ -52,292 +52,387 @@ BOOST_FIXTURE_TEST_SUITE(net, NetFixture)
...
 {
-	if (test::Options::get().nonetwork)
-	{
-		clog << "Skipping test network/net/kademlia. --nonetwork flag is set.\n";
-		return;
-	}
-
-	TestNodeTableHost node(8);
-	node.start();
-	node.setup();
```
gumb0: `setup()` and `populateAll()` generated a bunch of `NodeTable` instances, which doesn't seem to influence anything in the test

gumb0:
Rebased

gumb0:
Somehow the test coverage of each touched file increased, but overall coverage still decreased

gumb0:
One new test fails on clang builds, I'll try to reproduce it locally

libp2p/NodeTable.cpp
gumb0: Yes, I've brought it back now https://github.com/ethereum/cpp-ethereum/pull/4920/files#diff-b79ec8f203ebd2b76e119c5c3bc8b242R355
`contested` is just renamed to `nodeToEvict`

libp2p/NodeTable.cpp
```diff
@@ -381,7 +374,7 @@ void NodeTable::dropNode(shared_ptr<NodeEntry> _n)
     {
         Guard l(x_state);
         NodeBucket& s = bucket_UNSAFE(_n.get());
-        s.nodes.remove_if([&_n](weak_ptr<NodeEntry> n) { return n.lock() == _n; });
+        s.nodes.remove_if([&_n](weak_ptr<NodeEntry> _bucketEntry) { return _bucketEntry == _n; });
```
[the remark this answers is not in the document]
gumb0: Did you mean here that I should pass `weak_ptr` by reference to lambda?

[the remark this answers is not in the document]
gumb0: Ah you probably meant capturing `_n` by reference doesn't ensure it will live as long as lambda.
But this lambda lives only in this statement, I think it's fine to capture locals by reference.

gumb0:
Now the test began to fail on GCC, too

gumb0:
I couldn't reproduce it with GCC7 build and ctest with 4 threads

gumb0:
Ok, I reproduced it with 8 threads of ctest, great

gumb0:
I got what's going on: `NodeTable` instances created in tests open UDP socket on port 30311; when several testeth instances are being run at once by ctest, it can happen that the port is occupied by one of them, and other instances that create `NodeTable` fail to open the socket, then `NodeTable` is still constructed, but for example eviction mechanism doesn't work (because it is supposed to send Ping packet). So eviction-related test fails in this case.

Not sure what to do with it yet, maybe try to mock the socket.

chfast:
Try assigning different port per test or pick a random one.

gumb0:
(I selected "rebase and merge")

PR #4922, opened 2018-03-29
fix setting author in setChainParams for --test mode

winsvega:
when setChainParams was called. coinbase author was not reset correctly

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -273,7 +273,6 @@ void Client::reopenChain(ChainParams const& _p, WithExisting _we)
         WriteGuard l2(x_preSeal);
         WriteGuard l3(x_working);
 
-        auto author = m_preSeal.author();   // backup and restore author.
         m_preSeal = Block(chainParams().accountStartNonce);
         m_postSeal = Block(chainParams().accountStartNonce);
         m_working = Block(chainParams().accountStartNonce);
@@ -283,7 +282,7 @@ void Client::reopenChain(ChainParams const& _p, WithExisting _we)
         m_stateDB = State::openDB(Defaults::dbPath(), bc().genesisHash(), _we);
 
         m_preSeal = bc().genesisBlock(m_stateDB);
-        m_preSeal.setAuthor(author);
+        m_preSeal.setAuthor(_p.author);
         m_postSeal = m_preSeal;
         m_working = Block(chainParams().accountStartNonce);
     }
--- a/libethereum/ClientTest.cpp
+++ b/libethereum/ClientTest.cpp
@@ -61,7 +61,6 @@ void ClientTest::setChainParams(string const& _genesis)
             BOOST_THROW_EXCEPTION(ChainParamsNotNoProof() << errinfo_comment("Provided configuration is not well formatted."));
 
         reopenChain(params, WithExisting::Kill);
-        setAuthor(params.author); //for some reason author is not being set
     }
     catch (...)
     {
--- a/test/unittests/libethereum/ClientTest.cpp
+++ b/test/unittests/libethereum/ClientTest.cpp
@@ -1,0 +1,107 @@
+/*
+    This file is part of cpp-ethereum.
+
+    cpp-ethereum is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    cpp-ethereum is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/** @date 2018
+ */
+
+#include <libethereum/ChainParams.h>
+#include <libethereum/ClientTest.h>
+#include <libp2p/Network.h>
+#include <libwebthree/WebThree.h>
+#include <test/tools/libtesteth/TestOutputHelper.h>
+#include <boost/test/unit_test.hpp>
+
+using namespace std;
+using namespace dev;
+using namespace dev::eth;
+using namespace dev::test;
+using namespace dev::p2p;
+namespace fs = boost::filesystem;
+
+class TestClientFixture : public TestOutputHelperFixture
+{
+public:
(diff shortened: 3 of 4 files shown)


gumb0:
Interesting, could it fix RPC tests?

libethereum/ClientTest.cpp
```diff
@@ -60,8 +60,9 @@ void ClientTest::setChainParams(string const& _genesis)
         if (params.sealEngineName != "NoProof")
             BOOST_THROW_EXCEPTION(ChainParamsNotNoProof() << errinfo_comment("Provided configuration is not well formatted."));
 
+        setAuthor(params.author);  // Must be called before reopenChain !!!
```
gumb0: Why must it?

libethereum/ClientTest.cpp
```diff
@@ -60,8 +60,9 @@ void ClientTest::setChainParams(string const& _genesis)
...
         reopenChain(params, WithExisting::Kill);
-        setAuthor(params.author); //for some reason author is not being set
+        completeSync();  // set sync state to idle for mining
```
gumb0: Why this? What happens without it?

winsvega: without this. chain state is not being set to idle and test_mineblock does not work

libethereum/ClientTest.cpp
winsvega: see at reopenChain  it replaces the author from m_preSeal. 
so setAuthor would set an author for the m_preSeal

winsvega: yes it should fix author for rpc tests. but we don't run that script anymore.

gumb0: It looks like this `reopenChain` is called only from here.
(also from `Client::killChain()` which looks to be not used and should be removed)

If so, set the author from `params.author` inside `reopwnChain` instead of "backup and restore author" logic there

libethereum/ClientTest.cpp
gumb0: This function doesn't do anything related to the sync, I don't think it's the right place for this... Can you maybe move it up the stack, to where `setChainParams()` is called? (`comleteSync()` is a public method)

winsvega: complete*Sync* does not do anything related to Sync ? 
anyway without this function the sync state is not idle and new blocks not being mined in test mode.

libethereum/Client.cpp
```diff
@@ -290,7 +290,10 @@ void Client::reopenChain(ChainParams const& _p, WithExisting _we)
     }
 
     if (auto h = m_host.lock())
+    {
         h->reset();
+        h->completeSync();  // set sync state to idle for mining
```
winsvega: @gumb0 this looks like Kostil` around m_state variable that must be set to Idle.  but I am afraid that changing it's in internal funcs might break the client.

chfast:
Why nobody is adding unit tests along with the fixes?

libethereum/ClientTest.cpp
gumb0: I meant not `completeSync` but `setChainParams`

libethereum/Client.cpp
gumb0: You move it down the stack instead of up the stack, as I asked. This is probably even worse.

gumb0: Where is `setChainParams` called for your case? Can you put `completeSync` call there?

gumb0:
Yeah, maybe try to to create a unit test for `Client::reopenChain`
We don't have any for test mode-specific functions though, I think

libethereum/Client.cpp
winsvega: setChainParams is test client method. so the best place is to keep it where it was. or make it a separate method of the test client. (which is unnecessary)

gumb0: There is already `ClientTest::completeSync()`, you could call it from from RPC method implementation `Test::test_setChainParams()` (add `asClientTest(m_eth).completeSync();` there), similar to what happens in `Test::test_rewindToBlock()`. It still would be ugly, but less so.

Starting `eth --test` doesn't enable sync by itself, can you find out at which moment does it become enabled? (the stack where `BlockChainSync::m_state` changes)

winsvega: the moment where it changes is strange. it works for 1 time and then it set to notSynced after blockchain reset.

chfast:
Can you also squash commits before merging?

test/unittests/libethereum/ClientTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <boost/test/unit_test.hpp>
+
+using namespace std;
+using namespace dev;
+using namespace dev::eth;
+using namespace dev::test;
+using namespace dev::p2p;
+namespace fs = boost::filesystem;
+
+std::unique_ptr<dev::WebThreeDirect> g_web3;
```
gumb0: Why do you make it global and not the member of the fixture class?

gumb0:
Please convert all tabs to spaces in `libweb3jsonrpc/Test.cpp`

winsvega:
huh set author still does not work on mac

test/unittests/libethereum/ClientTest.cpp
winsvega: how to access fixture from the test case?

test/unittests/libethereum/ClientTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+    cpp-ethereum is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/** @file ClientTest.cpp
```
chfast: Don't add this. There is no information added with this comment.

test/unittests/libethereum/ClientTest.cpp
chfast: test cases are methods of the fixture.

winsvega:
@chfast the test was failing on macOS when the web3 was declared as global variable in cpp.
when I moved it into the class it seems to work.

chfast:
> RandomCodeTests/rndStateTest (Failed)

winsvega:
random test failure is a bug. sometimes gasLimit  is wrong and transaction is not being executed. thus post state check errors out. better fix in a separate PR

PR #4923, opened 2018-03-29
fix creation transaction rpc

winsvega:
an issue occured when `to` field was set in json as empty string

### DIFF
--- a/libweb3jsonrpc/JsonHelper.cpp
+++ b/libweb3jsonrpc/JsonHelper.cpp
@@ -360,8 +360,8 @@ TransactionSkeleton toTransactionSkeleton(Json::Value const& _json)
 
 	if (!_json["from"].empty())
 		ret.from = jsToAddress(_json["from"].asString());
-	if (!_json["to"].empty() && _json["to"].asString() != "0x")
-		ret.to = jsToAddress(_json["to"].asString());
+    if (!_json["to"].empty() && _json["to"].asString() != "0x" && !_json["to"].asString().empty())
+        ret.to = jsToAddress(_json["to"].asString());
 	else
 		ret.creation = true;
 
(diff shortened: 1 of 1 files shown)


libweb3jsonrpc/JsonHelper.cpp
```diff
@@ -360,8 +360,8 @@ TransactionSkeleton toTransactionSkeleton(Json::Value const& _json)
 
 	if (!_json["from"].empty())
 		ret.from = jsToAddress(_json["from"].asString());
-	if (!_json["to"].empty() && _json["to"].asString() != "0x")
-		ret.to = jsToAddress(_json["to"].asString());
+    if (!_json["to"].empty() && _json["to"].asString() != "0x" && !_json["to"].asString().empty())
```
gumb0: An empty string in `to` field sounds to me like an invalid json for transaction Why do you want to ignore this error and create and invalid `TransactionSkeleton` ?

winsvega: but `"to" : "0x"`  is not considered to be invalid? 

hm. in tests it is empty `to` string that indicates creation.

gumb0: Ah yes, right, sorry empty recipient means contract creation

winsvega:
converting all tabs will change the whole file. are you sure?
changed lines are already clang-formated.

gumb0:
I am sure, we do this for all touched files, so that none of the files have a mixture of tabs and spaces

gumb0:
ah now you reformatted all of it.
Could you reformat only the changed lines and for the rest of the file just convert tabs?

winsvega:
why just tabs? whats wrong with converting the whole file? 
I thought if we are converting the file then its good to format all the file. otherwise why do we use clang?

gumb0:
Reformatting the whole file would mess up with `git blame` results (that is, it would be difficult to find out who was the author of particular lines and when)
While the whitespace only can be ignored by the option of `git blame`

PR #4924, opened 2018-03-29
add test_addTransaction

winsvega:
test method that allows to import unsigned transactions with explicit private key field that will be used to sign it.  

https://github.com/ethereum/retesteth/issues/6


gumb0:
I don't quite like the name, there are `eth_sendTransaction` and `eth_sendRawTransaction`, how about `test_signAndSendTransaction` ?

winsvega:
the discussion is here 
https://github.com/ethereum/retesteth/issues/6

winsvega:
ok. I will implement transaction signing at retesteth side.

PR #4931, opened 2018-04-03
cpp-build-env: Install old clang compiler

chfast:


axic:
Why is this needed and can you install both at the same time?

PR #4932, opened 2018-04-03
Always print state root upon completion of transaction

jwasinger:
EWASM is not compatible with the `--jsontrace` option and we need this feature for the trace equivalence tester.


jwasinger:
I need to make a change that enables this with a CLI flag.

jwasinger:
According to @winsvega this change will break testeth.  Do not merge until I find a better fix.

winsvega:
@jwasinger 
if this changes the behaviour of testeth debug options. 
please make a unit test for --statediff  and  --jsontrace

winsvega:
This PR seem to break debug options. I can't merge it. Shall we continue working on it? It looks like functionality it propose not used by anyone.

PR #4934, opened 2018-04-04
fix random transaction OOG before execution

winsvega:
this should prevent random state test from generating a transaction that is rejected
a test expect section was failing when transaction had not enough gas intrinsic.

### DIFF
--- a/test/tools/fuzzTesting/createRandomTest.cpp
+++ b/test/tools/fuzzTesting/createRandomTest.cpp
@@ -1,18 +1,18 @@
 /*
-	This file is part of cpp-ethereum.
+    This file is part of cpp-ethereum.
 
-	cpp-ethereum is free software: you can redistribute it and/or modify
-	it under the terms of the GNU General Public License as published by
-	the Free Software Foundation, either version 3 of the License, or
-	(at your option) any later version.
+    cpp-ethereum is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
 
-	cpp-ethereum is distributed in the hope that it will be useful,
-	but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


test/tools/fuzzTesting/createRandomTest.cpp
```diff
@@ -35,364 +35,364 @@ namespace dev { namespace test {
...
+                int random = randomPercent();
+                if (random < 30)
+                    replace = "0x1c";
+                else if (random < 60)
+                    replace = "0x1d";
+                else
+                    replace = "0x" + rndByteSequence(1);
+            }
+            else if (type == "[BLOCKGASLIMIT]")
+                replace = randomUniIntHex(dev::u256("51000000"), dev::u256("36028797018963967"));
```
chfast: This value does not make sense.

winsvega: min or max?

chfast: The min (the changed one). It's quite far from real values.

Why is this change needed?

winsvega: so that transaction wont go OOG intrinsic

` maxTransactionGasPrice * maxTransactionGasLimit  < minBlockGasLimit `

chfast: So what are the values of maxTransactionGasPrice and maxTransactionGasLimit?

winsvega: maxTransactionGasPrice = 5
maxTransactionGasLimit = 10 000 000   (Vitalik's theoretical block gas limit for the near future)

chfast: Ok, so I'd change it to 50000001.
Secondly, you don't need string literals to create these values.

winsvega:
`50000001`
what about transaction data gas usage? it is also included in intrinsic price.

chfast:
Please squash the commits.

PR #4938, opened 2018-04-05
move transaction checks to sealEngineFace

winsvega:
**move transaction checks to sealEngineFace**
Some of the transaction checks were disabled when using NoProof
so those checks are moved out of Ethash 

**NoProof sets minimum difficulty to blockHeader**
current difficulty opcode return `1` if not to set difficulty directly when using NoProof
@gumb0 please check that this does not break you unit tests with mining on NoProof


libethashseal/Ethash.cpp
```diff
@@ -166,12 +166,6 @@ void Ethash::verifyTransaction(ImportRequirements::value _ir, TransactionBase co
 		else
 			_t.checkChainId(-4);
 	}
-	if (_ir & ImportRequirements::TransactionBasic && _t.baseGasRequired(evmSchedule(_header.number())) > _t.gas())
```
gumb0: Do you mean you would like to have these checks to be executed for `NoProof`, too? Then would it be possible to move it into base class, to `SealEngineFace::verifyTransaction()` ?

The problem with moving it into `Executive` is that these checks won't be done until transaction is being executed, and we would like to reject pending transactions that don't satisfy this earlier, before adding them into pending block.

chfast: I think this is a consensus rule that it is not allowed to "accept" transactions that could exceed block gas limit. I'd prefer to move it somewhere else than Executive. Executive is a messy part that does not do anything useful per se.

winsvega: we decided with @pirapira that it does not related to sealing. @pirapira ?

gumb0: Actually I was not right, we don't call this for pending transactions, because these checks depend on the block header params, so don't make sense until we include tx into block

gumb0: But all other checks in `verifyTransaction` are not related to sealing either. I think it was put there because all these checks require `ChainOperationParams`

pirapira: `SealEngineFace` is fine for me.  That's the place for things not dependent on any particular sealing.

gumb0: But still: before this change we were able to reject incoming blocks containing invalid transactions earlier (in `BlockQueue` validation thread) than we begin to try to execute and import them

gumb0:
Do you know where does difficulty `1` come from?

libethashseal/Ethash.cpp
winsvega: without performing those checks for NoProof engine a transaction with negative gas is executed.

winsvega:
https://github.com/ethereum/cpp-ethereum/blob/develop/libethcore/BlockHeader.cpp#L208

libethashseal/Ethash.cpp
winsvega: some transaction tests are now failing it seems to be. but is it because cpp implementation was wrong without those checks?

gumb0: I think transaction tests validate transactions using `SealEngine` and now you removed some checks, so it's failing

winsvega:
I see. We need to move all transaction checks somewhere out of ethash. Cause if test mode uses NoProof, we still want to distinguish Byzantium transaction and Homestead transactions. 

Cpp importing 1 transaction enters SelaEngineFace constructor 3 times.

winsvega:
failures int builds:
```
CXX: clang++-5.0
/bin/bash: line 1: clang++-5.0: command not found
Exited with code 127
```

chfast:
Rebase!

libethereum/State.cpp
```diff
@@ -49,7 +49,14 @@ namespace
...
 {
-    _e.initialize(_t);
+    try
+    {
+        _e.initialize(_t);
+    }
+    catch (...)
+    {
+        throw;
+    }
```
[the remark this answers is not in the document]
winsvega: It was like this in the other place.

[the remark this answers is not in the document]
gumb0: Where?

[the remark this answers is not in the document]
winsvega: https://github.com/ethereum/cpp-ethereum/blob/develop/libethcore/TransactionBase.cpp#L100

search for throw;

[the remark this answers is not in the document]
pirapira: I see.  I was wrong about "stripped of informatinon", but still the change does nothing useful and wastes lines.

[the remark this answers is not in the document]
gumb0: @winsvega there it adds additional info to the exception before rethrowing

libethcore/SealEngine.cpp
```diff
@@ -55,6 +56,18 @@ void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, Transactio
...
+
+    // Pre calculate the gas needed for execution
+    if (_t.baseGasRequired(schedule) > _t.gas())
+        BOOST_THROW_EXCEPTION(OutOfGasIntrinsic() << RequirementError(
+                                  (bigint)(_t.baseGasRequired(schedule)), (bigint)_t.gas()));
+
+    // Avoid transactions that would take us beyond the block gas limit.
+    if (_gasUsed + (bigint)_t.gas() > _header.gasLimit())
+        BOOST_THROW_EXCEPTION(BlockGasLimitReached() << RequirementError(
+                                  (bigint)(_header.gasLimit() - _gasUsed), (bigint)_t.gas()));
```
pirapira: The error should contain the compared two values above
`_gasUsed + (bigint)_t.gas()` and `_header.gasLimit()`.

libethereum/Client.cpp
```diff
@@ -828,6 +828,7 @@ void Client::rewind(unsigned _n)
     auto h = m_host.lock();
     if (h)
         h->reset();
+    m_tq.clear();
```
pirapira: Is this change related to the other changes?

winsvega: if there are any transactions in the queue. with error status or not. it should be removed from the queue after calling the rewind method.

gumb0: the same for `m_bq` then, too?

winsvega: if this wont affect blockchain memory of which blocks could potentially be valid, then I think yes
m_bq is a client/networking variable, right?

libethcore/SealEngine.cpp
winsvega: requirenment error says that transaction gas should have been not more then 
`gasLimit - gasUsed`
but it got `t.gas` instead

the if could be 
```
if (_t.gas() > _header.gasLimit() - _gasUsed)
```
maybe it was written like this. because  of the bigint ?

pirapira: I'd keep the `if` condition as it is, and change the exception.

winsvega: to what?

winsvega:
```
required:  block gasLimit 
but got:    blockGasUsed + t.gas() 
```
?

libethereum/Client.cpp
pirapira: So `m_bq` was added below this line.

chfast:
Can you squash "fix issues" commit if it fixes only previous commit.

libethereum/BlockChain.cpp
```diff
@@ -1544,7 +1544,7 @@ VerifiedBlockRef BlockChain::verifyBlock(bytesConstRef _block, std::function<voi
             try
             {
                 Transaction t(d, (_ir & ImportRequirements::TransactionSignatures) ? CheckTransaction::Everything : CheckTransaction::None);
-                m_sealEngine->verifyTransaction(_ir, t, h, 0);
+                m_sealEngine->verifyTransaction(_ir, t, h, 0); // the gasUsed vs blockGasLimit is checked later in enact function
```
gumb0: This comment isn't true anymore, is it?

gumb0:
@winsvega I didn't get how these issues were resolved?

> Some tests execution time multiplied by 10
> it spends a lot in Tried to seal sealed block...

> Cpp importing 1 transaction enters SelaEngineFace constructor 3 times.
> 
> lookd like an issue. if transaction throw an exception with this checks it still gets imported into a block and on next check the block throws exception onBadBlock and goes to a loop tried to seal sealed block.

libethcore/SealEngine.h
```diff
@@ -140,6 +140,11 @@ class NoProof: public eth::SealEngineBase
 public:
 	std::string name() const override { return "NoProof"; }
 	static void init();
+    void populateFromParent(BlockHeader& _bi, BlockHeader const& _parent) const override
+    {
+        SealEngineFace::populateFromParent(_bi, _parent);
+        _bi.setDifficulty(chainParams().minimumDifficulty);
```
gumb0: I see that in both `c_genesisInfoFrontierNoProofTest` and `c_genesisInfoMainNetworkNoProofTest` genesis block difficulty is higher than minimumDifficulty.

Should we maybe for consistency change the difficulty of those genesis blocks to minimum? 
If we do that, then the easier fix here could be change the following line to `m_difficulty = _parent.m_difficulty;` https://github.com/ethereum/cpp-ethereum/blob/aa73807c9a6f79114d36ce738658dcba0bc7fbb3/libethcore/BlockHeader.cpp#L208

libethereum/BlockChain.cpp
winsvega: this comment is here because I spent some time trying to figure out where the actual check of the gasLimit is performed

winsvega:
```
Some tests execution time multiplied by 10
it spends a lot in Tried to seal sealed block...
```
This was resolved by throwing an exception at SealEngineFace on GasLimitCheck. 
When that issue occured I used wrong gasLimit from header info in SealEngineFace, but needed to use gasLimit provided into the SealEngineFace verify function.

winsvega:
```
Cpp importing 1 transaction enters SelaEngineFace constructor 3 times.
```

This is still the case. A transaction verify function called 3 times. Thats how cpp client works

libethcore/SealEngine.h
winsvega: This sounds reasonable. Let me run the tests with your change.

libethcore/SealEngine.cpp
```diff
@@ -55,6 +56,19 @@ void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, Transactio
 
 	if (_header.number() >= chainParams().homesteadForkBlock && (_ir & ImportRequirements::TransactionSignatures) && _t.hasSignature())
 		_t.checkLowS();
+
+    eth::EVMSchedule const& schedule = evmSchedule(_header.number());
+
+    // Pre calculate the gas needed for execution
+    if (_t.baseGasRequired(schedule) > _t.gas())
```
gumb0: This was earlier checked only if `_ir & ImportRequirements::TransactionBasic` - basic checks requested. Any reason why did you drop this condition?

libethereum/BlockChain.cpp
gumb0: Ah right, we don't know how much gas will be used until we actually execute the transaction.
Ok, let this stay if you find it helpful.

libethcore/SealEngine.h
winsvega: ```m_difficulty = _parent.m_difficulty``` does not work

libethcore/SealEngine.cpp
winsvega: hm. it looks like there is not. should return it back.

libethcore/SealEngine.h
winsvega: I think difficulty related tests should use Ethash. otherwise it is unclear what should be the return value of difficulty instruction. here it is set to c_minDifficulty. but I guess such tests should say to other clients that it should be executed with normal mining algo.

gumb0: Why didn't it work?

PR #4939, opened 2018-04-05
Quicker termination for eth --test

pirapira:
Before this commit, `eth --test` took half seconds to terminate after a SIGINT.
```
14:03:40.627|eth  Terminate worker 965 ms
```

After this commit, the termination takes less than 100 milliseconds (so
the timecount does not appear on the console).

Since retesteth required this after every test, most of the testing time
was spent during this half seconds.  ping @winsvega


pirapira:
I don't know, to be honest.  The mutex in the `wait_for` call is not used anywhere else.  If I don't put a timeout, it hangs.  The proper fix will be something bigger.

Issue #4942, opened 2018-04-06
ThreadSanitizer warning involving Worker


pirapira:
On `develop` built with
```
$ cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DVMTRACE=1 -DCMAKE_BUILD_TYPE=Debug -DSANITIZE=thread
$ make
```

I run
```
eth/eth --ropsten --kill --peerset "required:20c9ad97c081d63397d7b685a412227a40e23c8bdc6688c6f37e97cfbc22d2b4d1db1510d8f61e6a8866ad7f0e17c02b14182d37ea7c3c8b9c2683aeb6b733a1@52.169.14.227:30303 required:6ce05930c72abc632c58e2e4324f7c7ea478cec0ed4fa2528982cf34483094e9cbc9216e7aa349691242576d552a2a56aaeae426c5303ded677ce455ba1acd9d@13.84.180.240:30303"
```
wait for five seconds, and send ctrl-c.

I see a warning from ThreadSanitizer.
```
==================
WARNING: ThreadSanitizer: use of an invalid mutex (e.g. uninitialized or destroyed) (pid=48566)
    #0 pthread_mutex_lock <null> (libtsan.so.0+0x400ef)
    #1 boost::posix::pthread_mutex_lock(pthread_mutex_t*) /home/yh/.hunter/_Base/8f7cf6d/041e590/8b65c20/Install/include/boost/thread/pthread/mutex.hpp:62 (eth+0xf73d2)
    #2 boost::mutex::lock() /home/yh/.hunter/_Base/8f7cf6d/041e590/8b65c20/Install/include/boost/thread/pthread/mutex.hpp:116 (eth+0xf73d2)
    #3 boost::unique_lock<boost::mutex>::lock() /home/yh/.hunter/_Base/8f7cf6d/041e590/8b65c20/Install/include/boost/thread/lock_types.hpp:346 (eth+0xf73d2)
    #4 boost::shared_lock<boost::shared_mutex>::lock() <null> (eth+0xf9809)
    #5 boost::shared_lock<boost::shared_mutex>::shared_lock(boost::shared_mutex&) /home/yh/.hunter/_Base/8f7cf6d/041e590/8b65c20/Install/include/boost/thread/lock_types.hpp:520 (eth+0x2812ab)
[...]

PR #4943, opened 2018-04-06
~WebThreeDirect() destroys m_net before m_ethereum

pirapira:
This fixes https://github.com/ethereum/cpp-ethereum/issues/4942, a thread sanitizer warning.

### DIFF
--- a/libwebthree/WebThree.cpp
+++ b/libwebthree/WebThree.cpp
@@ -79,7 +79,6 @@ WebThreeDirect::~WebThreeDirect()
     // the guarantee is that m_ethereum is only reset *after* all sessions have ended (sessions are allowed to
     // use bits of data owned by m_ethereum).
     m_net.stop();
-    m_ethereum.reset();
 }
 
 std::string WebThreeDirect::composeClientVersion(std::string const& _client)
--- a/libwebthree/WebThree.h
+++ b/libwebthree/WebThree.h
@@ -211,9 +211,9 @@ public:
 private:
     std::string m_clientVersion;                    ///< Our end-application client's name/version.
 
-    p2p::Host m_net;                                ///< Should run in background and send us events when blocks found and allow us to send blocks as required.
-
     std::unique_ptr<eth::Client> m_ethereum;        ///< Client for Ethereum ("eth") protocol.
+
+    p2p::Host m_net;                                ///< Should run in background and send us events when blocks found and allow us to send blocks as required.
 };
 
 


pirapira:
I'll investigate the test failure in CircleCI.

libethereum/Client.h
```diff
@@ -220,8 +220,8 @@ class Client: public ClientBase, protected Worker
         boost::filesystem::path const& _snapshotPath, WithExisting _forceAction, u256 _networkId);
 
     /// InterfaceStub methods
-    BlockChain& bc() override { return m_bc; }
-    BlockChain const& bc() const override { return m_bc; }
+    std::shared_ptr<BlockChain> bc() override { return m_bc; }
```
chfast: This should keep returning reference. Otherwise you bump ref count on every access.

pirapira:
The reason of the whole change is this.  Sometimes `EthereumHost` had a reference to the `Client::m_bc` after the `Client` object was destroyed.  https://github.com/ethereum/cpp-ethereum/issues/4942

libethereum/Client.h
pirapira: I'll try to fix this one and similar ones.

chfast:
Shoudn't we instead fix the fact that EthereumHost outlives the Client?

pirapira:
Perhaps.  I'll look into that first.

pirapira:
* `WebThree::m_net` is a `Host`, and `Host::m_capabilities` contains a shared pointer to `EthereumHost`.
* `WebThree::m_ethereum` is a unique pointer to `Client`.
* When I reorder `m_net` and `m_ethereum` in `WebThree.h`, I still get the same ThreadSanitizer warning.  I guess the shared pointer to `EthereumHost` has been duplicated into somewhere else.

I thought this (on `develop`) might destroy `EthereumHost` before the `Client`, but didn't work.
```
diff --git a/libwebthree/WebThree.h b/libwebthree/WebThree.h
index 610607e7a..23857171f 100644
--- a/libwebthree/WebThree.h
+++ b/libwebthree/WebThree.h
@@ -211,9 +211,9 @@ public:
 private:
     std::string m_clientVersion;                    ///< Our end-application client's name/version.
 
-    p2p::Host m_net;                                ///< Should run in background and send us events when blocks found and allow us to send blocks as required.
-
     std::unique_ptr<eth::Client> m_ethereum;        ///< Client for Ethereum ("eth") protocol.
+
+    p2p::Host m_net;                                ///< Should run in background and send us events when blocks found and allow us to send blocks as required.
 };
```

pirapira:
Oh I see `WebThreeDirect`'s destructor resets its unique pointer of `Client`.

PR #4946, opened 2018-04-09
Add missing virtual destructor to NetworkFace

chfast:


libwebthree/WebThree.h
```diff
@@ -128,7 +128,7 @@ class WebThreeDirect: public NetworkFace
         bytesConstRef _network = bytesConstRef(), bool _testing = false);
```
axic: Why does this still have `bzz` and `shh` in the `_interfaces` above?

gumb0: Yeah, they can be deleted, they are ignored in `WebThreeDirect` constructor implementation anyway

PR #4944, opened 2018-04-06
Unify the condition variables in Client and Worker

pirapira:
A Client object had two condition variables in Client class and in
Worker class.  This commit unifies these.  Then, I can revert the quick
workaround to make `eth --test` terminate quicker:

https://github.com/ethereum/cpp-ethereum/commit/ccfa8c705cc5dd144765d21fcb53804f34207d24

~Some tests fail, but it's Friday evening, so I file it~

Tests pass in CircleCI.  Maybe my local machine was busy syncing Ropsten.

PR #4948, opened 2018-04-09
notify Client class's condition variable before terminating the thread

pirapira:
This reverts https://github.com/ethereum/cpp-ethereum/commit/ccfa8c705cc5dd144765d21fcb53804f34207d24 and applies a cleaner way to terminate `eth --test` faster.

This comes after a discussion in https://github.com/ethereum/cpp-ethereum/pull/4944

### DIFF
--- a/libethashseal/EthashClient.cpp
+++ b/libethashseal/EthashClient.cpp
@@ -57,6 +57,7 @@ EthashClient::EthashClient(ChainParams const& _params, int _networkID, p2p::Host
 
 EthashClient::~EthashClient()
 {
+    m_signalled.notify_all(); // to wake up the thread from Client::doWork()
     terminate();
 }
 
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -695,7 +695,7 @@ void Client::doWork(bool _doWait)
     if (!m_syncBlockQueue && !m_syncTransactionQueue && (_doWait || isSealed))
     {
         std::unique_lock<std::mutex> l(x_signalled);
-        m_signalled.wait_for(l, chrono::milliseconds(30));
+        m_signalled.wait_for(l, chrono::seconds(1));
     }
 }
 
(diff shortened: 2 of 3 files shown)


libethashseal/EthashClient.cpp
```diff
@@ -57,6 +57,7 @@ EthashClient::EthashClient(ChainParams const& _params, int _networkID, p2p::Host
 
 EthashClient::~EthashClient()
 {
+    m_signalled.notify_all(); // to wake up the thread from Client::doWork()
```
chfast: Why it must be in all derived classes?

pirapira: Because the derived classes' destructors override `~Client()`.

(Similarly, all drived class of `Worker` need to call `terminate()` in the destructors.)

chfast: In case of destructors "override" does not mean that the destructor in the base class is replaced. The base class one will be called as the last one.

pirapira: `terminate()` must be called first, not the last.  Otherwise, the worker thread tries to continue working, using the most derived object.

chfast: Agree. Is this the case also for `.notify_all()`?

pirapira: Yes, without the `notify_all()`, `terminate()` would block, waiting for the thread sleeping for one second.  For anything except `ClientTest`, I just reasoned this, not tried the behavior.

libethereum/Client.cpp
```diff
@@ -695,7 +696,7 @@ void Client::doWork(bool _doWait)
     if (!m_syncBlockQueue && !m_syncTransactionQueue && (_doWait || isSealed))
```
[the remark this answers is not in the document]
pirapira: I'm adding it here.  That way it's easier to figure out later why the change happened.

gumb0:
Please rebase to update clang on CI

pirapira:
This one is on top of 10f1435a1e520f7c813bb6886bf947b43c0de375 (current `develop`) already.

pirapira:
A similar failure happens on `develop` too https://circleci.com/gh/ethereum/cpp-ethereum/3667

gumb0:
Ah right `circle.yml` changes are not merged yet, sorry. Let's merge it when AppVeyor finishes

PR #4949, opened 2018-04-09
Circle CI: Use clang 6

chfast:


chfast:
I'm waiting for clang6 build only and I will merge it.

Issue #4941, opened 2018-04-06
Consensus failure on Ropsten


gumb0:
I am getting this on the Ropsten node
```
  ✘  12:17:40 PM.393|eth


    Import Failure     /home/ubuntu/Ethereum/cpp-ethereum/libethereum/Block.cpp(520): Throw in function dev::u256 dev::eth::Block::enact(const dev::eth::VerifiedBlockRef&, const dev::eth::BlockChain&)
Dynamic exception type: boost::exception_detail::clone_impl<dev::eth::InvalidReceiptsStateRoot>
[dev::tag_get_h256*] = f69e0ac8e52b7e3245608a81f8f5362c06a26cb8aa7ac206e36c58c38e036c49
[dev::tag_extraData*] = [ type: std::vector<unsigned char, std::allocator<unsigned char> >, size: 24, dump: 50 22 00 c4 43 7f 00 00 57 22 00 c4 43 7f 00 00 ]
[dev::tag_required_h256*] = 8c7644aeebf88bb7153f783d5ec5685e86c27a903a2615a6f6150a5e7dc57567
[dev::eth::tag_receipts*] = [ type: std::vector<std::vector<unsigned char, std::allocator<unsigned char> >, std::allocator<std::vector<unsigned char, std::allocator<unsigned char> > > >, size: 24, dump: c0 7e 08 c4 43 7f 00 00 80 7f 08 c4 43 7f 00 00 ]
[dev::eth::tag_now*] = 1523017060
[dev::eth::tag_block*] = [ type: std::vector<unsigned char, std::allocator<unsigned char> >, size: 24, dump: 40 8d 08 c4 43 7f 00 00 f3 94 08 c4 43 7f 00 00 ]


                       Bad Block #02770748.a146494a…
```

gumb0:
It syncs further fine after clean rebuild of `eth`

PR #4951, opened 2018-04-10
New unit test for consensus failure of ECMUL contract

gumb0:
It fails for me with GCC7 build on Trusty

It was the reason for https://github.com/ethereum/cpp-ethereum/issues/4941

### DIFF
--- a/test/unittests/libdevcrypto/LibSnark.cpp
+++ b/test/unittests/libdevcrypto/LibSnark.cpp
@@ -1,18 +1,18 @@
 /*
-	This file is part of cpp-ethereum.
+    This file is part of cpp-ethereum.
 
-	cpp-ethereum is free software: you can redistribute it and/or modify
-	it under the terms of the GNU General Public License as published by
-	the Free Software Foundation, either version 3 of the License, or
-	(at your option) any later version.
+    cpp-ethereum is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
 
-	cpp-ethereum is distributed in the hope that it will be useful,
-	but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


pirapira:
For me the new test succeeds on Linux
```
 $ ./testeth -t '*/ECMULRopstenBug'
Running 1 test case...

*** No errors detected
```

gumb0:
@pirapira Looks like it's GCC7 only

pirapira:
Still I cannot reproduce the failure.  Maybe I'm using a different cmake option.
```
$ export CC=gcc
$ export CXX=g++
$ cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DVMTRACE=1 -DCMAKE_BUILD_TYPE=Debug
$ make
$ cd test
$ ./testeth -t '*/ECMULRopstenBug'
Running 1 test case...

*** No errors detected
~/src/cpp-ethereum/build-gcc7/test(ecmul-consensus-bug) $ ./testeth --version
cpp-1.3.0+commit.d91d1c2e.Linux.g++
~/src/cpp-ethereum/build-gcc7/test(ecmul-consensus-bug) $ g++ --version
g++ (Ubuntu 7.3.0-15ubuntu2) 7.3.0
```

gumb0:
Hmm for me all other LibSnark tests fail, too

gumb0:
Now with the clean build it doesn't happen 🤦‍♂️

chfast:
@gumb0 if you have access to linux, you can instrument the test with UB sanitizer: -DSANITIZE=undefined`.

gumb0:
The problem disappeared after clean rebuild

@chfast thanks for the tip, I'll try that with my another linux problem...

chfast:
In this case I expected undefined behavior exploited by the newer compiler for optimization.

But we should add sanitizers to CIs

chfast:
We can still merge this PR.

gumb0:
Ok, but I'll come up with another name for the test

gumb0:
It needs a review now

PR #4953, opened 2018-04-10
Handle EVMC internal error

chfast:


libevm/VMFace.h
```diff
@@ -35,6 +35,16 @@ ETH_SIMPLE_EXCEPTION_VM(StackUnderflow);
 ETH_SIMPLE_EXCEPTION_VM(DisallowedStateChange);
 ETH_SIMPLE_EXCEPTION_VM(BufferOverrun);
 
+/// Reports VM internal error. This is not based on VMException because it must be handled
+/// differently than defined consensus exceptions.
+struct InternalVMError : Exception
```
gumb0: Boost.exception recommendation is to virtually inherit it and to contain no data members.
Then define
```
using errinfo_evmcStatusCode = boost::error_info<struct tag_evmcStatusCode, evmc_status_code>;
```
and at the place of throw
```
BOOST_THROW_EXCEPTION(InternalVMError() << errinfo_evmcStatusCode(r.status()));
```

gumb0: It does improve the usage -
1. you won't need to add the code for printing this member, it will be printed by any `ex.what()`
2. I'm for discouraging any data members in exception classes in general because it can be dangerous sometimes. Here it will work fine, because it's just an `enum`, but for any nontrivial class `BOOST_THROW_EXCEPTION` will behave weirdly - see my problems in https://github.com/ethereum/cpp-ethereum/pull/4864 - having a `std::string` member is not allowed because it has a throwing copy constructor.

Declaration is somewhat the same number of lines.
Depenency on `error_info` is already there in `<libdevcore/Exceptions.h>`

chfast: Done.

chfast:
Thanks for the review!

Issue #4583, opened 2017-10-06
Receipt json in RPC output should have "status" field


gumb0:
Currently it's called `statusCode`.

LoveULin:
Hi, I submitted a PR #4954 that simply replaced "statusCode" with "status". Please tell me if I have misunderstood your intentions:D

PR #4954, opened 2018-04-11
Receipt json in RPC output should have "status" field

LoveULin:
replace field name "statusCode" with "status" in RPC output of receipt json #4583

### DIFF
--- a/test/tools/fuzzTesting/createRandomTest.cpp
+++ b/test/tools/fuzzTesting/createRandomTest.cpp
@@ -1,18 +1,18 @@
 /*
-	This file is part of cpp-ethereum.
+    This file is part of cpp-ethereum.
 
-	cpp-ethereum is free software: you can redistribute it and/or modify
-	it under the terms of the GNU General Public License as published by
-	the Free Software Foundation, either version 3 of the License, or
-	(at your option) any later version.
+    cpp-ethereum is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


LoveULin:
Sure. I simply replaced the `toJS` with `toString`, and cast the `uint8_t` returned from `_t.statusCode()` to `uint16_t`. Please help review. Thank you!!:D @gumb0

gumb0:
It's would be nicer to fix `toString()` itself instead of doing `static_cast<uint16_t>` here two times.
I suggest to add a template specialization for `toString<uint8_t>`

gumb0:
We return incorrect value for receipt's `gasUsed`, I filed a separate issue https://github.com/ethereum/cpp-ethereum/issues/4956
@LoveULin maybe you would like to fix that, too

LoveULin:
OK. I think what you said makes sense. I will fix it as soon as possible.
@gumb0 Yes, I'd like to. I will try to fix that (#4956 ) once I have finished my recent work.

gumb0:
Some tests that use `toString<uint8_t>()` fail now, but they look wrong to me, I will fix them.

gumb0:
@LoveULin  Sorry, pushing to your `develop` branch went wrong, now your `develop` corresponds to upstream `develop`, but your & mine changes are here in the upstream branch, I created a new PR to replace this: https://github.com/ethereum/cpp-ethereum/pull/4957

gumb0:
(git magic would be less confusing to me if you created PR from some your branch other than `develop`)

PR #4955, opened 2018-04-12
test_closeClient method

winsvega:
https://github.com/ethereum/retesteth/issues/17

### DIFF
--- a/libweb3jsonrpc/test.json
+++ b/libweb3jsonrpc/test.json
@@ -1,4 +1,5 @@
 [
+{ "name": "test_closeClient", "params": [], "order": [], "returns": false},
 { "name": "test_setChainParams", "params": [{}], "order": [], "returns": false},
 { "name": "test_mineBlocks", "params": [0], "returns": false },
 { "name": "test_modifyTimestamp", "params": [0], "returns": false },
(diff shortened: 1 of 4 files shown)


libweb3jsonrpc/test.json
```diff
@@ -1,4 +1,5 @@
 [
+{ "name": "test_closeClient", "params": [], "order": [], "returns": false},
```
pirapira: I'm just curious; what does this "returns" field mean?

gumb0:
Couldn't you just send a SIGTERM from retesteth?

libweb3jsonrpc/test.json
winsvega: I guess. false or true  means bool.  and "" means string. judgind on how other .json descriptions are made

winsvega:
how ? 
pclose  wont send SIGTERM

winsvega:
hm. in theory there should be a way to call kill $pid on unix system. but would it be a correct way to close the client ?

gumb0:
Yes, sending `SIGTERM` or `SIGINT` should be a correct way, the client is responsible for handling it correctly.

But to send it you need to know PID and you have to be either the owner of the client process or a superuser

pirapira:
@winsvega I guess, if `retesteth` doesn't kill `eth`, after running the test suite, there will be 400 `eth` processes left.  Is this correct?

winsvega:
not really. just the number of threads retesteth runs the eth instances.

winsvega:
found a way to send sigterm on unix

PR #4960, opened 2018-04-16
Handle EVMC Internal Error in Executive

chfast:
Follow up #4953.

### DIFF
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -459,6 +459,12 @@ bool Executive::go(OnOpFunc const& _onOp)
             m_excepted = toTransactionException(_e);
             revert();
         }
+        catch (InternalVMError const& _e)
+        {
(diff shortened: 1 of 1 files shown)


libethereum/Executive.cpp
```diff
@@ -459,6 +459,12 @@ bool Executive::go(OnOpFunc const& _onOp)
             m_excepted = toTransactionException(_e);
             revert();
         }
+        catch (InternalVMError const& _e)
+        {
+            cwarn << "Internal VM Error (" << *boost::get_error_info<errinfo_evmcStatusCode>(_e) << ")\n"
```
axic: The other exception messages here don't have an explicit new line character.

chfast: It looks a bit better if the diagnostic_information is on the next lines.

PR #4965, opened 2018-04-18
Refactor logging - MemoryDB, OverlayDB, Executive, State

gumb0:
Also added a simple macro for sending a log message without creating a Logger like this:
```
clogSimple(4, "channel") << "message";
```
I want to rename it to just `clog`, but I need to get rid of the old `clog` macro first (that won't happen before switching most of the log statements to the new system)

### DIFF
--- a/libdevcore/Log.h
+++ b/libdevcore/Log.h
@@ -238,7 +238,7 @@ public:
 #define LOG BOOST_LOG
 
 // Simple cout-like stream objects for accessing common log channels.
-// Dirties the global namespace, but oh so convenient...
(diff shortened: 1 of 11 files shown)


chfast:
Does this improve performance?

gumb0:
I didn't measure anything, but theoretically it might, because it won't execute anything to the right of `LOG(...) << ` if it's channel and verbosity are filtered out

gumb0:
Also because sending messages to log is asynchronous; there's a separate thread responsible for flushing the log into backend (stdout)

PR #4967, opened 2018-04-19
Refactor logging - libp2p

gumb0:
In general I think it's preferable to use loggers local to the task (say, loggers declared as class members), this way you avoid the overhead of thread-safety needed for global loggers.

libp2p though used the number of channels shared among all classes of the library.
For now I mostly replicated this approach by declaring several global loggers (in `p2p` namespace). Still I simplified it somewhat, reducing the number of loggers (some were merged into one and some I was able to make class members)

All libp2p loggers write to `"net"` channel (to be able to filter all of them out) and vary by verbosity.

It doesn't include node discovery logs from `NodeTable` class, they will still go to another channel I think (in the following PR)

### DIFF
--- a/libethereum/EthereumHost.cpp
+++ b/libethereum/EthereumHost.cpp
@@ -178,9 +178,9 @@ public:
         if (_blockId.size() == 32) // block id is a hash
         {
             blockHash = _blockId.toHash<h256>();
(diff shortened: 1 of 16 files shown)


gumb0:
Rebased

PR #4969, opened 2018-04-19
fix transactionQueue

winsvega:
+ faster client worker restart

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -260,6 +260,7 @@ void Client::reopenChain(WithExisting _we)
 
 void Client::reopenChain(ChainParams const& _p, WithExisting _we)
 {
+    m_signalled.notify_all(); // to wake up the thread from Client::doWork()
     bool wasSealing = wouldSeal();
     if (wasSealing)
         stopSealing();
--- a/libethereum/TransactionQueue.cpp
+++ b/libethereum/TransactionQueue.cpp
@@ -348,6 +348,7 @@ void TransactionQueue::clear()
 	WriteGuard l(m_lock);
 	m_known.clear();
 	m_current.clear();
+	m_dropped.clear();
 	m_currentByAddressAndNonce.clear();
 	m_currentByHash.clear();
 	m_future.clear();


libethereum/TransactionQueue.cpp
```diff
@@ -348,6 +348,7 @@ void TransactionQueue::clear()
 	WriteGuard l(m_lock);
 	m_known.clear();
 	m_current.clear();
+	m_dropped.clear();
```
pirapira: This looks good.

libethereum/Client.cpp
```diff
@@ -696,7 +696,7 @@ void Client::doWork(bool _doWait)
     if (!m_syncBlockQueue && !m_syncTransactionQueue && (_doWait || isSealed) && isWorking())
     {
         std::unique_lock<std::mutex> l(x_signalled);
-        m_signalled.wait_for(l, chrono::seconds(1));
+        m_signalled.wait_for(l, chrono::milliseconds(30));
```
pirapira: Does this make any difference?

gumb0: This shouldn't go in

winsvega:
1 sec increases rpc test time significantly.

pirapira:
@winsvega there should be a problem in the previous fix https://github.com/ethereum/cpp-ethereum/pull/4948 . Will you investigate?

winsvega:
@pirapira fixed

libethereum/ClientTest.cpp
```diff
@@ -61,6 +61,7 @@ void ClientTest::setChainParams(string const& _genesis)
         if (params.sealEngineName != "NoProof")
             BOOST_THROW_EXCEPTION(ChainParamsNotNoProof() << errinfo_comment("Provided configuration is not well formatted."));
 
+        m_signalled.notify_all(); // to wake up the thread from Client::doWork()
```
pirapira: Looks good, too.

pirapira:
@gumb0 now the wait-for-one-second is back.

libethereum/ClientTest.cpp
gumb0: Can we move this to `Client::reopenChain(ChainParams const& _p, WithExisting _we)` before `stopWorking()` call?

libethereum/TransactionQueue.cpp
gumb0: This is fine, as this method is used only by `ClientTest` I think

winsvega:
moved

libethereum/TransactionQueue.cpp
winsvega: No. this fixes a bug. if transaction is considered invalid but then after rewind it is being imported again it might become valid.

gumb0: Right, and rewind is used only in test mode

Issue #4072, opened 2017-05-09
Cleanup Instruction table


chfast:
In files Instruction.(h|cpp), there are a lot of unnecessary data and unused functions.

seusher:
It looks like this issue is resolved in EIP211: RETURNDATA #4062 (pending merge) and is no longer up-for-grabs.

chfast:
No. #4062 has some changes for this, but everything. E.g. "side effect" is not used anywhere I think. So we should wait with starting this until #4062 is merged to avoid conflicts, but there is more to be done here.

AnthonyBroadCrawford:
@chfast Is it okay if I run with this issue?  I'm new to the codebase and this seems like it may be a good place to start.  Let me know as I'm excited to contribute.

chfast:
Yes, sure. It's yours.

Some things might have been improved recently, but check out the Instruction.* files for unused data.

AnthonyBroadCrawford:
@chfast Awesome.  I’ll start to dig in this evening and hopefully have something to review sometime next week.  

Where’s the best place to find all coding standards, preferred syntax, etc?  Want to make sure once the PR is ready it’s in the best shape possible.  

Cheers.

chfast:
See https://github.com/ethereum/cpp-ethereum/blob/develop/CODING_STYLE.md.

The best is to use clang-format to format your changes. In case big tables are not readable after reformatting with clang-format, you can come up with manual formatting by disabling clang-format it that section.

chfast:
1. The `additional` field does not seem to be needed.
2. You can change the type of `name` from `std::string` to `const char* const`.
3. Changing `std::map` to flat array would also be nice, but probably should be done as an another PR.

AnthonyBroadCrawford:
@chfast got it, helpful, thank you.

AnthonyBroadCrawford:
@chfast I made the changes and opened this PR ( https://github.com/ethereum/cpp-ethereum/pull/4970 )

Please provide any/all feedback and I'll be happy to make the recommended changes.

PR #4970, opened 2018-04-19
Cleaned-up unused data fields in Instruction table

AnthonyBroadCrawford:
# PR for issue #4072  
This PR is now development complete and ready for review. 

## Decisions 
I followed the guidance of @chfast 

1. The additional field does not seem to be needed.
2. You can change the type of name from std::string to const char* const.

##Feedback 
I'm brand new to the codebase so please feel free to provide any/all feedback and I'll be happy to make the changes (I may ask `why` to help better understand)


libevm/Instruction.h
```diff
@@ -245,8 +245,7 @@ enum class Tier : unsigned
 /// Information structure for a particular instruction.
 struct InstructionInfo
 {
-	std::string name;   ///< The name of the instruction.
-	int additional;     ///< Additional items required in memory for this instructions (only for PUSH).
+	const char* const name;   ///< The name of the instruction.
```
chfast: If you made it `const`, can you also try to const other fields?

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -280,7 +280,7 @@ std::string RandomCodeBase::generate(int _maxOpNumber, RandomCodeOptions const&
 			opcode = makeOpcodeDefined(opcode);
 			eth::Instruction inst = (eth::Instruction) opcode;
 			eth::InstructionInfo info = eth::instructionInfo(inst);
-			if (info.name.find("PUSH") != std::string::npos)
+			if (strcmp(info.name, "PUSH") == 0)
```
chfast: It's not the same. The `.find()` searches for a substring, so it will match all `PUSHxx` instructions.
Better leave it as it was:
```
if (std::string{info.name}.find("PUSH") != std::string::npos)
```

libevm/Instruction.h
chfast: Also initialized this field: `const char* const name = nullptr`.

test/tools/fuzzTesting/fuzzHelper.cpp
```diff
@@ -245,7 +245,7 @@ bool isOpcodeDefined(uint8_t _opcode)
 {
 	eth::Instruction inst = (eth::Instruction) _opcode;
 	eth::InstructionInfo info = eth::instructionInfo(inst);
-	return (info.gasPriceTier != dev::eth::Tier::Invalid && !info.name.empty()
+	return (info.gasPriceTier != dev::eth::Tier::Invalid && (info.name != NULL)
```
chfast: `info.name != nullptr`. And skip `()` to match the style of the first check.

libevm/Instruction.h
```diff
@@ -245,8 +245,7 @@ enum class Tier : unsigned
...
 	int args;           ///< Number of items required on the stack for this instruction (and, for the purposes of ret, the number taken from the stack).
 	int ret;            ///< Number of items placed (back) on the stack by this instruction, assuming args items were removed.
 	Tier gasPriceTier;   ///< Tier for gas pricing.
```
chfast: Init it with "invalid" value.

libevm/Instruction.cpp
```diff
@@ -24,204 +24,204 @@ namespace eth
...
+	{ Instruction::JUMPCI,       { "JUMPCI",              2,    0, Tier::High } },
 }; 
  
 InstructionInfo instructionInfo(Instruction _inst)
 {
 	auto it = c_instructionInfo.find(_inst);
 	if (it != c_instructionInfo.end())
 		return it->second;
-	return InstructionInfo{{}, 0, 0, 0, Tier::Invalid};
+	return InstructionInfo{{}, 0, 0, Tier::Invalid};
```
chfast: If you have all fields initialized in the declaration of `InstructionInfo`, you can just do `return InstructionInfo{}` or even `return {}`.

test/tools/fuzzTesting/fuzzHelper.cpp
AnthonyBroadCrawford: @chfast  Ah, thank you.  My apologies for not fully understanding the difference.  I'll make that change.

Issue #4726, opened 2017-12-28
Improve error handling in RPC


gumb0:
1. Exceptions should be caught in RPC methods implementation (and re-thrown as `jsonrpc::JsonRpcException`) in order to generate error response and not just crash

At least in these methods I don't see proper handling of possible errors:
```
admin_eth_findBlock // invalid block hash will throw
admin_eth_newAccount // I think writing file on disk might throw
admin_addPeer // incorrect node address will throw
debug_traceBlock // will throw if RLP is incorrect hex string
debug_traceBlockByHash
personal_unlockAccount
```

1a. Many methods could be improved with better error messages (many report just "invalid params" for any possible error).

2. Sending RPC request with invalid json in body on Linux through HTTP results for me in `curl: (52) Empty reply from server` and `Request timed out: timeout('timed out',)` in jsonrpcproxy log.
On Windows curl just hangs never getting the response.

It should generate proper response with error message for this case.

Manugal:
Hi, 
I am a new contributor and I'd like to take care of this issue. Could you give me some guidance on how to test the methods listed above? I have already implemented exception handling on some of them, but I need to test.

gumb0:
Hi @Manugal, thank you for your interest!
There's no good way to write autotests for RPC methods at the moment, but you should be able to manually check it by starting `eth` (without arguments it will just start the sync with the main net) and then using `ethereum-console` (https://github.com/ethereum/ethereum-console) to connect to the node and send RPC requests.

So something like this should work
```
$ npm install -g ethereum-console
$ ethconsole
> web3.debug.traceBlockByHash("0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3")
```

There were some changes in `ethconsole` lately, and I haven't tried it with a latest version, so let me know if it's not working at all and you need some assistance (maybe in Gitter it would be more convenient)

Note also: it's better to submit your changes as a PR as early as possible, than trying to finish big chunk of work first.

PR #4972, opened 2018-04-20
test_getPostState

winsvega:
return current actual post state of the blockchain.

### DIFF
--- a/libweb3jsonrpc/Test.cpp
+++ b/libweb3jsonrpc/Test.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "Test.h"
+#include <libdevcore/CommonJS.h>
 #include <jsonrpccpp/common/errors.h>
 #include <jsonrpccpp/common/exception.h>
 #include <libethereum/ClientTest.h>
@@ -32,6 +33,60 @@ using namespace dev::rpc;
 using namespace jsonrpc;
 
 Test::Test(eth::Client& _eth): m_eth(_eth) {}
+
+Json::Value fillJsonWithState(eth::State const& _state)
+{
+    Json::Value oState;
+    for (auto const& a : _state.addresses())
+    {
+        Json::Value o;
+        o["balance"] = toCompactHexPrefixed(_state.balance(a.first), 1);
+        o["nonce"] = toCompactHexPrefixed(_state.getNonce(a.first), 1);
+
+        Json::Value store;
+        for (auto const& s : _state.storage(a.first))
+            store[toCompactHexPrefixed(s.second.first, 1)] =
+                toCompactHexPrefixed(s.second.second, 1);
+        o["storage"] = store;
+        o["code"] = toHexPrefixed(_state.code(a.first));
+        oState[toHexPrefixed(a.first)] = o;
+    }
+    return oState;
+}
+
+string exportLog(eth::LogEntries const& _logs)
+{
+    RLPStream s;
+    s.appendList(_logs.size());
+    for (eth::LogEntry const& l : _logs)
+        l.streamRLP(s);
+    return toHexPrefixed(sha3(s.out()));
+}
+
+const string c_postStateJustHashVersion = "justhash";
+const string c_postStateFullStateVersion = "fullstate";
+const string c_postStateLogHashVersion = "justloghash";
+string Test::test_getPostState(Json::Value const& param1)
+{
+    Json::Value out;
+    Json::FastWriter fastWriter;
+    if (param1["version"].asString() == c_postStateJustHashVersion)
+        return toJS(m_eth.postState().state().rootHash());
+    else if (param1["version"].asString() == c_postStateFullStateVersion)
+    {
+        out = fillJsonWithState(m_eth.postState().state());
+        return fastWriter.write(out);
+    }
+    else if (param1["version"].asString() == c_postStateLogHashVersion)
+    {
+        if (m_eth.blockChain().receipts().receipts.size() != 0)
+            return exportLog(m_eth.blockChain().receipts().receipts[0].log());
+        else
+            return toJS(EmptyListSHA3);
+    }
+    BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS));
+    return "error";
+}
 
 bool Test::test_setChainParams(Json::Value const& param1)
 {
--- a/libweb3jsonrpc/Test.h
+++ b/libweb3jsonrpc/Test.h
@@ -43,6 +43,7 @@ public:
 		return RPCModules{RPCModule{"test", "1.0"}};
 	}
 
+	virtual std::string test_getPostState(const Json::Value& param1) override;
 	virtual bool test_setChainParams(const Json::Value &param1) override;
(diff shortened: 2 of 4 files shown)


libweb3jsonrpc/Test.cpp
```diff
@@ -33,6 +34,76 @@ using namespace jsonrpc;
...
+{
+    bool mapEmpty = (_map.size() == 0);
+    Json::Value oState;
+    for (auto const& a : _state.addresses())
+    {
+        if (_map.size() && _map.find(a.first) == _map.end())
+            continue;
+
+        Json::Value o;
+        if (mapEmpty || _map.at(a.first).hasBalance())
```
pirapira: Why is this empty case treated in a special way?

winsvega: this is actually a copy of the same method from tests. but that one works with json_spirit. 
I guess the account map here could be removed. because it is needed for expect sections with incomplete post states.

libweb3jsonrpc/Test.cpp
```diff
@@ -33,6 +34,64 @@ using namespace jsonrpc;
...
+}
+
+const int c_postStateJustHashVersion = 1;
+const int c_postStateFullStateVersion = 2;
+const int c_postStateLogHashVersion = 3;
+string Test::test_getPostState(Json::Value const& param1)
+{
+    Json::Value out;
+    Json::FastWriter fastWriter;
+    if (u256(param1["version"].asString()) == c_postStateJustHashVersion)
```
pirapira: Does the RPC method take `1`, `2`, or `3`, indicating different versions?  In other parts of the RPC, "latest" means "latest", and so on.  I guess strings like "jushash" "fullstate" or "loghash" are easier.

libweb3jsonrpc/Test.cpp
```diff
@@ -33,6 +34,64 @@ using namespace jsonrpc;
...
+        return fastWriter.write(out);
+    }
+    else if (u256(param1["version"].asString()) == c_postStateLogHashVersion)
+    {
+        if (m_eth.blockChain().receipts().receipts.size() != 0)
+            return exportLog(m_eth.blockChain().receipts().receipts[0].log());
+        else
+            return toJS(EmptyListSHA3);
+    }
+    return "";
```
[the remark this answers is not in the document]
winsvega: how do we report an error via RPC reply ?

libweb3jsonrpc/Test.cpp
winsvega: no problem here. it could be anything.
the discussion is here 
https://github.com/ethereum/retesteth/issues/7

libweb3jsonrpc/Test.cpp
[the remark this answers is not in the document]
pirapira: I don't know, look at other methods.

libweb3jsonrpc/Test.cpp
```diff
@@ -33,6 +34,64 @@ using namespace jsonrpc;
...
+        Json::Value store;
+        for (auto const& s : _state.storage(a.first))
+            store[toCompactHexPrefixed(s.second.first, 1)] =
+                toCompactHexPrefixed(s.second.second, 1);
+        o["storage"] = store;
+
+        if (_state.code(a.first).size() > 0)
+            o["code"] = toHexPrefixed(_state.code(a.first));
+        else
+            o["code"] = "";
```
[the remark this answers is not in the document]
winsvega: you right. changed this now

libweb3jsonrpc/Test.cpp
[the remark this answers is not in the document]
winsvega: well there is ```BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS));```

but it stops the client. which is not a good thing. you dont want a client to stop on malicious rpc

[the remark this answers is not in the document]
gumb0: It shouldn't stop the client, all RPC methods report errors kind of like that and the last time I tried it I saw correct error message on the caller side.

Could you investigate whether it crashes because of unhandled exception or for other reason?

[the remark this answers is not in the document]
gumb0: Also related issue is https://github.com/ethereum/cpp-ethereum/issues/4726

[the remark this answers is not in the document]
winsvega: no. it does not crash the client. I was wrong. the rpc error is returned.

winsvega:
@pirapira  do you have any more proposals? 
I have one change: missing method note in test.json file

winsvega:
Actually should this method be a test method only ?

gumb0:
API-design-wise it should better be three different API methods instead of one method doing completely different things based on "version" param
Like `getStateRootHash`, `getState`... and I'm not sure why you need the last one

libweb3jsonrpc/Test.cpp
```diff
@@ -33,6 +34,60 @@ using namespace jsonrpc;
...
+        return toJS(m_eth.postState().state().rootHash());
+    else if (param1["version"].asString() == c_postStateFullStateVersion)
+    {
+        out = fillJsonWithState(m_eth.postState().state());
+        return fastWriter.write(out);
+    }
+    else if (param1["version"].asString() == c_postStateLogHashVersion)
+    {
+        if (m_eth.blockChain().receipts().receipts.size() != 0)
+            return exportLog(m_eth.blockChain().receipts().receipts[0].log());
```
gumb0: This gets the receipt of only the first transaction of the block, is this your intention?

gumb0: Couldn't you just use receipts root which is in the block header?

libweb3jsonrpc/Test.cpp
```diff
@@ -33,6 +34,60 @@ using namespace jsonrpc;
...
+
+const string c_postStateJustHashVersion = "justhash";
+const string c_postStateFullStateVersion = "fullstate";
+const string c_postStateLogHashVersion = "justloghash";
+string Test::test_getPostState(Json::Value const& param1)
+{
+    Json::Value out;
+    Json::FastWriter fastWriter;
+    if (param1["version"].asString() == c_postStateJustHashVersion)
+        return toJS(m_eth.postState().state().rootHash());
```
gumb0: Is this any different from getting `eth.getBlock("latest").stateRoot` ?

libweb3jsonrpc/Test.cpp
```diff
@@ -33,6 +34,60 @@ using namespace jsonrpc;
 
 Test::Test(eth::Client& _eth): m_eth(_eth) {}
 
+Json::Value fillJsonWithState(eth::State const& _state)
```
gumb0: Put free functions into the unnamed namespace

chfast:
> API-design-wise it should better be three different API methods instead of one method doing completely different things based on "version" param

I'm also for it. It's better to have many simple RPC methods than not many but complex RPC methods. Having simple ones you can who is using what and why. And also able to deprecate individual methods in future.

PR #4973, opened 2018-04-20
parse different versions of genesis config

winsvega:
parse test genesis format
https://github.com/ethereum/retesteth/issues/4

```
# Genesis config for running a StateTest via RPC
{
       "version" : "1",
       "params": {
            "miningMethod" : "NoProof",
	    "forkRules" : "Frontier",
            "blockReward" : "0x00"
       },
       "genesis" : {
            "coinbase" : "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
            "gasLimit" : "0x0f4240",
            "timestamp" : "0x03e8"
       },
       "state": {
            "0x095e7baea6a6c7c4c2dfeb977efac326af552d87" : {
                "balance" : "0x0de0b6b3a76586a0",
                "code" : "0x6001600101600055",
                "nonce" : "0x00",
                "storage" : {
                    "0x00" : "0x02"
                }
            },
            "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba" : {
                "balance" : "0x29a2241af62ca034",
                "code" : "",
                "nonce" : "0x00",
                "storage" : {
                }
            }
       }
}
```

### DIFF
--- a/libethcore/ChainOperationParams.h
+++ b/libethcore/ChainOperationParams.h
@@ -77,10 +77,13 @@ struct ChainOperationParams
 	/// General chain params.
 private:
 	u256 m_blockReward;
+	bool m_allowRewardOverwrite; //allow overwrite of the block reward that is set in genesis.json
+
 public:
 	EVMSchedule const& scheduleForBlockNumber(u256 const& _blockNumber) const;
 	u256 blockReward(EVMSchedule const& _schedule) const;
 	void setBlockReward(u256 const& _newBlockReward);
+	void setBlockRewardOvewrite(bool _ovewrite) { m_allowRewardOverwrite = _ovewrite; }
 	u256 maximumExtraDataSize = 1024;
 	u256 accountStartNonce = 0;
 	bool tieBreakingGas = true;
--- a/libethcore/ChainOperationParams.cpp
+++ b/libethcore/ChainOperationParams.cpp
@@ -43,6 +43,7 @@ PrecompiledContract::PrecompiledContract(
 
 ChainOperationParams::ChainOperationParams():
 	m_blockReward("0x4563918244F40000"),
+	m_allowRewardOverwrite(true),
 	minGasLimit(0x1388),
 	maxGasLimit("0x7fffffffffffffff"),
 	gasLimitBoundDivisor(0x0400),
@@ -73,7 +74,7 @@ EVMSchedule const& ChainOperationParams::scheduleForBlockNumber(u256 const& _blo
 
 u256 ChainOperationParams::blockReward(EVMSchedule const& _schedule) const
 {
-	if (_schedule.blockRewardOverwrite)
+	if (_schedule.blockRewardOverwrite && m_allowRewardOverwrite)
 		return *_schedule.blockRewardOverwrite;
 	else
 		return m_blockReward;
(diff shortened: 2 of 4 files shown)


pirapira:
What is the benefit of having a new format?

libethcore/ChainOperationParams.h
```diff
@@ -77,10 +77,13 @@ struct ChainOperationParams
 	/// General chain params.
 private:
 	u256 m_blockReward;
+	bool m_allowRewardOverwrite; //allow overwrite of the block reward that is set in genesis.json
```
gumb0: Why do you need this hack?

winsvega: without this, Byzantium blockreward is reset to non zero value even if I want to run Byzantium tests with 0 mining reward that I set in chain params.

winsvega:
@pirapira  this is a minimalistic genesis format for running a test.
besides every client has its own unique genesis format. it would be good to have a common genesis format for test executions via rpc.

libethcore/ChainOperationParams.h
gumb0: You're not running Byzantium rules then. Maybe create another config similar to Byzantium but without block reward?

gumb0:
Still not clear what's exactly wrong for you with our regular genesis format.

Provide an example of new format please

libethcore/ChainOperationParams.h
gumb0: You need to create new `EVMSchedule` in `EVMSchedule.h`, add it to `ChainOperationParams::scheduleForBlockNumber()`, then use new param like "byzantiumTestForkBlock" in actual chain config

winsvega: isn't it too much just for the test purpose?

winsvega:
so basically there should be setGenesis method which allows you to enable/disable or set mining algorythm (PoW, NoProof)  and initial state and genesis block header. 
Plus it should be possible to set chain rules. (homestead, eip158)

Perhaps there should be a config section that describes those chain rules. cause currently chain rules are hardcoded into the client.

libethcore/ChainOperationParams.h
gumb0: Well ad-hoc hacking lower-level data structures and mechanisms just for the test purpose seems too much to me

gumb0:
So is this a special kind of config where instead of fork block numbers you just set which fork it is and it is in effect for the whole chain starting from genesis?

libethcore/ChainOperationParams.h
winsvega: it is not really a hack. its an option to set mining reward in genesis config

winsvega:
yes and no. we define `Frontier` as a test network that has Frontier rules enabled by default and so on for other forks. but we also have FrontierToHomesteadAt5  which has a transition at block 5. 

so I see, you mean that chain options like forkBlock number should also be set in this method probably. 
thus there will be no need to set Frontier and EIP158 names. 

the issue is still that such transition block options are depend on a client.

libethcore/ChainOperationParams.h
gumb0: To make it work you could also try just overriding `NoProof::blockReward()` and always returning 0 from there.

gumb0: Though I wouldn't fully support such change, because it makes `NoProof` engine further away from real network mechanics and for unit tests I would like it to stay closer. But if you'd made another engine just for your tests like `NoProofZeroRewards` that would be fine I guess.

winsvega:
Looks like the current set_chainParams method. (chain param init function) needs refactoring. 
And we could support default values for the fields that are not set in .json

PR #4974, opened 2018-04-21
Add "gasUsed" data to the LocalisedTransactionReceipt class (fix #4956)

### LINKED ISSUE #4956: gasUsed value is incorrect in eth_getTransactionReceipt response
We currently return the same value for `gasUsed` and `cumulativeGasUsed`, see https://github.com/ethereum/cpp-ethereum/blob/a17f30a41aa2af506b6b453f052c53bd323a36e0/libweb3jsonrpc/JsonHelper.cpp#L203

For both values we return `TransactionReceipt::gasUsed()`, which is in fact cumulative gas.

We need to
1. rename `TransactionReceipt::gasUsed()` to `cumulativeGasUsed()`
2. add `gasUsed()` to `LocalizedTransactionReceipt`, calculate it by subtracting the previous transaction receipt's gas
3. fill it correctly in `toJson()` function linked above

halfalicious:
Fixes #4956 

Add "gasUsed" data (the amount of gas used by a transaction) to the LocalisedTransactionReceipt class so that one can access this data via the web3 APIs. Note that without these changes, one only has access to "cumulative gas used" data. As a part of these changes, I've also renamed the "gasUsed()" getter in the TransactionReceipt class to "cumulativeGasUsed()" so that its name more accurately reflects what's being returned.

I've verified my changes locally by retrieving a transaction receipt and verifying that gasUsed and cumulativeGasUsed are different, and that gasUsed accurately reflects the amount of gas used by the transaction.

### DIFF
--- a/libethereum/BasicGasPricer.cpp
+++ b/libethereum/BasicGasPricer.cpp
@@ -53,7 +53,7 @@ void BasicGasPricer::update(BlockChain const& _bc)
 			for (auto const& tr: r[1])
 			{
 				Transaction tx(tr.data(), CheckTransaction::None);
-				u256 gu = brs.receipts[i].gasUsed();
+				u256 gu = brs.receipts[i].cumulativeGasUsed();
 				dist[tx.gasPrice()] += gu;
 				total += gu;
 				i++;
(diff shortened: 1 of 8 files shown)


winsvega:
will see if it has effect on rpc tests after it will be merged

yann300:
looks good too

halfalicious:
Thanks guys!

PR #4975, opened 2018-04-24
Add support for eth_chainId

### LINKED ISSUE #4810: Support eth_chainId
See https://github.com/ethereum/EIPs/pull/695 and https://github.com/paritytech/parity/pull/6329 (merged into parity).

This would be useful for the upcoming ewasm testnet & tooling around it.

halfalicious:
Fixes #4810 
Implements https://github.com/ethereum/EIPs/blob/master/EIPS/eip-695.md

I added a `chainId()` function to the `libethereum\Interface` class which returns 0 by default (since chainId values start @ 0x1 with MainNet), along with the corresponding function to `libethereum\ClientBase` which retrieves the current chain id from `BlockChain::chainParams`. I then surfaced the function in the JSON-RPC API by adding it to `libweb3jsonrpc\eth.json` and regenerating `EthFace.h` via `jsonrpcstub`, and removing unnecessary changes from the generated file.

I tested my changes on Ubuntu with eth connected to Ropsten and a JSON-RPC server setup via the `jsonrpcproxy.py` script:

```
localadminuser@ELLIOTT:~$ curl -X POST --data '{"jsonrpc":"2.0","method":"eth_chainId","params":[],"id":1}' http://localhost:8545
{"id":1,"jsonrpc":"2.0","result":"0x3"}
localadminuser@ELLIOTT:~$ 
```

### DIFF
--- a/libethereum/ClientBase.h
+++ b/libethereum/ClientBase.h
@@ -174,7 +174,9 @@ public:
     }
 
     Block block(BlockNumber _h) const;
-
+    
+    virtual int chainId() const override;
+    
 protected:
     /// The interface that must be implemented in any class deriving this.
     /// {
--- a/libethereum/ClientBase.h
+++ b/libethereum/ClientBase.h
@@ -82,7 +82,7 @@ public:
     /// Estimate gas usage for call/create.
     /// @param _maxGas An upper bound value for estimation, if not provided default value of c_maxGasEstimate will be used.
     /// @param _callback Optional callback function for progress reporting
-    virtual std::pair<u256, ExecutionResult> estimateGas(Address const& _from, u256 _value, Address _dest, bytes const& _data, int64_t _maxGas, u256 _gasPrice, BlockNumber _blockNumber, GasEstimationCallback const& _callback) override;
+    std::pair<u256, ExecutionResult> estimateGas(Address const& _from, u256 _value, Address _dest, bytes const& _data, int64_t _maxGas, u256 _gasPrice, BlockNumber _blockNumber, GasEstimationCallback const& _callback) override;
 
     using Interface::balanceAt;
     using Interface::countAt;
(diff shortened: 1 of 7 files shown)


libethereum/ClientBase.h
```diff
@@ -174,7 +174,9 @@ class ClientBase: public Interface
     }
 
     Block block(BlockNumber _h) const;
-
+    
+    virtual int chainId() const override;
```
chfast: Do not repeat `virtual` keyword if `override` is used.

halfalicious: Thanks for the feedback @chfast! I followed the approach used by other functions in the class e.g.:

https://github.com/ethereum/cpp-ethereum/blob/3f2ac7c4f9e1543eb4f31cbdcdf8028eb9bfe0b6/libethereum/ClientBase.h#L130

After reading more about the virtual and override keywords, it looks like neither of them are required:
* Virtual: This keyword is redundant because ClientBase::chainId() is implicitly virtual because it overrides a virtual function
* Override: This keyword isn't required, but it can be help one avoid runtime bugs since it tells the compiler to verify that there's a function in the parent class with the same signature and to raise a compiler error if that's not the case

That said, I can also see the value in keeping both keywords in the interest of code consistency. What are your thoughts?

gumb0: Yes, `override` is helpful and having both `virtual` and `override` is redundant. I'm fine with making it less consistent and leaving only `override`.
You could also remove all other `virtual` from other overridden methods.

gumb0:
It would be great if you could also document how you generate stub files with jsonrpcstub, because usually no one can't make it work.

libethereum/ClientBase.h
halfalicious: Thanks @gumb0, I've removed virtual from all of the ClientBase functions which override other functions.

PR #4977, opened 2018-04-24
Refactor logging - fix operator<< overloads

gumb0:
0. Define overloads for `formatting_ostream` (char stream) instead of generic `basic_formatting_ostream` to simplify things a bit
1. Remove `operator<<` overloads for primitive types and `std::string` because they already exist in boost.log and we can't redefine them.
2. Add overloads for non-const references aside const references - without it `operator<<` with non-const references gets resolved into generic overload for `T&` defined in boost.log. I haven't found more elegant solution for this.
3. Overloads for the types of `std` namespace can't be defined in `dev` namespace because of Argument-Dependent Lookup. Putting them into `std` is forbidden, so I define them in `boost::log`

### DIFF
--- a/libdevcore/Log.h
+++ b/libdevcore/Log.h
@@ -282,132 +282,120 @@ inline Logger createLogger(int _severity, std::string const& _channel)
 
 
 // Log stream output operators to apply special formatting to the values sent to log
-template <typename CharT, typename TraitsT, typename AllocatorT>
-inline boost::log::basic_formatting_ostream<CharT, TraitsT, AllocatorT>& operator<<(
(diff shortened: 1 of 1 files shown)


gumb0:
I'd say `const_cast` emphasizes that we don't change the type, but only adding constness, but I'm fine with changing to `static_cast`

chfast:
Hm... maybe do a trick like this:
```
auto const& constValue = _value;
_strm.stream() << constValue;
```
We don't need casting at all.

gumb0:
Ok, this is more readable

gumb0:
@chfast fixed

PR #4978, opened 2018-04-24
Refactor logging - NodeTable

gumb0:
…g new logging. Replace logging in DiscoveryDatagram factory with global logger.


gumb0:
Some network tests are failing, it's probably not directly related to PR, but I'll investigate

PR #4979, opened 2018-04-24
EVMC: Passthrough status code

chfast:
It's not perfect, because some information is lost during status_code -> VM exception -> TransactionException enum -> status_code transition. I did MVP.


libevm/ExtVMFace.h
```diff
@@ -169,6 +169,33 @@ class EnvInfo
...
+/// Represents a CREATE result.
+///
+/// @todo: Replace with evmc_result in future.
+struct CreateResult
+{
+    evmc_status_code status = EVMC_INTERNAL_ERROR;
+    h160 address;
+    owning_bytes_ref output;
+
+    CreateResult(evmc_status_code status, h160 address, owning_bytes_ref&& output)
```
gumb0: Better pass 20 bytes of `h160` by const reference

libevm/ExtVMFace.h
```diff
@@ -169,6 +169,33 @@ class EnvInfo
     u256 m_gasUsed;
 };
 
+/// Represents a call result.
+///
+/// @todo: Replace with evmc_result in future.
+struct CallResult
+{
+    evmc_status_code status = EVMC_INTERNAL_ERROR;
```
gumb0: For alignment reasons yhis struct might be more compact if you put `enum` member after `owning_bytes_ref` member

gumb0: Also no real need to use inline initialization here as it's always initialized by a single constuctor

libevm/ExtVMFace.h
chfast: I think I wanted to use `std::move` here. But for address it does not make sense anyway.

libevm/ExtVMFace.h
chfast: Logically, the status should be first, because the rest of fields depend on value here.

chfast: I prefer to have every field initialized, but in this case I can skip it to avoid confusion.

chfast: I swapped address with output. It matches the order in evmc_result. Other suggestions also applied.

PR #4986, opened 2018-04-27
string test_getLogHash(txHash)

winsvega:
`string test_getLogHash(string _txHash)`
Get pre-calculated log has of the _txHash. hash of emptyRLPList if transaction not found.
Log hash could be calculated from transaction receipt, but some transactions has too many logs, so we need a method to get just the log hash.

looks like this would be an optional method. because it is not required for state tests to check log hash.

https://github.com/ethereum/retesteth/issues/7

### DIFF
--- a/libweb3jsonrpc/Test.cpp
+++ b/libweb3jsonrpc/Test.cpp
@@ -23,15 +23,51 @@
 #include "Test.h"
 #include <jsonrpccpp/common/errors.h>
 #include <jsonrpccpp/common/exception.h>
-#include <libethereum/ClientTest.h>
+#include <libdevcore/CommonJS.h>
 #include <libethereum/ChainParams.h>
+#include <libethereum/ClientTest.h>
 
 using namespace std;
 using namespace dev;
+using namespace dev::eth;
 using namespace dev::rpc;
 using namespace jsonrpc;
 
 Test::Test(eth::Client& _eth): m_eth(_eth) {}
+
+namespace
+{
+string logEntriesToLogHash(eth::LogEntries const& _logs)
+{
+    RLPStream s;
+    s.appendList(_logs.size());
+    for (eth::LogEntry const& l : _logs)
+        l.streamRLP(s);
+    return toHexPrefixed(sha3(s.out()));
+}
+}
+
+string Test::test_getLogHash(string const& _txHash)
+{
+    h256 txHash;
+    try
+    {
+        txHash = h256(_txHash);
+    }
+    catch (...)
+    {
+        BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS));
+    }
+
+    if (m_eth.blockChain().isKnownTransaction(txHash))
+    {
+        LocalisedTransaction t = m_eth.localisedTransaction(txHash);
+        BlockReceipts const& blockReceipts = m_eth.blockChain().receipts(t.blockHash());
+        if (blockReceipts.receipts.size() != 0)
+            return logEntriesToLogHash(blockReceipts.receipts[t.transactionIndex()].log());
+    }
+    return toJS(dev::EmptyListSHA3);
+}
 
 bool Test::test_setChainParams(Json::Value const& param1)
 {
--- a/libweb3jsonrpc/Test.cpp
+++ b/libweb3jsonrpc/Test.cpp
@@ -43,7 +43,7 @@ string logEntriesToLogHash(eth::LogEntries const& _logs)
     s.appendList(_logs.size());
     for (eth::LogEntry const& l : _logs)
         l.streamRLP(s);
-    return toHexPrefixed(sha3(s.out()));
+    return toJS(sha3(s.out()));
 }
 }
 
@@ -59,13 +59,21 @@ string Test::test_getLogHash(string const& _txHash)
         BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS));
     }
 
-    if (m_eth.blockChain().isKnownTransaction(txHash))
+    try
     {
-        LocalisedTransaction t = m_eth.localisedTransaction(txHash);
-        BlockReceipts const& blockReceipts = m_eth.blockChain().receipts(t.blockHash());
-        if (blockReceipts.receipts.size() != 0)
-            return logEntriesToLogHash(blockReceipts.receipts[t.transactionIndex()].log());
+        if (m_eth.blockChain().isKnownTransaction(txHash))
+        {
+                LocalisedTransaction t = m_eth.localisedTransaction(txHash);
+                BlockReceipts const& blockReceipts = m_eth.blockChain().receipts(t.blockHash());
+                if (blockReceipts.receipts.size() != 0)
+                    return logEntriesToLogHash(blockReceipts.receipts[t.transactionIndex()].log());
+        }
     }
+    catch(...)
+    {
+        BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INTERNAL_ERROR));
+    }
+
     return toJS(dev::EmptyListSHA3);
 }
 
--- a/libweb3jsonrpc/Test.cpp
+++ b/libweb3jsonrpc/Test.cpp
@@ -49,32 +49,32 @@ string logEntriesToLogHash(eth::LogEntries const& _logs)
 
 string Test::test_getLogHash(string const& _txHash)
 {
-    h256 txHash;
     try
     {
-        txHash = h256(_txHash);
-    }
-    catch (...)
-    {
-        BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS));
-    }
+        h256 txHash;
+        try
+        {
+            txHash = h256(_txHash);
+        }
+        catch (BadHexCharacter const&)
+        {
+            throw JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS);
+        }
 
-    try
-    {
         if (m_eth.blockChain().isKnownTransaction(txHash))
         {
-                LocalisedTransaction t = m_eth.localisedTransaction(txHash);
-                BlockReceipts const& blockReceipts = m_eth.blockChain().receipts(t.blockHash());
-                if (blockReceipts.receipts.size() != 0)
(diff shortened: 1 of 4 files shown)


libweb3jsonrpc/Test.cpp
```diff
@@ -25,14 +25,37 @@
...
+string Test::test_getLogHash(string const& _txHash)
+{
+    if (m_eth.blockChain().isKnownTransaction(h256(_txHash)))
+    {
+        LocalisedTransaction t = m_eth.localisedTransaction(h256(_txHash));
+        BlockReceipts const& blockReceipts = m_eth.blockChain().receipts(t.blockHash());
+        if (blockReceipts.receipts.size() != 0)
+            return exportLog(blockReceipts.receipts[t.transactionIndex()].log());
+     }
+     return toJS(dev::EmptyListSHA3);
```
pirapira: Indentation around here is a bit strange.  Did you run git-clang-format?

gumb0:
What about logs bloom filter in transaction receipt - couldn't you just use it instead of this hash?
It's just 256 bytes long.

winsvega:
current tests has log hash. 
ask @felix from the go team. he might sometimes be in the berlin office. 
He is mainly the one who uses this log hash information.

gumb0:
Ok, so it's a part of the filled test format, I see, ok then I guess we need it for client compatibility

libweb3jsonrpc/Test.cpp
```diff
@@ -23,16 +23,39 @@
...
 
 using namespace std;
 using namespace dev;
+using namespace dev::eth;
 using namespace dev::rpc;
 using namespace jsonrpc;
 
 Test::Test(eth::Client& _eth): m_eth(_eth) {}
 
+string exportLog(eth::LogEntries const& _logs)
```
gumb0: Put free functions into unnamed namespace

gumb0: "export" seems too generic in this context, I'd name it `logsToRlpListHex`

libweb3jsonrpc/Test.cpp
```diff
@@ -23,16 +23,39 @@
...
+string Test::test_getLogHash(string const& _txHash)
+{
+    if (m_eth.blockChain().isKnownTransaction(h256(_txHash)))
+    {
+        LocalisedTransaction t = m_eth.localisedTransaction(h256(_txHash));
+        BlockReceipts const& blockReceipts = m_eth.blockChain().receipts(t.blockHash());
+        if (blockReceipts.receipts.size() != 0)
+            return exportLog(blockReceipts.receipts[t.transactionIndex()].log());
+    }
+    return toJS(dev::EmptyListSHA3);
```
gumb0: Why should it be empty list in case transaction not found? This way the user of the method can't understand whether the transaction produced no log or it doesn't exist at all.

Maybe we should return error in case transaction not found?

libweb3jsonrpc/Test.cpp
```diff
@@ -23,16 +23,39 @@
...
+    RLPStream s;
+    s.appendList(_logs.size());
+    for (eth::LogEntry const& l : _logs)
+        l.streamRLP(s);
+    return toHexPrefixed(sha3(s.out()));
+}
+
+string Test::test_getLogHash(string const& _txHash)
+{
+    if (m_eth.blockChain().isKnownTransaction(h256(_txHash)))
```
gumb0: What happens when you send here invalid hash string? Does it crash or is appropriate error returned?

libweb3jsonrpc/Test.h
```diff
@@ -42,8 +42,8 @@ class Test: public TestFace
 	{
 		return RPCModules{RPCModule{"test", "1.0"}};
 	}
-
-	virtual bool test_setChainParams(const Json::Value &param1) override;
+    virtual std::string test_getLogHash(std::string const& param1) override;
```
gumb0: `_` prefix for parameters

libweb3jsonrpc/Test.cpp
winsvega:
```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<dev::BadHexCharacter>'
  what():  /home/wins/Ethereum/cpp-ethereum/libdevcore/CommonData.cpp(93): Throw in function dev::bytes dev::fromHex(const string&, dev::WhenError)
Dynamic exception type: boost::exception_detail::clone_impl<dev::BadHexCharacter>
```

gumb0: So you should catch it here and rethrow as `JsonRpcException`, then method will correctly return error.

libweb3jsonrpc/Test.cpp
gumb0: I think returning error is cleaner than magic value of hash.
You should also be able to make the error more specific by adding error message like `throw JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS, "Transaction not found")`

gumb0: Or you could introduce new error code and return it like `throw JsonRpcException(TxNotFoundErrorCode)`

Maybe check what geth returns from `eth_getTransactionByHash` in this case

winsvega: ideally all json requests response should have a common scheme that all clients follow.

libweb3jsonrpc/Test.cpp
```diff
@@ -23,16 +23,52 @@
...
 
+namespace
+{
+string logEntriesToLogHash(eth::LogEntries const& _logs)
+{
+    RLPStream s;
+    s.appendList(_logs.size());
+    for (eth::LogEntry const& l : _logs)
+        l.streamRLP(s);
+    return toHexPrefixed(sha3(s.out()));
```
gumb0: It would be the same result, but for consistency I'd use either `toHexPrefixed` or `toJS` in both places - here and in `return toJS(dev::EmptyListSHA3);`

libweb3jsonrpc/Test.cpp
```diff
@@ -23,16 +23,52 @@
...
+        txHash = h256(_txHash);
+    }
+    catch (...)
+    {
+        BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS));
+    }
+
+    if (m_eth.blockChain().isKnownTransaction(txHash))
+    {
+        LocalisedTransaction t = m_eth.localisedTransaction(txHash);
```
gumb0: This could throw, too (that would mean something is not right in DB), so I would wrap all of this `if` in `try..catch`, too and rethrow as `JsonRpcException(Errors::ERROR_RPC_INTERNAL_ERROR)`

winsvega: Yeap, I thought of that

libweb3jsonrpc/Test.cpp
```diff
@@ -23,16 +23,60 @@
...
+string Test::test_getLogHash(string const& _txHash)
+{
+    try
+    {
+        h256 txHash;
+        try
+        {
+            txHash = h256(_txHash);
+        }
+        catch (BadHexCharacter const&)
```
gumb0: Let's not mask all possible errors, but only catch the errors of parsing the  hexstring. If at some point (maybe in the future) it start to throw other exceptions, we won't fail with confusing "invalid params" error.

gumb0:
Oh wait I made a mistake, I didn't mean to throw an enum

winsvega:
you could squash the changes then and make a one finale commit

gumb0:
Didn't squash all of them, sorry

libweb3jsonrpc/Test.cpp
```diff
@@ -23,16 +23,60 @@
...
+        {
+            throw JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS);
```
gumb0: There was no any sense really in using `BOOST_THROW_EXCEPTION` for `JsonRpcException` which is not a boost exception

libweb3jsonrpc/Test.cpp
```diff
@@ -23,16 +23,60 @@
...
+            LocalisedTransaction t = m_eth.localisedTransaction(txHash);
+            BlockReceipts const& blockReceipts = m_eth.blockChain().receipts(t.blockHash());
+            if (blockReceipts.receipts.size() != 0)
+                return logEntriesToLogHash(blockReceipts.receipts[t.transactionIndex()].log());
+        }
+        return toJS(dev::EmptyListSHA3);
+    }
+    catch (std::exception const& ex)
+    {
+        cwarn << ex.what();
```
gumb0: When something goes wrong we don't want to just silently return "internal error", we'd like to have a way to understand what happened, so let's output the error to the log.

gumb0:
Ok now, but I still think returning empty list for "not found" is not the best design.

winsvega:
the function should return in all it's paths. some compilers does not allow not returning the value. 
is it ok if you use throw and no return after {} ? 

actually in many throw cases right now client will just silently return json rpc internall error without any comment.

gumb0:
> the function should return in all it's paths. some compilers does not allow not returning the value.
is it ok if you use throw and no return after {} ?

Sure it's allowed

> actually in many throw cases right now client will just silently return json rpc internall error without any comment.

That should be fixed

libweb3jsonrpc/Test.cpp
winsvega: _ex ?

winsvega: you could still use boost exception for nicer error output like this: 
`BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INTERNAL_ERROR, _ex.what()));`

emprisupriatna:
Embuh gak mudeng

libweb3jsonrpc/Test.cpp
gumb0: No
1. `BOOST_THROW_EXCEPTION` wraps your exception with additional info like filename and line in debug build, but it can be useful only if it's caught and accessed with boost.exception facilities. Here `JsonRpcException` is always caught by our JSON-RPC framework, which doesn't know anything about boost.exception and just transforms it into JSON response.
2. The second argument to `JsonRpcException` constructor will also be added to JSON response, but `what()` string is usually for our exceptions not user-friendly at all, I don't think it will look appropriate at the client side. We should just log it in our log.

gumb0: `ex` is not a function parameter, so no `_` prefix needed

PR #4988, opened 2018-04-27
 Dynamicaly load EVMC VM provided by --vm <path>

chfast:
This extends `--vm` option to accept also paths to shared libraries with EVMC VMs. E.g. `--vm libhera.so`.
I wanted separate loading plugins from selecting a VM, but program_options makes it super hard to do. So this is much simpler variant, but might be good enough.

Some code cleanups are possible.

### DIFF
--- a/libevm/VMFactory.cpp
+++ b/libevm/VMFactory.cpp
@@ -62,28 +62,22 @@ VMKindTableEntry vmKindsTable[] = {
     {VMKind::Hera, "hera"},
 #endif
 };
-}
 
-void validate(boost::any& v, const std::vector<std::string>& values, VMKind* /* target_type */, int)
+void setVMKind(const std::string& name)
 {
-    // Make sure no previous assignment to 'v' was made.
-    po::validators::check_first_occurrence(v);
-
-    // Extract the first string from 'values'. If there is more than
-    // one string, it's an error, and exception will be thrown.
-    const std::string& s = po::validators::get_single_string(values);
-
     for (auto& entry : vmKindsTable)
     {
         // Try to find a match in the table of VMs.
-        if (s == entry.name)
+        if (name == entry.name)
         {
-            v = entry.kind;
+            g_kind = entry.kind;
             return;
         }
     }
 
-    throw po::validation_error(po::validation_error::invalid_option_value);
+    BOOST_THROW_EXCEPTION(
+        po::validation_error(po::validation_error::invalid_option_value, "vm", name, 1));
+}
 }
 
 namespace
@@ -137,10 +131,10 @@ po::options_description vmProgramOptions(unsigned _lineLength)
     auto add = opts.add_options();
 
     add("vm",
-        po::value<VMKind>()
+        po::value<std::string>()
             ->value_name("<name>")
-            ->default_value(VMKind::Legacy, "legacy")
-            ->notifier(VMFactory::setKind),
+            ->default_value("legacy")
+            ->notifier(setVMKind),
         description.data());
 
     add(c_evmcPrefix,
--- a/libevm/VMFactory.cpp
+++ b/libevm/VMFactory.cpp
@@ -20,6 +20,9 @@
 #include "LegacyVM.h"
 #include "interpreter.h"
 
+#include <boost/dll.hpp>
+#include <boost/filesystem.hpp>
+
 #if ETH_EVMJIT
 #include <evmjit.h>
 #endif
@@ -28,7 +31,11 @@
 #include <hera.h>
 #endif
 
+namespace dll = boost::dll;
+namespace fs = boost::filesystem;
 namespace po = boost::program_options;
+
+using evmc_create_fn = evmc_instance*();
 
 namespace dev
 {
@@ -37,6 +44,7 @@ namespace eth
 namespace
 {
 auto g_kind = VMKind::Legacy;
+fs::path g_dllPath;
 
 /// A helper type to build the tabled of VM implementations.
 ///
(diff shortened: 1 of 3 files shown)


libevm/VMFactory.cpp
```diff
@@ -63,6 +71,27 @@ VMKindTableEntry vmKindsTable[] = {
 #endif
 };
 
+evmc_instance* createDllEVMC(const fs::path path)
+{
+    static auto createFn = [&] {
+        auto symbols = dll::library_info{path}.symbols();
+        auto it = std::find_if(symbols.begin(), symbols.end(),
+            [](const std::string& symbol) { return symbol.find("evmc_create_") == 0; });
```
axic: Will this stop on the first match?

Should this rather throw an error if there are multiple matches? (So it is consistent with what's loaded.)

chfast: Yes, it stops. To be consistent with what?
I would keep it simple. Still you don't have any guarantees that the function will be ABI compatible and everything will not explode.

axic: If there are two symbols with the `evmc_create_` prefix is it consistent which one will be picked up across platforms? However, if there are multiple matches I'd rather just bail out.

libevm/VMFactory.cpp
```diff
@@ -62,28 +70,49 @@ VMKindTableEntry vmKindsTable[] = {
     {VMKind::Hera, "hera"},
 #endif
 };
-}
 
-void validate(boost::any& v, const std::vector<std::string>& values, VMKind* /* target_type */, int)
+evmc_instance* createDllEVMC(const fs::path path)
```
gumb0: Better pass path by reference

gumb0: Also `_path`

libevm/VMFactory.cpp
```diff
@@ -62,28 +70,49 @@ VMKindTableEntry vmKindsTable[] = {
...
+        return dll::import<evmc_create_fn>(path, *it);
+    }();
 
-    // Extract the first string from 'values'. If there is more than
-    // one string, it's an error, and exception will be thrown.
-    const std::string& s = po::validators::get_single_string(values);
+    return createFn();
+}
 
+void setVMKind(const std::string& name)
```
gumb0: `std::string const& _name`

libevm/VMFactory.cpp
```diff
@@ -28,15 +31,20 @@
...
 
+using evmc_create_fn = evmc_instance*();
+
 namespace dev
 {
 namespace eth
 {
 namespace
 {
 auto g_kind = VMKind::Legacy;
```
gumb0: logically these two globals should be private static members of `VMFactory`

libevm/VMFactory.cpp
```diff
@@ -137,10 +166,10 @@ po::options_description vmProgramOptions(unsigned _lineLength)
...
 
     add("vm",
-        po::value<VMKind>()
-            ->value_name("<name>")
-            ->default_value(VMKind::Legacy, "legacy")
-            ->notifier(VMFactory::setKind),
+        po::value<std::string>()
+            ->value_name("<name>|<path>")
+            ->default_value("legacy")
+            ->notifier(setVMKind),
```
gumb0: Remove `VMFactory::setKind` f it's not used anymore

libevm/VMFactory.cpp
```diff
@@ -62,28 +70,49 @@ VMKindTableEntry vmKindsTable[] = {
...
 #endif
 };
-}
 
-void validate(boost::any& v, const std::vector<std::string>& values, VMKind* /* target_type */, int)
+evmc_instance* createDllEVMC(const fs::path path)
 {
-    // Make sure no previous assignment to 'v' was made.
-    po::validators::check_first_occurrence(v);
+    static auto createFn = [&] {
```
gumb0: I'd capture only `path&` explicitly

libevm/VMFactory.cpp
chfast: This is public API, but not used now. I will remove it because not thread-safe.

libevm/VMFactory.cpp
chfast: True, but then I will have to expose them in the header file. I prefer not to. However, not sure if the prefix should be `s_` or `g_` :)

gumb0: here `g_`; `s_` is for local statics

libevm/VMFactory.cpp
chfast: Addressed.

libevm/VMFactory.cpp
chfast: Removed.

chfast:
I have rewritten it a bit. Now we load the DLL when processing `--vm` option and pointer to EVMC create function (and I believe also DLL handle) is kept as `boost::function` object.

libevm/VMFactory.cpp
```diff
@@ -62,29 +75,58 @@ VMKindTableEntry vmKindsTable[] = {
...
+        if (_name == entry.name)
         {
-            v = entry.kind;
+            g_kind = entry.kind;
             return;
         }
     }
 
-    throw po::validation_error(po::validation_error::invalid_option_value);
+    // The name does not match any built-in VM, try to use it as a path to load a DLL VM.
```
gumb0: If you'd put the rest of the function into separate function, `setVMKind` itself would be simple and this comment might be not needed

chfast: I was thinking the same so let's do it.

libevm/VMFactory.cpp
axic: Thanks!

PR #4991, opened 2018-04-29
Organize the --help option output

johndebord:
The output of `eth --help` was rather messy, so I gave the it a more consistent and cleaner look.


johndebord:
Here's what it looks like if you want to compare quick.

help-option

eth/AccountManager.cpp
```diff
@@ -31,16 +31,16 @@ using namespace dev::eth;
...
 {
 	_out
-		<< "    account list  List all keys available in wallet.\n"
-		<< "    account new	Create a new key and add it to the wallet.\n"
-		<< "    account update [<uuid>|<address> , ... ]  Decrypt and re-encrypt given keys.\n"
-		<< "    account import [<uuid>|<file>|<secret-hex>]	Import keys from given source and place in wallet.\n";
+		<< "   account list                                List all keys available in wallet\n"
+		<< "   account new                                 Create a new key and add it to wallet\n"
+		<< "   account update [<uuid>|<address> , ... ]    Decrypt and re-encrypt keys\n"
+		<< "   account import [<uuid>|<file>|<secret-hex>] Import keys from file and place in wallet\n";
```
gumb0: But this should work not only for files, shouldn't it?

eth/MinerAux.h
```diff
@@ -223,23 +223,27 @@ class MinerCLI
...
-			<< "Mining configuration:" << endl
-			<< "    -C,--cpu  When mining, use the CPU." << endl
-			<< "    -t, --mining-threads <n> Limit number of CPU/GPU miners to n (default: use everything available on selected platform)" << endl
-			<< "    --current-block Let the miner know the current block number at configuration time. Will help determine DAG size and required GPU memory." << endl
-			<< "    --disable-submit-hashrate  When mining, don't submit hashrate to node." << endl;
+			<< "ETHASH VERIFY MODE:" << endl
+			<< "  -w,--check-pow <headerHash> <seedHash> <difficulty> <nonce> Check PoW credentials for validity" << endl
+			<< endl
+			<< "BENCHMARKING MODE:" << endl
+			<< "  -M,--benchmark               Benchmark for mining and exit; use with --cpu and --opencl" << endl
```
gumb0: We don't have `--opencl` anymore, maybe just remove the "use with" part

eth/MinerAux.h
```diff
@@ -223,23 +223,27 @@ class MinerCLI
 	static void streamHelp(ostream& _out)
 	{
 		_out
-			<< "Work farming mode:" << endl
-			<< "    --no-precompute  Don't precompute the next epoch's DAG." << endl
-			<< "Ethash verify mode:" << endl
-			<< "    -w,--check-pow <headerHash> <seedHash> <difficulty> <nonce>  Check PoW credentials for validity." << endl
+			<< "WORK FARMING MODE:" << endl
```
gumb0: Could you please also replace all `endl` in this function with `\n`

eth/main.cpp
```diff
@@ -773,9 +773,12 @@ int main(int argc, char** argv)
...
     {
         cout
-                << "Usage eth [OPTIONS]\n"
-                << "Options:\n\n"
-                << "Wallet usage:\n";
+			<< "NAME:\n"
+			<< "   eth - the cpp-ethereum command line interface\n\n"
+			<< "   Copyright 2013-2018 The cpp-ethereum Authors\n\n"
+			<< "USAGE:\n"
+			<< "   eth [options] command [command options] [arguments...]\n\n"
```
gumb0: I don't think this is accurate, we don't really have "commands" as geth does. I'd leave it like `eth [OPTIONS]`

eth/main.cpp
```diff
@@ -773,9 +773,12 @@ int main(int argc, char** argv)
     if (vm.count("help"))
     {
         cout
-                << "Usage eth [OPTIONS]\n"
-                << "Options:\n\n"
-                << "Wallet usage:\n";
+			<< "NAME:\n"
+			<< "   eth - the cpp-ethereum command line interface\n\n"
+			<< "   Copyright 2013-2018 The cpp-ethereum Authors\n\n"
```
gumb0: This would be the second place where we'd need to keep the year up-to-date. Could we somehow not duplicate it here and somehow reuse the string returned by `credits()` from `MinerAux.h` ?

johndebord:
I added an extra space to the `credits()` function in `MinerAux.h` to keep a consistent look, is that ok?

gumb0:
Please rebase

PR #4993, opened 2018-04-30
Refactor logging - minimize Log.h

gumb0:


gumb0:
@chfast clang6 and xcode90 are already at 8, I've increased gcc6 to 8 now

circle.yml
```diff
@@ -138,7 +138,7 @@ jobs:
       - CC:  gcc-6
       - GCOV: gcov-6
       - GENERATOR: Ninja
-      - BUILD_PARALLEL_JOBS: 3
+      - BUILD_PARALLEL_JOBS: 8
```
chfast: Oh, sorry. I think it was 3 before LOG changes. You can increase it to 4 to check, but that's not required if it was not decreased by LOG changes.

gumb0: Ok (i think it was decreased in the beginning but then I rebased and kept the value from `develop`)

gumb0:
Gcc6 still failed with 4 threads, I'll revert it

circle.yml
chfast: GCC cannot handle 8 for sure, mostly because it also produces debug symbols.

gumb0: It's back to 3 now

PR #5000, opened 2018-05-03
Refactor logging - get rid of ThreadContext

gumb0:


gumb0:
Ah now I realized it covers less than `ThreadContext` covered...

PR #5001, opened 2018-05-03
Refactor logging - get rid of ThreadContext

gumb0:


gumb0:
Found better and simpler approach: https://github.com/ethereum/cpp-ethereum/pull/5000

libp2p/Session.cpp
```diff
@@ -332,72 +335,73 @@ void Session::doRead()
...
-        catch (std::exception const& _e)
-        {
-            cnetwarn << "Exception decoding frame header RLP: " << _e.what() << " "
-                     << bytesConstRef(m_data.data(), h128::size).cropped(3);
-            drop(BadProtocol);
-            return;
-        }
+    ba::async_read(m_socket->ref(), boost::asio::buffer(m_data, h256::size),
+        [this, self](boost::system::error_code ec, std::size_t length) {
+            LOG_SCOPED_CONTEXT(m_logContext);
```
gumb0: clang-format reformatted it a lot, but the change is just replacing `ThreadContext` declarations with `LOG_SCOPED_CONTEXT`

Issue #4984, opened 2018-04-27
Refactor logging 


gumb0:
This PR accumulates all smaller PRs related to rewriting logging system using boost.log

- [x] Create basic logging primitives in `Log.h`
- [x] Custom `operator<<` for log stream to apply special formatting to logged values - this will replace `LogOutputStreamBase::append()` methods
- [x] Change all logging statements to use new logging
- [x] Rename `clogSimple` to `clog`
- [x] Replace `ThreadContext` mechanism with some additional log context attributes
- [x] Get rid of `VerbosityHolder`
- [x] `setupLogging()` in all executables
- [x] Get rid of global verbosity variable (make it argument to `setupLogging()` instead)
- [x] Create separate global logger for critical errors
- [x] Handle exceptions from logging system
- [x] new command line parameter for filtering channels (probably create program options group to handle it easily in all tools)
- [x] Limit used verbosities to 5 levels: Error, Warning, Info, Debug & Trace

PR #5002, opened 2018-05-03
Refactor logging - get rid of VerbosityHolder

gumb0:
Part of log refactoring #4984

`VerbosityHolder` was used for temporary changing the log verbosity. Currently it's used only in unit tests, not sure what for, and just removing it doesn't affect anything at least at default verbosity.
(It might be that it improves output of testeth for these tests at higher verbosity, but then it should be done some other way. CI doesn't use higher verbosity and me neither and I don't feel like untangling it right now)

Also I removed the rest of `--nonetwork` flag support from `testeth`.

### DIFF
--- a/libdevcore/Log.h
+++ b/libdevcore/Log.h
@@ -39,14 +39,6 @@ namespace dev
 /// The logging system's current verbosity.
 extern int g_logVerbosity;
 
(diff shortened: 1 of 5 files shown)


gumb0:
AppVeyor doesn't move for some reason...

PR #5006, opened 2018-05-07
eth command line options to filter log by channel

gumb0:
Part of log refactoring #4984

I went with specifying channels as a space-separated list because it's more natural for boost.program_options.
So the command line can be like
```
./eth -v 10 --log-channels host client warn bq --log-exclude-channels net discov
```

Program options group to reuse in other tools is not implemented yet, probably will go to separate PR.

### DIFF
--- a/eth/main.cpp
+++ b/eth/main.cpp
@@ -185,8 +185,6 @@ int main(int argc, char** argv)
     Ethash::init();
     NoProof::init();
 
-    int logVerbosity = 1;
-
     /// Operating mode.
     OperationMode mode = OperationMode::Node;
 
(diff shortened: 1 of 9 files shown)


gumb0:
I'd like to postpone supporting it in other tools for now, it's not so critical to filter the logs there.

eth/main.cpp
```diff
@@ -390,12 +388,28 @@ int main(int argc, char** argv)
     addImportExportOption("import-snapshot", po::value<string>()->value_name("<path>"),
         "Import blockchain and state data from the Parity Warp Sync snapshot.\n");
 
+    int logVerbosity = 1;
+    std::vector<std::string> includeLogChannels;
+    std::vector<std::string> excludeLogChannels;
+    po::options_description loggingOptions("Logging Options", c_lineWidth);
```
chfast: Is this shared by other tools? Probably should be.

Issue #4216, opened 2017-07-03
testeth should warn or error if --statediff is specified but verbosity is below 2.


pirapira:
The `testeth` option `--statediff` works only when `--verbosity 2` or a higher verbosity is specified.  When the `--statediff` option does not work, `testeth` should warn or error out.

pirapira:
@winsvega do you prefer a warning or an error?

pirapira:
I learned that `--statediff` option also writes something in the filled test.  This works without the `--verbosity` option.  So close the issue.

winsvega:
--statediff writes to the filled test if --filltests is set. 
this work as another debug option if you want to get statediff into a file, not just the cmd output (which might be hard to read)
same here, statediff should set verbosity itself. so you see the output (when used without --filltests)

PR #5011, opened 2018-05-09
fix verbosity for testeth

winsvega:
add another level of notes
0 - no output
1 - test progress output
2 - test progress + test names

### DIFF
--- a/test/tools/libtesteth/Options.cpp
+++ b/test/tools/libtesteth/Options.cpp
@@ -20,14 +20,18 @@
 
 #include <libevm/VMFactory.h>
 #include <libweb3jsonrpc/Debug.h>
-#include <test/tools/libtesteth/Options.h>
 #include <test/tools/fuzzTesting/fuzzHelper.h>
+#include <test/tools/libtesteth/Options.h>
+#include <boost/iostreams/device/null.hpp>
+#include <boost/iostreams/stream.hpp>
 #include <boost/program_options.hpp>
 
 using namespace std;
 using namespace dev::test;
 using namespace dev::eth;
 
+namespace
+{
 void printHelp()
 {
     cout << "Usage: \n";
@@ -72,6 +76,25 @@ void printVersion()
 {
     cout << prepareVersionString() << "\n";
 }
+}
+
+void Options::setVerbosity(int _level)
+{
+    static boost::iostreams::stream<boost::iostreams::null_sink> nullOstream(
+        (boost::iostreams::null_sink()));
+    dev::LoggingOptions logOptions;
+    verbosity = _level;
+    logOptions.verbosity = verbosity;
+    if (_level <= 0)
+    {
+        logOptions.verbosity = -1;
+        std::cout.rdbuf(nullOstream.rdbuf());
+        std::cerr.rdbuf(nullOstream.rdbuf());
+    }
+    else if (_level == 1 || _level == 2)
+        logOptions.verbosity = -1;
+    dev::setupLogging(logOptions);
+}
 
 Options::Options(int argc, const char** argv)
 {
@@ -100,7 +123,6 @@ Options::Options(int argc, const char** argv)
     trGasIndex = -1;
     trValueIndex = -1;
     bool seenSeparator = false; // true if "--" has been seen.
-    int verbosity = 5;
     for (auto i = 0; i < argc; ++i)
     {
         auto arg = std::string{argv[i]};
@@ -145,7 +167,7 @@ Options::Options(int argc, const char** argv)
         {
 #if ETH_VMTRACE
             vmtrace = true;
-            verbosity = 13;
+            verbosity = std::max(verbosity, 13);
 #else
             cerr << "--vmtrace option requires a build with cmake -DVMTRACE=1\n";
             exit(1);
@@ -207,22 +229,7 @@ Options::Options(int argc, const char** argv)
         else if (arg == "--verbosity")
         {
             throwIfNoArgumentFollows();
-            static std::ostringstream strCout; //static string to redirect logs to
-            std::string indentLevel = std::string{argv[++i]};
-            if (indentLevel == "0")
-            {
-                logVerbosity = Verbosity::None;
-                std::cout.rdbuf(strCout.rdbuf());
-                std::cerr.rdbuf(strCout.rdbuf());
-            }
-            else if (indentLevel == "1")
-                logVerbosity = Verbosity::NiceReport;
-            else
-                logVerbosity = Verbosity::Full;
-
-            int indentLevelInt = atoi(argv[i]);
-            if (indentLevelInt > verbosity)
-                verbosity = indentLevelInt;
+            verbosity = std::max(verbosity, atoi(argv[++i]));
         }
         else if (arg == "--options")
         {
@@ -263,7 +270,10 @@ Options::Options(int argc, const char** argv)
             testpath = std::string{argv[++i]};
         }
         else if (arg == "--statediff")
+        {
             statediff = true;
+            verbosity = std::max(verbosity, 5);
+        }
         else if (arg == "--randomcode")
         {
             throwIfNoArgumentFollows();
@@ -332,13 +342,8 @@ Options::Options(int argc, const char** argv)
                     "--seed <uint> could be used only with --createRandomTest \n"));
     }
 
-    //Default option
-    if (logVerbosity == Verbosity::NiceReport)
-        verbosity = -1;  // disable cnote but leave cerr and cout
-
-    LoggingOptions logginOptions;
-    logginOptions.verbosity = verbosity;
-    setupLogging(logginOptions);
+    // If no verbosity is set. use default
+    setVerbosity(verbosity == -1 ? 1 : verbosity);
 }
 
 Options const& Options::get(int argc, const char** argv)
--- a/test/tools/libtesteth/Options.h
+++ b/test/tools/libtesteth/Options.h
@@ -29,18 +29,12 @@ namespace dev
 namespace test
 {
 
-enum class Verbosity
-{
-    None,
-    NiceReport,
-    Full
-};
-
 DEV_SIMPLE_EXCEPTION(InvalidOption);
 
 class Options
 {
 public:
+    int verbosity = -1;     ///< Verbosity level
     bool vmtrace = false;	///< Create EVM execution tracer
     bool filltests = false; ///< Create JSON test files from execution results
     bool fillchain = false; ///< Fill tests as a blockchain tests if possible
@@ -55,7 +49,6 @@ public:
     bool jsontrace = false; ///< Vmtrace to stdout in json format
     eth::StandardTrace::DebugOptions jsontraceOptions; ///< output config for jsontrace
     std::string testpath;	///< Custom test folder path
-    Verbosity logVerbosity = Verbosity::NiceReport;
     boost::optional<boost::filesystem::path> randomCodeOptionsPath; ///< Options for random code generation in fuzz tests
 
     /// Test selection
@@ -77,6 +70,7 @@ public:
 private:
     Options(int argc = 0, const char** argv = 0);
     Options(Options const&) = delete;
+    void setVerbosity(int _level);
 };
 
 } //namespace test
--- a/test/tools/libtesteth/TestSuite.cpp
+++ b/test/tools/libtesteth/TestSuite.cpp
@@ -232,11 +232,11 @@ void TestSuite::executeTest(string const& _testFolder, fs::path const& _testFile
 	}
 
 	// Test is generated. Now run it and check that there should be no errors
-	if ((Options::get().singleTest && Options::get().singleTestName == testname) || !Options::get().singleTest)
-		cnote << "TEST " << testname << ":";
+    if (Options::get().verbosity > 1)
+        std::cout << "TEST " << testname << ":\n";
 
-	Listener::notifySuiteStarted(testname);	// Outdated logging
-	executeFile(boostTestPath);
+    Listener::notifySuiteStarted(testname);  // Outdated logging
+    executeFile(boostTestPath);
 }
 
 void TestSuite::executeFile(boost::filesystem::path const& _file) const
--- a/test/tools/libtesteth/TestOutputHelper.cpp
+++ b/test/tools/libtesteth/TestOutputHelper.cpp
@@ -52,7 +52,6 @@ bool TestOutputHelper::checkTest(std::string const& _testName)
 	if (test::Options::get().singleTest && test::Options::get().singleTestName != _testName)
 		return false;
 
-	cnote << _testName;
 	m_currentTestName = _testName;
 	return true;
 }
--- a/test/tools/libtesteth/boostTest.cpp
+++ b/test/tools/libtesteth/boostTest.cpp
@@ -180,6 +180,7 @@ int main(int argc, const char* argv[])
         framework::master_test_suite().add(ts1);
     }
 
+    std::cout << "Running tests using path: " << test::getTestPath() << std::endl;
     int result = 0;
     auto fakeInit = [](int, char* []) -> boost::unit_test::test_suite* { return nullptr; };
     if (opt.jsontrace || opt.vmtrace || opt.statediff)


libdevcore/Log.h
```diff
@@ -73,10 +73,15 @@ BOOST_LOG_INLINE_GLOBAL_LOGGER_CTOR_ARGS(g_debugLogger,
...
     boost::log::sources::severity_channel_logger_mt<>,
-    (boost::log::keywords::severity = 1)(boost::log::keywords::channel = "note"))
-#define cnote LOG(dev::g_noteLogger::get())
+    (boost::log::keywords::severity = 2)(boost::log::keywords::channel = "note2"))
+#define cnote LOG(dev::g_note2Logger::get())
+
+BOOST_LOG_INLINE_GLOBAL_LOGGER_CTOR_ARGS(g_note3Logger,
+    boost::log::sources::severity_channel_logger_mt<>,
+    (boost::log::keywords::severity = 3)(boost::log::keywords::channel = "note3"))
+#define cnote3 LOG(dev::g_note3Logger::get())
```
gumb0: I'd like to avoid additional global loggers. I'd like to minimize them in general and maybe will remove `cnote` at some point, too.

Global loggers here pollute `dev` namespace and they have an overhead for being thread-safe. It's better to use loggers more local to the task

libdevcore/Log.h
```diff
@@ -73,10 +73,15 @@ BOOST_LOG_INLINE_GLOBAL_LOGGER_CTOR_ARGS(g_debugLogger,
     (boost::log::keywords::severity = 0)(boost::log::keywords::channel = "debug"))
 #define cdebug LOG(dev::g_debugLogger::get())
 
-BOOST_LOG_INLINE_GLOBAL_LOGGER_CTOR_ARGS(g_noteLogger,
+BOOST_LOG_INLINE_GLOBAL_LOGGER_CTOR_ARGS(g_note2Logger,
     boost::log::sources::severity_channel_logger_mt<>,
-    (boost::log::keywords::severity = 1)(boost::log::keywords::channel = "note"))
-#define cnote LOG(dev::g_noteLogger::get())
+    (boost::log::keywords::severity = 2)(boost::log::keywords::channel = "note2"))
```
gumb0: You are changing with this a lot of messages going into `cnote`. I'm not sure it's a good idea. What are you trying to achieve?

test/tools/libtesteth/Options.h
```diff
@@ -77,6 +69,7 @@ class Options
 private:
     Options(int argc = 0, const char** argv = 0);
     Options(Options const&) = delete;
+    void setVerbosity(size_t _level);
```
gumb0: It could be static or better just free function in unnamed namespace in Options.cpp

test/tools/libtesteth/Options.cpp
```diff
@@ -263,7 +267,10 @@ Options::Options(int argc, const char** argv)
             testpath = std::string{argv[++i]};
         }
         else if (arg == "--statediff")
+        {
             statediff = true;
+            setVerbosity(5);
```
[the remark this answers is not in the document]
winsvega: a static variable inside that function should handle that

libethereum/BlockChain.cpp
```diff
@@ -225,7 +225,7 @@ unsigned BlockChain::open(fs::path const& _path, WithExisting _we)
         DEV_IGNORE_EXCEPTIONS(lastMinor = (unsigned)RLP(status));
     if (c_minorProtocolVersion != lastMinor)
     {
-        cnote << "Killing extras database (DB minor version:" << lastMinor << " != our miner version: " << c_minorProtocolVersion << ").";
+        cnote3 << "Killing extras database (DB minor version:" << lastMinor << " != our miner version: " << c_minorProtocolVersion << ").";
```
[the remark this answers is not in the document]
winsvega: I use cnote to print test names. but I don't need any other output but those testnames. 
which logchannel should I use?

test/tools/libtesteth/Options.cpp
```diff
@@ -73,6 +73,25 @@ void printVersion()
     cout << prepareVersionString() << "\n";
 }
 
+void Options::setVerbosity(size_t _level)
+{
+    static int verbosity = -1;
```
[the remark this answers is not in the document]
winsvega: forgot about that.

libdevcore/Log.h
winsvega: I think in this case I could just remote those debug traces from blockchain

test/tools/libtesteth/Options.h
gumb0: Why do you make verbosity parameter `size_t`? Make it `int` like in the rest of the codebase

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es106-dont-try-to-avoid-negative-values-by-using-unsigned

libdevcore/Log.h
winsvega: I need another log level for test output. 
0) no output
1) boost output
2) test output (uses cnote)
3) cnote ouput
4) ctrace output

gumb0: Why are you ignoring in this all other channels, what should happen to them?

What's your idea about distinction between cnote and ctrace? What makes them different, what's the purpose of each?

> test output (uses cnote)

I would propose for these messages to go into separate test-specific channel/logger.

gumb0: (If you need just specific channels it would be much easier and better instead of dancing around verbosity to setup a special filter based on channels)

gumb0:
General note: from a testeth user perspective I wouldn't like a verbosity to be silently changed to some untransparent value in case of `--vmtrace` or `--statediff`

I would prefer testeth to give out a warning like `You need to set verbosity to 13 to see vmtrace` when I run it with `--vmtrace` but don't give verbosity.
(I think it's actually 11)

Maybe worse but acceptable: output a message like "Changing verbosity to 13 for VM trace" when you do it, for user to better understand what's happening

libethereum/BlockChain.cpp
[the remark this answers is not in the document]
gumb0: Better to have a separate channel for that. Where do you do that?

winsvega:
there was an issue about changing verbosity level with those options. otherwise it does not produce output if you dont set the verbosity

gumb0:
I know, I propose to make this more transparent to the user

libdevcore/Log.h
winsvega: boost output to std output. what I need is 
0)  no output at all 
1) stderr + boost output
2) stderr + boost output + more output from test helper
3) all + verbosity levels

so before I was not using cnote and ctrace. so you say you going to delete those. then I could just use the verbosity and std::cout  + if

gumb0:
Or actually for vmtrace and statediff you also could just setup a logging filter for specific channels, that's better

libdevcore/Log.h
gumb0: `std::cout` + `if` is a bad idea if I understand it correctly. boost.log is awesome at filtering your logs for you, so let's use it for this and not have an `if`s  with each log message.

So if I understand your needs correctly, you could
instead of `cnote` and `ctrace` use 2 new test-specific channels. It could be one channel with a name `test` and macros to log messages with 2 different verbosities.

for 0 & 1 - disable all boost.log logs
for 2 - enable only `test` channel
for 3 - enable channels based on verbosity (call `setupLogging` like you do now)

pirapira:
Yes, when `--vmtrace` is set, the vm trace should appear whatever the verbosity is.

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -232,10 +232,10 @@ void TestSuite::executeTest(string const& _testFolder, fs::path const& _testFile
 	}
 
 	// Test is generated. Now run it and check that there should be no errors
-	if ((Options::get().singleTest && Options::get().singleTestName == testname) || !Options::get().singleTest)
-		cnote << "TEST " << testname << ":";
+    if (Options::get().verbosity > 1)
+        std::cout << "TEST " << testname << ":\n";
```
gumb0: It's only this single place for which you first changed verbosity of all `cnote`? Ok then.

winsvega:
https://github.com/ethereum/cpp-ethereum/issues/4216
Do you really think it is better to print a warning? 
1) Print a warning
type a command
get a warning 
type a command + verbosity 
2) correct verbosity
type a command

test/tools/libtesteth/Options.cpp
```diff
@@ -72,6 +74,26 @@ void printVersion()
...
+        if (_level <= 0)
+        {
+            dev::setupLogging(-1);
+            std::cout.rdbuf(strCout.rdbuf());
+            std::cerr.rdbuf(strCout.rdbuf());
+        }
+        else if (_level == 1 || _level == 2)
+            dev::setupLogging(-1);
+        else
+            dev::setupLogging(_level);
```
gumb0: So there's no way for the user of testeth to set cpp logs verbosity to 0, 1 or 2, is it ok for you?

gumb0:
> Do you really think it is better to print a warning?

No, now I think it's better just to ignore verbosity and just output the trace when user asks for it.

test/tools/libtesteth/Options.cpp
```diff
@@ -72,6 +74,26 @@ void printVersion()
 {
     cout << prepareVersionString() << "\n";
 }
+}
+
+void Options::setVerbosity(int _level)
+{
+    static std::ostringstream strCout;  // static string to redirect logs to
+    if (_level > verbosity)
```
gumb0: I think the code was / would be much simpler when you don't need this reentrance of `setVerbosity` support and call it just once in the end of `Options` constructor.

gumb0:
Also please rebase,  `setupLogging` now has another type of parameter

test/tools/libtesteth/Options.cpp
winsvega: then this if will be each time you set a verbosity

gumb0: You should `setVerbosity` only once

test/tools/libtesteth/Options.cpp
winsvega: usually the user of testeth is me ) 
by default testeth output logs for travis. it is like it has its own verbosity reserved for 0 1 and 2.

gumb0:
If you don't need any other cpp-ethereum logs when vmtrace or statediff is requested, then it should be possible just to 
```
LoggingOptions options;
options.verbosity = 11; // 5 for statediff
options.includeChannels = {"vmtrace"}; // "state" for statediff
setupLogging(options);
```

test/tools/libtesteth/Options.cpp
```diff
@@ -72,6 +74,32 @@ void printVersion()
...
+    static std::ostringstream strCout;  // static string to redirect logs to
+    static dev::LoggingOptions logOptions;
+    if (_level > verbosity)
+    {
+        verbosity = _level;
+        logOptions.verbosity = verbosity;
+        if (_level <= 0)
+        {
+            logOptions.verbosity = -1;
+            dev::setupLogging(logOptions);
```
gumb0: Now you can call this just once in the end of the function, instead of in each branch

winsvega:
No I guess its better be all logs included.

test/tools/libtesteth/Options.cpp
```diff
@@ -72,6 +74,27 @@ void printVersion()
...
+    static dev::LoggingOptions logOptions;
+    if (_level > verbosity)
+    {
+        verbosity = _level;
+        logOptions.verbosity = verbosity;
+        if (_level <= 0)
+        {
+            logOptions.verbosity = -1;
+            std::cout.rdbuf(strCout.rdbuf());
+            std::cerr.rdbuf(strCout.rdbuf());
```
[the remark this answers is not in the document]
gumb0: I think it accumulates all the logs in the string in memory, never throws them out.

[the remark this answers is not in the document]
winsvega: thats the way I found to disable all logs from output

test/tools/libtesteth/Options.cpp
```diff
@@ -72,6 +74,27 @@ void printVersion()
...
+        }
+        else if (_level == 1 || _level == 2)
+            logOptions.verbosity = -1;
+        dev::setupLogging(logOptions);
+    }
+}
```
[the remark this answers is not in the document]
winsvega: what do you mean by tricks here?

test/tools/libtesteth/Options.cpp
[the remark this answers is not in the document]
pirapira: Something called a null stream exists https://stackoverflow.com/a/8244321/103518 .

test/tools/libtesteth/Options.cpp
```diff
@@ -72,6 +74,27 @@ void printVersion()
...
+        logOptions.verbosity = verbosity;
+        if (_level <= 0)
+        {
+            logOptions.verbosity = -1;
+            std::cout.rdbuf(strCout.rdbuf());
+            std::cerr.rdbuf(strCout.rdbuf());
+        }
+        else if (_level == 1 || _level == 2)
+            logOptions.verbosity = -1;
+        dev::setupLogging(logOptions);
```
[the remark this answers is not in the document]
winsvega: the message is not being output twice. 
if you want to call setVerbosity once. then there is no need in setVerbosity. 
I just copy if (verbosity > x)  to all places where it was called and then call setupLogging

test/tools/libtesteth/Options.cpp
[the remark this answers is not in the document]
pirapira: * Translating some numbers into -1.
* changing the standard output.

test/tools/libtesteth/Options.cpp
```diff
@@ -332,13 +342,8 @@ Options::Options(int argc, const char** argv)
...
 
-    //Default option
-    if (logVerbosity == Verbosity::NiceReport)
-        verbosity = -1;  // disable cnote but leave cerr and cout
-
-    LoggingOptions logginOptions;
-    logginOptions.verbosity = verbosity;
-    setupLogging(logginOptions);
+    // If no verbosity is set. use default
+    setVerbosity(1);
```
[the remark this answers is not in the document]
winsvega: default is 1 
but testeth default is -1. because you should be able to set verbosity to 0

test/tools/libtesteth/Options.cpp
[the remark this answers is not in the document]
winsvega: -1 means disable cpp log output

test/tools/libtesteth/Options.cpp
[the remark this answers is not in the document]
gumb0: Yeah right, that's what I meant, I think it would be simpler (I'd still keep `setVerbosity` for the sake of smaller functions)
But ok, it's up to you,

test/tools/libtesteth/Options.cpp
```diff
@@ -23,11 +23,15 @@
 #include <test/tools/libtesteth/Options.h>
 #include <test/tools/fuzzTesting/fuzzHelper.h>
 #include <boost/program_options.hpp>
+#include "boost/iostreams/stream.hpp"
```
gumb0: Sorry one more thing: this should be in angle brackets; and run clang-format

test/tools/libtesteth/Options.cpp
```diff
@@ -72,6 +76,24 @@ void printVersion()
 {
     cout << prepareVersionString() << "\n";
 }
+}
+
+void Options::setVerbosity(int _level)
+{
+    static boost::iostreams::stream< boost::iostreams::null_sink > nullOstream( ( boost::iostreams::null_sink() ) );
+    static dev::LoggingOptions logOptions;
```
gumb0: This does not need to be static now

test/tools/libtesteth/Options.cpp
```diff
@@ -72,6 +76,24 @@ void printVersion()
 {
     cout << prepareVersionString() << "\n";
 }
+}
+
+void Options::setVerbosity(int _level)
+{
+    static boost::iostreams::stream< boost::iostreams::null_sink > nullOstream( ( boost::iostreams::null_sink() ) );
```
gumb0: Did clang-format leave all these extra spaces?

pirapira:
Does anybody know why AppVeyor is failing?  Timeout?

gumb0:
@pirapira Looke like timeout
```
Build execution time has reached the maximum allowed time for your plan (60 minutes).
```
But tests pass.

pirapira:
OK.

PR #5016, opened 2018-05-11
Use Cable and make a package

chfast:


.bumpversion.cfg
```diff
@@ -0,0 +1,21 @@
+[bumpversion]
```
gumb0: What does this file do?

libwebthree/WebThree.cpp
```diff
@@ -56,13 +56,12 @@ WebThreeDirect::WebThreeDirect(std::string const& _clientVersion,
...
 
-        string bp = DEV_QUOTED(ETH_BUILD_PLATFORM);
-        vector<string> bps;
-        boost::split(bps, bp, boost::is_any_of("/"));
-        bps[0] = bps[0].substr(0, 5);
-        bps[1] = bps[1].substr(0, 3);
-        bps.back() = bps.back().substr(0, 3);
-        m_ethereum->setExtraData(rlpList(0, string(dev::Version) + "++" + string(DEV_QUOTED(ETH_COMMIT_HASH)).substr(0, 4) + (ETH_CLEAN_REPO ? "-" : "*") + string(DEV_QUOTED(ETH_BUILD_TYPE)).substr(0, 1) + boost::join(bps, "/")));
+        const auto* buildinfo = eth_get_buildinfo();
+        m_ethereum->setExtraData(rlpList(0, string{buildinfo->project_version}.substr(0, 5) + "++" +
```
gumb0: We've lost `ETH_CLEAN_REPO` here? Do you know whether it worked before?

.bumpversion.cfg
chfast: It's config for [bumpversion](https://github.com/peritus/bumpversion) tool.
It defines how version numbers are progressing and use can use the tool to release a new version: it will update relevant places and create a git commit and tag.

Here it is: 1.4.0.dev0 ->  1.4.0.dev1 -> ... -> 1.4.0rc0 -> 1.4.0rc1 -> ... -> 1.4.0.
I don't like `rc0` so usually I manually override it.

libwebthree/WebThree.cpp
gumb0: These magic values for substr lengths look like they will break the first time some of the formats of build info strings changes.
Maybe we should have a function in buildinfo lib that produces a compact string that includes all info?

Also, extraData is 32 bytes I think, we could fit there more symbols if needed.

chfast: Yes, we lost it. I believe it was working before.
The dirty part is included in `project_version`, because it looks like `1.4.0.dev0-N+commit.XXXXXXXX-dirty`. But here this "project version" is trimmed to 5 chars.
I can add the "dirty" information to `buildinfo` struct similarly as there is `git_commit_hash`. But I don't think this information is super important.

chfast: I agree the extra data here is messy and fragile. Still `string{...}.substr(0, x)` pattern should always work. I would also get rid of RLP here.
I didn't wanted to touch it in this PR.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -624,13 +625,7 @@ void requireJsonFields(json_spirit::mObject const& _o, string const& _section,
...
 string prepareVersionString()
 {
-    // cpp-1.3.0+commit.6be76b64.Linux.g++
-    string commit(DEV_QUOTED(ETH_COMMIT_HASH));
-    string version = "cpp-" + string(ETH_PROJECT_VERSION);
-    version += "+commit." + commit.substr(0, 8);
-    version +=
-        "." + string(DEV_QUOTED(ETH_BUILD_OS)) + "." + string(DEV_QUOTED(ETH_BUILD_COMPILER));
-    return version;
+    return string{"testeth "} + eth_get_buildinfo()->project_version;
```
gumb0: It was useful for Dimitry to have exact commit here, I think it writes it to the filled test files.
cc @winsvega

chfast: Ok, let me know what exactly is needed.

gumb0: Ah `project_version` it includes commit already?
Ok then

gumb0: The comment you deleted here showed how it looks in test files.
I think commit and "cpp" are the most important parts

chfast: Not always. Only if the git is not directly on a version tag. https://github.com/ethereum/cable/blob/master/buildinfo/buildinfo.cmake#L32-L49

For tagged releases the commit hash is omitted.

gumb0: Probably that's fine

PR #5024, opened 2018-05-15
testeth touch coinbase in state tests

winsvega:
fix testeth to pass state tests as retesteth https://github.com/ethereum/tests/pull/437

### DIFF
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -292,8 +292,7 @@ std::tuple<eth::State, ImportTest::ExecOutput, eth::ChangeLog> ImportTest::execu
         {
             // Touch here bacuse coinbase might be suicided above
             initialState.addBalance(_env.author(), 0);  // imitate mining reward
-            initialState.commit(removeEmptyAccounts ? State::CommitBehaviour::RemoveEmptyAccounts :
-                                                      State::CommitBehaviour::KeepEmptyAccounts);
+            initialState.commit(State::CommitBehaviour::KeepEmptyAccounts);
         }
         return std::make_tuple(initialState, out, changeLog);
     }
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -266,6 +266,10 @@ std::tuple<eth::State, ImportTest::ExecOutput, eth::ChangeLog> ImportTest::execu
     try
     {
         unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(_sealEngineNetwork)).createSealEngine());
+        bool removeEmptyAccounts = m_envInfo->number() >= se->chainParams().EIP158ForkBlock;
+        if (!removeEmptyAccounts)
+            initialState.addBalance(_env.author(), 0);  // touch coinbase no matter what
+
         if (Options::get().jsontrace)
         {
             StandardTrace st;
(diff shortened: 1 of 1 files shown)


test/tools/libtesteth/ImportTest.cpp
```diff
@@ -266,6 +266,10 @@ std::tuple<eth::State, ImportTest::ExecOutput, eth::ChangeLog> ImportTest::execu
     try
     {
         unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(_sealEngineNetwork)).createSealEngine());
+        bool removeEmptyAccounts = m_envInfo->number() >= se->chainParams().EIP158ForkBlock;
+        if (!removeEmptyAccounts)
+            initialState.addBalance(_env.author(), 0);  // touch coinbase no matter what
```
pirapira: Why should the coinbase be touched both before and after the transaction?

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -283,8 +287,15 @@ std::tuple<eth::State, ImportTest::ExecOutput, eth::ChangeLog> ImportTest::execu
...
 
         //Finalize the state manually (clear logs)
-        bool removeEmptyAccounts = m_envInfo->number() >= se->chainParams().EIP158ForkBlock;
         initialState.commit(removeEmptyAccounts ? State::CommitBehaviour::RemoveEmptyAccounts : State::CommitBehaviour::KeepEmptyAccounts);
+
+        if (!removeEmptyAccounts)
+        {
+            // Touch here bacuse coinbase might be suicided above
+            initialState.addBalance(_env.author(), 0);  // touch coinbase no matter what
+            initialState.commit(removeEmptyAccounts ? State::CommitBehaviour::RemoveEmptyAccounts :
```
pirapira: What if we move this out of the `if` body?  Maybe we can skip the first `initialState.commit`?

winsvega: no. the first commit apply suicide instruction to coinbase. making it dissappear from the post state.

winsvega: thus simulating touch and commiting the state again.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -292,7 +291,7 @@ std::tuple<eth::State, ImportTest::ExecOutput, eth::ChangeLog> ImportTest::execu
         if (!removeEmptyAccounts)
         {
             // Touch here bacuse coinbase might be suicided above
-            initialState.addBalance(_env.author(), 0);  // touch coinbase no matter what
+            initialState.addBalance(_env.author(), 0);  // imitate mining reward
```
pirapira: Why is this done again?  Before, this is a line
```
initialState.addBalance(_env.author(), 0);  // imitate mining reward
```

winsvega: because suicide coibase will delete coinbase account from the state. we need to touch it again after suicide.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -283,8 +286,15 @@ std::tuple<eth::State, ImportTest::ExecOutput, eth::ChangeLog> ImportTest::execu
...
 
         //Finalize the state manually (clear logs)
-        bool removeEmptyAccounts = m_envInfo->number() >= se->chainParams().EIP158ForkBlock;
         initialState.commit(removeEmptyAccounts ? State::CommitBehaviour::RemoveEmptyAccounts : State::CommitBehaviour::KeepEmptyAccounts);
+
+        if (!removeEmptyAccounts)
+        {
+            // Touch here bacuse coinbase might be suicided above
+            initialState.addBalance(_env.author(), 0);  // imitate mining reward
+            initialState.commit(removeEmptyAccounts ? State::CommitBehaviour::RemoveEmptyAccounts :
```
pirapira: What happens if you move this line down out of the if body?

PR #5025, opened 2018-05-15
Circle CI: Enable ccache

chfast:
Added ccache to Circle CI and fixed deps cache (the path was wrong).
This is rather an experiment to see if the ccache gives anything in long run. In this PR the build step goes from 9 mins to 3 mins, but ccache restore and save takes additional 0.5 + 1 mins. This is 50% better build time in total for 100% cache hit.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -60,9 +60,9 @@ defaults:
   save-deps-cache: &save-deps-cache
     cache-save:
       name: "Save dependencies cache"
-      key: &deps-cache-key deps-2-{{arch}}-{{checksum "compiler.version"}}-{{checksum "cmake/ProjectLibFF.cmake"}}
+      key: &deps-cache-key deps-4-{{arch}}-{{checksum "compiler.version"}}-{{checksum "cmake/ProjectLibFF.cmake"}}
       paths:
-        - build/deps
+        - ~/build/deps
 
   restore-deps-cache: &restore-deps-cache
     cache-restore:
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -89,13 +89,49 @@ defaults:
...
+      key: ccache-1-{{arch}}-{{checksum "compiler.version"}}-{{.Branch}}-{{epoch}}
+      paths:
+        - ~/.ccache
+
+  restore-ccache: &restore-ccache
+    restore_cache:
+      name: "Restore ccache"
+      keys:
+        - ccache-1-{{arch}}-{{checksum "compiler.version"}}-{{.Branch}}
+        - ccache-1-{{arch}}-{{checksum "compiler.version"}}
```
chriseth: The bad thing about this fallback key is that it will be soon horribly outdated. It might make more sense to use the current week or something as another level. On the other hand, perhaps you prefer full builds for the first build in a PR anyway.

chfast: I believe it works in a different way. The Circle CI documentation is not clear about it, but from their and other examples it looks the cache key matching is prefix based and the newest matched cache blog is taken.

First of all, we update the ccache cache on every build (see `-{{epoch}}` suffix in the `save-cache`).

Here, we restore the newest ccache cache matching the current branch (I'm not sure what the "branch" means in foreign PRs). If nothing found, the newest ccache cache is taken considering all branches (including all PRs).

Considering all this, maybe better strategy would be to fallback to develop branch only, not to other branches.

I took it from https://github.com/hyperledger/quilt/blob/master/.circleci/config.yml, the only good example I've found with ccache config for Circle CI.

Issue #4503, opened 2017-09-15
Select Win64 architecture on Windows


chfast:
By default, on Windows CMake generates project for Visual Studio for Win32 architecture.
Tweak this to always use Win64 by setting `CMAKE_GENERATOR_PLATFORM`: https://cmake.org/cmake/help/v3.9/variable/CMAKE_GENERATOR_PLATFORM.html.

Also update README and simplify instructions for building on Windows.

bahelit:
Would like to pick this up, will give me a reason to setup up the build environment on my laptop.

chfast:
@bahelit it's yours. Do you have any practice with CMake?

bahelit:
Just cloned the repo, and successfully built the project.

So we are wanting to detect if we are on Windows and set the correct compiler within CMake without passing the argument -G "compiler version"?

chfast:
I don't know exactly what's possible, but ideally:

1. `cmake ..` should build with "Visual Studio 14 2015 Win64" (build with "Visual Studio 14 2015" now).
2. `cmake .. -G "Visual Studio 14 2015"` should be an error.
3. `cmake .. -G "MinGW Something"` should at least try to use MinGW.

In other words, for people that have "Visual Studio 2015" installed, they should be allowed to skip `-G` option. This https://cmake.org/cmake/help/v3.9/variable/CMAKE_GENERATOR_PLATFORM.html might be helpful, but I have investigated it a lot.

Try to inform us on progress. If it would take long time, might not be worth it.

bahelit:
CMake will automatically detect the presence of the visual studio compiler, the only problem is that it defaults to creating the project as an x86 build. 

Successfully created and built the solution with Visual Studio 2017 community edition with the following commands:
`mkdir build && cd build`
`cmake .. -A x64`

It's not as simple as just having `cmake ..` but is better then having to specify the Visual Studio version, which is unnecessary as cmake automatically does this.

Have already modified the README.md with the new steps and a recommendation of Visual Studio 2017 instead of 2015.

PR #5029, opened 2018-05-16
Select x64 platform for Visual Studio if not specified

chfast:
Resolves https://github.com/ethereum/cpp-ethereum/issues/4503.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,10 @@ include(CableToolchains)
 include(GNUInstallDirs)
 
 cable_configure_toolchain(DEFAULT cxx11)
+if(${CMAKE_GENERATOR} MATCHES "Visual Studio" AND NOT CMAKE_GENERATOR_PLATFORM)
+    # Select x64 platform for Visual Studio if not specified.
+    set(CMAKE_GENERATOR_PLATFORM x64 CACHE INTERNAL "Name of generator platform.")
+endif()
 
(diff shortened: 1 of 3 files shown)


gumb0:
I have both VS 2015 and 2017 installed and just `cmake ..` chooses 2017 (and fails)

gumb0:
Ah cmake command for windows didn't change. sorry then

gumb0:
But it still fails
```
C:\dev\cpp-ethereum\build_temp>cmake -G "Visual Studio 14 2015 Win64" ..
-- [cable ] Cable 0.2.2 initialized
CMake Error at CMakeLists.txt:69 (project):
  Generator

    Visual Studio 14 2015 Win64

  does not support platform specification, but platform

    x64

  was specified.


-- Configuring incomplete, errors occurred!
See also "C:/dev/cpp-ethereum/build_temp/CMakeFiles/CMakeOutput.log".
```

chfast:
What CMake version? I think this was added in 3.9.

gumb0:
`cmake version 3.9.3`

chfast:
Did you use it for existing build or a new one?

If causes problems probably not worth it.

chfast:
It seems it also caused full Hunter rebuild...

gumb0:
I made new build directory

chfast:
It was working for me for CMake 3.10 :)

PR #5030, opened 2018-05-16
explicit chainParams check

winsvega:
Instead of validating fields names with c_fieldname constant
I moved requireJson function to libdevcode
so you could specify how json object should look like, and what fields it must have and of what time those fields could be. 

example code in validating the chainParams. requires review. 
this is a step toward understanding what chain params are required and what could be set optional.


libdevcore/JsonUtils.h
```diff
@@ -24,4 +24,14 @@ namespace dev
 {
 // Throws UnknownField() if _obj contains field names not listed in _allowedFields.
 void validateFieldNames(json_spirit::mObject const& _obj, std::set<std::string> const& _allowedFields);
+
+// Converts json value type to string
+std::string jsonTypeAsString(json_spirit::Value_type _type);
+
+// Check json _o with validation map that reuires certain field of certain type to be present in
+// json
+typedef std::set<json_spirit::Value_type> possibleJsonType;
```
pirapira: For me `jsonTypeSet` sounds better.

First I thought "possibly it is a json type, but possibly it is not a json type".

libdevcore/JsonUtils.cpp
```diff
@@ -31,3 +31,77 @@ void dev::validateFieldNames(json_spirit::mObject const& _obj, std::set<std::str
...
+            BOOST_THROW_EXCEPTION(MissingField() << errinfo_comment(comment));
+        }
+
+        bool matched = false;
+        std::string sTypes;
+        for (auto const& type : vmap.second)
+        {
+            if (sTypes.size())
+                sTypes += ", or ";
+            sTypes += jsonTypeAsString(type);
```
pirapira: You need this concatenation only when there is an error.

winsvega:
https://github.com/ethereum/EIPs/issues/1085

libdevcore/JsonUtils.cpp
winsvega: so you think two loops on `for (auto const& type : vmap.second)` is better ?

winsvega:
@gumb0 @pirapira 
Cpp client has specific declaration of a precompiled contracts in the genesis state. 

My question is do we really need it? Precompiled contracts are hardcoded anyway. non other clients support this specific account desctiption anyway. retesteth and hive has to prepare a specific genesis for cpp in this case.  could we move those params to hardcode of the client? 

```
"0000000000000000000000000000000000000001": { "precompiled": { "name": "ecrecover", "linear": { "base": 3000, "word": 0 } } },
		"0000000000000000000000000000000000000002": { "precompiled": { "name": "sha256", "linear": { "base": 60, "word": 12 } } },
		"0000000000000000000000000000000000000004": { "precompiled": { "name": "identity", "linear": { "base": 15, "word": 3 } } },
		"0000000000000000000000000000000000000005": { "precompiled": { "name": "modexp" } },
		"0000000000000000000000000000000000000006": { "precompiled": { "name": "alt_bn128_G1_add", "linear": { "base": 500, "word": 0 } } },
		"0000000000000000000000000000000000000007": { "precompiled": { "name": "alt_bn128_G1_mul", "linear": { "base": 40000, "word": 0 } } },
		"0000000000000000000000000000000000000008": { "precompiled": { "name": "alt_bn128_pairing_product" } }
```

libethereum/ChainParams.cpp
```diff
@@ -98,7 +133,8 @@ ChainParams ChainParams::loadConfig(
 	json_spirit::read_string_or_throw(_json, val);
 	js::mObject obj = val.get_obj();
 
-	validateFieldNames(obj, c_knownChainConfigFields);
+    validateConfigJson(obj);
+    validateFieldNames(obj, c_knownChainConfigFields);
```
gumb0: Did you mean to leave both of these validations here?
So now we have two separate list of allowed fields to maintain?

winsvega: the other one could be removed from this place. once we come to a standart in genesis config.
the question remains is wether we should use constants instead of strings in scheme validation functions.

libdevcore/JsonUtils.cpp
pirapira: yes

libethereum/ChainParams.cpp
gumb0: I don't think we should validate it two times. So let's not merge it then until we have final version

libethereum/ChainParams.cpp
```diff
@@ -90,6 +90,41 @@ set<string> const c_knownParamNames = {c_minGasLimit, c_maxGasLimit, c_gasLimitB
     c_durationLimit, c_chainID, c_networkID, c_allowFutureBlocks};
 } // anonymous namespace
 
+void validateConfigJson(js::mObject const& _obj)
+{
+    requireJsonFields(_obj, "ChainParams::loadConfig",
+        {{"sealEngine", {json_spirit::str_type}}, {"params", {json_spirit::obj_type}},
```
gumb0: So why did you choose not to use the constants for the field names here?

libdevcore/JsonUtils.cpp
```diff
@@ -31,3 +31,77 @@ void dev::validateFieldNames(json_spirit::mObject const& _obj, std::set<std::str
...
+    case json_spirit::real_type:
+        return "json Real";
+    case json_spirit::null_type:
+        return "json Null";
+    default:
+        return "json n/a";
+    }
+}
+
+void dev::requireJsonFields(json_spirit::mObject const& _o, std::string const& _config,
```
gumb0: I'm not sure about having it in `libdevcore`

libethereum/ChainParams.cpp
winsvega: I guess this check is like a constant itself. if it does not match, then it throws an error. 
so if you have a typo, it will break the test run. 
although it is said that it is not good to use char in the code this way. maybe we need a dictionary class that would return a string by id for all text checks and messages in the client.

libdevcore/JsonUtils.cpp
winsvega: the previous checks were put into this file. @pirapira ?

libethereum/ChainParams.cpp
gumb0: We shouldn't have duplicate string literals in the code here, it's maintenance pain and might waste resources. So use here `c_sealEngine`, `c_params`  etc.

winsvega:
looks like client devs wont come up with the common scheme any time soon. 
I suggest we figure out the minimum required config fields for cpp and create our own scheme validation.

libdevcore/JsonUtils.cpp
pirapira: If there is a better place, feel free to move.

winsvega: so we could remove jsonUtils from libdevcore. and keep it in libethereum. but I dont think its a better choise

pirapira: If `libdevcore` or `libethereum`, I'd vote for `libdevcore`.  This is not so particularly about Ethereum.

chfast:
Have you considered using JSON Schema to specify and validate JSON configs? I've found validator for JSON Schema: https://github.com/tristanpenman/valijson.

winsvega:
Hmm. That repo looks like a lot of code to add for dependencies. 
Here we have one small function that does a verification

libdevcore/JsonUtils.h
```diff
@@ -24,4 +24,19 @@ namespace dev
...
+// Check json _o with validation map that reuires certain field of certain type to be present in
+// json
+enum jsonField
+{
+    Required,
+    Optional
+};
+using jsonTypeSet = std::set<json_spirit::Value_type>;
+using jsonType = std::pair<jsonTypeSet, jsonField>;
+void requireJsonFields(json_spirit::mObject const& _o, std::string const& _config,
```
pirapira: The function does a little bit different job from what "require Json Fields" suggests.

Please document the function.

winsvega: could you clarify what "document the function" should look like ?

libdevcore/JsonUtils.h
```diff
@@ -24,4 +24,25 @@ namespace dev
 {
 // Throws UnknownField() if _obj contains field names not listed in _allowedFields.
 void validateFieldNames(json_spirit::mObject const& _obj, std::set<std::string> const& _allowedFields);
+
+// Converts json value type to string
+std::string jsonTypeAsString(json_spirit::Value_type _type);
+
+enum jsonField
```
gumb0: `enum class` is preferable to `enum`. Also start type names with capital letter. Also this should be better named something like `JsonFieldPresence` maybe

libdevcore/JsonUtils.h
```diff
@@ -24,4 +24,25 @@ namespace dev
...
+{
+    Required,
+    Optional
+};
+using jsonTypeSet = std::set<json_spirit::Value_type>;
```
gumb0: Start type name with capital letter

libdevcore/JsonUtils.h
```diff
@@ -24,4 +24,25 @@ namespace dev
...
+using jsonType = std::pair<jsonTypeSet, jsonField>;
```
gumb0: Start type name with capital letter, also this is something like `JsonTypesAndPresence` I'd say

libdevcore/JsonUtils.h
```diff
@@ -24,4 +24,25 @@ namespace dev
...
+//! Check the json object with validation map that reuires certain field of certain type to be
```
gumb0: Our coding style prescribes @notation rather than \notation in doxygen comments (that is, change the following lines to use `@param`)
Also the rest of the code tends to mark doxygen comment with triple slash `///` instead of `//!`

Though to be honest I don't see much point in using doxygen comments at all for us.

libdevcore/JsonUtils.cpp
```diff
@@ -31,3 +31,86 @@ void dev::validateFieldNames(json_spirit::mObject const& _obj, std::set<std::str
...
+            std::string const comment =
+                "Unexpected field '" + field.first + "' in config: " + _config;
+            std::cerr << comment << "\n"
+                      << json_spirit::write_string((json_spirit::mValue)_o, true) << "\n";
+            BOOST_THROW_EXCEPTION(UnknownField() << errinfo_comment(comment));
+        }
+    }
+
+    // check field types with validation map
+    for (auto const vmap : _validationMap)
```
gumb0: I'd maybe rename `vmap` to something more clearly expressing what it is, because it's difficult to follow what all `vmap.second.first`, `vmap.second.second` etc. are.

Or maybe as first lines of the loop body do
```
auto const& fieldName = vmap.first;
auto const& fieldTypesAndPresence = vmap.second;
```

libdevcore/JsonUtils.cpp
```diff
@@ -31,3 +31,86 @@ void dev::validateFieldNames(json_spirit::mObject const& _obj, std::set<std::str
...
+                          << json_spirit::write_string((json_spirit::mValue)_o, true) << "\n";
+                BOOST_THROW_EXCEPTION(MissingField() << errinfo_comment(comment));
+            }
+            else if (vmap.second.second == jsonField::Optional)
+                continue;
+        }
+
+        // check that field type is one of allowed field types
+        bool matched = false;
+        for (auto const& type : vmap.second.first)
```
gumb0: You shouldn't iterate over the `set` to find if the element is present in it. Use `std::set::find()`

libdevcore/JsonUtils.cpp
```diff
@@ -31,3 +31,86 @@ void dev::validateFieldNames(json_spirit::mObject const& _obj, std::set<std::str
...
+        {
+            if (_o.at(vmap.first).type() == type)
+                matched = true;
+        }
+        if (matched == false)
+        {
+            std::string sTypes;
```
gumb0: I'd prefer this following error message construction to go into separate function. This one function is doing too much stuff.

libdevcore/JsonUtils.cpp
```diff
@@ -31,3 +31,86 @@ void dev::validateFieldNames(json_spirit::mObject const& _obj, std::set<std::str
...
+            for (auto const& type : vmap.second.first)
+            {
+                if (sTypes.size())
+                    sTypes += ", or ";
```
gumb0: There shouldn't be comma before `or` in English here

libdevcore/JsonUtils.cpp
```diff
@@ -31,3 +31,86 @@ void dev::validateFieldNames(json_spirit::mObject const& _obj, std::set<std::str
...
+        bool matched = false;
+        for (auto const& type : vmap.second.first)
+        {
+            if (_o.at(vmap.first).type() == type)
+                matched = true;
+        }
+        if (matched == false)
+        {
+            std::string sTypes;
+            for (auto const& type : vmap.second.first)
```
gumb0: I'd use `boost::algorithm::join` here instead of reinventing joining the strings, but it's up to you

libdevcore/JsonUtils.cpp
winsvega: why?

winsvega: Ah, I see.

libdevcore/JsonUtils.cpp
```diff
@@ -31,3 +31,86 @@ void dev::validateFieldNames(json_spirit::mObject const& _obj, std::set<std::str
...
+            {
+                if (sTypes.size())
+                    sTypes += ", or ";
+                sTypes += jsonTypeAsString(type);
+            }
+            std::string const comment =
+                "Field '" + vmap.first + "' expected to be " + sTypes + ", but set to " +
```
gumb0: is (or was) expected to be, but is (or was) set

libethereum/ValidationSchemes.h
```diff
@@ -0,0 +1,82 @@
...
+
+    cpp-ethereum is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/** @file ValidationSchemes.h
```
gumb0: remove these `@file` and `@date`
Use `@file` only if you provide some meaningful description of the file

libdevcore/JsonUtils.h
gumb0: Actually sometimes there's `/**` too, but I don't remember many `//!` in our code

libethereum/Account.cpp
```diff
@@ -119,12 +100,12 @@ AccountMap dev::eth::jsonToAccountMap(std::string const& _json, u256 const& _def
...
 	js::mObject o = val.get_obj();
-	validateAccountMapObj(o);
-	for (auto const& account: o)
+    for (auto const& account: o)
 	{
-		Address a(fromHex(account.first));
+        Address a(fromHex(account.first));
 		// FIXME: Do not copy every account object.
 		auto o = account.second.get_obj();
+        validateAccountMapObj(o);
```
gumb0: Why do you do this now inside the loop? Validating the whole map object in each iteration?

libethereum/ValidationSchemes.h
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+#include <string>
+#include <json_spirit/JsonSpiritHeaders.h>
+
+namespace dev {
+namespace eth {
+namespace validation {
+
+extern std::string const c_sealEngine;
```
gumb0: I liked these constants more in `ChainParams.cpp` / `Account.cpp` ... where they are closer to the actual useful work being done with them.
Are you sure with moving all of this into separate file? I'd leave the three validation functions closer to the other work with the json configs

winsvega: this constants could grow over time into a big amount of variables

libethereum/Account.cpp
winsvega: checking one account instead of a list of accounts

gumb0: Ah wow, don't use `o` for both the whole map and one account object.
Use more meaningful names than `o` anyway

libethereum/ValidationSchemes.h
```diff
@@ -0,0 +1,82 @@
...
+void validateConfigJson(json_spirit::mObject const& _obj);
+
+// Validate account json object
+void validateAccountObj(json_spirit::mObject const& _o);
+
+// Validate account Map json object. Map indicates which fields are set
+void validateAccountMapObj(json_spirit::mObject const& _o);
+
+
+}}}
```
gumb0: This doesn't look like it was clang-formatted

libethereum/ValidationSchemes.cpp
```diff
@@ -0,0 +1,142 @@
...
+    // A map object with possibly defined fields
+    requireJsonFields(_obj, "validateAccountMapObj",
+        {{c_storage, {{json_spirit::obj_type}, JsonFieldPresence::Optional}},
+            {c_balance, {{json_spirit::str_type}, JsonFieldPresence::Optional}},
+            {c_nonce, {{json_spirit::str_type}, JsonFieldPresence::Optional}},
+            {c_code, {{json_spirit::str_type}, JsonFieldPresence::Optional}},
+            {c_precompiled, {{json_spirit::obj_type}, JsonFieldPresence::Optional}},
+            {c_shouldnotexist, {{json_spirit::str_type}, JsonFieldPresence::Optional}},
+            {c_wei, {{json_spirit::str_type}, JsonFieldPresence::Optional}}});
+    }
```
gumb0: Indentation is off here. Use clang-format

libethereum/ValidationSchemes.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    You should have received a copy of the GNU General Public License
+    along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#include "ValidationSchemes.h"
+#include <libdevcore/JsonUtils.h>
+#include <string>
+
+using namespace std;
+using namespace dev;
```
gumb0: You don't need this, everything is defined inside `dev` anyway.
You probably don't need `using namespace std;` either

libethereum/ValidationSchemes.cpp
```diff
@@ -0,0 +1,142 @@
...
+namespace js = json_spirit;
```
gumb0: You don't use this below, it's `json_spirit` everywhere. So maybe just delete it.

libethereum/ValidationSchemes.cpp
```diff
@@ -0,0 +1,142 @@
...
+            if (_obj.count(c_balance))
+                requireJsonFields(_obj, "validateAccountObj",
+                    {{c_balance, {{json_spirit::str_type}, JsonFieldPresence::Required}}});
+            else
+                requireJsonFields(_obj, "validateAccountObj",
+                    {{c_wei, {{json_spirit::str_type}, JsonFieldPresence::Required}}});
+        }
+        else
+        {
+            // A standart account with all fields
```
gumb0: What's the difference between standard account and genesis account? Do these standard accounts really always have code and storage?

libethereum/ChainParams.cpp
```diff
@@ -27,15 +27,16 @@
 #include <libethcore/SealEngine.h>
 #include <libethcore/BlockHeader.h>
 #include <libethcore/Precompiled.h>
+#include "ValidationSchemes.h"
```
gumb0: Use clang-format here, it should reorder includes

libethereum/ValidationSchemes.cpp
winsvega: I used clang-format

libethereum/ValidationSchemes.cpp
winsvega: yes. standart account required to have all 4 fields. 
in genesis it is not required.
and account map could be empty

libdevcore/JsonUtils.cpp
winsvega: git does not show. but this peace is updated with user friendly names

libethereum/ValidationSchemes.cpp
```diff
@@ -0,0 +1,141 @@
...
+        if (_obj.count(c_balance))
+            requireJsonFields(_obj, "validateAccountObj",
+                {{c_balance, {{js::str_type}, JsonFieldPresence::Required}}});
+        else
+            requireJsonFields(_obj, "validateAccountObj",
+                {{c_wei, {{js::str_type}, JsonFieldPresence::Required}}});
+    }
+    else
+    {
+        // A standart account with all fields
```
gumb0: In which configs are they used?

libethereum/ValidationSchemes.h
```diff
@@ -0,0 +1,81 @@
...
+extern std::string const c_networkID;
+extern std::string const c_allowFutureBlocks;
+
+// Validate config.json that contains chain params and genesis state
+void validateConfigJson(json_spirit::mObject const& _obj);
+
+// Validate account json object
+void validateAccountObj(json_spirit::mObject const& _o);
+
+// Validate account Map json object. Map indicates which fields are set
```
gumb0: Second part of comment is not needed I guess?

libethereum/ValidationSchemes.cpp
gumb0: Ah it's allowed to have in genesis accounts either balance only or all 4 fields? ok then

libethereum/Account.cpp
```diff
@@ -118,35 +99,36 @@ AccountMap dev::eth::jsonToAccountMap(std::string const& _json, u256 const& _def
...
-	for (auto const& account: o)
-	{
-		Address a(fromHex(account.first));
+
+    for (auto const& account : val.get_obj())
+    {
+        Address a(fromHex(account.first));
 		// FIXME: Do not copy every account object.
-		auto o = account.second.get_obj();
+        auto const& accountMapJson = account.second.get_obj();
```
gumb0: Maybe just `accountJson` or `accountObject`, because `acountMap` sounds like a map of accounts, but as I understand it refers to single account

libethereum/ValidationSchemes.h
```diff
@@ -0,0 +1,81 @@
...
+void validateAccountMapObj(json_spirit::mObject const& _o);
```
gumb0: Also this name, if it validates just one account object I would name it differently

libethereum/ChainParams.cpp
```diff
@@ -135,48 +97,33 @@ ChainParams ChainParams::loadConfig(
...
     cp.genesisState = jsonToAccountMap(
         genesisStateStr, cp.accountStartNonce, nullptr, &cp.precompiled, _configPath);
-    cp.stateRoot = _stateRoot ? _stateRoot : cp.calculateStateRoot(true);
 
-    return cp;
-}
+    // Strict account check
+    js::read_string_or_throw(genesisStateStr, val);
+    for (auto const& account: val.get_obj())
+        validateAccountObj(account.second.get_obj());
```
gumb0: Do you need his here at all? You call `validateConfigJson(obj);` above and inside it it already goes over all accounts, no?

libethereum/ValidationSchemes.cpp
winsvega: In tests also

libethereum/Account.cpp
winsvega: Map is because it indicates which fields are set and which are not. there is a class for that: AccountMap

libethereum/ChainParams.cpp
winsvega: you right

libethereum/ValidationSchemes.h
winsvega: the confusion is because it should be account mask here

libethereum/ValidationSchemes.h
winsvega: accountMaskObj

libethereum/ValidationSchemes.h
```diff
@@ -0,0 +1,81 @@
...
+extern std::string const c_durationLimit;
+extern std::string const c_chainID;
+extern std::string const c_networkID;
+extern std::string const c_allowFutureBlocks;
+
+// Validate config.json that contains chain params and genesis state
+void validateConfigJson(json_spirit::mObject const& _obj);
+
+// Validate account json object
+void validateAccountObj(json_spirit::mObject const& _o);
```
gumb0: This is now used only by `validateConfigJson` so make it internal to `ValidationSchemes.cpp`, remove from header

gumb0:
I'm confused now about these account masks. 

`dev::eth::jsonToAccountMap()` function parses json with accounts and returns all accounts as `AccountMap`. It also produces `AccountMaskMap` but that's a side-effect, not it's main purpose.

Now you want to validate input json to this function. Input json has account description, not "account masks". So the validation of json of each account should be called something like "validateAccountJsonObject()" not "validateAccountMaskObj()"

Also: I'm looking at `ChainParams::loadConfig()` which is supposed to parse config including genesis header and accounts. Now it looks like it would validate first the accounts list inside `validateConfigJson()` (calling `validateAccountObj()`) and then it would validate the same account list inside `dev::eth::jsonToAccountMap()` calling `validateAccountMaskObj()` for each.
So do you even need both of `validateAccountObj()` and `validateAccountMaskObj()`?
It looks to me like that they both validate the same json structure but in a different way for some weird reason.

libethereum/ValidationSchemes.h
winsvega: I don't think it shoud be removed. it should be used in testeth.

winsvega:
your confusion might be because 

1) accountMask is totally a different object. it is not a json object of account.
2) accountMask used only in the tests. but for some reason Gavin implemented it in libethereum.

gumb0:
Sure, ping me in discord for a call, you're offline now

PR #5041, opened 2018-05-23
fix testeth -- --exectimelog option

winsvega:
fix --exectimelog array being reset with each testcase

### DIFF
--- a/test/tools/libtesteth/TestOutputHelper.cpp
+++ b/test/tools/libtesteth/TestOutputHelper.cpp
@@ -37,9 +37,8 @@ void TestOutputHelper::initTest(size_t _maxTests)
 	BasicAuthority::init();
 	m_currentTestName = "n/a";
 	m_currentTestFileName = string();
-	m_execTimeResults = std::vector<execTimeName>();
-	m_timer = Timer();
-	m_timer.restart();
+    m_timer = Timer();
+    m_timer.restart();
 	m_currentTestCaseName = boost::unit_test::framework::current_test_case().p_name;
 	if (!Options::get().createRandomTest)
 		std::cout << "Test Case \"" + m_currentTestCaseName + "\": \n";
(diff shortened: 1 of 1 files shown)


test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -86,9 +85,14 @@ void TestOutputHelper::printTestExecStats()
 {
 	if (Options::get().exectimelog)
 	{
-		std::cout << std::left;
+        int totalTime = 0;
+        std::cout << std::left;
```
chfast: The stream flags should be restored in the end of the function.

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -37,9 +37,8 @@ void TestOutputHelper::initTest(size_t _maxTests)
 	BasicAuthority::init();
 	m_currentTestName = "n/a";
 	m_currentTestFileName = string();
-	m_execTimeResults = std::vector<execTimeName>();
-	m_timer = Timer();
-	m_timer.restart();
+    m_timer = Timer();
+    m_timer.restart();
```
gumb0: This is redundant, constructor does `restart()`

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -86,9 +85,14 @@ void TestOutputHelper::printTestExecStats()
 {
 	if (Options::get().exectimelog)
 	{
-		std::cout << std::left;
+        int totalTime = 0;
+        std::cout << std::left;
 		std::sort(m_execTimeResults.begin(), m_execTimeResults.end(), [](execTimeName _a, execTimeName _b) { return (_b.first < _a.first); });
-		for (size_t i = 0; i < m_execTimeResults.size(); i++)
+        for (size_t i = 0; i < m_execTimeResults.size(); i++)
```
gumb0: Use range for or better `std::accumulate`

test/tools/libtesteth/TestOutputHelper.cpp
gumb0: @winsvega maybe output to `strstream` and then the whole string to `cout` to avoid dealing with restoring `cout` state

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -86,9 +85,14 @@ void TestOutputHelper::printTestExecStats()
...
+            totalTime += m_execTimeResults[i].first;
+        std::cout << setw(45) << "Total Time: " << setw(25) << "     : " + toString(totalTime)
+                  << "\n";
+        for (size_t i = 0; i < m_execTimeResults.size(); i++)
```
gumb0: use range `for`

test/tools/libtesteth/TestOutputHelper.cpp
chfast: You can use `boost::io::ios_flags_saver` for it.

gumb0:
@winsvega `#include <numeric>`

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -18,7 +18,9 @@
  * Fixture class for boost output when running testeth
  */
 
+#include <numeric>
```
chfast: Reverse the order of the includes.

PR #5050, opened 2018-05-28
Log message when sync discovers new chain head

gumb0:


libethereum/BlockChainSync.cpp
```diff
@@ -261,10 +261,15 @@ void BlockChainSync::syncPeer(std::shared_ptr<EthereumPeer> _peer, bool _force)
 
     if (_force || _peer->m_totalDifficulty > syncingDifficulty)
     {
+        LOG(m_logger) << "Discovered new highest difficulty";
```
chfast: It's not clear that the log verbosity is.

gumb0: `m_logger` definition is
```
    Logger m_logger{createLogger(VerbosityDebug, "sync")};
```
So it's DEBUG

gumb0: Run with e.g. `-v 3 --log-exclude-channels net discov` to be able to see it

chfast: I have a small problem with this. I'd like this message to be INFO. It's not good idea to bind channel with verbosity.

gumb0: We can make it INFO, I'm just afraid that it could lead to many such messages (every time `syncPeer` is called (that's not only when new peer connects) and highest difficulty changes, which is often in live network). Also `new highest difficulty` sounds quite low-level for default log level which I think should be more average-user-oriented.

Maybe output to INFO only `Starting full sync` below and leave this in DEBUG?

chfast: Yes, INFO "Starting full sync" sounds good.

gumb0:
Yeah, syncing with the main net I'm seeing `Discovered new highest difficulty` ~once in several seconds to once a minute, so I'd prefer to leave it in DEBUG
I moved `Starting full sync` to INFO.

libethereum/BlockChainSync.h
```diff
@@ -161,6 +161,7 @@ class BlockChainSync final: public HasInvariants
     u256 m_syncingTotalDifficulty;				///< Highest peer difficulty
 
     Logger m_logger{createLogger(VerbosityDebug, "sync")};
+    Logger m_loggerInfo{createLogger(VerbosityInfo, "sync")};
```
chfast: Do you need to create a logger for every combination? Can't we apply the verbosity level at the place of `LOG` macro?

gumb0: The logger should be defined somewhere anyway (either global or local/member variable, the latter preferable)
But one alternative could be to have a logger without fixed verbosity, then a macro like `LOGV(m_logger, VerbosityInfo) << ...`
which we can make `LOG_INFO(m_logger) << ...`

PR #5051, opened 2018-05-28
Use ethash library fully

chfast:
This uses ethash library in CPU miner. The hashrate goes from 0.6 Mh/s to 0.9 Mh/s.


libethashseal/EthashCPUMiner.cpp
```diff
@@ -83,25 +86,20 @@ void EthashCPUMiner::minerBody()
...
-    unsigned hashCount = 1;
-    for (; !m_shouldStop; tryNonce++, hashCount++)
+    for (unsigned hashCount = 1; !m_shouldStop; tryNonce++, hashCount++)
     {
-        ethashReturn = ethash_full_compute(dag->full, *(ethash_h256_t*)w.headerHash().data(), tryNonce);
-        h256 value = h256((uint8_t*)&ethashReturn.result, h256::ConstructFromPointer);
-        if (value <= boundary && submitProof(EthashProofOfWork::Solution{(h64)(u64)tryNonce, h256((uint8_t*)&ethashReturn.mix_hash, h256::ConstructFromPointer)}))
+        auto result = ethash::hash(ethashContext, toEthash(w.headerHash()), tryNonce);
+        h256 value = h256(result.final_hash.bytes, h256::ConstructFromPointer);
+        if (value <= boundary && submitProof(EthashProofOfWork::Solution{(h64)(u64)tryNonce,
```
gumb0: Can we maybe instead of double cast of `tryNonce` here declare it as `h64` and just increment each iteration? (`FixedHash` has `operator++`)

gumb0: Ah then we would need to cast in `ethash;:hash` call, ok nevermind

libethashseal/Ethash.cpp
```diff
@@ -278,26 +270,14 @@ bool Ethash::verifySeal(BlockHeader const& _blockHeader) const
...
-        m_farm.setWork(m_sealing);        // TODO: take out one before or one after...
-    }
-    bytes shouldPrecompute = option("precomputeDAG");
-    if (!shouldPrecompute.empty() && shouldPrecompute[0] == 1)
-        ensurePrecomputed((unsigned)_bi.number());
+    Guard l(m_submitLock);
+    m_sealing = _bi;
+    m_farm.setWork(m_sealing);
+    m_farm.start(m_sealer);
+    m_farm.setWork(m_sealing);  // TODO: take out one before or one after...
```
gumb0: This comment can be removed, if you don't know what it's about (I don't)

bkukanov:
Why has been removed the libethash library?

chfast:
Replaced with https://github.com/chfast/ethash.

PR #5053, opened 2018-05-28
debug_accountRangeAt and debug_storageRangeAt for "pending" and "latest"

gumb0:
WIP, still need to handle invalid strings passed as block number (currently `jsToBlockNumber()` will always return 0 for invalid string)

cc @winsvega


winsvega:
getting pending state didn't work for me. always got a state without any transactions applied.

gumb0:
Closing in favor of https://github.com/ethereum/cpp-ethereum/pull/5071

Issue #3121, opened 2016-08-25
eth_sendTransaction should return errors


bobsummerwill:
_From @chriseth on March 3, 2016 12:37_

The transaction should be checked for
- gas price too low
- not enough gas
- not enough funds

and potentially others and errors should be returned.
The same applies for eth_sendRawTransaction

_Copied from original issue: ethereum/webthree-umbrella#298_

halfalicious:
I'm taking a crack at this and I had some questions regarding propagating errors from `Client::submitTransaction` to `AccountHolder::authenticate` - 2 options come to mind for addressing this issue: 
* **Add the new errors (e.g. gas price to low, account balance too low) to the `TransactionRepercussion `struct and modify `Client::submitTransaction` to return a `TransactionNotification`:** Adding errors to `TransactionRepercussion `seems logical because it appears to represent conditions which can block transaction submission (e.g. account locked) which is exactly what I'm checking for and there's already a struct which contains all of the data I'd need to return from `Client::submitTransaction` (`TransactionNotification`). But taking this approach would introduce a dependency from `Client `on `AccountHolder `(`TransactionRepercussion `and `TransactionNotification `are defined in `AccountHolder.h`) which doesn't seem good since this means the core Ethereum functionality would depend on the jsonrpc component (though I could address this by moving `TransactionRepercussion `and `TransactionNotification `to `client.h`). 
	
* **Use `TransactionException `(which already appears to have some of the error codes I need)**: However, this means that I'd need to create a new wrapper struct to return the tx hash, from address, and error code from `Client::submitTransaction`. Additionally, I'm a little conflicted about using `TransactionException `because it appears to represent errors that occur during transaction execution which doesn't exactly map my use case (pre-submission tx validation). 

@gumb0 / @chfast: Any thoughts?

halfalicious:
No problem @gumb0 , it's super interesting learning how an Ethereum node works so I'm happy to help out! Thank you for all of your prompt and detailed responses, they're very helpful. 

Regarding geth's behavior - yup, they reject right away: 
```
> web3.eth.sendTransaction({to:web3.personal.listAccounts[0], from:web3.personal.listAccounts[3], value:web3.toWei(0.25)})
	Error: insufficient funds for gas * price + value
	    at web3.js:3143:20
	    at web3.js:6347:15
	    at web3.js:5081:36
	    at <anonymous>:1:1
```
[...]

PR #5060, opened 2018-05-30
Fix bug where eth rejects transactions with 0 gas price

### LINKED ISSUE #5059: eth rejects transactions with 0 gas price
Eth rejects transactions with 0 gas price: 

```
	> web3.eth.sendTransaction({from: web3.personal.listAccounts[0], to: web3.personal.listAccounts[1], value: 0.1, gasPrice: 0})
	Error: INVALID_PARAMS: Invalid method parameters (invalid name and/or type) recognised
	    at web3.js:3143:20
	    at web3.js:6347:15
	    at web3.js:5081:36
	    at <anonymous>:1:1
```

Geth does not. Note that 0 gas price transactions aren't explicitly prohibited by the protocol and therefore I think we should accept them.

halfalicious:
Fix #5059 

It turns out that this is due to a divide-by-zero in `Client::submitTransaction`:

https://github.com/ethereum/cpp-ethereum/blob/8d1c17ba324455c228d90a1997f281be8266adc9/libethereum/Client.cpp#L854

The fix is trivial - fall back to using the value returned by `gasBidPrice()` if `gasPrice `is 0.

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -844,6 +844,10 @@ pair<h256, Address> Client::submitTransaction(TransactionSkeleton const& _t, Sec
 {
     prepareForTransaction();
 
+    // Default gas value meets the intrinsic gas requirements of both
+    // send value and create contract transactions and is the same default
+    // value used by geth and testrpc.
+    const u256 defaultTransactionGas = 90000;
     TransactionSkeleton ts(_t);
     ts.from = toAddress(_secret);
     if (_t.nonce == Invalid256)
@@ -851,7 +855,7 @@ pair<h256, Address> Client::submitTransaction(TransactionSkeleton const& _t, Sec
     if (ts.gasPrice == Invalid256)
         ts.gasPrice = gasBidPrice();
     if (ts.gas == Invalid256)
-        ts.gas = min<u256>(gasLimitRemaining() / 5, ts.gasPrice > 0 ? balanceAt(ts.from) / ts.gasPrice : balanceAt(ts.from) / gasBidPrice());
+        ts.gas = defaultTransactionGas;
 
     Transaction t(ts, _secret);
     m_tq.import(t.rlp());
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -851,7 +851,7 @@ pair<h256, Address> Client::submitTransaction(TransactionSkeleton const& _t, Sec
     if (ts.gasPrice == Invalid256)
         ts.gasPrice = gasBidPrice();
     if (ts.gas == Invalid256)
-        ts.gas = min<u256>(gasLimitRemaining() / 5, balanceAt(ts.from) / ts.gasPrice);
+        ts.gas = min<u256>(gasLimitRemaining() / 5, ts.gasPrice > 0 ? balanceAt(ts.from) / ts.gasPrice : balanceAt(ts.from) / gasBidPrice());
 
     Transaction t(ts, _secret);
     m_tq.import(t.rlp());


libethereum/Client.cpp
```diff
@@ -851,7 +851,7 @@ pair<h256, Address> Client::submitTransaction(TransactionSkeleton const& _t, Sec
     if (ts.gasPrice == Invalid256)
         ts.gasPrice = gasBidPrice();
     if (ts.gas == Invalid256)
-        ts.gas = min<u256>(gasLimitRemaining() / 5, balanceAt(ts.from) / ts.gasPrice);
+        ts.gas = min<u256>(gasLimitRemaining() / 5, ts.gasPrice > 0 ? balanceAt(ts.from) / ts.gasPrice : balanceAt(ts.from) / gasBidPrice());
```
chfast: Why do you use the big price here?

halfalicious: The bid price is used as a default value when one doesn't explicitly supply a gas price:
https://github.com/ethereum/cpp-ethereum/blob/8d1c17ba324455c228d90a1997f281be8266adc9/libethereum/Client.cpp#L851-L852
As a result, I thought it was a suitable value to fall back to when computing the gas limit. I'm of course open to suggestions for a better value to use. 

An alternate approach could be to replace the gas computation entirely with a default gas constant (I think geth and testrpc use 90,000) - in practice this would probably result in lower gas limits being used most of the time and cover value transfer and many contract creation txs (which have base gas costs of 21K and 32K, respectively). What do you think?

halfalicious: You bring up a good point - I actually chatted with @gumb0  about this last week (see #3121), I think that we should allow transactions with gasPrice = 0 since this appears to be allowed by the protocol (and geth accepts them). To address this we can remover the gas computation entirely and fall back to a default gas limit (I'm in favor of using geth/testrpc's 90,000). What do you think?

chfast: Ok, I'm for using 90000 constant. Just expose the constant somewhere by name "defaultTransactionGas". This calculation might be more confusing than useful.

halfalicious: Agreed, the computation is confusing (especially to someone who isn't intimately familiar with the codebase) and tends to result in very high gas limits (unless you have an extremely small account balance).

halfalicious:
```
> eth.sendTransaction({from: personal.listAccounts[0], to: personal.listAccounts[1], value: web3.toWei(0.01), gasPrice: 0})
"0x7d84ee3e4d932fb84605f6970673dd0f61e1893dbb094b3e82d57975e6f2b7e1"
> eth.pendingTransactions
[{
    blockHash: null,
    blockNumber: 0,
    data: "0x0000000000000000000000000000000000000000000000000000000000000000",
    from: "0xd97047048eddd54e083f5afebe0d4ba2331aec13",
    gas: 90000,
    gasPrice: 0,
    hash: "0x7d84ee3e4d932fb84605f6970673dd0f61e1893dbb094b3e82d57975e6f2b7e1",
    nonce: 0,
    r: "0x812fb31403032b2cd1dcf637a3f763142c6ecb5316b63b929196712407c0252c",
    s: "0x53923bd627a359bc198043de70e7477d20d1ed7a0b8a062a14a5810b8e048aeb",
    sighash: "0x1eeb7483e8260ecc43df0e42e7ed594608f49c867381793244238a747fd2281e",
    to: "0xec75ed1426d02b0237385b596cb3a55e6bbe6c3e",
    transactionIndex: 0,
    v: "0x\x00",
    value: 10000000000000000
}]
```

halfalicious:
@chfast: Regarding geth using 90K as its default tx gas limit - here's where geth sets it's default gas limit:

https://github.com/ethereum/go-ethereum/blob/90b22773e99244b37f68f3a8401c405254e0af4e/internal/ethapi/api.go#L1118-L1123

And here's a test to confirm that the value is actually used:

```
> eth.sendTransaction({from: personal.listAccounts[0], to: personal.listAccounts[1], value:web3.toWei(0.01)})
"0xed1e74c506286bb837a886b135a7a66674e1309fc3428448290836f02f181f71"
> eth.pendingTransactions
[{
    blockHash: null,
    blockNumber: null,
    from: "0x1cbb11b72a335b709a61d22bca90f7e54bfcf8b1",
    gas: 90000,
    gasPrice: 11080332409,
    hash: "0xed1e74c506286bb837a886b135a7a66674e1309fc3428448290836f02f181f71",
    input: "0x",
    nonce: 9,
    r: "0xa978176fb39d9fbaa406e1dd2789a9be27a2f81c2b3ced30ac71b91d0a85d380",
    s: "0x425911e0dfd9e301c6a7e4c92129a550951ece3c32e8f2e2797521088bf19dff",
    to: "0x1cac8fadc0711615bfca975be292cbd093a0db58",
    transactionIndex: 0,
    v: "0x29",
    value: 10000000000000000
}]
> web3.version
{
  api: "0.20.1",
  ethereum: "0x3f",
  network: "3",
  node: "Geth/v1.7.2-stable-1db4ecdc/windows-amd64/go1.9",
  whisper: undefined,
  getEthereum: function(callback),
  getNetwork: function(callback),
  getNode: function(callback),
  getWhisper: function(callback)
}
```

halfalicious:
Awesome, thanks @chfast !

PR #5062, opened 2018-05-31
Print one opcode trace per line

lrettig:
Rather than printing a massive JSON blob that needs to be manually unrolled

### DIFF
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -28,6 +28,8 @@
 #include <json/json.h>
 #include <boost/timer.hpp>
 
+#include <numeric>
+
 using namespace std;
 using namespace dev;
 using namespace dev::eth;
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -165,16 +165,15 @@ void StandardTrace::operator()(uint64_t _steps, uint64_t PC, Instruction inst, b
 
 string StandardTrace::json(bool _styled) const
 {
-	if (m_trace.empty())
-		return string();
+    if (m_trace.empty())
+        return {};
 
-	if (_styled)
-		return Json::StyledWriter().write(m_trace);
+    if (_styled)
+        return Json::StyledWriter().write(m_trace);
 
-	return std::accumulate(std::next(m_trace.begin()), m_trace.end(), Json::FastWriter().write(m_trace[0]),
-			[](std::string a, Json::Value b) {
-			return a + Json::FastWriter().write(b);
-			});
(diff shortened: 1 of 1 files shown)


libethereum/Executive.cpp
```diff
@@ -163,7 +163,13 @@ void StandardTrace::operator()(uint64_t _steps, uint64_t PC, Instruction inst, b
 
 string StandardTrace::json(bool _styled) const
 {
-    return _styled ? Json::StyledWriter().write(m_trace) : Json::FastWriter().write(m_trace);
+	if (_styled)
+		return Json::StyledWriter().write(m_trace);
+
+	return std::accumulate(std::next(m_trace.begin()), m_trace.end(), Json::FastWriter().write(m_trace[0]),
```
chfast: You have to include `accumulate` function, probably by `#include <algorithm>`.

lrettig: Done, also rebased

lrettig:
@chfast This would be very useful, is this good to go or do you need anything else here to merge? Thanks!

lrettig:
Wait, I think there's a bug here, I'm getting a segfault with the latest cpp-ethereum code

libethereum/Executive.cpp
```diff
@@ -165,16 +165,15 @@ void StandardTrace::operator()(uint64_t _steps, uint64_t PC, Instruction inst, b
...
-	if (_styled)
-		return Json::StyledWriter().write(m_trace);
+    if (_styled)
+        return Json::StyledWriter().write(m_trace);
 
-	return std::accumulate(std::next(m_trace.begin()), m_trace.end(), Json::FastWriter().write(m_trace[0]),
-			[](std::string a, Json::Value b) {
-			return a + Json::FastWriter().write(b);
-			});
+    return std::accumulate(std::next(m_trace.begin()), m_trace.end(),
```
chfast: FYI, I reformatted changes with clang-format.

libethereum/Executive.cpp
```diff
@@ -165,16 +165,15 @@ void StandardTrace::operator()(uint64_t _steps, uint64_t PC, Instruction inst, b
 
 string StandardTrace::json(bool _styled) const
 {
-	if (m_trace.empty())
-		return string();
+    if (m_trace.empty())
+        return {};
```
chfast: Here wasn't anything wrong, but `string{}` syntax is preferred to `string()` because of some C++ parser edge cases.
I also omitted the type `string` because the compiler will auto-deduct it.

lrettig: Thanks, will keep this and the clang-format in mind for the future.

PR #5071, opened 2018-06-12
Support getting final state of the block in debug_accountRangeAt and debug_storageRangeAt

gumb0:
Call it like `debug_accountRangeAt(blockNumber, txCount, beginKey, maxResults)` to get the final state of the block.

@winsvega You can try how this branch works for you


gumb0:
I wanted to try to add tests for this and for that I'm working on reviving unit tests of RPC https://github.com/ethereum/cpp-ethereum/pull/5072
But that takes a while now and maybe we can merge this first.

winsvega:
ok. fixed. the issue was on retesteth side.

PR #5072, opened 2018-06-13
Revive RPC unit tests

gumb0:
This brings to life the tests for RPC disabled long ago.

It uses mock implementations for RPC server & client given as transport layer to jsonrpc lib.
Not many changes in the tests themselves.

I'd like to see the future work on RPC methods to grow this test suite further on.


winsvega:
I could add RPC tests into retesteth. 
so you could run retesteth -t RPCTests  and it will run a series or requests to a local client chacking the response of the methods.

gumb0:
@winsvega  Not sure it's very easy/would not require more complications around RPC API.

I'd still like to have these unit tests, which are internal to the project and are more tightly coupled with the code.

Integration-level tests in retesteth independent from the client is a good idea, too, but a separate task for you perhaps :)

winsvega:
the idea is to have retesteth as tool that would test RPC api against all aclients. thus we will be sure that clients has no issues with requests-response queries by RPC 

to test RPC api. no new methods required. because we test already existing ones. 
an RPC standart for each method would be required.

PR #5074, opened 2018-06-14
Log synchronously in Debug

gumb0:
I think it's more convenient in the debugger when all log messages are output synchronously and you see exactly what already happened in the log when you trace.

So I keep async logging only in Release.

### DIFF
--- a/libdevcore/Log.cpp
+++ b/libdevcore/Log.cpp
@@ -29,12 +29,21 @@
 #include <boost/log/attributes/clock.hpp>
 #include <boost/log/attributes/function.hpp>
 #include <boost/log/expressions.hpp>
-#include <boost/log/sinks/async_frontend.hpp>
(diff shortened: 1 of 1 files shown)


libdevcore/Log.cpp
```diff
@@ -29,13 +29,22 @@
...
 #include <boost/log/attributes/function.hpp>
 #include <boost/log/expressions.hpp>
-#include <boost/log/sinks/async_frontend.hpp>
 #include <boost/log/sinks/text_ostream_backend.hpp>
 #include <boost/log/sources/global_logger_storage.hpp>
 #include <boost/log/sources/severity_channel_logger.hpp>
 #include <boost/log/support/date_time.hpp>
 #include <boost/log/utility/exception_handler.hpp>
 
+#if defined(NDEBUG)
```
gumb0: Note that RelWithDebInfo doesn't have it so it will also log synchronously

Issue #5014, opened 2018-05-11
web3.eth.signTransaction appears to also submit the transaction


halfalicious:
When poking around in the codebase I noticed that `web3.eth.signTransaction` appears to both sign AND submit the signed transaction, whereas according to the [API behavior documentation](https://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html#signtransaction) it's supposed to only sign the transaction. This signed transaction can then be submitted via `web3.eth.sendSignedTransaction` ([API doc link](https://web3js.readthedocs.io/en/1.0/web3-eth.html#sendsignedtransaction)). Note that eth doesn't currently implement `web3.eth.sendSignedTransaction`, I've filed issue #5013  to track this. 

Additional details:

`Eth::eth_signTransaction()` calls `AccountHolder::Authenticate()`, and both `AccountHolder `subclasses (`FixedAccountHolder `and `SimpleAccountHolder`) implement this abstract function with implementations which submit the signed transaction e.g.:

https://github.com/ethereum/cpp-ethereum/blob/3db4290a7dcca4bb3241ae1fd9189ec518edbd1b/libweb3jsonrpc/AccountHolder.cpp#L132-L136

gumb0:
Does go-ethereum not send it?

halfalicious:
Correct, as far as I can tell geth only signs but doesn't send. Here's output of a test I ran yesterday, note that the signed tx (hash: 0xa91fc20b2...) doesn't show up in the pending tx set:

```
> web3.eth.signTransaction({to: '0x884ae368b946a7e5b11462b1333248388b463cc3', value:1000})
{
  raw: "0xf8670285029470747983015f9094884ae368b946a7e5b11462b1333248388b463cc38203e8802aa09b7eb00e2675c4c8e905d70b98466b7e93a394905b28438c4d6c294b03313f59a0442739b945a09562e1656d69c3b129775dbe7cde772d819fac43dd756a16eb05",
  tx: {
    gas: "0x15f90",
    gasPrice: "0x294707479",
    hash: "0xa91fc20b253af1e23f42d3820179c5cabf9385e3a0f0de69f1777bc297ddab8b",
    input: "0x",
    nonce: "0x2",
    r: "0x9b7eb00e2675c4c8e905d70b98466b7e93a394905b28438c4d6c294b03313f59",
    s: "0x442739b945a09562e1656d69c3b129775dbe7cde772d819fac43dd756a16eb05",
[...]

PR #5076, opened 2018-06-15
Fix 3121: eth_sendTransaction should return errors

### LINKED ISSUE #3121: eth_sendTransaction should return errors

halfalicious:
Fix #3121 

Perform additional validation on transactions sent via the client before importing them into the transaction queue and surface validation and transaction import errors to callers of the RPC function `Eth::eth_submitTransaction`. 

Note that we should also do this for `Eth::eth_sendRawTransaction` but I think those changes will be slightly different since that function calls `Client::injectTransaction` and I don't think one will be able to test their changes until #5014 is fixed. I'll file a new issue to track this. 

With respect to my changes, I've tested the following cases: 

**Valid transaction:**
```
> eth.sendTransaction({from: personal.listAccounts[0], to: personal.listAccounts[1], value: web3.toWei(0.01), gas:21000})
"0x36aafcf8feec89e4293e700f6a169acaad42251204610b28b6cc5cf2448cefa4"
```

**Invalid account balance (value + gas * gas price > balance):**
```
> eth.getBalance(personal.listAccounts[0])
2000000000000000000
> eth.sendTransaction({from: personal.listAccounts[0], to: personal.listAccounts[1], value: web3.toWei(5)})
Error: Account balance is too low (balance < value + gas * gas price).
    at web3.js:3143:20
    at web3.js:6347:15
    at web3.js:5081:36
    at <anonymous>:1:1
```

**Gas limit too low:**
```
> eth.sendTransaction({from: personal.listAccounts[0], to: personal.listAccounts[1], value: web3.toWei(0.01), gas:20999})
Error: Transaction gas amount is less than the intrinsic gas amount for this transaction type.
    at web3.js:3143:20
    at web3.js:6347:15
    at web3.js:5081:36
    at <anonymous>:1:1
```

**Invalid nonce (nonce of already mined transaction)**
Mined transaction: [0x36aafcf8feec89e4293e700f6a169acaad42251204610b28b6cc5cf2448cefa4](https://ropsten.etherscan.io/tx/0x36aafcf8feec89e4293e700f6a169acaad42251204610b28b6cc5cf2448cefa4)

```
> eth.sendTransaction({from: personal.listAccounts[0], to: personal.listAccounts[1], value: web3.toWei(0.01), gas:21000, nonce:0})
Error: Invalid transaction nonce.
    at web3.js:3143:20
    at web3.js:6347:15
    at web3.js:5081:36
    at <anonymous>:1:1
```

**Same nonce but <= gas price:**
```
> eth.sendTransaction({from: personal.listAccounts[0], to: personal.listAccounts[1], value: web3.toWei(0.01), gasPrice:20000000000, nonce:4})
"0xe875c0f76a15fbc53c85367c5e590da5720958f239ac6c015bfa608a32e17ac0"
> eth.sendTransaction({from: personal.listAccounts[0], to: personal.listAccounts[1], value: web3.toWei(0.01), gasPrice:2000000000, nonce:4})
Error: Pending transaction with same nonce but higher gas price exists.
    at web3.js:3143:20
    at web3.js:6347:15
    at web3.js:5081:36
    at <anonymous>:1:1
```

**Duplicate pending transaction:**
```
> eth.sendTransaction({from: personal.listAccounts[0], to: personal.listAccounts[1], value: web3.toWei(0.01), gas:21000, nonce:2})
"0xcf58494520b5b2c0bc732c9be45f8872ea5cc183feee109479986033c88c6055"
> eth.sendTransaction({from: personal.listAccounts[0], to: personal.listAccounts[1], value: web3.toWei(0.01), gas:21000, nonce:2})
Error: Same transaction already exists in the pending transaction queue.
    at web3.js:3143:20
    at web3.js:6347:15
    at web3.js:5081:36
    at <anonymous>:1:1
```


halfalicious:
@chfast , @gumb0 : Would love your eyes on this, thanks much!

halfalicious:
Looks like some client tests are failing, will investigate and address the failures.

libweb3jsonrpc/Eth.cpp
```diff
@@ -271,6 +271,42 @@ string Eth::eth_sendTransaction(Json::Value const& _json)
 	{
 		throw;
 	}
+	catch (ZeroSignatureTransaction&)
+	{
+		BOOST_THROW_EXCEPTION(JsonRpcException("Zero signature transaction."));
+	}
+	catch (InvalidTransactionFormat&)
```
gumb0: Where can this be thrown from?

libweb3jsonrpc/Eth.cpp
```diff
@@ -271,6 +271,42 @@ string Eth::eth_sendTransaction(Json::Value const& _json)
 	{
```
gumb0: Please remove this `catch`, it's useless

libweb3jsonrpc/Eth.cpp
```diff
@@ -271,6 +271,42 @@ string Eth::eth_sendTransaction(Json::Value const& _json)
...
 		throw;
 	}
+	catch (ZeroSignatureTransaction&)
+	{
+		BOOST_THROW_EXCEPTION(JsonRpcException("Zero signature transaction."));
```
gumb0: Let's not use `BOOST_THROW_EXCEPTION` for `JsonRpcException`. I know it's like this in many other places, but it doesn't make sense to me, as `JsonRpcException` is not a boost exception, and `BOOST_THROW_EXCEPTION` will wrap it in boost exception, but this does not improve anything, as it is just caught by libjsonrpc and returned through RPC to the caller. So use just `throw`.

gumb0:
I don't think there's a helper to generate non-empty account. I'm working on something similar in  https://github.com/ethereum/cpp-ethereum/pull/5072 
There are some old RPC tests and there's even `jsonrpc_transact` test which does the very same thing - calls `eth_sendTransaction`
To get the non-zero balance there it mines a block and gets a block reward to the generated address.

You could try to do the similar thing in this `Personal` test or maybe if we merge  https://github.com/ethereum/cpp-ethereum/pull/5072 first, we could move `Personal` test to `JsonRpcSuite` closer to other RPC tests.

To run it locally you need to build `testeth` (`make testeth` if you didn't build it with the full `make` before) and then just
```
test/testeth -t ClientTests/Personal
```

libweb3jsonrpc/Eth.cpp
halfalicious: I throw this in Client::submitTransaction in response to receiving ImportResult::ZeroSignature from TransactionQueue::import:
* Client::submitTransaction (from my PR branch): https://github.com/halfalicious/cpp-ethereum/blob/afe1842cf815b54c53ee3f3cfedfa523ebc60c51/libethereum/Client.cpp#L869-L875

* TransactionQueue::Import: 
https://github.com/ethereum/cpp-ethereum/blob/bcc98a31190e99fe536dc6ea97307fb1e4a188cb/libethereum/TransactionQueue.cpp#L79-L82

libweb3jsonrpc/Eth.cpp
halfalicious: Right you are, I'll remove it.

libweb3jsonrpc/Eth.cpp
gumb0: But you throw `ZeroSignatureTransaction` there

halfalicious: Ah sorry about that, I glanced at the code fragment you posted and saw the 

`BOOST_THROW_EXCEPTION(JsonRpcException("Zero signature transaction.")); `

so I thought you were referring to the `ZeroSignatureTransaction` exception rather than the `InvalidTransactionFormat` exception, my mistake!

Regarding `InvalidTransactionFormat`, it looks like this is only thrown in the `TransactionBase` constructor when a tx is being created from an RLP source: 

https://github.com/ethereum/cpp-ethereum/blob/bcc98a31190e99fe536dc6ea97307fb1e4a188cb/libethcore/TransactionBase.cpp#L47-L96

I don't think this code path can be hit by `Eth::eth_sendTransaction` (the version of `TransactionBase::TransactionBase` which accepts a `TransactionSkeleton` will be hit instead) so therefore I don't need to catch `InvalidTransactionFormat` at the RPC level. I'll remove the catch, thanks for pointing this out!

halfalicious:
@gumb0 Thanks for the tip about the `jsonrpc_transact` test - I updated `ClientTests/Personal` to mine which addressed the balance check failure hit by the first `sendingShouldSucceed` call, but I'm now hitting an invalid nonce failure in the next `sendingShouldSucceed` call. I tried to mine again but the mine function is hanging (the client's block # isn't increasing), any ideas as to how to address this?

gumb0:
@halfalicious Well it makes sense, we now don't allow incorrect nonce and the test tries to send transaction with nonce 0 twice.
So this is expected, you can change the second `sendingShouldSucceed` to `sendingShouldFail`, then you can call `mine()`, then try `tx["nonce"] = "1"` and `sendingShouldSucceed()`

To make the second `mine` work, try this:
https://github.com/ethereum/cpp-ethereum/blob/226d834a4afbc7fd94fe5096506311e38d416188/test/unittests/libweb3jsonrpc/jsonrpc.cpp#L77-L79

I'm not exactly sure why, but setting `allowFutureBlocks` to `true` helped me with the same problem in RPC tests.

halfalicious:
@gumb0 : Sorry, to clarify - the invalid nonce error made sense, what I was confused about was the second `mine `call hanging (which I was making to fix the nonce error, more details below). Setting the chain parameters fixed things, thanks much! 

Additional details on the nonce error since the scenario is slightly different than you described - I'm actually sending out 2 txs with different nonces but the Executive's expected nonce wasn't being incremented because I didn't make a mine call after the first `sendShouldSucceed `call, and therefore the `Executive::initialize` nonce check performed in `Client::submitTransaction` failed. Note that the Client personal test doesn't explicitly set nonces in the txs so the Client automatically generates the nonce based on the state of the pending block / transaction queue - from `Client::submitTransaction`: 

https://github.com/ethereum/cpp-ethereum/blob/5146908fec45aff9ccce4db591d5050a5abafd80/libethereum/Client.cpp#L853-L854

libweb3jsonrpc/Eth.cpp
halfalicious: Actually, the catch is required to surface the account / user errors to the user and without it 5 tests in `ClientTests/Personal` fail. I think that it's best to remove the catch though since a cleaner way to surface the errors would be by having `AccountHolder::authenticate` throw on failure and catch / re-throw at the RPC level like I'm doing for the other errors (as we discussed in #3121). As a result, I've updated the client tests with the new generic error message so they should all pass now.

halfalicious:
@gumb0 : I don't think the jsonrpc tests are reliable, the test which failed ci/circleci: linux-gcc6 `(416 - JsonRpcSuite/jsonrpc_number)` passed locally but I'm seeing another test fail intermittently:

```
C:\Users\nilse\Documents\Code\cpp-ethereum-01\build\test\RelWithDebInfo>testeth -t JsonRpcSuite
Running 12 test cases...
Test Case "jsonrpc_gasPrice":
Test Case "jsonrpc_isListening":
Test Case "jsonrpc_accounts":
Test Case "jsonrpc_number":
Test Case "jsonrpc_peerCount":
Test Case "jsonrpc_setListening":
Test Case "jsonrpc_setMining":
Test Case "jsonrpc_stateAt":
Test Case "jsonrpc_transact":
c:/users/nilse/documents/code/cpp-ethereum-01/test/unittests/libweb3jsonrpc/jsonrpc.cpp(241): error: in "JsonRpcSuite/jsonrpc_transact": check toJS(balance) == balanceString has failed [0xc77e4256863d80000 != 0xcbd47b6eaa8cc0000]
Test Case "simple_contract":
Test Case "contract_storage":
Test Case "sha3":

C:\Users\nilse\Documents\Code\cpp-ethereum-01\build\test\RelWithDebInfo>testeth -t JsonRpcSuite
Running 12 test cases...
Test Case "jsonrpc_gasPrice":
Test Case "jsonrpc_isListening":
Test Case "jsonrpc_accounts":
Test Case "jsonrpc_number":
Test Case "jsonrpc_peerCount":
Test Case "jsonrpc_setListening":
Test Case "jsonrpc_setMining":
Test Case "jsonrpc_stateAt":
Test Case "jsonrpc_transact":
Test Case "simple_contract":
Test Case "contract_storage":
Test Case "sha3":

*** No errors detected

C:\Users\nilse\Documents\Code\cpp-ethereum-01\build\test\RelWithDebInfo>
```

gumb0:
Yeah intermittent failures hopefully should be fixed in https://github.com/ethereum/cpp-ethereum/pull/5088

libethereum/Client.cpp
```diff
@@ -858,7 +858,30 @@ pair<h256, Address> Client::submitTransaction(TransactionSkeleton const& _t, Sec
...
+        case ImportResult::ZeroSignature:
+            BOOST_THROW_EXCEPTION(ZeroSignatureTransaction());
+        case ImportResult::OverbidGasPrice:
+            BOOST_THROW_EXCEPTION(GasPriceTooLow());
+        case ImportResult::AlreadyKnown:
+            BOOST_THROW_EXCEPTION(PendingTransactionAlreadyExists());
+        case ImportResult::AlreadyInChain:
+            BOOST_THROW_EXCEPTION(TransactionAlreadyInChain());
+        default:
+            BOOST_THROW_EXCEPTION(UnknownError());
```
gumb0: I'd prefer this to be a little more specific, like `UnknownTransactionValidationError`

libweb3jsonrpc/Eth.cpp
```diff
@@ -258,24 +258,52 @@ string Eth::eth_sendTransaction(Json::Value const& _json)
...
+			throw JsonRpcException("Account unknown.");
 		case TransactionRepercussion::Locked:
-			BOOST_THROW_EXCEPTION(JsonRpcException("Account is locked."));
+			throw JsonRpcException("Account is locked.");
 		case TransactionRepercussion::Refused:
-			BOOST_THROW_EXCEPTION(JsonRpcException("Transaction rejected by user."));
+			throw JsonRpcException("Transaction rejected by user.");
 		case TransactionRepercussion::Unknown:
-			BOOST_THROW_EXCEPTION(JsonRpcException("Unknown reason."));
+			throw JsonRpcException("Unknown reason.");
```
gumb0: Here we could handle also `default` case and then we could get rid of weirdly-looking throw before `return std::string()` i the end of the funcition

gumb0: This error message if it's surfaced to the caller (I see now it's not) should better be a little more specific like "Unknown authentication error."

libweb3jsonrpc/Eth.cpp
gumb0: I see, this was weird.
Maybe then instead of throwing `JsonRpcException` in the `try` block above we can now throw some other exception (maybe some new `TransactionAuthenticationError`), catch it and rethrow as `JsonRpcException`.

If you plan to refactor `AccountHolder::authenticate` we can defer this.

gumb0: Ah I see the issue you created is exactly about this. Ok then, let's leave like this for now

libweb3jsonrpc/Eth.cpp
halfalicious: Thanks, I wanted to keep my changes as scoped as possible but since I'm already touching so much of `Eth::eth_sendTransaction` I'll go ahead and handle the default case and update the error message. Even though the error isn't surfaced to the caller (yet), I think that changing it now will leave the function in a good state for whomever fixes #5090

halfalicious:
Thanks for the detailed review and great feedback @gumb0 ! I learned a lot from making these changes, your time is much appreciated :)

gumb0:
Thank you for your work!

PR #5065, opened 2018-06-05
EVMC tracing for interpreter

chfast:
This is a preview of the EVMC feature https://github.com/ethereum/evmc/pull/32.

I picked up single VM test as an example: VMTests/vmIOandFlowOperations/mstore0

"Classic" VM trace:
```
test/testeth -t VMTests/vmIOandFlowOperations -- --vmtrace --singletest mstore0 --vm legacy        
Running tests using path: "/home/chfast/Projects/ethereum/cpp-ethereum/test/jsontests"
Running 1 test case...
Test Case "vmIOandFlowOperations": 
100%
TEST mstore0:
2018-06-05 18:26:00 testeth EVM 
    STACK
    MEMORY
    STORAGE

2018-06-05 18:26:00 testeth EVM  | 0 | @0f572e52… | #1 | 0000 : PUSHC | 100000 | -3 | 0x32 ]
2018-06-05 18:26:00 testeth EVM 
    STACK
ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
    MEMORY
    STORAGE

2018-06-05 18:26:00 testeth EVM  | 0 | @0f572e52… | #2 | 0021 : PUSH1 | 99997 | -3 | 0x32 ]
2018-06-05 18:26:00 testeth EVM 
    STACK
ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
0000000000000000000000000000000000000000000000000000000000000001
    MEMORY
    STORAGE

2018-06-05 18:26:00 testeth EVM  | 0 | @0f572e52… | #3 | 0023 : MSTORE | 99994 | -3 | 0x32 ]
2018-06-05 18:26:00 testeth EVM 
    STACK
    MEMORY
0000 ???????? 00 ff ff ff ff ff ff ff 
0008 ???????? ff ff ff ff ff ff ff ff 
0010 ???????? ff ff ff ff ff ff ff ff 
0018 ???????? ff ff ff ff ff ff ff ff 
0020 ???????? ff 00 00 00 00 00 00 00 
0028 ???????? 00 00 00 00 00 00 00 00 
0030 ???????? 00 00 00 00 00 00 00 00 
0038 ???????? 00 00 00 00 00 00 00 00 
    STORAGE

2018-06-05 18:26:00 testeth EVM  | 0 | @0f572e52… | #4 | 0024 : PUSH1 | 99985 | -3 | 0x32 ]
2018-06-05 18:26:00 testeth EVM 
    STACK
[...]

PR #5080, opened 2018-06-19
aleth-interpreter library

chfast:


libaleth-interpreter/VM.h
```diff
@@ -17,9 +17,10 @@
 
 #pragma once
 
-#include "Instruction.h"
 #include "VMConfig.h"
-#include "VMFace.h"
+
+#include <libevm/VMFace.h>
```
gumb0: So libaleth-interpreter depends on libevm? This is going to change later, right?

libaleth-interpreter/VMConfig.h
```diff
@@ -0,0 +1,433 @@
+/*
```
gumb0: Hm, why does github doesn't show this file as moved

gumb0: Did you change it?

gumb0: Ah is it copied?

libaleth-interpreter/VM.h
chfast: Only on headers. This should change, but is not a priority. Home helper functions are defined in VMFace.h. I didn't want to do a lot of changes when moving files.

At least this PR allow building aleth-interpreter as a shared library and distribute it in this form.

libaleth-interpreter/VMConfig.h
chfast: It's a copy. The config was shared between legacy and interpreter VMs.

chfast:
I believe there are 2 easy changes that can be done here later. Use instruction metrics from evmc library and do not include headers from libevm. Maybe third would be to remove some compile time optimization options (only leave the default values).

But we have to decide what to do with https://github.com/ethereum/cpp-ethereum/pull/5065 as these 2 are in conflict.

libaleth-interpreter/VMConfig.h
axic: Perhaps you want to move the file to `libaleth-interpreter` and place a copy into the original directory?

That way if `libaleth-interpreter` is split out of the main tree using "git magic" the history will stay proper.

axic:
The above is my only commit and I like the change FWIW. Would vote to merge this first, the other PR should detect if files are moved during a rebase and apply the changes properly.

libaleth-interpreter/VMConfig.h
chfast: I don't think VMConfig is super important, but I can try to do it.

libevm/LegacyVMConfig.h
```diff
@@ -0,0 +1,444 @@
+/*
```
chfast: This is a copy now.

PR #5082, opened 2018-06-19
replace test_addBlock with test_importRawBlock

winsvega:
a correct way of importing a block from rlp

### DIFF
--- a/libweb3jsonrpc/Test.cpp
+++ b/libweb3jsonrpc/Test.cpp
@@ -73,7 +73,7 @@ string Test::test_getLogHash(string const& _txHash)
     catch (std::exception const& ex)
     {
         cwarn << ex.what();
-        throw JsonRpcException(Errors::ERROR_RPC_INTERNAL_ERROR);
+        throw JsonRpcException(Errors::ERROR_RPC_INTERNAL_ERROR, ex.what());
     }
 }
 
@@ -122,19 +122,6 @@ bool Test::test_modifyTimestamp(int _timestamp)
     return true;
 }
 
-bool Test::test_addBlock(std::string const& _rlp)
-{
-    try
-    {
-        asClientTest(m_eth).addBlock(_rlp);
-    }
-    catch (std::exception const&)
-    {
-        BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INTERNAL_ERROR));
-    }
-    return true;
-}
-
 bool Test::test_rewindToBlock(int _number)
 {
     try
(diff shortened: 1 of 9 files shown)


libweb3jsonrpc/Test.cpp
```diff
@@ -135,16 +136,26 @@ bool Test::test_addBlock(std::string const& _rlp)
...
+            tie(ignore, moreToImport, ignore) = client.syncQueue(100000);
+            this_thread::sleep_for(chrono::milliseconds(100));
+        }
+        return toJS(blockHash);
     }
-    catch (std::exception const&)
+    catch (std::exception const& ex)
     {
-        BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INTERNAL_ERROR));
+        cwarn << ex.what();
```
chfast: Include this in the JSON-RCP error.

winsvega: added

libweb3jsonrpc/Test.cpp
```diff
@@ -135,16 +136,26 @@ bool Test::test_addBlock(std::string const& _rlp)
...
+std::string Test::test_importRawBlock(string const& _blockRLP)
 {
     try
     {
-        m_eth.rewind(_number);
-        asClientTest(m_eth).completeSync();
+        ClientTest& client = asClientTest(m_eth);
+        bytes blockBytes = jsToBytes(_blockRLP, OnFailed::Throw);
+        h256 blockHash = BlockHeader::headerHashFromBlock(blockBytes);
+        client.queueBlock(blockBytes, true);
```
gumb0: Why do you do this now instead of `TestClient::addBlock` ? `addBlock` also notified `EthreumHost` to propagate new block over network, not it doesn't do that

libweb3jsonrpc/Test.cpp
```diff
@@ -135,16 +136,26 @@ bool Test::test_addBlock(std::string const& _rlp)
...
 }
 
-bool Test::test_rewindToBlock(int _number)
+std::string Test::test_importRawBlock(string const& _blockRLP)
 {
     try
     {
-        m_eth.rewind(_number);
-        asClientTest(m_eth).completeSync();
+        ClientTest& client = asClientTest(m_eth);
```
gumb0: I think `asClient` is enough now?

gumb0:
I'd also appreciate a new test for this in `JsonRpcSuite`

libweb3jsonrpc/Test.cpp
winsvega: oh. now I think I remember where it was used. in experemental rpc tests? 
the issue is that addBlock was not working properly. we could notify ethereum host from importRawBlock as well

gumb0: I don't know where it was used.

Maybe better move this logic into `ClientTest` and leave only call to it in `Test::test_importRawBlock`

gumb0: Maybe just move new stuff to `TestClient::addBlock` and call from here

winsvega:
Do NOt merge. JsonRpcSuite tests coming

winsvega:
rpc tests not ready

PR #5083, opened 2018-06-20
more info on RLP exception

winsvega:
gives a hint on which rlp data might be responsible for this exception
@gumb0 if no changes required. with next PR I could make all tabs into spaces

### DIFF
--- a/libdevcore/RLP.cpp
+++ b/libdevcore/RLP.cpp
@@ -26,21 +26,32 @@ using namespace dev;
 bytes dev::RLPNull = rlp("");
 bytes dev::RLPEmptyList = rlpList();
 
+namespace {
+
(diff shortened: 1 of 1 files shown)


libdevcore/RLP.cpp
```diff
@@ -32,15 +32,23 @@ RLP::RLP(bytesConstRef _d, Strictness _s):
 	if ((_s & FailIfTooBig) && actualSize() < _d.size())
 	{
 		if (_s & ThrowOnFail)
-			BOOST_THROW_EXCEPTION(OversizeRLP());
-		else
+        {
+            std::stringstream s;
+            s << "Actual size: " << actualSize() << ", data size: " << _d.size();
```
gumb0: Don't duplicate it, make a helper function constructing this string

PR #5086, opened 2018-06-20
fix vmtrace

winsvega:


libethereum/State.cpp
```diff
@@ -612,12 +612,12 @@ std::pair<ExecutionResult, TransactionReceipt> State::execute(EnvInfo const& _en
     auto onOp = _onOp;
 #if ETH_VMTRACE
 	// Run the existing onOp function and the tracer
-	onOp = [&_onOp, &e](uint64_t _steps, uint64_t PC, Instruction inst, bigint
-			newMemSize, bigint gasCost, bigint gas, VMFace const* _vm,
-			ExtVMFace const* voidExt) {
-		_onOp(_steps, PC, inst, newMemSize, gasCost, gas, _vm, voidExt);
-		e.simpleTrace();
-	};
+    onOp = [&_onOp, &e](uint64_t _steps, uint64_t PC, Instruction inst, bigint newMemSize,
```
chfast: How about we just do
```
if (!onOp)
    onOp = e.simpleTrace();
```

I don't think having both traces at the same time is useful.

PR #5088, opened 2018-06-20
Fix mining in RPC tests

gumb0:
This adds to the `Client` ability to send callbacks about mined block and imported block, so that the unit tests can be notified and keep exact track of how many blocks were generated, instead of waiting in the loop.
(This central change is in TestHelper.cpp)

Also some minor changes, e.g. fixing RPC calls for "latest" block.

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -122,6 +122,7 @@ void Client::init(p2p::Host* _extNet, fs::path const& _dbPath, fs::path const& _
     bc().setOnBlockImport([=](BlockHeader const& _info) {
         if (auto h = m_host.lock())
             h->onBlockImported(_info);
+        m_onBlockImport(_info);
     });
 
     if (_forceAction == WithExisting::Rescue)
@@ -493,9 +494,12 @@ void Client::resyncStateFromChain()
     DEV_READ_GUARDED(x_working)
         if (bc().currentHash() == m_working.info().parentHash())
             return;
-        
-    // RESTART MINING
 
+    restartMining();
+}
+
+void Client::restartMining()
+{
     bool preChanged = false;
     Block newPreMine(chainParams().accountStartNonce);
     DEV_READ_GUARDED(x_preSeal)
@@ -512,7 +516,7 @@ void Client::resyncStateFromChain()
             m_working = newPreMine;
         DEV_READ_GUARDED(x_postSeal)
             if (!m_postSeal.isSealed() || m_postSeal.info().hash() != newPreMine.info().parentHash())
-                for (auto const& t: m_postSeal.pending())
+                for (auto const& t : m_postSeal.pending())
                 {
                     LOG(m_loggerDetail) << "Resubmitting post-seal transaction " << t;
                     //                      ctrace << "Resubmitting post-seal transaction " << t;
@@ -604,6 +608,8 @@ void Client::rejigSealing()
                     LOG(m_logger) << "Tried to seal sealed block...";
                     return;
                 }
+                // TODO is that needed? we have "Generating seal on" below
+                LOG(m_loggerDetail) << "Starting to seal block #" << m_working.info().number();
                 m_working.commitToSeal(bc(), m_extraData);
             }
             DEV_READ_GUARDED(x_working)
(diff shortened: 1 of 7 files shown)


winsvega:
could you run this comman locally: 
```
./testeth -t BlockchainTests/bcStateTests -- --filltests
```

and see if its generating the tests correctly it should return 100% and no errors.

libethereum/Client.cpp
```diff
@@ -615,11 +621,14 @@ void Client::rejigSealing()
 
             if (wouldSeal())
             {
-                sealEngine()->onSealGenerated([=](bytes const& header){
-                    if (!this->submitSealed(header))
+                sealEngine()->onSealGenerated([=](bytes const& header) {
+                    LOG(m_logger) << "Block seal generated";
```
chfast: `"Block sealed"`?

gumb0:
@winsvega this command finishes for me with
```
.
C:/dev/cpp-ethereum/test/tools/libtesteth/TestHelper.cpp(444): error: in "BlockchainTests/bcStateTests": executeCmd() has not been implemented for Windows.
C:/dev/cpp-ethereum/test/tools/libtesteth/TestHelper.cpp(444): error: in "BlockchainTests/bcStateTests": executeCmd() has not been implemented for Windows.
C:/dev/cpp-ethereum/test/tools/libtesteth/TestHelper.cpp(444): error: in "BlockchainTests/bcStateTests": executeCmd() has not been implemented for Windows.
C:/dev/cpp-ethereum/test/tools/libtesteth/TestHelper.cpp(444): error: in "BlockchainTests/bcStateTests": executeCmd() has not been implemented for Windows.
C:/dev/cpp-ethereum/test/tools/libtesteth/TestHelper.cpp(444): error: in "BlockchainTests/bcStateTests": executeCmd() has not been implemented for Windows.
22%...
C:/dev/cpp-ethereum/test/tools/libtesteth/TestHelper.cpp(144): error: in "BlockchainTests/bcStateTests": blockhashNonConstArg network not found: ALL
C:/dev/cpp-ethereum/test/tools/libtesteth/TestHelper.cpp(474): error: in "BlockchainTests/bcStateTests": LLL compilation only supported on posix systems.
C:/dev/cpp-ethereum/test/tools/libtesteth/TestHelper.cpp(358): fatal error: in "BlockchainTests/bcStateTests": Bytecode is missing! '{ [[ 1 ]] (BLOCKHASH (BALANCE 0x1000000000000000000000000000000000000000)) }' blockhashNonConstArg_Frontier

*** 8 failures are detected (5 failures are expected) in the test module "Master Test Suite"
```
which seems to be windows-specific.

Actually I think blockhcain tests use another `mine()` function which I didn't touch.

gumb0:
@chfast I don't think it's related, the function I fix here is used only in RPC tests, not in private networks.

Do we have an issue for that? I could debug it.

chfast:
> Do we have an issue for that? I could debug it.

To be reported soon.

libethereum/Client.cpp
gumb0: Well right after this message there goes (on trace verbosity) "Sealing block!" https://github.com/ethereum/cpp-ethereum/blob/dacd74b4d0bba7ef131b19726e4346662f3dd4de/libethereum/Block.cpp#L840

So now it's like seal generated, then this seal is appended to the block.

chfast: Yes, I've seen also "Generating seal..." which sounds better than "Sealing..."

gumb0: Yeah, I think I'll just delete "Sealing block!" which is confusing

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -43,17 +43,24 @@ namespace dev
...
+    int importedBlocks = 0;
+    std::promise<void> allBlocksImported;
+    auto importHandler =
+        _c.setOnBlockImport([_numBlocks, &importedBlocks, &allBlocksImported](BlockHeader const&) {
+            if (++importedBlocks == _numBlocks)
+                allBlocksImported.set_value();
+        });
+
+    _c.startSealing();
+    allBlocksImported.get_future().get();
```
chfast: Replace `.get()` with `.wait()`?

libethereum/Client.h
```diff
@@ -264,6 +281,7 @@ class Client: public ClientBase, protected Worker
 
     /// Called after processing blocks by onChainChanged(_ir)
     void resyncStateFromChain();
+    void restartMining();
```
chfast: A documentation comment needed?

winsvega:
@gumb0  sorry. this command require solidity

gumb0:
@winsvega can you give me the command to generate just one blockchain test, I can try to compare it on my machine

winsvega:
blockchain tests has issues. let me commit to this branch with a fix

gumb0:
Well this is not related to blockchain tests, so maybe better another PR?

winsvega:
ok. you could run the test command locally with my PR
```./testeth -t BlockchainTests/bcStateTests --  --all --singletest blockhashNonConstArg --verbosity 10```

but remove the PR before merging. I'll open the new PR

winsvega:
here the separate PR:
https://github.com/ethereum/cpp-ethereum/pull/5092/files

you could remove my last commit from here

gumb0:
Actually we should have measured blockhain tests filling performance when we changed ethash library

gumb0:
@winsvega Well I didn't get lllc to work, but current develop doesn't seem faster for me, it also often displays 5 or more "." when mining in this command.

Could it be that you compare not with current develop but the version from while ago? We replaced libethash ~month ago https://github.com/ethereum/cpp-ethereum/pull/5051 and it could change performance in your environment

PR #5095, opened 2018-06-26
Don't start the sync when not needed

gumb0:
In the private network with two nodes, when they connected, the sync was started on both nodes. The node with higher difficulty shouldn't start the sync.

### DIFF
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -226,7 +226,7 @@ void BlockChainSync::onPeerStatus(std::shared_ptr<EthereumPeer> _peer)
         // Before starting to exchange the data with the node, let's verify that it's on our chain
         if (!requestDaoForkBlockHeader(_peer))
             // DAO challenge not needed
-            syncPeer(_peer, true); 
+            syncPeer(_peer, false); 
     }
 }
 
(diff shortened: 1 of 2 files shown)


libweb3jsonrpc/IpcServerBase.cpp
```diff
@@ -36,6 +36,7 @@ int const c_bufferSize = 1024;
 template <class S> IpcServerBase<S>::IpcServerBase(string const& _path):
     m_path(_path)
 {
+    clog(VerbosityInfo, "rpc") << "JSONRPC socket path: " << _path;
```
chfast: `JSON-RPC`.

libethereum/BlockChainSync.cpp
```diff
@@ -689,6 +690,7 @@ void BlockChainSync::collectBlocks()
         case ImportResult::UnknownParent:
             if (headers.first + i > m_lastImportedBlock)
             {
+                LOG(m_logger) << "Already known or future time unknown or unknown parent, block #" << headers.first + i << ". Resetting sync.";
```
chfast: clang-format

libethereum/BlockChainSync.cpp
```diff
@@ -676,6 +676,7 @@ void BlockChainSync::collectBlocks()
             break;
         case ImportResult::Malformed:
         case ImportResult::BadChain:
+            LOG(m_logger) << "Malformed or bad chain, block #" << headers.first + i << ". Restarting sync.";
```
chfast: Can you separate the logs? One per "Malformed", one per "BadChain".

gumb0:
@chfast fixed

PR #5097, opened 2018-06-27
Tests for BlockChainSync

gumb0:


gumb0:
The first test improved coverage of BlockChainSync.cpp by 30%, if I interpret codecov correctly

PR #5098, opened 2018-06-27
move block verification to sealEngineFace

winsvega:
make verification possible on NoProof engine

### DIFF
--- a/libethcore/SealEngine.cpp
+++ b/libethcore/SealEngine.cpp
@@ -1,18 +1,18 @@
 /*
-	This file is part of cpp-ethereum.
+    This file is part of cpp-ethereum.
 
-	cpp-ethereum is free software: you can redistribute it and/or modify
-	it under the terms of the GNU General Public License as published by
-	the Free Software Foundation, either version 3 of the License, or
-	(at your option) any later version.
+    cpp-ethereum is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
 
-	cpp-ethereum is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


libethcore/SealEngine.cpp
```diff
@@ -31,7 +31,57 @@ void NoProof::init()
 
 void SealEngineFace::verify(Strictness _s, BlockHeader const& _bi, BlockHeader const& _parent, bytesConstRef _block) const
 {
-	_bi.verify(_s, _parent, _block);
+    _bi.verify(_s, _parent, _block);
+    if (_parent)
```
gumb0: Let's move this after `if (_s != CheckNothingNew)` as it was before

libethcore/SealEngine.cpp
```diff
@@ -31,7 +31,57 @@ void NoProof::init()
...
+
+        if (_bi.gasLimit() < chainParams().minGasLimit)
+            BOOST_THROW_EXCEPTION(InvalidGasLimit() << RequirementError(
+                                      bigint(chainParams().minGasLimit), bigint(_bi.gasLimit())));
+
+        if (_bi.gasLimit() > chainParams().maxGasLimit)
+            BOOST_THROW_EXCEPTION(InvalidGasLimit() << RequirementError(
+                                      bigint(chainParams().maxGasLimit), bigint(_bi.gasLimit())));
+
+        if (_bi.number())
```
gumb0: `if (_bi.number() && _bi.extraData().size() > chainParams().maximumExtraDataSize)` is more readable I think than nested `if`

gumb0:
Solidity tests use NoProof, so check how it works and submit a PR to https://github.com/ethereum/solidity/blob/develop/test/RPCSession.cpp if change in config needed

winsvega:
soltest fix 
https://github.com/ethereum/solidity/pull/4389

PR #5099, opened 2018-06-27
Improve handling of future blocks in sync

gumb0:
Sometimes during sync several blocks with a future timestamp arrive in a row. In this case `BlockQueue` marks the first one as "future block" (adds it to `m_future` and returns `FutureTimeKnown` from `import()`), but for the following ones it's parent is not present neither in blockchain nor in already verified blocks, so it returns `FutureTimeUnknown`, i.e. future time and parent unknown. Then `BlockChainSync` resets the sync interpreting this as block with unknown parent encountered.

The fix is for each future block to check whether it's parent is also future block, and return  `FutureTimeKnown` (future time and parent known) if it is. 
In this case sync continues further without resetting, and future blocks will be imported eventually.

### DIFF
--- a/libdevcore/CommonData.h
+++ b/libdevcore/CommonData.h
@@ -358,4 +358,9 @@ bool contains(T const& _t, V const& _v)
 	return std::end(_t) != std::find(std::begin(_t), std::end(_t), _v);
 }
 
+template <class V>
+bool contains(std::unordered_set<V> const& _set, V const& _v)
+{
+    return _set.find(_v) != _set.end();
+}
 }
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -694,8 +694,10 @@ void BlockChainSync::collectBlocks()
         case ImportResult::UnknownParent:
             if (headers.first + i > m_lastImportedBlock)
             {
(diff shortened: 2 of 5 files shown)


libdevcore/CommonData.h
```diff
@@ -358,4 +358,10 @@ bool contains(T const& _t, V const& _v)
 	return std::end(_t) != std::find(std::begin(_t), std::end(_t), _v);
 }
 
+template <class V>
+bool contains(std::unordered_set<V> const& _set, V const& _v)
+{
+    return _set.find(_v) != _set.end();
```
chfast: You can still use `.count()` here.

libdevcore/CommonData.h
```diff
@@ -358,4 +358,10 @@ bool contains(T const& _t, V const& _v)
...
+    ;
```
chfast: Dangling `;`.

libdevcore/CommonData.h
gumb0: I'm not a fan of this idiom, I think `find()` better expresses the intent.

libdevcore/CommonData.h
gumb0: fixed

libdevcore/CommonData.h
chfast: :+1:

PR #5100, opened 2018-06-27
test_getBlockStatus

winsvega:
returns the status of a last block that has been mined or imported in the test mode.

required for upcoming malicious block tests that will try to import a block and check that client has reported an error upon reading that block. 

also should be used in soltest instead of waiting for the blocknumber to change. 
because if mining goes wrong, blocknumber will not change and soltest would wait for a long time for blocknumber to update.


winsvega:
the solution here is to make:
test_mineBlocks to be synchronous request (retrun only when finished mining) 
test_importRawBlock  to be synchronous request

winsvega:
instead of blockStatus 
test_mineBlocks and test_importRawBlock   will be synchronous

PR #5102, opened 2018-06-29
Ship the interpreter 

chfast:
It adds the `ALETH_INTERPRETER_SHARED` CMake option to additionally build the interpreter as a shared library and include it in the final package. This is useful because this shared library can be used in geth as the VM and people will not have to build it from source.

The interpreter is still linked statically in the aleth client. It's easier to distribute it this way (no install needed on linux, no DLL issue on Windows).

### DIFF
--- a/cmake/EthCompilerSettings.cmake
+++ b/cmake/EthCompilerSettings.cmake
@@ -50,6 +50,13 @@ if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MA
         endif ()
     endif ()
 
+    # Hide all symbols by default.
+    add_compile_options(-fvisibility=hidden)
+    if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+        # Do not export symbols from dependencies.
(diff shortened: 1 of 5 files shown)


cmake/EthCompilerSettings.cmake
```diff
@@ -18,103 +18,110 @@ include(EthCheckCXXCompilerFlag)
...
+    # Hide all symbols by default.
+    add_compile_options(-fvisibility=hidden)
+    if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+        # Do not export symbols from dependencies.
+        set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--exclude-libs,ALL")
+    endif()
+
+    # Check GCC compiler version.
+    if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
+        if("${CMAKE_CXX_COMPILER_VERSION}" VERSION_LESS 4.8)
```
gumb0: We don't support 4.8 anymore, maybe we should change it in this PR

chfast: Will do. Anything else?

Issue #4269, opened 2017-07-17
Implement eth_sign RPC method


pirapira:
On `develop`, when I run `eth --test` and connect to it, I get a `METHOD_NOT_FOUND` for `ethl.sign()`.  This issue keeps track of implementing this method.

```
eth.sign(eth.accounts[0], "a b c")
Error: METHOD_NOT_FOUND: The method being requested is not available on this server
```

PR #5103, opened 2018-06-30
Update web3.eth.signTransaction so that it only signs the transaction

### LINKED ISSUE #5014: web3.eth.signTransaction appears to also submit the transaction

halfalicious:
Fix #5014 

I've updated `web3.eth.signTransaction` so that it just signs the transaction, rather than signing and submitting the tx. Overview of the changes:
1. I updated the signature of `AccountHolder::authenticate` so that it returns a `Secret `and a `TransactionRepercussion `, rather than a `TransactionNotification `(we no longer need to return the tx hash since the tx isn't being submitted). I've also removed the `TransactionNotification` struct definition since it's no longer used.
2. I modified `FixedAccountHolder::authenticate` and `SimpleAccountHolder::authenticate` to implement the new behavior
3. I updated the RPC function (`Eth::eth_signTransaction`) to create the transaction using the secret returned from `SimpleAccountHolder::authenticate`, populate unintialized tx fields with default values (e.g. gas, gas price, nonce), sign and jsonify both the signed tx and its RLP-encoded bytes, and return the json data to the caller.
* Since `signTransaction `needs to return a json object containing both the signed transaction and the RLP-encoded signed tx bytes (more on this below), I created a new json helper function to generate this data. 
* Being able to populate the tx with default values required moving the tx initialization functionality out of `Client::submitTransaction` into a helper function (`Client::populateTransactionWithDefaults`). Here's the code in question:
https://github.com/ethereum/cpp-ethereum/blob/7d0d31c65ea348ad12054f9bc0b11f3baeb9f8d1/libethereum/Client.cpp#L856-L867

Note: `signTransaction `doesn't appear to be defined in the [web3 0.2x.x API docs](https://github.com/ethereum/wiki/wiki/JavaScript-API). As such, I chose to have the API's return data conform to geth's format which is a json object containing 2 items - the signed tx's RLP bytes (stored in a property named `raw`) and the jsonified transaction (stored in a property called `tx`). 

I verified that the modified function operates as expected - `web3.eth.signTransaction` now returns a json object containing the tx json and the RLP'd bytes, and the RLP'd bytes can be sent via `web3.eth.sendRawTransaction`. Additionally, for good measure I verified that `web3.eth.sendTransaction` still works as expected:
```

> var tx = eth.signTransaction({from: personal.listAccounts[0], to: personal.listAccounts[1], value: web3.toWei(0.01)})
undefined
> tx
{
  raw: "0xf86c068504a817c80083015f9094248221ca1dec4d19fad6491a7cf8888c72103d20872386f26fc10000801ba0d761c6808de697b86ce35c742d29d15cad8330271b0d767bdec63b2f14259adaa053dcac0f87cc591fe760cf85453a0313757906541aec7db631aaf6dc07e877b0",
  tx: {
    data: "0x0000000000000000000000000000000000000000000000000000000000000000",
    from: "0x18bf436622de9e94109bea48219a458df1aef261",
    gas: "0x15f90",
    gasPrice: "0x4a817c800",
    hash: "0x234a64f382f06b01d9fb2ff25a6ed881e996bf2238b341566341ac898be0e46a",
    nonce: "0x6",
    r: "0xd761c6808de697b86ce35c742d29d15cad8330271b0d767bdec63b2f14259ada",
    s: "0x53dcac0f87cc591fe760cf85453a0313757906541aec7db631aaf6dc07e877b0",
    sighash: "0xab6d4c06fbe3a2b65220e951b74f3d3cdd94a74ce07a8cd78b26a9f00b9a15fa",
    to: "0x248221ca1dec4d19fad6491a7cf8888c72103d20",
    v: "0x\x00",
    value: "0x2386f26fc10000"
  }
}
> eth.sendRawTransaction(tx.raw)
"0x234a64f382f06b01d9fb2ff25a6ed881e996bf2238b341566341ac898be0e46a"
> eth.pendingTransactions
[{
    blockHash: null,
    blockNumber: 0,
    data: "0x0000000000000000000000000000000000000000000000000000000000000000",
    from: "0x18bf436622de9e94109bea48219a458df1aef261",
    gas: 90000,
    gasPrice: 20000000000,
    hash: "0x234a64f382f06b01d9fb2ff25a6ed881e996bf2238b341566341ac898be0e46a",
    nonce: 6,
    r: "0xd761c6808de697b86ce35c742d29d15cad8330271b0d767bdec63b2f14259ada",
    s: "0x53dcac0f87cc591fe760cf85453a0313757906541aec7db631aaf6dc07e877b0",
    sighash: "0xab6d4c06fbe3a2b65220e951b74f3d3cdd94a74ce07a8cd78b26a9f00b9a15fa",
    to: "0x248221ca1dec4d19fad6491a7cf8888c72103d20",
    transactionIndex: 0,
    v: "0x\x00",
    value: 10000000000000000
> eth.sendTransaction({from: personal.listAccounts[1], to: personal.listAccounts[0], value: web3.fromWei(0.01)})
"0x465a90fba85ad35d1f9784337735cb8b6608add53710cec070bb49ff77ee8b24"
> eth.pendingTransactions
[{
    blockHash: null,
    blockNumber: 0,
    data: "0x0000000000000000000000000000000000000000000000000000000000000000",
    from: "0x18bf436622de9e94109bea48219a458df1aef261",
    gas: 90000,
    gasPrice: 20000000000,
    hash: "0x234a64f382f06b01d9fb2ff25a6ed881e996bf2238b341566341ac898be0e46a",
    nonce: 6,
    r: "0xd761c6808de697b86ce35c742d29d15cad8330271b0d767bdec63b2f14259ada",
    s: "0x53dcac0f87cc591fe760cf85453a0313757906541aec7db631aaf6dc07e877b0",
    sighash: "0xab6d4c06fbe3a2b65220e951b74f3d3cdd94a74ce07a8cd78b26a9f00b9a15fa",
    to: "0x248221ca1dec4d19fad6491a7cf8888c72103d20",
    transactionIndex: 0,
    v: "0x\x00",
    value: 10000000000000000
}, {
    blockHash: null,
    blockNumber: 0,
    data: "0x0000000000000000000000000000000000000000000000000000000000000000",
    from: "0x248221ca1dec4d19fad6491a7cf8888c72103d20",
    gas: 90000,
    gasPrice: 20000000000,
    hash: "0x465a90fba85ad35d1f9784337735cb8b6608add53710cec070bb49ff77ee8b24",
    nonce: 0,
    r: "0xab039903c3851c1d8994a77b773ef5d0573262d5c291d8bc71bbf701fcb71c4b",
    s: "0x7585581e0821496480c999ec8b802bcdf1aeebae544eec8f9838fd1b62cd0751",
    sighash: "0x42d4e2255b57f12922b1973231cd76a99a7765c0ac81c45da0befeb238ca48fa",
    to: "0x18bf436622de9e94109bea48219a458df1aef261",
    transactionIndex: 0,
    v: "0x\x00",
    value: 0
}]

```

### DIFF
--- a/libweb3jsonrpc/Eth.cpp
+++ b/libweb3jsonrpc/Eth.cpp
@@ -320,7 +320,7 @@ Json::Value Eth::eth_signTransaction(Json::Value const& _json)
 	{
 		TransactionSkeleton ts = toTransactionSkeleton(_json);
 		setTransactionDefaults(ts);
-		client()->populateTransactionWithDefaults(ts);
+		ts = client()->populateTransactionWithDefaults(ts);
 		pair<TransactionRepercussion, Secret> ar = m_ethAccounts.authenticate(ts);
 		switch (ar.first)
 		{
@@ -334,12 +334,12 @@ Json::Value Eth::eth_signTransaction(Json::Value const& _json)
 		}
 		default:
 			// TODO: provide more useful information in the exception.
-			BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS));
+			throw JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS);
 		}
 	}
 	catch (...)
 	{
-		BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS));
+		throw JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS);
 	}
 }
 
--- a/libweb3jsonrpc/Eth.cpp
+++ b/libweb3jsonrpc/Eth.cpp
@@ -250,13 +250,20 @@ string Eth::eth_sendTransaction(Json::Value const& _json)
 	{
 		TransactionSkeleton t = toTransactionSkeleton(_json);
 		setTransactionDefaults(t);
-		TransactionNotification n = m_ethAccounts.authenticate(t);
-		switch (n.r)
+		pair<TransactionRepercussion, Secret> ar = m_ethAccounts.authenticate(t);
+		switch (ar.first)
 		{
 		case TransactionRepercussion::Success:
-			return toJS(n.hash);
+		{
+			pair<h256, Address> txResult = client()->submitTransaction(t, ar.second);
+			return toJS(txResult.first);
+		}
 		case TransactionRepercussion::ProxySuccess:
-			return toJS(n.hash);// TODO: give back something more useful than an empty hash.
+		{
+			m_ethAccounts.queueTransaction(t);
+			h256 emptyHash;
+			return toJS(emptyHash); // TODO: give back something more useful than an empty hash.
+		}
 		case TransactionRepercussion::UnknownAccount:
 			throw JsonRpcException("Account unknown.");
 		case TransactionRepercussion::Locked:
@@ -307,19 +314,24 @@ string Eth::eth_sendTransaction(Json::Value const& _json)
 	return string();
 }
 
-string Eth::eth_signTransaction(Json::Value const& _json)
+Json::Value Eth::eth_signTransaction(Json::Value const& _json)
 {
 	try
 	{
-		TransactionSkeleton t = toTransactionSkeleton(_json);
-		setTransactionDefaults(t);
-		TransactionNotification n = m_ethAccounts.authenticate(t);
-		switch (n.r)
+		TransactionSkeleton ts = toTransactionSkeleton(_json);
+		setTransactionDefaults(ts);
+		client()->populateTransactionWithDefaults(ts);
+		pair<TransactionRepercussion, Secret> ar = m_ethAccounts.authenticate(ts);
+		switch (ar.first)
 		{
 		case TransactionRepercussion::Success:
-			return toJS(n.hash);
 		case TransactionRepercussion::ProxySuccess:
-			return toJS(n.hash);// TODO: give back something more useful than an empty hash.
+		{
+			Transaction t(ts, ar.second);
+			RLPStream s;
+			t.streamRLP(s);
+			return toJson(t, s.out());
+		}
 		default:
 			// TODO: provide more useful information in the exception.
 			BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS));
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -807,6 +807,20 @@ SyncStatus Client::syncStatus() const
     return status;
 }
 
+void Client::populateTransactionWithDefaults(TransactionSkeleton& _t) const
+{
+    // Default gas value meets the intrinsic gas requirements of both
+    // send value and create contract transactions and is the same default
+    // value used by geth and testrpc.
+    const u256 defaultTransactionGas = 90000;
+    if (_t.nonce == Invalid256)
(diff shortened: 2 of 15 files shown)


halfalicious:
@chfast  / @gumb0 : Would love a review when you get a chance

libweb3jsonrpc/Eth.cpp
```diff
@@ -250,13 +250,20 @@ string Eth::eth_sendTransaction(Json::Value const& _json)
 	{
 		TransactionSkeleton t = toTransactionSkeleton(_json);
 		setTransactionDefaults(t);
-		TransactionNotification n = m_ethAccounts.authenticate(t);
-		switch (n.r)
+		pair<TransactionRepercussion, Secret> ar = m_ethAccounts.authenticate(t);
```
chfast: I wander if that would be possible to sign the transaction here (and return the already signed transaction). It's a bit bad that the RPC wrapper has to deal with private keys.

halfalicious:
Thanks for the feedback guys! 

@gumb0 I'll make the function signature change (which will remove the need for adding the "io_" prefix for the TransactionSkeleton parameter, I wasn't aware of this convention thanks for calling it out!)

@chfast In retrospect, I agree that having the RPC layer handle private keys is undesirable so I'll keep the signing out of the RPC layer. I'll also try to add an RPC test for the signTransaction function and will let you know if I have any questions. I expect to have the next iteration ready for review early/mid next week.

halfalicious:
@chfast : So unfortunately keeping the `Secret `within `AccountHolder::authenticate` looks like it will be a little tricky, the problem being that `authenticate `cannot return a `TransactionSkeleton` since `TransactionSkeleton `doesn't support signatures - this functionality is supported by `TransactionBase`: https://github.com/ethereum/cpp-ethereum/blob/d91f7ce6a0973b915352a1aec0207c1a91d71ae7/libethcore/TransactionBase.h#L183

`AccountHolder::authenticate` also cannot return a signed `Transaction ` because it looks like the proxy account tx import function only supports `TransactionSkeleton `objects: https://github.com/ethereum/cpp-ethereum/blob/d91f7ce6a0973b915352a1aec0207c1a91d71ae7/libweb3jsonrpc/AccountHolder.cpp#L82-L88

I agree that we should keep the `Secret `out of the RPC layer but I think we're going to need to do some refactoring of `AccountHolder `in order to do this and I'll need to do some research to figure out what changes I'll need to make. I can either take care of this as a part of this PR or in a future PR (e.g. as a part of fixing #4269).

Issue #5091, opened 2018-06-22
In private network, cpp-ethereum's cpu miner stops once a new node joining this network


azazhu:
**OS:**
 two Ubuntu 16.04.4 LTS

**Issue:**
In private network, if there is only one node in network, everything works fine: node A can mine blocks and results can be accepted by itself. But once new node B joins this private network, node A's cpu miner stops. Is it a issue?

**Repro:**
1. launch node A, use geth as console and start mining. 2. launch node B, use geth as console and join node A's private network. 3. node A's miner stops

**Details:**
1. launch node A and start mining:
cmdline of cpp-ethereum:
```
./aleth --config /data/local/abc/opensource/orig_eth_cppclient/genesis.json -m on -v 5 --network-id 45676 --ipc --listen 30307 --port 30307 --db-path /data/local/abc/opensource/orig_eth_cppclient/data --private "eth_45676"
```
cmdline of go-ethereum:
```
./geth attach /data/local/abc/opensource/orig_eth_cppclient/data/geth.ipc
		>>miner.setEtherbase(eth.accounts[0])
		>>miner.start(1)
```
output of cpp-ethereum:
```
...
2018-06-22 17:33:38 eth trace Generating seal on#c1d77054…#36
2018-06-22 17:33:38 ethsync host Sending a new block (current is #372151a1…, was #7a003d2c…)
2018-06-22 17:33:38 miner2 block Sealing block!
2018-06-22 17:33:38 miner2 bq Queuing block #314aeb70… for import...
2018-06-22 17:33:38 miner2 blockhdr Expected trie root: 56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421
2018-06-22 17:33:38 miner2 blockhdr Expected uncle hash: 1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347
2018-06-22 17:33:38 miner2 bq Block #314aeb70… is 36 parent is #372151a1…
2018-06-22 17:33:38 miner2 bq OK - ready for chain insertion.
2018-06-22 17:33:38 verifier5 blockhdr Expected trie root: 56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421
2018-06-22 17:33:38 verifier5 blockhdr Expected uncle hash: 1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347
[...]

PR #5109, opened 2018-07-04
Fix unexpected sync start

gumb0:
Continuing to fix issues from https://github.com/ethereum/cpp-ethereum/issues/5091

Sometimes `BlockChainSync::continueSync()` is called after we determined that peer is not suitable for sync (e.g. different genesis hash) and disabled it, but before it is actually dropped - then it tries to start the sync from it anyway.

Here I change it to immediately disconnect from peer if it's not suitable for sync. Just disabling "eth" capability doesn't make much sense, as `BlockChainSync` works only with it and with no other capability, so we could just disconnect.

### DIFF
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -209,24 +209,32 @@ void BlockChainSync::onPeerStatus(std::shared_ptr<EthereumPeer> _peer)
     std::shared_ptr<SessionFace> session = _peer->session();
     if (!session)
         return; // Expired
+    char const* disconnectReason = nullptr;
     if (_peer->m_genesisHash != host().chain().genesisHash())
-        _peer->disable("Invalid genesis hash");
+        disconnectReason = "Invalid genesis hash.";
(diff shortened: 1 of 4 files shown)


libethereum/BlockChainSync.cpp
```diff
@@ -209,24 +209,32 @@ void BlockChainSync::onPeerStatus(std::shared_ptr<EthereumPeer> _peer)
     std::shared_ptr<SessionFace> session = _peer->session();
     if (!session)
         return; // Expired
+    std::string disconnectReason;
```
chfast: Use `char const*` not to construct `std::string` when the LOG is no going to be displayed anyway.

libp2p/Capability.cpp
```diff
@@ -35,6 +35,13 @@ Capability::Capability(std::shared_ptr<SessionFace> _s, HostCapabilityFace* _h,
                 << "; idOffset: " << m_idOffset;
 }
 
+void Capability::disconnect()
+{
+    auto s = session();
+    if (s)
```
chfast: A bit into C++17 spirit, you can `if (auto s = session())`.

gumb0:
@chfast fixed

PR #5111, opened 2018-07-11
make test_mineBlocks synchronous

winsvega:
wait for blocks to be successfully imported in the test mode. 
so no need to call for a block number in order to get the information that the blocks were mined

### DIFF
--- a/libethereum/ClientTest.cpp
+++ b/libethereum/ClientTest.cpp
@@ -24,6 +24,7 @@
 #include <libethereum/ClientTest.h>
 #include <libethereum/EthereumHost.h>
 #include <boost/filesystem/path.hpp>
+#include <future>
 
 using namespace std;
 using namespace dev;
@@ -99,18 +100,24 @@ void ClientTest::modifyTimestamp(int64_t _timestamp)
     onPostStateChanged();
 }
 
-void ClientTest::mineBlocks(unsigned _count)
+bool ClientTest::mineBlocks(unsigned _count)
 {
-    m_blocksToMine = _count;
+    if (wouldSeal())
+        return false;
+    std::promise<void> allBlocksImported;
+    int blocksLeftToImport = _count;
+    auto importHandler =
+        setOnBlockImport([this, &blocksLeftToImport, &allBlocksImported](BlockHeader const&) {
+            if (--blocksLeftToImport == 0)
+            {
+                stopSealing();
+                allBlocksImported.set_value();
+            }
+        });
     startSealing();
-}
-
-void ClientTest::onNewBlocks(h256s const& _blocks, h256Hash& io_changed)
-{
-    Client::onNewBlocks(_blocks, io_changed);
-
-    if (--m_blocksToMine <= 0)
-        stopSealing();
+    future_status ret = allBlocksImported.get_future().wait_for(
+        std::chrono::seconds(m_singleBlockMaxMiningTimeInSeconds * _count));
+    return (ret == future_status::ready);
 }
 
 bool ClientTest::completeSync()
--- a/libethereum/ClientTest.cpp
+++ b/libethereum/ClientTest.cpp
@@ -100,24 +100,32 @@ void ClientTest::modifyTimestamp(int64_t _timestamp)
     onPostStateChanged();
 }
 
-bool ClientTest::mineBlocks(unsigned _count)
+bool ClientTest::mineBlocks(unsigned _count) noexcept
 {
     if (wouldSeal())
         return false;
-    std::promise<void> allBlocksImported;
-    int blocksLeftToImport = _count;
-    auto importHandler =
-        setOnBlockImport([this, &blocksLeftToImport, &allBlocksImported](BlockHeader const&) {
-            if (--blocksLeftToImport == 0)
-            {
-                stopSealing();
-                allBlocksImported.set_value();
-            }
-        });
-    startSealing();
-    future_status ret = allBlocksImported.get_future().wait_for(
-        std::chrono::seconds(m_singleBlockMaxMiningTimeInSeconds * _count));
-    return (ret == future_status::ready);
+    try
+    {
+        std::promise<void> allBlocksImported;
+        int blocksLeftToImport = _count;
+        auto importHandler =
+            setOnBlockImport([this, &blocksLeftToImport, &allBlocksImported](BlockHeader const&) {
+                if (--blocksLeftToImport == 0)
+                {
+                    stopSealing();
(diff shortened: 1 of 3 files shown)


mkalinin:
I see one issue with making `test_mineBlocks` synchronous. 
HTTP servers and clients have request timeout parameter which must be considered when long processing requests like block mining takes place. Possible solutions are:
- use WebSocket only, not a good option, client may not support RPC over WebSocket
- turn off proof-of-work when that method is processed to avoid big delays, here we will have a delay caused by block import only, but that must be very little time
- use a couple of methods: `test_mineBlocks` in asynchronous mode and the one which checks block status (a synchronous method)

From my point of view the best solution is to disable proof-of-work if it doesn't affect test runs @winsvega ?

mkalinin:
Well, if block processing time for some tests tend to be comparable to request timeouts (~60 seconds) then only an option with a pair of methods makes sense. This approach will also work perfect for tests that require proof-of-work.

libweb3jsonrpc/Test.cpp
```diff
@@ -101,7 +102,15 @@ bool Test::test_mineBlocks(int _number)
...
     {
+        std::promise<void> allBlocksImported;
+        int importedBlocks = 0;
+        auto importHandler = m_eth.setOnBlockImport(
+            [_number, &importedBlocks, &allBlocksImported](BlockHeader const&) {
+                if (++importedBlocks == _number)
+                    allBlocksImported.set_value();
+            });
         asClientTest(m_eth).mineBlocks(_number);
+        allBlocksImported.get_future().wait();
```
gumb0: Actually it would be a good idea not to wait here indefinitely if something went wrong with the mining. 
Let's use `wait_for` here and return an error if it failed to mine in some limited amount of time (probably depending on number of blocks requested to mine, minute per block perhaps?)

winsvega:
we could implement what ever socket we want to. IPC/HTTP over TCP and so on. 
the issue is that some malicious blocks might timeout even with mining disabled. 
or a node might go into a loop and never respond. 

so would we use timeout exception or error code with get block status ?

mkalinin:
Guys, I stand for solid `retesteth` design. 
From my point of view it's not a big one in terms of development to add `getBlockStatus` while it can bring desired solidity into `retesteth` high level design cause it seems that this solution (with two methods) is able to cover all possible cases. It would be worse if this part had to be changed since time causing client updates.

gumb0:
@winsvega @mkalinin This discussion seems to conflate two different matters/use cases - mining the blocks (with `test_mineBlocks`) and importing raw blocks (with `test_importRawBlock`)

@winsvega mining shouldn't produce any invalid blocks ever. And equally, the node shouldn't go into infinite loop while mining, otherwise it's a buggy node. But yeah, we should have some timeout, as I noted above, not to make `test_mineBlocks` wait too long.

@winsvega @mkalinin Actually I'm fine with leaving mining async as it is now. The current approach is using the second method `eth_blockNumber` to poll the node to see if mining finished. (note: `getBlockStatus` doesn't have anything to do with it)

The idea behind `getBlockStatus` I think was to similarly poll the node after `test_importRawBlock` call.
But I'm not convinced importing should be async, it should be quite fast and imports only one block.

If you don't agree, what is the motivation to make importing async? What is the exact proposed spec for `getBlockStatus` ?

gumb0:
Getting the status of block after you tried to import invalid one seems like a node-implementation-dependent feature to me.

cpp-ethereum I think saves only the hash of encountered bad blocks, and only if some preliminary checks passed (e.g. if it's malformed RLP or incorrect block structure, we just ingore it, remembering nothing about it)
geth seems to save the full bad blocks (as it has `debug_getBadBlocks` method).

I don't know how other clients deal with importing bad blocks.

mkalinin:
Agree, storing bad block information is an optional feature that not every client might want to have. If we go with synchronous option then it needs additional attention to request timeouts. 
Btw, @gumb0 how hard would it be to implement import statuses for test purposes in cpp-ethereum?

winsvega:
when speaking about importRawBlock we don't know if it is valid or not and how much time will it take to process the block. we could agree on a timeout however.

gumb0:
@mkalinin I guess not very hard

mkalinin:
@gumb0 ok, thanks; it's easy to be implemented in Harmony as well.
So, @winsvega looks like we have all pros and cons of both approaches here. You, as an author of `retesteth`, better decide on what to go with.

mkalinin:
Timeout is defined by the server, e.g. Tomcat's default is 60 seconds

mkalinin:
> So that said I think you could set it to any value you want?

yes, sure

gumb0:
(tomcat is http server)
@winsvega I don't think there's any timeout on the IPC socket that you use to talk to the node.

libethereum/ClientTest.cpp
```diff
@@ -101,8 +102,17 @@ void ClientTest::modifyTimestamp(int64_t _timestamp)
 
 void ClientTest::mineBlocks(unsigned _count)
 {
+    std::promise<void> allBlocksImported;
+    int blocksLeftToImport = _count;
+    auto importHandler =
+        setOnBlockImport([&blocksLeftToImport, &allBlocksImported](BlockHeader const&) {
+            if (--blocksLeftToImport == 0)
+                allBlocksImported.set_value();
```
gumb0: You can also `stopSealing()` here and then get rid of `onNewBlocks` handler

libethereum/ClientTest.cpp
```diff
@@ -101,8 +102,17 @@ void ClientTest::modifyTimestamp(int64_t _timestamp)
...
+        });
     m_blocksToMine = _count;
     startSealing();
+    allBlocksImported.get_future().wait_for(
+        std::chrono::seconds(m_singleBlockMaxMiningTimeInSeconds * _count));
```
gumb0: We should return the error (from `ClientTest::mineBlocks` and from RPC method) in case timeout expired.
See `future::wait_for` return value.

libethereum/ClientTest.cpp
```diff
@@ -101,8 +102,17 @@ void ClientTest::modifyTimestamp(int64_t _timestamp)
 
 void ClientTest::mineBlocks(unsigned _count)
 {
+    std::promise<void> allBlocksImported;
```
gumb0: This whole method will not work quite correctly in case some blocks are already being imported or mined, ideally we should check if mining is already started and just fail in this case, but this is not critical, we can leave it like this.

winsvega:
thanks!

Issue #4999, opened 2018-05-03
Recovery id ("v") value in transaction objects returned by eth_pendingTransactions is escaped


halfalicious:
The recovery id ("v") value in the transaction objects returned by `eth_pendingTransactions` is escaped (once PR #4998 has been merged). For example:
```
> web3.eth.pendingTransactions
[{
    blockHash: null,
    blockNumber: 0,
    data: "0x0000000000000000000000000000000000000000000000000000000000000000",
    from: "0xec75ed1426d02b0237385b596cb3a55e6bbe6c3e",
    gas: 275924,
    gasPrice: 20000000000,
    hash: "0xef625e1b091d7933d0b783e1bd757c8408cad59416f4b478e20d8560127c94d5",
    nonce: 0,
    r: "0xaa77bc1412e9f8947078c983cd6ba67e44efc3e927b6f6b4cbe804b98f7f5e04",
    s: "0x1bf75a7c043cafdb68d3ddcba5c2378dcc5a357d8531fbd2d231c9e5b15db4c7",
    sighash: "0x335f04c01e045d636528e31b556979c0189d9b66b7ac56b81d19bc9e4ce85bfd",
    to: "0x884ae368b946a7e5b11462b1333248388b463cc3",
    transactionIndex: 0,
    v: "0x\x01",
    value: 1000000000000000
}]

```
Note that this is not the case with the pending transactions returned by geth - e.g.:
```
> web3.eth.pendingTransactions
[{
    blockHash: null,
    blockNumber: null,
    from: "0x1cbb11b72a335b709a61d22bca90f7e54bfcf8b1",
    gas: 90000,
    gasPrice: 10526315789,
    hash: "0x7375026ff56417cb3e9c4c73a705e956174393261dd0ff057e74bd87434e2c72",
    input: "0x",
    nonce: 0,
    r: "0x98e84508f9f79b38cd240841e65486441b6838194c46e429880d02b6ae38951e",
    s: "0x6236869b5e094d36dd5b663f6f86289f7adf7183a708e68c28b135a893970f57",
    to: "0x884ae368b946a7e5b11462b1333248388b463cc3",
    transactionIndex: 0,
    v: "0x2a",
    value: 1000000000000000
}]
```

I'm not sure if this has functional impact, but the escaping seems unnecessary.

gumb0:
[...]

PR #5112, opened 2018-07-11
web3RPC export transaction with signature info

winsvega:
allows export of transaction via RPC

### DIFF
--- a/libweb3jsonrpc/JsonHelper.cpp
+++ b/libweb3jsonrpc/JsonHelper.cpp
@@ -130,6 +130,9 @@ Json::Value toJson(dev::eth::Transaction const& _t, std::pair<h256, unsigned> _l
         res["blockHash"] = toJS(_location.first);
         res["transactionIndex"] = toJS(_location.second);
         res["blockNumber"] = toJS(_blockNumber);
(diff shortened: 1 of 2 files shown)


libweb3jsonrpc/JsonHelper.cpp
```diff
@@ -130,6 +130,9 @@ Json::Value toJson(dev::eth::Transaction const& _t, std::pair<h256, unsigned> _l
         res["blockHash"] = toJS(_location.first);
         res["transactionIndex"] = toJS(_location.second);
         res["blockNumber"] = toJS(_blockNumber);
+        res["v"] = toJS(u256(_t.signature().v));
```
gumb0: Why do you need the cast? `v` is not 32 bytes, but 1 byte.
`res["v"] = toJS(_t.signature().v);` works somehow for `toJson(dev::eth::Transaction const& _t)`

gumb0: Ah I think it doesn't work well https://github.com/ethereum/cpp-ethereum/issues/4999 Maybe you could fix that too,  returning 62 `0`s here is not cool anyway

PR #5114, opened 2018-07-12
[FROZEN] nest ClientTest from EthashClient

winsvega:
so that aleth --test mode would work with RPC requests that would try to cast *client to EthashClient

### DIFF
--- a/libethereum/ClientTest.cpp
+++ b/libethereum/ClientTest.cpp
@@ -44,7 +44,7 @@ ClientTest* dev::eth::asClientTest(Interface* _c)
 ClientTest::ClientTest(ChainParams const& _params, int _networkID, p2p::Host* _host,
     std::shared_ptr<GasPricer> _gpForAdoption, fs::path const& _dbPath, WithExisting _forceAction,
     TransactionQueue::Limits const& _limits)
-  : Client(
+  : EthashClient(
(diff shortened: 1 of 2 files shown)


gumb0:
Tests are failing now with `InvalidSealEngine` exception

gumb0:
Which RPC methods do you mean?

I'm not a fan of this change, `ClientTest` is something that supports both seal engines + provides some additional RPC methods. We can't say it's related to `EthashClient` with is-a relation, i.e. not any `ClientTest` is a `EthashClient`

Maybe we should get rid of `EthashClient` and the casts you mention altogether...

winsvega:
Yes. Casts are bad. I did some RPC tests. So isMining method does not send any response when sent to eth with --test mode.

gumb0:
Do you intend to use `eth_mining` only when you do full PoW or for NoProof, too?

Currently `isMining` makes sense only for real Ethash - it asks the mining farm whether worker threads are actually working on the mining task.

winsvega:
This looks like a big refactoring project

I am just covering RPC requests being handeled correctly. A node should return correct json in those cases.

winsvega:
This request / result could be tested with retesteth

gumb0:
Should we close this?

winsvega:
I think we can. Right now I focus on other issues

PR #5116, opened 2018-07-16
Cleanup p2p::HostCapabilityFace

gumb0:
- Make `HostCapabilityFace` pure virtual interface, move all concrete members and data to `HostCapability`
- Remove all friends of `HostCapabilityFace`
- Move `HostCapabilityFace::peerSessions()` implementation into `Host`.


libp2p/HostCapability.h
```diff
@@ -36,46 +37,28 @@ namespace p2p
...
+    virtual void onStopping() = 0;
 };
 
 template<class PeerCap>
 class HostCapability: public HostCapabilityFace
 {
 public:
-    HostCapability() {}
-    virtual ~HostCapability() {}
+    explicit HostCapability(p2p::Host* _host) : m_host(_host) { assert(_host); }
```
chfast: If possible, the `_host` should be passed by reference here.

gumb0: Changed to reference

Issue #5090, opened 2018-06-22
Descriptive error messages should be returned when web3.eth.sendTransaction fails due to account / user issues 


halfalicious:
Once #5076 is merged, descriptive error messages will be returned when `web3.eth.sendTransaction` fails, with the exception of account / user-related errors (e.g. account is locked, user-rejected tx). If sending a transaction fails because of an account or user-related error, a generic RPC "invalid method parameters" message is returned. For example, take the following case of a user-rejected tx:

	
```C:\Users\nilse\Documents\temp>geth attach ipc:\\.\pipe\geth.ipc
	Welcome to the Geth JavaScript console!
	
	instance: aleth/1.4.0.dev1+commit.05a43ac3.dirty/windows/msvc19.14.26430.0/relwithdebinfo
	coinbase: 0x0000000000000000000000000000000000000000
	> web3.eth.sendTransaction({from: personal.listAccounts[0], to: personal.listAccounts[1], value:web3.toWei(0)})
	Error: INVALID_PARAMS: Invalid method parameters (invalid name and/or type) recognised
	    at web3.js:3143:20
	    at web3.js:6347:15
	    at web3.js:5081:36
	    at <anonymous>:1:1
```

Note that the expected exception error messages in the `sendingShouldFail `tests in `ClientTests/Personal `will also need to be updated. These tests can be found here: https://github.com/ethereum/cpp-ethereum/blob/dacd74b4d0bba7ef131b19726e4346662f3dd4de/test/unittests/libweb3jsonrpc/Client.cpp#L109-L132

halfalicious:
@gumb0: This can be closed too, it's been addressed by #5117.

PR #5117, opened 2018-07-17
Update sendRawTransaction RPC code path to perform pre-import tx validation and return informative error messages

### LINKED ISSUE #5093: Eth::eth_sendRawTransaction should return descriptive error messages on failure
Similar to #3121 , `Eth::eth_sendRawTransaction` should return descriptive error messages on failure. It currently returns: 

	Error: INVALID_PARAMS: Invalid method parameters (invalid name and/or type) recognised

Note that it will probably make sense to first fix #5014 so it will be easier to test a fix for this issue.

The code in question is here: https://github.com/ethereum/cpp-ethereum/blob/80c4e31aef3d040ca1857d5eb0c048192b952231/libweb3jsonrpc/Eth.cpp#L346-L362

halfalicious:
Fix #5093 

I've updated the `sendRawTransaction `RPC code path to perform pre-import tx validation and to return informative error messages, similar to how `sendTransaction `does things. I've also added a jsonrpc unit test which validates a positive case and a variety of the error cases.

The changes are straight-forward - I'm able to re-use the tx pre-submission validation and tx queue import code in `Client::submitTransaction` by moving it into a new function called `Client::importTransaction`. This also allows me to remove `Client::injectTransaction`.

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -868,10 +868,16 @@ void Client::rewind(unsigned _n)
 
 pair<h256, Address> Client::submitTransaction(TransactionSkeleton const& _t, Secret const& _secret)
 {
-    prepareForTransaction();
     TransactionSkeleton ts = populateTransactionWithDefaults(_t);
     ts.from = toAddress(_secret);
     Transaction t(ts, _secret);
+    importTransaction(t);
+    return make_pair<h256, Address>(t.sha3(), toAddress(ts.from, ts.nonce));
+}
+
+void Client::importTransaction(Transaction const& _t)
+{
+    prepareForTransaction();
 
     // Use the Executive to perform basic validation of the transaction
     // (e.g. transaction signature, account balance) using the state of
@@ -879,8 +885,8 @@ pair<h256, Address> Client::submitTransaction(TransactionSkeleton const& _t, Sec
     // we'll catch the exception at the RPC level.
     Block currentBlock = block(bc().currentHash());
     Executive e(currentBlock, bc());
-    e.initialize(t);
-    ImportResult res = m_tq.import(t.rlp());
+    e.initialize(_t);
+    ImportResult res = m_tq.import(_t.rlp());
     switch (res)
     {
         case ImportResult::Success:
@@ -896,8 +902,6 @@ pair<h256, Address> Client::submitTransaction(TransactionSkeleton const& _t, Sec
         default:
             BOOST_THROW_EXCEPTION(UnknownTransactionValidationError());
     }
-
-    return make_pair(t.sha3(), toAddress(ts.from, ts.nonce));
 }
 
 // TODO: remove try/catch, allow exceptions
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -866,16 +866,15 @@ void Client::rewind(unsigned _n)
     m_bq.clear();
 }
 
-pair<h256, Address> Client::submitTransaction(TransactionSkeleton const& _t, Secret const& _secret)
+h256 Client::submitTransaction(TransactionSkeleton const& _t, Secret const& _secret)
 {
     TransactionSkeleton ts = populateTransactionWithDefaults(_t);
     ts.from = toAddress(_secret);
     Transaction t(ts, _secret);
-    importTransaction(t);
-    return make_pair<h256, Address>(t.sha3(), toAddress(ts.from, ts.nonce));
+    return importTransaction(t);
 }
 
-void Client::importTransaction(Transaction const& _t)
+h256 Client::importTransaction(Transaction const& _t)
 {
     prepareForTransaction();
 
@@ -902,6 +901,8 @@ void Client::importTransaction(Transaction const& _t)
         default:
             BOOST_THROW_EXCEPTION(UnknownTransactionValidationError());
     }
+
+    return _t.sha3();
 }
 
 // TODO: remove try/catch, allow exceptions
--- a/libethereum/Transaction.cpp
+++ b/libethereum/Transaction.cpp
@@ -73,47 +73,6 @@ TransactionException dev::eth::toTransactionException(Exception const& _e)
 	return TransactionException::Unknown;
 }
 
-string dev::eth::transactionExceptionToErrorMessage(Exception const& _e)
-{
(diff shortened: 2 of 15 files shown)


halfalicious:
@chfast or @gumb0 : Would appreciate a review when you get a chance

libethereum/Client.cpp
```diff
@@ -868,19 +868,25 @@ void Client::rewind(unsigned _n)
...
 {
-    prepareForTransaction();
     TransactionSkeleton ts = populateTransactionWithDefaults(_t);
     ts.from = toAddress(_secret);
     Transaction t(ts, _secret);
+    importTransaction(t);
+    return make_pair<h256, Address>(t.sha3(), toAddress(ts.from, ts.nonce));
+}
+
+void Client::importTransaction(Transaction const& _t)
```
gumb0: `submitTransaction` and `importTransaction` are very similar in what they do, I think it would be nice if their signature is similar, so I propose to return `pair<h256, Address` here, too. Then you won't need to calculate transaction hash in the RPC layer in `eth_sendRawTransaction`.

Also, I think `Address` part of result might not be used anywhere, if so we can return just hash.

libweb3jsonrpc/Eth.cpp
```diff
@@ -359,17 +359,51 @@ string Eth::eth_sendRawTransaction(std::string const& _rlp)
...
-		}
-		else
-			return toJS(h256());
+		// Don't need to check the transaction signature (CheckTransaction::None) since it will
+		// be checked as a part of transaction import
+		Transaction t(jsToBytes(_rlp, OnFailed::Throw), CheckTransaction::None);
+		client()->importTransaction(t);
+		return toJS(t.sha3());
+	}
+	catch (InvalidSignature&)
```
gumb0: Can we get rid of duplication between `eth_sendRawTransaction` and `eth_sendTransaction` by catching `Exception` in both and having a helper like `string transactionExceptionToErrorMessage(Exception const&)` ?

test/unittests/libweb3jsonrpc/jsonrpc.cpp
```diff
@@ -319,6 +336,90 @@ BOOST_AUTO_TEST_CASE(eth_sendTransaction)
...
+
+    BOOST_TEST_CHECKPOINT("Sending a signed transaction from 0 balance account should fail");
+
+    int blockNumber = 0;
+    BOOST_CHECK_EQUAL(0, web3->ethereum()->balanceAt(senderAddress, blockNumber));
+
+    Json::Value t;
+    t["from"] = toJS(senderAddress);
+    t["to"] = toJS(receiver.address());
+    t["value"] = jsToDecimal(toJS(0.01));
```
gumb0: I think value is always in wei (so should be integer) and this will result in transaction sending 0 value

test/unittests/libweb3jsonrpc/jsonrpc.cpp
```diff
@@ -319,6 +336,90 @@ BOOST_AUTO_TEST_CASE(eth_sendTransaction)
     BOOST_CHECK_EQUAL(txAmount, balance2);
 }
 
+BOOST_AUTO_TEST_CASE(eth_sendRawTransaction)
+{
+    auto senderAddress = coinbase.address();
+    auto receiver = KeyPair::create();
+    
+
+    BOOST_TEST_CHECKPOINT("Sending a signed transaction from 0 balance account should fail");
```
gumb0: I'd prefer them to be separate tests instead of several test cases separated by checkpoints, to make them more isolated, but that's not critical.

halfalicious:
Note that my latest commit will also fix #5090 (due to the changed exception handling in `eth::eth_sendTransaction`)

halfalicious:
@gumb0: this is ready for another review whenever you get a chance.

libethereum/Transaction.cpp
```diff
@@ -73,6 +73,47 @@ TransactionException dev::eth::toTransactionException(Exception const& _e)
 	return TransactionException::Unknown;
 }
 
+string dev::eth::transactionExceptionToErrorMessage(Exception const& _e)
```
gumb0: These error messages are very RPC-specific, I would leave this function in `libweb3jsonrpc`, probably in `Eth.cpp`

libethereum/Transaction.cpp
```diff
@@ -73,6 +73,47 @@ TransactionException dev::eth::toTransactionException(Exception const& _e)
...
+{
+	if (!!dynamic_cast<ZeroSignatureTransaction const*>(&_e))
```
gumb0: Wouldn't `throw _e` and then the series of `catch`es be less ugly than this series of casts?

libweb3jsonrpc/Eth.cpp
```diff
@@ -275,43 +275,14 @@ string Eth::eth_sendTransaction(Json::Value const& _json)
 			throw JsonRpcException("Unknown authentication error.");
 		}
 	}
-	catch (ZeroSignatureTransaction&)
+	catch (JsonRpcException&)
```
gumb0: Now to avoid this weird pattern surfacing up again, can we maybe `throw / catch(Exception const&)` around just `submitTransaction` call?

Or alternatively - instead of `throw JsonRpcException` in `case`s above assign some variable like `string authenticateError` and then do `throw JsonRpcException(authenticateError)` after `catch Exception`.

libethereum/Transaction.cpp
halfalicious: I'll change the code - the reason I went with this ugly style was that it's used in ```Transaction::toTransactionException``` and I've been trying to keep my changes consistent with the existing code where possible: https://github.com/ethereum/cpp-ethereum/blob/4ae74c6ef7d53bf2434ce43d472396ce0ce46d03/libethereum/Transaction.cpp#L42-L74

This seems like a case where it would make sense to use a different approach so I'll go with something like what you've mentioned. I'll also refactor ```Transaction::toTransactionException```.

halfalicious: Actually I'll leave ```toTransactionException``` as is - refactoring it would require changing the signature (since we no longer need the ```_e``` parameter) and it's referenced in a bunch of unit tests and in the Executive and I'd like to avoid the unnecessary code churn.

gumb0:
Thank you for your work!

PR #5120, opened 2018-07-17
Update SealEngine.cpp

sctChain:
modify for Transaction call, Signature is not needed for call transactions


gumb0:
This is touching the code that is not active in any of usual configs, what are you trying to achieve?

Issue #5119, opened 2018-07-17
Implement Skinny CREATE2 opcode


gumb0:
https://eips.ethereum.org/EIPS/eip-1014

I think we should have CREATE2 implementation already in place, but the formula in the latest spec slightly changed.

Some backgrund:
Account abstraction EIP where CREATE2 idea originated https://github.com/ethereum/EIPs/pull/208
Some discussion in Skinny CREATE2 PR https://github.com/ethereum/EIPs/pull/1014

gumb0:
This will involve passing `salt` parameter through EVMC

axic:
I think this is done now, via #5121 and #5125.

PR #5121, opened 2018-07-17
CREATE2 implementation in LegacyVM

gumb0:
https://github.com/ethereum/cpp-ethereum/issues/5119

### DIFF
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -365,7 +365,7 @@ bool Executive::createOpcode(Address const& _sender, u256 const& _endowment, u25
 
 bool Executive::create2Opcode(Address const& _sender, u256 const& _endowment, u256 const& _gasPrice, u256 const& _gas, bytesConstRef _init, Address const& _origin, u256 const& _salt)
 {
-    m_newAddress = right160(sha3(_sender.asBytes() + toBigEndian(_salt) + sha3(_init).asBytes()));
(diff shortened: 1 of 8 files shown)


gumb0:
Somehow the test succeeds on Windows but generated `BadInstruction` elsewhere 😕

gumb0:
Oh wow, switch statement in VM is implemented with some gcc extention magic on *nixes, I didn't notice it

gumb0:
Still some details are not finalized:
- whether the address formula has additional `0xff` prefix
- whether salt is appended as big endian integer or as byte array
see https://github.com/ethereum/EIPs/pull/1014#issuecomment-406193182

Ortherwise this can be merged I guess.

libethereum/Executive.cpp
```diff
@@ -365,7 +365,7 @@ bool Executive::createOpcode(Address const& _sender, u256 const& _endowment, u25
 
 bool Executive::create2Opcode(Address const& _sender, u256 const& _endowment, u256 const& _gasPrice, u256 const& _gas, bytesConstRef _init, Address const& _origin, u256 const& _salt)
 {
-    m_newAddress = right160(sha3(_sender.asBytes() + toBigEndian(_salt) + sha3(_init).asBytes()));
+    m_newAddress = right160(sha3(_sender.asBytes() + toBigEndian(_salt) + _init));
```
karalabe: I think the consensus is going towards adding the `0xff` prefix. The benefit is that old addresses can't clash with new ones, which is a win because it prevents a class of corner cases that might appear in old create code.

PR #5125, opened 2018-07-24
CREATE2 implementation in aleth-interpreter

gumb0:
#5119 

Required change in EVMC https://github.com/ethereum/evmc/pull/45


libaleth-interpreter/VMCalls.cpp
```diff
@@ -147,11 +138,24 @@ void VM::caseCreate()
...
+        {
+            msg.input_data = &m_mem[off];
+            msg.input_size = size;
+            msg.kind = EVMC_CREATE;  // FIXME: In EVM-C move the kind to the top.
+        }
+        else
+        {
+            assert(m_OP == Instruction::CREATE2);
+            // Pass salt data through EVMC as 32-byte prefix in evmc_message::input_data
+            saltAndInputData = toBigEndian(salt) + bytesConstRef(&m_mem[off], size);
```
axic: If we go with this, it must be documented in evmc properly.

I think it is a stretch, however uncontrollably growing `evmc_message` isn't a better option.

chfast: Because using the message input to pass the salt requires a copy of the init code, I'm more into adding the salt param to the message.

More generic approach would be to ask the VM to come up with the address and put it in the `message.destination`.

gumb0: I'll change it to salt as a separate member of `evmc_message` then.

gumb0:
Still need to support CREATE2 calls through `EVM::execute` wrapper

libaleth-interpreter/VMCalls.cpp
```diff
@@ -138,24 +138,12 @@ void VM::caseCreate()
...
-            msg.input_data = saltAndInputData.data();
-            msg.input_size = saltAndInputData.size();
-            msg.kind = EVMC_CREATE2;
-        }
+        msg.input_data = &m_mem[off];
+        msg.input_size = size;
+        msg.salt = toEvmC(salt);
         msg.sender = m_message->destination;
         msg.depth = m_message->depth + 1;
+        msg.kind = m_OP == Instruction::CREATE ? EVMC_CREATE : EVMC_CREATE2;  // FIXME: In EVM-C move the kind to the top.
```
axic: "EVM-C" -> "EMVC" :)

gumb0:
I realized that no more changes are needed in `EVM` and `ExtVMFace`, because if it's the execution of `CREATE2` itself (executing contract init code), then it's just the same call to `EVM::execute` with `ExtVMFace::isCreate == true` as for `CREATE`, just with another address in `ExtVMFace::myAddress`.

So this is ready for review.

PR #5131, opened 2018-07-26
Update EVMC to a4f70293 (ABI 3)

axic:


chfast:
Done in https://github.com/ethereum/cpp-ethereum/pull/5132.

chfast:
Sorry for not taking it, but it was caused by pure race condition.

Issue #5118, opened 2018-07-17
Implement EXTCODEHASH opcode


gumb0:
https://eips.ethereum.org/EIPS/eip-1052

PR #5134, opened 2018-07-26
Implement EXTCODEHASH in LegacyVM

gumb0:
https://github.com/ethereum/cpp-ethereum/issues/5118


axic:
Is it possible to introduce it in both LegacyVM and aleth-interpreter in the same PRs?

PR #5139, opened 2018-07-26
Release automation

chfast:
This this change we can make a binary release with single button push (use bumpversion for this).
This also links libstdc++ and leveldb statically in the binary for linux, so solidity and hera will be able to use them easily.


axic:
Cool, with this I hope we can simplify Hera's CI.

circle.yml
```diff
@@ -27,7 +27,11 @@ defaults:
     run:
       name: "Build"
       working_directory: ~/build
-      command: cmake --build . --target package -- -j $BUILD_PARALLEL_JOBS
+      command: |
+        cmake --build . --target package -- -j $BUILD_PARALLEL_JOBS
+        mkdir -p ../package
+        . aleth/buildinfo.sh
+        mv aleth.tar.gz ../package/$PROJECT_NAME-$PROJECT_VERSION-$SYSTEM_NAME-$SYSTEM_PROCESSOR.tar.gz
```
axic: Can/should this include the commit hash?

circle.yml
```diff
@@ -196,16 +207,49 @@ jobs:
...
+      - run:
+          name: "Install ghr"
+          command: go get -u github.com/tcnksm/ghr
+      - attach_workspace:
+          at: ~/package
+      - run:
+          name: "Create GitHub release"
+          command: |
+            prerelease_flag=$([[ $CIRCLE_TAG =~ ^v[0-9\.]+$ ]] || echo '-prerelease')
+            ghr -u ethereum -r cpp-ethereum $prerelease_flag $CIRCLE_TAG ~/package
```
axic: Just checked github.com/tcnksm/ghr, nice.

I wonder what the "prerelease" looks like.

axic: Oh its that red icon. In that case I think having commit hash is very useful, but could be done in a separate PR.

circle.yml
chfast: The commit hash is included in `$PROJECT_VERSION`.

PR #5142, opened 2018-07-27
EVMC: do not map unknown non-critical errors as internal errors

axic:
Addresses https://github.com/ewasm/tests/pull/52#issuecomment-408276981

### DIFF
--- a/libevm/EVMC.cpp
+++ b/libevm/EVMC.cpp
@@ -73,8 +73,14 @@ owning_bytes_ref EVMC::exec(u256& io_gas, ExtVMFace& _ext, const OnOpFunc& _onOp
         cwarn << "Execution rejected by EVMC, executing with default VM implementation";
         return VMFactory::create(VMKind::Legacy)->exec(io_gas, _ext, _onOp);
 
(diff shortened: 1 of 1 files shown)


libevm/EVMC.cpp
```diff
@@ -73,8 +73,14 @@ owning_bytes_ref EVMC::exec(u256& io_gas, ExtVMFace& _ext, const OnOpFunc& _onOp
         cwarn << "Execution rejected by EVMC, executing with default VM implementation";
         return VMFactory::create(VMKind::Legacy)->exec(io_gas, _ext, _onOp);
 
+    case EVMC_INTERNAL_ERROR:
```
chfast: Why did you added this case?

axic: Only for clarity for the reader, but it is not necessary of course.

axic: Should we have a comment here that negative ones are internal errors while positive ones are non-internal?

axic:
Merging this now to have it fixed (needed by ewasm tests). This could be improved/changed in the future.

PR #5145, opened 2018-07-27
Fix docker build of aleth

chfast:
- Fixes the docker file for aleth image (`scripts/docker/aleth.dockerfile`). It uses `ADD` instead of `git clone` so this depends on the local source being properly checked out.
- Fixes some small CMake issues.
- Add Circle CI job for building and uploading aleth docker image to https://hub.docker.com/r/ethereum/aleth/tags/. Tagged versions are pushed with this tag, besides all pushes to the main branch are built as `:latest`.


axic:
Please rebase.

axic:
I'm not sure I can review the entire PR, but can approve parts of it.

CMakeLists.txt
```diff
@@ -92,8 +92,10 @@ find_package(ethash CONFIG REQUIRED)
 include(ProjectSecp256k1)
 include(ProjectLibFF)
 
-# Import binaryen library before hera to make it available also to testeth.
-include(hera/cmake/ProjectBinaryen.cmake)
+if(TESTS)
```
axic: Approved.

chfast:
Updated.

circle.yml
```diff
@@ -228,6 +228,33 @@ jobs:
             prerelease_flag=$([[ $CIRCLE_TAG =~ ^v[0-9\.]+$ ]] || echo '-prerelease')
             ghr -u ethereum -r cpp-ethereum $prerelease_flag $CIRCLE_TAG ~/package
 
+  # Builds aleth docker image and uploads it to Docker Hub.
+  docker-aleth:
+    docker:
+      - image: circleci/golang
```
christianparpart: Why is the base image a circleci/golang?

cmake/ProjectMPIR.cmake
```diff
@@ -1,8 +1,7 @@
 include(ExternalProject)
-include(GNUInstallDirs)
 
 set(prefix "${CMAKE_BINARY_DIR}/deps")
-set(MPIR_LIBRARY "${prefix}/${CMAKE_INSTALL_LIBDIR}/${CMAKE_STATIC_LIBRARY_PREFIX}mpir${CMAKE_STATIC_LIBRARY_SUFFIX}")
+set(MPIR_LIBRARY "${prefix}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}mpir${CMAKE_STATIC_LIBRARY_SUFFIX}")
```
christianparpart: Below, `CMAKE_INSTALL_LIBDIR` is being added to cmake args list, but here it's being removed and hard-coded to `lib`?

circle.yml
chfast: CircleCI had this one in their example: https://circleci.com/docs/2.0/building-docker-images/#example.

cmake/ProjectMPIR.cmake
chfast: Yes. I'm asking MPIR build to use `lib` as the value of  `CMAKE_INSTALL_LIBDIR`.

chfast: Same as here: https://github.com/ethereum/solidity/blob/develop/cmake/jsoncpp.cmake#L32.

scripts/aleth.py
```diff
@@ -22,7 +22,7 @@
 
 try:
     if wrapper_args.rpc:
-        run_daemon()
+        run_daemon(wrapper_args.rpc)
```
gumb0: Just realized that this will not work if `aleth.py` will be called with `--db-path` or `--ipcpath`.
Here for this container it will be fine of course

circle.yml
```diff
@@ -243,13 +270,29 @@ workflows:
...
       - deploy:
           requires:
             - macos-xcode90
             - linux-clang6
+            - linux-gcc6
           filters:
             branches:
               ignore: /.*/
             tags:
               only: /^v.*/
```
gumb0: I don't know how it works exactly, but maybe this should be something more like `/v[0-9].*/` ?
For the possibility of creating a tag starting with `v` but not a version

CMakeLists.txt
```diff
@@ -25,10 +25,10 @@ list(APPEND CMAKE_MODULE_PATH ${ETH_CMAKE_DIR})
 set(CMAKE_MAP_IMPORTED_CONFIG_DEBUG Release)
 set(CMAKE_MAP_IMPORTED_CONFIG_RELWITHDEBINFO Release)
 
-set(HUNTER_CONFIGURATION_TYPES Release)
-set(HUNTER_JOBS_NUMBER 4)
-set(HUNTER_CACHE_SERVERS "https://github.com/ethereum/hunter-cache")
-set(HUNTER_PASSWORDS_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/Hunter/passwords.cmake)
+set(HUNTER_CONFIGURATION_TYPES Release CACHE STRING "Hunter configuration types")
```
gumb0: What does this change do?

scripts/aleth.py
chfast: The `--db-path` will work, but it will be a path inside the docker.

I don't want to put a lot of effort to make it perfectly robust now, however the idea of having aleth extendable with Python scripting is quite nice (but probably with different architecture).
Anyway, I will try to quickly intercept `--ipcpath` and `--no-ipc` flags.

chfast:
Updated. Can you review again?

scripts/aleth.py
gumb0: I meant that `--dp-path` also changes the IPC socket path

circle.yml
```diff
@@ -228,6 +228,33 @@ jobs:
...
+      - run:
+          name: "Build aleth docker image"
+          command: docker build -t ethereum/aleth:latest -f scripts/docker/aleth.dockerfile .
+      - run:
+          name: "Push aleth docker image"
+          command: |
+            echo $DOCKER_PASSWORD | docker login -u $DOCKER_USER_ID --password-stdin
+            docker push ethereum/aleth:latest
+            echo "Pushed ethereum/aleth:latest"
+            if [[ $CIRCLE_TAG =~ ^v.* ]]; then
```
gumb0: Should we change regexp here, too?

chfast: Yes.

scripts/aleth.py
chfast: Ok, `--db-path` also handled.

scripts/aleth.py
```diff
@@ -5,13 +5,16 @@
...
-from jsonrpcproxy import run_daemon, DEFAULT_PROXY_URL
+from jsonrpcproxy import run_daemon, DEFAULT_PROXY_URL, DEFAULT_BACKEND_PATH
 
 DEFAULT_ALETH_EXEC = '/usr/bin/aleth'
 
 parser = argparse.ArgumentParser(add_help=False)
 parser.add_argument('--aleth-exec', default=DEFAULT_ALETH_EXEC)
-parser.add_argument('--rpc', nargs='?', const=DEFAULT_PROXY_URL, default=False)
+parser.add_argument('--rpc', metavar='URL', nargs='?', const=DEFAULT_PROXY_URL, default=False)
+parser.add_argument('--db-path', help=argparse.SUPPRESS)
```
gumb0: It also can be `-d`

gumb0:
`--db-path` can also be passed as `-d`

chfast:
Fixed.

chfast:
Fingers crossed...

PR #5146, opened 2018-07-27
fixes graceful termination when `aleth --test` is invoked but require…

christianparpart:
Fixes graceful termination when `aleth --test` is invoked but required a `SIGKILL` to terminate.

This was because the UnixSocketServer requires at least one accepting
connection *after* SIGTERM/SIGINT has been received, which usually never
is the case.

This PR implements a trivial workaround by first waiting for up to N
msecs (N=1000 for now) for any input to arrive, and if not, it'll just
retry, hence, evaluating m_running, resulting in `aleth --test` to
properly terminate upon receival of `SIGINT` or `SIGTERM`.


libweb3jsonrpc/UnixSocketServer.cpp
```diff
@@ -103,11 +104,27 @@ bool UnixDomainSocketServer::StopListening()
...
 }
 
+static bool waitForReadable(int fd, unsigned timeoutMillis)
+{
+	fd_set in, out, err;
+	FD_ZERO(&in);
+	FD_ZERO(&out);
+	FD_ZERO(&err);
+	FD_SET(fd, &in);
+	struct timeval tv;
```
gumb0: In a more C++ way this could be `timeval const tv = {  timeoutMillis / 1000, timeoutMillis % 1000 }`

libweb3jsonrpc/UnixSocketServer.cpp
```diff
@@ -103,11 +104,27 @@ bool UnixDomainSocketServer::StopListening()
 	return false;
 }
 
+static bool waitForReadable(int fd, unsigned timeoutMillis)
```
gumb0: Please move it into unnamed namespace above.

gumb0:
Can there be any better approach than calling `select` every second? Like maybe is there a way to make `select` interrupted by SIGINT/SIGTERM ?

christianparpart:
hey @gumb0. of course there is, by using `signalfd()` (Linux-only, not OS/X), or with the help with `eventfd()` in order to break out of the waiting loop (such as `select()`). The latter can be emulated via `pipe()` on OS/X too.

However, I didn't wanna be too intrusive for a very first PR in this repo. I can however look into finding a somewhat minimal solution of the latter suggestion I made.

gumb0:
Let's take a step back, can you describe how exactly do you reproduce it?
I saw this kind of deadlock sometimes, but not every time.

Now when I tried the following steps
- start `aleth -- test`
- send some RPC calls
- send SIGINT to aleth

`accept` call is being interrupted and returns -1. And aleth process immediately finishes.

chfast:
> start aleth -- test
> send some RPC calls
> send SIGINT to aleth

Try without " send some RPC calls".

gumb0:
Still terminates fine without RPC calls.

Also tried running soltest and SIGINT after that, also aleth is terminated fine.

chfast:
I tried it tool and it terminates quickly. Can it be macOS @christianparpart ?

christianparpart:
So I was facing this issue on "Windows Subsystem for Linux" (which basically is a Linux kernel & ABI re-implementation within the Windows NT kernel).

However, I was trying to reproduce this behavior with a minimal test case, and I believe that I get the same behavior even on Ubuntu 16.04's Linux kernel 4.8.0 with the following test case (link below). So I may still be absolutely brain-dead right now, not finding the bug in my head (or we need this patch)

https://gist.github.com/christianparpart/8b89c9dbfba83478d1308ecc32b9753d

This code is a minimalistic TCP pong server that shares the same behavior this PR tries to fix within aleth's AF_UNIX listener code.

christianparpart:
I digged a little bit into it. it's actually not accept() being cancelled with -1 (EINTR) due to the SIGINT but a -1 (EINVAL) due to the unlink() on the socket. While that one should work on all UNIX'es, I don't think this is a clean solution, as only a side effect of closing a socket while being in a blocking I/O is made use of, which seems unreliable wrt. platform independence (?). I do like your solution (lean, but unfortunately wasn't quite obvious from the beginning), but it doesn't work on my end it seems.

gumb0:
So we had in place a solution I hoped for, it just seems to not work on exotic systems.

I would prefer to not merge this (it complicates things quite a bit) and resolving this by you @christianparpart submitting a bug report to Microsoft (is this even a bug?)

Otherwise needs some work on better logging / better error reporting / better code formatting.

gumb0:
@chfast this PR somehow is still to `develop`, not to `master`

christianparpart:
I'd adapt the code to conform to coding style and "codecov" once you say I can proceed (:+1:).

libweb3jsonrpc/UnixSocketServer.cpp
```diff
@@ -92,6 +124,10 @@ bool UnixDomainSocketServer::StartListening()
 
 bool UnixDomainSocketServer::StopListening()
 {
+	int dummy = 0x90;
+	if (::write(m_exitChannel[1], &dummy, sizeof(dummy)) < 0)
+		cerr << "Failed to notify UNIX domain server loop. " << strerror(errno) << "\n";
```
gumb0: Better output it to `clog(VerbosityWarn, "rpc")`, It will not need `\n` in the end.

libweb3jsonrpc/UnixSocketServer.cpp
```diff
@@ -54,16 +57,45 @@ fs::path getIpcPathOrDataDir()
...
+
 }
 
 UnixDomainSocketServer::UnixDomainSocketServer(string const& _appId):
-	IpcServerBase((getIpcPathOrDataDir() / fs::path(_appId + ".ipc")).string().substr(0, c_socketPathMaxLength))
+	IpcServerBase((getIpcPathOrDataDir() / fs::path(_appId + ".ipc")).string().substr(0, c_socketPathMaxLength)),
+	m_exitChannel{-1, -1}
 {
+	if (pipe(m_exitChannel) < 0)
+		abort(); // I don't like. Replace me with throw.
```
gumb0: Throw exception here.
Declare it in `IpcServerBase.h` as
```
DEV_SIMPLE_EXCEPTION(FailedToCreatePipe);
```

Then throw here like
```
    BOOST_THROW_EXCEPTION(FailedToCreatePipe());
```

PR #5147, opened 2018-07-27
support blockchain tests with noProof in testeth

winsvega:
changes only to the testeth part

### DIFF
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -34,12 +34,30 @@ using namespace dev;
 using namespace dev::eth;
 namespace fs = boost::filesystem;
 
+namespace
+{
+// Parse sealEngine field from .json test
+test::TestBlockChain::MiningType getMiningType(json_spirit::mObject const& _testObj)
+{
+    test::TestBlockChain::MiningType miningType = test::TestBlockChain::MiningType::Default;
+    if (_testObj.count("sealEngine") > 0)
+    {
+        if (_testObj.at("sealEngine") == eth::NoProof::name())
+            miningType = test::TestBlockChain::MiningType::ForceNoProof;
+        else if (_testObj.at("sealEngine") == eth::Ethash::name())
+            miningType = test::TestBlockChain::MiningType::ForceEthash;
+    }
+    return miningType;
+}
+}
+
 namespace dev {
 
 namespace test {
 
 eth::Network ChainBranch::s_tempBlockchainNetwork = eth::Network::MainNetwork;
 eth::Network TestBlockChain::s_sealEngineNetwork = eth::Network::FrontierTest;
+
 json_spirit::mValue BlockchainTestSuite::doTests(json_spirit::mValue const& _input, bool _fillin) const
 {
 	json_spirit::mObject tests;
@@ -179,14 +197,15 @@ fs::path TransitionTestsSuite::suiteFillerFolder() const {
 	return fs::path("BlockchainTestsFiller") / "TransitionTests";
 }
 
-ChainBranch::ChainBranch(TestBlock const& _genesis): blockchain(_genesis)
+ChainBranch::ChainBranch(TestBlock const& _genesis, TestBlockChain::MiningType _miningType)
+  : blockchain(_genesis, _miningType)
 {
 	importedBlocks.push_back(_genesis);
 }
 
-void ChainBranch::reset()
+void ChainBranch::reset(TestBlockChain::MiningType _miningType)
 {
-	blockchain.reset(importedBlocks.at(0));
+    blockchain.reset(importedBlocks.at(0), _miningType);
 }
 
 void ChainBranch::restoreFromHistory(size_t _importBlockNumber)
@@ -223,10 +242,13 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input)
 	TestBlock genesisBlock(_input.at("genesisBlockHeader").get_obj(), _input.at("pre").get_obj());
 	genesisBlock.setBlockHeader(genesisBlock.blockHeader());
 
-	TestBlockChain testChain(genesisBlock);
-	assert(testChain.getInterface().isKnown(genesisBlock.blockHeader().hash(WithSeal)));
+    TestBlockChain::MiningType miningType = getMiningType(_input);
+    TestBlockChain testChain(genesisBlock, miningType);
+    assert(testChain.getInterface().isKnown(genesisBlock.blockHeader().hash(WithSeal)));
 
-	output["genesisBlockHeader"] = writeBlockHeaderToJson(genesisBlock.blockHeader());
+    if (_input.count("sealEngine"))
+        output["sealEngine"] = _input.at("sealEngine");
+    output["genesisBlockHeader"] = writeBlockHeaderToJson(genesisBlock.blockHeader());
 	output["genesisRLP"] = toHexPrefixed(genesisBlock.bytes());
 	BOOST_REQUIRE(_input.count("blocks"));
 
@@ -234,9 +256,10 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input)
 	size_t importBlockNumber = 0;
 	string chainname = "default";
 	string chainnetwork = "default";
-	std::map<string, ChainBranch*> chainMap = { {chainname , new ChainBranch(genesisBlock)}};
+    std::map<string, ChainBranch*> chainMap = {
+        {chainname, new ChainBranch(genesisBlock, miningType)}};
 
-	if (_input.count("network") > 0)
+    if (_input.count("network") > 0)
 		output["network"] = _input.at("network");
 
 	for (auto const& bl: _input.at("blocks").get_array())
@@ -277,16 +300,16 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input)
 			if (_input.count("noBlockChainHistory") == 0)
 			{
 				ChainBranch::forceBlockchain(chainnetwork);
-				chainMap[chainname]->reset();
-				ChainBranch::resetBlockchain();
+                chainMap[chainname]->reset(miningType);
+                ChainBranch::resetBlockchain();
 				chainMap[chainname]->restoreFromHistory(importBlockNumber);
 			}
 		}
 		else
 		{
 			ChainBranch::forceBlockchain(chainnetwork);
-			chainMap[chainname] = new ChainBranch(genesisBlock);
-			ChainBranch::resetBlockchain();
+            chainMap[chainname] = new ChainBranch(genesisBlock, miningType);
+            ChainBranch::resetBlockchain();
 		}
 
 		TestBlock block;
@@ -415,10 +438,15 @@ void testBCTest(json_spirit::mObject const& _o)
 {
 	string testName = TestOutputHelper::get().testName();
 	TestBlock genesisBlock(_o.at("genesisBlockHeader").get_obj(), _o.at("pre").get_obj());
-	TestBlockChain blockchain(genesisBlock);
 
-	TestBlockChain testChain(genesisBlock);
-	assert(testChain.getInterface().isKnown(genesisBlock.blockHeader().hash(WithSeal)));
+    TestBlockChain::MiningType const miningType = getMiningType(_o);
+    eth::IncludeSeal includeSeal = (miningType == TestBlockChain::MiningType::ForceEthash ||
+                                       miningType == TestBlockChain::MiningType::Default) ?
+                                       WithSeal :
+                                       WithoutSeal;
+    TestBlockChain blockchain(genesisBlock, miningType);
+    TestBlockChain testChain(genesisBlock, miningType);
+    assert(testChain.getInterface().isKnown(genesisBlock.blockHeader().hash(WithSeal)));
 
 	if (_o.count("genesisRLP") > 0)
 	{
@@ -511,8 +539,9 @@ void testBCTest(json_spirit::mObject const& _o)
 		}
 
 		//Check that imported block to the chain is equal to declared block from test
-		bytes importedblock = testChain.getInterface().block(blockFromFields.blockHeader().hash(WithSeal));
-		TestBlock inchainBlock(toHex(importedblock));
+        bytes importedblock =
+            testChain.getInterface().block(blockFromFields.blockHeader().hash(includeSeal));
+        TestBlock inchainBlock(toHex(importedblock));
 		checkBlocks(inchainBlock, blockFromFields, testName);
 
 		string blockNumber = toString(testChain.getInterface().number());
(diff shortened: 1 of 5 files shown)


test/tools/libtesteth/TestHelper.h
```diff
@@ -82,6 +82,7 @@ std::set<eth::Network> const& getNetworks();
 /// >Homestead to EIP150, EIP158, Byzantium, ...
 /// <=Homestead to Frontier, Homestead
 std::set<std::string> translateNetworks(std::set<std::string> const& _networks);
+std::string prepareGenesisConfig(eth::Network _net);
```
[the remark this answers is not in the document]
winsvega: yes

test/tools/libtesteth/BlockChainHelper.h
```diff
@@ -137,10 +137,19 @@ class TestBlock
...
+    {
+        Ethash,
+        NoProof
+    };
+    TestBlockChain() : TestBlockChain(defaultGenesisBlock(), MiningType::Ethash) {}
+    TestBlockChain(TestBlock const& _genesisBlock, MiningType _mining = MiningType::Ethash);
+
+    // Replace sealEngine depending on a MiningType"
+    // Genesis config is parsed from _net in libethashseal/GenesisInfo.cpp
+    static std::string prepareGenesisConfig(eth::Network _net, MiningType _mining);
```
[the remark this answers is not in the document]
winsvega: it is used in ImportTest.cpp when generating the bc out of state test.
however if it is possible to change sealengine in already parese chain params it might not be needed at all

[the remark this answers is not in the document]
gumb0: Yeah I guess you should be able to modify `ChainOperationParams::sealEngineName`

test/unittests/libethereum/Block.cpp
```diff
@@ -37,7 +37,7 @@ BOOST_FIXTURE_TEST_SUITE(FrontierBlockSuite, FrontierNoProofTestFixture)
 
 BOOST_AUTO_TEST_CASE(bStates)
 {
-	TestBlockChain testBlockchain(TestBlockChain::defaultGenesisBlock());
```
gumb0: No need to change this file, the tests are already in `FrontierNoProofTestFixture`.
Others are  in `ConstantinopleTransitionTestFixture` which also uses `NoProof`

test/unittests/libethereum/BlockChain.cpp
```diff
@@ -38,7 +38,7 @@ BOOST_FIXTURE_TEST_SUITE(BlockChainFrontierSuite, FrontierNoProofTestFixture)
 BOOST_AUTO_TEST_CASE(output)
 {
     TestBlock genesis = TestBlockChain::defaultGenesisBlock();
-    TestBlockChain bc(genesis);
```
gumb0: No changes needed in this file, all tests are already in `FrontierNoProofTestFixture` or in `MainNetworkNoProofTestFixture`

test/unittests/libethereum/BlockQueue.cpp
```diff
@@ -35,8 +35,8 @@ BOOST_FIXTURE_TEST_SUITE(BlockQueueSuite, MainNetworkNoProofTestFixture)
 BOOST_AUTO_TEST_CASE(BlockQueueImport)
 {
     TestBlock genesisBlock = TestBlockChain::defaultGenesisBlock();
-    TestBlockChain blockchain(genesisBlock);
```
gumb0: No changes needed, test is in `MainNetworkNoProofTestFixture`

test/unittests/libethereum/ExtVMTest.cpp
```diff
@@ -31,7 +31,7 @@ class ExtVMTestFixture : public TestOutputHelperFixture
 public:
     ExtVMTestFixture()
       : networkSelector(eth::Network::ConstantinopleTransitionTest),
-        testBlockchain(TestBlockChain::defaultGenesisBlock()),
```
gumb0: No need to change this, `ConstantinopleTransitionTest` uses `NoProof`

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -179,14 +179,14 @@ fs::path TransitionTestsSuite::suiteFillerFolder() const {
...
+ChainBranch::ChainBranch(TestBlock const& _genesis, TestBlockChain::MiningType _miningType): blockchain(_genesis, _miningType)
 {
 	importedBlocks.push_back(_genesis);
 }
 
-void ChainBranch::reset()
+void ChainBranch::reset(TestBlockChain::MiningType _miningType)
 {
-	blockchain.reset(importedBlocks.at(0));
+    blockchain.reset(importedBlocks.at(0), _miningType);
```
gumb0: Is this ok that this `importedBlocks.at(0)` can stay the same even with the change of `_miningType`? i.e. Can the same `TestBlock` work for both types?

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -223,7 +223,15 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input)
...
-	TestBlockChain testChain(genesisBlock);
+    TestBlockChain::MiningType miningType = TestBlockChain::MiningType::Ethash;
+    if (_input.count("sealEngine") > 0)
+    {
+        output["sealEngine"] = _input.at("sealEngine");
+        if (_input.at("sealEngine") == eth::NoProof::name())
+            miningType = TestBlockChain::MiningType::NoProof;
+    }
+
+    TestBlockChain testChain(genesisBlock, miningType);
```
gumb0: if this exists only for assert below, you could do `assert(TestBlockChain(genesisBlock, miningType).getInterface().isKnown(...));`

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -234,7 +242,7 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input)
 	size_t importBlockNumber = 0;
 	string chainname = "default";
 	string chainnetwork = "default";
-	std::map<string, ChainBranch*> chainMap = { {chainname , new ChainBranch(genesisBlock)}};
+    std::map<string, ChainBranch*> chainMap = { {chainname , new ChainBranch(genesisBlock, miningType)}};
```
gumb0: Never use raw pointers. Use `unique_ptr` here.
Extra space before comma.

test/unittests/libethereum/Block.cpp
winsvega: This controdicts your previous requirement of explicit constructor in test blockchain class

gumb0: My comment was about `explicit` keyword in constructor declaration

test/tools/libtesteth/BlockChainHelper.h
```diff
@@ -137,10 +137,16 @@ class TestBlock
...
-
-    void reset(TestBlock const& _genesisBlock);
+    enum class MiningType
+    {
+        Ethash,
+        NoProof,
+        Default
+    };
+    TestBlockChain() : TestBlockChain(defaultGenesisBlock(), MiningType::Default) {}
+    TestBlockChain(TestBlock const& _genesisBlock, MiningType _mining = MiningType::Default);
```
gumb0: This should be `explicit` constructor

test/tools/libtesteth/BlockChainHelper.h
```diff
@@ -137,10 +137,16 @@ class TestBlock
...
 public:
-    TestBlockChain() : TestBlockChain(defaultGenesisBlock()) {}
-    TestBlockChain(TestBlock const& _genesisBlock);
-
-    void reset(TestBlock const& _genesisBlock);
+    enum class MiningType
+    {
+        Ethash,
+        NoProof,
+        Default
```
gumb0: I'd add a comment that `Default` means "don't override the engine from given config"

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -40,6 +40,7 @@ namespace test {
 
 eth::Network ChainBranch::s_tempBlockchainNetwork = eth::Network::MainNetwork;
 eth::Network TestBlockChain::s_sealEngineNetwork = eth::Network::FrontierTest;
+TestBlockChain::MiningType getMiningType(json_spirit::mObject const& _testObj);
```
gumb0: Better put some empty lines between your constants and function declarations, definitions

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -223,9 +224,12 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input)
 	TestBlock genesisBlock(_input.at("genesisBlockHeader").get_obj(), _input.at("pre").get_obj());
 	genesisBlock.setBlockHeader(genesisBlock.blockHeader());
 
-	TestBlockChain testChain(genesisBlock);
+	TestBlockChain::MiningType miningType = getMiningType(_input);
+	TestBlockChain testChain(genesisBlock, miningType);
 	assert(testChain.getInterface().isKnown(genesisBlock.blockHeader().hash(WithSeal)));
 
+	if (_input.count("sealEngine"))
```
gumb0: Is this a new field in blockchain test source files that didn't exist before?

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -415,9 +419,11 @@ void testBCTest(json_spirit::mObject const& _o)
 {
 	string testName = TestOutputHelper::get().testName();
 	TestBlock genesisBlock(_o.at("genesisBlockHeader").get_obj(), _o.at("pre").get_obj());
-	TestBlockChain blockchain(genesisBlock);
 
-	TestBlockChain testChain(genesisBlock);
+    TestBlockChain::MiningType miningType = getMiningType(_o);
+    eth::IncludeSeal includeSeal = (miningType == TestBlockChain::MiningType::Ethash || miningType == TestBlockChain::MiningType::Default) ? WithSeal : WithoutSeal;
```
gumb0: Define a variable closer to where you use it.

test/tools/jsontests/BlockChainTests.cpp
winsvega: possibly could make tests where valid block has invalid PoW and invalid block has valid PoW proof

test/tools/jsontests/BlockChainTests.cpp
winsvega: yes. but it is not required. we have to support the legacy tests.

test/tools/libtesteth/BlockChainHelper.h
winsvega: That contradicts your other requirenment. (not to change the unit tests)

test/tools/jsontests/BlockChainTests.cpp
winsvega: the declaration must be before the line 
```
for (auto const& bl: _o.at("blocks").get_array())
```

test/tools/libtesteth/BlockChainHelper.h
winsvega: fixed

test/tools/jsontests/BlockChainTests.cpp
winsvega: not for this PR

test/tools/jsontests/BlockChainTests.cpp
winsvega: no testChain is required for BC tests logic

test/tools/libtesteth/BlockChainHelper.h
gumb0: No, it shouldn't. You can still create it like `TestBlockChain bc(genesis);` even if it's `explicit`

winsvega: ok. I've added explicit keyword to this line. it works!

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -40,6 +40,10 @@ namespace test {
 
 eth::Network ChainBranch::s_tempBlockchainNetwork = eth::Network::MainNetwork;
 eth::Network TestBlockChain::s_sealEngineNetwork = eth::Network::FrontierTest;
+
+// Parse sealEngine field from .json test
+TestBlockChain::MiningType getMiningType(json_spirit::mObject const& _testObj);
```
gumb0: This better could be in unnamed namespace

test/tools/jsontests/BlockChainTests.cpp
gumb0: Right, but not at the top of the function

winsvega: do you want me to move it 4 lines below ? 
just what is the difference?

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -34,12 +34,31 @@ using namespace dev;
 using namespace dev::eth;
 namespace fs = boost::filesystem;
 
+namespace {
```
gumb0: Brace should be on a separate line. Better use clang-format. Also convert tabs to spaces (AFTER clang-formatting, maybe better in separate commit) in the files you touch

test/tools/jsontests/BlockChainTests.cpp
gumb0: Ideally of course I want you to never ever write loops 110 lines long and functions 150 lines long. Giant scope of a variable is only a consequence of this...

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -415,9 +437,11 @@ void testBCTest(json_spirit::mObject const& _o)
 {
 	string testName = TestOutputHelper::get().testName();
 	TestBlock genesisBlock(_o.at("genesisBlockHeader").get_obj(), _o.at("pre").get_obj());
-	TestBlockChain blockchain(genesisBlock);
 
-	TestBlockChain testChain(genesisBlock);
+    TestBlockChain::MiningType miningType = getMiningType(_o);
+    eth::IncludeSeal includeSeal = (miningType == TestBlockChain::MiningType::ForceEthash || miningType == TestBlockChain::MiningType::Default) ? WithSeal : WithoutSeal;
```
gumb0: Why now we include seal for `Default`, too? What if default changes, will you remember to change this line? (I would not)

test/tools/jsontests/BlockChainTests.cpp
gumb0: I meant here that you have one instance of `TestBlock` (namely `importedBlocks.at(0)`) which you can use after call to `reset` with different seal engine. That looked suspicious.

test/tools/jsontests/BlockChainTests.cpp
winsvega: I will run clang-format on all touched files in a separate commit.

test/tools/jsontests/BlockChainTests.cpp
winsvega: yes. that is if you change _miningType inside the for loop. should set it to const.

test/tools/jsontests/BlockChainTests.cpp
winsvega: Then tests will fail. By default blockchain tests are done with PoW. 
I told you that we should better use explicit constructor with mining type + in unittests.
So that you could remove  HomsteadTestNoProof chain configs.

gumb0:
Don't reformat the whole files please.
What you should do with any of your changes:

1. Apply clang-format to your changes only.
2. Convert tabs to spaces in the touched files.

PR #5149, opened 2018-07-30
Aleth Python wrapper

chfast:
Add Python script wrapping aleth with JSON-RPC HTTP server

### DIFF
--- a/scripts/aleth.py
+++ b/scripts/aleth.py
@@ -1,0 +1,21 @@
+#!/usr/bin/env python3
+
+import argparse
+import subprocess
+
+from jsonrpcproxy import run_daemon
+
+DEFAULT_ALETH_EXEC = '/usr/bin/aleth'
+
+parser = argparse.ArgumentParser(add_help=False)
+parser.add_argument('--aleth-exec', default=DEFAULT_ALETH_EXEC)
+parser.add_argument('--rpc', action='store_true')
+wrapper_args, aleth_args = parser.parse_known_args()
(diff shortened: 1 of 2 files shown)


gumb0:
Does it print any reasonable error message when run without args and aleth is not found?

I'd output the string with expected parameters in this case. Also maybe add `--help`

scripts/aleth.py
```diff
@@ -0,0 +1,21 @@
...
+DEFAULT_ALETH_EXEC = '/usr/bin/aleth'
+
+parser = argparse.ArgumentParser(add_help=False)
+parser.add_argument('--aleth-exec', default=DEFAULT_ALETH_EXEC)
+parser.add_argument('--rpc', action='store_true')
+wrapper_args, aleth_args = parser.parse_known_args()
+
+try:
+    if wrapper_args.rpc:
+        run_daemon()
```
gumb0: Will it still output log messages when run as a daemon

chfast:
> Does it print any reasonable error message when run without args and aleth is not found?
> 
> I'd output the string with expected parameters in this case. Also maybe add --help

I added the error message. It also prints the usage message in case of incorrect path to aleth. The `--help` is not supported because I want it to be passed to aleth, so aleth.py options are hidden. aleth.py is going to be used as docker image entry point.

scripts/aleth.py
chfast: Yes.

PR #5154, opened 2018-08-01
fix filler hash check

winsvega:
git might change the file contents depending on the OS (probably line endings)
that leads to wrong test source hash error check in testeth sometimes.

this pr change check to the hash of a json structure rather than the hash of filler.json file.
requrires all tests update. (which is coming anyway with new hardfork. filling Constantinople rule)

### DIFF
--- a/test/tools/libtesteth/TestSuite.cpp
+++ b/test/tools/libtesteth/TestSuite.cpp
@@ -82,10 +82,15 @@ void addClientInfo(json_spirit::mValue& _v, fs::path const& _testSource, h256 co
 void checkFillerHash(fs::path const& _compiledTest, fs::path const& _sourceTest)
 {
 	json_spirit::mValue v;
-	string const s = asString(dev::contents(_compiledTest));
+	string s = asString(dev::contents(_compiledTest));
 	BOOST_REQUIRE_MESSAGE(s.length() > 0, "Contents of " + _compiledTest.string() + " is empty.");
 	json_spirit::read_string(s, v);
-	h256 const fillerHash = sha3(dev::contents(_sourceTest));
+
+	json_spirit::mValue v2;
+	s = asString(dev::contents(_sourceTest));
+	BOOST_REQUIRE_MESSAGE(s.length() > 0, "Contents of " + _sourceTest.string() + " is empty.");
+	json_spirit::read_string(s, v2);
+	h256 const fillerHash = sha3(json_spirit::write_string(v2, true));
 
 	for (auto& i: v.get_obj())
 	{


gumb0:
Oh great, at last I will be able to run tests on windows!

Should we not merge it until test are regenerated?

winsvega:
yes. I think first to refill tests with 
1) noProof
2) SSTORE change

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -82,10 +82,15 @@ void addClientInfo(json_spirit::mValue& _v, fs::path const& _testSource, h256 co
 void checkFillerHash(fs::path const& _compiledTest, fs::path const& _sourceTest)
 {
 	json_spirit::mValue v;
-	string const s = asString(dev::contents(_compiledTest));
+	string s = asString(dev::contents(_compiledTest));
```
gumb0: you probably could use just `dev::contentsString`

gumb0:
I see the checking code changed, but where's the code that writes that hash to test file, should that be changed, too?

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -82,10 +82,15 @@ void addClientInfo(json_spirit::mValue& _v, fs::path const& _testSource, h256 co
...
 	BOOST_REQUIRE_MESSAGE(s.length() > 0, "Contents of " + _compiledTest.string() + " is empty.");
 	json_spirit::read_string(s, v);
-	h256 const fillerHash = sha3(dev::contents(_sourceTest));
+
+	json_spirit::mValue v2;
```
gumb0: What's `v` & `v2`, give them more meaningful names, like `jsonSource` & `jsonCompiled`

gumb0:
When further changes are needed here, pleaso don't overwrite these commits, because 413e4077 is now mentioned in generated tests

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -18,17 +18,23 @@
...
-#include <test/tools/libtesteth/JsonSpiritHeaders.h>
+#include <boost/algorithm/string.hpp>
 #include <string>
 using namespace std;
 using namespace dev;
 namespace fs = boost::filesystem;
 
 //Helper functions for test proccessing
 namespace {
+struct testFileData
```
gumb0: Type names start with uppercase

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -55,6 +61,25 @@ void removeComments(json_spirit::mValue& _obj)
 	}
 }
 
+testFileData readTestFile(fs::path _testFileName)
+{
+    testFileData testData;
+    bytes const byteContents = dev::contents(_testFileName);
```
gumb0: There's `dev::contentsString` for these two lines

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -55,6 +61,25 @@ void removeComments(json_spirit::mValue& _obj)
...
+
+    if (_testFileName.extension() == ".json")
+        json_spirit::read_string(s, testData.data);
+    else if (_testFileName.extension() == ".yml")
+        testData.data = test::parseYamlToJson(s);
+    else
+        BOOST_ERROR("Unknow test format!" + test::TestOutputHelper::get().testFile().string());
+
+    testData.hash = h256(sha3(json_spirit::write_string(testData.data, false)));
+    removeComments(testData.data);
```
gumb0: A bit weird that `readTestFile` returns hash of json differing from json that it returns, maybe move removing comments to the calling side where it's only needed (to `TestSuite::executeTest`)

gumb0:
1. Can we limit what goes into this PR? It is now already quite big.
2. Please provide more meaningful commit messages than "testeth fixes"
3. What's happening with all Int/U256 conversion?

gumb0:
Is it possible to make a separate PR out of all Int/U256 changes? It's not related to hash calculation I guess?

winsvega:
this will be rather a big PR because I have to make all tests green. 
so apparantly some tests expected negative timestamp so I had to implement Int64 version of PositiveInt64 function.  then I noticed that toInt is actually convert data into u256  which is not exactly Int

gumb0:
Can't we make separate PR about Int64, merge it, then rebase this one?

winsvega:
Would not it override the commit hash in this PR?

gumb0:
It will, but you'd need to regenerate all tests with final version of this PR anyway, so that doesn't matter anymore?

libethcore/BlockHeader.cpp
```diff
@@ -204,6 +204,7 @@ void BlockHeader::populateFromParent(BlockHeader const& _parent)
     m_parentHash = _parent.m_hash;
     m_gasLimit = _parent.m_gasLimit;
     m_difficulty = _parent.m_difficulty;
+    m_timestamp = _parent.m_timestamp + 1;
```
gumb0: Why is this needed?

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -650,7 +651,8 @@ void overwriteBlockHeaderForTest(mObject const& _blObj, TestBlock& _block, Chain
             ho.count("number") ? toU256(ho["number"]) : header.number(),
             ho.count("gasLimit") ? toU256(ho["gasLimit"]) : header.gasLimit(),
             ho.count("gasUsed") ? toU256(ho["gasUsed"]) : header.gasUsed(),
-            ho.count("timestamp") ? toU256(ho["timestamp"]) : header.timestamp(),
+            ho.count("timestamp") ? max<u256>(toU256(ho["timestamp"]), 0) :
```
gumb0: `u256` is unsigned, so I don't see the point of `max<u256>(toU256(ho["timestamp"]), 0)`

libethcore/BlockHeader.cpp
gumb0: I think you should put `header.setTimestamp()` where you construct the block, instead of here.

This change could affect basically everything in the client. And looking at the code it seems that after this all our PoW mining will produce timestamps incremented from parent instead of real timestamps.

Also - better do this in a separate PR, this is not related to the test hashes.

winsvega:
ok. makes sense move it to separate PR. 
it does not affect mining tests because when mining is used setTimestamp is called.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -286,10 +285,30 @@ int64_t toPositiveInt64(const json_spirit::mValue& _v)
...
+uint64_t toPositiveInt64(const json_spirit::mValue& _v)
+{
+    uint64_t n = 0;
+    switch (_v.type())
+    {
+    case json_spirit::str_type:
+    {
+        long long readval = std::stoll(_v.get_str(), nullptr, 0);
+        if (readval < 0)
+            throw std::out_of_range{
```
gumb0: I recommend not to throw standard exceptions, but the ones derived from them, we have `DEV_SIMPLE_EXCEPTION` macro to define them.

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -286,10 +285,30 @@ int64_t toPositiveInt64(const json_spirit::mValue& _v)
...
+                "TestOutputHelper::toPositiveInt64: unexpected negative value: " +
+                std::to_string(readval)};
+        n = readval;
+    }
+    break;
```
gumb0: indentation is off, use clang-format

test/tools/libtesteth/TestHelper.h
```diff
@@ -84,9 +84,12 @@ std::set<eth::Network> const& getNetworks();
 std::set<std::string> translateNetworks(std::set<std::string> const& _networks);
 u256 toU256(json_spirit::mValue const& _v);
 
-/// Parses a JSON value as an 64-bit signed integer.
+/// Parses a JSON value as an 64-bit unsigned integer.
 /// Throws std::out_of_range exception in case the value is too big or negative.
-int64_t toPositiveInt64(const json_spirit::mValue& _v);
+uint64_t toPositiveInt64(const json_spirit::mValue& _v);
```
gumb0: you could name it `toUint64` then, it'd better correspond to what it does

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -55,6 +61,23 @@ void removeComments(json_spirit::mValue& _obj)
 	}
 }
 
+TestFileData readTestFile(fs::path _testFileName)
```
gumb0: I recommend passing `boost::filesystem::path` by const reference, it's what boost itself does

test/unittests/libethereum/BlockChain.cpp
```diff
@@ -209,7 +209,8 @@ BOOST_AUTO_TEST_CASE(Mining_5_BlockFutureTime)
     uncleBlock.mine(bc);
 
     BlockHeader uncleHeader = uncleBlock.blockHeader();
-    uncleHeader.setTimestamp(uncleHeader.timestamp() + 10000);
+    uncleHeader.setTimestamp(
+        dev::utcTime() + 1000000);  // mining does not reset timestamp on NoProof
```
gumb0: I don't understand this comment and what this change achieves.

test/unittests/libethereum/GasPricer.cpp
```diff
@@ -119,8 +119,8 @@ BOOST_AUTO_TEST_CASE(basicGasPricer_bcUncleTest_Frontier)
 
 BOOST_AUTO_TEST_CASE(basicGasPricer_bcUncleTest_Homestead)
 {
-	u256 _expectedAsk = 155633980282;
-	u256 _expectedBid = 1;
+    u256 _expectedAsk = 155632494086;
```
gumb0: Why did these change?

gumb0: If it's because these tests use NoProof now, I'm not sure, why does it affect gas pricing?

test/unittests/libethereum/BlockChain.cpp
winsvega: it makes uncleHeader timestamp in the future. 
with NoProof setTimestamp were not called. so this appeared to be in the past

test/unittests/libethereum/GasPricer.cpp
winsvega: I don't know.

test/unittests/libethereum/BlockChain.cpp
gumb0: So should your `TestBlock::mine` or `TestBlock` constructor set the correct timestamp maybe?

winsvega: setting timestamp in the future is logic of this test case

winsvega:
need to rebase on this PR https://github.com/ethereum/aleth/pull/5299
after I fix the failing test

test/unittests/libethereum/BlockChain.cpp
gumb0: Ok, but `uncleHeader` should have the valid timestamp I think and then this change won't be needed

test/unittests/libethereum/GasPricer.cpp
gumb0: These tests seem to check something that is not used anywhere, so it's fine for now, we will probably delete them later.

But I think this change was needed because NoProof didn't calculate block gas limit, so you might need to roll this back after rebase on #5299

test/unittests/libethereum/BlockChain.cpp
winsvega: need to investigate why this is failing. cause in calculateDifficulty PR it was fine

test/tools/libtesteth/TestHelper.cpp
winsvega: clang-format is used.

libdevcore/RLP.h
```diff
@@ -277,9 +277,9 @@ class RLP
 		return fromBigEndian<_T>(p);
 	}
 
-	int64_t toPositiveInt64(int _flags = Strict) const
-	{
-		int64_t i = toInt<int64_t>(_flags);
+    int64_t toUint64(int _flags = Strict) const
```
gumb0: Don't change this name please, it was ok and it doesn't return uint64.

Changing the name was ok in the testeth code where you changed the implementation...

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -81,22 +104,25 @@ void addClientInfo(json_spirit::mValue& _v, fs::path const& _testSource, h256 co
 
 void checkFillerHash(fs::path const& _compiledTest, fs::path const& _sourceTest)
 {
-	json_spirit::mValue v;
-	string const s = asString(dev::contents(_compiledTest));
+    json_spirit::mValue filledTest;
+    string s = dev::contentsString(_compiledTest);
```
gumb0: Why is this not const anymore?

libethereum/SnapshotImporter.cpp
```diff
@@ -227,4 +227,4 @@ void SnapshotImporter::importBlockChunks(SnapshotStorageFace const& _snapshotSto
 }
 
 }  // namespace eth
-}  // namespace dev
\ No newline at end of file
+}  // namespace dev
```
gumb0: no changes here, please don't touch this file

test/tools/libtesteth/TestHelper.h
```diff
@@ -84,9 +85,12 @@ std::set<eth::Network> const& getNetworks();
 std::set<std::string> translateNetworks(std::set<std::string> const& _networks);
 u256 toU256(json_spirit::mValue const& _v);
 
-/// Parses a JSON value as an 64-bit signed integer.
+/// Parses a JSON value as an 64-bit unsigned integer.
 /// Throws std::out_of_range exception in case the value is too big or negative.
```
gumb0: It throws `UnexpectedNegative` now

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -55,6 +61,23 @@ void removeComments(json_spirit::mValue& _obj)
...
+    BOOST_REQUIRE_MESSAGE(s.length() > 0, "Contents of " + _testFileName.string() + " is empty.");
+
+    if (_testFileName.extension() == ".json")
+        json_spirit::read_string(s, testData.data);
+    else if (_testFileName.extension() == ".yml")
+        testData.data = test::parseYamlToJson(s);
+    else
+        BOOST_ERROR("Unknow test format!" + test::TestOutputHelper::get().testFile().string());
+
+    testData.hash = h256(sha3(json_spirit::write_string(testData.data, false)));
```
gumb0: I think `dev::sha3` returns `h256`, no need for `h256(...)` ?

PR #5158, opened 2018-08-02
Update README.md

melvinljy96:
Adding more descriptions and and check for grammatical error. The reason to do that is to make the README.md file more user-friendly so that user can learn and understand more quickly.


melvinljy96:
Good day sir, may I know when will it be merged?
Your kind consideration is highly appreciated.
Thank you for your attention.

PR #5170, opened 2018-08-08
JSON-RPC Proxy improvements

chfast:


scripts/jsonrpcproxy.py
```diff
@@ -111,6 +129,9 @@ def __init__(self, ipc_path):
     def is_connected(self):
         return True
 
+    def check_connection(self, timeout):
+        pass
```
axic: Perhaps add a todo comment here?

scripts/jsonrpcproxy.py
```diff
@@ -82,6 +84,22 @@ def close(self):
     def is_connected(self):
         return self._socket is not None
 
+    def check_connection(self, timeout):
+        SLEEPTIME = 0.1
```
axic: Is this in seconds?

scripts/jsonrpcproxy.py
chfast: I'm not sure it should be on any todo list. Looks like for named pipes on Windows it is always "connected".

scripts/jsonrpcproxy.py
chfast: Yes.

scripts/jsonrpcproxy.py
```diff
@@ -221,6 +238,14 @@ def process(self, request):
 
         return response
 
+    def run(self):
+        self.conn = get_ipc_connector(self.backend_address)
+        self.conn.check_connection(timeout=10.0)
```
axic: There is a `BUFSIZE=32` in the top. Maybe move this as a configuration value up there too?

axic: Also suggest to bump the `VERSION` there.

chfast: Bumped to 0.2

PR #5171, opened 2018-08-08
Use storage status from EVMC

chfast:
Depends on https://github.com/ethereum/evmc/pull/52.

### DIFF
--- a/libaleth-interpreter/VM.cpp
+++ b/libaleth-interpreter/VM.cpp
@@ -189,15 +189,6 @@ void VM::adjustStack(int _removed, int _added)
         throwBadStack(_removed, _added);
 }
 
-void VM::updateSSGas()
-{
-    evmc_uint256be key = toEvmC(m_SP[0]);
-    evmc_uint256be rawValue;
-    m_context->fn_table->get_storage(&rawValue, m_context, &m_message->destination, &key);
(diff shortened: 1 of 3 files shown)


libaleth-interpreter/VM.cpp
```diff
@@ -1365,12 +1356,13 @@ void VM::interpretCases()
...
-            updateIOGas();
-
             evmc_uint256be key = toEvmC(m_SP[0]);
             evmc_uint256be value = toEvmC(m_SP[1]);
-            m_context->fn_table->set_storage(m_context, &m_message->destination, &key, &value);
+            auto status =
+                m_context->fn_table->set_storage(m_context, &m_message->destination, &key, &value);
+            m_runGas = (status == EVMC_STORAGE_ADDED) ? VMSchedule::sstoreSetGas :
+                                                        VMSchedule::sstoreResetGas;
+            updateIOGas();  // Update the gas after the change when the full cost is known.
```
chfast: I see one problem here. When we have 0 gas left, we will do a contract storage lookup (which should cost at least 200). Maybe we should charge `sstoreResetGas` up front, and then optionally `VMSchedule::sstoreSetGas - VMSchedule::sstoreResetGas`?

BTW, wasn't the cost of deletion 0 in Frontier?

PR #5173, opened 2018-08-08
Fix debug_traceTransaction

gumb0:
- Rename `StandardTrace::json()` to `multilineTrace()`, because it's not a valid json now, but json objects for each opcode trace concantenated as seaparate line each.
- Move styled json option into separate `styledJson()` method.
- Add `jsonValue()` method to return json without converting to string.

### DIFF
--- a/libethereum/Executive.h
+++ b/libethereum/Executive.h
@@ -61,8 +61,9 @@ public:
     void setShowMnemonics() { m_showMnemonics = true; }
     void setOptions(DebugOptions _options) { m_options = _options; }
 
-    std::string json(bool _styled = false) const;
     Json::Value jsonValue() const { return m_trace; }
+    std::string styledJson() const;
+    std::string multilineTrace() const;
 
     OnOpFunc onOp()
     {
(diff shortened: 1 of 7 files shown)


libethereum/Executive.h
```diff
@@ -61,7 +61,9 @@ class StandardTrace
     void setShowMnemonics() { m_showMnemonics = true; }
     void setOptions(DebugOptions _options) { m_options = _options; }
 
-    std::string json(bool _styled = false) const;
+    Json::Value jsonValue() const { return m_trace; }
```
chfast: I would shorten it to `.json()` but I understand that refactoring was easier this way.

gumb0: Yeah I considered it, then decided `jsonValue` is more explicit here, to show that it's not a string

chfast: :+1:

gumb0:
docker-aleth on circleci now takes longer than AppVeyor...

chfast:
Yes. Sorry for that. It should not be running every build. I was about to fix that.

PR #5174, opened 2018-08-09
Rename MinerCLI variable to miner

tiendq:
`m` is changed to `miner` for reading code easier, especially when there're many line of code between its declaration and where it's used. And there're already many local scope `m` variables for other purposes.


chfast:
All PR should target master branch from now on (I changed the base here, no action is needed).

PR #5176, opened 2018-08-10
Upgrade EVMC to v5.0.0

chfast:


axic:
Should also upgrade Hera/evmjit to EVMC5 and then upgrade the Hera/evmjit submodule here, otherwise someone trying to do a build with those from the aleth release source will fail.

chfast:
I will upgrade Hera. EVMJIT probably does not build for some time with Constantinople changes introduces recently. I will leave it for other time.

PR #5178, opened 2018-08-10
Fix incorrect logging of NodeIPEndpoint

gumb0:
Fix is to move `operator<<` for `NodeIPEndpoint` into namespace `dev::p2p`

### DIFF
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -98,11 +98,15 @@ shared_ptr<NodeEntry> NodeTable::addNode(Node const& _node, NodeRelation _relati
     if (!_node.id)
     {
         DEV_GUARDED(x_nodes)
-        LOG(m_logger) << "Sending public key discovery Ping to "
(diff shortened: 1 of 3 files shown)


libp2p/NodeTable.cpp
```diff
@@ -98,11 +98,15 @@ shared_ptr<NodeEntry> NodeTable::addNode(Node const& _node, NodeRelation _relati
     if (!_node.id)
     {
         DEV_GUARDED(x_nodes)
-        LOG(m_logger) << "Sending public key discovery Ping to "
-                      << (bi::udp::endpoint)_node.endpoint
-                      << " (Advertising: " << (bi::udp::endpoint)m_node.endpoint << ")";
+        {
```
gumb0: Tried here adding braces to coerce clang-format to better handle code after `DEV_GUARDED` and it worked almost as needed. The only problem that it made lines 119-122 into one-liner `DEV_GUARDED(x_nodes) {m_nodes[_node.id] = ret; }`

PR #5179, opened 2018-08-11
remove EthashClient

winsvega:
Remove EthashClient.cpp
Move EthashClient methods to Ethash.cpp
relates to https://github.com/ethereum/aleth/pull/5114


winsvega:
@gumb0  
```std::tuple<h256, h256, h256> Ethash::getEthashWork(Client const& _client)```

has both Client and Ethash logic. What is the best way to resolve this?

File: libethashseal/Ethash.cpp [hunk 1]
```diff
@@ -247,3 +247,82 @@ bool Ethash::shouldSeal(Interface*)
...
+}
+
+Ethash& asEthash(SealEngineFace* _f)
+{
+    if (dynamic_cast<Ethash*>(_f))
+        return &dynamic_cast<Ethash*>(_f);
+    throw InvalidSealEngine();
+}
+
+/*EthashClient::~EthashClient()
```
gumb0: I think it's fine to remove this, this code is needed only in descendants of `Worker` class

libethashseal/Ethash.h
```diff
@@ -82,6 +83,34 @@ class Ethash: public SealEngineBase
...
+
+    /// Check the progress of the mining.
+    WorkingProgress miningProgress() const;
+
+    /// Update to the latest transactions and get hash of the current block to be mined minus the
+    /// nonce (the 'work hash') and the difficulty to be met.
+    /// @returns Tuple of hash without seal, seed hash, target boundary.
+    std::tuple<h256, h256, h256> getEthashWork(Client const& _client);
+
+    /** @brief Submit the proof for the proof-of-work.
```
gumb0: Please change this to use `///`

libethashseal/Ethash.h
```diff
@@ -92,7 +121,16 @@ class Ethash: public SealEngineBase
...
+    u256 externalHashrate() const;
+
+    // external hashrate
+    mutable std::unordered_map<h256, std::pair<u256, std::chrono::steady_clock::time_point>>
+        m_externalRates;
+    mutable SharedMutex x_externalRates;
 };
 
+DEV_SIMPLE_EXCEPTION(InvalidSealEngine);
+Ethash& asEthash(SealEngineFace* _f);
```
gumb0: Maybe change this to get reference parameter, it's more logical to pass reference and transform it to reference,

libethashseal/Ethash.cpp
```diff
@@ -247,3 +247,82 @@ bool Ethash::shouldSeal(Interface*)
...
+    WriteGuard writeGuard(x_externalRates);
+    for (auto i = m_externalRates.begin(); i != m_externalRates.end();)
+        if (chrono::steady_clock::now() - i->second.second > chrono::seconds(5))
+            i = m_externalRates.erase(i);
+        else
+            ret += i++->second.first;
+    return ret;
+}
+
+Ethash& asEthash(SealEngineFace* _f)
```
gumb0: Shouldn't this be defined in namespace `eth`?

libethashseal/Ethash.h
```diff
@@ -64,7 +65,7 @@ class Ethash: public SealEngineBase
     void generateSeal(BlockHeader const& _bi) override;
     bool shouldSeal(Interface* _i) override;
 
-    eth::GenericFarm<EthashProofOfWork>& farm() { return m_farm; }
+    eth::GenericFarm<EthashProofOfWork> const& farm() const { return m_farm; }
```
gumb0: Can we remove this and just use `m_farm` in internal methods?

libethashseal/Ethash.h
```diff
@@ -82,6 +83,34 @@ class Ethash: public SealEngineBase
...
+    /** @brief Submit the proof for the proof-of-work.
+     * @param _s A valid solution.
+     * @return true if the solution was indeed valid and accepted.
+     */
+    bool submitEthashWork(h256 const& _mixHash, h64 const& _nonce);
+
+    void submitExternalHashrate(u256 const& _rate, h256 const& _id);
+
+    /// @returns true only if it's worth bothering to prep the mining block.
+    bool shouldServeWork(Client const& _c) const
```
gumb0: Can we make this private?

libweb3jsonrpc/Eth.cpp
```diff
@@ -613,8 +613,8 @@ Json::Value Eth::eth_getWork()
 	try
 	{
 		Json::Value ret(Json::arrayValue);
-		auto r = asEthashClient(client())->getEthashWork();
-		ret.append(toJS(get<0>(r)));
+        auto r = asEthash(client()->sealEngine())->getEthashWork();
```
gumb0: Doesn't `getEthashWork` require now the `Client` argument?

libweb3jsonrpc/Eth.cpp
```diff
@@ -58,9 +58,9 @@ string Eth::eth_coinbase()
 string Eth::eth_hashrate()
 {
 	try
-	{
-		return toJS(asEthashClient(client())->hashrate());
-	}
+    {
+        return toJS(asEthash(client()->sealEngine())->hashrate());
```
gumb0: Would be great to create a function out of repeating code
```
try
{
  return asEthash(client()->sealEngine());
}
catch(InvalidSealEngine&)
{
 ...
}
```

gumb0:
Ah the tests still don't compile. But I think this is the right direction.

libethashseal/Ethash.cpp
winsvega: the previous one was not.

winsvega:
fixed your comments. 
getEthashWork logically belongs to Ethash
but it has private methods and members of the client. 
as well as signals. so I wonder should I make a public method in Client to resolve this?

libethashseal/Ethash.cpp
gumb0: Waht do you mean I see this
```
EthashClient& dev::eth::asEthashClient(Interface& _c)
```
I think this will lead to linker error without `dev::eth::`
Better approach: don't use `using namespace dev;` etc. Instead put everything into `namespace dev { ... }`

libethashseal/Ethash.cpp
```diff
@@ -247,3 +247,76 @@ bool Ethash::shouldSeal(Interface*)
...
+    return r;
+}
+
+std::tuple<h256, h256, h256> Ethash::getEthashWork(Client const& _client)
+{
+    // lock the work so a later submission isn't invalidated by processing a transaction elsewhere.
+    // this will be reset as soon as a new block arrives, allowing more transactions to be
+    // processed.
+    bool oldShould = shouldServeWork(_client);
+    m_lastGetWork = chrono::system_clock::now();
```
gumb0: 1. Add `getWork` to `SealEngineFace`, rename `getEthashWork` to `getWork` (and you'll need some dummy implementation in `NoProof`)
2. Then create `Client::getWork` that will update `m_lastGetWork` like here and then call `sealEngine()->getWork()`
3. Call `Client::getWork` from `eth_getWork`

libethashseal/Ethash.cpp
```diff
@@ -247,3 +247,76 @@ bool Ethash::shouldSeal(Interface*)
...
+
+    if (!sealEngine()->shouldSeal(this))
```
gumb0: Move this also to `Client::getWork`

libethashseal/Ethash.cpp
```diff
@@ -247,3 +247,76 @@ bool Ethash::shouldSeal(Interface*)
...
+        return std::tuple<h256, h256, h256>();
+
+    // if this request has made us bother to serve work, prep it now.
+    if (!oldShould && shouldServeWork(_client))
```
gumb0: Move this also to `Client::getWork`

libethashseal/Ethash.cpp
```diff
@@ -247,3 +247,76 @@ bool Ethash::shouldSeal(Interface*)
...
+        onPostStateChanged();
+    else
+        // otherwise, set this to true so that it gets prepped next time.
+        m_remoteWorking = true;
+    manuallySetWork(m_sealingInfo);
```
gumb0: This will be start of `Ethash::getWork`. Pass BlockHeader (m_sealingInfo) as a parameter to `SealEngineFace::getWork`

libethashseal/Ethash.cpp
winsvega: 1. -> 
sealEngineFace could not include Client. I could not declare 
```std::tuple<h256, h256, h256> SealEngineFace::getWork(Client const& _client)```

libethashseal/Ethash.cpp
winsvega: this depends on Ethash::isMining()

PR #5188, opened 2018-08-14
 rpc: Return empty bytes as "" not as "0x" 

chfast:


winsvega:
this might break the tests. check that when filling it is creating empty data and empty code as ```0x```

```
testeth -t GeneralStateTests/stExample -- --filltests
```

Filled test in
```
TestPath / GeneralStateTests / stExample / add1.json
```

chfast:
> check that when filling it is creating empty data and empty code as 0x

No. Please create a unit test for it if you care about that. Otherwise the problem is non-existing for me.

winsvega:
added more checks to the unit test around test filling logic.

chfast:
So is this a problem for test filling or not?

winsvega:
no. the filling works.

chfast:
So can you review this PR then?

axic:
> Return empty bytes as "" not as "0x"

Is this the "standard" now? I remember a lot of arguments for these.

chfast:
> > Return empty bytes as "" not as "0x"
> 
> Is this the "standard" now? I remember a lot of arguments for these.

Probably no standards in this field, but from JavaScript point of view this is the best option because empty bytes is always `""` no matter if raw or hex encoded.

axic:
As this changes every single RPC method, does this matches geth entirely? If so, it is better than now.

chfast:
Ok, I'm confused now, because geth returns `{"jsonrpc":"2.0","id":67,"result":"0x"}`.

test/unittests/libweb3jsonrpc/jsonrpc.cpp
```diff
@@ -558,6 +558,8 @@ BOOST_AUTO_TEST_CASE(contract_storage)
         "505600a165627a7a72305820d8407d9cdaaf82966f3fa7a3e665b8cf4e6"
         "5ee8909b83094a3f856b9051274500029";
 
+    const string runtimeCode = compiled.substr(58);
```
gumb0: Is there an easy way to find this offset? or just disassemble and look for RETURN

chfast:
I'm really confused now, because the `0x` is consistent with geth, but @lrettig confirmed that this change fixes the issue in truffle.

Maybe there is some other method that has been changed with this PR. @lrettig, can you please investigate exactly what method and result causes problems in truffle?

chfast:
I expect the problem is with some RPC method returning an integer, not bytes.

winsvega:
after this PR I could provide you with retesteth that would run RPC methods checks. 
https://github.com/ethereum/aleth/pull/5179

lrettig:
@chfast it appears to be `eth_call`. Full transcript [here](https://gist.github.com/lrettig/89d778cbd51c35b1ad28809bdef6d1bd).

chfast:
> after this PR I could provide you with retesteth that would run RPC methods checks.
> #5179

We have JSON-RPC compatibility tests for 2 years done by @cdetrio, but nobody seems to care. Please do not create your own test set. However, I would be great if you take it over and start testing clients regularly.

chfast:
@lrettig big thanks for the log. If you have more time, can you check what geth is returning?

chfast:
@lrettig Would you be so kind to get a logs from truffle with the Aleth patched with this PR changes?

lrettig:
@chfast here you go: https://gist.github.com/lrettig/e9b83d2f0c8575578b16a1488afe7cb0

PR #5190, opened 2018-08-14
Fix use-after-free in AccountHolder::defaultTransactAccount() 

chfast:


winsvega:
does this also fix the data race around AccountHolder ?

chfast:
> does this also fix the data race around AccountHolder ?

No.

PR #5193, opened 2018-08-15
Remove duplicated Ethash::init and NoProof::init calls

### LINKED ISSUE #5187: Duplicated call to Ethash::init() and NoProof::init()
These functions are called first in `aleth/main.cpp` `main()` and then called again in `MinerCLI::MinerCLI`. They register seal engines and I think they do no harm, just duplicate work. Which one should be removed?

tiendq:
close #5187

### DIFF
--- a/aleth/MinerAux.h
+++ b/aleth/MinerAux.h
@@ -54,7 +54,6 @@ public:
 
     explicit MinerCLI(OperationMode _mode = OperationMode::None): mode(_mode)
     {
-        BasicAuthority::init();
     }
 
     bool interpretOption(size_t& i, vector<string> const& argv)
--- a/aleth/MinerAux.h
+++ b/aleth/MinerAux.h
@@ -52,11 +52,8 @@ public:
         Benchmark
     };
 
-
     explicit MinerCLI(OperationMode _mode = OperationMode::None): mode(_mode)
     {
-        Ethash::init();
-        NoProof::init();
         BasicAuthority::init();
     }
 
--- a/libwebthree/WebThree.cpp
+++ b/libwebthree/WebThree.cpp
@@ -44,10 +44,9 @@ WebThreeDirect::WebThreeDirect(std::string const& _clientVersion,
 {
     if (_dbPath.size())
         Defaults::setDBPath(_dbPath);
+
     if (_interfaces.count("eth"))
     {
-        Ethash::init();
-        NoProof::init();
         if (_testing)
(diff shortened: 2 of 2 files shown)


aleth/MinerAux.h
```diff
@@ -52,11 +52,8 @@ class MinerCLI
         Benchmark
     };
 
-
     explicit MinerCLI(OperationMode _mode = OperationMode::None): mode(_mode)
     {
-        Ethash::init();
-        NoProof::init();
         BasicAuthority::init();
```
chfast: We probably should also drop `BasicAuthority::init()`.

tiendq: Ok, I'll remove it too. By the way, I just found these `init` in `WebThreeDirect::WebThreeDirect` too. `WebThreeDirect` is initialized in `main` at line 854 so I'm pretty sure those init are redundant too, but I need your confirmation.

tiendq: [The inits in `WebThreeDirect` ctor](https://github.com/ethereum/aleth/blob/master/libwebthree/WebThree.cpp#L49)

chfast: @gumb0 can you look here. Should WebTreeDirect init seal engines? Do we need to init BasicAuthority engine? It seems to be abandoned.

gumb0: Some unit tests create `WebThreeDirect` instances, but it seems testeth code calls these `init`s anyway for each test. So I think we can remove it from `WebThreeDirect` if tests don't break

gumb0: Seems that there are some unit tests for `BlockChain` that use `BasicAuthority` (`BlockChainInsertTests`)

@winsvega Do you know if `BasicAuthority` is useful for anything?

winsvega: Without this init. Windows was having build errors

winsvega: My theory is that visual studio does not build cpp file if nothing from this file is called.

tiendq: @gumb0 I have removed those inits, from `WebThreeDirect` too. No failed unit tests.

@chfast meant to drop `BasicAuthority::init()` from `MinerCLI` ctor, not the `BasicAuthority` itself. I just checked and see `BasicAuthority` is used in [BlockChainInsert.cpp](https://github.com/ethereum/aleth/blob/master/test/unittests/libethereum/BlockChainInsert.cpp) test.

PR #5198, opened 2018-08-16
Cable

chfast:


CMakeLists.txt
```diff
@@ -4,19 +4,13 @@
...
 include(GNUInstallDirs)
 
-if(UNIX)
-    # Build deps with PIC to allow building shared libraries.
-    set(toolchain cxx11-pic)
-else()
-    set(toolchain cxx11)
-endif()
-cable_configure_toolchain(DEFAULT ${toolchain})
+cable_configure_toolchain(DEFAULT cxx11-pic)
```
axic: `fpic`?

chfast: No, I'm testing the simpler one because Hunter said it might work.

chfast: Seems to be working nicely, so `fpic` variant is not needed.

PR #5200, opened 2018-08-17
Don't try to connect to oneself

gumb0:
Fixes https://github.com/ethereum/aleth/issues/5199

### DIFF
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -527,6 +527,12 @@ void Host::addNode(NodeID const& _node, NodeIPEndpoint const& _endpoint)
         else
             return;
 
+    if (_node == id())
+    {
+        cnetdetails << "Ingoring the request to connect to host " << _node;
+        return;
+    }
+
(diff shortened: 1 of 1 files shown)


libp2p/Host.cpp
```diff
@@ -592,6 +592,9 @@ void Host::connect(std::shared_ptr<Peer> const& _p)
     if (!m_run)
         return;
     
+    if (_p->id == id())
+        return;
```
chfast: Add comment, and maybe a log?

gumb0:
@chfast I moved the check up the stack, so that we don't even add our own nodeID to NodeTable; and added a log.

libp2p/Host.cpp
```diff
@@ -527,6 +527,12 @@ void Host::addNode(NodeID const& _node, NodeIPEndpoint const& _endpoint)
         else
             return;
 
+    if (_node == id())
+    {
+        cnetdetails << "Ingoring the request to connect to host " << _node;
```
axic: host -> self ?

gumb0: changed

PR #5211, opened 2018-08-21
Shows error if config file not found or empty

### LINKED ISSUE #5205: aleth should complain when --config argument file is not found
When aleth is passed a file that doesn't exist in the `--config` param, it defaults to a mainnet config and starts syncing to the mainnet. Instead, aleth should complain.

tiendq:
Fix #5205

### DIFF
--- a/aleth/main.cpp
+++ b/aleth/main.cpp
@@ -560,6 +560,12 @@ int main(int argc, char** argv)
         {
             configPath = vm["config"].as<string>();
             configJSON = contentsString(configPath.string());
+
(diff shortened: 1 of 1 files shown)


gumb0:
Could you squash the commits?

tiendq:
@gumb0 I accidentally did several wrong merges which resulted in "merge" commits above. Just cleaned up my repo and made new PR (#5218). Please review the new one (exactly same this one), and ignore this PR. Thanks.

PR #5213, opened 2018-08-22
libp2p - more Capability refactoring

gumb0:
- make `Capability` inherit virtual interface `PeerCapabilityFace`
- rename to `PeerCapability` for less confusion


libethereum/EthereumHost.cpp
```diff
@@ -626,7 +626,8 @@ void EthereumHost::onTransactionImported(ImportResult _ir, h256 const& _h, h512
     }
 }
 
-shared_ptr<Capability> EthereumHost::newPeerCapability(shared_ptr<SessionFace> const& _s, unsigned _idOffset, p2p::CapDesc const& _cap)
+shared_ptr<PeerCapabilityFace> EthereumHost::newPeerCapability(
+    shared_ptr<SessionFace> const& _s, unsigned _idOffset, p2p::CapDesc const& _cap)
 {
     auto ret = HostCapability<EthereumPeer>::newPeerCapability(_s, _idOffset, _cap);
```
chfast: Why is this created first?

libp2p/Session.cpp
```diff
@@ -431,7 +431,7 @@ bool Session::checkRead(std::size_t _expected, boost::system::error_code _ec, st
     return true;
 }
 
-void Session::registerCapability(CapDesc const& _desc, std::shared_ptr<Capability> _p)
+void Session::registerCapability(CapDesc const& _desc, std::shared_ptr<PeerCapabilityFace> _p)
 {
     DEV_GUARDED(x_framing)
     {
```
chfast: Move the shared ptr.

libp2p/Session.h
```diff
@@ -87,99 +88,103 @@ class SessionFace
...
+    PeerSessionInfo info() const override { Guard l(x_info); return m_info; }
+    std::chrono::steady_clock::time_point connectionTime() override { return m_connect; }
 
-	void registerCapability(CapDesc const& _desc, std::shared_ptr<Capability> _p) override;
+    void registerCapability(CapDesc const& _desc, std::shared_ptr<PeerCapabilityFace> _p) override;
 
-	std::map<CapDesc, std::shared_ptr<Capability>> const& capabilities() const override { return m_capabilities; }
+    std::map<CapDesc, std::shared_ptr<PeerCapabilityFace>> const& capabilities() const override
+    {
+        return m_capabilities;
```
chfast: Why capabilities are shared? Cannot the Session be the only owner of them?

gumb0: Good question, I think it can, but then we don't have a guarantee that someone won't store reference to capability longer than `Session` lives.

Solution could be to remove capabilities getter and provide only the method to apply callback to the capabilities.

libethereum/EthereumHost.cpp
gumb0: This is a call to a base class method that creates a capability and stores it in session.
(which is virtual, so returns shared_ptr to a base interface of capabilities `PeerCapabilityFace`)

libp2p/Session.cpp
gumb0: Fixed

PR #5217, opened 2018-08-23
Improve log messages on fatal errors from EVM

gumb0:
See https://github.com/ethereum/aleth/issues/5212

### DIFF
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -497,15 +497,16 @@ bool Executive::go(OnOpFunc const& _onOp)
         }
         catch (InternalVMError const& _e)
         {
-            cwarn << "Internal VM Error (" << *boost::get_error_info<errinfo_evmcStatusCode>(_e) << ")\n"
-                  << diagnostic_information(_e);
+            cerror << "Internal VM Error (EVMC status code: "
+                 << *boost::get_error_info<errinfo_evmcStatusCode>(_e) << ")";
             revert();
             throw;
         }
         catch (Exception const& _e)
         {
             // TODO: AUDIT: check that this can never reasonably happen. Consider what to do if it does.
-            cwarn << "Unexpected exception in VM. There may be a bug in this implementation. " << diagnostic_information(_e);
+            cerror << "Unexpected exception in VM. There may be a bug in this implementation. "
+                 << diagnostic_information(_e);
             exit(1);
             // Another solution would be to reject this transaction, but that also
             // has drawbacks. Essentially, the amount of ram has to be increased here.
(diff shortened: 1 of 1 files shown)


chfast:
That's not what I had in mind. Let me try to explain this point by point.

1. The names `cwarn`, `cnote` and others unnecessary clashes with `cerr`. I'd rename them to something different like `logWarn` or `LOG(WARN)`, or `LOG(warning)`.
2. We should also add `logError` (and `logFatal` if needed).
3. Then here I want to move the severity level from warning to error by using `logError` instead of `logWarn`.
4. Sending to standard error output will make it just disappear.
5. Currently the severity level is not shown in logs. I think it should be. It's more important than the channel name. Also we could use colors. Geth has it done quite nicely.

gumb0:
Oh I meant it to be `cerror`, sorry

gumb0:
Not sure about renaming `cwarn` etc. yet, but I plan to add severity info to messages

gumb0:
`LOG(g_errorLogger::get())` is possible now, too, by the way

libethereum/Executive.cpp
```diff
@@ -497,23 +497,24 @@ bool Executive::go(OnOpFunc const& _onOp)
...
             revert();
             throw;
         }
         catch (Exception const& _e)
         {
             // TODO: AUDIT: check that this can never reasonably happen. Consider what to do if it does.
-            cwarn << "Unexpected exception in VM. There may be a bug in this implementation. " << diagnostic_information(_e);
+            cerror << "Unexpected exception in VM. There may be a bug in this implementation. "
+                 << diagnostic_information(_e);
             exit(1);
```
axic: That these two exceptions are doing an abort, just reinforces the comment (https://github.com/ethereum/aleth/issues/5212#issuecomment-414805672) reasoning that `InternalVMError` should do the same.

chfast: These are truly unexpected. Too defensive programming to my taste, I'd let them go to be handled in the main(). But this not the scope of this PR.

axic: Well `InternalVMError` is truly unexpected too, isn't it?

chfast:
If not a problem, can you move it to `release/1.4`?

libethereum/Executive.cpp
chfast: No it is not, because it can come from external DLLs.
In other cases there is no way to throw `Exception` other than the ones handled above. And `std::exception` can only be `std::bad_alloc`.

gumb0:
Moved to `release/1.4`

PR #5218, opened 2018-08-23
Shows error if config file not found or empty (rework)

### LINKED ISSUE #5205: aleth should complain when --config argument file is not found

tiendq:
close #5205

### DIFF
--- a/aleth/main.cpp
+++ b/aleth/main.cpp
@@ -566,6 +566,12 @@ int main(int argc, char** argv)
         {
             configPath = vm["config"].as<string>();
             configJSON = contentsString(configPath.string());
+
+            if (configJSON.empty())
+            {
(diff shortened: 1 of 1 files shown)


aleth/main.cpp
```diff
@@ -566,6 +566,12 @@ int main(int argc, char** argv)
         {
             configPath = vm["config"].as<string>();
             configJSON = contentsString(configPath.string());
+
+            if (configJSON.empty())
+            {
+                cout << "Config file not found or empty (" << configPath.string() << ")\n";
```
gumb0: Output to `cerr` like other messages in this file

tiendq: @gumb0 Amended :)

PR #5219, opened 2018-08-23
fix NoProof block signature

winsvega:
Other devs report that It is better if NoProof blocks still have the signature. but the signature is just ignored. 
Rather then absence of signature as it is now.

### DIFF
--- a/libethcore/BasicAuthority.cpp
+++ b/libethcore/BasicAuthority.cpp
@@ -50,8 +50,11 @@ void BasicAuthority::generateSeal(BlockHeader const& _bi)
 	BlockHeader bi = _bi;
 	h256 h = bi.hash(WithoutSeal);
 	Signature s = sign(m_secret, h);
-	setSig(bi, s);
-	SealEngineBase::generateSeal(bi);
+    setSig(bi, s);
+    RLPStream ret;
+    bi.streamRLP(ret);
+    if (m_onSealGenerated)
+        m_onSealGenerated(ret.out());
 }
 
 bool BasicAuthority::onOptionChanging(std::string const& _name, bytes const& _value)
--- a/libethcore/SealEngine.cpp
+++ b/libethcore/SealEngine.cpp
@@ -29,6 +29,17 @@ void NoProof::init()
     ETH_REGISTER_SEAL_ENGINE(NoProof);
 }
 
+void NoProof::generateSeal(BlockHeader const& _bi)
+{
+    BlockHeader header(_bi);
+    header.setSeal(NonceField, h64{0});
+    header.setSeal(MixHashField, h256{0});
+    RLPStream ret;
+    header.streamRLP(ret);
+    if (m_onSealGenerated)
+        m_onSealGenerated(ret.out());
+}
+
 void SealEngineFace::verify(Strictness _s, BlockHeader const& _bi, BlockHeader const& _parent, bytesConstRef _block) const
 {
     _bi.verify(_s, _parent, _block);
--- a/libethcore/SealEngine.h
+++ b/libethcore/SealEngine.h
@@ -97,14 +97,12 @@ private:
 class SealEngineBase: public SealEngineFace
 {
 public:
-	void generateSeal(BlockHeader const& _bi) override
-	{
-		RLPStream ret;
-		_bi.streamRLP(ret);
-		if (m_onSealGenerated)
-			m_onSealGenerated(ret.out());
-	}
-	void onSealGenerated(std::function<void(bytes const&)> const& _f) override { m_onSealGenerated = _f; }
+    enum
+    {
+        MixHashField = 0,
+        NonceField = 1
+    };
+    void onSealGenerated(std::function<void(bytes const&)> const& _f) override { m_onSealGenerated = _f; }
 	EVMSchedule const& evmSchedule(u256 const& _blockNumber) const override;
 	u256 blockReward(u256 const& _blockNumber) const override;
 
@@ -139,6 +137,7 @@ class NoProof: public eth::SealEngineBase
 public:
     static std::string name() { return "NoProof"; }
     static void init();
+    void generateSeal(BlockHeader const& _bi) override;
 };
 
 }
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -109,7 +109,7 @@ json_spirit::mValue BlockchainTestSuite::doTests(json_spirit::mValue const& _inp
                     set<string> netlist;
                     json_spirit::mObject const& expectObj = expect.get_obj();
                     ImportTest::parseJsonStrValueIntoSet(expectObj.at("network"), netlist);
-
+                    netlist = test::translateNetworks(netlist);
                     if (netlist.count(test::netIdToString(network)) || netlist.count("ALL"))
                     {
                         jObjOutput["expect"] = expectObj.at("result");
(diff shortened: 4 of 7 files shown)


test/tools/libtesteth/BlockChainHelper.cpp
```diff
@@ -273,6 +273,14 @@ void TestBlock::mine(TestBlockChain const& _bc)
     }
     else
         recalcBlockHeaderBytes();
+
+    // Create dummy signature for NoProof blocks
+    if (Ethash::mixHash(m_blockHeader) == h256(0))
```
chfast: `h256{0}`.

test/tools/libtesteth/BlockChainHelper.cpp
```diff
@@ -273,6 +273,14 @@ void TestBlock::mine(TestBlockChain const& _bc)
...
+    {
+        m_blockHeader.setSeal(Ethash::NonceField, h64(42));
```
chfast: Maybe make it null?

test/tools/libtesteth/BlockChainHelper.cpp
```diff
@@ -273,6 +273,14 @@ void TestBlock::mine(TestBlockChain const& _bc)
     }
     else
         recalcBlockHeaderBytes();
+
+    // Create dummy signature for NoProof blocks
+    if (Ethash::mixHash(m_blockHeader) == h256{0})
+    {
+        m_blockHeader.setSeal(Ethash::NonceField, h64{0});
+        m_blockHeader.setSeal(Ethash::MixHashField, h256{0});
```
gumb0: Why do you assign it here if you just checked that it's 0?

winsvega: because it returns 0 also when there is no signature set.

winsvega: We could implenent a method hasSignature() if you woukd like. That will change libethereum. By default I try not to touch libethereum for test related logic

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -109,7 +109,7 @@ json_spirit::mValue BlockchainTestSuite::doTests(json_spirit::mValue const& _inp
                     set<string> netlist;
                     json_spirit::mObject const& expectObj = expect.get_obj();
                     ImportTest::parseJsonStrValueIntoSet(expectObj.at("network"), netlist);
-
+                    netlist = test::translateNetworks(netlist);
```
gumb0: Did you just sneak in this unrelated change? :)

gumb0:
I don't get - if there's a change to `NoProof` blocks format - why only testeth files are changed?
Will `test_mineBlocks` generate correct blocks with `NoProof` config?

gumb0:
Do you mean both blocks witthout any signature and blocks with zero signature are "correct"?
This doesn't make sense to me, only one format should be valid (the one that other clients would also support)

winsvega:
Correct in what context? 
blocks without signature are incorrect for the test format. 
blocks with or without signature could still be invalid or correct depending on a block contents. 
blocks without signature or with invalid signature are incorret on Ethash

winsvega:
There is no consensus on blocks without signature cause NoProof is smth that we use for testing purposes. So we define what a block on NoProof should look like (with zero signature / random signature  or no signature at all) 
It is also possible to support blocks without signature in test format. It just that other clients devs agreed that it is easier for them to parse blocks with at least some signature.

gumb0:
> So we define what a block on NoProof should look like (with zero signature / random signature or no signature at all)

So I understood that intention of this PR is to establish that NoProof blocks have zero signature.

> It just that other clients devs agreed that it is easier for them to parse blocks with at least some signature.

Right, so I'm saying we should generate blocks with signature everywhere.

> blocks with or without signature could still be invalid or correct depending on a block contents.

Why should blocks without signature be valid?

winsvega:
Blocks without signature are valid on NoProof

gumb0:
But you say they are not valid for other clients

winsvega:
Other clients parsing state test format expect signature in blocks. Only in test format. They dont have NoProof defenition. We could make it anything.

gumb0:
We could, but why would we want two different block formats in tests and out of the tests? This will only lead to confusion and possible incompatibilities between parts of the codebase.

chfast:
> Thats why this pr adding a zero signature to the blocks without signature

But you don't change NoProof design, just hack the final test output.

I'm also a bit into the solution to make NoProof more strict and do not allow block headers without signature fields.

winsvega:
By default I try not to touch core libraries logic for testing purposes. 
So If you think that NoProof should be modified on a core level to require signature at the block but not verifying it we could open another PR. 

To summarize the discussion: 
Is it concluded that NoProof should generate 0 signature at corelib mining method? 
Is it concluded that aleth should require signature to be present in a block rlp for any Ethash/NoProof

chfast:
Yes, if that makes the whole system more consistent and simple to understand.

Can someone check if Clique is adding/remove any fields to the block hash?

gumb0:
@chfast It does not change the format, just changes the meaning of some fields https://github.com/ethereum/EIPs/issues/225

winsvega:
updatet this PR. now 0 signature is set by default inside sealEngineBase generateSeal method. 
that is used when NoProof is enabled.

gumb0:
`bcBasicInsert` test is failing

libethcore/SealEngine.h
```diff
@@ -99,8 +99,11 @@ class SealEngineBase: public SealEngineFace
 public:
 	void generateSeal(BlockHeader const& _bi) override
 	{
+        BlockHeader header(_bi);
```
gumb0: I think this should happen in `NoProof`, not in `SealEngineBase`
Also I don't see any reason for `SealEngineBase` to have `generateSeal` implementation at all, so try to just move it to `NoProof`

winsvega:
removed generateSeal from SealEngineBase
moved generateSeal from SealEngineBase to NoProof.
Fixed bcBasicImport by moving the old implementation of generateSeal from SealEngineBase to BasicAuthority generateSeal

There should be an error throwing in case no callback is called from BasicAuthority::generateSeal. 
That leads to endless loop.

gumb0:
Where is the endless loop?

libethcore/SealEngine.cpp
```diff
@@ -29,6 +29,17 @@ void NoProof::init()
     ETH_REGISTER_SEAL_ENGINE(NoProof);
 }
 
+void NoProof::generateSeal(BlockHeader const& _bi)
+{
+    BlockHeader header(_bi);
+    header.setSeal(1, h64{0});   // NonceField
```
gumb0: Maybe move `enum { MixHashField = 0, NonceField = 1 };` to `SealEngineBase` and use it here

winsvega:
Endless loop somewhere where it waits for a callback of m_onSealGenerated

libethcore/SealEngine.cpp
winsvega: this is a good question. what if with sealEngine = Ethash2  there will be 3 or 1 fields for signature?

gumb0: We'll see then, let's not try to guess the future

libethcore/BasicAuthority.cpp
```diff
@@ -50,8 +50,12 @@ void BasicAuthority::generateSeal(BlockHeader const& _bi)
...
 	h256 h = bi.hash(WithoutSeal);
 	Signature s = sign(m_secret, h);
-	setSig(bi, s);
-	SealEngineBase::generateSeal(bi);
+    setSig(bi, s);
+    RLPStream ret;
+    bi.streamRLP(ret);
+    if (m_onSealGenerated)
+        m_onSealGenerated(ret.out());
+    // Else throw an error!!!
```
gumb0: I'd say it's a problem of the calling code if it would not set this handler and then wait somehow (how?) for it to be called.
So I don't think anything need to be done here.

PR #5221, opened 2018-08-26
EVMC: show logs on loading and warning on failing to set option

axic:


axic:
@chfast please advise what is the bet logging method to be used.

libevm/EVMC.cpp
```diff
@@ -15,10 +15,13 @@ EVM::EVM(evmc_instance* _instance) noexcept : m_instance(_instance)
     assert(m_instance != nullptr);
     assert(m_instance->abi_version == EVMC_ABI_VERSION);
 
+    cwarn << "EVMC name=" << m_instance->name << " version=" << m_instance->version << "\n";
```
gumb0: I'd say this should go to `cnote`

libevm/VMFactory.cpp
```diff
@@ -85,6 +85,7 @@ void setVMKind(const std::string& _name)
     }
 
     // If not match for predefined VM names, try loading it as an EVMC DLL.
+    cwarn << "Loading EVMC module: " << _name << "\n";
```
gumb0: This is too, more like `cnote` to me

libevm/EVMC.cpp
```diff
@@ -15,10 +15,13 @@ EVM::EVM(evmc_instance* _instance) noexcept : m_instance(_instance)
...
+
     // Set the options.
     if (m_instance->set_option)
         for (auto& pair : evmcOptions())
-            m_instance->set_option(m_instance, pair.first.c_str(), pair.second.c_str());
+            if (m_instance->set_option(m_instance, pair.first.c_str(), pair.second.c_str()) != 1)
+                cwarn << "Failed to set EVMC parameter '" << pair.first << "'";
```
axic: @gumb0 @chfast should this have a `\n`?

The `cwarn << "Execution rejected by EVMC, executing with default VM implementation";` below doesn't have one, is that a bug?

chfast: No.

axic: So I should just remove the `\n` I'm adding?

chfast: Yes.

axic:
Updated, should be ready.

libevm/EVMC.cpp
```diff
@@ -15,10 +15,13 @@ EVM::EVM(evmc_instance* _instance) noexcept : m_instance(_instance)
     assert(m_instance != nullptr);
     assert(m_instance->abi_version == EVMC_ABI_VERSION);
 
+    cnote << "EVMC name=" << m_instance->name << " version=" << m_instance->version;
```
chfast: i've found an issue. The log cannot be here, because it will be shown for every transaction. You can change it that EVM wrapper is created only once.

axic: That might be too big of a work for me in aleth, will just remove the message. Unless you want to push a fix yourself?

axic: Created an issue to track this https://github.com/ethereum/aleth/issues/5223.

PR #5228, opened 2018-08-28
Add "from" and "to" fields to the output of eth_getTransactionReceipt 

dchenmei:


gumb0:
Would you like to add a unit test for this? 
See `test/unittests/libweb3jsonrpc/jsonrpc.cpp` for other tests for RPC methods, some of them even call `eth_getTransactionReceipt`, it should be not hard to add a similar new test

libweb3jsonrpc/JsonHelper.cpp
```diff
@@ -203,6 +203,8 @@ Json::Value toJson(dev::eth::LocalisedTransactionReceipt const& _t)
     res["transactionIndex"] = _t.transactionIndex();
     res["blockHash"] = toJS(_t.blockHash());
     res["blockNumber"] = _t.blockNumber();
+	res["from"] = toJS(_t.from());
```
tiendq: Indent.

tiendq: Indentation.

dchenmei: Got it! Thanks for the reminder.

dchenmei:
Thanks for the help!

PR #5229, opened 2018-08-29
 evm: Use EVMC VM DLL singleton

chfast:


libevm/VMFactory.cpp
```diff
@@ -175,30 +175,32 @@ po::options_description vmProgramOptions(unsigned _lineLength)
 }
 
 
-std::unique_ptr<VMFace> VMFactory::create()
```
axic: Is this change also needed or is enough to change the loader not to return a `unique_ptr`?

gumb0:
Maybe we should not re-create for all other VMs, too (maybe just create them lazily when first requested) and just return a reference, will that work?

libevm/VMFactory.cpp
```diff
@@ -105,6 +103,8 @@ void setVMKind(const std::string& _name)
                 "loading " + _name + " failed"));
     }
     g_kind = VMKind::DLL;
+
+    cnote << "Loaded EVMC module: " << g_evmcDll->name() << " " << g_evmcDll->version() << " (" << _name << ")";
```
gumb0: I guess this will not be formatted properly not only in testeth but in aleth either, because `setupLogging` is called after handling all command line parameters

gumb0: I don't see any easy fix for it other than just output to `cout`

libevm/VMFactory.cpp
chfast: Yes, because now we need two kinds of pointers, one with default deleter, one with null deleter.

chfast:
> Maybe we should not re-create for all other VMs, too (maybe just create them lazily when first requested) and just return a reference, will that work?

It would work for EVMCs, but not for LegacyVM. For LegacyVM we need a wrapper. But this is my goal for VMFactory. I just wanted minimal set of changes for 1.4 release.

axic:
> For LegacyVM we need a wrapper.

Would merging tracing enable the removal of LegacyVM and thus sparing this work? :)

chfast:
Ping.

axic:
I don't mind. I suppose this can be simplified again when LegacyVM is removed.

libevm/VMFactory.h
```diff
@@ -43,17 +43,19 @@ std::vector<std::pair<std::string, std::string>>& evmcOptions() noexcept;
 boost::program_options::options_description vmProgramOptions(
     unsigned _lineLength = boost::program_options::options_description::m_default_line_length);
 
+using vm_ptr = std::unique_ptr<VMFace, void (*)(VMFace*)>;
```
gumb0: I'd name this `VmPtr`, it's a type, should start with a capital letter

libevm/VMFactory.cpp
```diff
@@ -175,30 +176,34 @@ po::options_description vmProgramOptions(unsigned _lineLength)
...
 {
+    static const auto default_delete = [](VMFace * _vm) noexcept { delete _vm; };
+    static const auto null_delete = [](VMFace*) noexcept {};
+
     switch (_kind)
     {
 #ifdef ETH_EVMJIT
     case VMKind::JIT:
-        return std::unique_ptr<VMFace>(new EVMC{evmjit_create()});
+        return {new EVMC{evmjit_create()}, default_detete};
```
gumb0: `default_delete`

libevm/VMFactory.cpp
```diff
@@ -175,30 +176,34 @@ po::options_description vmProgramOptions(unsigned _lineLength)
...
-std::unique_ptr<VMFace> VMFactory::create()
+vm_ptr VMFactory::create()
 {
     return create(g_kind);
 }
 
-std::unique_ptr<VMFace> VMFactory::create(VMKind _kind)
+vm_ptr VMFactory::create(VMKind _kind)
 {
+    static const auto default_delete = [](VMFace * _vm) noexcept { delete _vm; };
```
gumb0: `static auto const`

libevm/VMFactory.cpp
axic: So the CI doesn't even compile `ETH_EVMJIT` otherwise this would have been caught.

Can evmjit be compiled as a shared lib? If so, any point keeping `ETH_EVMJIT` in aleth anymore?

gumb0:
Why is it bad to create a new instance each time?

axic:
> Why is it bad to create a new instance each time?

There might be a lot of set up time involved.

Can we merge this into 1.4 and include it in the final 1.4 release?

chfast:
@gumb0 @axic Can you approve then?

gumb0:
@chfast at least fix the uncompilable line under `#ifdef ETH_EVMJIT`

libevm/VMFactory.cpp
axic: Tracked in #5253.

chfast:
Done.

axic:
@chfast this seems to be broken, as least in `testeth`. The destructor is not called: https://codecov.io/gh/ewasm/hera/pull/404/changes

PR #5231, opened 2018-08-29
Implement net gas metering for SSTORE in LegacyVM

gumb0:
https://github.com/ethereum/aleth/issues/5226


gumb0:
I will better move tabs conversion to separate PR

gumb0:
Code in Account/State still should be improved and better documented, but I'll add some tests first.

PR #5233, opened 2018-08-29
Upgrade submodules

chfast:


chfast:
Hera update is to fix CMake issue in it.

chfast:
But we should consider removing Hera submodule.

PR #5235, opened 2018-08-30
feat: add support enode:// in peerset option

### LINKED ISSUE #5197: Support enode:// addresses in --peerset command line argument handing
This causes confusion all the time.

Currently the only allowed way to use it is like
```
./aleth --peerset "required:68e6383c649935649723fc44bb4a94bd2019c29a828cb7ee8151abda9567417797fec213f4b60b59d84faac62577814d63cc5d7cc7d4bc09730ea4f43e9f44db@127.0.0.1:30303"
```
We should allow also
```
./aleth --peerset "required:enode://68e6383c649935649723fc44bb4a94bd2019c29a828cb7ee8151abda9567417797fec213f4b60b59d84faac62577814d63cc5d7cc7d4bc09730ea4f43e9f44db@127.0.0.1:30303"
```

See also how parsing is done in `NodeSpec` constructor and maybe reuse this `NodeSpec` class for this.

tiendq:
- Refactored peerset argument handling, now parsing peerset option with regex
- Allow optional `enode://` in peerset

Close #5197


gumb0:
Peerset is essentially a set of node specs. I'd like to break the pattern into pieces, to make enode strings be parsed in one place (in p2p library).

gumb0:
@tiendq do you plan to continue working on this?

tiendq:
@gumb0 Sorry, I'm afraid I can't continue in the near future.

gumb0:
As discussed we shouldn't change the format of input of RPC methods, not introduce inconsistencies with other clients.
We could change command line format though if it's needed and you see how it could be improved.

Also as I noted above I think `(default|required)` part is not inherently part of enode address format, so its parsing shouldn't be in `NodeSpec` class.

halfalicious:
> As discussed we shouldn't change the format of input of RPC methods, not introduce inconsistencies with other clients.
> We could change command line format though if it's needed and you see how it could be improved.
> 
> Also as I noted above I think `(default|required)` part is not inherently part of enode address format, so its parsing shouldn't be in `NodeSpec` class.

Agreed, I'm going with that approach. As for the format of URL that we will support in `NodeSpec`, I'm going with the format outlined here: https://github.com/ethereum/wiki/wiki/enode-url-format. Example string:

`enode://6f8a80d14311c39f35f516fa664deaaaa13e85b2f7493f37f6144d86991ec012937307647bd3b9a82abe2974e1407241d54947bbb39763a4cac9f77166ad92a0@10.3.58.6:30303?discport=30301`

Both geth and web3.py support this format of URL, though there's no mention in their docs of the trailing "?discport=<udpport>"...this is optional though so I think it's okay if we support it as long as we treat it as such.

PR #5237, opened 2018-08-31
Test for CREATE2 over the account with non-empty storage

gumb0:


test/unittests/libevm/VMTest.cpp
```diff
@@ -433,6 +446,11 @@ BOOST_AUTO_TEST_CASE(LegacyVMCreate2isForbiddenInStaticCall)
     testCreate2isForbiddenInStaticCall();
 }
 
+BOOST_AUTO_TEST_CASE(LegacyVMCreate2collisionWithNonEmptyStorage)
+{
+    testCreate2collisionWithNonEmptyStorage();
```
chfast: Why this is layout this way?

PR #5238, opened 2018-09-03
Implement SSTORE net gas metering in aleth-interpreter

### LINKED ISSUE #5226: Implement net gas metering for SSTORE without dirty maps
https://eips.ethereum.org/EIPS/eip-1283

gumb0:
Closes #5226 

Includes the changes due to EVMC API change (in the separate commit)

### DIFF
--- a/libaleth-interpreter/VM.cpp
+++ b/libaleth-interpreter/VM.cpp
@@ -1353,22 +1353,19 @@ void VM::interpretCases()
             if (m_message->flags & EVMC_STATIC)
                 throwDisallowedStateChange();
 
-            static_assert(
-                VMSchedule::sstoreResetGas <= VMSchedule::sstoreSetGas, "Wrong SSTORE gas costs");
-            m_runGas = VMSchedule::sstoreResetGas;  // Charge the modification cost up front.
-            updateIOGas();
-
-            evmc_uint256be key = toEvmC(m_SP[0]);
-            evmc_uint256be value = toEvmC(m_SP[1]);
+            evmc_uint256be const key = toEvmC(m_SP[0]);
+            evmc_uint256be const value = toEvmC(m_SP[1]);
             auto status =
                 m_context->host->set_storage(m_context, &m_message->destination, &key, &value);
 
             if (status == EVMC_STORAGE_ADDED)
-            {
-                // Charge additional amount for added storage item.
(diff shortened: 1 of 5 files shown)


libaleth-interpreter/VM.cpp
```diff
@@ -1356,22 +1356,19 @@ void VM::interpretCases()
             if (m_message->flags & EVMC_STATIC)
                 throwDisallowedStateChange();
 
-            static_assert(
```
gumb0: For the sake of simplicity I got rid of the trick to first charge `sstoreSetGas`, and possibly `sstoreSetGas - sstoreResetGas` afterwards.

To keep it I think we would need to handle pre-Constantinople as an additional special case. In Constantinople we can't know before actually calling `set_storage` whether the change will require DB write, or will be just a dirty write (unless we decide to provide access to "original storage values" to EVM)

gumb0: Now I'm not sure if this is much simpler because it turns out we still need to handle Constantinople here differently (difference in the cost of X->X)
Here's the version with Constantinople as a separate case and keeping the old optimization: https://github.com/ethereum/aleth/pull/5240

chfast: I think to be safe we should charge more/less the SLOAD cost, so `sstoreUnchangedGas` is good candidate and handles both pre and post Constantinople cases. 
We want to make sure the `sstoreUnchangedGas` is the smallest from all SSTORE possible costs. Would be nice to use `std::min` here but it started be a `constexpr` function in C++14.

libevm/ExtVMFace.cpp
```diff
@@ -58,23 +58,49 @@ evmc_storage_status setStorage(evmc_context* _context, evmc_address const* _addr
...
+    if (newValue == currentValue)
         return EVMC_STORAGE_UNCHANGED;
 
+    EVMSchedule const& schedule = env.evmSchedule();
     auto status = EVMC_STORAGE_MODIFIED;
-    if (oldValue == 0)
-        status = EVMC_STORAGE_ADDED;
-    else if (value == 0)
+    u256 const originalValue = env.originalStorageValue(index);
+    if (originalValue == currentValue || !schedule.eip1283Mode)
```
chfast: I think this should be simplified. We should report `EVMC_STORAGE_MODIFIED_DIRTY` also in pre-Constantinople. It will be simply handled together with `EVMC_STORAGE_MODIFIED`.

PR #5240, opened 2018-09-03
Alternative code in aleth-interpreter SSTORE

gumb0:
Maybe this version is better, not sure.

### DIFF
--- a/libaleth-interpreter/VM.cpp
+++ b/libaleth-interpreter/VM.cpp
@@ -1356,22 +1356,19 @@ void VM::interpretCases()
             if (m_message->flags & EVMC_STATIC)
                 throwDisallowedStateChange();
 
-            static_assert(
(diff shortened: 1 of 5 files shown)


libaleth-interpreter/VM.cpp
```diff
@@ -1355,23 +1355,42 @@ void VM::interpretCases()
...
+                if (status == EVMC_STORAGE_ADDED)
+                {
+                    // Charge additional amount for added storage item.
+                    m_runGas = VMSchedule::sstoreSetGas - VMSchedule::sstoreResetGas;
+                    updateIOGas();
+                }
             }
+            else
+            {
+                auto const status =
```
gumb0: We could also charge `sstoreUnchangedGas` upfront similar to pre-Constantinople way, but maybe it doesn't make much sense because doesn't prevent extra writes to DB

PR #5241, opened 2018-09-04
testeth select BlockchainTest by --singletest, --singletestnet

winsvega:
run single test from BCGeneralStateTests

usage:
```
./testeth -t BCGeneralStateTests/stCreate2 --  --all  --singletest create2collisionStorage_d0g0v0 --singlenet Constantinople --vmtrace
```

it was impossible to run a single test from BCGeneralStateTests before.

### DIFF
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -67,8 +67,20 @@ json_spirit::mValue BlockchainTestSuite::doTests(json_spirit::mValue const& _inp
         json_spirit::mObject const& inputTest = i.second.get_obj();
 
         //Select test by name if --singletest is set and not filling state tests as blockchain
-        if (!Options::get().fillchain && !TestOutputHelper::get().checkTest(testname))
-            continue;
+        if (!Options::get().fillchain)
+        {
+            if (test::Options::get().singleTest)
+                BOOST_REQUIRE_MESSAGE(!test::Options::get().singleTestNet.empty(),
+                    "Selecting a single Blockchain test requires --singlenet argument!");
+
+            // Select BC Test by singleTest + singleTestNet
+            if (test::Options::get().singleTest &&
+                test::Options::get().singleTestName + "_" + test::Options::get().singleTestNet !=
+                    testname)
+                continue;
+
+            TestOutputHelper::get().setCurrentTestName(testname);
+        }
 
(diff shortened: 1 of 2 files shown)


gumb0:
GCC6 build failed

winsvega:
Fixed

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -67,8 +67,20 @@ json_spirit::mValue BlockchainTestSuite::doTests(json_spirit::mValue const& _inp
         json_spirit::mObject const& inputTest = i.second.get_obj();
 
         //Select test by name if --singletest is set and not filling state tests as blockchain
-        if (!Options::get().fillchain && !TestOutputHelper::get().checkTest(testname))
-            continue;
+        if (!Options::get().fillchain)
+        {
+            if (test::Options::get().singleTest)
+                BOOST_REQUIRE_MESSAGE(!test::Options::get().singleTestNet.empty(),
+                    "Selecting a single Blockchain test requires --singlenet argument!");
```
gumb0: Maybe you could instead of this check if `test::Options::get().singleTestName` equals beginning of the  `testname` and separately if `test::Options::get().singleTestNet` equals the end of `testname`, then two options would work independently I guess

winsvega:
updated

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -130,13 +130,29 @@ json_spirit::mValue BlockchainTestSuite::doTests(json_spirit::mValue const& _inp
...
+            if (!Options::get().fillchain)
+            {
+                Options const& opt = test::Options::get();
+
+                // Select BC Test by singleTest
+                if (opt.singleTest && testname.find(opt.singleTestName) == string::npos)
+                    continue;
+
+                // Select BC Test by singleNet
+                if (!opt.singleTestNet.empty() && testname.rfind(opt.singleTestNet) == string::npos)
```
gumb0: There's no much difference between using `find` and `rfind` here if you do it like this (it will search the whole string anyway)
There are `boost::algorithm::starts_with`, `boost::algorithm::ends_with` for this

gumb0: So if you're worried, you should change it, it will help.

(`rfind("Byzantium")` will find it in your case)

winsvega:
Changed

PR #5242, opened 2018-09-04
Fix clearing storage in case of CREATE2 with empty init code

gumb0:
Storage of the newly created account was cleared only if init code is non-empty (in `Executive::go()`.

That could be a problem in case of new account address collision when nonce is zero and code is empty, but storage is not empty. Creation should succeed in this case (account still is considered empty) and the storage should be cleared.


gumb0:
@winsvega Please recreate CREATE2 (and other create collision) tests

PR #5245, opened 2018-09-04
Clear storage original cache together with storage overlay

gumb0:
I cant't think of any case where this could be a problem (that's why I didn't do it right away), but still it feels safer / more logical to clear original cache when we reset storage.

(For it to affect any consensus it would be necessary to overwrite account which previously has read some storage in the same transaction)


chfast:
Any ideas how to make a test case for it?

PR #5252, opened 2018-09-06
Add assertion to EVMC factory create

axic:


axic:
@chfast do you plan to make a final build off 1.4? If so should we merge this there instead?

axic:
@chfast @gumb0 moved onto the 1.4 branch.

PR #5255, opened 2018-09-06
Update EVMC to 5.2.0

axic:
Mostly needed in order to support `evmc_create` in the loader.


gumb0:
You forgot to update aleth code :) (doesn't compile now)

axic:
Oh wow, 5.2.0 should have been a non-breaking change. Will check.

chfast:
It's not breaking on the ABI level, but some function definitions were physically moved and renamed.

axic:
@chfast hope fixed it

chfast:
You did not. I will handle this.

PR #5266, opened 2018-09-14
Install codespell on cpp-build-env

axic:


chfast:
Tagged with `build-env-v6`. Should be available as docker `ethereum/cpp-build-env:6`.

PR #5267, opened 2018-09-14
Display firendlier ABI version mismatch error in VMFactory

axic:


axic:
> Does this solve the Hera + Aleth 1.4.0rc6 issue?

No, but I think #5268 does.

PR #5268, opened 2018-09-14
Fix testeth crash if EVMC load failed

axic:
Fixes #5229.

### DIFF
--- a/libevm/VMFactory.cpp
+++ b/libevm/VMFactory.cpp
@@ -85,8 +85,15 @@ void setVMKind(const std::string& _name)
     }
 
     // If no match for predefined VM names, try loading it as an EVMC VM DLL.
+    g_kind = VMKind::DLL;
+
+    // Release previous instance
+    g_evmcDll.reset();
+
     evmc_loader_error_code ec;
-    g_evmcDll.reset(new EVMC{evmc_load_and_create(_name.c_str(), &ec)});
+    evmc_instance *instance = evmc_load_and_create(_name.c_str(), &ec);
+    assert(ec == EVMC_LOADER_SUCCESS || instance == nullptr);
+
     switch (ec)
     {
     case EVMC_LOADER_SUCCESS:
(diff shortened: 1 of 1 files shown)


libevm/VMFactory.cpp
```diff
@@ -86,7 +86,8 @@ void setVMKind(const std::string& _name)
 
     // If no match for predefined VM names, try loading it as an EVMC VM DLL.
     evmc_loader_error_code ec;
-    g_evmcDll.reset(new EVMC{evmc_load_and_create(_name.c_str(), &ec)});
+    evmc_instance *instance = evmc_load_and_create(_name.c_str(), &ec);
+    g_evmcDll.reset(nullptr);
```
axic: Not sure we need this.

Also could have an assertion that if `ec != EVMC_LOADER_SUCCESS` then `instance == nullptr`

chfast: I think this is ok to release the previous VM. You can even do it before `evmc_load_and_create`. 
Also, you don't need `nullptr` for argument, it can be just `.reset()`.

chfast:
So the issue is that assert is triggered inside EVMC class before the exception has a chance to being handled?

axic:
> So the issue is that assert is triggered inside EVMC class before the exception has a chance to being handled?

Yep.

axic:
@chfast pushed I think the version I think is the best.

libevm/VMFactory.cpp
```diff
@@ -85,8 +85,15 @@ void setVMKind(const std::string& _name)
     }
 
     // If no match for predefined VM names, try loading it as an EVMC VM DLL.
+    g_kind = VMKind::DLL;
+
+    // Release previous instance
+    g_evmcDll.reset();
```
axic: With these two above, even if testeth would ignore the exceptions below, the factory will be in an expected failing state and the assertion in `VMFactory::create` will take of that.

axic:
Tested locally, seems to be working.

PR #5272, opened 2018-09-17
Fix for serialising Boost bigint's when using msvc.

sipsorcery:
With msvc the `(byte)_i` cast is being applied to the underlying bigint data structure and not the number the bigint represents. By adding the `& oxff ` before the cast it ensures the cast is being applied to the least significant byte of the underlying bigint number.


axic:
@gumb0 Is it possible to add a test case for this?

sipsorcery:
If it's of any use this the test case I used is below. Without this fix it fails on msvc. Passes on Linux before and after. Other platforms not tested.

```
dev::u256 startVal("10000000000000000000000");
std::cout << "Start val=" << startVal << "." << std::endl;
dev::RLPStream rlpStm{};
rlpStm << startVal;
auto buffer = rlpStm.out();
std::cout << "Serialised RLP u256=" << dev::toHex(buffer) << "." << std::endl;
dev::RLP rlp{ buffer };
dev::u256 returnedU256 = rlp.toInt<dev::u256>();
std::cout << "Deserialised u256=" << returnedU256 << "." << std::endl;
bool testResult = (startVal == returnedU256);
```

sipsorcery:
@chfast unit test added.

winsvega:
there are RLP tests: https://github.com/ethereum/tests/blob/develop/RLPTests/rlptest.json#L154
but it is up to the client to run this check. so if you could construct theoretical invalid RLP that should happen if this is not fixed, then we could create malicious blockchain test and check that client is rejecting such a block because of invalid RLP encoding. (a block that has invalid bigint encoding)

sipsorcery:
Windows 10 Pro 10.0.17134; Visual Studio Community 2017 15.8.4; Boost 1.68.0 x64.

Test run without patch:

```
[DEBUG] CheckUint256Serialisation test commencing...
Start val=10000000000000000000000.
Serialised RLP u256=8a02ffffffffffffffffff.
Deserialised u256=14167099448608935641087.
[DEBUG] CheckUint256Serialisation result false.
```

Test run with patch:

```
[DEBUG] CheckUint256Serialisation test commencing...
Start val=10000000000000000000000.
Serialised RLP u256=8a021e19e0c9bab2400000.
Deserialised u256=10000000000000000000000.
[DEBUG] CheckUint256Serialisation result true.
```

Looking at the [Boost documentation](https://www.boost.org/doc/libs/1_66_0/libs/multiprecision/doc/html/boost_multiprecision/tut/ints/cpp_int.html) for one of the multi precision integer backends does seem to indicate that the underlying data structure may not be a direct byte representation of the integer. When debugging with Visual Studio the `(byte)_i` cast is being applied to `boost::multiprecision::backends::cpp_int_base<0,4294967295,1,0,std::allocator<unsigned long>,0> = {m_data={ld={capacity=2990538752 data=0xcccccccc0000021e {???} } la=0x000000ca80efdf38 {2990538752, ...} ...} ...}`. 

I guess the cast works as long as `data` is the first item in the data structure AND data is stored LSB first. Seems a lot safer to do a standard numeric operation such as `&` rather than rely on a cast against a data structure that does appear to vary based on compiler/platform implementation.

gumb0:
We use only VS 2015 and boost 1.65.1 currently to build aleth and I see in the debugger that conversion to `byte` goes for me into conversion method which has a loop to go over the limbs and it certainly doesn't work like `reinterpret_cast` of cpp_int representation.

It might be that something is broken it newer boost (I see there were some changes around this conversion in recent versions https://github.com/boostorg/multiprecision/commit/bbe819f8034a3c854deffc6191410b91ac27b3d6 ), I'm just curious what's really happening there. Maybe I'll try it with new boost

BTW did you manage to build aleth / testeth with VS 2017 & boost 1.68 or only some part of it somehow?

sipsorcery:
> BTW did you manage to build aleth / testeth with VS 2017 & boost 1.68 or only some part of it somehow?

I did the very first time I tried it which was approximately 3 weeks ago.

I then tried again a week ago and haven't been able to get the build to work on my Windows dev machine. I suspect I've got a dependency issue and probably need to clean out something somewhere. I haven't looked into it too deeply yet.

Here's the last error message cmake spat out at me (I used `cmake .. -G "Visual Studio 15 2017 Win64"` which could also be the problem as the dependencies might only be set for VS14):

```
[hunter ** FATAL ERROR **] Build step failed (dir: C:/.hunter/_Base/aa1facc/72c99c2/4de1f3d/Build/yaml-cpp
[hunter ** FATAL ERROR **] [Directory:C:/.hunter/_Base/Download/Hunter/0.21.16/aa1facc/Unpacked/cmake/projects/yaml-cpp]
```

sipsorcery:
@gumb0  Could you try this test program?
```
#include <iostream>
#include <boost/multiprecision/cpp_int.hpp>

using u256 = boost::multiprecision::number<boost::multiprecision::cpp_int_backend<256, 256, boost::multiprecision::unsigned_magnitude, boost::multiprecision::unchecked, void>>;

int main() {
	u256 i{ "10000000000000000000000" };
	for (; i; i >>= 8)
		std::cout << std::hex << "(byte)i=" << +(uint8_t)i << ", (byte)(i & 0xff)=" << +(uint8_t)(i & 0xff) << "." << std::endl;
}
```

My result:
```
(byte)i=ff, (byte)(i & 0xff)=0.
(byte)i=ff, (byte)(i & 0xff)=0.
(byte)i=ff, (byte)(i & 0xff)=40.
(byte)i=ff, (byte)(i & 0xff)=b2.
(byte)i=ff, (byte)(i & 0xff)=ba.
(byte)i=ff, (byte)(i & 0xff)=c9.
(byte)i=ff, (byte)(i & 0xff)=e0.
(byte)i=ff, (byte)(i & 0xff)=19.
(byte)i=ff, (byte)(i & 0xff)=1e.
(byte)i=2, (byte)(i & 0xff)=2.
```

Visual Studio 2017 build log (runtime library set to "Multi-threaded Debug"):
```
1>------ Rebuild All started: Project: ConsoleApplication2, Configuration: Debug x64 ------
1>Microsoft (R) C/C++ Optimizing Compiler Version 19.15.26729 for x64
1>Copyright (C) Microsoft Corporation.  All rights reserved.
1>
1>cl /c /I"N:\Dev\vcpkg\installed\x64-windows-static\include" /ZI /JMC /W3 /WX- /diagnostics:classic /sdl /Od /D _DEBUG /D _CONSOLE /D _UNICODE /D UNICODE /Gm- /EHsc /RTC1 /MTd /GS /fp:precise /permissive- /Zc:wchar_t /Zc:forScope /Zc:inline /Fo"x64\Debug\\" /Fd"x64\Debug\vc141.pdb" /Gd /TP /FC /errorReport:prompt ConsoleApplication2.cpp
1>
1>ConsoleApplication2.cpp
1>Microsoft (R) Incremental Linker Version 14.15.26729.0
1>Copyright (C) Microsoft Corporation.  All rights reserved.
1>
1>"/OUT:C:\Users\aaron\source\repos\ConsoleApplication2\x64\Debug\ConsoleApplication2.exe" /INCREMENTAL "/LIBPATH:N:\Dev\vcpkg\installed\x64-windows-static\debug\lib" "/LIBPATH:N:\Dev\vcpkg\installed\x64-windows-static\debug\lib\manual-link" kernel32.lib user32.lib gdi32.lib winspool.lib comdlg32.lib advapi32.lib shell32.lib ole32.lib oleaut32.lib uuid.lib odbc32.lib odbccp32.lib "N:\Dev\vcpkg\installed\x64-windows-static\debug\lib\*.lib" /MANIFEST "/MANIFESTUAC:level='asInvoker' uiAccess='false'" /manifest:embed /DEBUG:FASTLINK "/PDB:C:\Users\aaron\source\repos\ConsoleApplication2\x64\Debug\ConsoleApplication2.pdb" /SUBSYSTEM:CONSOLE /TLBID:1 /DYNAMICBASE /NXCOMPAT "/IMPLIB:C:\Users\aaron\source\repos\ConsoleApplication2\x64\Debug\ConsoleApplication2.lib" /MACHINE:X64 x64\Debug\ConsoleApplication2.obj
1>ConsoleApplication2.vcxproj -> C:\Users\aaron\source\repos\ConsoleApplication2\x64\Debug\ConsoleApplication2.exe
========== Rebuild All: 1 succeeded, 0 failed, 0 skipped ==========
```

ras0219-msft:
>I suspect I've got a dependency issue and probably need to clean out something somewhere.

@sipsorcery Let us know if it turns out to be an issue with vcpkg :)

winsvega:
got an error when refilling the tests: 

```
Test Case "bcBlockGasLimitTest": 
100%
Filling GasUsedHigherThanBlockGasLimitButNotWithRefundsSuicideLast_EIP150
/home/wins/Ethereum/cpp-ethereum/libdevcore/RLP.h(272): fatal error: in "_T dev::RLP::toInt(int) const [with _T = long int]": boost::exception_detail::clone_impl<dev::BadCast>: /home/wins/Ethereum/cpp-ethereum/libdevcore/RLP.h(272): Throw in function _T dev::RLP::toInt(int) const [with _T = long int]
Dynamic exception type: boost::exception_detail::clone_impl<dev::BadCast>
[dev::eth::tag_data*] = a0ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
[dev::eth::tag_field*] = invalid block header format
[dev::eth::tag_field*] = 11
```

could this be related?

gumb0:
@winsvega this one is only about windows

gumb0:
@sipsorcery my result
```
(byte)i=0, (byte)(i & 0xff)=0.
(byte)i=0, (byte)(i & 0xff)=0.
(byte)i=40, (byte)(i & 0xff)=40.
(byte)i=b2, (byte)(i & 0xff)=b2.
(byte)i=ba, (byte)(i & 0xff)=ba.
(byte)i=c9, (byte)(i & 0xff)=c9.
(byte)i=e0, (byte)(i & 0xff)=e0.
(byte)i=19, (byte)(i & 0xff)=19.
(byte)i=1e, (byte)(i & 0xff)=1e.
(byte)i=2, (byte)(i & 0xff)=2.
```

Also the same on some online VC++ compiler http://rextester.com/JUACWO36965

chfast:
@sipsorcery Wait with this, I will simplify that.

chfast:
So I rebased it and also simplified the test a bit.

sipsorcery:
@chfast thanks! I was battling some merge conflicts and pushed a commit before I saw your post. Not sure best approach now? Should I take the branch back to a specific commit?

Issue #5262, opened 2018-09-10
In-memory database implementation


gumb0:
This issue is the continuation of https://github.com/ethereum/aleth/issues/4669 and describes one point from it:

>  Implement MemoryDB as a simple wrapper of std::unordered_map.

`OverlayDB` refactoring can be done as a separate task later. Until then we need to rename currently existing `MemoryDB`

- [x] Rename `MemoryDB`, I suggest the name `StateMemoryDB`

- [x] Implement `DatabaseFace` using `std::unordered_map` as backend key/value storage. This will be the new `MemoryDB`

- [x] Allow selection of either `MemoryDB` or `LevelDB` at run-time (Currently DB implementation is selected at compile time via `DBImpl` alias) I think it can be done in the fashion simiar to how VM implementation is created currently, see `setVMKind()` / `VMFactory::create()` 
Make sure the DB factory is used in `BlockChain` and when creating state DB (`OverlayDB`)

- [x] Add aleth command line parameter to enable `MemoryDB` (maybe `--memory-db`). We could create something similar to `vmProgramOptions()` for this.
(But I'd say we don't need to list all supported DB engines in CLI, so DB program options would include just `--db-path` and `--memory-db` (mutually exclusive))

cc @chfast

chfast:
I don't get why `StateMemoryDB`, but I have not seen its implementation.

For CLI I prefer having common prefix for groups. It could be something like this:

1. `--db-path` and `--db-in-memory`.
2. `--db-path=memory`.
3. `--db={leveldb, memory}` and `--db-path` acceptable only for memory kind.
4. `--db=leveldb://path`, `--db=memory`.

gumb0:
> I don't get why `StateMemoryDB`, but I have not seen its implementation.

This class is confusingly used as both a cache when accessing State DB and as in-memory DB (at least in some tests, this is the part we want to replace with new `MemoryDB` later).
So it's kind of neither exactly cache nor exactly DB and so coming up with great name is difficult. Anyway I hope this would a temporary thing and we'll refactor it out.

For CLI I think I like option 1.

chfast:
How about StateCacheDB?

gumb0:
Fine with me.

halfalicious:
[...]

PR #5286, opened 2018-09-25
Implement new MemoryDB and make DB selectable at runtime

halfalicious:
These changes take care of the first 2 items in #5262 - renaming the "old" MemoryDB to StateCacheDB and creating a new thread-safe MemoryDB implementation which implements `DatabaseFace `and is backed by `std::unordered_map`. I've verified that the changes compile clean in both the Debug and RelWithDebInfo configurations and that the StateCacheDB tests all still pass: 

```
C:\Users\nilse\Documents\Code\aleth\build\test\Debug>testeth -t stateCacheDB
Running 7 test cases...
Test Case "kill":
Test Case "purgeMainMem":
Test Case "purgeMainMem_Refs":
Test Case "purgeAuxMem":
Test Case "copy":
Test Case "lookUp":
Test Case "stream":

*** No errors detected

C:\Users\nilse\Documents\Code\aleth\build\test\Debug>
```


libdevcore/MemoryDB.cpp
```diff
@@ -1,189 +1,104 @@
...
-        else
-            it = m_aux.erase(it);
+    auto* batchPtr = dynamic_cast<MemoryDBWriteBatch*>(_batch.get());
+    if (!batchPtr)
+    {
+        BOOST_THROW_EXCEPTION(DatabaseError() << errinfo_comment("Invalid batch type passed to MemoryDB::commit"));
+    }
+    MemoryDBBatch batch = batchPtr->writeBatch();
+    std::lock_guard<std::mutex> lock(m_mutex);
+    for (const auto& e : batch)
```
gumb0: we put `const` after type specifier, so this should be `auto const&`

libdevcore/MemoryDB.cpp
```diff
@@ -1,189 +1,104 @@
...
-#if DEV_GUARDED_DB
-    ReadGuard l(x_this);
-#endif
-    h256Hash ret;
-    for (auto const& i: m_main)
-        if (i.second.second)
-            ret.insert(i.first);
-    return ret;
+    std::lock_guard<std::mutex> lock(m_mutex);
+    for (const auto& e : m_db)
```
gumb0: `auto const&`

gumb0:
Note that we usually rebase PR branch on the `master` instead of merging `master` into branch, to keep history more linear.

chfast:
I have not reviewed it all yet, but will be easier to handle the review if the first two commits are sent as a separate PR.

halfalicious:
> Note that we usually rebase PR branch on the `master` instead of merging `master` into branch, to keep history more linear.

Makes sense, I'll look into how to do that after the PR is signed off.

libdevcore/MemoryDB.h
```diff
@@ -14,79 +14,37 @@
...
+    // A database must implement the `forEach` method that allows the caller
+    // to pass in a function `f`, which will be called with the key and value
+    // of each record in the database. If `f` returns false, the `forEach`
+    // method must return immediately.
+    void forEach(std::function<bool(Slice, Slice)> f) const override;
 
 private:
-    MemoryDB const& m_o;
-    bool m_r;
+        std::unordered_map<std::string, std::string> m_db;
```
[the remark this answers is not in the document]
halfalicious: DatabaseFace::lookup returns std::string, I'll look into how LevelDB goes from slices to strings before returning the data.

libdevcore/MemoryDB.cpp
```diff
@@ -1,189 +1,104 @@
...
+std::string MemoryDB::lookup(Slice _key) const
 {
-#if DEV_GUARDED_DB
-    WriteGuard l(x_this);
-#endif
-    auto it = m_main.find(_h);
-    if (it != m_main.end())
+    std::string value;
+    std::lock_guard<std::mutex> lock(m_mutex);
+    if (m_db.count(_key.data()))
```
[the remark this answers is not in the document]
halfalicious: Find can return empty so I still need to check for that before returning. Doing that check is probably cheaper than calling count() though so I'll go with that approach instead. Thanks for calling this out!

gumb0:
Last commit is not linked to your account, I think you can fix it by adding an email you used (in git) to your github profile

libdevcore/MemoryDB.h
```diff
@@ -40,11 +40,11 @@ class MemoryDB : public DatabaseFace
...
     // of each record in the database. If `f` returns false, the `forEach`
     // method must return immediately.
-    void forEach(std::function<bool(Slice, Slice)> f) const override;
+    void forEach(std::function<bool(Slice, Slice)> _f) const override;
 
 private:
-        std::unordered_map<std::string, std::string> m_db;
-        mutable std::mutex m_mutex;
+    std::unordered_map<std::string, std::string> m_db;
+    mutable dev::Mutex m_mutex;
```
gumb0: `dev::` is not really needed, we're in `dev`

libdevcore/MemoryDB.cpp
```diff
@@ -33,30 +49,31 @@ void MemoryDBWriteBatch::kill(Slice _key)
...
+        value = it->first;
     }
     return value;
 }
 
 bool MemoryDB::exists(Slice _key) const
 {
-    std::lock_guard<std::mutex> lock(m_mutex);
+    dev::Guard lock(m_mutex);
     return m_db.count(_key.data()) != 0;
```
gumb0: must be `_key.toString()`

gumb0: Also in `MemoryDBWriteBatch` methods above

libdevcore/MemoryDB.cpp
```diff
@@ -33,30 +49,31 @@ void MemoryDBWriteBatch::kill(Slice _key)
...
-        value = m_db.find(_key.data())->first;
+        value = it->first;
     }
     return value;
 }
 
 bool MemoryDB::exists(Slice _key) const
 {
-    std::lock_guard<std::mutex> lock(m_mutex);
+    dev::Guard lock(m_mutex);
```
gumb0: can be just `Guard`

libdevcore/MemoryDB.cpp
```diff
@@ -88,17 +106,17 @@ void MemoryDB::commit(std::unique_ptr<WriteBatchFace> _batch)
...
-void MemoryDB::forEach(std::function<bool(Slice, Slice)> f) const
+void MemoryDB::forEach(std::function<bool(Slice, Slice)> _f) const
 {
-    std::lock_guard<std::mutex> lock(m_mutex);
-    for (const auto& e : m_db)
+    dev::Guard lock(m_mutex);
+    for (auto const& e : m_db)
     {
-        if (!f(Slice(e.first.c_str()), Slice(e.second.c_str())))
+        if (!_f(Slice(e.first.c_str()), Slice(e.second.c_str())))
```
gumb0: Constructing `Slice` from `c_str()` also assumes it's zero-terminated string, which it is not. Look at other `vector_ref` constructors, is there constructor just from `std::string` ?

libdevcore/MemoryDB.cpp
halfalicious: Sorry about that, I thought I had gotten all of them.

halfalicious:
> Last commit is not linked to your account, I think you can fix it by adding an email you used (in git) to your github profile

Thanks for the heads up I didn't even notice that, I've fixed it.

chfast:
Also, this should be rebased to remove the "Merge" commit.

halfalicious:
I've fixed the issues that @gumb0  called out and have created a set of unit tests which cover everything but `foreach `(I'm planning on creating the `forEach `tests tomorrow). Once I've created the foreach tests and verified they all pass I'll push my latest changes (I'm hoping to do this tomorrow evening) and resume validating my `DBFactory `changes. Some notes/questions:

- I added a size function to both `MemoryDB `and `MemoryDBWriteBatch `for testing purposes (so the unit tests can validate how many elements have been inserted into the db / write batch, respectively). I chose this approach over making the MemoryDB test suite a `friend `class since it seems cleaner plus the functions can potentially have utility outside of testing. Thoughts? 
- I moved the `MemoryDBWriteBatch `class declaration from the CPP file to the header file so that the unit tests have access to it so they can validate it. Let me know if you'd like me to go with a different approach. 
- I'm running into build issues again on my home machine, even after completely uninstalling Visual Studio 2017. I'll create a new issue for tracking purposes and try to untangle it tomorrow / this weekend.

gumb0:
The testing approach is ok I think.

libdevcore/MemoryDB.cpp
```diff
@@ -14,176 +14,109 @@
...
-    auto it = m_main.find(_h);
-    if (it != m_main.end())
+    std::string value;
+    dev::Guard lock(m_mutex);
+    auto const& it = m_db.find(_key.toString());
+    if (it != m_db.end())
     {
-        it->second.first = _v.toString();
-        it->second.second++;
+        value = it->first;
```
halfalicious: This should be it->second

halfalicious:
> I've fixed the issues that @gumb0 called out and have created a set of unit tests which cover everything but `foreach`(I'm planning on creating the `forEach`tests tomorrow). Once I've created the foreach tests and verified they all pass I'll push my latest changes (I'm hoping to do this tomorrow evening) and resume validating my `DBFactory`changes. Some notes/questions:
> 
> * I added a size function to both `MemoryDB`and `MemoryDBWriteBatch`for testing purposes (so the unit tests can validate how many elements have been inserted into the db / write batch, respectively). I chose this approach over making the MemoryDB test suite a `friend`class since it seems cleaner plus the functions can potentially have utility outside of testing. Thoughts?
> * I moved the `MemoryDBWriteBatch`class declaration from the CPP file to the header file so that the unit tests have access to it so they can validate it. Let me know if you'd like me to go with a different approach.
> * I'm running into build issues again on my home machine, even after completely uninstalling Visual Studio 2017. I'll create a new issue for tracking purposes and try to untangle it tomorrow / this weekend.

U

> I've fixed the issues that @gumb0 called out and have created a set of unit tests which cover everything but `foreach`(I'm planning on creating the `forEach`tests tomorrow). Once I've created the foreach tests and verified they all pass I'll push my latest changes (I'm hoping to do this tomorrow evening) and resume validating my `DBFactory`changes. Some notes/questions:
> 
> * I added a size function to both `MemoryDB`and `MemoryDBWriteBatch`for testing purposes (so the unit tests can validate how many elements have been inserted into the db / write batch, respectively). I chose this approach over making the MemoryDB test suite a `friend`class since it seems cleaner plus the functions can potentially have utility outside of testing. Thoughts?
> * I moved the `MemoryDBWriteBatch`class declaration from the CPP file to the header file so that the unit tests have access to it so they can validate it. Let me know if you'd like me to go with a different approach.
> * I'm running into build issues again on my home machine, even after completely uninstalling Visual Studio 2017. I'll create a new issue for tracking purposes and try to untangle it tomorrow / this weekend.

I've added the foreach tests and verified that everything builds clean and that all tests pass locally. I'll revisit the DBFactory integration over the weekend.

halfalicious:
@chfast :Thanks for the tip about commit messages, makes sense (so one can easily tell by glancing at the commit what it covers), I'll be sure to do that in the future.

halfalicious:
My latest set of commits complete this issue (barring any bugs that need to be fixed before rebasing). The `DBFactory ` has been integrated and one can now select the database (LevelDB, which is the default selection, or the new MemoryDB) at runtime via the new '--db' parameter. I haven't performed any qualitative perf benchmarks but anecdotally, I kicked off a new sync using MemoryDB and blocks started being imported nearly immediately and were imported faster than in the LevelDB case. Some additional notes:
* Here's the new command-line output for the database options:
```

DATABASE OPTIONS:
  --db <name> (=leveldb)                                      Select database implementation. Available options are: leveldb, memorydb.
  --db-path <path> (=C:\Users\nilse\AppData\Roaming\Ethereum) Database path (for non-memory database options)
```

* @gumb0: I know that you mentioned that we don't need to log all database options, but I decided to do so because I think it improves usability (one can easily glance at the help output and know what the options are without having to go digging for documentation or looking through the code), the list isn't very long, and we list the various VM types for the VM option. If you still feel like we shouldn't list the DB types let me know and I can pull them out. 

* I've repurposed `--db-path` to specify the database path and created a new parameter (`--data-dir`) for other files (which, as far as I know, are the keystore, secret, and configuration files). 
* Given that `DBFactory `now includes the database path, I've pulled out the `Defaults` class. 
* Both the DB path and the "other files" path default to the same path
* I've created what I think to be a pretty exhaustive set of tests for the new MemoryDB. They are pretty straightforward and all pass.
* I updated the OverlayDB tests to explicitly specify LevelDB when creating the new DB via `DBFactory` despite it already being the default option in case we want to change the default in the future, and so it's clear to code readers which DB is being created. I've verified that the OverlayDB tests all still pass as well. 
* I pass the DB path to the `DBFactory::create` function, even when the MemoryDB option is set (the path is silently ignored), because I think this is cleaner than having a MemoryDB-specific special case for each DB creation call where a path isn't supplied.

aleth/main.cpp
```diff
@@ -769,7 +770,7 @@ int main(int argc, char** argv)
         AccountManager::streamAccountHelp(cout);
         AccountManager::streamWalletHelp(cout);
         cout << clientDefaultMode << clientTransacting << clientNetworking << clientMining << minerOptions;
-        cout << importExportMode << vmOptions << loggingProgramOptions << generalOptions;
+        cout << importExportMode << vmOptions << dbOptions << loggingProgramOptions << generalOptions;
```
chfast: I think "db options" should be before "vm options".

libdevcore/DBFactory.cpp
```diff
@@ -0,0 +1,151 @@
...
+        return std::unique_ptr<DatabaseFace>(new LevelDB(_path));
+        break;
+    case DatabaseKind::MemoryDB:
+        // Silently ignore path since the concept of a db path doesn't make sense
+        // when using an in-memory database
+        return std::unique_ptr<DatabaseFace>(new MemoryDB());
+        break;
+    default:
+        assert(false);
+        return std::unique_ptr<DatabaseFace>(new LevelDB(_path));
```
chfast: Return null unique_ptr. Can be simple as `return {}`.

libdevcore/DBFactory.cpp
```diff
@@ -0,0 +1,151 @@
...
+    return create(g_kind, _path);
+}
+
+std::unique_ptr<DatabaseFace> DBFactory::create(DatabaseKind _kind, fs::path _path)
+{
+    switch (_kind)
+    {
+    case DatabaseKind::LevelDB:
+        return std::unique_ptr<DatabaseFace>(new LevelDB(_path));
+        break;
```
chfast: `break`s after `return`s are redundant.

libdevcore/MemoryDB.cpp
```diff
@@ -14,176 +14,95 @@
...
+std::string MemoryDB::lookup(Slice _key) const
 {
-#if DEV_GUARDED_DB
-    WriteGuard l(x_this);
-#endif
-    auto it = m_main.find(_h);
-    if (it != m_main.end())
+    std::string value;
+    Guard lock(m_mutex);
+    auto const& it = m_db.find(_key.toString());
```
chfast: This can be simplified a bit as
```
if (it != m_db.end())
    return it->second;
return {};
```

chfast: Also, for single-liners after `if` we skip `{}`.

libdevcore/MemoryDB.cpp
```diff
@@ -14,176 +14,95 @@
...
 }
 
-void MemoryDB::insertAux(h256 const& _h, bytesConstRef _v)
+std::unique_ptr<WriteBatchFace> MemoryDB::createWriteBatch() const
 {
-#if DEV_GUARDED_DB
-    WriteGuard l(x_this);
-#endif
-    m_aux[_h] = make_pair(_v.toBytes(), true);
+    return std::unique_ptr<WriteBatchFace>(new MemoryDBWriteBatch());
```
chfast: Watch out with creating objects as `Object()`, compiler sometimes thinks it is a function definition (not in this context however). Always safe alternatives: `new Object`, `new Object{}`.

libdevcore/MemoryDB.cpp
```diff
@@ -14,176 +14,95 @@
...
-
-    // purge m_aux
-    for (auto it = m_aux.begin(); it != m_aux.end(); )
-        if (it->second.second)
-            ++it;
-        else
-            it = m_aux.erase(it);
+    if (!_batch)
+    {
+        BOOST_THROW_EXCEPTION(DatabaseError() << errinfo_comment("Cannot commit null batch"));
```
chfast: Can this happen?

aleth/main.cpp
```diff
@@ -394,12 +393,13 @@ int main(int argc, char** argv)
 
     po::options_description generalOptions("GENERAL OPTIONS", c_lineWidth);
     auto addGeneralOption = generalOptions.add_options();
-    addGeneralOption("db-path,d", po::value<string>()->value_name("<path>"),
+    addGeneralOption("data-dir,d", po::value<string>()->value_name("<path>"),
         ("Load database from path (default: " + getDataDir().string() + ")").c_str());
```
gumb0: Option description should be changed (Maybe `Load config files and keystore from path`)

libdevcore/DBFactory.cpp
```diff
@@ -0,0 +1,151 @@
...
+{
+namespace db
+{
+namespace fs = boost::filesystem;
+namespace po = boost::program_options;
+
+auto g_kind = DatabaseKind::LevelDB;
+fs::path g_dbPath = getDataDir();
+
+/// A helper type to build the tabled of DB implementations.
```
gumb0: the table

libdevcore/DBFactory.cpp
```diff
@@ -0,0 +1,151 @@
...
+///
+/// More readable than std::tuple.
+/// Fields are not initialized to allow usage of construction with initializer lists {}.
+struct DBKindTableEntry
+{
+    DatabaseKind kind;
+    const char* name;
```
gumb0: `char const*`

libdevcore/DBFactory.cpp
```diff
@@ -0,0 +1,151 @@
...
+    for (auto& entry : dbKindsTable)
+    {
+        if (_name == entry.name)
+        {
+            g_kind = entry.kind;
+            return;
+        }
+    }
+
+    BOOST_THROW_EXCEPTION(std::system_error(std::make_error_code(std::errc::invalid_argument),
```
gumb0: I see this comes from `setVMKind`, but I don't think it's a good idea to throw `system_error` here which is usually thrown from standard library classes working with the OS functions.

In general in C++ it's usually advised to define custom exception types deriving `std::exception`, we also utilize `boost::exception`, and the usual approach is to define in some relevant header exception type like
```
DEV_SIMPLE_EXCEPTION(InvalidDatabaseKind);
```

Then throw it here like
```
BOOST_THROW_EXCEPTION(InvalidDatabaseKind());
```
or if you want to include comment
```
BOOST_THROW_EXCEPTION(InvalidDatabaseKind() << errinfo_comment("invalid database kind supplied: " + _name));
```

libdevcore/DBFactory.h
```diff
@@ -0,0 +1,60 @@
...
+
+/// Provide a set of program options related to databases
+///
+/// @param _lineLength  The line length for description text wrapping, the same as in
+///                     boost::program_options::options_description::options_description().
+boost::program_options::options_description dbProgramOptions(
+    unsigned _lineLength = boost::program_options::options_description::m_default_line_length);
+
+bool isMemoryDB();
+DatabaseKind getDatabaseKind();
```
gumb0: Getters are usually without `get` prefix, so `databaseKind()`

libdevcore/DBFactory.h
```diff
@@ -0,0 +1,60 @@
...
+boost::filesystem::path getDatabasePath();
```
gumb0: and `databasePath()`

libdevcore/DBFactory.cpp
```diff
@@ -0,0 +1,151 @@
...
+        "invalid database kind supplied: " + _name));
+}
+
+void setDatabasePath(const std::string& _path)
```
gumb0: `std::string const&`

libdevcore/MemoryDB.cpp
```diff
@@ -14,176 +14,95 @@
...
+    }
+
+    auto* batchPtr = dynamic_cast<MemoryDBWriteBatch*>(_batch.get());
+    if (!batchPtr)
+    {
+        BOOST_THROW_EXCEPTION(
+            DatabaseError() << errinfo_comment("Invalid batch type passed to MemoryDB::commit"));
+    }
+    std::unordered_map<std::string, std::string> batch = batchPtr->writeBatch();
```
gumb0: This is unnecessary copy of the batch. If you wanted to simplify for loop below, you could just define this as const reference.

libdevcore/MemoryDB.cpp
gumb0: In practice it should, but can if someone calls `db->commit({});`

gumb0:
> Can we also consider improving consistency of naming convention? I.e. "db" vs "database" name.

I'd rename `dmProgramOptions` to `databaseProgramOptions`

gumb0:
> @gumb0: I know that you mentioned that we don't need to log all database options, but I decided to do so because I think it improves usability (one can easily glance at the help output and know what the options are without having to go digging for documentation or looking through the code), the list isn't very long, and we list the various VM types for the VM option. If you still feel like we shouldn't list the DB types let me know and I can pull them out.

My concern is that when we finish intergating RocksDB support, I think we shouldn't build and link both LevelDB and RocksDB at the same time, in other words it should be a compile-time option. So this way `--db` will always have just two possible values.
We can leave your variant now as it is, but I expect some code around setting DB kind might go away with RocksDB integration.

chfast:
> My concern is that when we finish intergating RocksDB support, I think we shouldn't build and link both LevelDB and RocksDB at the same time, in other words it should be a compile-time option. So this way `--db` will always have just two possible values.
> We can leave your variant now as it is, but I expect some code around setting DB kind might go away with RocksDB integration.

Building both is to so bad idea (not by default of curse) for easier performance comparison.

chfast:
For me it is ok to have only 2 options in the final binary at this point.

libdevcore/DBFactory.cpp
halfalicious: Yes, thought I needed this to satisfy the compiler, since I don't I'll remove it. Thanks for calling this out!

libdevcore/MemoryDB.cpp
halfalicious: Ah interesting, thanks for calling this out!

aleth/main.cpp
halfalicious: Ah yes, thank you for catching that!

libdevcore/MemoryDB.cpp
halfalicious: Nice catch!

libdevcore/MemoryDB.cpp
halfalicious: Thanks! Was aware of the skipping {} for single-line ifs but forgot in this case, thanks for catching this!

libdevcore/MemoryDB.cpp
chfast: Both exceptions here are rather candidates for asserts.

chfast:
I can rebase this PR when ready.

Why is AppVeyor slow again?

libethereum/BlockChain.cpp
```diff
@@ -212,58 +211,70 @@ void BlockChain::init(ChainParams const& _p)
...
         {
-            cwarn <<
-                "Database " <<
-                (chainPath / fs::path("blocks")) <<
-                "or " <<
-                (extrasPath / fs::path("extras")) <<
-                "already open. You appear to have another instance of ethereum running. Bailing.";
-            BOOST_THROW_EXCEPTION(DatabaseAlreadyOpen());
+            cwarn << "Unknown database error occurred during in-memory database creation";
+            BOOST_THROW_EXCEPTION(UnknownDatabaseError());
```
gumb0: I would `throw;` here, current exception `ex` could contain some valuable info from lower level, it's better to make it visible than to suppress it.

libethereum/BlockChain.cpp
```diff
@@ -343,10 +354,10 @@ void BlockChain::rebuild(fs::path const& _path, std::function<void(unsigned, uns
 
     // Keep extras DB around, but under a temp name
     m_extrasDB.reset();
-    fs::rename(extrasPath / fs::path("extras"), extrasPath / fs::path("extras.old"));
-    std::unique_ptr<db::DatabaseFace> oldExtrasDB(
-        new db::DBImpl(extrasPath / fs::path("extras.old")));
-    m_extrasDB.reset(new db::DBImpl(extrasPath / fs::path("extras")));
+    if (!db::isMemoryDB())
```
gumb0: I think this `rebuild()` method doesn't make any sense for MemoryDB (it rebuilds the state DB by replaying all blocks from blockchain DB), so let's just return immediately in the beginning if it's MemoryDB (maybe output some log message, too)

libethereum/State.cpp
```diff
@@ -60,40 +59,48 @@ State::State(State const& _s):
...
             cwarn <<
                 "Database " <<
                 (path / fs::path("state")) <<
                 "already open. You appear to have another instance of ethereum running. Bailing.";
             BOOST_THROW_EXCEPTION(DatabaseAlreadyOpen());
         }
+        else
+        {
+            cwarn << "Unknown error when creating in-memory database.";
+            BOOST_THROW_EXCEPTION(UnknownDatabaseError());
```
gumb0: Here too, I would instead do `if (db::isMemoryDB()) throw;` in the beginning of `catch`

test/unittests/libweb3core/overlaydb.cpp
```diff
@@ -20,22 +20,23 @@
...
+using namespace db;
 using namespace dev::test;
 
 BOOST_FIXTURE_TEST_SUITE(OverlayDBTests, TestOutputHelperFixture)
 
 BOOST_AUTO_TEST_CASE(basicUsage)
 {
     TransientDirectory td;
-    std::unique_ptr<db::DBImpl> db(new db::DBImpl(td.path()));
+	std::unique_ptr<db::DatabaseFace> db = DBFactory::create(DatabaseKind::LevelDB, td.path());
```
gumb0: In all the tests in this file, I'd like to use `MemoryDB` instead of `LevelDB` and not use `TransientDirectory`, but this may go into separate PR.

gumb0:
> Why isn't it a good idea to build with both if we only use one at runtime, is it due to the increased binary size?

binary size and build time

gumb0:
> Why is AppVeyor slow again?


I don't know, I thought it's the problem of timeout going down from 2 hours to 1 hour sometimes for unknown reason?

test/unittests/libweb3core/overlaydb.cpp
halfalicious: Got it, I'll take care of that in a separate PR

libdevcore/DBFactory.h
```diff
@@ -0,0 +1,60 @@
...
+boost::filesystem::path databasePath();
+
+class DBFactory
+{
+public:
+    DBFactory() = delete;
+    ~DBFactory() = delete;
+
+    static std::unique_ptr<DatabaseFace> create();
+    static std::unique_ptr<DatabaseFace> create(boost::filesystem::path _path);
```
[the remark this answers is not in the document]
halfalicious: Makes sense!

libethereum/BlockChain.cpp
halfalicious: I took a look at when rebuild is called and it's only called when re-opening an existing database, which obviously doesn't make sense when running MemoryDB. I'll go ahead and return immediately.

test/unittests/libweb3core/overlaydb.cpp
halfalicious: Should we just create a copy of this file which uses MemoryDB since we presumably want test coverage for both the LevelDB and MemoryDB implementations (since LevelDB is the database actually used by users, at least until I get RocksDB online)?

libdevcore/MemoryDB.h
```diff
@@ -14,79 +14,53 @@
...
-
-    bytes lookupAux(h256 const& _h) const;
-    void removeAux(h256 const& _h);
-    void insertAux(h256 const& _h, bytesConstRef _v);
+    void insert(Slice _key, Slice _value) override;
+    void kill(Slice _key) override;
 
-    h256Hash keys() const;
+    std::unordered_map<std::string, std::string> const& writeBatch() const { return m_batch; }
+    std::unordered_map<std::string, std::string>& writeBatch() { return m_batch; }
```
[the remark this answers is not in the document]
halfalicious: Correct, I don't use it anywhere. I'll remove it.

test/unittests/libweb3core/overlaydb.cpp
gumb0: I don't think we need this. I believe unit tests by their nature should be as isolated as possible, so `OverlayDB` tests should test only the code in `OverlayDB`
(this is not quite the case here, since `MemoryDB` is used, so it's not ideal)

We could have additionally some set of tests checking how our `LevelDBDatabase` implementation works, but then `OverlayDB` shouldn't interfere with that.

chfast: @gumb0 can you propose a code change and push it to this PR?

gumb0: I meant that we don't need additional test file as Nils proposed above, this code in this PR is fine

chfast:
The blockchain tests are 6x faster with MemoryDB!
```sh
# LevelDB:
test/testeth -t BlockchainTests  76,08s user 18,78s system 23% cpu 6:39,06 total
# MemoryDB:
test/testeth -t BlockchainTests  68,12s user 10,07s system 107% cpu 1:13,04 total
```

halfalicious:
> I did the rebase with some squashes. It's not perfect, but making it perfect would require splitting some commits (I don't know how to do it easily with preserving the author).

Awesome, thanks! I can take care of that next time :)

Issue #5278, opened 2018-09-19
EthereumHost::randomSelection improvements


gumb0:
`EthereumHost::randomSelection()` sucks.
1. It returns a tuple of 
- vector of requested percent of peers
- vector of the rest of peers
- vector of sessions corresponding to all returned peers.
Sessions are never used but I assume it was a hack to keep session alive while we send blocks/transaction (`EthereumPeer` itself has only `weak_ptr<SessionFace>`)
But I think we can simplify this by just ignoring those peers where session expired, it wouldn't be much of a problem to propagate to less peers.

2. This code
https://github.com/ethereum/aleth/blob/1114b251b8e09d3631557debee3ca1d40f002fa6/libethereum/EthereumHost.cpp#L528-L535
looks quadratic (erasing from the middle of the vector on each iteration) and doesn't do proper uniform random selection (`rand() % allowed.size()` is not uniformly distributed).

See also https://github.com/ethereum/aleth/issues/5277, we probably don't need to select the requested percent of peers, but a square root of peers satisfying the predicate (but at least requested minimum)

Creris:
Hey,

I noticed that this is still open and that the pull request has not been accepted yet. I wonder if I could try to have a go at it?

gumb0:
2 is resolved in #5289, 1 will be resolved when #5305 is merged

PR #5288, opened 2018-09-26
randomSelection() improvements

yuvalzach:
Solved some of the issues expressed in #5278.
In addition, for some reason, elements were removed from the allowed vector when they were places into the chosen vector. I changed this behavior, such that allowed and chosen contain the elements their names imply they do.


yuvalzach:
Welp. Looks like I misunderstood the allowed vector. Sorry :)

PR #5289, opened 2018-09-26
EthereumHost::randomSelection() improvements

lhiny:
Resolved some of the issues brought up in #5278.
Specifically, optimized out the removal of elements in the middle of the 'allowed' vector. Also replaced the modulo operation such that uniform distribution is now present when picking EthereumPeers for the 'chosen' vector.

### DIFF
--- a/libethereum/EthereumHost.cpp
+++ b/libethereum/EthereumHost.cpp
@@ -383,6 +383,8 @@ EthereumHost::EthereumHost(Host const& _host, BlockChain const& _ch, OverlayDB c
     m_peerObserver = make_shared<EthereumPeerObserver>(m_sync, m_tq);
     m_latestBlockSent = _ch.currentHash();
     m_tq.onImport([this](ImportResult _ir, h256 const& _h, h512 const& _nodeId) { onTransactionImported(_ir, _h, _nodeId); });
+    std::random_device seed;
+    m_urng = std::mt19937_64(seed());
 }
 
 EthereumHost::~EthereumHost()
@@ -512,27 +514,29 @@ tuple<vector<shared_ptr<EthereumPeer>>, vector<shared_ptr<EthereumPeer>>, vector
     vector<shared_ptr<EthereumPeer>> chosen;
     vector<shared_ptr<EthereumPeer>> allowed;
     vector<shared_ptr<SessionFace>> sessions;
-
-    size_t peerCount = 0;
+    double percentDecimal = _percent / 100.0;
     foreachPeer([&](std::shared_ptr<EthereumPeer> _p)
     {
         if (_allow(_p.get()))
         {
-            allowed.push_back(_p);
             sessions.push_back(_p->session());
+            allowed.push_back(_p);
         }
-        ++peerCount;
         return true;
     });
 
-    size_t chosenSize = (peerCount * _percent + 99) / 100;
-    chosen.reserve(chosenSize);
-    for (unsigned i = chosenSize; i && allowed.size(); i--)
+    if (_percent == 0 || allowed.size() == 0)
     {
-        unsigned n = rand() % allowed.size();
-        chosen.push_back(std::move(allowed[n]));
-        allowed.erase(allowed.begin() + n);
+        return make_tuple(move(chosen), move(allowed), move(sessions));
     }
+
+    std::shuffle(allowed.begin(), allowed.end(), m_urng);
+
+   // Remove elements from the end of the shuffled allowed vector and move them to chosen.
+    size_t chosenSize = percentDecimal * allowed.size();
+    chosen.reserve(chosenSize);
+    std::move(allowed.begin() + chosenSize, allowed.end(), std::back_inserter(chosen));
+    allowed.erase(allowed.begin() + chosenSize, allowed.end());
     return make_tuple(move(chosen), move(allowed), move(sessions));
 }
 
--- a/libethereum/EthereumHost.h
+++ b/libethereum/EthereumHost.h
@@ -28,6 +28,7 @@
 #include <memory>
 #include <utility>
 #include <thread>
+#include <random>
 
 #include <libdevcore/Guards.h>
 #include <libdevcore/Worker.h>
@@ -139,6 +140,8 @@ private:
 
     std::shared_ptr<EthereumHostDataFace> m_hostData;
     std::shared_ptr<EthereumPeerObserverFace> m_peerObserver;
+
+    std::mt19937_64 m_urng; // Mersenne Twister psuedo-random number generator
 
     Logger m_logger{createLogger(VerbosityDebug, "host")};
 };


libethereum/EthereumHost.cpp
```diff
@@ -512,27 +512,25 @@ tuple<vector<shared_ptr<EthereumPeer>>, vector<shared_ptr<EthereumPeer>>, vector
...
-    size_t peerCount = 0;
+    double percentDecimal = _percent / 100.0;
     foreachPeer([&](std::shared_ptr<EthereumPeer> _p)
     {
         if (_allow(_p.get()))
         {
-            allowed.push_back(_p);
             sessions.push_back(_p->session());
+            // Random selection of peers according to percent requested
+            if(((double)rand() / RAND_MAX) < percentDecimal)
```
gumb0: Please use C++ random number tools from `<random>` instead of C stuff.

(I recommend this video to quit the bad habit of `rand()` https://channel9.msdn.com/Events/GoingNative/2013/rand-Considered-Harmful)

gumb0: You could also use `std::random_shuffle`

lhiny: Is it assumed that allowed's elements are in a certain order? Or can I just return a shuffled allowed vector?

gumb0: The goal is to randomly select them and order doesn't matter

libethereum/EthereumHost.cpp
```diff
@@ -512,26 +512,32 @@ tuple<vector<shared_ptr<EthereumPeer>>, vector<shared_ptr<EthereumPeer>>, vector
...
         return true;
     });
 
-    size_t chosenSize = (peerCount * _percent + 99) / 100;
+    if(_percent == 0 || allowed.size() == 0)
+    {
+        return make_tuple(move(chosen), move(allowed), move(sessions));
+    }
+
+    std::random_device randomSeed;
```
gumb0: Sorry for the long silence, could you please make `random_device` and `mt19937` objects members of the `EthereumHost` class instead of local variables (their initialization might be costly, we don't want to do that each time we call `randomSelection`)

libethereum/EthereumHost.h
```diff
@@ -140,6 +141,9 @@ class EthereumHost: public p2p::HostCapability<EthereumPeer>, Worker
     std::shared_ptr<EthereumHostDataFace> m_hostData;
     std::shared_ptr<EthereumPeerObserverFace> m_peerObserver;
 
+    std::shared_ptr<std::random_device> m_randomSeed;
```
gumb0: We don't to share them with anyone, so we don't need the overhead of `shared_ptr`, so make them just value members (`std::random_device m_randomDevice;`)

libethereum/EthereumHost.cpp
```diff
@@ -512,26 +514,30 @@ tuple<vector<shared_ptr<EthereumPeer>>, vector<shared_ptr<EthereumPeer>>, vector
...
-            allowed.push_back(_p);
             sessions.push_back(_p->session());
+            allowed.push_back(_p);
         }
-        ++peerCount;
         return true;
     });
 
-    size_t chosenSize = (peerCount * _percent + 99) / 100;
+    if(_percent == 0 || allowed.size() == 0)
```
gumb0: There should be whitespace after `if`
Better use clang-format as described in https://github.com/ethereum/aleth/blob/master/CONTRIBUTING.md

libethereum/EthereumHost.cpp
```diff
@@ -512,26 +514,30 @@ tuple<vector<shared_ptr<EthereumPeer>>, vector<shared_ptr<EthereumPeer>>, vector
...
+    size_t chosenSize = percentDecimal * allowed.size();
     chosen.reserve(chosenSize);
-    for (unsigned i = chosenSize; i && allowed.size(); i--)
+    for(size_t i = allowed.size() - 1; i >= allowed.size() - chosenSize; i--)
     {
-        unsigned n = rand() % allowed.size();
-        chosen.push_back(std::move(allowed[n]));
-        allowed.erase(allowed.begin() + n);
+        chosen.push_back(move(allowed[i]));
+        allowed.erase(allowed.end() - i);
```
gumb0: this loop body was supposed to move an element from `allowed` to `chosen`. You seem to first move value of `allowed[i]` but then erase `allowed[size - i]`, doesn't look correct

This whole loop is confusing, maybe just create new vectors like
```
vector<shared_ptr<EthereumPeer>> chosen(allowed.begin(), allowed.begin() + chosenSize);
vector<shared_ptr<EthereumPeer>> rest(allowed.begin() + chosenSize, allowed.end());
```
?

lhiny:
Yeah, I do. Sorry, this week has just been very stressful so I didn't have time to finish this

gumb0:
@lhiny ok, no worries

libethereum/EthereumHost.cpp
```diff
@@ -383,6 +383,7 @@ EthereumHost::EthereumHost(Host const& _host, BlockChain const& _ch, OverlayDB c
     m_peerObserver = make_shared<EthereumPeerObserver>(m_sync, m_tq);
     m_latestBlockSent = _ch.currentHash();
     m_tq.onImport([this](ImportResult _ir, h256 const& _h, h512 const& _nodeId) { onTransactionImported(_ir, _h, _nodeId); });
+    m_urng = std::mt19937(m_randomSeed());
```
chfast: Don't keep the random device around. Just init mt19937 here and forget about it.

libethereum/EthereumHost.h
```diff
@@ -140,6 +141,9 @@ class EthereumHost: public p2p::HostCapability<EthereumPeer>, Worker
     std::shared_ptr<EthereumHostDataFace> m_hostData;
     std::shared_ptr<EthereumPeerObserverFace> m_peerObserver;
 
+    std::random_device m_randomSeed;
+    std::mt19937 m_urng; // Mersenne Twister psuedo-random number generator
```
chfast: Shouldn't we use mt19937_64?

gumb0:
```suggestion
    std::mt19937_64 m_urng; // Mersenne Twister psuedo-random number generator
```

libethereum/EthereumHost.cpp
gumb0:
```suggestion
    m_urng = std::mt19937(std::random_device());
```

libethereum/EthereumHost.h
```diff
@@ -140,6 +141,9 @@ class EthereumHost: public p2p::HostCapability<EthereumPeer>, Worker
     std::shared_ptr<EthereumHostDataFace> m_hostData;
     std::shared_ptr<EthereumPeerObserverFace> m_peerObserver;
 
+    std::random_device m_randomSeed;
```
gumb0:
```suggestion
```

libethereum/EthereumHost.cpp
```diff
@@ -512,26 +513,31 @@ tuple<vector<shared_ptr<EthereumPeer>>, vector<shared_ptr<EthereumPeer>>, vector
...
+   // Remove elements from the end of the shuffled allowed vector and move them to chosen.
+    size_t chosenSize = percentDecimal * allowed.size();
     chosen.reserve(chosenSize);
-    for (unsigned i = chosenSize; i && allowed.size(); i--)
+    for (auto it = allowed.end(); it >= allowed.begin() + chosenSize; it--)
     {
-        unsigned n = rand() % allowed.size();
-        chosen.push_back(std::move(allowed[n]));
-        allowed.erase(allowed.begin() + n);
+        chosen.push_back(move(*it));
```
gumb0: On the first iteration you dereference `allowed.end()`, not correct at all

libethereum/EthereumHost.cpp
```diff
@@ -512,26 +513,31 @@ tuple<vector<shared_ptr<EthereumPeer>>, vector<shared_ptr<EthereumPeer>>, vector
...
+        return make_tuple(move(chosen), move(allowed), move(sessions));
+    }
+
+    std::shuffle(allowed.begin(), allowed.end(), m_urng);
+
+   // Remove elements from the end of the shuffled allowed vector and move them to chosen.
+    size_t chosenSize = percentDecimal * allowed.size();
     chosen.reserve(chosenSize);
-    for (unsigned i = chosenSize; i && allowed.size(); i--)
+    for (auto it = allowed.end(); it >= allowed.begin() + chosenSize; it--)
```
gumb0: Still I think you don't need a loop for this
```suggestion
    std::move(allowed.begin() + chosenSize, allowed.end(), std::back_inserter(chosen)); 
    allowed.erase(allowed.begin() + chosenSize, allowed.end());
```

gumb0:
Looks fine to me now, could you squash the commits please @lhiny ?

gumb0:
@chfast please review

PR #5293, opened 2018-09-30
Install leveldb using hunter package manager

speqtr:
Most of the dependencies for aleth are already installed using [hunter](https://github.com/ruslo/hunter/) package manager, which makes it possible to easily install leveldb using just a few [lines of code](https://docs.hunter.sh/en/latest/packages/pkg/leveldb.html). 

With this simple change in `libdevcore/CMakeLists.txt` file _Getting Started_ process become as simple as _clone, build and run_. (I believe that...) Simplifying _Getting Started_ process for new devs gives a huge benefit for any open source project. And it also makes project more _self-contained_.

NOTE: Changes tested with CLion IDE on macOS. If requested I can also test on Linux, Windows and other environments.

NOTE2 (by Captain Obvious): [RocksDB](https://docs.hunter.sh/en/latest/packages/pkg/rocksdb.html) is also available as a package for [hunter](https://github.com/ruslo/hunter/).


speqtr:
This change makes even more sense since @gumb0 is already using [hunter](https://github.com/ruslo/hunter/) to install RocksDB in #4844.

gumb0:
Unfortunately this will not work on Windows, because LevelDB package on hunter doesn't support windows (and won't because LevelDB itself doesn't) This is the reason we use custom built LevelDB instead of hunter package.

cc @chfast

chfast:
Leveldb is not available in Hunter. It can be added but will never work on Windows. There is no problem with using in from Hunter on Linux only.

speqtr:
Thank you for your feedback!

Can we use hunter for leveldb on Linux/macOS and fallback to previous code on Windows (like in pseudo code example below)?
```
if (not Windows) {
  // use hunter for leveldb
} else {
  // fallback to manually build leveldb
}
```

speqtr:
And there are also two other options to consider:
1. Use bitcoin-leveldb with custom configuration for hunter;
2. Create a patch from the following [PR](https://github.com/google/leveldb/pull/544) and install it using custom configuration for hunter.

cc @gumb0 @chfast

chfast:
Sorry, I did not realized this is a PR. 

The solution with "if" in CMake looks good.

We don't want to use Bitcoin fork, because it diverged a lot.

Using an unmerged branch is not an option either. Compilation is not so big issue (they accept changes fixing compilation with msvc) be leveldb explicitly refuses to merge code sitting like supporting Windows. 

For this reason we want to switch to rocksdb by default.

speqtr:
In a few additional commits to this PR I'm planning to:
1. Add solution with "if" in related `libdevcore/CMakeLists.txt` file;
2. Update CI scripts (remove manual leveldb installation/compilation on Linux and macOS);
3. Update README according to the changes (leveldb installation instaructions for Linux and macOS will be irrelevant).

Does this sound like a good plan?

cc @chfast

chfast:
Yes. Perfect

speqtr:
Tests on all platforms failed after rebasing on top of latest changes from `master` branch (only `test/jsontests` submodule was updated to newer version). Before this update all tests were green, but codecov tool failed due to decreased coverage by `0.42%`.

gumb0:
@winsvega Do you know what could be the reason for test hash mismatch error in this branch?

winsvega:
the test repo uses develop branch as the latest. 
why do we change main branch name all the time?  

for the last three years go-ethereum  changed it from master to develop then to master again. For as long as I remember I was always using develop in the test repo. because cpp-ethereum also named it develop.

gumb0:
@speqtr Somehow your branch now points to the very old revision of tests instead of the one current aleth@master is using

chfast:
> the test repo uses develop branch as the latest.

I can ask question like: why is tests repo using "develop" branch? What are developing there?

winsvega:
it was like this all the time. just mimiced that from webthree-umbrella

speqtr:
I rebased `test/jsontests` submodule on top of the latest changes from `ethereum/tests` repo (`origin/develop` branch).

speqtr:
I have two additional questions:
1. Why rebasing on top of `aleth/master` switches `test/jsontests` submodule to the incorrect revision?
2. What about installation of `*-dev` packages for LevelDB inside some of the Dockerfile's in `test/jsontests` submodule? Should I open another PR for the changes under `test/jsontests`?

speqtr:
@gumb0 Cool! But now there are 82 failed test cases on Linux. Why did the tests start to fail on Linux? And what is the best approach to debug this failures? Maybe we just need to rerun the tests for Linux?

gumb0:
I restarted it and it still fails.
Did you try running `testeth` locally?

speqtr:
Yes, I did run `testeth` successfully on local machine.

(Also I think that this is worth mentioning) In this PR commit `1de3499` passed all tests successfully. After that I only removed one line from `snapcraft.yaml` and updated `test/jsontests` module.

speqtr:
In file `test/jsontests/ansible/test-files/docker-cpp/Dockerfile` package `libleveldb-dev` is installed using `apt-get`. The same is true for `test/jsontests/ansible/test-files/docker-cppjit/Dockerfile`. May this somehow affect tests on Linux platform? (since we are now installing `leveldb` from `hunter` package manager on Linux and macOS).

chfast:
> Before commit `58f1a23` all tests were green, but codecov tool failed due to decreased coverage by `0.42%`.

Codecov is really flaky and seems it is their fault. I tried to debug it recently but it's complicated...



>     1. Why rebasing on top of `aleth/master` switches `test/jsontests` submodule to the incorrect revision?

It should not. The submodule change requires an explicit change in a commit. You probably committed that by accident.

> Cool! But now there are 82 failed test cases on Linux. Why did the tests start to fail on Linux? And what is the best approach to debug this failures? Maybe we just need to rerun the tests for Linux?

It was failing for me too. The `leveldb-dev` is installed on Circle CI so the failures might be because we are including system headers or linking wrong shared library at runtime. I will try to check that locally.

speqtr:
I will also create a Linux VM for testing this issue on my local machine.

speqtr:
**UPDATE**

I created a clean VM using Vagrant and VirtualBox. `Vagrantfile` content:
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure("2") do |config|
    config.vm.define "linux" do |linux|
        linux.vm.box = "bento/ubuntu-18.04"
        linux.vm.box_check_update = false
        linux.vm.hostname = "linux"
        linux.vm.synced_folder ".", "/tmp/synced_folder"

        linux.vm.provision "shell", inline: <<-SHELL
            export DEBIAN_FRONTEND=noninteractive
            sudo apt-get -yq update
            sudo apt-get -yq install build-essential cmake
        SHELL
    end

    config.vm.provider "virtualbox" do |v|
        v.gui = false
        v.cpus = 2
        v.memory = "4096"
    end
end
```
I built project and ran `testeth` multiple times. It always exited with `*** No errors detected` message and the end.

Can we just remove `leveldb-dev` package from Circle CI? Or should I apply some additional changes to the project sources?

cc @gumb0 @chfast @winsvega

speqtr:
> Also run ./testeth -- --all

OK, will do.

speqtr:
This is just a hunch, but... is it possible that `aleth/cmake/FindLevelDB.cmake` is breaking a Linux build somehow? Maybe we should disable this script for Linux and macOS platforms?

cc @chfast @gumb0

speqtr:
@gumb0 could you please rerun the CI build for Windows? 

[Latest CI build on AppVeyor](https://ci.appveyor.com/project/ethereum/cpp-ethereum/builds/19914576) failed with the following error message: `Build execution time has reached the maximum allowed time for your plan (60 minutes)`.

speqtr:
ping @chfast @gumb0 @halfalicious @winsvega

speqtr:
All the checks are green. Can we merge this PR now?

cc @chfast @gumb0 @halfalicious @winsvega

chfast:
I believe the test are green now because they don't use leveldb any more ("memory" db instead). I have investigate why when running tests in parallel we get failures.

speqtr:
@chfast Is there any way I can help you with resolving an issue with tests?

speqtr:
This PR now has one conflicting file with `origin/master`. I think that a proper fix for `libdevcore/CMakeLists.txt` should look something like the code below:
```
if(USE_LEVELDB AND MSVC)
    find_package(LevelDB)
    target_include_directories(devcore SYSTEM PUBLIC ${LEVELDB_INCLUDE_DIRS})
    target_link_libraries(devcore PRIVATE ${LEVELDB_LIBRARIES})
elseif(USE_LEVELDB AND NOT MSVC)
    hunter_add_package(leveldb)
    find_package(leveldb CONFIG REQUIRED)
    target_link_libraries(devcore PRIVATE leveldb::leveldb)
else()
    hunter_add_package(rocksdb)
    find_package(RocksDB CONFIG REQUIRED)
    target_link_libraries(devcore PRIVATE RocksDB::rocksdb)
endif()
```

cc @chfast @gumb0

chfast:
I rebased the PR.
I will do one more test.

PR #5295, opened 2018-10-02
Charge for hashing init code in CREATE2

gumb0:
Changing the gas cost of CREATE2 according to the latest changes in spec https://github.com/ethereum/EIPs/pull/1375

cc @winsvega

### DIFF
--- a/libaleth-interpreter/VMCalls.cpp
+++ b/libaleth-interpreter/VMCalls.cpp
@@ -116,7 +116,11 @@ void VM::caseCreate()
 
     u256 salt;
     if (m_OP == Instruction::CREATE2)
+    {
         salt = m_SP[3];
+        // charge for hashing initCode = GSHA3WORD * ceil(len(init_code) / 32)
+        m_runGas += toInt63((u512{initSize} + 31) / 32 * uint64_t{VMSchedule::sha3WordGas});
+    }
 
     updateMem(memNeed(initOff, initSize));
     updateIOGas();
--- a/libevm/LegacyVMCalls.cpp
+++ b/libevm/LegacyVMCalls.cpp
@@ -127,7 +127,11 @@ void LegacyVM::caseCreate()
 
     u256 salt;
     if (m_OP == Instruction::CREATE2)
+    {
         salt = m_SP[3];
+        // charge for hashing initCode = GSHA3WORD * ceil(len(init_code) / 32)
+        m_runGas += toInt63((u512{initSize} + 31) / 32 * m_schedule->sha3WordGas);
+    }
 
     updateMem(memNeed(initOff, initSize));
     updateIOGas();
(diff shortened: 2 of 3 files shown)


chfast:
It cannot find sha3WordGas symbol :/

gumb0:
Linker can't find the symbol, but it's `constexpr`, weird. And it's fine with `VMSchedule::createGas` in the same function 😕

libaleth-interpreter/VMCalls.cpp
```diff
@@ -116,7 +116,11 @@ void VM::caseCreate()
 
     u256 salt;
     if (m_OP == Instruction::CREATE2)
+    {
         salt = m_SP[3];
+        // charge for hashing initCode = GSHA3WORD * ceil(len(init_code) / 32)
+        m_runGas += toInt63((static_cast<u512>(initSize) + 31) / 32 * uint64_t(VMSchedule::sha3WordGas));
```
chfast: Write it as `u512(initSize)` because there is not data loss here.

gumb0:
So `uint64_t{VMSchedule::sha3WordGas}` fixes the linker error.

Another option is to add definitions of `VMSchedule` static members to cpp file like
```
constexpr int64_t VMSchedule::sha3WordGas;
...
```

chfast:
I don't care which workaround is used, but I've found this for `SHA3` instructions:
```
        CASE(SHA3)
        {
            ON_OP();
            constexpr int64_t sha3Gas = VMSchedule::sha3Gas;
            constexpr int64_t sha3WordGas = VMSchedule::sha3WordGas;
            m_runGas = toInt63(sha3Gas + (u512(m_SP[1]) + 31) / 32 * sha3WordGas);
```

gumb0:
I see, this is third kind of workaround.

So as I understand `constexpr int64_t sha3WordGas = VMSchedule::sha3WordGas;` works and `(u512(m_SP[1]) + 31) / 32 * VMSchedule::sha3WordGas` doesn't, because the second one probably tries to pass `VMSchedule::sha3WordGas` by reference to `boost::multiprecision::operator*` and creating a reference requires definition outside of class in C++11

chfast:
I don't know why it work and I don't want to know. I think it was me who add this but I have not investigated the reasons.

libevm/LegacyVMCalls.cpp
```diff
@@ -127,7 +127,11 @@ void LegacyVM::caseCreate()
 
     u256 salt;
     if (m_OP == Instruction::CREATE2)
+    {
         salt = m_SP[3];
+        // charge for hashing initCode = GSHA3WORD * ceil(len(init_code) / 32)
+        m_runGas += toInt63((static_cast<u512>(initSize) + 31) / 32 * m_schedule->sha3WordGas);
```
chfast: Same `static_cast` -> `u512{}` change here.

gumb0: done

PR #5298, opened 2018-10-04
Circle CI: Decouple docker nightly build from "all tests" job

chfast:
@winsvega We have `testeth --all` tests for some time, but because nobody is paying attention to them they are also failing for some time. https://circleci.com/gh/ethereum/aleth/6946

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -319,6 +319,4 @@ workflows:
             - master
     jobs:
       - linux-gcc6-nightly
-      - docker-aleth-nightly:
-          requires:
-            - linux-gcc6-nightly
+      - docker-aleth-nightly


winsvega:
we used to run a server with cron job and mail notification. should we  bring it back again?

gumb0:
@winsvega CircleCI is better for this (minus email notifications)

chfast:
To be clear:
1. Currently, there is no way to configure Circle CI to send any notification about failing job. You can only enabled email notification about every Aleth build in your personal account settings. I did this, but expect to get too much spam.
2. There is no way to get a badge for this nightly build (because it runs on master branch). But even if there is a way, I'm not sure anyone here is regularly checking the build status in the README.

PR #5299, opened 2018-10-05
[READY] make calculateDifficulty separate

winsvega:
NoProof support calculateDifficulty
Ethash support calculateDifficulty

### DIFF
--- a/libethcore/SealEngine.cpp
+++ b/libethcore/SealEngine.cpp
@@ -19,14 +19,22 @@
 #include "TransactionBase.h"
 
 using namespace std;
-using namespace dev;
-using namespace eth;
-
+namespace dev
+{
+namespace eth
+{
 SealEngineRegistrar* SealEngineRegistrar::s_this = nullptr;
 
 void NoProof::init()
 {
     ETH_REGISTER_SEAL_ENGINE(NoProof);
+}
+
+void NoProof::populateFromParent(BlockHeader& _bi, BlockHeader const& _parent) const
+{
+    SealEngineFace::populateFromParent(_bi, _parent);
+    _bi.setDifficulty(calculateEthashDifficulty(chainParams(), _bi, _parent));
+    _bi.setGasLimit(calculateGasLimit(chainParams(), _bi));
 }
 
 void NoProof::generateSeal(BlockHeader const& _bi)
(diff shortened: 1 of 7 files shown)


gumb0:
`BlockChainMainNetworkSuite/Mining_5_BlockFutureTime` fails everywhere

libethcore/SealEngine.h
```diff
@@ -138,7 +138,13 @@ class NoProof: public eth::SealEngineBase
     static std::string name() { return "NoProof"; }
     static void init();
     void generateSeal(BlockHeader const& _bi) override;
+    void populateFromParent(BlockHeader& _bi, BlockHeader const& _parent) const override;
 };
 
+u256 calculateDifficulty(
```
gumb0: Call it `calculateEthashDifficulty`

libethcore/SealEngine.h
```diff
@@ -138,7 +138,13 @@ class NoProof: public eth::SealEngineBase
...
+    SealEngineFace const& sealEngine, BlockHeader const& _bi, BlockHeader const& _parent);
```
gumb0: you don't need the whole `SealEngine`, pass `ChainParams` instead

libethcore/SealEngine.h
winsvega: Sounds weird calling EthashDifficulty from NoProof

libethcore/SealEngine.cpp
```diff
@@ -151,3 +158,80 @@ u256 SealEngineBase::blockReward(u256 const& _blockNumber) const
     EVMSchedule const& schedule{evmSchedule(_blockNumber)};
     return chainParams().blockReward(schedule);
 }
+
+namespace dev
```
gumb0: Better put everything in the file into these and remove `using namespace dev` `using namespace eth`

libethcore/SealEngine.h
gumb0: right, but that's what we need to do, I think it better expresses what's happening

winsvega: is it also calculateEthashGasPrice ?

gumb0: I'd leave `calculateGasLimit`, it doesn't seem to me to be inherent to Ethash
(could also go to the base class, but this global function is fine, too)

libethcore/SealEngine.h
```diff
@@ -138,7 +138,13 @@ class NoProof: public eth::SealEngineBase
     static std::string name() { return "NoProof"; }
     static void init();
     void generateSeal(BlockHeader const& _bi) override;
+    void populateFromParent(BlockHeader& _bi, BlockHeader const& _parent) const override;
 };
 
+u256 calculateEthashDifficulty(
+    ChainOperationParams const& chainParams, BlockHeader const& _bi, BlockHeader const& _parent);
```
gumb0: `_chainParams`

libethcore/SealEngine.h
```diff
@@ -138,7 +138,13 @@ class NoProof: public eth::SealEngineBase
...
+
+u256 calculateGasLimit(ChainOperationParams const& chainParams, BlockHeader const& _bi,
```
gumb0: `_chainParams`

test/unittests/libethashseal/EthashTest.cpp
```diff
@@ -55,7 +55,8 @@ BOOST_AUTO_TEST_CASE(calculateDifficultyByzantiumWithoutUncles)
     header.setNumber(0x2001);
     header.setTimestamp(130);
 
-    BOOST_CHECK_EQUAL(ethash.calculateDifficulty(header, parentHeader), 999024);
```
gumb0: I'd leave this without a change, the tests here check `Ethash` class, we'd like to be sure that it calls the correct calculation

winsvega: how?
we've just removed calculateDifficulty from ethash into a separate function

winsvega: the test doesnt fail. that should be enough. it actually tests that function work as it was before

gumb0: Ah right, for a moment it seemed to me that `Ethash::calculateDifficulty` still existed. Ok then, this is fine

PR #5302, opened 2018-10-06
Added "Running Aleth" section to README.md

### LINKED ISSUE #5301: Create "running Aleth" documentation
I think it would be great if we had some "running Aleth" documentation which covers how to run Aleth in various configurations (e.g. on mainnet, the various testnets, configuring private networks) so that new users can easily and quickly learn how to get up and running. I think it would make sense to have both "quickstart"-style documentation and more detailed deep-dive-style docs. 

I took a quick look at Geth's documentation in their GitHub repo and I think it's a great bar to try and hit: https://github.com/ethereum/go-ethereum/blob/master/README.md

Note that there is already documentation in the [Aleth doc folder](https://github.com/ethereum/aleth/tree/master/doc) - perhaps we can reuse some of this and link to it from the main [readme](https://github.com/ethereum/aleth/blob/master/README.md)?

halfalicious:
Fix #5301 

Added output of "aleth -h" to a new "Running" section in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,9 +141,12 @@ thoroughly before making alterations to the code base.
 
 All development goes in develop branch.
 
-## Running
+## Usage
+*Note: The following is the output of ```aleth.exe -h [--help]```*
 
 ```
+NAME:
+   aleth 1.4.0
 USAGE:
    aleth [options]
 
@@ -182,6 +185,11 @@ CLIENT TRANSACTING:
   --bid <wei>            Set the bid gas price to pay for transactions (default: 20000000000)
   --unsafe-transactions  Allow all transactions to proceed without verification; EXTREMELY UNSAFE
                          
+CLIENT MINING:
+  -a [ --address ] <addr>         Set the author (mining payout) address (default: auto)
+  -m [ --mining ] <on/off/number> Enable mining; optionally for a specified number of blocks (default: off)
+  --extra-data arg                Set extra data for the sealed blocks
+                                  
 CLIENT NETWORKING:
   -b [ --bootstrap ]              Connect to the default Ethereum peer servers (default unless --no-discovery used)
   --no-bootstrap                  Do not connect to the default Ethereum peer servers (default only when --no-discovery is used)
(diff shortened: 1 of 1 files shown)


halfalicious:
Note that this doesn't completely fix #5301 , I think it would be good to have additional details for some of the options for running Aleth e.g. setting up a private network.

gumb0:
I'm not sure if this is very helpful at all. Maybe just a sentence like "see `aleth --help` output for usage instructions" would be enough?

halfalicious:
Thanks for the feedback guys! So what motivated me to create this PR in the first place was that I found it easier to read through the help output in a browser rather than in a console window, here's what I see in the default-size command prompt window on my system (note the wrapped-around lines)

aleth_help_output

I also found it useful to have something that I knew would always be in the same place that I could quickly reference without having to generate new console output each time. I know that I could just save the console output to a text file and look at that, but I figured that if I was going to do that I might as well put something in the readme since then others could benefit from it as well. I also think this usage info could also be useful to people who want to check what Aleth supports without having to download or build their own binaries just to run `aleth -h`

Finally, I thought that this usage section would be just a starting point for something more detailed e.g. we could include additional details and usage examples on some of the more complicated commands such as db import/export. 

I ultimately don't feel very strongly about including usage info in the readme and can understand how it can be seen as unncessary, so if you'd prefer me to leave it out just let me know and I can close this PR :)

README.md
```diff
@@ -141,7 +141,110 @@ thoroughly before making alterations to the code base.
 
 All development goes in develop branch.
 
+## Usage
+*Note: The following is the output of ```aleth.exe -h [--help]```*
```
chfast: For me this can stay, but include version number of aleth (preferably 1.4.0).

halfalicious:
Updated output to Aleth 1.4.0 and squashed commits

gumb0:
@chfast Why 1.4.0? So this is now outdated (doesn't have `--db`)
Do you think we should update once when we prepare stable release?

halfalicious:
I think we should keep this in sync with stable releases since I expect that's what most people will be running, we can update this once we release 1.5.0

PR #5305, opened 2018-10-08
libp2p refactoring - get rid of PeerCapability and its descendants

gumb0:
This is mostly moving things around with the goal of handling all network messages in `HostCapability`-derived classes instead of `PeerCapability`-derived classes.

## Overview of what changed
Before:
- `Host` has a list of `shared_ptr<HostCapabilityFace>`s - capabilities supported by the client
- The client implements `HostCapabilityFace` and `PeerCapability` interfaces to support some devp2p subprotocol
- When the new peer connects, libp2p asks `HostCapabilityFace` implementation to create a new instance of `PeerCapability` implementation class
- Each `Session` (corresponding to each peer we're connected to) has a different list of `shared_ptr<PeerCapability>`s - subprotocols through which we communicate with this peer.

After:
- `Host` has a list of `shared_ptr<HostCapabilityFace>`s - capabilities supported by the client
- The client implements `HostCapabilityFace` interface to support some devp2p subprotocol; this implementation uses `CapabilityHostFace` interface to interact with libp2p library
- When the new peer connects, libp2p calls `HostCapabilityFace::onConnect()`
- Each `Session` has a list of `shared_ptr<HostCapability>`s.

## More details of what went where
- `PeerCapability` data members (offset and enabled flag) are now stored in `Session`;
- `SessionFace` is not a part of public interface anymore, capability implementations use the methods of `CapabilityHostFace` instead of accessing session directly;
- `EthereumPeer` instances are created by `EthereumHost` and stored in the map `EthereumHost::m_peers`;
- Handling messages went from `EthereumPeer::interpretCapabilityPacket` to `EthereumHost::interpretCapabilityPacket`
- `BlockChainSync` keeps track of connected peers with sets of `NodeID`s instead of `weak_ptr<EthereumPeer>`s;
- `WarpPeerCapability` data members are now in `WarpPeerStatus` struct, stored in the map `WarpHostCapability::m_peers`;
- Methods for sending messages `WarpPeerCapability::request...` are now in `WarpHostCapability`;
- Handling messages went from `WarpPeerCapability::interpretCapabilityPacket` to `WarpHostCapability::interpretCapabilityPacket`;
- In general, instead of passing around `EtherereumPeer`/ `WarpPeerCapability` peers are referred to by their `NodeID` only;
- Iterating over peers ordered by rating moved from `EthereumHost::foreachPeer` to `Host::forEachPeer`, now accessible to any capability implementation.

## Next steps
In the next simple PR I want to rename

- `HostCapabilityFace` -> `CapabilityFace` (then there'll be less confusion with `CapabilityHostFace`)
- `EthereumHost` -> `EthereumCapability`
- `WarpHostCapability` -> `WarpCapability`

In the next not-so-simple PR I have to rethink the thread-safety of `EthereumHost` and `WarpHostCapability` classes, because now it actually got worse (`EthereumHost::m_peers` and `WarpHostCapability::m_peers` are containers accessed from several threads.)
I want to get rid of additional threads created by these classes (i.e. don't derive them from `Worker`) and instead introduce the method for scheduling any callback execution from network handling thread. This way all the work in these classes, including "background" work, will be done from the single thread.

### DIFF
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -1041,7 +1041,7 @@ void Host::forEachPeer(
             std::vector<CapDesc> capabilities = s->capabilities();
             for (auto const& cap : capabilities)
                 if (cap.first == _capabilityName)
-                    sessions.push_back(s);
+                    sessions.emplace_back(std::move(s));
         }
 
     // order peers by rating, connection age
@@ -1053,7 +1053,7 @@ void Host::forEachPeer(
     };
     std::sort(sessions.begin(), sessions.end(), sessionLess);
 
-    for (auto s : sessions)
+    for (auto const& s : sessions)
         if (!_f(s->id()))
             return;
 }


gumb0:
Also need to think how to improve the test coverage, it seems very poor.

gumb0:
Added to the top comment a short intro to how it worked vs how it works now.

libethereum/BlockChainSync.cpp
```diff
@@ -819,18 +812,18 @@ void BlockChainSync::onPeerNewBlock(std::shared_ptr<EthereumPeer> _peer, RLP con
     case ImportResult::FutureTimeUnknown:
     case ImportResult::UnknownParent:
     {
-        _peer->m_unknownNewBlocks++;
-        if (_peer->m_unknownNewBlocks > c_maxPeerUknownNewBlocks)
+        m_host.incrementPeerUnknownNewBlocks(_peerID);
+        if (m_host.peerStatus(_peerID).m_unknownNewBlocks > c_maxPeerUknownNewBlocks)
```
chfast: If used this way, the `m_` prefix should be removed from `m_unknownNewBlocks`.

libethereum/BlockChainSync.cpp
```diff
@@ -819,18 +812,18 @@ void BlockChainSync::onPeerNewBlock(std::shared_ptr<EthereumPeer> _peer, RLP con
     case ImportResult::FutureTimeUnknown:
     case ImportResult::UnknownParent:
     {
-        _peer->m_unknownNewBlocks++;
-        if (_peer->m_unknownNewBlocks > c_maxPeerUknownNewBlocks)
+        m_host.incrementPeerUnknownNewBlocks(_peerID);
```
chfast: This is quite different style than the next line where a filed of peer status struct is accessed directly.

libethereum/BlockChainSync.cpp
```diff
@@ -908,9 +902,8 @@ void BlockChainSync::onPeerNewHashes(std::shared_ptr<EthereumPeer> _peer, std::v
     for (auto const& p: _hashes)
     {
         h256 const& h = p.first;
-        _peer->addRating(1);
-        DEV_GUARDED(_peer->x_knownBlocks)
-            _peer->m_knownBlocks.insert(h);
+        m_host.capabilityHost().addRating(_peerID, 1);
```
chfast: Maybe rename `addRating` to `updateRating`?

libethereum/BlockChainSync.cpp
```diff
@@ -935,7 +928,7 @@ void BlockChainSync::onPeerNewHashes(std::shared_ptr<EthereumPeer> _peer, std::v
     if (unknowns > 0)
     {
         LOG(m_loggerDetail) << "Not syncing and new block hash discovered: syncing.";
-        syncPeer(_peer, true);
+        syncPeer(_peerID, true);
```
chfast: (finished here)

libethereum/BlockChainSync.cpp
gumb0: Good point, I think I'll make a class out of `EthereumPeerStatus`

halfalicious:
Should we start changing all instances of "cpp-ethereum" to "Aleth" in the licensing comments?

halfalicious:
> This is mostly moving things around with the goal of handling all network messages in `HostCapability`-derived classes instead of `PeerCapability`-derived classes.

Naive question from someone who isn't familiar with these classes (yet) - what is the benefit of this work?

libethereum/EthereumHost.h
```diff
@@ -51,20 +51,95 @@ class TransactionQueue;
...
+struct EthereumPeerStatus
+{
+    /// What, if anything, we last asked the other peer for.
+    Asking m_asking = Asking::Nothing;
+    /// When we asked for it. Allows a time out.
+    std::atomic<time_t> m_lastAsk;
+    /// Peer's protocol version.
+    unsigned m_protocolVersion;
+    /// Peer's network id.
+    u256 m_networkId;
```
chfast: What is this?

libethereum/EthereumHost.h
```diff
@@ -51,20 +51,95 @@ class TransactionQueue;
...
+    /// These are determined through either a Status message or from NewBlock.
+    /// Peer's latest block's hash that we know about or default null value if no need to sync.
+    h256 m_latestHash;
+    /// Peer's latest block's total difficulty.
+    u256 m_totalDifficulty;
+    h256 m_genesisHash;            ///< Peer's genesis hash
+    /// Protocol version this peer supports received as capability
+    u256 m_peerCapabilityVersion;
```
chfast: Probably can be smaller type.

libp2p/Host.cpp
```diff
@@ -238,26 +240,26 @@ bool Host::isRequiredPeer(NodeID const& _id) const
 void Host::startPeerSession(Public const& _id, RLP const& _rlp, unique_ptr<RLPXFrameCoder>&& _io, std::shared_ptr<RLPXSocket> const& _s)
 {
     // session maybe ingress or egress so m_peers and node table entries may not exist
-    shared_ptr<Peer> p;
+    shared_ptr<Peer> peer;
     DEV_RECURSIVE_GUARDED(x_sessions)
     {
         if (m_peers.count(_id))
-            p = m_peers[_id];
+            peer = m_peers[_id];
```
chfast: Can't you use find() here?

libp2p/Host.cpp
```diff
@@ -315,33 +317,39 @@ void Host::startPeerSession(Public const& _id, RLP const& _rlp, unique_ptr<RLPXF
...
+        m_sessions[_id] = session;
+
         unsigned offset = (unsigned)UserPacket;
 
         // todo: mutex Session::m_capabilities and move for(:caps) out of mutex.
-        for (auto const& i: caps)
+        for (auto const& capDesc : caps)
         {
-            auto pcap = m_capabilities[i];
+            auto pcap = m_capabilities[capDesc];
```
chfast: This will fill the map with unknown capabilities. Use find().

libp2p/Host.cpp
```diff
@@ -1020,14 +1028,30 @@ bool Host::addNodeToNodeTable(Node const& _node, NodeTable::NodeRelation _relati
...
     for (auto const& i : m_sessions)
         if (std::shared_ptr<SessionFace> s = i.second.lock())
-            if (s->capabilities().count(std::make_pair(_name, _version)))
-                ret.push_back(make_pair(s, s->peer()));
-    return ret;
+        {
+            std::vector<CapDesc> capabilities = s->capabilities();
+            for (auto const& cap : capabilities)
+                if (cap.first == _capabilityName)
+                    sessions.push_back(s);
```
```suggestion
                    sessions.emplace_back(std::move(s));
```

libp2p/Host.cpp
```diff
@@ -1020,14 +1028,30 @@ bool Host::addNodeToNodeTable(Node const& _node, NodeTable::NodeRelation _relati
...
+    // order peers by rating, connection age
+    auto sessionLess = [](std::shared_ptr<SessionFace> const& _left,
+                           std::shared_ptr<SessionFace> const& _right) {
+        return _left->rating() == _right->rating() ?
+                   _left->connectionTime() < _right->connectionTime() :
+                   _left->rating() > _right->rating();
+    };
+    std::sort(sessions.begin(), sessions.end(), sessionLess);
+
+    for (auto s : sessions)
```
```suggestion
    for (auto const& s : sessions)
```

gumb0:
> Naive question from someone who isn't familiar with these classes (yet) - what is the benefit of this work?

@halfalicious The benefit is making the library interface cleaner and simpler (with a prospect of potentially making it a stand-alone library); which should result in making it easier to reason about the code of both the library itself and the code using it; and make further modifications easier.

libethereum/EthereumHost.h
gumb0: Network ID that the peer sends us in eth Status message

libethereum/BlockChainSync.cpp
gumb0: Unfortunately we don't have any convention about capitalizing abbreviations in the names, so both ways can be found in the code

libethereum/BlockChainSync.cpp
gumb0: done

libethereum/EthereumHost.h
gumb0: In general it comes from devp2p handshake protocol where capability version can be any positive integer, so probably I'd keep it as `u256` in libp2p interface, but I guess for eth implementation we can narrow it here.

gumb0: Actually there's no much sense it saving it (it is always 63 now), so I removed it.

libp2p/Host.cpp
gumb0: fixed

gumb0:
@chfast all issues addressed now

gumb0:
Rebased.

PR #5310, opened 2018-10-10
Trying to fix running tests on AppVeyor

gumb0:


scripts/tests.bat
```diff
@@ -32,9 +32,8 @@ if "%TESTS%"=="On" (
 
     REM Copy the DLLs into the test directory which need to be able to run.
     cd %APPVEYOR_BUILD_FOLDER%\build\test\%CONFIGURATION%
-    copy %APPVEYOR_BUILD_FOLDER%\build\evmjit\libevmjit\%CONFIGURATION%\evmjit.dll .
     copy "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\redist\x86\Microsoft.VC140.CRT\msvc*.dll" .
-    copy %ETHEREUM_DEPS_PATH%\win64\bin\OpenCl.dll .
+    copy %APPVEYOR_BUILD_FOLDER%\build\libaleth-interpreter\%CONFIGURATION%\aleth-interpreter.dll .
```
chfast: I don't think we run any tests for this DLL. I'm surprised this is even built.

gumb0: For some reason we build with `-DALETH_INTERPRETER_SHARED=ON`

Should we better statically link instead?

chfast: The name is confusing, but `ALETH_INTERPRETER_SHARED=ON` build DLL additionally. Still aleth-interpreter is statically linked to aleth / testeth. This build tests if the DLL can be linked but no test are executed.

To test statically linked interpreter you need `testeth -- --vm interpreter`.
To test DLL you need `testeth -- --vm path/aleth-interpreter.dll`.

chfast: We don't run tests to safe time.

chfast: I'm suggesting to remove this line, because in case we stop building interpreter.dll the script will stop working again.

chfast: I'm suggesting to remove this line, because in case we stop building interpreter.dll the script will stop working again.

gumb0: It won't stop because of the failed copy, it doesn't stop now when it can't copy `OpenCl.dll`

gumb0:
@chfast I removed building interpreter dll

appveyor.yml
```diff
@@ -46,7 +46,7 @@ install:
 before_build: |
   if not exist build mkdir build
   cd build
-  cmake -G "Visual Studio 14 2015 Win64" .. -DALETH_INTERPRETER_SHARED=ON
```
chfast: Ok, but this should stay :)

gumb0: Then testeth doesn't start because dll is required. Do you mean we should find the way to fix this?

chfast: It should not be required, but maybe there is a build issue. The static build produces `aleth-interpreter.lib` and the DLL build produces `aleth-interpreter.lib` and `aleth-interpreter.dll`. Maybe the second build overwrites the static build. This should be handled separately.

PR #5311, opened 2018-10-10
[READY] difficulty verification on NoProof

winsvega:


test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -771,12 +774,13 @@ void overwriteUncleHeaderForTest(mObject& uncleHeaderObj, TestBlock& uncle, std:
     BlockHeader uncleHeader;
     if (uncleHeaderObj.count("populateFromBlock"))
     {
-        uncleHeader.setTimestamp(time(0));
         size_t number = (size_t)toU256(uncleHeaderObj.at("populateFromBlock"));
         uncleHeaderObj.erase("populateFromBlock");
         if (number < importedBlocks.size())
         {
+            uncleHeader.setTimestamp(importedBlocks.at(number).blockHeader().timestamp() + 1);
```
gumb0: maybe `max(utcTime(), importedBlocks.at(number).blockHeader().timestamp() + 1)` could be closer to real world somehow (more similar to Ethash mining)

gumb0: Well `sleep_for` can't guarantee to sleep for exactly 1 second

PR #5314, opened 2018-10-12
Fix BLOCKHASH cost in Constantinople

gumb0:
Fixes https://github.com/ethereum/aleth/issues/5313


gumb0:
@winsvega Please help with updating tests here.

These needs to be regenerated: https://gist.github.com/holiman/0a9a057c90079b70a47d7f248a8fb75a#gistcomment-2730618

PR #5316, opened 2018-10-15
Constantinople SSTORE refund fix

chfast:
cc @sorpaas.

There are 2 commits: the first one cleans up some code to make the fix in the second commit more obvious.

### DIFF
--- a/libevm/ExtVMFace.h
+++ b/libevm/ExtVMFace.h
@@ -90,7 +90,7 @@ struct SubState
 {
     std::set<Address> suicides;  ///< Any accounts that have suicided.
     LogEntries logs;             ///< Any logs.
-    u256 refunds;                ///< Refund counter for storage changes.
+    int64_t refunds = 0;         ///< Refund counter for storage changes.
 
     SubState& operator+=(SubState const& _s)
     {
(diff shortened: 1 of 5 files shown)


libevm/ExtVMFace.h
```diff
@@ -90,7 +90,7 @@ struct SubState
 {
     std::set<Address> suicides;  ///< Any accounts that have suicided.
     LogEntries logs;             ///< Any logs.
-    u256 refunds;                ///< Refund counter for storage changes.
+    int64_t refunds = 0;         ///< Refund counter for storage changes.
```
gumb0: Wouldn't it be safer to use signed bigint?

chfast: Yes, but if 63 bits are not enough I'd like to constructs a test for this case.

sorpaas: Yeah. For this counter, geth uses `u64`, and parity uses `i128` (which is basically signed `u64`). If we make every opcode give maximum amount of refund as we currently have, it would still take decades of execution for that to ever overflow.

chfast: Thanks for clarification @sorpaas.

PR #5317, opened 2018-10-15
evm: Change Executive::m_gas to int64_t

chfast:


libethereum/Executive.h
```diff
@@ -200,7 +200,11 @@ class Executive
     unsigned m_depth = 0;				///< The context's call-depth.
     TransactionException m_excepted = TransactionException::None;	///< Details if the VM's execution resulted in an exception.
     int64_t m_baseGasRequired;			///< The base amount of gas requried for executing this transaction.
-    u256 m_gas = 0;						///< The gas for EVM code execution. Initial amount before go() execution, final amount after go() execution.
+
+    /// The amount of gas for EVM execution.
+    ///
+    /// Initial amount before go() execution, the amount of gas left after go() execution.
+    int64_t m_gas = 0;
```
gumb0: So what is the advantage of using signed instead of unsigned here?

chfast: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es102-use-signed-types-for-arithmetic

gumb0:
Some SSTORE tests need updating

PR #5322, opened 2018-10-16
ethash: Optimize difficulty to boundary implementation

chfast:


libethashseal/Ethash.h
```diff
@@ -69,7 +69,13 @@ class Ethash: public SealEngineBase
     static h256 seedHash(BlockHeader const& _bi);
     static Nonce nonce(BlockHeader const& _bi) { return _bi.seal<Nonce>(NonceField); }
     static h256 mixHash(BlockHeader const& _bi) { return _bi.seal<h256>(MixHashField); }
-    static h256 boundary(BlockHeader const& _bi) { auto d = _bi.difficulty(); return d ? (h256)u256((bigint(1) << 256) / d) : h256(); }
+
+    static h256 boundary(BlockHeader const& _bi)
+    {
+        const auto& d = _bi.difficulty();
```
gumb0: `auto const&`

libethashseal/Ethash.h
```diff
@@ -69,7 +69,13 @@ class Ethash: public SealEngineBase
...
+        return d > 1 ? h256{u256((u512(1) << 256) / d)} : ~h256{};
```
gumb0: It returned 0 before for d == 0, now it returns `ff...ff`, is this correct?

gumb0: Maybe make a `constexpr` constant out of `~h256{}`

chfast: > Maybe make a `constexpr` constant out of `~h256{}`

How? We don't have any `constexpr` support in `h256`.

gumb0: Could be just const then.

chfast: You mean `const h256 maxBoundary = ~h256{}`?

gumb0: Yes. Then it's guaranteed not to calculate negation each time.

PR #5325, opened 2018-10-18
Upgrade EVMC to version 6.0.0

chfast:
Upgrades the EVMC to version 6.0.0.


chfast:
I know there is `const` misplacement and I'll fix it. Any other comments?

PR #5328, opened 2018-10-21
Enable usage of MemoryDB in the unit tests and consensus tests

### LINKED ISSUE #5318: Update Blockchain tests to use MemoryDB for block and state storage
Update the Blockchain tests to have the option of being run with the new MemoryDB or LevelDB. Update the unit tests to only run using MemoryDB. The rationale behind having the Blockchain tests run using either MemoryDB or LevelDB is that they are considered to be integration tests and exercise user scenarios, whereas the unit tests are targeted at very specific functionality.

halfalicious:
Fix #5318 

These changes enable usage of MemoryDB in the unit (the tests in [test/unittests](https://github.com/ethereum/aleth/tree/master/test/unittests)) and consensus tests ([test/jsontests](https://github.com/ethereum/tests/tree/a69eb762fc5665814ac21cf091eff3bd0bf4ef27)). 

For the unit tests, MemoryDB is turned on for all tests except for 1 - `BlockChainFrontierSuite!opendb`: https://github.com/ethereum/aleth/blob/873a36e634bc7f57cabaf0685afb36713e2d35e1/test/unittests/libethereum/BlockChain.cpp#L53-L61

This is because this test case exercises a user scenario (opening the same database twice) so we want to still test this same scenario. Additionally, this scenario doesn't make sense when running MemoryDB. 

For the consensus tests, MemoryDB is on by default but optional - one can instruct the tests to use LevelDB via the new `--diskdb` command-line parameter. We have the option of these 2 databases for the consensus tests because these tests are integration-style tests which exercise user scenarios (instead of extremely targeted unit tests) so we therefore want to be able to hit the same code that users hit when running the Aleth client.

A few notes:
* I've made a couple of assumptions - that the fixtures defined in the test/jsontests directory are only used in the consensus tests and that `TestOutputHelperFixture `is only used in the unit tests. I validated these assumptions somewhat (more in the next bullet) but this validation doesn't catch every single case so it would be great if someone with more knowledge of these tests could comment on if these assumptions are correct (perhaps @winsvega). 
* I've verified that `--diskdb` turns on LevelDB for the consensus tests by running a private testeth which throws on MemoryDB creation, running the consensus tests, and verifying no exceptions were thrown. I verified that LevelDB was not turned on for the unit tests (either by default or when the `--diskdb` parameter is supplied) by running one set of unit tests (`BlockChainFrontierSuite`) using this same private and verifying that each test case threw (except for `opendb` since it always uses LevelDB).
* I explicitly set a database kind in all of the consensus test fixtures except for `VMTestFixture ` despite not knowing if a block or state database is used in the tests which use all of these fixtures because I assumed that even if these tests don't use a block/state database now, they might at some point in the future. I didn't set a database kind in VMTestFixture because I verified that these tests don't use a block/state db and I don't think it would make sense to use such functionality in the future.

### DIFF
--- a/libdevcore/DBFactory.cpp
+++ b/libdevcore/DBFactory.cpp
@@ -50,7 +50,7 @@ DBKindTableEntry dbKindsTable[] = {
     {DatabaseKind::MemoryDB, "memorydb"},
 };
 
-void setDatabaseKind(std::string const& _name)
+void setDatabaseKindByName(std::string const& _name)
 {
     for (auto& entry : dbKindsTable)
     {
@@ -61,8 +61,13 @@ void setDatabaseKind(std::string const& _name)
         }
     }
 
-    BOOST_THROW_EXCEPTION(eth::InvalidDatabaseKind()
-                          << errinfo_comment("invalid database kind supplied: " + _name));
+    BOOST_THROW_EXCEPTION(
+        eth::InvalidDatabaseKind() << errinfo_comment("invalid database name supplied: " + _name));
+}
+
+void setDatabaseKind(DatabaseKind _kind)
+{
+    g_kind = _kind;
 }
 
 void setDatabasePath(std::string const& _path)
@@ -105,7 +110,7 @@ po::options_description databaseProgramOptions(unsigned _lineLength)
 
     add("db",
         po::value<std::string>()->value_name("<name>")->default_value("leveldb")->notifier(
-            setDatabaseKind),
+            setDatabaseKindByName),
         description.data());
 
     add("db-path",
--- a/libdevcore/DBFactory.h
+++ b/libdevcore/DBFactory.h
@@ -42,6 +42,8 @@ boost::program_options::options_description databaseProgramOptions(
 
 bool isMemoryDB();
 DatabaseKind databaseKind();
+void setDatabaseKindByName(std::string const& _name);
+void setDatabaseKind(DatabaseKind _kind);
 boost::filesystem::path databasePath();
 
 class DBFactory
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -425,10 +425,18 @@ string BlockChain::dumpDatabase() const
 {
     ostringstream oss;
     oss << m_lastBlockHash << '\n';
-    m_extrasDB->forEach([&oss](db::Slice key, db::Slice value) {
-        oss << toHex(key) << "/" << toHex(value) << '\n';
+
+    // We need to first insert the db data into an ordered map so that the string returned from this function
+    // always has data in the same order, regardless of the underlying database implementation
+    std::map<std::string, std::string> dbData;
+    m_extrasDB->forEach([&dbData](db::Slice key, db::Slice value) {
+        dbData[key.toString()] = value.toString();
         return true;
     });
+
+    for (auto const& it : dbData)
+        oss << toHex(it.first) << "/" << toHex(it.second) << '\n';
+
     return oss.str();
 }
 
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -22,6 +22,7 @@
  */
 
 #include <libdevcore/FileSystem.h>
+#include <libdevcore/DBFactory.h>
 #include <test/tools/fuzzTesting/fuzzHelper.h>
 #include <test/tools/jsontests/BlockChainTests.h>
 #include <test/tools/libtesteth/TestHelper.h>
@@ -32,6 +33,7 @@
 using namespace std;
 using namespace json_spirit;
 using namespace dev;
+using namespace dev::db;
 using namespace dev::eth;
 namespace fs = boost::filesystem;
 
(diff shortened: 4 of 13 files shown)


halfalicious:
I thought this was ready but one of the unit tests is failing (BlockChainInsertTests/bcBasicInsert), I'll investigate

gumb0:
I suspect that this test inserts the same key/values in two databases, but then `dumpDatabase()` outputs them in different order for MemoryDB (because it uses `unordered_map` maybe?)
I guess we should change the test or/and `dumpDatabase` function so that comparison doesn't depend on the order of iteration over the database.

gumb0:
I suspect that this test inserts the same key/values in two databases, but then `dumpDatabase()` outputs them in different order for MemoryDB (because it uses `unordered_map` maybe?)
I guess we should change the test or/and `dumpDatabase` function so that comparison doesn't depend on the order of iteration over the database.

gumb0:
I suspect that this test inserts the same key/values into two databases, but then `dumpDatabase()` outputs them in different order for MemoryDB (because it uses `unordered_map` maybe?)
I guess we should change the test or/and `dumpDatabase` function so that comparison doesn't depend on the order of iteration over the database.

gumb0:
I suspect that this test inserts the same key/values into two databases, but then `dumpDatabase()` outputs them in different order for MemoryDB (because it uses `unordered_map` maybe?)
I guess we should change the test or/and `dumpDatabase` function so that comparison doesn't depend on the order of iteration over the database.

halfalicious:
> Fix #5318
> 
> These changes enable usage of MemoryDB in the unit (the tests in [test/unittests](https://github.com/ethereum/aleth/tree/master/test/unittests)) and consensus tests ([test/jsontests](https://github.com/ethereum/tests/tree/a69eb762fc5665814ac21cf091eff3bd0bf4ef27)).
> 
> For the unit tests, MemoryDB is turned on for all tests except for 1 - `BlockChainFrontierSuite!opendb`:
> 
> [aleth/test/unittests/libethereum/BlockChain.cpp](https://github.com/ethereum/aleth/blob/873a36e634bc7f57cabaf0685afb36713e2d35e1/test/unittests/libethereum/BlockChain.cpp#L53-L61)
> 
> Lines 53 to 61 in [873a36e](/ethereum/aleth/commit/873a36e634bc7f57cabaf0685afb36713e2d35e1)
> 
>  BOOST_AUTO_TEST_CASE(opendb) 
>  { 
>      TestBlock genesis = TestBlockChain::defaultGenesisBlock(); 
>      TransientDirectory tempDirBlockchain; 
>      ChainParams p(genesisInfo(eth::Network::TransitionnetTest), genesis.bytes(), genesis.accountMap()); 
>      BlockChain bc(p, tempDirBlockchain.path(), WithExisting::Kill); 
>      auto is_critical = []( std::exception const& _e) { return string(_e.what()).find("DatabaseAlreadyOpen") != string::npos; }; 
>      BOOST_CHECK_EXCEPTION(BlockChain bc2(p, tempDirBlockchain.path(), WithExisting::Verify), DatabaseAlreadyOpen, is_critical); 
>  } 
> This is because this test case exercises a user scenario (opening the same database twice) so we want to still test this same scenario. Additionally, this scenario doesn't make sense when running MemoryDB.
> 
> For the consensus tests, MemoryDB is on by default but optional - one can instruct the tests to use LevelDB via the new `--diskdb` command-line parameter. We have the option of these 2 databases for the consensus tests because these tests are integration-style tests which exercise user scenarios (instead of extremely targeted unit tests) so we therefore want to be able to hit the same code that users hit when running the Aleth client.
> 
> A few notes:
> 
> * I've made a couple of assumptions - that the fixtures defined in the test/jsontests directory are only used in the consensus tests and that `TestOutputHelperFixture `is only used in the unit tests. I validated these assumptions somewhat (more in the next bullet) but this validation doesn't catch every single case so it would be great if someone with more knowledge of these tests could comment on if these assumptions are correct (perhaps @winsvega).
> * I've verified that `--diskdb` turns on LevelDB for the consensus tests by running a private testeth which throws on MemoryDB creation, running the consensus tests, and verifying no exceptions were thrown. I verified that LevelDB was not turned on for the unit tests (either by default or when the `--diskdb` parameter is supplied) by running one set of unit tests (`BlockChainFrontierSuite`) using this same private and verifying that each test case threw (except for `opendb` since it always uses LevelDB).
> * I explicitly set a database kind in all of the consensus test fixtures except for `VMTestFixture ` despite not knowing if a block or state database is used in the tests which use all of these fixtures because I assumed that even if these tests don't use a block/state database now, they might at some point in the future. I didn't set a database kind in VMTestFixture because I verified that these tests don't use a block/state db and I don't think it would make sense to use such functionality in the future.

Based on a discussion that @gumb0  and I had in today's team chat I've modified the `--diskdb` setting to affect all tests, rather than just the consensus tests in the interest of a cleaner design (`BlockChainFrontierSuite!opendb` still always uses LevelDB though).

halfalicious:
@chfast: I'd prefer to leave the diskdb option as is because an option with a "DB" prefix implies that the option supports multiple possible values which is not the case here (though I might change this when adding RocksDB support)

chfast:
Ok, merge it then.

chfast:
Full clang6 build takes now 6 mins. Well done.

PR #5337, opened 2018-11-03
Allow enode in peerset

### LINKED ISSUE #5335: One cannot supply multiple peer URLs to Aleth via peerset argument
If one supplies multiple (space delimited) peer URLs to Aleth via the `--peerset` argument, all but the first are silently dropped. The peerset argument parsing happens here: https://github.com/ethereum/aleth/blob/37ead28874c5adb380d4c60eadff161ca9736ece/aleth/main.cpp#L454-L510

halfalicious:
Fix #5335 and #5197 

This is a continuation of #5235:
* Update the Aleth URL parsing code so that the URL format outlined here is supported: https://github.com/ethereum/wiki/wiki/enode-url-format
* Fix a bug where if one supplies multiple peer URLs to Aleth via the command line "--peerset" argument, all but the first one are silently dropped


halfalicious:
The change history got messed up so closing this PR, will track review in new PR #5338

PR #5338, opened 2018-11-03
Update required node URL format

### LINKED ISSUE #5335: One cannot supply multiple peer URLs to Aleth via peerset argument

halfalicious:
Fix #5335 and #5197 

This is a continuation of #5235:
* Update the Aleth URL parsing code so that the URL format outlined here is supported: https://github.com/ethereum/wiki/wiki/enode-url-format
* Fix a bug where if one supplies multiple peer URLs to Aleth via the command line "--peerset" argument, all but the first one are silently dropped


libp2p/Common.cpp
```diff
@@ -19,6 +19,7 @@
  * @date 2014
  */
 
+#include <regex>
```
gumb0: Please apply clang-format to your changes, it should put this below other includes

libp2p/Common.cpp
```diff
@@ -212,18 +214,23 @@ std::string NodeSpec::enode() const
...
-        ret += ":" + toString(m_udpPort);
-
+    }
+   
     if (m_id)
         return "enode://" + m_id.hex() + "@" + ret;
     return ret;
 }
 
+bool NodeSpec::isInitialized() const
```
gumb0: Maybe `isValid` ?

libp2p/Common.cpp
```diff
@@ -180,25 +181,26 @@ Node::Node(NodeSpec const& _s, PeerType _p):
...
-        m_address = m_address.substr(0, colon);
-        size_t p2 = ports.find_first_of(".");
-        if (p2 != string::npos)
+        m_id = p2p::NodeID(match.str(2));
+        m_address = match.str(3);
+        if (match[5].matched)
         {
-            m_udpPort = stoi(ports.substr(p2 + 1));
-            m_tcpPort = stoi(ports.substr(0, p2));
+            m_udpPort = m_tcpPort = stoi(match.str(5).substr(1));
```
gumb0: I guess with another set of parenthesis in regex you could get rid of `substr` call here

aleth/main.cpp
```diff
@@ -451,63 +452,43 @@ int main(int argc, char** argv)
...
     if (vm.count("peer-stretch"))
         peerStretch = vm["peer-stretch"].as<int>();
+
     if (vm.count("peerset"))
     {
-        string peerset = vm["peerset"].as<string>();
-        if (peerset.empty())
+        string peersetStr = vm["peerset"].as<string>();
+        vector<string> peersetList;
+        boost::split(peersetList, peersetStr, boost::is_any_of(","));
```
gumb0: I think should be no need for `is_any_of`

gumb0: Actually, couldn't we instead use regex itself to get an array of matches in the `peersetStr`?

libp2p/Common.cpp
halfalicious: Thanks for the tip, I tend to not do that since it causes a lot of churn in some files (e.g. libethereum\Blockchain.cpp) but I'll do so in this case.

halfalicious: > Thanks for the tip, I tend to not do that since it causes a lot of churn in some files (e.g. libethereum\Blockchain.cpp) but I'll do so in this case.

Actually, it causes a lot of churn in this file as well...I'll manually move the #include.

libp2p/Common.cpp
halfalicious: I purposely used substr here since the regex was getting a bit unwieldy with all of the parentheses in the ports section...I'd like to leave this as-is unless you feel strongly about removing the substr.

aleth/main.cpp
halfalicious: > I think should be no need for `is_any_of`
split needs a predicate to determine if the current char is a delimiter...what did you have in mind instead of is_any_of?

> Actually, couldn't we instead use regex itself to get an array of matches in the `peersetStr`?
We could, but I think that's a bit overkill and would be more cumbersome since we'll need to use regex capture groups and iterate over the groups. I'd prefer to leave this code as-is.

gumb0: It was a guess about `is_any_of`, I thought maybe you can pass just a character delimiter to it, but looks like maybe you can't, sorry

libp2p/Common.cpp
halfalicious: > You should appply it only to your changes not

Thanks, I'll take another look at that.

PR #5343, opened 2018-11-08
Fix Binaryen build for testeth

chfast:


winsvega:
Oh, I actually get this error every time I build the project after cmake + reboot. I had to remove the previous build information and rerun cmake in order to resolve this

winsvega:
Binaryen compilation error. 
It could only be fixed if I remove data from build folder and rerun cmake again.

PR #5344, opened 2018-11-08
dockerfile for testeth

holiman:
This is an attempt to create a docker file for testeth, so we can put that as an automated build on docker hub. cc @holgerd77

### DIFF
--- a/scripts/docker/testeth.dockerfile
+++ b/scripts/docker/testeth.dockerfile
@@ -1,0 +1,21 @@
+# Multistage Dockerfile for the testeth tool.
+
+# Build stage
+FROM alpine:latest as builder
+RUN apk add --no-cache \
+        git \
+        cmake \
+        g++ \
+        make \
+        linux-headers
+ADD . /source
+WORKDIR /build
+RUN cmake /source -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Release -DTOOLS=OFF -DHUNTER_JOBS_NUMBER=$(nproc)
+RUN make -j $(nproc) && make install
+
+# Install stage
+FROM alpine:latest
+RUN apk add --no-cache libstdc++
+COPY --from=builder /build/test/testeth /usr/bin/testeth
+
+ENTRYPOINT ["/usr/bin/testeth"]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,3 @@
 /build
 /deps
-/hera
-/test
+/test/jsontests


scripts/docker/testeth.dockerfile
```diff
@@ -0,0 +1,21 @@
+# Build stage 
+FROM alpine:latest as builder
+RUN apk add --no-cache \
+        git \
+        cmake \
+        g++ \
+        make \
+        linux-headers \
+        leveldb-dev --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing/
```
chfast: Should not be needed any more.

scripts/docker/testeth.dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN git clone --recursive https://github.com/ethereum/aleth --branch fix-binaryen-build --single-branch --depth 1
```
chfast: Fix merged. Switch back to master.

scripts/docker/testeth.dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN mkdir /build
+RUN cd /build && cmake /aleth
+RUN cd /build && cmake --build .
+
+FROM alpine:latest
+RUN apk add --no-cache \
+        libstdc++ \
+        leveldb --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing/
```
chfast: Should not be needed any more.

holgerd77:
@chfast Thanks for reviewing this!

holgerd77:
If at the end this could be put on the organizational ethereum docker hub accout (it's probably https://hub.docker.com/r/ethereum/ ?) I think this would be given a bit more trust, not necessarily needed though.

chfast:
Possible, but I think we had to combine testeth and aleth into a single container image.

holgerd77:
In doubt do what is easier.

holiman:
I'm testing it out right now.. https://hub.docker.com/r/ethereum/client-cpp/builds/

holiman:
Failing:
```
Build failed: The command '/bin/sh -c cmake /source -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Release -DTOOLS=OFF -DHUNTER_JOBS_NUMBER=$(nproc)' returned a non-zero code: 1
```
I wonder... the dockerfiles are in `/scripts/docker` -- and we're referencing the repo root with `.`. Are you sure that works, @chfast ? Could we try placing the dockerfile in the repo root instead ?

chfast:
I'm building it as
```
docker build -t ethereum/testeth -f scripts/docker/testeth.dockerfile . 
```

holiman:
Right, the question is what docker hub builds it with...

holiman:
A few years old, but still: https://forums.docker.com/t/automated-build-on-docker-hub-is-there-a-way-to-specify-a-dockerfile/1078 :

> For now, you cannot specify specific Dockerfile and also the location you keep your Dockerfile, it is all the resources it can have access. You cannot even create symlinks to locate to parent folders. If we make changes, we would definitely let everyone know, like in cli.

PR #5348, opened 2018-11-12
 test: Introduce aleth-unittests

chfast:
This adds new executable aleth-unittests powered by GTest.
Two files were converted so far:
- libdevcore/CommonJS.cpp,
- libdevcore/core.cpp.

### DIFF
--- a/test/unittests/libdevcore/CommonJS.cpp
+++ b/test/unittests/libdevcore/CommonJS.cpp
@@ -1,136 +1,128 @@
 /*
-	This file is part of cpp-ethereum.
+    This file is part of cpp-ethereum.
 
-	cpp-ethereum is free software: you can redistribute it and/or modify
-	it under the terms of the GNU General Public License as published by
-	the Free Software Foundation, either version 3 of the License, or
-	(at your option) any later version.
+    cpp-ethereum is free software: you can redistribute it and/or modify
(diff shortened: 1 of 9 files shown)


winsvega:
so unittests are now to be migrated from testeth to aleth-unittests and separated from boost? 
this will also require automatition script updates on travis and such.

chfast:
> this will also require automatition script updates on travis and such.

No. These are included in ctest.

chfast:
Please review!

test/unittests/libdevcore/CommonJS.cpp
```diff
@@ -1,136 +1,128 @@
...
+#include <gtest/gtest.h>
 
 using namespace dev;
 using namespace std;
-using namespace test;
-
-BOOST_FIXTURE_TEST_SUITE(CommonJSTests, TestOutputHelperFixture)
 
-BOOST_AUTO_TEST_CASE(test_toJS)
+TEST(common_js, toJS)
```
gumb0: Maybe let's try to use camelCase more consistently for both test case names and test names, so this is perhaps `CommonJS`

chfast:
> What about codecov, will it count only testeth or only unittests or both?

Both. At least for now.

PR #5355, opened 2018-11-15
Make EthereumCapability thread-safe again

gumb0:
- Change background tasks of `EthereumCapability::doWork` to be performed from the networking thread instead of separate `Worker` thread
- Make `onTransactionImported` handler and `reset()` method code be run from the networking thread as well
- Make `atomic` some members of `EthereumCapability` that are still accessed from separate threads (like RPC handler thread)

### DIFF
--- a/libethereum/EthereumCapability.cpp
+++ b/libethereum/EthereumCapability.cpp
@@ -34,6 +34,7 @@ using namespace dev::eth;
 static unsigned const c_maxSendTransactions = 256;
 static unsigned const c_maxHeadersToSend = 1024;
 static unsigned const c_maxIncomingNewHashes = 1024;
+static int const c_backroundWorkPeriodMs = 1000;
 
 char const* const EthereumCapability::s_stateNames[static_cast<int>(SyncState::Size)] = {
     "NotSynced", "Idle", "Waiting", "Blocks", "State"};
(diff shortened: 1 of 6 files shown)


libethereum/EthereumCapability.h
```diff
@@ -200,8 +191,9 @@ class EthereumCapability : public p2p::CapabilityFace, Worker
 
     std::unordered_map<NodeID, EthereumPeer> m_peers;
 
-    std::mt19937_64 m_urng;  // Mersenne Twister psuedo-random number generator
+    std::atomic<bool> m_backgroundWork = {false};
```
chfast: Can you rename it to `m_backgroundWorkEnabled`?

libethereum/EthereumCapability.cpp
```diff
@@ -409,9 +409,15 @@ EthereumCapability::EthereumCapability(shared_ptr<p2p::CapabilityHostFace> _host
     m_urng = std::mt19937_64(seed());
 }
 
-EthereumCapability::~EthereumCapability()
+void EthereumCapability::onStarting()
 {
-    terminate();
+    m_backgroundWork = true;
```
chfast: If you want something fancy
```suggestion
    m_backgroundWork.store(true, std::memory_order_relaxed);
```

Similarly for load.

libethereum/EthereumCapability.h
gumb0: Renamed

libethereum/EthereumCapability.cpp
gumb0: I'll leave it; not convinced yet that it gives enough improvement here to account for more confusing code

chfast: I was curious.
https://godbolt.org/z/11A51F

PR #5356, opened 2018-11-15
move dockerfiles to root

holiman:
I think this solves the problem of building on docker hub, as discussed in https://github.com/ethereum/aleth/pull/5344


holiman:
Ping @chfast

chfast:
I believe it was working before, just Docker Hub was not able to finish the build (no errors reported, probably out-of-memory).

holiman:
I don't think so.. Could we please then make copies of at least one of the files into the root, so I can try it out?

chfast:
Done.

holiman:
Interesting.. So this 'works' -- docker hub now does not fail on an early stage, instead it builds for two hours, and then aborts the build.

chfast:
Well, I believe I had similar behavior when the docker file was specified as `scripts/docker/aleth.dockerfile`.

Anyway, seems it is too heavy for Docker Hub. I can have it done on Circle CI, but I have to figure out how to build them (aleth and testeth) once.

holiman:
Build them both into an `alltools`image, then use a multistage build to just pick out the cherries from `alltools`?

chfast:
The automation seems to be working. You can test `ethereum/testeth:20181122` docker image.

Issue #5334, opened 2018-10-30
Log execution trace of aleth-vm


ireneGP:
Hello,

I am opening an issue about the execution trace log of `aleth-vm`.  I have talked with Paweł Bylica via the public channel of `aleth` on the gitter.

I am aware that `evm` and `parity-vm` has a debug mode which prints out the stack usage at each step. I cannot find such an option in `aleth-vm` (newest version, aleth-vm 1.5.0-alpha.3-4+commit.816e52de).

I tried the following commands but it does not work.

    ./aleth-vm --code 60026000556001600055 --log-verbosity 4
    ./aleth-vm trace --code 60026000556001600055 --log-verbosity 4

Here are some outputs on my computer, by running the `aleth-vm` with some sample inputs:

    aleth-vm git:(master) ./aleth-vm --code 60026000556001600055 --log-verbosity 20
     Gas used: 25012 (+21000 for transaction, -0 refunded)
    Output:
    0 logs.
    7 operations in 1.5e-05 seconds. 
    Maximum memory usage: 0 bytes
    Expensive operations:
      SSTORE x 2 (0 gas)
    DEBUG 10-30 14:55:56 aleth-vm overlaydb Closing state DB

gumb0:
I believe you need to compile it with `VMTRACE` flag to enable it (`cmake .. -DVMTRACE=ON`)

ireneGP:
@werew  Thanks. One followup question, why it cannot print out the corresponding opcode each step?

werew:
@ireneGP adding `st.setShowMnemonics();` in the main function will do the job:
```
$ ./aleth-vm trace --code 60026000556001600055                                                                                     
[                                                                                                                                                     
   {                                                                                                                                                  
      "depth" : "0",                                                                                                                                  
      "gas" : "9223372036854775807",
      "gasCost" : "3",
      "memory" : [],
      "op" : "PUSH1",
      "pc" : "0",
      "stack" : [],
      "storage" : {}
   },
   {
      "depth" : "0",
      "gas" : "9223372036854775804",
      "gasCost" : "3",
      "op" : "PUSH1",
      "pc" : "2",
      "stack" : [ "0x02" ]
   },
   ...
]
```

ireneGP:
@werew Perfect, thanks a lot!

gumb0:
I believe this is resolved, thanks @werew

PR #5357, opened 2018-11-17
Fixing log execution trace of aleth-vm

werew:
This is just a fix for #5334.

### DIFF
--- a/aleth-vm/main.cpp
+++ b/aleth-vm/main.cpp
@@ -348,7 +348,7 @@ int main(int argc, char** argv)
         };
     }
     else if (mode == Mode::Trace)
-        onOp = st;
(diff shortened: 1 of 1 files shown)


aleth-vm/main.cpp
```diff
@@ -348,7 +348,7 @@ int main(int argc, char** argv)
         };
     }
     else if (mode == Mode::Trace)
-        onOp = st;
```
gumb0: @chfast I think assigning a function object to a `std::function` creates a copy of it inside `std::function`

But then in the end (line 388) we output the accumulated trace from the original object.

chfast: Yes, your are right. I did not realized that `std::function` copies objects.

PR #5359, opened 2018-11-19
[WIP] Remove "mode" command-line argument

### LINKED ISSUE #5358: Aleth's "peer" mode option isn't actually supported
According to Aleth's command-line help message, one can supply "peer" and "full" values for the "mode" argument:
`
  -o [ --mode ] <full/peer>               Start a full node or a peer node (default: full)`

However, I've looked at the processing of this argument in main.cpp and it appears that nothing actually happens when the "peer" option is supplied. I've confirmed that running Aleth with the "peer" option results in Aleth immediately quitting:

```
C:\Users\nilse\Documents\Code\aleth\build\aleth\Debug>aleth --mode peer
aleth, a C++ Ethereum client
INFO  11-18 20:35:31 main net    Id: ##132397b7ΓÇª
aleth 1.5.0-alpha.6-7+commit.3f95c5f6.dirty
INFO  11-18 20:35:33 p2p  info   UPnP device not found.
Node ID: enode://132397b7af351f81826080e0ff9ba2187d5c1f1b5fb010b656722f15e26c922787cfcc8a8005a28f1400aff6699cfccb1cde560f7290e9a2aae05bd4d4cab231@0.0.0.0:0

C:\Users\nilse\Documents\Code\aleth\build\aleth\Debug>
```

halfalicious:
Fix #5358
Remove the "mode" command-line argument (supplied via --mode or --o) since one of the two options listed (peer) doesn't actually do anything.

### DIFF
--- a/aleth/main.cpp
+++ b/aleth/main.cpp
@@ -124,12 +124,6 @@ void importPresale(KeyManager& _km, string const& _file, function<string()> _pas
     _km.import(k.secret(), "Presale wallet" + _file + " (insecure)");
 }
 
-enum class NodeMode
-{
-    PeerServer,
-    Full
-};
-
 enum class OperationMode
 {
     Node,
@@ -198,9 +192,6 @@ int main(int argc, char** argv)
     string exportFrom = "1";
     string exportTo = "latest";
     Format exportFormat = Format::Binary;
-
-    /// General params for Node operation
-    NodeMode nodeMode = NodeMode::Full;
 
     bool ipc = true;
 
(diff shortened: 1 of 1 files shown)


gumb0:
Agree with what Pawel said, and I would prefer it to be fixed and the command line options probably renamed.

Here's the issue that tracked it https://github.com/ethereum/aleth/issues/4703

I believe it should be not difficult to fix: `main` function should just skip anything accessing `web3.ethereum()` in this mode.
Or maybe better we could make a separate executable for it, which would have only networking CLI options (I think in go-ethereum it's a separate binary)

halfalicious:
Got it, I’ll investigate what geth does and change this PR so that Aleth follows that behavior.

halfalicious:
Closing since this will be tracked in a new PR: #5378

PR #5363, opened 2018-11-22
Publish docker image of testeth

chfast:
Closes #5356.


winsvega:
Is this masterbranch docker image?

chfast:
> Is this masterbranch docker image?

If you mean git master branch, then not exactly.

We publish docker images for every "release" git tag (starting with `v...`) and we also run builds nightly.
We do not build every PR merged to master.

If you want the "recent" build the `ethereum/testeth:nightly` is the closest approximation.

If you want something special, you can always push git tag like e.g. `v1.5.0-special-build-for-dimitry' or something to `docker` branch.

See https://hub.docker.com/r/ethereum/aleth/tags/ which has longer history for reference.

PR #5364, opened 2018-11-22
Fix test_mineBlocks returning too early with soltest

gumb0:
Fixes two problems:
- Blocks are first sealed, then imported into the chain. `test_mineBlocks` stopped sealing after all requested blocks are imported; it should better stop as soon as all requested blocks are sealed, otherwise as sealing continues in parallel with import, more blocks than requested can be sealed.
- `test_mineBlocks` returned after all requested blocks are imported into the chain. But it turns out, that at this moment `Client` class did not update the pending block yet, and `eth_call(..., "pending")` immediately after was accessing incorrect block. Now `test_mineBlocks` returns only after `Client` updates its stuff.

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -122,7 +122,6 @@ void Client::init(p2p::Host& _extNet, fs::path const& _dbPath,
     bc().setOnBlockImport([=](BlockHeader const& _info) {
         if (auto h = m_host.lock())
             h->onBlockImported(_info);
-        m_onBlockImport(_info);
     });
 
     if (_forceAction == WithExisting::Rescue)
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -566,6 +566,7 @@ void Client::onChainChanged(ImportRoute const& _ir)
     if (!isMajorSyncing())
         resyncStateFromChain();
     noteChanged(changeds);
+    m_onChainChanged(_ir.deadBlocks, _ir.liveBlocks);
 }
 
 bool Client::remoteActive() const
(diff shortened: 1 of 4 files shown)


gumb0:
@winsvega please check that this still works with retesteth

gumb0:
Also I made a similar fix in `test::mine()` function used in unit tests and removed `Client::m_onBlockImport`

@chfast please review

winsvega:
some tests work, some not. 
mining does not work. 

I get this error: 
```
Error on JSON-RPC call (createInitFailBadJumpDestination): Block gas limit reached. Request: {"jsonrpc":"2.0","method":"eth_sendRawTransaction","params":["0xf86480018405f5e10094095e7baea6a6c7c4c2dfeb977efac326af552d87830186a0801ba088544c93a564b4c28d2ffac2074a0c55fdd4658fe0d215596ed2e32e3ef7f56ba07fb4075d54190f825d7c47bb820284757b34fd6293904a93cddb1d3aa961ac28"],"id":3}
Error on JSON-RPC call (CallRecursiveBombPreCall): Block gas limit reached. Request: {"jsonrpc":"2.0","method":"eth_sendRawTransaction","params":["0xf8658001887ffffffffffffff094095e7baea6a6c7c4c2dfeb977efac326af552d8780801ca08503ee9dc6d31d42de8457475bf2ab1ebedf4ce7cae9e37d03a7af523012d816a039d25114f6bb11240492c0818ad9cbbd95339520e15ecf4bd6dfd475b2121196"],"id":33}
```

but it might be due to a long time since I last rebased retesteth. (4 months)

gumb0:
@winsvega If it's no worse than current `master`, then I think I'll merge it now

PR #5372, opened 2018-11-29
Clean-up Discovery implementation

gumb0:
1. Renaming some methods/data members.
2. Remove the feature to add a node by IP+port only and discover its public key by sending PING - this is not used anyway and complicates the code.

### DIFF
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -134,9 +134,10 @@ list<NodeEntry> NodeTable::snapshot() const
 Node NodeTable::node(NodeID const& _id)
 {
     Guard l(x_nodes);
-    if (m_allNodes.count(_id))
+    auto const it = m_allNodes.find(_id);
+    if (it != m_allNodes.end())
     {
-        auto entry = m_allNodes[_id];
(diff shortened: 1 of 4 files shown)


libp2p/NodeTable.cpp
```diff
@@ -562,8 +537,8 @@ void NodeTable::doCheckEvictions()
             Guard ln(x_nodes);
             for (auto& e: m_evictions)
                 if (chrono::steady_clock::now() - e.second.evictedTimePoint > c_reqTimeout)
-                    if (m_nodes.count(e.second.newNodeID))
-                        drop.push_back(m_nodes[e.second.newNodeID]);
+                    if (m_allNodes.count(e.second.newNodeID))
```
chfast: `.find()`.

gumb0:
I planned to add the real changes here, too, but now I think we can also merge this first and further work will go to https://github.com/ethereum/aleth/pull/5380

libp2p/NodeTable.cpp
gumb0: this is fixed

PR #5373, opened 2018-12-01
fix rlp tests to require prefixed values

winsvega:
tests on top of latest develop


winsvega:
rebase after this https://github.com/ethereum/aleth/pull/5194

PR #5376, opened 2018-12-02
Increase default gas limit to 8M

naikmyeong:
Geth and Parity already has 8M gas cap as a default .

ref)

https://github.com/paritytech/parity-ethereum/pull/9564
https://github.com/ethereum/go-ethereum/pull/17546/files#diff-21609b3ea13b7a6695113bee0fd055acR51

cc @chfast

### DIFF
--- a/libethcore/SealEngine.cpp
+++ b/libethcore/SealEngine.cpp
@@ -235,7 +235,7 @@ u256 calculateEthashDifficulty(
 u256 calculateGasLimit(
     ChainOperationParams const& _chainParams, BlockHeader const& _bi, u256 const& _gasFloorTarget)
 {
-    u256 gasFloorTarget = _gasFloorTarget == Invalid256 ? 3141562 : _gasFloorTarget;
+    u256 gasFloorTarget = _gasFloorTarget == Invalid256 ? 8000000 : _gasFloorTarget;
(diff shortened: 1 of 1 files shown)


halfalicious:
Can we turn the gas limit into a global constant?

test/unittests/libethashseal/EthashTest.cpp
```diff
@@ -181,7 +181,7 @@ TEST(Ethash, etashVerify)
     header.setLogBloom({});
     header.setDifficulty(h256{131072});
     header.setNumber(1);
-    header.setGasLimit(3141562);
+    header.setGasLimit(8000000);
```
axic: Updating this field in the test requires the hashes to be updated too. Though I do not think it is important to change the gas limit in this test.

naikmyeong: @axic Maybe aleth maintainers can?

axic: I am not sure what you mean, but if you remove this part of your commit, it has a better chance of merging.

naikmyeong: @axic reverted, thanks 👍

naikmyeong:
@axic @chfast Resolved

Issue #4703, opened 2017-12-13
--mode peer is broken


gumb0:
Starting with `--mode peer` leads to uncaught exception.
```
> eth/eth --mode peer
cpp-ethereum, a C++ Ethereum client
cpp-ethereum 1.3.0
  By cpp-ethereum contributors, (c) 2013-2016.
  See the README for contributors and credits.
  ℹ  01:52:56 PM.317|p2p  UPnP device not found.
Node ID: enode://493d52068ec72230688da539926f47a452b762bc348d2ab1491f399b532186d71d7c512e09ffb8e9c24d292d064c00f6234ef1221bc0d86093d2de32358d33da@0.0.0.0:0
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<dev::InterfaceNotSupported>'
  what():  std::exception
fish: Job 1, “eth/eth --mode peer” terminated by signal SIGABRT (Abort)
```

I'm not sure why this mode existed and probably we should just remove it.

chfast:
Inspect and remove if you can't find anything there.

gumb0:
It was supposed to run a node without any capabilities (no eth, no whisper)
Maybe it's something like a bootnode of geth - not acting as a full client, but participating in node discovery?

chfast:
We will be setting up a testnet for eWASM today, so I will try it.

gumb0:
The fix could be for aleth's `main` function just to skip anything accessing `web3.ethereum()` in this mode.

Or better we could make a separate executable for it, which would have only networking CLI options (in go-ethereum there's a separate binary for the bootnode)

halfalicious:
Information on geth's bootnode executable: https://github.com/ethereum/go-ethereum/wiki/Setting-up-private-network-or-local-cluster

halfalicious:
I've looked at the Aleth networking code (`Hosts` and `NodeTable`) and the Geth bootnode documentation and here's how I'm planning on addressing this issue:
* I'm going to create a new executable (called something like `bootnode`) which will listen on a specific UDP port for discovery requests. This port will be configurable via a command-line argument.
* This new bootnode mode will only be relevant if you want to run a private chain (since all it does is build a node table via discovery requests and then service discovery "find neighbor" requests using this table...it will not connect to the Ethereum bootnodes defined in `Host::PoCHosts` or any hosts supplied via the command line, which appears to be the same behavior as Geth's bootnode executable)
[...]

PR #5378, opened 2018-12-04
Create aleth-bootnode executable and remove "mode" argument from Aleth

### LINKED ISSUE #4855: Lack of bootnode mechanism
The current implementation (at the time of writing) of cpp-ethereum lacks of the bootnode implementation when compared against the go-ethereum.
We do need to have a C++ implementation of bootnode. Otherwise, we must borrow the "bootnode" from go-ethereum. However, it is impossible to use this kind of "bootnode" for bootstrapping the "eth" nodes to p2p network. This is because "bootnode" initiates UDP endpoints whereas the "eth" nodes accept only TCP endpoints. 
Alternative to the C++ implementation of bootnode, we can change the cpp-ethereum implementation in such a way that it accepts UDP endpoint. Particularly, the change can happen inside the following function:
```

//libp2p/Host.cpp
void Host::connect(std::shared_ptr<Peer> const& _p)
{
    ....
    bi::tcp::endpoint ep(_p->endpoint);
}
```
Furthermore, go-ethereum client supports the option "--bootnodes" which is also missing in cpp-ethereum. We should support this nice option as well.

halfalicious:
Fix #4855, #4703 

Create the aleth-bootnode executable which just starts the `Host `thread in discovery mode with no p2p functionality. This is the same approach that Geth's "bootnode" binary takes (please see #4703 for details).

Configuration options:
```

NAME:
   aleth-bootnode
USAGE:
   aleth-bootnode [options]

GENERAL OPTIONS:
  -h [ --help ]         Show this help message and exit

NETWORKING:
  --public-ip <ip>          Force advertised public IP to the given IP (default: auto)
  --listen-ip <ip>(:<port>) Listen on the given IP for incoming connections (default: 0.0.0.0)
  --listen <port>           Listen on the given port for incoming connections (default: 30303)

LOGGING OPTIONS:
  -v [ --log-verbosity ] <0 - 4>        Set the log verbosity from 0 to 4 (default: 2).
  --log-channels <channel_list>         Space-separated list of the log channels to show (default: show all channels).
  --log-exclude-channels <channel_list> Space-separated list of the log channels to hide.
```

I've also removed the "mode" argument from Aleth's command line since it doesn't provide any purpose.

Note that I still need to verify that the bootnode works correctly - in my tests yesterday (running aleth-bootnode and 2 aleth clients) I didn't see the bootnode or either of the clients send FINDNODE packets so I need to investigate and figure out why.

I also need to add `composeClientVersion `functionality before merging.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,6 +123,7 @@ add_subdirectory(libwebthree)
 add_subdirectory(libweb3jsonrpc)
 
 add_subdirectory(aleth)
+add_subdirectory(aleth-bootnode)
 
 if (TOOLS)
     add_subdirectory(aleth-key)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,12 +123,12 @@ add_subdirectory(libwebthree)
 add_subdirectory(libweb3jsonrpc)
 
 add_subdirectory(aleth)
-add_subdirectory(aleth-bootnode)
 
 if (TOOLS)
     add_subdirectory(aleth-key)
     add_subdirectory(aleth-vm)
     add_subdirectory(rlp)
+    add_subdirectory(aleth-bootnode)
 endif()
 
 if (TESTS)
--- a/aleth-bootnode/CMakeLists.txt
+++ b/aleth-bootnode/CMakeLists.txt
@@ -6,7 +6,7 @@ set(
 add_executable(aleth-bootnode ${sources})
 target_link_libraries(
     aleth-bootnode
-    PRIVATE devcore Boost::program_options p2p
+    PRIVATE p2p devcore Boost::program_options
 )
 
 install(TARGETS aleth-bootnode DESTINATION bin)
--- a/aleth-bootnode/CMakeLists.txt
+++ b/aleth-bootnode/CMakeLists.txt
@@ -1,0 +1,12 @@
+set(
+    sources
+    main.cpp
+)
+
+add_executable(aleth-bootnode ${sources})
+target_link_libraries(
+    aleth-bootnode
+    PRIVATE devcore Boost::program_options p2p
+)
+
+install(TARGETS aleth-bootnode DESTINATION bin)
--- a/aleth-bootnode/main.cpp
+++ b/aleth-bootnode/main.cpp
@@ -1,0 +1,154 @@
+/*
+    This file is part of aleth.
+
+    aleth is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    aleth is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with aleth.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <libdevcore/FileSystem.h>
+#include <libdevcore/LoggingProgramOptions.h>
+#include <libp2p/Host.h>
+#include <boost/program_options.hpp>
+#include <boost/program_options/options_description.hpp>
+#include <iostream>
+#include <thread>
+
+namespace po = boost::program_options;
+namespace fs = boost::filesystem;
+
+using namespace dev;
+using namespace dev::p2p;
+using namespace std;
+
+int main(int argc, char** argv)
+{
+    setDefaultOrCLocale();
+
+    string const ProgramName = "aleth-bootnode";
+    string const NetworkConfigFileName = ProgramName + "-network.rlp";
+
+    /// Networking params.
+    string listenIP;
+    unsigned short listenPort = c_defaultIPPort;
+    string publicIP;
+    bool upnp = true;
+
+    po::options_description generalOptions("GENERAL OPTIONS", lineWidth());
+    auto addGeneralOption = generalOptions.add_options();
+    addGeneralOption("help,h", "Show this help message and exit\n");
+
+    LoggingOptions loggingOptions;
+    po::options_description loggingProgramOptions(
+        createLoggingProgramOptions(lineWidth(), loggingOptions));
+
+    po::options_description clientNetworking("NETWORKING", lineWidth());
+    auto addNetworkingOption = clientNetworking.add_options();
+#if ETH_MINIUPNPC
+    addNetworkingOption(
+        "upnp", po::value<string>()->value_name("<on/off>"), "Use UPnP for NAT (default: on)");
+#endif
+    addNetworkingOption("public-ip", po::value<string>()->value_name("<ip>"),
+        "Force advertised public IP to the given IP (default: auto)");
+    addNetworkingOption("listen-ip", po::value<string>()->value_name("<ip>(:<port>)"),
+        "Listen on the given IP for incoming connections (default: 0.0.0.0)");
+    addNetworkingOption("listen", po::value<unsigned short>()->value_name("<port>"),
+        "Listen on the given port for incoming connections (default: 30303)\n");
+
+    po::options_description allowedOptions("Allowed options");
+    allowedOptions.add(generalOptions).add(loggingProgramOptions).add(clientNetworking);
+
+    po::variables_map vm;
+    try
+    {
+        po::parsed_options parsed = po::parse_command_line(argc, argv, allowedOptions);
+        po::store(parsed, vm);
+        po::notify(vm);
+    }
+    catch (po::error const& e)
+    {
+        cout << e.what() << "\n";
+        return -1;
+    }
+
+    if (vm.count("help"))
+    {
+        cout << "NAME:\n"
+             << "   " << ProgramName << "\n"
+             << "USAGE:\n"
+             << "   " << ProgramName << " [options]\n\n";
+        cout << generalOptions << clientNetworking << loggingProgramOptions;
+        return 0;
+    }
+
+    /// Networking params.
+    string listenIP;
+    unsigned short listenPort = c_defaultIPPort;
+    string publicIP;
+    bool upnp = true;
+
+#if ETH_MINIUPNPC
+    if (vm.count("upnp"))
+    {
+        string m = vm["upnp"].as<string>();
+        if (isTrue(m))
+            upnp = true;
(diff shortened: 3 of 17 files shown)


libdevcore/Common.h
```diff
@@ -303,4 +303,16 @@ enum class WithExisting: int
 int64_t utcTime();
 
 void setDefaultOrCLocale();
+
+unsigned lineWidth();
```
chfast: Just make it `static constexpr unsigned c_lineWidth = 160`.

CMakeLists.txt
```diff
@@ -123,6 +123,7 @@ add_subdirectory(libwebthree)
 add_subdirectory(libweb3jsonrpc)
 
 add_subdirectory(aleth)
+add_subdirectory(aleth-bootnode)
```
gumb0: Maybe this should go under `if (TOOLS)` ?

chfast: Yes.

chfast:
@halfalicious it builds for me without a problem. Finding `aleth-buildinfo.h` in VS might be tricky, but you can try finding `aleth-buildinfo` project in VS solution and/or check `build dir/aleth/` directory.

aleth-bootnode/main.cpp
```diff
@@ -0,0 +1,129 @@
...
+    setupLogging(loggingOptions);
+    if (loggingOptions.verbosity > 0)
+        cout << EthGrayBold << ProgramName << ", a C++ Ethereum bootnode implementation" EthReset
+             << "\n";
+
+    auto netPrefs = publicIP.empty() ? NetworkConfig(listenIP, listenPort, upnp) :
+                                       NetworkConfig(publicIP, listenIP, listenPort, upnp);
+    auto netData = contents(getDataDir() / fs::path(NetworkConfigFileName));
+
+    // TODO: Compose client version
```
gumb0: Not that much important for the bootnode I think (it's used in RLPx handshake, but we'll return no capabilities and it won't go any further)

aleth/main.cpp
```diff
@@ -828,8 +790,7 @@ int main(int argc, char** argv)
         chainParams.allowFutureBlocks = true;
 
     dev::WebThreeDirect web3(WebThreeDirect::composeClientVersion("aleth"), db::databasePath(),
-        snapshotPath, chainParams, withExisting, nodeMode == NodeMode::Full ? caps : set<string>(),
-        netPrefs, &nodesState, testingMode);
+        snapshotPath, chainParams, withExisting, caps, netPrefs, &nodesState, testingMode);
```
gumb0: `caps` parameter doesn't do anything useful anymore, we could remove it

aleth/main.cpp
```diff
@@ -989,15 +950,12 @@ int main(int argc, char** argv)
...
-    eth::Client* c = nodeMode == NodeMode::Full ? web3.ethereum() : nullptr;
-    if (c)
-    {
-        c->setGasPricer(gasPricer);
-        c->setSealer(miner.minerType());
-        c->setAuthor(author);
-        if (networkID != NoNetworkID)
-            c->setNetworkId(networkID);
-    }
+    eth::Client* c = web3.ethereum();
```
gumb0: I would use a reference here to avoid raw pointers

libp2p/NodeTable.cpp
```diff
@@ -32,6 +33,30 @@ namespace
...
     boost::log::sources::severity_channel_logger_mt<>,
     (boost::log::keywords::severity = 0)(boost::log::keywords::channel = "discov"))
+
+struct DatagramTypeTableEntry
+{
+    char const* name;
+    uint8_t type;
+};
+
+DatagramTypeTableEntry datagramTypesTable[] = {
```
gumb0:
```suggestion
DatagramTypeTableEntry const c_datagramTypesTable[] = {
```

libp2p/NodeTable.cpp
```diff
@@ -32,6 +33,30 @@ namespace
...
+    {"PING", PingNode::type},
+    {"PONG", Pong::type},
+    {"FINDNODE", FindNode::type},
+    {"NEIGHBORS", Neighbours::type}
+};
+
+string datagramNameFromType(uint8_t _t)
```
gumb0: I'd suggest adding `typeName()` method to `RLPXDatagramFace` (and each implementation returning its name) instead of this. Then you won't need to handle unknown packet case and no need to throw.

libp2p/NodeTable.cpp
```diff
@@ -409,6 +434,7 @@ void NodeTable::onReceived(UDPSocketFace*, bi::udp::endpoint const& _from, bytes
             return;
         }
 
+        cnetdetails << "Received " << datagramNameFromType(packet->packetType()) << " from " << packet->sourceid << "@" << _from.address() << ":" << _from.port();
```
gumb0: I'd suggest outputting this and other your logs to `LOG(m_logger)` (it's lower verbosity (3 - Debug), but for me Debug verbosity is to log everything that happens, and 4 - Trace is for dumping into log actual binary contents of all data packets, and for tracing VM execution)

Also `m_logger` is more efficient than global `cnetdetails`, because it doesn't need to be thread-safe.

aleth-bootnode/main.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+int main(int argc, char** argv)
+{
+    setDefaultOrCLocale();
+
+    string const ProgramName = "aleth-bootnode";
+    string const NetworkConfigFileName = ProgramName + "-network.rlp";
+
+    /// Networking params.
+    string listenIP;
```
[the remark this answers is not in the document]
halfalicious: Got it - I do that if I'm going to be using them in multiple places, but I'll go ahead just declare them right before I use them.

libp2p/NodeTable.cpp
halfalicious: Good idea, I'll go ahead and do that

CMakeLists.txt
halfalicious: Thanks for the tip, done!

aleth-bootnode/main.cpp
halfalicious: Got it, that makes things simpler (no need for buildinfo.h). Thanks for the tip!

aleth/main.cpp
halfalicious: Confirmed - it's only used in the WebThree constructor to determine if the client should be configured / started and we always do this when running Aleth so it can be removed.

halfalicious:
> 
> 
> @halfalicious it builds for me without a problem. Finding `aleth-buildinfo.h` in VS might be tricky, but you can try finding `aleth-buildinfo` project in VS solution and/or check `build dir/aleth/` directory.

I always build from the console so it was pretty easy to find `aleth-buildinfo.h`..the problem was that there was no` aleth-bootnode-buildinfo.h`. It's a moot point now anyway since I don't need to compose the client version (since it's not used when performing discovery, only when establishing p2p connections)

halfalicious:
Note quite ready for merging yet - I want to add a log message for outgoing ping packets and take another look at the log output and compare it to Geth.

chfast:
> he problem was that there was no` aleth-bootnode-buildinfo.h`

There will be no `aleth-bootnode-buildinfo.h`. The prefix in `aleth-buildinfo.h` is for Aleth project and the same header is used by all the tools.

aleth-bootnode/main.cpp
```diff
@@ -0,0 +1,129 @@
...
+    unsigned short listenPort = c_defaultIPPort;
+    string publicIP;
+    bool upnp = true;
+
+    po::options_description generalOptions("GENERAL OPTIONS", lineWidth());
+    auto addGeneralOption = generalOptions.add_options();
+    addGeneralOption("help,h", "Show this help message and exit\n");
```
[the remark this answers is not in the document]
gumb0: Yeah I think it provides only the conversion of `options_description` to a string, but `-h` still need to be handled

aleth-bootnode/main.cpp
```diff
@@ -0,0 +1,150 @@
...
+    if (loggingOptions.verbosity > 0)
+        cout << EthGrayBold << c_programName << ", a C++ Ethereum bootnode implementation" EthReset
+             << "\n";
+
+    auto netPrefs = publicIP.empty() ? NetworkConfig(listenIP, listenPort, upnp) :
+                                       NetworkConfig(publicIP, listenIP, listenPort, upnp);
+    auto netData = contents(getDataDir() / fs::path(c_networkConfigFileName));
+
+    // Pass in empty client version since it's not used in the discovery process
+    Host h("", netPrefs, &netData);
```
gumb0: Still I think better to give `c_programName` as client name; if someone tries to handshake with us, they'll see it.

aleth-bootnode/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+set(
+    sources
+    main.cpp
+)
+
+add_executable(aleth-bootnode ${sources})
+target_link_libraries(
+    aleth-bootnode
+    PRIVATE devcore Boost::program_options p2p
```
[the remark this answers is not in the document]
halfalicious: For future reference, what's the general rule used for ordering the dependencies? Is it from most specialized to least specialized internal libraries followed by 3rd-party libraries?

halfalicious:
This should be good to go now...I addressed the remaining issues, added ping logging, and cleaned up the change history a bit and removed the merge commits. I also slightly simplified the discovery packet log statements (I didn't have to explicitly log the destination endpoint's port as the << operator is overloaded).

libp2p/NodeTable.cpp
```diff
@@ -299,7 +295,8 @@ void NodeTable::evict(shared_ptr<NodeEntry> _leastSeen, shared_ptr<NodeEntry> _n
 
     if (evicts == 1)
         doCheckEvictions();
-    ping(_leastSeen.get());
+    if (_leastSeen)
```
gumb0: I think this `if` is not needed, it's already checked at the calling side, and `_lastSeen` is already used above in this function anyway

aleth-bootnode/CMakeLists.txt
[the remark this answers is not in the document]
chfast: > most specialized to least specialized internal libraries followed by 3rd-party libraries?

Exactly. But this is not very important though.

libp2p/NodeTable.cpp
chfast: @gumb0 Can you fix it now? I'd like to make a build with aleth-bootnode included.

PR #5380, opened 2018-12-04
Validate PONGs

gumb0:
Addresses point 1 of https://github.com/ethereum/aleth/issues/4959
Saves info about every PING sent into `m_sentPings`; then validates each received PONG against the saved hash.

Next steps:
- more unit tests for timeouts and for evictions https://github.com/ethereum/aleth/pull/5390;
- Save last PONG time into `network.rlp` with each node (currently it trusts `network.rlp` to have valid nodes and doesn't re-PING them after aleth restart)

### DIFF
--- a/libdevcore/Assertions.h
+++ b/libdevcore/Assertions.h
@@ -25,7 +25,7 @@
 #pragma once
 
 #include <boost/exception/info.hpp>
-#include <iosfwd>
+#include <iostream>
 
 namespace dev
 {
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -29,16 +29,9 @@ BOOST_LOG_INLINE_GLOBAL_LOGGER_CTOR_ARGS(g_discoveryWarnLogger,
     boost::log::sources::severity_channel_logger_mt<>,
     (boost::log::keywords::severity = 0)(boost::log::keywords::channel = "discov"))
 
-const unsigned c_handleTimeoutsIntervalMs = 5000;
+constexpr unsigned c_handleTimeoutsIntervalMs = 5000;
 
 }  // namespace
-
-std::chrono::seconds const DiscoveryDatagram::c_timeToLive{60};
-std::chrono::milliseconds const NodeTable::c_evictionCheckInterval{75};
-std::chrono::milliseconds const NodeTable::c_reqTimeout{300};
-std::chrono::milliseconds const NodeTable::c_bucketRefresh{7200};
-uint32_t const NodeTable::c_bondingTimeSeconds = 12 * 60 * 60;
-
 
 inline bool operator==(
     std::weak_ptr<NodeEntry> const& _weak, std::shared_ptr<NodeEntry> const& _shared)
@@ -416,14 +409,14 @@ void NodeTable::onPacketReceived(
         {
             case Pong::type:
             {
-                auto pong = dynamic_cast<Pong const&>(*packet);
-                auto const sourceId = pong.sourceid;
+                auto const& pong = dynamic_cast<Pong const&>(*packet);
+                auto const& sourceId = pong.sourceid;
 
                 // validate pong
                 auto const sentPing = m_sentPings.find(sourceId);
                 if (sentPing == m_sentPings.end())
                 {
-                    LOG(m_logger) << "Unsolicited PONG from " << _from.address().to_string() << ":"
+                    LOG(m_logger) << "Unexpected PONG from " << _from.address().to_string() << ":"
                                   << _from.port();
                     return;
                 }
@@ -475,7 +468,7 @@ void NodeTable::onPacketReceived(
 
             case Neighbours::type:
             {
-                auto in = dynamic_cast<Neighbours const&>(*packet);
+                auto const& in = dynamic_cast<Neighbours const&>(*packet);
                 bool expected = false;
                 auto now = chrono::steady_clock::now();
                 DEV_GUARDED(x_findNodeTimeout)
@@ -502,7 +495,7 @@ void NodeTable::onPacketReceived(
 
             case FindNode::type:
             {
-                auto in = dynamic_cast<FindNode const&>(*packet);
+                auto const& in = dynamic_cast<FindNode const&>(*packet);
                 vector<shared_ptr<NodeEntry>> nearest = nearestNodeEntries(in.target);
                 static unsigned constexpr nlimit = (NodeSocket::maxDatagramSize - 109) / 90;
                 for (unsigned offset = 0; offset < nearest.size(); offset += nlimit)
@@ -519,7 +512,7 @@ void NodeTable::onPacketReceived(
 
             case PingNode::type:
             {
-                auto in = dynamic_cast<PingNode const&>(*packet);
+                auto& in = dynamic_cast<PingNode&>(*packet);
                 in.source.setAddress(_from.address());
                 in.source.setUdpPort(_from.port());
                 addNode(Node(in.sourceid, in.source));
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -33,6 +33,11 @@ constexpr unsigned c_handleTimeoutsIntervalMs = 5000;
 
 }  // namespace
 
+constexpr std::chrono::seconds DiscoveryDatagram::c_timeToLive;
+constexpr std::chrono::milliseconds NodeTable::c_reqTimeout;
+constexpr std::chrono::milliseconds NodeTable::c_bucketRefresh;
+constexpr std::chrono::milliseconds NodeTable::c_evictionCheckInterval;
+
 inline bool operator==(
     std::weak_ptr<NodeEntry> const& _weak, std::shared_ptr<NodeEntry> const& _shared)
 {
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -276,9 +276,10 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID _target)
     return ret;
 }
 
-void NodeTable::ping(NodeEntry const& _nodeEntry)
+void NodeTable::ping(NodeEntry const& _nodeEntry, boost::optional<NodeID> const& _replacementNodeID)
 {
-    m_timers.schedule(0, [this, _nodeEntry](boost::system::error_code const& _ec) {
+    m_timers.schedule(0, [this, _nodeEntry, _replacementNodeID](
+                             boost::system::error_code const& _ec) {
         if (_ec || m_timers.isStopped())
             return;
 
@@ -289,7 +290,7 @@ void NodeTable::ping(NodeEntry const& _nodeEntry)
 	    LOG(m_logger) << p.typeName() << " to " << _nodeEntry.id << "@" << p.destination;
         m_socket->send(p);
 
-        m_sentPings[_nodeEntry.id] = std::make_pair(chrono::steady_clock::now(), pingHash);
+        m_sentPings[_nodeEntry.id] = {chrono::steady_clock::now(), pingHash, _replacementNodeID};
     });
 }
 
@@ -297,18 +298,8 @@ void NodeTable::evict(NodeEntry const& _leastSeen, NodeEntry const& _new)
 {
     if (!m_socket->isOpen())
         return;
-    
-    unsigned evicts = 0;
-    DEV_GUARDED(x_evictions)
-    {
-        EvictionTimeout evictTimeout{_new.id, chrono::steady_clock::now()};  
-        m_evictions.emplace(_leastSeen.id, evictTimeout);
-        evicts = m_evictions.size();
-    }
 
-    if (evicts == 1)
-        doCheckEvictions();
-    ping(_leastSeen);
+    ping(_leastSeen, _new.id);
 }
 
 void NodeTable::noteActiveNode(Public const& _pubk, bi::udp::endpoint const& _endpoint)
@@ -426,37 +417,25 @@ void NodeTable::onPacketReceived(
                     return;
                 }
 
-                if (pong.echo != sentPing->second.second)
+                if (pong.echo != sentPing->second.pingHash)
                 {
                     LOG(m_logger) << "Invalid PONG from " << _from.address().to_string() << ":"
                                   << _from.port();
                     return;
                 }
 
-                m_sentPings.erase(sentPing);
                 auto const sourceNodeEntry = nodeEntry(sourceId);
                 assert(sourceNodeEntry.get());
                 sourceNodeEntry->lastPongReceivedTime = RLPXDatagramFace::secondsSinceEpoch();
 
-                // Whenever a pong is received, check if it's in m_evictions.
-                // Don't evict it if valid PONG received.
-                EvictionTimeout evictionEntry;
-                NodeID replacementNodeID;
-                DEV_GUARDED(x_evictions)
-                {
-                    auto e = m_evictions.find(sourceId);
-                    if (e != m_evictions.end() &&
-                        e->second.evictedTimePoint + c_reqTimeout >=
-                        std::chrono::steady_clock::now())
-                    {
-                        replacementNodeID = e->second.newNodeID;
-                        m_evictions.erase(e);
-                    }
-                }
-                // if we don't want to evict, we don't need to remember replacement anymore
-                if (replacementNodeID)
-                    if (auto replacementNode = nodeEntry(replacementNodeID))
+                // Valid PONG received, so we don't want to evict this node,
+                // and we don't need to remember replacement node anymore
+                auto const& optionalReplacementID = sentPing->second.replacementNodeID;
+                if (optionalReplacementID)
+                    if (auto replacementNode = nodeEntry(*optionalReplacementID))
                         dropNode(move(replacementNode));
+
+                m_sentPings.erase(sentPing);
 
                 // update our endpoint address and UDP port
                 DEV_GUARDED(x_nodes)
@@ -466,8 +445,6 @@ void NodeTable::onPacketReceived(
                         m_hostNode.endpoint.setAddress(pong.destination.address());
                     m_hostNode.endpoint.setUdpPort(pong.destination.udpPort());
                 }
-
-                LOG(m_logger) << "PONG from " << sourceId << " " << _from;
                 break;
             }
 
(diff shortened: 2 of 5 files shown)


lyy4j:
ok，I have doubts stilly。fistly, when the new node was put into the node talbe if the oldest node doesn't recevie a pong ,I can't find the  implement node。 secondly , if  current self node the never  receive the oldest node's pong reponse,  and it will never be remove  which in you branch

gumb0:
@lyy4j Thanks for looking into this!

> fistly, when the new node was put into the node talbe if the oldest node doesn't receive a pong in time ,I can't find the implement code

If the PONG is received but later than expected, it is discarded here
https://github.com/ethereum/aleth/blob/86234554bd6e91a9a958663d338e2eae3e4517db/libp2p/NodeTable.cpp#L394-L399
This part doesn't change in this PR

> secondly , if current self node never receive the oldest node's pong reponse, and will it never be remove 

If the evicted node doesn't get a valid PONG, it will be dropped in `doCheckEvictions` - is this what you're asking? This part doesn't change in this PR either

Not sure I understood your questions correctly, sorry.

To be clear: I plan to merge first your PR https://github.com/ethereum/aleth/pull/5377, which makes useful fixes to evictions mechanism, and after that rebase this one.

gumb0:
I realized it would be a good idea to drop the node in `doHandleTimeout` when no PONG was received after timeout (besides deleting the item from `m_sentPings`). I've added it in the last commit.
Maybe it is what you meant by your questions @lyy4j

chfast:
When ready?

gumb0:
@chfast today

gumb0:
Rebased.

libp2p/NodeTable.cpp
```diff
@@ -401,49 +418,60 @@ void NodeTable::onPacketReceived(
...
-                        if (e->second.evictedTimePoint + c_reqTimeout >=
-                            std::chrono::steady_clock::now())
-                        {
-                            found = true;
-                            leastSeenID = e->first;
-                            evictionEntry = e->second;
-                            m_evictions.erase(e);
-                        }
-                    }
+                auto pong = dynamic_cast<Pong const&>(*packet);
```
chfast: The packet is copied here.

libp2p/NodeTable.cpp
```diff
@@ -401,49 +418,60 @@ void NodeTable::onPacketReceived(
...
+                auto const sourceId = pong.sourceid;
+
+                // validate pong
+                auto const sentPing = m_sentPings.find(sourceId);
+                if (sentPing == m_sentPings.end())
+                {
+                    LOG(m_logger) << "Unsolicited PONG from " << _from.address().to_string() << ":"
```
chfast: Change "Unsolicited" to "Unexpected".

libp2p/NodeTable.cpp
```diff
@@ -401,49 +418,60 @@ void NodeTable::onPacketReceived(
...
+                                  << _from.port();
+                    return;
                 }
 
-                if (found)
+                if (pong.echo != sentPing->second.second)
```
chfast: In this case we should probably also remove it from `m_sentPings`.

libp2p/NodeTable.cpp
```diff
@@ -593,6 +618,29 @@ void NodeTable::doDiscovery()
...
+    m_timers.schedule(c_handleTimeoutsIntervalMs, [this](boost::system::error_code const& _ec) {
+        if ((_ec && _ec.value() == boost::asio::error::operation_aborted) || m_timers.isStopped())
+            return;
+
+        for (auto it = m_sentPings.begin(); it != m_sentPings.end();)
+        {
+            if (chrono::steady_clock::now() > it->second.first + DiscoveryDatagram::c_timeToLive)
+            {
+                if (auto node = nodeEntry(it->first))
+                    dropNode(node);
```
chfast: `dropNode()` looks to be quite inefficient by requiring `shared_ptr`. Still you can do `std::move()`.

libdevcore/Assertions.h
```diff
@@ -25,7 +25,7 @@
 #pragma once
 
 #include <boost/exception/info.hpp>
-#include <iosfwd>
+#include <iostream>
```
chfast: Why?

libp2p/NodeTable.cpp
```diff
@@ -32,8 +28,18 @@ namespace
 BOOST_LOG_INLINE_GLOBAL_LOGGER_CTOR_ARGS(g_discoveryWarnLogger,
     boost::log::sources::severity_channel_logger_mt<>,
     (boost::log::keywords::severity = 0)(boost::log::keywords::channel = "discov"))
+
+const unsigned c_handleTimeoutsIntervalMs = 5000;
+
 }  // namespace
 
+std::chrono::seconds const DiscoveryDatagram::c_timeToLive{60};
```
chfast: Can you make these `constexpr`?

libp2p/NodeTable.cpp
```diff
@@ -270,14 +279,21 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID _target)
...
-    PingNode p(src, _toEndpoint);
-    p.sign(m_secret);
-    LOG(m_logger) << "Sending " << p.typeName() << " to " << _toId << "@" << p.destination;
-    m_socketPointer->send(p);
+    m_timers.schedule(0, [this, _nodeEntry](boost::system::error_code const& _ec) {
+        if (_ec || m_timers.isStopped())
+            return;
+
+        NodeIPEndpoint src;
+        DEV_GUARDED(x_nodes) { src = m_hostNode.endpoint; }
```
chfast: Why do you have to lock this?

libp2p/NodeTable.cpp
```diff
@@ -270,14 +279,21 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID _target)
...
+        PingNode p(src, _nodeEntry.endpoint);
+        auto const pingHash = p.sign(m_secret);
+	    LOG(m_logger) << "Sending " << p.typeName() << " to " << _nodeEntry.id << "@" << p.destination;
```
chfast: Indentation is off.

libp2p/NodeTable.cpp
```diff
@@ -401,49 +418,60 @@ void NodeTable::onPacketReceived(
...
-                        n->pending = false;
+                    LOG(m_logger) << "Invalid PONG from " << _from.address().to_string() << ":"
+                                  << _from.port();
+                    return;
                 }
-                else
+
+                m_sentPings.erase(sentPing);
+                auto const sourceNodeEntry = nodeEntry(sourceId);
+                assert(sourceNodeEntry.get());
```
chfast: This assert fails.

libdevcore/Assertions.h
gumb0: Reordering of #includes somewhere (probably in tests) led to error here. `Assertions.h` uses `std::cerr`; `<iosfwd>` is not enough for it

libp2p/NodeTable.cpp
gumb0: Probably I don't anymore, I wanted to go over unnecessary locks later

libp2p/NodeTable.cpp
gumb0: They will be removed by timeout anyway, but maybe we could remove it here right away, too.

libp2p/NodeTable.cpp
gumb0: It shouldn't, because we checked that `m_sendPings` had this `sourceId` and we always add a node to `m_sentPings` after it is already in `m_allNodes`

chfast: > It shouldn't, because we checked that `m_sendPings` had this `sourceId` and we always add a node to `m_sentPings` after it is already in `m_allNodes`

Well...
```
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##8152c411…@35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Invalid PONG from 35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##8152c411…@35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Invalid PONG from 35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##8152c411…@35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Invalid PONG from 35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##8152c411…@35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Invalid PONG from 35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##d5f4476e…@149.28.150.71:30303
DEBUG 12-10 18:15:27 p2p  discov PONG from ##d5f4476e… 149.28.150.71:30303
DEBUG 12-10 18:15:27 p2p  discov Noting active node: ##d5f4476e… 149.28.150.71:30303
DEBUG 12-10 18:15:27 p2p  discov Sending Ping to ##8152c411…@35.188.130.212 UDP 30303 TCP 30303
DEBUG 12-10 18:15:27 p2p  discov Received Ping from ##d5f4476e…@149.28.150.71:30303
DEBUG 12-10 18:15:27 p2p  discov Sending Pong to ##d5f4476e…@149.28.150.71:30303
DEBUG 12-10 18:15:27 p2p  discov Noting active node: ##d5f4476e… 149.28.150.71:30303
DEBUG 12-10 18:15:27 p2p  discov Sending Ping to ##8152c411…@35.188.130.212 UDP 30303 TCP 30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##d9e5dc65…@150.109.58.135:30303
DEBUG 12-10 18:15:27 p2p  discov PONG from ##d9e5dc65… 150.109.58.135:30303
DEBUG 12-10 18:15:27 p2p  discov Noting active node: ##d9e5dc65… 150.109.58.135:30303
DEBUG 12-10 18:15:27 p2p  discov Sending Ping to ##8152c411…@35.188.130.212 UDP 30303 TCP 30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##67f8e81c…@52.83.138.158:30307
DEBUG 12-10 18:15:27 p2p  discov PONG from ##67f8e81c… 52.83.138.158:30307
DEBUG 12-10 18:15:27 p2p  discov Noting active node: ##67f8e81c… 52.83.138.158:30307
DEBUG 12-10 18:15:27 p2p  discov Sending Ping to ##8152c411…@35.188.130.212 UDP 30303 TCP 30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##8152c411…@35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Invalid PONG from 35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov p2p.nodes.drop ##8152c411…
DEBUG 12-10 18:15:27 p2p  discov Noting active node: ##b57b2647… 46.101.253.211:30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##8152c411…@35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Invalid PONG from 35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##8152c411…@35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Invalid PONG from 35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##8152c411…@35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Invalid PONG from 35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##8152c411…@35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Invalid PONG from 35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##8152c411…@35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Invalid PONG from 35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##8152c411…@35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov Invalid PONG from 35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  net    p2p.host.nodeTable.events.NodeEntryDropped ##8152c411…
DEBUG 12-10 18:15:27 p2p  net    p2p.host.nodeTable.events.nodeEntryAdded ##b57b2647…
DEBUG 12-10 18:15:27 p2p  net    p2p.host.peers.events.peerAdded ##b57b2647… 46.101.253.211 UDP 30303 TCP 30303
DEBUG 12-10 18:15:27 p2p  discov Received Pong from ##8152c411…@35.188.130.212:30303
DEBUG 12-10 18:15:27 p2p  discov ASSERT FAILURE 35.188.130.212:30303
aleth-bootnode: /home/chfast/Projects/ethereum/aleth/libp2p/NodeTable.cpp:446: virtual void dev::p2p::NodeTable::onPacketReceived(dev::p2p::UDPSocketFace*, const endpoint&, dev::bytesConstRef): Assertion `sourceNodeEntry.get()' failed.
```

gumb0: That's interesting, the same node sends us several invalid PONGs, and then the valid one...

libp2p/NodeTable.h
```diff
@@ -177,26 +177,26 @@ class NodeTable : UDPSocketEvents
 protected:
     /// Constants for Kademlia, derived from address space.
 
-    static unsigned const s_addressByteSize = h256::size;					///< Size of address type in bytes.
-    static unsigned const s_bits = 8 * s_addressByteSize;					///< Denoted by n in [Kademlia].
-    static unsigned const s_bins = s_bits - 1;  ///< Size of m_buckets (excludes root, which is us).
-    static unsigned const s_maxSteps = boost::static_log2<s_bits>::value;	///< Max iterations of discovery. (discover)
+    static constexpr unsigned s_addressByteSize = h256::size;					///< Size of address type in bytes.
```
chfast: Unfortunately, even when they are `constexpr` now you still have to allocate memory for them. This is some kind of C++11 bug, fixed in newer revisions.

Easy way to workaround this, is to move them out of class to .cpp file. This also saves us including `boost::static_log2` in the header file.

libp2p/NodeTable.cpp
gumb0: So I think what happens is the node is being evicted via eviction mechanism and then the problem is its record in `m_sentPings` is not deleted at the same time.

Now I realize that eviction and timeout handling is basically the same thing. Both drop a node after PING timeouts, the only difference that in eviction the new node is added after dropping the old one.
Maybe I'll try to simplify it by unifying them...

chfast: Good idea.

libp2p/NodeTable.h
gumb0: There's `m_buckets` member definition that depends on `s_bins`, so I can't just move it...

gumb0: I've added definitions to cpp for the ones clang was complaining about

libp2p/NodeTable.cpp
chfast: We have to wait for this fix, right?

gumb0: Yes, this currently fails for me, too

gumb0:
For some unknown reason I don't see any log output anymore when running under VS debugger, this sucks...
Running debug build without debugger outputs log as usual.

libp2p/NodeTable.h
```diff
@@ -257,6 +254,8 @@ class NodeTable : UDPSocketEvents
     /// Looks up a random node at @c_bucketRefresh interval.
     void doDiscovery();
 
+    void doHandleTimeouts();
```
subtly: This is handled by evictions in the discovery process and the discovery process which is every 7200ms.

libp2p/NodeTable.cpp
subtly: > So I think what happens is the node is being evicted via eviction mechanism and then the problem is its record in `m_sentPings` is not deleted at the same time.
> 
> Now I realize that eviction and timeout handling is basically the same thing. Both drop a node after PING timeouts, the only difference that in eviction the new node is added after dropping the old one.
> Maybe I'll try to simplify it by unifying them...

This is exactly the case. Its based on kademlia paper and algorithm and there only needs to be one discovery process and separate timers/expirations aren't necessary.

gumb0:
> Its very critical not to respond to any packets except Ping if the node isn't in the node table. 

We plan add the checks for other requests, as outlined in points 2-4 of https://github.com/ethereum/aleth/issues/4959.
I'm limiting the scope of this particular PR, because it's already quite big.

> So, what you are working on here must be handled by whether or not the node is in the node table, rather than timeouts. 

The node is not going to be considered valid unless we received the valid PONG from it in the last 12 hours. That is, nodes in the node table can be both valid and not valid anymore. The validity check will be looking at the time of last pong receive of the node.

Moreover, we don't place the newly added node into the node table until we receive a valid Pong from it - it was actually like this even before my changes.
Otherwise some of our `Neigbours` responses might include not-yet-validated nodes.

> The exception is the Pong packet, which should only be sent or received when an appropriate Ping packet was received or sent; and so, this is covered by a check as to whether a Ping packet was sent to begin with. 

Before my changes there was no such check other than for the evicted nodes.

> I'd recommend reconsidering your approach and changes here. Happy to help and chat about it!

Why exactly do you think it should be reconsidered, if still? Where does it break?

> (PS: I'm the author you're removing)

I know. We're removing it everywhere, because it's not helpful in any way.

libp2p/NodeTable.cpp
```diff
@@ -14,10 +14,6 @@
  You should have received a copy of the GNU General Public License
  along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
  */
-/** @file NodeTable.cpp
```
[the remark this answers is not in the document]
gumb0: All this info is stored by git.

libp2p/NodeTable.cpp
gumb0: Right. Yet there's still one old separate timeout mechanism for `FindNode` requests. Maybe we'll simplify it later...

gumb0: @chfast Please check this with the latest changes, should be fixed.

libp2p/NodeTable.h
gumb0: No more separate evictions mechanism...

Before my changes the nodes that didn't answer to the initial Ping were never deleted (from `m_allNodes`)

gumb0: By the way, why 7200 ms?

libp2p/NodeTable.cpp
chfast: I did some commits ago. It's fixed.

halfalicious:
> 
> 
> For some unknown reason I don't see any log output anymore when running under VS debugger, this sucks...
> Running debug build without debugger outputs log as usual.

I think that's an issue with your VS, I built master and saw log output in both VS 2015 and WinDBG.

libp2p/NodeTable.cpp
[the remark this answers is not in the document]
halfalicious: Should we also replace cpp-ethereum with aleth?

PR #5382, opened 2018-12-05
remove strictness::JustSeal

winsvega:
remove JustSeal
fix tests mining uncle issue

### DIFF
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -654,7 +654,7 @@ void overwriteBlockHeaderForTest(mObject const& _blObj, TestBlock& _block, Chain
             ho.count("number") ? toU256(ho["number"]) : header.number(),
             ho.count("gasLimit") ? toU256(ho["gasLimit"]) : header.gasLimit(),
             ho.count("gasUsed") ? toU256(ho["gasUsed"]) : header.gasUsed(),
-            ho.count("timestamp") ? toU256(ho["timestamp"]) : header.timestamp() >= 0 ? header.timestamp() : utcTime(),
+            ho.count("timestamp") ? toU256(ho["timestamp"]) : header.timestamp(),
             ho.count("extraData") ? importByteArray(ho["extraData"].get_str()) :
                                     header.extraData());
 
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -654,7 +654,7 @@ void overwriteBlockHeaderForTest(mObject const& _blObj, TestBlock& _block, Chain
             ho.count("number") ? toU256(ho["number"]) : header.number(),
             ho.count("gasLimit") ? toU256(ho["gasLimit"]) : header.gasLimit(),
             ho.count("gasUsed") ? toU256(ho["gasUsed"]) : header.gasUsed(),
-            ho.count("timestamp") ? toU256(ho["timestamp"]) : header.timestamp(),
+            ho.count("timestamp") ? toU256(ho["timestamp"]) : header.timestamp() >= 0 ? header.timestamp() : utcTime(),
             ho.count("extraData") ? importByteArray(ho["extraData"].get_str()) :
                                     header.extraData());
 
@@ -780,6 +780,7 @@ void overwriteUncleHeaderForTest(mObject& uncleHeaderObj, TestBlock& uncle, std:
         {
             uncleHeader.setTimestamp(importedBlocks.at(number).blockHeader().timestamp() + 1);
             sealEngine->populateFromParent(uncleHeader, importedBlocks.at(number).blockHeader());
+            uncleHeader.setAuthor(h160::random()); // make each populated block unique
 
             //Set Default roots for empty block
             //m_transactionsRoot = _t; m_receiptsRoot = _r; m_sha3Uncles = _u; m_stateRoot = _s;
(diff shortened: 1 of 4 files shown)


test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -654,7 +654,7 @@ void overwriteBlockHeaderForTest(mObject const& _blObj, TestBlock& _block, Chain
             ho.count("number") ? toU256(ho["number"]) : header.number(),
             ho.count("gasLimit") ? toU256(ho["gasLimit"]) : header.gasLimit(),
             ho.count("gasUsed") ? toU256(ho["gasUsed"]) : header.gasUsed(),
-            ho.count("timestamp") ? toU256(ho["timestamp"]) : header.timestamp(),
+            ho.count("timestamp") ? toU256(ho["timestamp"]) : header.timestamp() >= 0 ? header.timestamp() : 1000,
```
gumb0: What's wrong with timestamp 0?

winsvega: header.timestamp() could be negative which is incorrect and breaks the test generation.

test/tools/libtesteth/BlockChainHelper.cpp
```diff
@@ -222,7 +222,7 @@ void TestBlock::mine(TestBlockChain const& _bc)
         // mining!");
         cnote << "Mining block difficulty of:  " << block.info().difficulty();
         dev::eth::mine(block, blockchain, blockchain.sealEngine());
-        blockchain.sealEngine()->verify(JustSeal, block.info());
+        blockchain.sealEngine()->verify(CheckNothingNew, block.info());
```
chfast: What this change does?

winsvega: I told you to change timestamp in blockheader. nobody agrees. now you ask me why and nobody unserstands. timestamp in blockheader has value of signed int. and it is incorrect by default when you create the blockheader. 

The intention of this check was to validate that that blocheader is constructed correct (but without signature). however if it throws then mining does not work. I guess this line could be removed, but who knows what might happen. I could try to remove line 225 but that will require to check that BC test filling still work.

gumb0: @chfast This was my suggestion to remove `JustSeal`, because it is not being handled correctly, and it's used only in json tests, but for them using `CheckNothingNew` is also fine

chfast:
For "JustSeal" it can go. For timestamp try this fix first: https://github.com/ethereum/aleth/pull/5383.

test/tools/jsontests/BlockChainTests.cpp
gumb0: If it's not RLP parsing as Pawel suggests, maybe you should change `TestBlock` so that it produces a valid header.

winsvega: perhaps in NoProof::generateSeal ?

winsvega: nope that does not work

winsvega: So I have a question then. if blockheader by default is invalid. how I make it valid?

chfast: Any seal engine should not seal an invalid block nor transform an invalid one into a valid one.

To make it valid set the timestamp proper value (a number >= 0).

winsvega: that exactly what I do at line 657

winsvega: It's `invalid` because it is invalid by default. what number should I use?

chfast: 0?

winsvega: testeth transform blockheader before mining for testing purposes.

winsvega: 0 and utcTime() both works.

winsvega:
rebased on master branch

winsvega:
@chfast, anything else to this PR?

chfast:
How about like this?

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -780,7 +780,6 @@ void overwriteUncleHeaderForTest(mObject& uncleHeaderObj, TestBlock& uncle, std:
         {
             uncleHeader.setTimestamp(importedBlocks.at(number).blockHeader().timestamp() + 1);
             sealEngine->populateFromParent(uncleHeader, importedBlocks.at(number).blockHeader());
-            uncleHeader.setAuthor(h160::random()); // make each populated block unique
```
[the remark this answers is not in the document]
chfast: Sorry. Added it back.

But do you really want to generate different test each time you run the generator?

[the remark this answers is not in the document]
winsvega: hmm. it is different anyway because build info is updated. but you right maybe a better solution would be to use determenistic addresses. like 10000..01, 10000...02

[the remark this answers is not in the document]
winsvega: test filling does not work after your changes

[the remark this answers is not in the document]
winsvega: no it is not related to timestamp issue. it is some other issue. it happens some time for some reason. but other time test filling works. strange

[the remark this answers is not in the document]
winsvega: so test is different all times anyway. timestamp is different when you regenerate the test. so blockhash would be different. different uncle author or static - the hash would be different anyway.

winsvega:
(could be merged)

PR #5383, opened 2018-12-06
Fix handling of negative header timestamp

chfast:


gumb0:
> in libethcore/BlockHeader.h :: 203
m_timestamp = -1 by default.

@winsvega It looks to be expected, `BlockHeader` is in invalid state by default.
(there's even a method `clear()` which puts it back into invalid state, so I guess it needs to have an invalid state for some reason)

winsvega:
Maybe timestamp should be set to valid in NoProof generate seal?

PR #5389, opened 2018-12-10
Discovery cleanups and logging improvements

chfast:
No functional changes expected.

### DIFF
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -103,7 +103,7 @@ void NodeTable::addNode(Node const& _node, NodeRelation _relation)
 
     auto ret = make_shared<NodeEntry>(m_hostNode.id, _node.id, _node.endpoint);
     DEV_GUARDED(x_nodes) { m_allNodes[_node.id] = ret; }
-    LOG(m_logger) << "addNode pending for " << _node.id << "@" << _node.endpoint;
+    LOG(m_logger) << "Pending node " << _node.id << "@" << _node.endpoint;
     ping(_node.id, _node.endpoint);
 }
 
@@ -426,15 +426,12 @@ void NodeTable::onPacketReceived(
                 {
                     if (auto n = nodeEntry(evictionEntry.newNodeID))
                         dropNode(n);
-                    if (auto n = nodeEntry(leastSeenID))
-                        n->pending = false;
                 }
-                else
-                {
-                    if (auto n = nodeEntry(in.sourceid))
-                        n->pending = false;
-                }
-                
+
+                auto n = nodeEntry(found ? leastSeenID : in.sourceid);
+                if (n && n->pending)
+                    n->pending = false;
+
                 // update our endpoint address and UDP port
                 DEV_GUARDED(x_nodes)
                 {
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -306,8 +306,7 @@ void NodeTable::noteActiveNode(Public const& _pubk, bi::udp::endpoint const& _en
     shared_ptr<NodeEntry> newNode = nodeEntry(_pubk);
     if (newNode && !newNode->pending)
     {
-        LOG(m_logger) << "Noting active node: " << _pubk << " " << _endpoint.address().to_string()
-                      << ":" << _endpoint.port();
+        LOG(m_logger) << "Active node " << _pubk << '@' << _endpoint;
         newNode->endpoint.setAddress(_endpoint.address());
         newNode->endpoint.setUdpPort(_endpoint.port());
 
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -442,22 +442,22 @@ void NodeTable::onPacketReceived(
 
                 break;
             }
-                
+
             case Neighbours::type:
             {
                 auto in = dynamic_cast<Neighbours const&>(*packet);
                 bool expected = false;
                 auto now = chrono::steady_clock::now();
                 DEV_GUARDED(x_findNodeTimeout)
-                m_findNodeTimeout.remove_if([&](NodeIdTimePoint const& t) noexcept {
-                    if (t.first == in.sourceid)
-                    {
-                        if (now - t.second < c_reqTimeout)
+                {
+                    m_findNodeTimeout.remove_if([&](NodeIdTimePoint const& _t) noexcept {
+                        if (_t.first != in.sourceid)
+                            return false;
+                        if (now - _t.second < c_reqTimeout)
                             expected = true;
                         return true;
-                    }
-                    return false;
-                });
+                    });
+                }
                 if (!expected)
                 {
                     cnetdetails << "Dropping unsolicited neighbours packet from "
@@ -465,7 +465,7 @@ void NodeTable::onPacketReceived(
                     break;
                 }
 
-                for (auto const& n: in.neighbours)
+                for (auto const& n : in.neighbours)
                     addNode(Node(n.node, n.endpoint));
                 break;
             }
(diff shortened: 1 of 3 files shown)


libp2p/NodeTable.cpp
```diff
@@ -46,20 +46,18 @@ NodeTable::NodeTable(
     ba::io_service& _io, KeyPair const& _alias, NodeIPEndpoint const& _endpoint, bool _enabled)
   : m_hostNode(Node(_alias.pub(), _endpoint)),
     m_secret(_alias.secret()),
-    m_socket(make_shared<NodeSocket>(
-        _io, *reinterpret_cast<UDPSocketEvents*>(this), (bi::udp::endpoint)m_hostNode.endpoint)),
-    m_socketPointer(m_socket.get()),
+    m_socket(make_shared<NodeSocket>(_io, *this, (bi::udp::endpoint)m_hostNode.endpoint)),
```
gumb0: I would prefer to keep the inheritance from `UDPSocketEvents` private.
Then bacause of `make_shared` you would still need a cast here, but `static_cast` will do.
Also `m_socket` actually can be made `unique_ptr` I think, and then there'll be no `make_shared` and no cast here if you just use `new`.

libp2p/NodeTable.cpp
```diff
@@ -177,9 +175,9 @@ void NodeTable::doDiscover(NodeID _node, unsigned _round, shared_ptr<set<shared_
             FindNode p(r->endpoint, _node);
             p.sign(m_secret);
             DEV_GUARDED(x_findNodeTimeout)
-                m_findNodeTimeout.push_back(make_pair(r->id, chrono::steady_clock::now()));
-            LOG(m_logger) << "Sending " << p.typeName() << " to " << _node << "@" << r->endpoint;
-            m_socketPointer->send(p);
+                m_findNodeTimeout.emplace_back(make_pair(r->id, chrono::steady_clock::now()));
```
gumb0: Should work without `make_pair`

libdevcore/Log.h
```diff
@@ -214,27 +214,6 @@ inline boost::log::formatting_ostream& operator<<(
     _strm.stream() << EthCyan "#" << _value.abridged() << EthReset;
     return _strm;
 }
-inline boost::log::formatting_ostream& operator<<(
```
gumb0: I remember that it was not working correctly for non-const variables without this overload https://github.com/ethereum/aleth/commit/d0af9c2cdc30d94084450481ffdc971afd74abfb

libp2p/NodeTable.cpp
```diff
@@ -453,22 +448,23 @@ void NodeTable::onPacketReceived(
                 bool expected = false;
                 auto now = chrono::steady_clock::now();
                 DEV_GUARDED(x_findNodeTimeout)
-                    m_findNodeTimeout.remove_if([&](NodeIdTimePoint const& t)
+                m_findNodeTimeout.remove_if([&](NodeIdTimePoint const& t) noexcept {
```
gumb0: I usually add braces around the statement under `DEV_GUARDED` to coerce clang-format to better indentation.
Maybe put lambda to a separate variable to simplify this.

libp2p/NodeTable.cpp
```diff
@@ -453,22 +448,23 @@ void NodeTable::onPacketReceived(
...
+                });
                 if (!expected)
                 {
                     cnetdetails << "Dropping unsolicited neighbours packet from "
                                 << _from.address();
                     break;
                 }
 
-                for (auto n: in.neighbours)
+                for (auto const& n: in.neighbours)
```
gumb0: this was not clang-formatted

libp2p/NodeTable.cpp
gumb0: Comment to `m_socket` mentions that `shared_ptr` is required by boost.asio, but I don't believe this is true. The only sense in which I see it could be "required" is `shared_ptr` could have been used to avoid async handlers accessing already deleted variables.
But nowhere we capture this `shared_ptr` by value, and anyway `p2p::Host` takes care (at least tries) to stop IO service and all handlers before destroying `NodeTable` and other stuff.

chfast: Unfortunately the comment is true. ASIO takes a weak_ptr of it somehow.

libdevcore/Log.h
```diff
@@ -214,27 +214,6 @@ inline boost::log::formatting_ostream& operator<<(
...
-    boost::log::formatting_ostream& _strm, h256& _value)
-{
-    auto const& constValue = _value;
-    _strm << constValue;
-    return _strm;
-}
-
-inline boost::log::formatting_ostream& operator<<(
```
gumb0: Removing these probably will result in generic `FixedHash` function used which outputs it with `#` instead of `##`

halfalicious:
Can you also simplify this log statement while you're at it 😃 https://github.com/ethereum/aleth/blob/dc0b5a2f706333951e94a56453186849a41a03ac/libp2p/NodeTable.cpp#L310-L311

You can just log `_endpoint`, you don't need to log both the address and port explicitly since the overloaded << operator will do that for you. I think it would also be good to include an "@" between `_pubk `and `_endpoint`.

halfalicious:
Out of curiosity, what was the point of having both `m_socket` and the raw pointer version `m_socketPointer`:
https://github.com/ethereum/aleth/blob/dc0b5a2f706333951e94a56453186849a41a03ac/libp2p/NodeTable.h#L278-L279

libdevcore/Log.h
chfast: Originally, I thought the `##` was a bug. That's why this change.

libdevcore/Log.h
chfast: It builds now.

libp2p/NodeTable.cpp
chfast: Why do you want to have `UDPSocketEvents` private?

chfast:
> Out of curiosity, what was the point of having both `m_socket` and the raw pointer version `m_socketPointer`:

I assume someone wrongly assumed that accessing `shared_ptr` has some additional overhead.

libdevcore/Log.h
gumb0: It was calling the wrong `operator<<` (generic `operator<<(formatting_ostream&, T&)` which ouput it differently from the const version)

chfast: Ok, I only removed the h512 overloads.

libdevcore/Log.h
gumb0: Why do you still remove these? it seems useful to see that `##38db89c6…` and `#bfd15dbb…` are actually different types of hashes

chfast: They have different color, isn't this enough?

libp2p/NodeTable.cpp
gumb0: Because deriving it is an implementation detail; the users of `NodeTable` don't need to know about it.

libp2p/NodeTable.cpp
```diff
@@ -446,29 +440,32 @@ void NodeTable::onPacketReceived(
...
                 bool expected = false;
                 auto now = chrono::steady_clock::now();
                 DEV_GUARDED(x_findNodeTimeout)
-                    m_findNodeTimeout.remove_if([&](NodeIdTimePoint const& t)
-                    {
-                        if (t.first == in.sourceid && now - t.second < c_reqTimeout)
-                            expected = true;
-                        else if (t.first == in.sourceid)
+                {
+                    m_findNodeTimeout.remove_if([&](NodeIdTimePoint const& t) noexcept {
```
gumb0: Parameter should be `_t`

libp2p/NodeTable.cpp
```diff
@@ -446,29 +440,32 @@ void NodeTable::onPacketReceived(
...
+                        if (t.first == in.sourceid)
```
gumb0: Would look a bit nicer if you return early `if (t.first != in.sourceid)`

chfast:
Fixed.

PR #5390, opened 2018-12-10
More Discovery unit tests

gumb0:
This adds a couple of tests for eviction/timeout mechanism, and I had to make timeout interval configurable in `NodeTable` for this.

### DIFF
--- a/test/unittests/libp2p/net.cpp
+++ b/test/unittests/libp2p/net.cpp
@@ -165,6 +165,7 @@ struct TestNodeTable: public NodeTable
     using NodeTable::m_sentPings;
     using NodeTable::m_socket;
     using NodeTable::noteActiveNode;
+    using NodeTable::setRequestTimeToLive;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


test/unittests/libp2p/net.cpp
```diff
@@ -430,6 +450,144 @@ BOOST_AUTO_TEST_CASE(noteActiveNodeReplacesNodeInFullBucketWhenEndpointChanged)
...
+    nodeSocketHost.start();
+    uint16_t nodePort = nodeSocketHost.port;
+
+    // add a node to node table, initiating PING
+    auto nodeEndpoint = NodeIPEndpoint{bi::address::from_string("127.0.0.1"), nodePort, nodePort};
+    auto nodeKeyPair = KeyPair::create();
+    auto nodePubKey = nodeKeyPair.pub();
+    nodeTable->addNode(Node{nodePubKey, nodeEndpoint});
+
+    this_thread::sleep_for(DiscoveryDatagram::c_timeToLive + std::chrono::seconds(5));
```
gumb0: Not sure if it's acceptable for a test to sleep for 25 seconds... I'll probably try to make it configurable

PR #5393, opened 2018-12-12
Add missing header guards

axic:


axic:
Appveyor doesn't seem to be a build failure.

PR #5396, opened 2018-12-12
Use dopple instead of jsonrpcproxy

### LINKED ISSUE #5324: Replace scripts/jsonrpcproxy.py with dopple
Dopple is ["maintained"](https://github.com/ethereum/dopple/issues/1) at: https://github.com/ethereum/dopple

It would make sense to import it as a git submodule or subtree and be used that way.

axic:
Fixes #5324.


gumb0:
No idea either.

axic:
@gumb0, @chfast mentioned that you use dopple for something. I think this docker image is still working after the change, but I haven't tried it.

gumb0:
I use it for ethstats reporting; I can try it from this PR.

scripts/dopple.py
```diff
@@ -0,0 +1 @@
+dopple/dopple.py
```
gumb0: does it skip command line arguments?

axic: This is a symlink, github shows it strangely.

axic:
@gumb0 @chfast anything blocking this?

chfast:
This is also used in docker images.

axic:
@chfast @gumb0 so should we merge? 😉 Please feel free to merge in that case.

chfast:
Does the code is exactly the same? If yes, we can merge it. If not I have to verify it works in docker images as they are used in the Hive testing.

axic:
> Does the code is exactly the same?

I've updated the other repo to match the on here when I made the PR. If there were changes to jsonrpcproxy in this repo since then, those aren't mirrored.

PR #5397, opened 2018-12-12
Reorder code (compileLLL) and leave comment

axic:


test/tools/libtesteth/TestHelper.cpp
```diff
@@ -497,9 +497,9 @@ string compileLLL(string const& _code)
     return "";
 #else
     fs::path path(fs::temp_directory_path() / fs::unique_path());
-    string cmd = string("lllc ") + path.string();
     writeFile(path.string(), _code);
-    string result = executeCmd(cmd);
+    // NOTE: this will abort if execution failed
+    string result = executeCmd(string("lllc ") + path.string());
```
winsvega: looks like nothing really changed. 
why do you need to fix this all of a suden?

axic: There is no fix, just some tiny cleanup. I was looking at the code whether it still requires `lllc` to present or not. I had to go through several functions to understand what it does, hence the comment.

Answer to my question: `lllc` is not required anymore.

PR #5398, opened 2018-12-12
Store binaryenVersion in test outputs

axic:
Will create a PR next which moves binaryen into an external binary (just like lllc).

### DIFF
--- a/test/tools/libtesteth/TestHelper.cpp
+++ b/test/tools/libtesteth/TestHelper.cpp
@@ -637,11 +637,10 @@ string prepareLLLCVersionString()
     return "Error getting LLLC Version";
 }
 
+#define STR(X) #X
 string prepareBinaryenVersionString()
 {
-    // NOTE: keep this in-sync with cmake/ProjectBinaryen.cmake
-    // FIXME: there is no internal API to retrieve this...
-    return "1.37.35";
+    return STR(BINARYEN_VERSION);
 }
 
 void copyFile(fs::path const& _source, fs::path const& _destination)
--- a/test/tools/libtesteth/TestHelper.cpp
+++ b/test/tools/libtesteth/TestHelper.cpp
@@ -637,6 +637,13 @@ string prepareLLLCVersionString()
     return "Error getting LLLC Version";
 }
 
+string prepareBinaryenVersionString()
+{
+    // NOTE: keep this in-sync with cmake/ProjectBinaryen.cmake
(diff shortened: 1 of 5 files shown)


test/tools/libtesteth/TestHelper.cpp
```diff
@@ -632,6 +632,13 @@ string prepareLLLCVersionString()
     return "Error getting LLLC Version";
 }
 
+string prepareBinaryenVersionString()
+{
+    // NOTE: keep this in-sync with cmake/ProjectBinaryen.cmake
+    // FIXME: there is no internal API to retrieve this...
+    return "1.37.35";
```
winsvega: there must be a way to prepare this with cmake.

test/tools/libtesteth/TestSuite.cpp
```diff
@@ -98,6 +98,7 @@ void addClientInfo(json_spirit::mValue& _v, fs::path const& _testSource, h256 co
 
 		clientinfo["filledwith"] = test::prepareVersionString();
 		clientinfo["lllcversion"] = test::prepareLLLCVersionString();
+		clientinfo["binaryenVersion"] = test::prepareBinaryenVersionString();
```
winsvega: please explain more who binaryen works and how it's related to evm.

axic: It used for ewasm tests. Binaryen is used in testeth for at least 10 months.

winsvega: could not this information somehow be retrived from aleth version ?

winsvega:
I still dont think it is a good idea to store yet another version info in tests. especially in case other clients would not have binaryen to provide this version info

axic:
The next PR is decoupling binaryen, e.g. it will not be part of testeth anymore.

test/tools/libtesteth/TestSuite.cpp
axic: It could right now. However I've stated the goal in the description of the PR:
> Will create a PR next which moves binaryen into an external binary (just like lllc).

Basically in the future it shouldn't have a built in binaryen, just rely on an external tool.

chfast:
I took the version number from CMake script. Not ideal, but better than it was. Merging this...

winsvega:
why do we need binaryen version string in the tests? 
cmake script does not work

PR #5400, opened 2018-12-15
Minor code cleanup in Host

halfalicious:
Made some (very) minor changes while going through the Host code, should have no functional impact: 
* Add constexpr
* Add some minor comments
* Remove unnecessary ReputationManager constructor
* Use boost::asio namespace alias
* Update license messages

### DIFF
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -1,18 +1,18 @@
 /*
-    This file is part of cpp-ethereum.
+    This file is part of aleth.
 
-    cpp-ethereum is free software: you can redistribute it and/or modify
+    aleth is free software: you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
     the Free Software Foundation, either version 3 of the License, or
     (at your option) any later version.
 
-    cpp-ethereum is distributed in the hope that it will be useful,
+    aleth is distributed in the hope that it will be useful,
     but WITHOUT ANY WARRANTY; without even the implied warranty of
     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
     GNU General Public License for more details.
 
     You should have received a copy of the GNU General Public License
-    along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+    along with aleth.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 #include "Host.h"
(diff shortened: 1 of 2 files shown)


libp2p/Host.cpp
```diff
@@ -1,18 +1,18 @@
 /*
-    This file is part of cpp-ethereum.
+    This file is part of aleth.
```
chfast: Please see #5403.

libp2p/Host.cpp
```diff
@@ -99,7 +97,7 @@ Host::Host(string const& _clientVersion, KeyPair const& _alias, NetworkConfig co
     m_clientVersion(_clientVersion),
     m_netConfig(_n),
     m_ifAddresses(Network::getInterfaceAddresses()),
-    m_ioService(2),
+    m_ioService(2), // concurrency hint, suggests how many threads it should allow to run simultaneously
```
chfast: Use clang-format.

libp2p/Host.cpp
```diff
@@ -705,7 +703,7 @@ void Host::run(boost::system::error_code const&)
     DEV_GUARDED(x_connecting)
         m_connecting.remove_if([](std::weak_ptr<RLPXHandshake> h){ return h.expired(); });
     DEV_GUARDED(x_timers)
-    m_timers.remove_if([](std::unique_ptr<boost::asio::deadline_timer> const& t) {
+    m_timers.remove_if([](std::unique_ptr<ba::deadline_timer> const& t) {
```
chfast: How about we use `using io = boost.asio`?

libp2p/Host.cpp
```diff
@@ -757,6 +755,8 @@ void Host::run(boost::system::error_code const&)
     m_timer->async_wait(runcb);
 }
 
+// Called after thread has been started to perform additional class-specific state initialization (e.g. start
```
chfast: Maybe move the comment inside the function. This looks like doxygen comment.

libp2p/Host.cpp
```diff
@@ -767,7 +767,7 @@ void Host::startedWorking()
         // time, stop will wait on m_timer and graceful network shutdown.
         Guard l(x_runTimer);
         // create deadline timer
-        m_timer.reset(new boost::asio::deadline_timer(m_ioService));
+        m_timer.reset(new ba::deadline_timer(m_ioService));
```
chfast: Do we need heap allocated timers? They cannot be used as values?

libp2p/Host.h
```diff
@@ -87,7 +87,7 @@ struct Reputation
 class ReputationManager
 {
 public:
-    ReputationManager();
+    ReputationManager() = default;
```
chfast: Is this needed at all?

libp2p/Host.cpp
gumb0: It looks we destroy and recreate it when the network is stopped / restarted, but I'm not sure if that's actually needed.

libp2p/Host.h
halfalicious: Ah you're right, C++ default constructors are declared & defined implicitly so we don't need this.

libp2p/Host.cpp
halfalicious: I like doxygen comments 😃 We have other functions that are commented like that e.g. https://github.com/ethereum/aleth/blob/b92056768376d42b809e2a80afd3178b2dead29f/libp2p/Common.cpp#L48-L51

Also, Geth functions are commented like that e.g: https://github.com/ethereum/go-ethereum/blob/472c23a8015cd84b193d2d0efb4592a664de3c62/internal/ethapi/addrlock.go#L43-L48

libp2p/Host.cpp
halfalicious: I'd like to leave the timer alone for these changes (they were meant to be trivial changes with absolutely no potential for functional impact), I'd be happy to look at it later.

libp2p/Host.cpp
```diff
@@ -759,6 +745,8 @@ void Host::run(boost::system::error_code const&)
 
 void Host::startedWorking()
 {
+    // Called after thread has been started to perform additional class-specific state initialization (e.g. start
+// capability threads, start TCP listener, and kick off timers)
```
chfast: Formatting off here

halfalicious: Thanks, will fix before merging

Issue #5391, opened 2018-12-11
Aleth bootnode improvements


chfast:
- [x] Does the bootnode need to open the TCP port?
- [ ] It runs with very low resources, but the memory usage is growing slightly over time. Check if we don't keep some data forever (like pending peers, evictions).
- [x] When running devp2p without any capabilities the Hello messages should be ignored.

halfalicious:
Regarding this:

> 
>     * [ ]  Does the bootnode need to open the TCP port?
> 

From what I understand discovery (i.e. no peer connections) only uses UDP - the [Node Discovery v4 page](https://github.com/ethereum/devp2p/blob/master/discv4.md)'s **Wire Protocol** section explicitly mentions that the discovery packets are UDP datagrams:

> Node discovery messages are sent as UDP datagrams. The maximum size of any packet is 1280 bytes.

Consequently I don't think we need to open a TCP port.

halfalicious:
Looked at the network code in Host, we don't need to start the TCP listener (`m_tcp4Acceptor`) since we don't need to support p2p connections at all, we just need to receive discovery packets (which is handled via UDP sockets in `NodeTable`). I'll take a closer look at how to avoid starting the TCP listener.

halfalicious:
> 
> 
> Looked at the network code in Host, we don't need to start the TCP listener (`m_tcp4Acceptor`) since we don't need to support p2p connections at all, we just need to receive discovery packets (which is handled via UDP sockets in `NodeTable`). I'll take a closer look at how to avoid starting the TCP listener (or maybe we can create the `NodeTable `without going through `Host`, though I doubt it since I think the `NodeTable `depends on the host e.g. it uses a reference to `Host`'s boost IO service to listen for discovery packets)

I'm thinking of adding a boolean parameter to `NetworkConfig` named something like `p2p `which aleth-bootnode's `main `would set to `false`. In `Host `we would then skip starting the capability threads and the tcp4 acceptor in `Host::startedWorking`:

https://github.com/ethereum/aleth/blob/ec3886029393fc9927b130ba8f8671480d810391/libp2p/Host.cpp#L774-L787

We could also turn `Host::m_tcp4Acceptor` and `Host::m_ioService` into pointer types and only create them if `NetworkConfig::p2p` is `true`. 

I think we should still restore the data from `aleth-bootnode-network.rlp` since that can contain node table data.

@gumb0,@chfast: Thoughts?

gumb0:
[...]

PR #5402, opened 2018-12-16
[WIP] Only start p2p if capabilities have been registered

### LINKED ISSUE #5388: use-after-free in aleth-bootnode
<pre>=================================================================
<font color="#EF2929"><b>==30369==ERROR: AddressSanitizer: heap-use-after-free on address 0x625000011368 at pc 0x55d76d0762b5 bp 0x7fc80c9886a0 sp 0x7fc80c988690</b></font>
<font color="#729FCF"><b>READ of size 1 at 0x625000011368 thread T1 (p2p)</b></font>
    #0 0x55d76d0762b4 in std::__atomic_base&lt;bool&gt;::load(std::memory_order) const /usr/include/c++/8/bits/atomic_base.h:396
    #1 0x55d76d0762b4 in std::atomic&lt;bool&gt;::operator bool() const /usr/include/c++/8/atomic:86
    #2 0x55d76d0762b4 in operator() /home/chfast/Projects/ethereum/aleth/libp2p/Common.cpp:165
    #3 0x55d76d0762b4 in _M_invoke /usr/include/c++/8/bits/std_function.h:297
    #4 0x55d76d0a19db in boost::asio::detail::wait_handler&lt;std::function&lt;void (boost::system::error_code const&amp;)&gt; &gt;::do_complete(boost::asio::detail::task_io_service*, boost::asio::detail::task_io_service_operation*, boost::system::error_code const&amp;, unsigned long) /usr/include/c++/8/bits/std_function.h:687
    #5 0x55d76d140b11 in boost::asio::detail::task_io_service_operation::complete(boost::asio::detail::task_io_service&amp;, boost::system::error_code const&amp;, unsigned long) /home/chfast/.hunter/_Base/c4cfcc0/af6c4e5/ef3c3eb/Install/include/boost/asio/detail/task_io_service_operation.hpp:38
    #6 0x55d76d140b11 in boost::asio::detail::task_io_service::do_poll_one(boost::asio::detail::scoped_lock&lt;boost::asio::detail::posix_mutex&gt;&amp;, boost::asio::detail::task_io_service_thread_info&amp;, boost::system::error_code const&amp;) /home/chfast/.hunter/_Base/c4cfcc0/af6c4e5/ef3c3eb/Install/include/boost/asio/detail/impl/task_io_service.ipp:436
    #7 0x55d76d140b11 in boost::asio::detail::task_io_service::poll(boost::system::error_code&amp;) /home/chfast/.hunter/_Base/c4cfcc0/af6c4e5/ef3c3eb/Install/include/boost/asio/detail/impl/task_io_service.ipp:198
    #8 0x55d76d0ecb9c in boost::asio::io_service::poll() /home/chfast/.hunter/_Base/c4cfcc0/af6c4e5/ef3c3eb/Install/include/boost/asio/impl/io_service.ipp:85
    #9 0x55d76d0f7670 in dev::p2p::Host::doneWorking() /home/chfast/Projects/ethereum/aleth/libp2p/Host.cpp:178
    #10 0x55d76d301b45 in operator() /home/chfast/Projects/ethereum/aleth/libdevcore/Worker.cpp:63
    #11 0x55d76d301e6c in __invoke_impl&lt;void, dev::Worker::startWorking()::&lt;lambda()&gt; &gt; /usr/include/c++/8/bits/invoke.h:60
    #12 0x55d76d301e6c in __invoke&lt;dev::Worker::startWorking()::&lt;lambda()&gt; &gt; /usr/include/c++/8/bits/invoke.h:95
    #13 0x55d76d301e6c in _M_invoke&lt;0&gt; /usr/include/c++/8/thread:234
    #14 0x55d76d301e6c in operator() /usr/include/c++/8/thread:243
    #15 0x55d76d301e6c in _M_run /usr/include/c++/8/thread:186
    #16 0x7fc810d8757e  (/usr/lib/x86_64-linux-gnu/libstdc++.so.6+0xbd57e)
    #17 0x7fc81105a6da in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x76da)
    #18 0x7fc81044488e in __clone (/lib/x86_64-linux-gnu/libc.so.6+0x12188e)

<font color="#8AE234"><b>0x625000011368 is located 8808 bytes inside of 8816-byte region [0x62500000f100,0x625000011370)</b></font>
<font color="#AD7FA8"><b>freed by thread T1 (p2p) here:</b></font>
    #0 0x7fc811362640 in operator delete(void*) (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xf0640)
    #1 0x55d76d128a49 in std::_Sp_counted_ptr_inplace&lt;dev::p2p::NodeTable, std::allocator&lt;dev::p2p::NodeTable&gt;, (__gnu_cxx::_Lock_policy)2&gt;::_M_destroy() /usr/include/c++/8/ext/new_allocator.h:125
    #2 0x55d76d108382 in std::__shared_count&lt;(__gnu_cxx::_Lock_policy)2&gt;::~__shared_count() /usr/include/c++/8/bits/shared_ptr_base.h:706
    #3 0x55d76d108382 in std::__shared_ptr&lt;dev::p2p::NodeTable, (__gnu_cxx::_Lock_policy)2&gt;::~__shared_ptr() /usr/include/c++/8/bits/shared_ptr_base.h:1145
    #4 0x55d76d108382 in std::__shared_ptr&lt;dev::p2p::NodeTable, (__gnu_cxx::_Lock_policy)2&gt;::reset() /usr/include/c++/8/bits/shared_ptr_base.h:1263
    #5 0x55d76d108382 in dev::p2p::Host::run(boost::system::error_code const&amp;) /home/chfast/Projects/ethereum/aleth/libp2p/Host.cpp:687
    #6 0x55d76d108d34 in do_complete /home/chfast/Projects/ethereum/aleth/libp2p/Host.cpp:755
    #7 0x55d76d13f7ef in boost::asio::detail::task_io_service_operation::complete(boost::asio::detail::task_io_service&amp;, boost::system::error_code const&amp;, unsigned long) /home/chfast/.hunter/_Base/c4cfcc0/af6c4e5/ef3c3eb/Install/include/boost/asio/detail/task_io_service_operation.hpp:38
    #8 0x55d76d13f7ef in boost::asio::detail::task_io_service::do_run_one(boost::asio::detail::scoped_lock&lt;boost::asio::detail::posix_mutex&gt;&amp;, boost::asio::detail::task_io_service_thread_info&amp;, boost::system::error_code const&amp;) /home/chfast/.hunter/_Base/c4cfcc0/af6c4e5/ef3c3eb/Install/include/boost/asio/detail/impl/task_io_service.ipp:372
    #9 0x55d76d13f7ef in boost::asio::detail::task_io_service::run(boost::system::error_code&amp;) /home/chfast/.hunter/_Base/c4cfcc0/af6c4e5/ef3c3eb/Install/include/boost/asio/detail/impl/task_io_service.ipp:149
    #10 0x55d76d0fcf9f in boost::asio::io_service::run() /home/chfast/.hunter/_Base/c4cfcc0/af6c4e5/ef3c3eb/Install/include/boost/asio/impl/io_service.ipp:59
    #11 0x55d76d0fcf9f in dev::p2p::Host::doWork() /home/chfast/Projects/ethereum/aleth/libp2p/Host.cpp:810
    #12 0x55d76d2feaa3 in dev::Worker::workLoop() /home/chfast/Projects/ethereum/aleth/libdevcore/Worker.cpp:140
    #13 0x55d76d301174 in operator() /home/chfast/Projects/ethereum/aleth/libdevcore/Worker.cpp:62
    #14 0x55d76d301e6c in __invoke_impl&lt;void, dev::Worker::startWorking()::&lt;lambda()&gt; &gt; /usr/include/c++/8/bits/invoke.h:60
    #15 0x55d76d301e6c in __invoke&lt;dev::Worker::startWorking()::&lt;lambda()&gt; &gt; /usr/include/c++/8/bits/invoke.h:95
    #16 0x55d76d301e6c in _M_invoke&lt;0&gt; /usr/include/c++/8/thread:234
    #17 0x55d76d301e6c in operator() /usr/include/c++/8/thread:243
    #18 0x55d76d301e6c in _M_run /usr/include/c++/8/thread:186
    #19 0x7fc810d8757e  (/usr/lib/x86_64-linux-gnu/libstdc++.so.6+0xbd57e)

<font color="#AD7FA8"><b>previously allocated by thread T1 (p2p) here:</b></font>
    #0 0x7fc811361790 in operator new(unsigned long) (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xef790)
    #1 0x55d76d1146a2 in std::shared_ptr&lt;dev::p2p::NodeTable&gt; std::make_shared&lt;dev::p2p::NodeTable, boost::asio::io_service&amp;, dev::KeyPair&amp;, dev::p2p::NodeIPEndpoint, bool&amp;&gt;(boost::asio::io_service&amp;, dev::KeyPair&amp;, dev::p2p::NodeIPEndpoint&amp;&amp;, bool&amp;) /usr/include/c++/8/ext/new_allocator.h:111
    #2 0x55d76d1146a2 in dev::p2p::Host::startedWorking() /home/chfast/Projects/ethereum/aleth/libp2p/Host.cpp:794
    #3 0x55d76d301b4c in operator() /home/chfast/Projects/ethereum/aleth/libdevcore/Worker.cpp:61
    #4 0x55d76d301e6c in __invoke_impl&lt;void, dev::Worker::startWorking()::&lt;lambda()&gt; &gt; /usr/include/c++/8/bits/invoke.h:60
    #5 0x55d76d301e6c in __invoke&lt;dev::Worker::startWorking()::&lt;lambda()&gt; &gt; /usr/include/c++/8/bits/invoke.h:95
    #6 0x55d76d301e6c in _M_invoke&lt;0&gt; /usr/include/c++/8/thread:234
    #7 0x55d76d301e6c in operator() /usr/include/c++/8/thread:243
    #8 0x55d76d301e6c in _M_run /usr/include/c++/8/thread:186
    #9 0x7fc810d8757e  (/usr/lib/x86_64-linux-gnu/libstdc++.so.6+0xbd57e)

Thread T1 (p2p) created by T0 here:
    #0 0x7fc8112bca8f in pthread_create (/usr/lib/x86_64-linux-gnu/libasan.so.5+0x4aa8f)
    #1 0x7fc810d87834 in std::thread::_M_start_thread(std::unique_ptr&lt;std::thread::_State, std::default_delete&lt;std::thread::_State&gt; &gt;, void (*)()) (/usr/lib/x86_64-linux-gnu/libstdc++.so.6+0xbd834)
    #2 0x55d76d0fb62c in dev::p2p::Host::start() /home/chfast/Projects/ethereum/aleth/libp2p/Host.cpp:126
    #3 0x55d76cf67095 in main /home/chfast/Projects/ethereum/aleth/aleth-bootnode/main.cpp:131
    #4 0x7fc810344b96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)

SUMMARY: AddressSanitizer: heap-use-after-free /usr/include/c++/8/bits/atomic_base.h:396 in std::__atomic_base&lt;bool&gt;::load(std::memory_order) const
Shadow bytes around the buggy address:
  0x0c4a7fffa210: <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font>
  0x0c4a7fffa220: <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font>
  0x0c4a7fffa230: <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font>
  0x0c4a7fffa240: <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font>
  0x0c4a7fffa250: <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font>
=&gt;0x0c4a7fffa260: <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font> <font color="#AD7FA8"><b>fd</b></font>[<font color="#AD7FA8"><b>fd</b></font>]<font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font>
  0x0c4a7fffa270: <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font>
  0x0c4a7fffa280: <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font>
  0x0c4a7fffa290: <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font>
  0x0c4a7fffa2a0: <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font>
  0x0c4a7fffa2b0: <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font> <font color="#EF2929"><b>fa</b></font>
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       <font color="#EF2929"><b>fa</b></font>
  Freed heap region:       <font color="#AD7FA8"><b>fd</b></font>
  Stack left redzone:      <font color="#EF2929"><b>f1</b></font>
  Stack mid redzone:       <font color="#EF2929"><b>f2</b></font>
  Stack right redzone:     <font color="#EF2929"><b>f3</b></font>
  Stack after return:      <font color="#AD7FA8"><b>f5</b></font>
  Stack use after scope:   <font color="#AD7FA8"><b>f8</b></font>
  Global redzone:          <font color="#EF2929"><b>f9</b></font>
  Global init order:       <font color="#34E2E2"><b>f6</b></font>
  Poisoned by user:        <font color="#729FCF"><b>f7</b></font>
  Container overflow:      <font color="#729FCF"><b>fc</b></font>
  Array cookie:            <font color="#EF2929"><b>ac</b></font>
  Intra object redzone:    <font color="#FCE94F"><b>bb</b></font>
  ASan internal:           <font color="#FCE94F"><b>fe</b></font>
  Left alloca redzone:     <font color="#729FCF"><b>ca</b></font>
  Right alloca redzone:    <font color="#729FCF"><b>cb</b></font>
==30369==ABORTING</pre>

halfalicious:
Address item 1 in #5391 and fix #5388 

Only start p2p functionality if capabilities have been registered for the host. Otherwise, only start discovery. This will prevent a TCP listener from being opened when running `aleth-bootnode` (which only needs to open a UDP listener since it only needs to run discovery)

### DIFF
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -378,54 +378,6 @@ void Host::onNodeTableEvent(NodeID const& _n, NodeTableEventType const& _e)
     }
 }
 
-void Host::determinePublic()
-{
-    // set m_tcpPublic := listenIP (if public) > public > upnp > unspecified address.
-    
-    auto ifAddresses = Network::getInterfaceAddresses();
-    auto laddr = m_netConfig.listenIPAddress.empty() ? bi::address() : bi::address::from_string(m_netConfig.listenIPAddress);
-    auto lset = !laddr.is_unspecified();
-    auto paddr = m_netConfig.publicIPAddress.empty() ? bi::address() : bi::address::from_string(m_netConfig.publicIPAddress);
-    auto pset = !paddr.is_unspecified();
-    
-    bool listenIsPublic = lset && isPublicAddress(laddr);
-    bool publicIsHost = !lset && pset && ifAddresses.count(paddr);
-    
-    bi::tcp::endpoint ep(bi::address(), m_listenPort);
-    if (m_netConfig.traverseNAT && listenIsPublic)
-    {
-        cnetnote << "Listen address set to Public address: " << laddr << ". UPnP disabled.";
-        ep.address(laddr);
-    }
-    else if (m_netConfig.traverseNAT && publicIsHost)
-    {
-        cnetnote << "Public address set to Host configured address: " << paddr << ". UPnP disabled.";
-        ep.address(paddr);
-    }
-    else if (m_netConfig.traverseNAT)
-    {
-        bi::address natIFAddr;
-        ep = Network::traverseNAT(lset && ifAddresses.count(laddr) ? std::set<bi::address>({laddr}) : ifAddresses, m_listenPort, natIFAddr);
-        
-        if (lset && natIFAddr != laddr)
-            // if listen address is set, Host will use it, even if upnp returns different
-            cwarn << "Listen address " << laddr << " differs from local address " << natIFAddr
-                  << " returned by UPnP!";
-
-        if (pset && ep.address() != paddr)
-        {
-            // if public address is set, Host will advertise it, even if upnp returns different
-            cwarn << "Specified public address " << paddr << " differs from external address "
-                  << ep.address() << " returned by UPnP!";
-            ep.address(paddr);
-        }
-    }
-    else if (pset)
-        ep.address(paddr);
-
-    m_tcpPublic = ep;
-}
-
 void Host::runAcceptor()
 {
     assert(m_listenPort > 0);
@@ -741,6 +693,7 @@ void Host::startedWorking()
     m_timer.reset(new io::deadline_timer(m_ioService));
     m_run = true;
 
+    NodeIPEndpoint publicEndpoint;
     if (!m_capabilities.empty())
     {
         // start capability threads (ready for incoming connections)
@@ -752,8 +705,9 @@ void Host::startedWorking()
         if (port > 0)
         {
             // Port may have been changed if the specified port was already in use
-            m_netConfig.listenPort = m_listenPort = port;
-            determinePublic();
+            m_listenPort = port;
+            publicEndpoint = Network::determinePublic(m_netConfig, m_listenPort);
+            m_tcpPublic = (bi::tcp::endpoint)publicEndpoint;
             runAcceptor();
         }
         else
@@ -762,16 +716,16 @@ void Host::startedWorking()
                 << " TCP Listen port is invalid or unavailable.";
         }
     }
+    else
+        publicEndpoint = Network::determinePublic(m_netConfig, m_netConfig.listenPort);
 
-    auto nodeTable = make_shared<NodeTable>(m_ioService, m_alias,
-        // Use data from network configuration rather than Host's settings because p2p might not
-        // have been started
-        NodeIPEndpoint(!m_netConfig.listenIPAddress.empty() ?
-                           bi::address::from_string(m_netConfig.listenIPAddress) :
-                           bi::address_v4(),
-            m_netConfig.listenPort /* UDP */, m_netConfig.listenPort /* TCP */),
-        m_netConfig.discovery, m_netConfig.allowLocalDiscovery);
-
+    auto nodeTable = make_shared<NodeTable>(
+        m_ioService,
+        m_alias,
+        publicEndpoint,
+        m_netConfig.discovery,
+        m_netConfig.allowLocalDiscovery
+        );
     if (!m_capabilities.empty())
         // Only set the event handler if there are capabilities since otherwise p2p
         // isn't running
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -621,9 +621,9 @@ size_t Host::peerCount() const
     return retCount;
 }
 
-void Host::run(boost::system::error_code const&)
+void Host::run(boost::system::error_code const& _ec)
 {
-    if (!m_run)
+    if (!m_run || _ec)
         return;
 
     if (auto nodeTable = this->nodeTable()) // This again requires x_nodeTable, which is why an additional variable nodeTable is used.
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -780,7 +780,8 @@ void Host::startedWorking()
         int port = Network::tcp4Listen(m_tcp4Acceptor, m_netConfig);
         if (port > 0)
         {
-            m_listenPort = port;
+            // Port may have been changed if the specified port was already in use
+            m_netConfig.listenPort = m_listenPort = port;
             determinePublic();
             runAcceptor();
         }
@@ -799,7 +800,7 @@ void Host::startedWorking()
         m_netConfig.discovery, m_netConfig.allowLocalDiscovery);
 
     if (m_capabilities.size())
-        // Don't need to set event handler since there are no peers for Host to update when p2p
+        // Only set the event handler if there are capabilities since otherwise p2p
         // isn't running
         nodeTable->setEventHandler(new HostNodeTableHandler(*this));
     DEV_GUARDED(x_nodeTable)
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -759,20 +759,24 @@ void Host::startedWorking()
         m_run = true;
     }
 
-    // start capability threads (ready for incoming connections)
-    for (auto const& h: m_capabilities)
-        h.second->onStarting();
-    
-    // try to open acceptor (todo: ipv6)
-    int port = Network::tcp4Listen(m_tcp4Acceptor, m_netConfig);
-    if (port > 0)
+    if (m_capabilities.size())
     {
-        m_listenPort = port;
-        determinePublic();
-        runAcceptor();
+        // start capability threads (ready for incoming connections)
+        for (auto const& h : m_capabilities)
+            h.second->onStarting();
+
+        // try to open acceptor (todo: ipv6)
+        int port = Network::tcp4Listen(m_tcp4Acceptor, m_netConfig);
+        if (port > 0)
+        {
+            m_listenPort = port;
+            determinePublic();
+            runAcceptor();
+        }
+        else
+            LOG(m_logger) << "p2p.start.notice id: " << id()
+                          << " TCP Listen port is invalid or unavailable.";
     }
-    else
-        LOG(m_logger) << "p2p.start.notice id: " << id() << " TCP Listen port is invalid or unavailable.";
 
     auto nodeTable = make_shared<NodeTable>(
         m_ioService,
@@ -786,7 +790,10 @@ void Host::startedWorking()
         m_nodeTable = nodeTable;
     restoreNetwork(&m_restoreNetwork);
 
-    LOG(m_logger) << "p2p.started id: " << id();
+    if (m_capabilities.size())
+        LOG(m_logger) << "p2p.started id: " << id();
+    else
+        LOG(m_logger) << "no capabilities detected, p2p not started";
 
     run(boost::system::error_code());
 }
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -139,11 +139,11 @@ void Host::stop()
 
 void Host::doneWorking()
 {
+    // reset ioservice (cancels all timers and allows manually polling network if there are
+    // capabilities. Also enables reuse of the ioservice)
+    m_ioService.reset();
     if (m_capabilities.size())
     {
-        // reset ioservice (cancels all timers and allows manually polling network, below)
-        m_ioService.reset();
-
         DEV_GUARDED(x_timers)
             m_timers.clear();
 
@@ -206,9 +206,6 @@ void Host::doneWorking()
         RecursiveGuard l(x_sessions);
         m_sessions.clear();
     }
-    else
-        // allows reusing ioservice in the future
-        m_ioService.reset();
 }
 
 bool Host::isRequiredPeer(NodeID const& _id) const
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -733,12 +733,10 @@ void Host::run(boost::system::error_code const&)
     m_timer->async_wait(runcb);
 }
 
+// Called after thread has been started to perform additional class-specific state
+// initialization (e.g. start capability threads, start TCP listener, and kick off timers) 
 void Host::startedWorking()
 {
-    // Called after thread has been started to perform additional class-specific state
-    // initialization (e.g. start capability threads, start TCP listener, and kick off timers)
-    asserts(!m_timer);
-
     // create deadline timer
     m_timer.reset(new io::deadline_timer(m_ioService));
     m_run = true;
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -90,9 +90,7 @@ Host::Host(string const& _clientVersion, KeyPair const& _alias, NetworkConfig co
     m_alias(_alias),
     m_lastPing(chrono::steady_clock::time_point::min()),
     m_capabilityHost(createCapabilityHost(*this))
-{
-    cnetnote << "Id: " << id();
-}
+{}
 
 Host::Host(string const& _clientVersion, NetworkConfig const& _n, bytesConstRef _restoreNetwork):
     Host(_clientVersion, networkAlias(_restoreNetwork), _n)
@@ -148,7 +146,7 @@ void Host::doneWorking()
 
         DEV_GUARDED(x_timers)
             m_timers.clear();
-        
+
         // shutdown acceptor
         m_tcp4Acceptor.cancel();
         if (m_tcp4Acceptor.is_open())
@@ -162,14 +160,14 @@ void Host::doneWorking()
             m_ioService.poll();
 
         // stop capabilities (eth: stops syncing or block/tx broadcast)
-        for (auto const& h: m_capabilities)
+        for (auto const& h : m_capabilities)
             h.second->onStopping();
 
         // disconnect pending handshake, before peers, as a handshake may create a peer
         for (unsigned n = 0;; n = 0)
         {
             DEV_GUARDED(x_connecting)
-                for (auto const& i: m_connecting)
+                for (auto const& i : m_connecting)
                     if (auto h = i.lock())
                     {
                         h->cancel();
@@ -179,12 +177,12 @@ void Host::doneWorking()
                 break;
             m_ioService.poll();
         }
-        
+
         // disconnect peers
         for (unsigned n = 0;; n = 0)
         {
             DEV_RECURSIVE_GUARDED(x_sessions)
-                for (auto i: m_sessions)
+                for (auto i : m_sessions)
                     if (auto p = i.second.lock())
                         if (p->isConnected())
                         {
@@ -785,7 +783,7 @@ void Host::startedWorking()
 
     if (m_capabilities.size())
     {
-        LOG(m_logger) << "p2p.started id: " << id();
+        LOG(m_logger) << "Capabilities detected, p2p started";
         run(boost::system::error_code());
     }
     else
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -132,11 +132,7 @@ void Host::stop()
     if (!m_run.exchange(false))
         return;
 
-    {
-        unique_lock<mutex> l(x_runTimer);
-        while (m_timer)
-            m_timerReset.wait(l);
-    }
+    m_ioService.stop();
 
     // stop worker thread
     if (isWorking())
@@ -519,7 +515,7 @@ void Host::addPeer(NodeSpec const& _s, PeerType _t)
 
 void Host::addNode(NodeID const& _node, NodeIPEndpoint const& _endpoint)
 {
-    // return if network is stopped while waiting on Host::run() or nodeTable to start
+    // return if network is stopped while waiting on Host::run or nodeTable to start
     while (!haveNetwork())
         if (isWorking())
             this_thread::sleep_for(chrono::milliseconds(50));
@@ -673,82 +669,62 @@ size_t Host::peerCount() const
 void Host::run(boost::system::error_code const&)
 {
     if (!m_run)
-    {
-        // reset NodeTable
-        DEV_GUARDED(x_nodeTable)
-            m_nodeTable.reset();
-
-        // stopping io service allows running manual network operations for shutdown
-        // and also stops blocking worker thread, allowing worker thread to exit
-        m_ioService.stop();
-
-        // resetting timer signals network that nothing else can be scheduled to run
-        DEV_GUARDED(x_runTimer)
-            m_timer.reset();
-
-        m_timerReset.notify_all();
         return;
-    }
 
-    // Don't need to process node table events (which results in peers being added/dropped) or
-    // perform any p2p functionality if we aren't running p2p
-    if (m_capabilities.size())
+    if (auto nodeTable = this->nodeTable()) // This again requires x_nodeTable, which is why an additional variable nodeTable is used.
+        nodeTable->processEvents();
+
+    // cleanup zombies
+    DEV_GUARDED(x_connecting)
+        m_connecting.remove_if([](std::weak_ptr<RLPXHandshake> h){ return h.expired(); });
+    DEV_GUARDED(x_timers)
+    m_timers.remove_if([](std::unique_ptr<boost::asio::deadline_timer> const& t) {
+        return t->expires_from_now().total_milliseconds() < 0;
+    });
+
+    keepAlivePeers();
+        
+    // At this time peers will be disconnected based on natural TCP timeout.
+    // disconnectLatePeers needs to be updated for the assumption that Session
+    // is always live and to ensure reputation and fallback timers are properly
+    // updated. // disconnectLatePeers();
+
+    // todo: update peerSlotsAvailable()
+        
+    list<shared_ptr<Peer>> toConnect;
+    unsigned reqConn = 0;
     {
-        if (auto nodeTable = this->nodeTable()) // This again requires x_nodeTable, which is why an additional variable nodeTable is used.
-            nodeTable->processEvents();
-
-        // cleanup zombies
-        DEV_GUARDED(x_connecting)
-            m_connecting.remove_if([](std::weak_ptr<RLPXHandshake> h){ return h.expired(); });
-        DEV_GUARDED(x_timers)
-        m_timers.remove_if([](std::unique_ptr<boost::asio::deadline_timer> const& t) {
-            return t->expires_from_now().total_milliseconds() < 0;
-        });
-
-        keepAlivePeers();
-        
-        // At this time peers will be disconnected based on natural TCP timeout.
-        // disconnectLatePeers needs to be updated for the assumption that Session
-        // is always live and to ensure reputation and fallback timers are properly
-        // updated. // disconnectLatePeers();
-
-        // todo: update peerSlotsAvailable()
-        
-        list<shared_ptr<Peer>> toConnect;
-        unsigned reqConn = 0;
+        RecursiveGuard l(x_sessions);
+        for (auto const& p: m_peers)
         {
-            RecursiveGuard l(x_sessions);
-            for (auto const& p: m_peers)
-            {
-                bool haveSession = havePeerSession(p.second->id);
-                bool required = p.second->peerType == PeerType::Required;
-                if (haveSession && required)
-                    reqConn++;
-                else if (!haveSession && p.second->shouldReconnect() && (!m_netConfig.pin || required))
-                    toConnect.push_back(p.second);
-            }
-        }
-
-        for (auto p: toConnect)
-            if (p->peerType == PeerType::Required && reqConn++ < m_idealPeerCount)
-                connect(p);
-        
-        if (!m_netConfig.pin)
-        {
-            unsigned const maxSlots = m_idealPeerCount + reqConn;
-            unsigned occupiedSlots = peerCount() + m_pendingPeerConns.size();
-            for (auto peerToConnect = toConnect.cbegin();
-                occupiedSlots <= maxSlots && peerToConnect != toConnect.cend(); ++peerToConnect)
-            {
-                if ((*peerToConnect)->peerType == PeerType::Optional)
-                {
-                    connect(*peerToConnect);
-                    ++occupiedSlots;
-                }
-            }
+            bool haveSession = havePeerSession(p.second->id);
+            bool required = p.second->peerType == PeerType::Required;
+            if (haveSession && required)
+                reqConn++;
+            else if (!haveSession && p.second->shouldReconnect() && (!m_netConfig.pin || required))
+                toConnect.push_back(p.second);
         }
     }
 
+    for (auto p: toConnect)
+        if (p->peerType == PeerType::Required && reqConn++ < m_idealPeerCount)
+            connect(p);
+        
+    if (!m_netConfig.pin)
+    {
+        unsigned const maxSlots = m_idealPeerCount + reqConn;
+        unsigned occupiedSlots = peerCount() + m_pendingPeerConns.size();
+        for (auto peerToConnect = toConnect.cbegin();
+            occupiedSlots <= maxSlots && peerToConnect != toConnect.cend(); ++peerToConnect)
+        {
+            if ((*peerToConnect)->peerType == PeerType::Optional)
+            {
+                connect(*peerToConnect);
+                ++occupiedSlots;
+            }
+        }
+    }
+    
     auto runcb = [this](boost::system::error_code const& error) { run(error); };
     m_timer->expires_from_now(boost::posix_time::milliseconds(c_timerInterval));
     m_timer->async_wait(runcb);
@@ -808,11 +784,12 @@ void Host::startedWorking()
     restoreNetwork(&m_restoreNetwork);
 
     if (m_capabilities.size())
+    {
         LOG(m_logger) << "p2p.started id: " << id();
+        run(boost::system::error_code());
+    }
     else
         LOG(m_logger) << "No capabilities detected, p2p not started";
-
-    run(boost::system::error_code());
 }
 
 void Host::doWork()
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -144,70 +144,76 @@ void Host::doneWorking()
     // reset ioservice (cancels all timers and allows manually polling network if there are
     // capabilities. Also enables reuse of the ioservice)
     m_ioService.reset();
-    if (m_capabilities.size())
+    if (m_capabilities.empty())
+        return;
+
+    DEV_GUARDED(x_timers) { m_timers.clear(); }
+
+    // shutdown acceptor
+    m_tcp4Acceptor.cancel();
+    if (m_tcp4Acceptor.is_open())
+        m_tcp4Acceptor.close();
+
+    // There maybe an incoming connection which started but hasn't finished.
+    // Wait for acceptor to end itself instead of assuming it's complete.
+    // This helps ensure a peer isn't stopped at the same time it's starting
+    // and that socket for pending connection is closed.
+    while (m_accepting)
+        m_ioService.poll();
+
+    // stop capabilities (eth: stops syncing or block/tx broadcast)
+    for (auto const& h : m_capabilities)
+        h.second->onStopping();
+
+    // disconnect pending handshake, before peers, as a handshake may create a peer
+    bool handshakeDisconnected;
+    while (true)
     {
-        DEV_GUARDED(x_timers)
-            m_timers.clear();
-
-        // shutdown acceptor
-        m_tcp4Acceptor.cancel();
-        if (m_tcp4Acceptor.is_open())
-            m_tcp4Acceptor.close();
-
-        // There maybe an incoming connection which started but hasn't finished.
-        // Wait for acceptor to end itself instead of assuming it's complete.
-        // This helps ensure a peer isn't stopped at the same time it's starting
-        // and that socket for pending connection is closed.
-        while (m_accepting)
-            m_ioService.poll();
-
-        // stop capabilities (eth: stops syncing or block/tx broadcast)
-        for (auto const& h : m_capabilities)
-            h.second->onStopping();
-
-        // disconnect pending handshake, before peers, as a handshake may create a peer
-        for (unsigned n = 0;; n = 0)
+        handshakeDisconnected = false;
+        DEV_GUARDED(x_connecting)
         {
-            DEV_GUARDED(x_connecting)
-                for (auto const& i : m_connecting)
-                    if (auto h = i.lock())
-                    {
-                        h->cancel();
-                        n++;
-                    }
-            if (!n)
-                break;
-            m_ioService.poll();
+            for (auto const& i : m_connecting)
+                if (auto h = i.lock())
+                {
+                    h->cancel();
+                    handshakeDisconnected = true;
+                }
         }
-
-        // disconnect peers
-        for (unsigned n = 0;; n = 0)
-        {
-            DEV_RECURSIVE_GUARDED(x_sessions)
-                for (auto i : m_sessions)
-                    if (auto p = i.second.lock())
-                        if (p->isConnected())
-                        {
-                            p->disconnect(ClientQuit);
-                            n++;
-                        }
-            if (!n)
-                break;
-
-            // poll so that peers send out disconnect packets
-            m_ioService.poll();
-        }
-
-        // stop network (again; helpful to call before subsequent reset())
-        m_ioService.stop();
-
-        // reset network (allows reusing ioservice in future)
-        m_ioService.reset();
-
-        // finally, clear out peers (in case they're lingering)
-        RecursiveGuard l(x_sessions);
-        m_sessions.clear();
+        if (!handshakeDisconnected)
+            break;
+        m_ioService.poll();
     }
+
+    // disconnect peers
+    bool peerDisconnected;
+    while (true)
+    {
+        peerDisconnected = false;
+        DEV_RECURSIVE_GUARDED(x_sessions)
+        {
+            for (auto i : m_sessions)
+                if (auto p = i.second.lock())
+                    if (p->isConnected())
+                    {
+                        p->disconnect(ClientQuit);
+                        peerDisconnected = true;
+                    }
+        }
+        if (!peerDisconnected)
+            break;
+        // poll so that peers send out disconnect packets
+        m_ioService.poll();
+    }
+
+    // stop network (again; helpful to call before subsequent reset())
+    m_ioService.stop();
+
+    // reset network (allows reusing ioservice in future)
+    m_ioService.reset();
+
+    // finally, clear out peers (in case they're lingering)
+    RecursiveGuard l(x_sessions);
+    m_sessions.clear();
 }
 
 bool Host::isRequiredPeer(NodeID const& _id) const
@@ -680,7 +686,7 @@ void Host::run(boost::system::error_code const&)
     });
 
     keepAlivePeers();
-        
+
     // At this time peers will be disconnected based on natural TCP timeout.
     // disconnectLatePeers needs to be updated for the assumption that Session
     // is always live and to ensure reputation and fallback timers are properly
@@ -737,7 +743,7 @@ void Host::startedWorking()
     m_timer.reset(new io::deadline_timer(m_ioService));
     m_run = true;
 
-    if (m_capabilities.size())
+    if (!m_capabilities.empty())
     {
         // start capability threads (ready for incoming connections)
         for (auto const& h : m_capabilities)
@@ -753,8 +759,10 @@ void Host::startedWorking()
             runAcceptor();
         }
         else
+        {
             LOG(m_logger) << "p2p.start.notice id: " << id()
-                          << " TCP Listen port is invalid or unavailable.";
+                << " TCP Listen port is invalid or unavailable.";
+        }
     }
 
     auto nodeTable = make_shared<NodeTable>(m_ioService, m_alias,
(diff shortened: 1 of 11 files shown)


halfalicious:
Need to prevent `aleth-bootnode` from attempting connections to nodes in the NodeTable:

```
C:\Users\nilse\Documents\Code\aleth\build\aleth-bootnode\Debug>aleth-bootnode -v4
aleth-bootnode, a C++ Ethereum bootnode implementation
INFO  12-16 14:02:36 main net    Id: ##43cdce4f…
DEBUG 12-16 14:02:36 p2p  discov Active node ##c2db46eb…@127.0.0.1:30305
DEBUG 12-16 14:02:36 p2p  net    no capabilities detected, p2p not started
DEBUG 12-16 14:02:36 p2p  net    p2p.host.nodeTable.events.nodeEntryAdded ##c2db46eb…
TRACE 12-16 14:02:36 p2p  net    Attempting connection to node ##c2db46eb…@127.0.0.1:30305 from ##43cdce4f…
Node ID: enode://43cdce4ffde8fc47dbdf6cdbac45c3cba9bddb615bd02239dc35c3799fdfa5fd7454983c1e55a3456c62e0aad8a8ee35e0cc4dda0c89d1ddacdb5767d8127fd0@0.0.0.0:0
TRACE 12-16 14:02:37 p2p  net    Connection refused to node ##c2db46eb…@127.0.0.1:30305 (No connection could be made because the target machine actively refused it)
DEBUG 12-16 14:02:43 p2p  discov performing random discovery
DEBUG 12-16 14:02:45 p2p  discov FindNode to ##2cb41268…@127.0.0.1 UDP 30305 TCP 30305
DEBUG 12-16 14:02:46 p2p  discov Terminating discover after 1 rounds.
TRACE 12-16 14:02:52 p2p  net    Attempting connection to node ##c2db46eb…@127.0.0.1:30305 from ##43cdce4f…
TRACE 12-16 14:02:53 p2p  net    Connection refused to node ##c2db46eb…@127.0.0.1:30305 (No connection could be made because the target machine actively refused it)
DEBUG 12-16 14:02:53 p2p  discov performing random discovery
DEBUG 12-16 14:02:53 p2p  discov FindNode to ##f961c8f3…@127.0.0.1 UDP 30305 TCP 30305
DEBUG 12-16 14:02:54 p2p  discov Terminating discover after 1 rounds.
```

halfalicious:
> 
> 
> Need to prevent `aleth-bootnode` from attempting connections to nodes in the NodeTable:
> 
> ```
> C:\Users\nilse\Documents\Code\aleth\build\aleth-bootnode\Debug>aleth-bootnode -v4
> aleth-bootnode, a C++ Ethereum bootnode implementation
> INFO  12-16 14:02:36 main net    Id: ##43cdce4f…
> DEBUG 12-16 14:02:36 p2p  discov Active node ##c2db46eb…@127.0.0.1:30305
> DEBUG 12-16 14:02:36 p2p  net    no capabilities detected, p2p not started
> DEBUG 12-16 14:02:36 p2p  net    p2p.host.nodeTable.events.nodeEntryAdded ##c2db46eb…
> TRACE 12-16 14:02:36 p2p  net    Attempting connection to node ##c2db46eb…@127.0.0.1:30305 from ##43cdce4f…
> Node ID: enode://43cdce4ffde8fc47dbdf6cdbac45c3cba9bddb615bd02239dc35c3799fdfa5fd7454983c1e55a3456c62e0aad8a8ee35e0cc4dda0c89d1ddacdb5767d8127fd0@0.0.0.0:0
> TRACE 12-16 14:02:37 p2p  net    Connection refused to node ##c2db46eb…@127.0.0.1:30305 (No connection could be made because the target machine actively refused it)
> DEBUG 12-16 14:02:43 p2p  discov performing random discovery
> DEBUG 12-16 14:02:45 p2p  discov FindNode to ##2cb41268…@127.0.0.1 UDP 30305 TCP 30305
> DEBUG 12-16 14:02:46 p2p  discov Terminating discover after 1 rounds.
> TRACE 12-16 14:02:52 p2p  net    Attempting connection to node ##c2db46eb…@127.0.0.1:30305 from ##43cdce4f…
> TRACE 12-16 14:02:53 p2p  net    Connection refused to node ##c2db46eb…@127.0.0.1:30305 (No connection could be made because the target machine actively refused it)
> DEBUG 12-16 14:02:53 p2p  discov performing random discovery
> DEBUG 12-16 14:02:53 p2p  discov FindNode to ##f961c8f3…@127.0.0.1 UDP 30305 TCP 30305
> DEBUG 12-16 14:02:54 p2p  discov Terminating discover after 1 rounds.
> ```

Need to selectively disable registering a NodeTable event handler and also make changes in Host::run...perhaps need to further decouple Host from NodeTable

halfalicious:
`aleth-bootnode` now just starts a UDP listener and runs discovery and I've verified that it can interact with an `aleth `node. I now need to fix the following failing tests:

```
The following tests FAILED:
	484 - libp2p/p2p/host (Failed)
	486 - libp2p/p2p/saveNodes (Failed)
	487 - libp2p/p2pPeer/requirePeer (Failed)
```

libp2p/Host.cpp
```diff
@@ -698,56 +698,61 @@ void Host::run(boost::system::error_code const&)
         return;
     }
 
-    if (auto nodeTable = this->nodeTable()) // This again requires x_nodeTable, which is why an additional variable nodeTable is used.
-        nodeTable->processEvents();
+    // Don't need to process node table events (which results in peers being added/dropped) or
+    // perform any p2p functionality if we aren't running p2p
+    if (m_capabilities.size())
```
gumb0: Hmm if everything in this function is not needed without capabilities, probably we should just not schedule it for run, that is skip this https://github.com/ethereum/aleth/blob/b92056768376d42b809e2a80afd3178b2dead29f/libp2p/Host.cpp#L802

halfalicious: Well the function deletes the node table, stops the io service, and stops the timer (which is used to schedule future calls to run()): https://github.com/ethereum/aleth/blob/b92056768376d42b809e2a80afd3178b2dead29f/libp2p/Host.cpp#L683-L699

I don't know enough about how boost's io service works and the interaction between the host and the node table to know if it's safe to skip this code or if it can be moved into another function. I'll do some more research.

halfalicious:
Rebased

libp2p/Host.cpp
```diff
@@ -759,34 +770,46 @@ void Host::startedWorking()
...
-    nodeTable->setEventHandler(new HostNodeTableHandler(*this));
+
+    auto nodeTable = make_shared<NodeTable>(m_ioService, m_alias,
+        // Use data from network configuration rather than Host's settings because p2p might not
+        // have been started
+        NodeIPEndpoint(!m_netConfig.listenIPAddress.empty() ?
+                           bi::address::from_string(m_netConfig.listenIPAddress) :
+                           bi::address_v4(),
+            m_netConfig.listenPort /* UDP */, m_netConfig.listenPort /* TCP */),
+        m_netConfig.discovery);
```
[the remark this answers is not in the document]
halfalicious: Ah that's strange, I'll fix that. Thanks @gumb0 !

libp2p/Host.cpp
gumb0: Some more thoughts about this

- I would try to move all this under `if (!m_run)` out of this function, perhaps to `Host::stop()` (the only place that sets `m_run` to `false`)
- `m_ioService.stop()` may be safely called from another thread, I found discussion about this here https://stackoverflow.com/a/4810965/1636308
- `m_nodeTable` should be deleted after `m_ioService.stop()` call anyway to fix https://github.com/ethereum/aleth/issues/5388. Or perhaps it should be not deleted at all when network is stopped (after `io_service` is stopped nothing will be called in `NodeTable` anyway) and we should just leave its destruction to `Host` destructor
- I don't see any good reason to destroy `m_timer` when network is stopped. Maybe these were some hacks to handle the edge case when `Host::stop()` is being called while its starting is still in progress: this comment points to it https://github.com/ethereum/aleth/blob/60a1075455feaf497cc8313b829f851fd3692574/libp2p/Host.cpp#L753-L755
but it seems weird to use `m_timer` for this

halfalicious:
> 
> 
> 2 tests are still failing:
> 
> ```
> The following tests FAILED:
> 	488 - libp2p/p2p/host (Failed)
> 	490 - libp2p/p2p/saveNodes (Failed)
> ```
> 
> Ran them locally and it looks like the hosts aren't peering with each other, I need to investigate and figure out why.

Looking at `libp2p/p2p/host`, there's an issue with the second node not being added to the first node's node table - the first node is pinging the second node but the second node isn't receiving the packet (`NodeTable::onPacketReceived` isn't being invoked). Will dig deeper.

gumb0:
I saw something similar when I created two `NodeTable`s in this test https://github.com/ethereum/aleth/blob/b545f53882e670eb0fa9ceed494ce05ff0364cb6/test/unittests/libp2p/net.cpp#L786-L816

They tried to open UDP socket on the same port, then the second one didn't fail due to this code (which I don't fully understand yet) 
https://github.com/ethereum/aleth/blob/60a1075455feaf497cc8313b829f851fd3692574/libp2p/UDP.h#L165-L172
but then the second one didn't receive the packets from the first one.
(I overcame this by ensuring that `NodeTable`s are created on different ports)

In `host` test I think it's supposed to get random listen port when the default one is occupied, and use it for `NodeTable`, too. So you could check whether `NodeTable`s use different ports here

halfalicious:
> ndom listen port when the default one is occupied, and use it for `NodeTable`, too. So you could check whether `NodeTa`

Thanks for the tip! They're using the same port (port 0), the reason being that my changes pass an endpoint created from the network config to the NodeTable ctor rather than the host's endpoint since if p2p isn't started the host's endpoint won't have been initialized. I think that the correct way to fix this is to update the NetworkConfig class to set its listen port to the default port (30303) if 0 is supplied, and to also check if the listen port being set is already in use by TCP or UDP and if so, to choose a random port that isn't in use.

halfalicious:
> nd one didn't receive the packets from the first one.
> (I overcame this by ensuring th

I think you need to set the `ba::socket_base::reuse_address(false)) ` option before binding in order for the binding to fail.

libp2p/Host.h
```diff
@@ -312,7 +312,6 @@ class Host: public Worker
     std::unique_ptr<io::deadline_timer> m_timer;					///< Timer which, when network is running, calls run() every c_timerInterval ms.
     mutable std::mutex x_runTimer;	///< Start/stop mutex.
```
gumb0: `x_runTimer` doesn't look to be useful anymore

libp2p/Host.cpp
```diff
@@ -145,70 +141,76 @@ void Host::stop()
...
-    // finally, clear out peers (in case they're lingering)
-    RecursiveGuard l(x_sessions);
-    m_sessions.clear();
+        // finally, clear out peers (in case they're lingering)
+        RecursiveGuard l(x_sessions);
+        m_sessions.clear();
+    }
+    else
+        // allows reusing ioservice in the future
+        m_ioService.reset();
```
gumb0: This can be simplified by moving it out of `if`, as it's present in both branches

libp2p/Host.cpp
```diff
@@ -145,70 +141,76 @@ void Host::stop()
...
-                    n++;
-                }
-        if (!n)
-            break;
-        m_ioService.poll();
-    }
-    
-    // disconnect peers
-    for (unsigned n = 0;; n = 0)
+    if (m_capabilities.size())
```
gumb0: Is this really needed? It seems we can leave `doneWorking` as it is, and not having capabilities / sessions / acceptor it won't do anything incorrect.

libp2p/Host.cpp
```diff
@@ -759,36 +746,50 @@ void Host::startedWorking()
...
-        m_nodeTable = nodeTable;
+    m_nodeTable = nodeTable;
     restoreNetwork(&m_restoreNetwork);
 
-    LOG(m_logger) << "p2p.started id: " << id();
-
-    run(boost::system::error_code());
+    if (m_capabilities.size())
+    {
+        LOG(m_logger) << "p2p.started id: " << id();
```
gumb0: This displays our NodeID, so might be useful for the bootnode, too. On the other hand, there's another log message somewhere also printing it at program start, so maybe one of them should be removed.

libp2p/Host.cpp
halfalicious: Good question - I'm not sure what the impact of some of these actions are when performed on objects that haven't been fully initialized (e.g. calling m_tcp4Acceptor.cancel() without first opening the acceptor). I'll take a closer look.

libp2p/Host.cpp
halfalicious: Correct - the host's enode info is already logged in both aleth's and aleth-bootnode's main. I think we should still log a status message regarding the started/not started p2p status of aleth/aleth-bootnode but I don't think we need to log the ID along with the status. I'll remove the id() call.

halfalicious: The ID is also logged in the Host ctor: https://github.com/ethereum/aleth/blob/91787f239566501a2b23f17946618ddf6a3cee15/libp2p/Host.cpp#L93-L95

I'll remove it from there as well.

libp2p/Host.cpp
halfalicious: I ran a quick test and there doesn't appear to be any functional impact but I do get a warning message now about there not being a valid file handle (or something to that affect). As such, I think it's best to leave the code as is.

libp2p/Host.cpp
halfalicious: I've actually ended up moving the enode URL logging code from the mains to right after network start in Host::start - this consolidates things slightly.

halfalicious:
Rebased

halfalicious:
This PR contains a lot of small changes, I'll squash most of them before merging

libp2p/Host.cpp
```diff
@@ -145,70 +141,73 @@ void Host::stop()
...
+            if (!n)
+                break;
+            m_ioService.poll();
+        }
 
-        // poll so that peers send out disconnect packets
-        m_ioService.poll();
-    }
+        // disconnect peers
+        for (unsigned n = 0;; n = 0)
```
chfast: WTF? Use `int` at least. `unsigned` is bad for a loop counter.

libp2p/Host.cpp
```diff
@@ -691,19 +675,19 @@ void Host::run(boost::system::error_code const&)
...
         m_connecting.remove_if([](std::weak_ptr<RLPXHandshake> h){ return h.expired(); });
     DEV_GUARDED(x_timers)
-    m_timers.remove_if([](std::unique_ptr<io::deadline_timer> const& t) {
+    m_timers.remove_if([](std::unique_ptr<boost::asio::deadline_timer> const& t) {
         return t->expires_from_now().total_milliseconds() < 0;
     });
 
     keepAlivePeers();
-    
+
```
chfast: Some added whitespace.

libp2p/Host.cpp
```diff
@@ -749,46 +733,54 @@ void Host::startedWorking()
...
+    m_run = true;
 
-    // start capability threads (ready for incoming connections)
-    for (auto const& h: m_capabilities)
-        h.second->onStarting();
-    
-    // try to open acceptor (todo: ipv6)
-    int port = Network::tcp4Listen(m_tcp4Acceptor, m_netConfig);
-    if (port > 0)
+    if (m_capabilities.size())
```
```suggestion
    if (!m_capabilities.empty())
```

libp2p/Host.cpp
```diff
@@ -749,46 +733,54 @@ void Host::startedWorking()
...
+        int port = Network::tcp4Listen(m_tcp4Acceptor, m_netConfig);
+        if (port > 0)
+        {
+            // Port may have been changed if the specified port was already in use
+            m_netConfig.listenPort = m_listenPort = port;
+            determinePublic();
+            runAcceptor();
+        }
+        else
+            LOG(m_logger) << "p2p.start.notice id: " << id()
```
chfast: I propose enclose with `{}` when the statement occupies more than single line.
We can have clang-tidy check for this.

libp2p/Host.cpp
```diff
@@ -749,46 +733,54 @@ void Host::startedWorking()
...
+    auto nodeTable = make_shared<NodeTable>(m_ioService, m_alias,
+        // Use data from network configuration rather than Host's settings because p2p might not
+        // have been started
+        NodeIPEndpoint(!m_netConfig.listenIPAddress.empty() ?
+                           bi::address::from_string(m_netConfig.listenIPAddress) :
+                           bi::address_v4(),
+            m_netConfig.listenPort /* UDP */, m_netConfig.listenPort /* TCP */),
+        m_netConfig.discovery, m_netConfig.allowLocalDiscovery);
+
+    if (m_capabilities.size())
```
```suggestion
    if (!m_capabilities.empty())
```

libp2p/Host.cpp
```diff
@@ -749,46 +733,54 @@ void Host::startedWorking()
...
+        nodeTable->setEventHandler(new HostNodeTableHandler(*this));
     DEV_GUARDED(x_nodeTable)
-        m_nodeTable = nodeTable;
+    m_nodeTable = nodeTable;
     restoreNetwork(&m_restoreNetwork);
 
-    LOG(m_logger) << "p2p.started id: " << id();
-
-    run(boost::system::error_code());
+    if (m_capabilities.size())
```
```suggestion
    if (!m_capabilities.empty())
```

libp2p/Host.cpp
```diff
@@ -749,46 +733,54 @@ void Host::startedWorking()
...
+    {
+        LOG(m_logger) << "Capabilities detected, p2p started";
```
```suggestion
        LOG(m_logger) << "P2P started (capabilities: " << m_capabilities.size() << ")";
```

libp2p/Host.cpp
```diff
@@ -749,46 +733,54 @@ void Host::startedWorking()
...
+        run(boost::system::error_code());
+    }
+    else
+        LOG(m_logger) << "No capabilities detected, p2p not started";
```
```suggestion
        LOG(m_logger) << "P2P not started (no capabilities)";
```

libp2p/Host.cpp
```diff
@@ -749,46 +733,54 @@ void Host::startedWorking()
...
+            m_netConfig.listenPort /* UDP */, m_netConfig.listenPort /* TCP */),
+        m_netConfig.discovery, m_netConfig.allowLocalDiscovery);
+
+    if (m_capabilities.size())
+        // Only set the event handler if there are capabilities since otherwise p2p
+        // isn't running
+        nodeTable->setEventHandler(new HostNodeTableHandler(*this));
     DEV_GUARDED(x_nodeTable)
-        m_nodeTable = nodeTable;
+    m_nodeTable = nodeTable;
```
chfast: This is formatting issue with using the `DEV_GUARDED` macro. Use `{}`.

libp2p/Host.cpp
gumb0: This is the old code, only indentation changed; to me using the for loop itself is strange here; counter not needed at all, it could be a flag.

chfast: Yes, I know. Use flag here. The rest should stay.

libp2p/Host.cpp
```diff
@@ -749,46 +733,54 @@ void Host::startedWorking()
     // initialization (e.g. start capability threads, start TCP listener, and kick off timers)
     asserts(!m_timer);
```
gumb0: As we don't delete `m_timer` now, this assert will fail when we stop and then restart network

libp2p/Host.cpp
halfalicious: Will replace this (and the code above it which disconnects pending handshakes) with a flag

halfalicious:
@gumb0 : From what I understand, before my changes `Host::stop()` waited on `Host::run()` because run ran some network shut down code (e.g. delete `m_timer ` and the node table, stop the ioservice). `Host::stop()` no longer waits on run with my changes because the network shutdown code has either been moved to `Host::stop()` (in the case of stopping the ioservice) or is now done implicitly as a part of `Host `destruction. I've structured the code this way based on the assumption that stopping the ioservice also stops future `run()` iterations from being scheduled via the timer and also cancels `run()` iterations that are already scheduled from executing (since after the ioservice is stopped I'm thinking any linked objects are effectively disabled and therefore the timer won't fire - please correct me if I'm wrong here). 

I'd be happy to test network restart, how do I go about doing this via RPC? I attached to aleth via geth but I don't see any functions for stopping/starting the network (I checked all of the modules e.g. admin and debug). I looked through the Aleth RPC source code and found a couple of functions which look promising (`AdminNet::admin_net_start` / `AdminNet::admin_net_stop`) but I don't know how to call them. Do I need to use another tool?

libp2p/Host.cpp
```diff
@@ -749,46 +733,54 @@ void Host::startedWorking()
...
-        m_ioService,
-        m_alias,
-        NodeIPEndpoint(bi::address::from_string(listenAddress()), listenPort(), listenPort()),
-        m_netConfig.discovery,
-        m_netConfig.allowLocalDiscovery
-    );
-    nodeTable->setEventHandler(new HostNodeTableHandler(*this));
+
+    auto nodeTable = make_shared<NodeTable>(m_ioService, m_alias,
+        // Use data from network configuration rather than Host's settings because p2p might not
```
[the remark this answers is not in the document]
halfalicious: I'm a little confused about the purpose of `determinePublic()` and `m_tcpPublic`:
* `determinePublic()` configures `m_tcpPublic` but `m_tcpPublic `doesn't appear to actually be used for anything i.e. it's not tied to the host's tcp acceptor - in `Network::tcp4Listen` (which is where we create the tcp acceptor) we create a new endpoint and use it when opening the tcp acceptor, we don't use `m_tcpPublic` here: https://github.com/ethereum/aleth/blob/91787f239566501a2b23f17946618ddf6a3cee15/libp2p/Network.cpp#L141-L156

* `listenAddress()` and `listenPort()` are host-specific i.e. they expose the host's listen information - I've updated them to check if capabilities exist and if so, they return the host's listen information but if not, they call into new NodeTable functions to obtain the node table's listen information. 
*` listenAddress() `and `listenPort()` query the host's listen information from the `m_tcpPublic `endpoint - since `m_tcpPublic `isn't actually used for anything, they should instead presumably query the host's tcp acceptor for listen information (e.g. they can obtain the acceptor's local endpoint via `boost::asio::ip::tcp::acceptor::local_endpoint()` and query it for the listen information)?

* I'm not completely clear on the difference between a public vs private IP address from Aleth's perspective - we enable the user to supply both (via the `--listen-ip` and `--public-ip` args) and both are saved in the network config (`NetworkConfig::listenIPAddress` and `NetworkConfig::publicIPAddress`). Why do we need to let the user supply an IP address at all, is this in case the user's machine has more than 1 network interface?

halfalicious:
The EIP8 tests are failing due to an AV:

```
The following tests FAILED:
	464 - eip8/test_handshake_plain_auth (Failed)
	465 - eip8/test_handshake_eip8_auth1 (Failed)
	466 - eip8/test_handshake_eip8_auth2 (Failed)
	467 - eip8/test_handshake_eip8_ack_plain (Failed)
	468 - eip8/test_handshake_eip8_ack1 (Failed)
	469 - eip8/test_handshake_eip8_ack2 (Failed)
```

These tests aren't built on Windows so I'm syncing/building by Ubuntu enlistment and will investigate. This might be due to my recent change which modifies the Host "listen" functions to query the node table if no capabilities have been registered (assuming the functions are called before the network thread is started)

libp2p/Host.cpp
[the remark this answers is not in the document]
gumb0: `m_tcpPublic` is used by the NodeTable as our IP we advertise to others (`NodeIPEndpoint` passed to `NodeTable` constructor uses `listenAddress()` which uses `m_tcpPublic`)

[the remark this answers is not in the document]
halfalicious: Ah of course, then presumably if we don’t advertise a public IP we can only connect to others but no one else can connect to us? Or does NAT magically take care of that?

halfalicious:
> 
> 
> Make new PR out of this commit.

Will do once I've ironed out most of the remaining issues, you're right this one has gotten a bit cluttered 😄

libp2p/Host.cpp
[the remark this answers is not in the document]
gumb0: >  However, why do we need to support both options rather than just a single IP

`listenIPAddress` is used to listen to new connections (see `Network::tcp4Listen` function), it's our address as we see it behind the NAT
`publicIPAddress` is what we give out to others through Discovery, it's our address as seen out of NAT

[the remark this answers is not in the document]
gumb0: > Ah of course, then presumably if we don’t advertise a public IP we can only connect to others but no one else can connect to us? 

More exactly they can't discover our public IP automatically; they still can add us as a peer manually if they know our public address through some other way.

libp2p/Host.cpp
```diff
@@ -667,22 +672,7 @@ size_t Host::peerCount() const
 void Host::run(boost::system::error_code const&)
```
gumb0: I think we should better check this error code, so that it doesn't proceed when io_service is stopped.

libp2p/Network.cpp
```diff
@@ -250,3 +250,32 @@ bi::tcp::endpoint Network::resolveHost(string const& _addr)
...
+     else if (publicIsHost)
+     {
+         cnetnote << "Public address set to Host configured address: " << paddr;
+         ep.setAddress(paddr);
+     }
+     else if (pset)
+         ep.setAddress(paddr);
+
+     return ep;
+}
```
gumb0: Did you remove `traverseNAT` part?

libp2p/NodeTable.h
```diff
@@ -269,7 +273,9 @@ class NodeTable : UDPSocketEvents
     std::unique_ptr<NodeTableEventHandler> m_nodeEventHandler;		///< Event handler for node events.
 
     NodeID const m_hostNodeID;
-    NodeIPEndpoint m_hostNodeEndpoint;
+
+    /// Host's endpoint if p2p is started, otherwise an endpoint specific to the NodeTable
+    NodeIPEndpoint m_nodeEndpoint;
```
gumb0: I think it was clearer as `m_hostNodeEndpoint`, less ambiguity about which node is that (as there are many nodes in `NodeTable` code)

> /// Host's endpoint if p2p is started, otherwise an endpoint specific to the NodeTable
What do you mean? it's the same thing, no?

libp2p/Host.h
```diff
@@ -174,10 +173,31 @@ class Host: public Worker
     size_t peerCount() const;
 
     /// Get the address we're listening on currently.
-    std::string listenAddress() const { return m_tcpPublic.address().is_unspecified() ? "0.0.0.0" : m_tcpPublic.address().to_string(); }
+    std::string listenAddress() const
```
gumb0: Why is this change needed?

libp2p/Host.cpp
```diff
@@ -691,19 +633,19 @@ void Host::run(boost::system::error_code const&)
...
-    
+
     // At this time peers will be disconnected based on natural TCP timeout.
     // disconnectLatePeers needs to be updated for the assumption that Session
     // is always live and to ensure reputation and fallback timers are properly
     // updated. // disconnectLatePeers();
 
     // todo: update peerSlotsAvailable()
-    
+
```
gumb0: extra whitespace

libp2p/Host.cpp
```diff
@@ -722,13 +664,13 @@ void Host::run(boost::system::error_code const&)
     for (auto p: toConnect)
         if (p->peerType == PeerType::Required && reqConn++ < m_idealPeerCount)
             connect(p);
-    
+
```
gumb0: extra whitespace

libp2p/Host.cpp
```diff
@@ -737,58 +679,69 @@ void Host::run(boost::system::error_code const&)
             }
         }
     }
-
+
```
gumb0: extra whitespace

libp2p/Host.cpp
```diff
@@ -164,38 +162,45 @@ void Host::doneWorking()
         m_ioService.poll();
 
     // stop capabilities (eth: stops syncing or block/tx broadcast)
-    for (auto const& h: m_capabilities)
+    for (auto const& h : m_capabilities)
         h.second->onStopping();
 
     // disconnect pending handshake, before peers, as a handshake may create a peer
-    for (unsigned n = 0;; n = 0)
+    bool handshakeDisconnected;
```
gumb0: It could be defined inside the loop.

libp2p/Host.cpp
```diff
@@ -164,38 +162,45 @@ void Host::doneWorking()
...
+        }
+        if (!handshakeDisconnected)
             break;
         m_ioService.poll();
     }
-    
+
     // disconnect peers
-    for (unsigned n = 0;; n = 0)
+    bool peerDisconnected;
```
gumb0: Move definition inside `while (true)` loop

gumb0:
> The EIP8 tests are failing due to an AV:

Interestingly those failures look similar to why it was disabled on Windows https://github.com/ethereum/aleth/issues/3273

I suspect it's some data race that by coincidence worked fine everywhere but not on Windows. And now something changed and it's not working anywhere.

So you might also enable it on Windows and try debug there, if it's easier.

(currently there are build failures anyway)

libp2p/NodeTable.h
halfalicious: No, from what I understand the node table only shares the host's endpoint if p2p is running (i.e. the UDP socket is bound to the same IP/port as the host). If there's no p2p then the endpoint isn't shared with any other component and is only used by the node table.

libp2p/Host.h
halfalicious: m_tcpPublic won't be initialized if devp2p isn't started, so we should instead return the node table's listen information rather than the host's (m_tcpPublic) in this case.

halfalicious:
The work in this PR has been split into two separate PRs:
* #5463 : [WIP] Don't start TCP listener without registered capabilities
* #5454 : Streamline host shutdown logic and fix use-after-free

PR #5406, opened 2018-12-19
Ping before sending FindNode

gumb0:
Addresses point 4 of https://github.com/ethereum/aleth/issues/4959

### DIFF
--- a/test/unittests/libp2p/net.cpp
+++ b/test/unittests/libp2p/net.cpp
@@ -21,6 +21,7 @@
 
 #include <libdevcore/Assertions.h>
 #include <libdevcore/Worker.h>
+#include <libdevcore/concurrent_queue.h>
 #include <libdevcrypto/Common.h>
 #include <libp2p/NodeTable.h>
 #include <libp2p/UDP.h>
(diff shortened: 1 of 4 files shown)


test/unittests/libp2p/net.cpp
```diff
@@ -172,10 +199,10 @@ struct TestNodeTable: public NodeTable
     {
         NodeTable::onPacketReceived(_socket, _from, _packet);
 
-        packetReceived.set_value(_packet.toBytes());
```
gumb0: This was not working correctly when more than one packet was received - `set_value` throws if `promise` already has a value.

gumb0:
The unit test is not great, but it's something. Don't want to spend more time on it now, maybe I'll improve it / add more tests later.

test/unittests/libp2p/net.cpp
```diff
@@ -388,7 +414,7 @@ BOOST_AUTO_TEST_CASE(noteActiveNodeAppendsNewNode)
     TestNodeTableHost nodeTableHost(1);
     nodeTableHost.populate(1);
 
-    auto nodeTable = nodeTableHost.nodeTable;
+    auto& nodeTable = nodeTableHost.nodeTable;
```
halfalicious: Nice catch :smile:

halfalicious:
As you mentioned, it would be nice to add some tests which verify that the ping is only sent before the FindNode if the pong time is older than 12 hours (this would probably require some changes to `TestNodeTableHost`) but that can be taken care of in a future change.

test/unittests/libp2p/net.cpp
```diff
@@ -56,6 +56,33 @@ class TestHost: public Worker
     ba::io_service m_io;
 };
 
+template <typename T>
+class ConcurrentQueue
```
chfast: Have you seen https://github.com/ethereum/aleth/blob/master/libdevcore/concurrent_queue.h ?

test/unittests/libp2p/net.cpp
chfast: Maybe block copy constructor?

gumb0: It's a `shared_ptr` anyway

test/unittests/libp2p/net.cpp
gumb0: Oh I haven't! Looks very similar, I'll remove mine then.

Where is it used though?

chfast: Seems nowhere currently. I was using it for SmartVM previously to JIT contracts in background thread. You can adjust it to your needs.

PR #5407, opened 2018-12-20
CMake: Use basic compiler settings from Cable

chfast:


gumb0:
Windows build still fails

chfast:
Yes I know. Looks like I have to spin off Windows machine to solve this.

Issue #4959, opened 2018-04-16
Node Discovery security improvements


gumb0:
1. When we receive Pong packet, we should validate that it's a response to some of the Pings that we've sent.
https://github.com/ethereum/devp2p/blob/master/discv4.md#endpoint-proof

See also related fix of the Pong packet https://github.com/ethereum/cpp-ethereum/pull/4915

gumb0:
Also before sending `FindNode` ourselves, we need to check whether we've send this node our `Pong` in the last 12 hours, and initiate `Ping` if we didn't
https://github.com/ethereum/go-ethereum/blob/a7501d0c419f8c1dc8b4004c115dd2d4e3b524b7/p2p/discover/udp.go#L326-L331

halfalicious:
I’ll work on #2, the fix should be fairly trivial so I should have something ready for PR (including unit tests) soon

gumb0:
3 is left to implement here

halfalicious:
I'm working on item 3 and should have something ready in the next day or 2 (I've finished the feature changes and have created a test but I think I want to make the test more stable).

PR #5412, opened 2018-12-31
Check for endpoint proof before processing FindNode request

halfalicious:
Address issue 2 in #4959 

Before processing a FindNode request, validate that the endpoint proof has been completed within the timeout window (12 hours). If so, service the request and send neighbors. Otherwise, ignore the request.

### DIFF
--- a/libdevcore/concurrent_queue.h
+++ b/libdevcore/concurrent_queue.h
@@ -1,64 +1,60 @@
-/*
-	This file is part of cpp-ethereum.
+// Aleth: Ethereum C++ client, tools and libraries.
+// Copyright 2018 Aleth Authors.
+// Licensed under the GNU General Public License, Version 3.
 
-	cpp-ethereum is free software: you can redistribute it and/or modify
-	it under the terms of the GNU General Public License as published by
-	the Free Software Foundation, either version 3 of the License, or
-	(at your option) any later version.
-
-	cpp-ethereum is distributed in the hope that it will be useful,
-	but WITHOUT ANY WARRANTY; without even the implied warranty of
-	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-	GNU General Public License for more details.
-
-	You should have received a copy of the GNU General Public License
-	along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
-*/
 #pragma once
 
 #include "Exceptions.h"
 
-#include <utility>
-#include <queue>
 #include <condition_variable>
 #include <mutex>
+#include <queue>
+#include <utility>
 
 namespace dev
 {
-
 /// Concurrent queue.
 /// You can push and pop elements to/from the queue. Pop will block until the queue is not empty.
 /// The default backend (_QueueT) is std::queue. It can be changed to any type that has
 /// proper push(), pop(), empty() and front() methods.
-template<typename _T, typename _QueueT = std::queue<_T>>
+template <typename _T, typename _QueueT = std::queue<_T>>
 class concurrent_queue
 {
 public:
-	template<typename _U>
-	void push(_U&& _elem)
-	{
-		{
-			std::lock_guard<decltype(x_mutex)> guard{x_mutex};
-			m_queue.push(std::forward<_U>(_elem));
-		}
-		m_cv.notify_one();
-	}
+    template <typename _U>
+    void push(_U&& _elem)
+    {
+        {
+            std::lock_guard<decltype(x_mutex)> guard{x_mutex};
+            m_queue.push(std::forward<_U>(_elem));
+        }
+        m_cv.notify_one();
+    }
 
-	_T pop(std::chrono::milliseconds const _waitDuration = std::chrono::milliseconds::max())
+	_T pop()
 	{
-		std::unique_lock<std::mutex> lock{x_mutex};
-        if (!m_cv.wait_for(lock, _waitDuration, [this] { return !m_queue.empty(); }))
-            BOOST_THROW_EXCEPTION(WaitTimeout());
-
+		std::unique_lock<std::mutex> lock{ x_mutex };
+		m_cv.wait(lock, [this] { return !m_queue.empty(); });
 		auto item = std::move(m_queue.front());
 		m_queue.pop();
 		return item;
 	}
-    
+
+    _T pop(std::chrono::milliseconds const& _waitDuration)
+    {
+        std::unique_lock<std::mutex> lock{x_mutex};
+        if (!m_cv.wait_for(lock, _waitDuration, [this] { return !m_queue.empty(); }))
+            BOOST_THROW_EXCEPTION(WaitTimeout());
+
+        auto item = std::move(m_queue.front());
+        m_queue.pop();
+        return item;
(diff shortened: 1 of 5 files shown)


halfalicious:
Still need to author unit tests

halfalicious:
Need to investigate failing tests after concurrent queue changes:

```
The following tests FAILED:
	481 - network/net/unexpectedPong (Failed)
	482 - network/net/invalidPong (Failed)
	483 - network/net/validPong (Failed)
	484 - network/net/pingTimeout (Failed)
	485 - network/net/evictionWithOldNodeAnswering (Failed)
	488 - network/net/findNodeIsSentAfterPong (Failed)
```

libdevcore/concurrent_queue.h
```diff
@@ -42,15 +44,17 @@ class concurrent_queue
 		m_cv.notify_one();
 	}
 
-	_T pop()
+	_T pop(std::chrono::milliseconds const _waitDuration = std::chrono::milliseconds::max())
 	{
 		std::unique_lock<std::mutex> lock{x_mutex};
-		m_cv.wait(lock, [this]{ return !m_queue.empty(); });
+        if (!m_cv.wait_for(lock, _waitDuration, [this] { return !m_queue.empty(); }))
```
halfalicious: Need to fix indentation

halfalicious:
Need to investigate why some tests are failing on Windows

halfalicious:
Rebased

libdevcore/concurrent_queue.h
```diff
@@ -1,60 +1,51 @@
...
+    void push(_U&& _elem)
+    {
+        {
+            std::lock_guard<decltype(x_mutex)> guard{x_mutex};
+            m_queue.push(std::forward<_U>(_elem));
+        }
+        m_cv.notify_one();
+    }
+
+    _T pop(std::chrono::milliseconds const _waitDuration = std::chrono::milliseconds{60000})
```
gumb0: `condition_variable::wait_for` gets duration by const reference, and you should do, too.

libp2p/NodeTable.cpp
```diff
@@ -482,6 +490,20 @@ void NodeTable::onPacketReceived(
...
+                if (!sourceNodeEntry)
+                {
+                    LOG(m_logger) << "Source node (" << packet->sourceid << "@" << _from
+                                  << ") not found in node table. Ignoring FindNode request.";
+                    return;
+                }
+                if (sourceNodeEntry->lastPongReceivedTime + c_bondingTimeSeconds <
+                    RLPXDatagramFace::secondsSinceEpoch())
+                {
+                    LOG(m_logger) << "Endpoint proof has expired! Ignoring FindNode request";
```
gumb0: Not quite precise, could be also that we haven't received any proof yet.

libp2p/NodeTable.cpp
```diff
@@ -482,6 +490,20 @@ void NodeTable::onPacketReceived(
...
             case FindNode::type:
             {
+                auto const sourceNodeEntry = nodeEntry(packet->sourceid);
+                if (!sourceNodeEntry)
+                {
+                    LOG(m_logger) << "Source node (" << packet->sourceid << "@" << _from
+                                  << ") not found in node table. Ignoring FindNode request.";
+                    return;
+                }
+                if (sourceNodeEntry->lastPongReceivedTime + c_bondingTimeSeconds <
```
gumb0: Maybe a helper for this check would be nice, because now it's repeated in two places.

libp2p/NodeTable.h
```diff
@@ -153,7 +153,7 @@ class NodeTable : UDPSocketEvents
     /// Called by implementation which provided handler to process NodeEntryAdded/NodeEntryDropped events. Events are coalesced by type whereby old events are ignored.
     void processEvents();
 
-    /// Add node. Node will be pinged.
+    /// Add node to the list of all nodes and if the node is known, also add it to the node table.
```
gumb0: Well it also pings the unknown one, maybe we should keep it in the comment.

libp2p/NodeTable.h
```diff
@@ -276,11 +277,15 @@ class NodeTable : UDPSocketEvents
     Secret m_secret;												///< This nodes secret key.
 
     mutable Mutex x_nodes;											///< LOCK x_state first if both locks are required. Mutable for thread-safe copy in nodes() const.
-    std::unordered_map<NodeID, std::shared_ptr<NodeEntry>> m_allNodes;  ///< Known Node Endpoints
+    std::unordered_map<NodeID, std::shared_ptr<NodeEntry>>
+        m_allNodes;  ///< Node endpoints. Includes all nodes that we've been in contact with and
```
gumb0: Formatting like this is weird, move multi-line comments to the lines preceding the variable please.
(you should also remove `<` symbols then)

libp2p/NodeTable.cpp
halfalicious: Good point...maybe a better log message would be something like “unexpected FindNode request”

libp2p/NodeTable.h
halfalicious: Will do...this is automatic clang formatting btw, we should also look into updating our format config

libdevcore/concurrent_queue.h
halfalicious: Thanks for calling this out, I specifically didn’t because the underlying type was something like long long which is the same size as pointer type on 64 bit systems. In the interest of conformance I’ll switch to const ref.

libp2p/NodeTable.cpp
halfalicious: Going to split this into 2 cases and log 2 different messages - 1 where we haven't performed the endpoint proof yet (lastPongReceivedTime == 0) and 1 where the endpoint proof has expired.

libp2p/NodeTable.cpp
halfalicious: Making this a helper on NodeEntry

libp2p/NodeTable.h
halfalicious: Good point - I'll rephrase the comment to call this out

halfalicious:
This test is also failing now: 
```
Running 1 test case...
Test Case "findNodeIsSentAfterPong": 
/home/builder/project/test/unittests/libp2p/net.cpp(832): error: in "network/net/findNodeIsSentAfterPong": check datagram3->typeName() == "FindNode" has failed [Ping != FindNode]
```

halfalicious:
@chfast : Thanks for the review! What's an example of a format issue that I missed? I ask because I format my changes using VS Code's clang format support (though occasionally I miss things, I still need to work on getting clang-format working from the command line) but sometimes the formatting comes out strange, I suspect this is due to issues in our `.clang-format` file...if there are problems with our format definitions it would be nice to catalog and fix them 😃

halfalicious:
@chfast : Thanks for the format examples! Those are whitespace-only issues which I think are pre-existing, I've reformatted them using clang-format. I'll work on getting clang-format working from the command line so I can automatically reformat all of my modified code before committing (for future changes).

halfalicious:
Rebased + squashed a bunch of changes

PR #5416, opened 2019-01-04
Allow less strict account in genesis config

winsvega:
https://github.com/ethereum/aleth/issues/5414

This also applies for tests (same code check accounts in tests). which is not really good as in tests account description is strict (4 fields only). but for tests we have a second scheme validation at python scripts in test repo.

### DIFF
--- a/libethereum/ValidationSchemes.cpp
+++ b/libethereum/ValidationSchemes.cpp
@@ -130,8 +130,7 @@ void validateAccountObj(js::mObject const& _obj)
                 {c_codeFromFile, {{js::str_type}, JsonFieldPresence::Optional}}});
 
         // At least one field must be set
-        if (!_obj.count(c_code) && !_obj.count(c_nonce) && !_obj.count(c_storage) &&
-            !_obj.count(c_balance) && !_obj.count(c_wei) && !_obj.count(c_codeFromFile))
+        if (_obj.empty())
         {
             string comment =
                 "Error in validateAccountObj: At least one field must be set (code, nonce, "
--- a/libethereum/ValidationSchemes.cpp
+++ b/libethereum/ValidationSchemes.cpp
@@ -118,36 +118,41 @@ void validateAccountObj(js::mObject const& _obj)
                 {c_wei, {{js::str_type}, JsonFieldPresence::Optional}},
                 {c_balance, {{js::str_type}, JsonFieldPresence::Optional}}});
     }
-    else if (_obj.size() == 1)
-    {
-        // A genesis account with only balance set
-        if (_obj.count(c_balance))
-            requireJsonFields(_obj, "validateAccountObj",
-                {{c_balance, {{js::str_type}, JsonFieldPresence::Required}}});
-        else
-            requireJsonFields(_obj, "validateAccountObj",
-                {{c_wei, {{js::str_type}, JsonFieldPresence::Required}}});
-    }
     else
     {
-        if (_obj.count(c_codeFromFile))
+        // Extendable account description. Check typo errors and fields type.
+        requireJsonFields(_obj, "validateAccountObj",
+            {{c_code, {{js::str_type}, JsonFieldPresence::Optional}},
+                {c_nonce, {{js::str_type}, JsonFieldPresence::Optional}},
+                {c_storage, {{js::obj_type}, JsonFieldPresence::Optional}},
+                {c_balance, {{js::str_type}, JsonFieldPresence::Optional}},
+                {c_wei, {{js::str_type}, JsonFieldPresence::Optional}},
+                {c_codeFromFile, {{js::str_type}, JsonFieldPresence::Optional}}});
+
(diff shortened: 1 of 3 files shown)


test/unittests/libethereum/ValidationSchemes.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include <test/tools/libtesteth/JsonSpiritHeaders.h>
+#include <test/tools/libtesteth/TestHelper.h>
+
+using namespace std;
+using namespace dev;
+using namespace dev::eth;
+using namespace dev::test;
+namespace js = json_spirit;
+
+BOOST_FIXTURE_TEST_SUITE(ValidationSchemes, TestOutputHelperFixture)
```
chfast: These tests must go to `aleth-unittests`

libethereum/ValidationSchemes.cpp
```diff
@@ -118,37 +118,42 @@ void validateAccountObj(js::mObject const& _obj)
...
+        requireJsonFields(_obj, "validateAccountObj",
+            {{c_code, {{js::str_type}, JsonFieldPresence::Optional}},
+                {c_nonce, {{js::str_type}, JsonFieldPresence::Optional}},
+                {c_storage, {{js::obj_type}, JsonFieldPresence::Optional}},
+                {c_balance, {{js::str_type}, JsonFieldPresence::Optional}},
+                {c_wei, {{js::str_type}, JsonFieldPresence::Optional}},
+                {c_codeFromFile, {{js::str_type}, JsonFieldPresence::Optional}}});
+
+        // At least one field must be set
+        if (!_obj.count(c_code) && !_obj.count(c_nonce) && !_obj.count(c_storage) &&
```
chfast: Why this is required?

winsvega: What might be the purpose of
```"account":{}``` ?

File: test/unittests/libethereum/ValidationSchemes.cpp (hunk 1 above)
Comment (winsvega) [reply test/unittests/libethereum/ValidationSchemes.cpp]: Where is that?

File: libethereum/ValidationSchemes.cpp (hunk 2 above)
Comment (winsvega) [reply libethereum/ValidationSchemes.cpp]: What is empty account? We have different definitions for different hardfork rules

Comment (chfast) [inline libethereum/ValidationSchemes.cpp]:
```suggestion
        if (_obj.empty())
```

Or whatever the equivalent of this is in the used JSON library.

winsvega: Yeah, wrong fields should be checked in verify json above.

test/unittests/libethereum/ValidationSchemes.cpp
winsvega: 1) why aleth-unittests ? 
2) how alet-unittest run? is it set to autorun on travis? 
3) to add it to aleth I have to exclude that file from unittests and rewrite it using GTest suite.

winsvega: do you plan to switch all unittests from testeth to aleth-unittest using gtest only ?

test/unittests/libethereum/ValidationSchemes.cpp
```diff
@@ -0,0 +1,192 @@
+/*
+    This file is part of cpp-ethereum.
```
chfast: Please change the note to this one: https://github.com/ethereum/aleth/issues/5403.

test/unittests/libethereum/ValidationSchemes.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+    cpp-ethereum is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/** @file ValidationSchemes.cpp
```
chfast: We don't put these tags here. Remove `@file`, `@author`, `@date`.

test/unittests/libethereum/ValidationSchemes.cpp
chfast: Only true unit tests: the ones that do not read JSON files.

winsvega:
gcc6 log:
```
    Uploading to S3...
Error: ('Connection broken: IncompleteRead(0 bytes read)', IncompleteRead(0 bytes read))
```

test/unittests/libethereum/ValidationSchemes.cpp
```diff
@@ -0,0 +1,177 @@
+// Aleth: Ethereum C++ client, tools and libraries.
+// Copyright 2018 Aleth Authors.
```
chfast: It's 2019.

PR #5419, opened 2019-01-08
Enable Constantinople on the main net

gumb0:
Setting fork block number according to https://github.com/ethereum/EIPs/pull/1642


chfast:
Why it show that the whole file was changed?

gumb0:
@chfast tabs->spaces

PR #5423, opened 2019-01-09
testeth: Remove dot output for Travis CI

chfast:
When I run single VM test, I can feel testeth lagging at the end because it waits for the dot thread to join.
It also pollutes the output with the dots in random way so comparing two outputs is harder.
And most importantly, we don't need this any more.

### DIFF
--- a/test/tools/libtesteth/boostTest.cpp
+++ b/test/tools/libtesteth/boostTest.cpp
@@ -1,19 +1,7 @@
-/*
-    This file is part of cpp-ethereum.
+// Aleth: Ethereum C++ client, tools and libraries.
+// Copyright 2018 Aleth Authors.
+// Licensed under the GNU General Public License, Version 3.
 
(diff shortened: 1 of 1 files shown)


winsvega:
why?

chfast:
Not needed and slows down testeth termination.

winsvega:
What if test run in a slow env and no output will be produced for  10 min?

chfast:
> Why not needed? Is travis 10 minute no output issue fixed?

We don't use Travis CI.

winsvega:
Oh. So Circleci does not have such a problem? 
Ok. we could try. this change was a workaround for travis anyway.

chfast:
> Oh. So Circleci does not have such a problem?

No. It is much faster and I don't think it has any "no output" timeouts. Besides, we are running test with ctest now, so the dots are not visible anyway.

PR #5424, opened 2019-01-10
Remove network start/stop RPC methods

halfalicious:
Removing these functions will enable us to significantly simplify the design of the Host class's start/stop logic due to the lack of needing to support network restart.

### DIFF
--- a/libweb3jsonrpc/AdminNet.cpp
+++ b/libweb3jsonrpc/AdminNet.cpp
@@ -12,20 +12,6 @@ using namespace dev::rpc;
 
 AdminNet::AdminNet(NetworkFace& _network, SessionManager& _sm): m_network(_network), m_sm(_sm) {}
 
-bool AdminNet::admin_net_start(std::string const& _session)
-{
-	RPC_ADMIN;
-	m_network.startNetwork();
-	return true;
-}
-
-bool AdminNet::admin_net_stop(std::string const& _session)
-{
-	RPC_ADMIN;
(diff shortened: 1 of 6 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
+{
```
chfast: Was this suppose to be here?

halfalicious: Ah definitely not, will remove

halfalicious:
cc @chfast

halfalicious:
@gumb0 : Maybe, if we make changes to WebThreeDirect...Aleth currently starts the network via WebThreeDirect::startNetwork() (but doesn't stop it, ideally it should stop the network before it writes its peer list to the network configuration file so the host/node table's state is in sync with the configuration file by the time Aleth fully exits). We might be able to remove this startNetwork call and start the network in the WebThreeDirect ctor instead and perhaps stop the network in the Host function which writes the network data to disk, though I'd have to do some investigation/testing to be sure.

I don't think this is super important to do right now, I'd be happy to take a task to do this once I've wrapped up my aleth-bootnode/discovery/rocksdb work though 😃

halfalicious:
> 
> 
> Something is wrong with the tests.

Yea, the failing test is one of the unreliable discovery tests:


```
The following tests FAILED:
	499 - network/net/evictionWithOldNodeAnswering (Failed)
Errors while running CTest
Exited with code 8
```

PR #5428, opened 2019-01-10
p2p: Fix PING handling

chfast:


libp2p/NodeTable.cpp
```diff
@@ -528,16 +530,17 @@ void NodeTable::onPacketReceived(
                 auto& in = dynamic_cast<PingNode&>(*packet);
                 in.source.setAddress(_from.address());
                 in.source.setUdpPort(_from.port());
-                addNode(Node(in.sourceid, in.source));
-                
+                auto entry = addNode(Node(in.sourceid, in.source));
+                if (entry)
+                    entry->lastPongSentTime = RLPXDatagramFace::secondsSinceEpoch();
+
+                // Send PONG response.
```
gumb0: We shouldn't send PONG if it failed to be added

libp2p/NodeTable.h
```diff
@@ -150,7 +150,9 @@ class NodeTable : UDPSocketEvents
     /// proof for it or it has been restored from the network config), also add it to the node table.
     /// If the node is unknown (i.e. we haven't completed the endpoint proof for it yet) then ping
     /// it to trigger the endpoint proof.
-    void addNode(Node const& _node, NodeRelation _relation = NodeRelation::Unknown);
+    ///
+    /// @return Pointer to added entry on null if node not added.
+    NodeEntry* addNode(Node const& _node, NodeRelation _relation = NodeRelation::Unknown);
```
gumb0: I'd prefer this not to be a raw pointer, especially because it's a public method of the class. Either return bool and lookup once again (don't bother micro-optimizing) or return shared_ptr

chfast: Changed to `bool`.

libp2p/NodeTable.cpp
chfast: Changed.

PR #5434, opened 2019-01-15
remove EthashClient

winsvega:
*rebased
continue to work on this topic from https://github.com/ethereum/aleth/pull/5179

Remove EthashClient.cpp
Move EthashClient methods to Ethash.cpp
relates to #5114


libethereum/Client.h
```diff
@@ -228,6 +228,11 @@ class Client: public ClientBase, protected Worker
         return m_onChainChanged.add(_handler);
     }
 
+    ///< Is there an active and valid remote worker?
+    bool remoteActive() const;
```
gumb0: Why did you make it public?

winsvega:
Could be unresolved function declaration. 
But error say different

libethcore/SealEngine.h
```diff
@@ -47,6 +47,11 @@ class SealEngineFace
 public:
 	virtual ~SealEngineFace() {}
 
+    virtual std::tuple<h256, h256, h256> getWork(BlockHeader const& _bi)
```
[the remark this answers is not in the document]
gumb0:
```suggestion
    /// @returns Tuple of hash of the current block to be mined minus nonce, seed hash, target boundary.
    virtual std::tuple<h256, h256, h256> getWork(BlockHeader const& _bi)
```

libethcore/SealEngine.h
```diff
@@ -47,6 +47,11 @@ class SealEngineFace
...
+    {
+        return std::tuple<h256, h256, h256>(_bi.hash(WithoutSeal), h256(0), h256(0));
```
gumb0: I guess this hash calculation is not needed
```suggestion
        return std::tuple<h256, h256, h256>{};
```

libethereum/Client.cpp
```diff
@@ -930,3 +930,21 @@ ExecutionResult Client::call(Address const& _from, u256 _value, Address _dest, b
     }
     return ret;
 }
+
+std::tuple<h256, h256, h256> Client::getWork()
+{
+    bool oldShould = shouldServeWork();
```
gumb0: The comment got lost
```suggestion
    // lock the work so a later submission isn't invalidated by processing a transaction elsewhere.	
    // this will be reset as soon as a new block arrives, allowing more transactions to be processed.
    bool oldShould = shouldServeWork();
```

libethereum/Client.h
```diff
@@ -260,6 +262,9 @@ class Client: public ClientBase, protected Worker
     /// Submit
     virtual bool submitSealed(bytes const& _s);
 
+    ///< Is there an active and valid remote worker?
+    bool remoteActive() const;
```
gumb0: Why is this public?

libethereum/ClientBase.cpp
```diff
@@ -527,3 +527,8 @@ int ClientBase::chainId() const
 {
 	return bc().chainParams().chainID;
 }
+
+std::tuple<h256, h256, h256> ClientBase::getWork()
+{
+    return std::tuple<h256, h256, h256>(h256(0), h256(0), h256(0));
```
gumb0:
```suggestion
    return std::tuple<h256, h256, h256>{};
```

libethereum/ClientBase.cpp
```diff
@@ -527,3 +527,8 @@ int ClientBase::chainId() const
 {
 	return bc().chainParams().chainID;
 }
+
+std::tuple<h256, h256, h256> ClientBase::getWork()
```
gumb0: Can we remove it or do we need it? (some tests create `ClientBase`?)

gumb0: If it's because of `ClientTest`, better move this into `ClientTest`

libweb3jsonrpc/Eth.cpp
```diff
@@ -57,26 +57,12 @@ string Eth::eth_coinbase()
...
 {
-	try
-	{
-		return toJS(asEthashClient(client())->hashrate());
-	}
-	catch (InvalidSealEngine&)
-	{
-		BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS));
-	}
+    return toJS(getEthash().hashrate());
```
gumb0: This is not quite good, this way we'll miss any exception coming from `hashrate()` or `toJS()` and it will crash... I suggest leaving `try/catch` here with two `catch` clauses - for `InvalidSealEngine` and for `std::exception`

libweb3jsonrpc/Eth.cpp
```diff
@@ -57,26 +57,12 @@ string Eth::eth_coinbase()
...
 {
-	try
-	{
-		return asEthashClient(client())->isMining();
-	}
-	catch (InvalidSealEngine&)
-	{
-		BOOST_THROW_EXCEPTION(JsonRpcException(Errors::ERROR_RPC_INVALID_PARAMS));
-	}
+    return getEthash().isMining();
```
gumb0: Similar exception handling needed here

libethereum/ClientBase.cpp
winsvega: this is because of line 602 in eth.cpp
```
auto r = client()->getWork();
```

where client() is the interface

gumb0: Some class cannot be instantiated with `getWork()` being pure virtual. If this class is `ClientTest`, move this into `ClientTest`

winsvega: it is FixedClient

gumb0: can you move it into `FixedClient` then?

libweb3jsonrpc/Eth.cpp
```diff
@@ -772,4 +781,16 @@ string dev::rpc::exceptionToErrorMessage()
...
+
+Ethash& Eth::getEthash()
+{
+    try
+    {
+        return asEthash(*client()->sealEngine());
+    }
+    catch (InvalidSealEngine&)
+    {
+        throw JsonRpcException("Seal Engine is not Ethash");
```
gumb0: Now this rethrowing doesn't make much sense, as all calling functions have either `catch(...)` or `catch(std::exception)`.
So I suggest to remove try/catch from this function

libethcore/SealEngine.h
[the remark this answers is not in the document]
gumb0: Please add comment here

PR #5442, opened 2019-01-19
Update network tests to use random ports

halfalicious:
The network tests can sometimes interfere with each other - update them
to use random ports which should reduce the chances of test failures due
to test interference.

### DIFF
--- a/libp2p/Network.h
+++ b/libp2p/Network.h
@@ -37,8 +37,8 @@ namespace dev
 namespace p2p
 {
 
-static constexpr unsigned short c_defaultListenPort = 30303;
-static constexpr const char* c_localhostIp = "127.0.0.1";
+constexpr unsigned short c_defaultListenPort = 30303;
+constexpr const char* c_localhostIp = "127.0.0.1";
 
 struct NetworkConfig
 {
--- a/libp2p/Network.h
+++ b/libp2p/Network.h
@@ -37,7 +37,8 @@ namespace dev
 namespace p2p
 {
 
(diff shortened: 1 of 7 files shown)


libp2p/Network.h
```diff
@@ -37,7 +37,8 @@ namespace dev
 namespace p2p
 {
 
-static const unsigned short c_defaultListenPort = 30303;
+static constexpr unsigned short c_defaultListenPort = 30303;
+static constexpr const char* c_localhostIp = "127.0.0.1";
```
chfast: The `static` is redundant. `constexpr` makes the linkage static.
If this is only used in tests, maybe define it only there?

halfalicious:
Run failed because it exceeded the max CircleCI runtime, is there a way we can increase it or should we look into optimizing the tests?

NAKsir-melody:
Any duplication check for the generated random port numbers?

chfast:
> Any duplication check for the generated random port numbers?

It will not be effective because they are going to run in multiple processes.

gumb0:
It looks like some tests are now hanging sometimes in the endless loop when trying to find port number

test/unittests/libp2p/net.cpp
```diff
@@ -289,7 +297,7 @@ class TestUDPSocketHost : UDPSocketEvents, public TestHost
             }
             catch (std::exception const&)
             {
-                ++port;
+                getRandomPortNumber();
```
gumb0: You forgot to assign it to `port` here

PR #5446, opened 2019-01-22
Fix endless loop in TestUDPSocketHost constructor

gumb0:


gumb0:
Still `network/net/evictionWithOldNodeAnswering` sometimes hangs for me (as before random port selection introduction), I'm trying to find the cause

PR #5447, opened 2019-01-22
Fix hanging evictionWithOldNodeAnswering test

gumb0:
There were sometimes not enough random `NodeEntry`s generated to fill the bucket.

### DIFF
--- a/test/unittests/libp2p/net.cpp
+++ b/test/unittests/libp2p/net.cpp
@@ -61,8 +61,8 @@ struct TestNodeTable: public NodeTable
     static std::vector<std::pair<Public, uint16_t>> createTestNodes(unsigned _count)
     {
         std::vector<std::pair<Public, uint16_t>> ret;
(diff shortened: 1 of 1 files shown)


test/unittests/libp2p/net.cpp
```diff
@@ -61,8 +61,8 @@ struct TestNodeTable: public NodeTable
     static std::vector<std::pair<Public, uint16_t>> createTestNodes(unsigned _count)
     {
         std::vector<std::pair<Public, uint16_t>> ret;
-        asserts(_count < 1000);
-
+        asserts(_count <= 1000);
```
halfalicious: Why is this capped at 1000? Is 1000 an arbitrary limit (but we just need some limit)?

gumb0: Yes, it's arbitrary, I  guess we could also just remove it

PR #5435, opened 2019-01-16
Only send ping after sending pong response to nodes with invalid endpoint proofs

halfalicious:
Fix item 3 in #4959 

When receiving a ping from a node for which we've already completed the endpoint proof, don't send a ping in addition to the pong response. Note that prior to these changes, we wouldn't send a ping after the pong to nodes whose endpoint proofs had expired. 

I've also added a unit test which validates this behavior.

PR #5449, opened 2019-01-23
Serialize Pong sent/received times to network.rlp

gumb0:


halfalicious:
You're probably already aware of this but you'll want to update the _relation == Known code path in NodeTable::addNode to check the deserialized last pong time and ping if it's expired.

gumb0:
@halfalicious Yes, I want to wait with this until https://github.com/ethereum/aleth/pull/5435 is merged

halfalicious:
> 
> 
> @halfalicious Yes, I want to wait with this until #5435 is merged

Excellent thank you 😄 I plan to have it merge-ready in the next couple of days.

gumb0:
I plan to try to split `addNode` into two methods - for usual adding and for known nodes, because these are pretty different cases, the latter used only when restoring from network.rlp and in tests

libp2p/Host.cpp
```diff
@@ -850,6 +844,19 @@ bytes Host::saveNetwork() const
 
     RLPStream network;
     int count = 0;
+    if (auto nodeTable = this->nodeTable())
```
gumb0: Changed the order in network.rlp - first node table contents are saved, then connected peers. Otherwise, as peers are probably present in the node table, too, adding them first would always initiate ping.

libp2p/NodeTable.cpp
```diff
@@ -88,62 +86,83 @@ void NodeTable::processEvents()
...
+
+    return true;
+}
+
+std::shared_ptr<NodeEntry> NodeTable::createNodeEntry(
+    Node const& _node, uint32_t _lastPongReceivedTime, uint32_t _lastPongSentTime)
 {
     if (!_node.endpoint || !_node.id)
     {
         LOG(m_logger) << "Supplied node has an invalid endpoint (" << _node.endpoint << ") or id ("
```
gumb0: Logs in this function are now executed under `x_nodes` mutex, but I don't think it's a big problem, as logging is now asynchronous (in Release build), and also the part to the right of `LOG(...) << ` is executed only when not disabled by log filters.

test/unittests/libp2p/net.cpp
```diff
@@ -546,7 +552,8 @@ BOOST_AUTO_TEST_CASE(noteActiveNodeReplacesNodeInFullBucketWhenEndpointChanged)
     // but added as most recently seen with new endpoint
     auto mostRecentNodeEntry = nodeTable->bucketLastNode(bucketIndex);
     BOOST_CHECK_EQUAL(mostRecentNodeEntry->id, leastRecentlySeenNodeId);
-    BOOST_CHECK_EQUAL(mostRecentNodeEntry->endpoint, newEndpoint);
+    BOOST_CHECK_EQUAL(mostRecentNodeEntry->endpoint.address(), newEndpoint.address());
```
gumb0: I changed the meaning of this test for now, but I think this whole case of changed endpoint should be revisited in https://github.com/ethereum/aleth/issues/5455, and this test probably will be removed.

libp2p/Host.cpp
```diff
@@ -904,76 +897,50 @@ void Host::restoreNetwork(bytesConstRef _b)
...
         {
+            // nodeRLP[0] - IP address
             // todo: ipv6
-            if (i[0].itemCount() != 4 && i[0].size() != 4)
+            if (nodeRLP[0].itemCount() != 4 && nodeRLP[0].size() != 4)
                 continue;
 
-            if (i.itemCount() == 4 || i.itemCount() == 11)
+            Node node((NodeID)nodeRLP[3], NodeIPEndpoint(nodeRLP));
+            if (nodeRLP.itemCount() == 6 && isAllowedEndpoint(node.endpoint))
```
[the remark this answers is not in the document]
chfast: Will the full content of this file be replaced on next save?

libp2p/Host.cpp
```diff
@@ -850,6 +844,19 @@ bytes Host::saveNetwork() const
```
chfast: FYI: This `peers.sort()` is buggy. It works by converting `Peer` objects to bool. Probably not what you wanted. You need to define `operator<` for `Node`.

libp2p/Host.cpp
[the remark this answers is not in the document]
gumb0: Yes, it just overwrites on next exit

libp2p/Host.cpp
gumb0: I removed sorting https://github.com/ethereum/aleth/issues/5470#issuecomment-461088152

gumb0:
Also rebased.

libp2p/Host.cpp
```diff
@@ -547,46 +547,40 @@ void Host::requirePeer(NodeID const& _n, NodeIPEndpoint const& _endpoint)
...
-        // required for discovery
-        addNodeToNodeTable(*p, NodeTable::NodeRelation::Unknown);
+        Guard l(x_requiredPeers);
+        m_requiredPeers.insert(_n);
+    }
+
+    Node const node(_n, _endpoint, PeerType::Required);
+    // create or update m_peers entry
+    shared_ptr<Peer> p;
+    DEV_RECURSIVE_GUARDED(x_sessions)
```
chfast: Add scope `{}`.

libp2p/Host.cpp
```diff
@@ -547,46 +547,40 @@ void Host::requirePeer(NodeID const& _n, NodeIPEndpoint const& _endpoint)
...
+    if (m_peers.count(_n))
```
chfast: Optionally, use `.find()`.

libp2p/Host.cpp
chfast: This looks like a copy of the table.

libp2p/Host.cpp
```diff
@@ -919,76 +910,50 @@ void Host::restoreNetwork(bytesConstRef _b)
...
         BOOST_THROW_EXCEPTION(NetworkStartRequired());
 
-    if (m_dropPeers)
-        return;
-    
     RecursiveGuard l(x_sessions);
     RLP r(_b);
-    unsigned fileVersion = r[0].toInt<unsigned>();
-    if (r.itemCount() > 0 && r[0].isInt() && fileVersion >= dev::p2p::c_protocolVersion - 1)
+    unsigned protocolVersion = r[0].toInt<unsigned>();
```
chfast: clang-tidy suggest not to repeat the type if obvious from the expression. Use `auto`?

libp2p/Host.cpp
```diff
@@ -919,76 +910,50 @@ void Host::restoreNetwork(bytesConstRef _b)
...
+    if (r.itemCount() > 0 && r[0].isInt() && protocolVersion >= dev::p2p::c_protocolVersion)
     {
         // r[0] = version
         // r[1] = key
         // r[2] = nodes
 
-        for (auto i: r[2])
+        for (auto nodeRLP : r[2])
```
chfast: Why handled by copy? The `auto&&` always works correctly.

libp2p/Host.cpp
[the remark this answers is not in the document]
chfast: Perfect

libp2p/Host.cpp
```diff
@@ -1020,13 +985,24 @@ Node Host::nodeFromNodeTable(Public const& _id) const
...
+}
+
+bool Host::addKnownNodeToNodeTable(
+    Node const& _node, uint32_t _lastPongReceivedTime, uint32_t _lastPongSentTime)
+{
+    auto nt = nodeTable();
+    if (!nt)
+        return false;
+
+    nt->addKnownNode(_node, _lastPongReceivedTime, _lastPongSentTime);
```
```suggestion
    return nt->addKnownNode(_node, _lastPongReceivedTime, _lastPongSentTime);
```

libp2p/Host.cpp
```diff
@@ -1020,13 +985,24 @@ Node Host::nodeFromNodeTable(Public const& _id) const
...
 
-bool Host::addNodeToNodeTable(Node const& _node, NodeTable::NodeRelation _relation /* = NodeTable::NodeRelation::Unknown */)
+bool Host::addNodeToNodeTable(Node const& _node)
 {
     auto nodeTable = this->nodeTable();
     if (!nodeTable)
         return false;
 
-    nodeTable->addNode(_node, _relation);
+    nodeTable->addNode(_node);
```
```suggestion
    return nodeTable->addNode(_node);
```

libp2p/Host.cpp
gumb0: It was copy of a `shared_ptr`, but I'll change it to access member variable directly

gumb0:
@chfast suggestions addressed

libp2p/Host.cpp
```diff
@@ -547,46 +547,42 @@ void Host::requirePeer(NodeID const& _n, NodeIPEndpoint const& _endpoint)
...
     {
-        if (!addNodeToNodeTable(node))
-            return;
-        scheduleExecution(600, [this, _n]() {
-            if (auto n = nodeFromNodeTable(_n))
-                requirePeer(n.id, n.endpoint);
-        });
+        if (contains(m_peers, _n))
+        {
+            p = m_peers[_n];
```
chfast: Sorry for not clearly describing the issue. If `_n` is found the iterator can be used here.

gumb0: ah right

gumb0: fixed

gumb0:
Ah looks like I accidentally committed jsontests, sorry

chfast:
> Ah looks like I accidentally committed jsontests, sorry

Uff...

PR #5451, opened 2019-01-24
Upgrade E0VMC to 6.1.0, use C++ wrappers available

### LINKED ISSUE #5227: EVMC: cleanup wrapper code
Merge
`EVM::Result EVM::execute(ExtVMFace& _ext, int64_t gas)`
and
`owning_bytes_ref EVMC::exec(u256& io_gas, ExtVMFace& _ext, const OnOpFunc& _onOp)`

This abstraction seems not to bring any benefits since the assertions and wrappers are present in both, more cohesiveness would make it easier to read the code.

One possible abstraction is to have `EVM::execute` take `evmc_message` as an input, but not sure it is worth it.

chfast:
Closes #5227.

### DIFF
--- a/libevm/EVMC.cpp
+++ b/libevm/EVMC.cpp
@@ -10,15 +10,33 @@ namespace dev
 {
 namespace eth
 {
-EVM::EVM(evmc_instance* _instance) noexcept : m_instance(_instance)
+namespace
 {
-    assert(m_instance != nullptr);
-    assert(evmc_is_abi_compatible(m_instance));
+evmc_revision toRevision(EVMSchedule const& _schedule) noexcept
+{
+    if (_schedule.haveCreate2)
+        return EVMC_CONSTANTINOPLE;
+    if (_schedule.haveRevert)
+        return EVMC_BYZANTIUM;
+    if (_schedule.eip158Mode)
+        return EVMC_SPURIOUS_DRAGON;
+    if (_schedule.eip150Mode)
+        return EVMC_TANGERINE_WHISTLE;
+    if (_schedule.haveDelegateCall)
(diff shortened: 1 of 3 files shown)


libevm/EVMC.cpp
```diff
@@ -64,19 +68,28 @@ owning_bytes_ref EVMC::exec(u256& io_gas, ExtVMFace& _ext, const OnOpFunc& _onOp
...
+    auto output = bytesConstRef{r.output_data, r.output_size};
+
+    switch (r.status_code)
     {
     case EVMC_SUCCESS:
-        io_gas = r.gasLeft();
+        io_gas = r.gas_left;
         // FIXME: Copy the output for now, but copyless version possible.
-        return {r.output().toVector(), 0, r.output().size()};
+        return {output.toVector(), 0, output.size()};
```
axic: Why keep `bytesConstRef` here if this makes a copy anyway? Wouldn't it be simpler creating the vector here from `r.output_data` directly?

libevm/EVMC.cpp
```diff
@@ -10,15 +10,33 @@ namespace dev
...
 {
-EVM::EVM(evmc_instance* _instance) noexcept : m_instance(_instance)
+namespace
 {
-    assert(m_instance != nullptr);
-    assert(evmc_is_abi_compatible(m_instance));
+evmc_revision toRevision(EVMSchedule const& _schedule) noexcept
+{
+    if (_schedule.haveCreate2)
+        return EVMC_CONSTANTINOPLE;
```
axic: Is there a "Constantinople2" hardfork in aleth yet?

chfast: Not yet.

PR #5452, opened 2019-01-24
Handle receiving a Pong with changed NodeID

gumb0:
Fixes https://github.com/ethereum/aleth/issues/5431

### DIFF
--- a/libp2p/NodeTable.h
+++ b/libp2p/NodeTable.h
@@ -177,24 +177,30 @@ public:
 // protected only for derived classes in tests
 protected:
     /**
-     * NodeValidation is used to record Pinged node's ID, the timepoint of sent PING,
-     * time of sending and the new node ID to replace unresponsive node.
+     * NodeValidation is used to record information about the nodes that we have sent Ping to.
      */
     struct NodeValidation
     {
+        // Public key of the target node
         NodeID nodeID;
+        // We receive TCP port in the Ping packet, and store it here until it passes endpoint proof
+        // (answers with Pong), then it will be added to the bucket of the node table
         uint16_t tcpPort = 0;
+        // Time we sent Ping - used to handle timeouts
         TimePoint pingSendTime;
+        // Hash of the sent Ping packet - used to validate received Pong
         h256 pingHash;
+        // Replacement is put into the node table,
+        // if original pinged node doesn't answer after timeout
         std::shared_ptr<NodeEntry> replacementNodeEntry;
 
-        NodeValidation(NodeID const& _nodeID, uint16_t _tcpPort, TimePoint _pingSendTime,
+        NodeValidation(NodeID const& _nodeID, uint16_t _tcpPort, TimePoint const& _pingSendTime,
             h256 const& _pingHash, std::shared_ptr<NodeEntry> _replacementNodeEntry)
-          : nodeID(_nodeID),
-            tcpPort(_tcpPort),
-            pingSendTime(_pingSendTime),
-            pingHash(_pingHash),
-            replacementNodeEntry(std::move(_replacementNodeEntry))
+          : nodeID{_nodeID},
+            tcpPort{_tcpPort},
+            pingSendTime{_pingSendTime},
+            pingHash{_pingHash},
+            replacementNodeEntry{std::move(_replacementNodeEntry)}
         {}
     };
 
--- a/libp2p/NodeTable.h
+++ b/libp2p/NodeTable.h
@@ -187,18 +187,18 @@ protected:
         // (answers with Pong), then it will be added to the bucket of the node table
         uint16_t tcpPort = 0;
         // Time we sent Ping - used to handle timeouts
-        TimePoint pingSendTime;
+        TimePoint pingSentTime;
         // Hash of the sent Ping packet - used to validate received Pong
         h256 pingHash;
         // Replacement is put into the node table,
         // if original pinged node doesn't answer after timeout
         std::shared_ptr<NodeEntry> replacementNodeEntry;
 
-        NodeValidation(NodeID const& _nodeID, uint16_t _tcpPort, TimePoint const& _pingSendTime,
+        NodeValidation(NodeID const& _nodeID, uint16_t _tcpPort, TimePoint const& _pingSentTime,
             h256 const& _pingHash, std::shared_ptr<NodeEntry> _replacementNodeEntry)
           : nodeID{_nodeID},
             tcpPort{_tcpPort},
-            pingSendTime{_pingSendTime},
+            pingSentTime{_pingSentTime},
             pingHash{_pingHash},
             replacementNodeEntry{std::move(_replacementNodeEntry)}
         {}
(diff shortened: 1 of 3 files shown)


libp2p/NodeTable.h
```diff
@@ -297,7 +307,7 @@ class NodeTable : UDPSocketEvents
     std::shared_ptr<NodeSocket> m_socket;							///< Shared pointer for our UDPSocket; ASIO requires shared_ptr.
 
     // The info about PING packets we've sent to other nodes and haven't received PONG yet
-    std::unordered_map<NodeID, NodeValidation> m_sentPings;
+    std::map<bi::udp::endpoint, NodeValidation> m_sentPings;
```
gumb0: Switched to `map` here, because it was easier than creating a hash function for `udp::endpoint`

gumb0:
`findNodeIsSentAfterPong` test fails intermittently, hopefully that will be fixed by @halfalicious PR https://github.com/ethereum/aleth/pull/5502/

Windows build timed out

gumb0:
Rebased, please review @chfast @halfalicious

test/unittests/libp2p/net.cpp
```diff
@@ -675,6 +675,54 @@ BOOST_AUTO_TEST_CASE(validPong)
...
+    uint16_t nodePort = nodeSocketHost.port;
+
+    // add a node to node table, initiating PING
+    auto nodeEndpoint = NodeIPEndpoint{bi::address::from_string(c_localhostIp), nodePort, nodePort};
+    auto nodeKeyPair = KeyPair::create();
+    auto nodePubKey = nodeKeyPair.pub();
+    nodeTable->addNode(Node{nodePubKey, nodeEndpoint});
+
+    // handle received PING
+    auto pingDataReceived = nodeSocketHost.packetsReceived.pop();
```
halfalicious: (Nit) Perhaps validate the packet name via BOOST_REQUIRE before doing the conversion? I think this would result in a more descriptive test failure message

test/unittests/libp2p/net.cpp
```diff
@@ -675,6 +675,54 @@ BOOST_AUTO_TEST_CASE(validPong)
...
+    // send PONG with different NodeID
+    Pong pong(nodeTable->m_hostNodeEndpoint);
+    pong.echo = ping.echo;
+    auto newNodeKeyPair = KeyPair::create();
+    auto newNodePubKey = newNodeKeyPair.pub();
+    pong.sign(newNodeKeyPair.secret());
+    nodeSocketHost.socket->send(pong);
+
+    // wait for PONG to be received and handled
+    nodeTable->packetsReceived.pop();
```
halfalicious: (Nit) Should verify that the received packet is a pong packet

test/unittests/libp2p/net.cpp
```diff
@@ -675,6 +675,54 @@ BOOST_AUTO_TEST_CASE(validPong)
...
+
+    BOOST_REQUIRE(nodeTable->nodeExists(newNodeKeyPair.pub()));
+    auto addedNode = nodeTable->nodeEntry(newNodePubKey);
+    BOOST_CHECK(addedNode->lastPongReceivedTime > 0);
```
halfalicious: Shouldn't this (and the rest of the BOOST_CHECKs in this test case) be BOOST_REQUIRE?

test/unittests/libp2p/net.cpp
```diff
@@ -675,6 +675,54 @@ BOOST_AUTO_TEST_CASE(validPong)
...
+
+    // wait for old node to be kicked out after timeout
+    this_thread::sleep_for(std::chrono::seconds(6));
```
halfalicious: I don't think you need to wait here - you've already verified that a node with the new pubkey is in the node table, and we drop the node with the old pubkey (in onPacketReceived) before adding the new node (in noteActiveNode)

libp2p/NodeTable.h
```diff
@@ -177,15 +177,25 @@ class NodeTable : UDPSocketEvents
...
     /**
-     * NodeValidation is used to record Pinged node's endpoint, the timepoint of sent PING,
+     * NodeValidation is used to record Pinged node's ID, the timepoint of sent PING,
      * time of sending and the new node ID to replace unresponsive node.
      */
     struct NodeValidation
     {
-        NodeIPEndpoint endpoint;
+        NodeID nodeID;
+        uint16_t tcpPort = 0;
```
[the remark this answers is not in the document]
gumb0: Yeah good idea, I'll add a comment.
(actually we store udp port in the `m_sentPings`, too - in the key part which is `udp::endpoint`)

test/unittests/libp2p/net.cpp
```diff
@@ -675,6 +675,54 @@ BOOST_AUTO_TEST_CASE(validPong)
...
+    // NodeTable sending PING
+    TestNodeTableHost nodeTableHost(0);
+    nodeTableHost.start();
+    auto& nodeTable = nodeTableHost.nodeTable;
+    nodeTable->setRequestTimeToLive(std::chrono::seconds(1));
+
+    // socket receiving PING
+    TestUDPSocketHost nodeSocketHost{randomPortNumber()};
+    nodeSocketHost.start();
+    uint16_t nodePort = nodeSocketHost.port;
```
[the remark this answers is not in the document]
gumb0: Yeah I don't pay that much attention to const correctness in tests

test/unittests/libp2p/net.cpp
gumb0: The reasoning is that updating the node in the node table, updating `m_sentPings` and removing the old node are kind of independent pieces of code, and if one fails, we can still get some useful info about whether other ones work

test/unittests/libp2p/net.cpp
gumb0: Ah right, I've written this before implementation, and I had idea then that it will be dropped after timeout,  but in the end it's removed immediately in the Pong handler

gumb0:
@halfalicious Thank you for the review, all comments addressed

test/unittests/libp2p/net.cpp
halfalicious: > 
> 
> The reasoning is that updating the node in the node table, updating `m_sentPings` and removing the old node are kind of independent pieces of code, and if one fails, we can still get some useful info about whether other ones work

Yes, but I think the checks will only provide value when you do a local run (let me know if I'm wrong here)? The problem is that failed boost_checks don't trigger test failures and boost_require triggers a test failure and ceases test execution and the boost_checks are all after a boost_require, meaning that if the boost_require fails the subsequent boost_checks won't be executed, and if the boost_require passes the boost_checks might fail but won't trigger a test failure (so we won't know to look at the logs)? Also, do boost_check failures show up in CircleCI logs even if the tests themselves passed?

halfalicious:
Approved, remaining issues aren't blockers

test/unittests/libp2p/net.cpp
gumb0: I believe BOOST_CHECK failures always show up in logs.

PR #5454, opened 2019-01-25
Streamline host shutdown logic and fix use-after-free

halfalicious:
Prior to these changes, host shutdown would involve the app thread setting m_run to false and waiting on a condition variable until the timer used to schedule the host's main work loop (Host::run) was destroyed on the network thread. Note that as a part of shutting down the network thread, the node table was destroyed and the boost io service was stopped (in that order). Because the timer was accessed on two threads, it was protected by a mutex. Additionally, because the node table was destroyed before the io service and the node table scheduled discovery round execution via boost deadline timers which used the node table's captured this pointer, the node table could be destroyed while there was an outstanding scheduled execution which would cause a use-after-free.

I've simplified host shutdown by moving io service stoppage to the app thread (Host::stop) and letting the node table and timer be destroyed automatically when the Host instance is destroyed. Since the timer can now only be accessed on the network thread, I've removed the associated mutex. Since the app thread no longer needs to wait while the network thread shuts down, we no longer need the timer condition variable either (m_timerReset).

This should also address the use-after-free since the io service is now stopped before node table is destroyed.

### DIFF
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -132,11 +132,9 @@ void Host::stop()
     if (!m_run.exchange(false))
         return;
 
-    {
-        unique_lock<mutex> l(x_runTimer);
-        while (m_timer)
-            m_timerReset.wait(l);
-    }
+    // stopping io service allows running manual network operations for shutdown
+    // and also stops blocking worker thread, allowing worker thread to exit
+    m_ioService.stop();
 
     // stop worker thread
     if (isWorking())
@@ -665,25 +663,10 @@ size_t Host::peerCount() const
     return retCount;
 }
 
-void Host::run(boost::system::error_code const&)
+void Host::run(boost::system::error_code const& _ec)
 {
-    if (!m_run)
-    {
-        // reset NodeTable
-        DEV_GUARDED(x_nodeTable)
-            m_nodeTable.reset();
-
-        // stopping io service allows running manual network operations for shutdown
-        // and also stops blocking worker thread, allowing worker thread to exit
-        m_ioService.stop();
-
-        // resetting timer signals network that nothing else can be scheduled to run
-        DEV_GUARDED(x_runTimer)
-            m_timer.reset();
-
-        m_timerReset.notify_all();
+    if (!m_run || _ec)
         return;
-    }
 
     if (auto nodeTable = this->nodeTable()) // This again requires x_nodeTable, which is why an additional variable nodeTable is used.
         nodeTable->processEvents();
@@ -739,9 +722,12 @@ void Host::run(boost::system::error_code const&)
         }
     }
 
-    auto runcb = [this](boost::system::error_code const& error) { run(error); };
-    m_timer->expires_from_now(boost::posix_time::milliseconds(c_timerInterval));
-    m_timer->async_wait(runcb);
+    if (m_run)
+    {
+        auto runcb = [this](boost::system::error_code const& error) { run(error); };
+        m_timer->expires_from_now(boost::posix_time::milliseconds(c_timerInterval));
+        m_timer->async_wait(runcb);
+    }
 }
 
 void Host::startedWorking()
@@ -750,15 +736,9 @@ void Host::startedWorking()
     // initialization (e.g. start capability threads, start TCP listener, and kick off timers)
     asserts(!m_timer);
 
-    {
-        // prevent m_run from being set to true at same time as set to false by stop()
-        // don't release mutex until m_timer is set so in case stop() is called at same
-        // time, stop will wait on m_timer and graceful network shutdown.
-        Guard l(x_runTimer);
-        // create deadline timer
-        m_timer.reset(new io::deadline_timer(m_ioService));
-        m_run = true;
-    }
+    // create deadline timer
+    m_timer.reset(new io::deadline_timer(m_ioService));
+    m_run = true;
 
     // start capability threads (ready for incoming connections)
     for (auto const& h: m_capabilities)
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -198,12 +198,6 @@ void Host::doneWorking()
         m_ioService.poll();
     }
 
-    // stop network (again; helpful to call before subsequent reset())
-    m_ioService.stop();
-
-    // reset network (allows reusing ioservice in future)
-    m_ioService.reset();
-
     // finally, clear out peers (in case they're lingering)
     RecursiveGuard l(x_sessions);
     m_sessions.clear();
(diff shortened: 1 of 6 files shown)


halfalicious:
Note that these changes assume that polling the io service after resetting it (as is done in Host::doneWorking) will not start the NodeTable timers running again.

I'm also planning on:
* Changing Host::stop() to a private function so one can't try to start/stop the network (since we no longer support that)  and making a call to it at the beginning of Host::saveNetwork().
* Changing m_timer to a non-heap-allocated object (as @chfast suggested a while back)

halfalicious:
@gumb0 / @chfast : Still a WIP but I'd appreciate your eyes on this since I'm still unsure about all of the ways that this can impact the inner workings of the host / node table.

libp2p/Host.cpp
```diff
@@ -750,15 +736,9 @@ void Host::startedWorking()
...
-        // don't release mutex until m_timer is set so in case stop() is called at same
-        // time, stop will wait on m_timer and graceful network shutdown.
-        Guard l(x_runTimer);
-        // create deadline timer
-        m_timer.reset(new io::deadline_timer(m_ioService));
-        m_run = true;
-    }
+    // create deadline timer
+    m_timer.reset(new io::deadline_timer(m_ioService));
+    m_run = true;
```
gumb0: I would move this to right before run `run(boost::system::error_code());` because it's closely related

gumb0:
> Changing Host::stop() to a private function so one can't try to start/stop the network (since we no longer support that) and making a call to it at the beginning of Host::saveNetwork().

Not sure about this, I would prefer `saveNetwork` to just fail in case network isn't stopped if we need this (it seems weird for function called `saveNetwork` to have such side effects)
Anyway is the network currently stopped before `saveNetwork`?

gumb0:
> Note that these changes assume that polling the io service after resetting it (as is done in Host::doneWorking) will not start the NodeTable timers running again.

I think it can start them (call the handlers scheduled previously)

Currently `NodeTable` prevents that by calling `m_timers.stop()` from its destructor. It seems we need now to introduce `NodeTable::stop` method that will do this and call it together with `m_ioService.stop`

libp2p/Host.cpp
halfalicious: That makes sense to me too...also I think it makes sense since that's after the network is considered to be fully up i.e. the node table has been started, the tcp and udp listeners are up, capability threads are running, etc.

halfalicious:
> 
> 
> > Changing Host::stop() to a private function so one can't try to start/stop the network (since we no longer support that) and making a call to it at the beginning of Host::saveNetwork().
> 
> Not sure about this, I would prefer `saveNetwork` to just fail in case network isn't stopped if we need this (it seems weird for function called `saveNetwork` to have such side effects)
> Anyway is the network currently stopped before `saveNetwork`?

It's up to the caller of saveNetwork, they need to call stop first.

libp2p/Host.cpp
halfalicious: Done, but in order for this to work I had to remove the dependency on m_run in `Host::runAcceptor`.

libp2p/Host.cpp
```diff
@@ -127,16 +134,18 @@ void Host::stop()
...
-        while (m_timer)
-            m_timerReset.wait(l);
-    }
+    // stopping io service allows running manual network operations for shutdown
+    // and also stops blocking worker thread, allowing worker thread to exit
+    m_ioService.stop();
+
+    // Close the node table socket and cancel deadline timers. This effectively stops
+    // discovery, even during subsequent io service polling
+    nodeTable()->stop();
```
halfalicious: I don't think I need to do this - we can let the node table continue to run since it's not deleted until the Host instance is deleted, meaning there's no risk of a UAF. @gumb0 : Thoughts?

halfalicious:
cc @chfast / @gumb0 : This is ready for review now

libp2p/Host.cpp
```diff
@@ -739,27 +725,18 @@ void Host::run(boost::system::error_code const&)
         }
     }
 
+    if (!m_run)
+        return;
+
     auto runcb = [this](boost::system::error_code const& error) { run(error); };
-    m_timer->expires_from_now(boost::posix_time::milliseconds(c_timerInterval));
-    m_timer->async_wait(runcb);
+    m_timer.expires_from_now(boost::posix_time::milliseconds(c_timerInterval));
```
chfast: Just a question: Can you use `std::chrono` instead of `boost` for milliseconds?

libp2p/Host.cpp
```diff
@@ -109,6 +110,10 @@ Host::~Host()
 void Host::start()
 {
     DEV_TIMED_FUNCTION_ABOVE(500);
+    auto nodeTable = this->nodeTable();
+    if (nodeTable)
+        BOOST_THROW_EXCEPTION(NetworkRestartNotSupported());
```
chfast: Can you explain this part? This requires a mutex lock and a shared_ptr copy.

halfalicious: Ah I don’t suppose I need to take the lock since I’m just checking if the handle is valid and don’t care about the state of the buckets or node list. I’ll change this before merging.

libp2p/Host.cpp
halfalicious: Unfortunately no, expires_from_now requires boost time.

halfalicious:
Addressed feedback, added unit test for network restart, and squashed some changes

halfalicious:
> 
> 
> It's hard to asses the quality of changes from my perspective.
> I tested it a bit with
> 
> ```
> aleth-bootnode/aleth-bootnode --log-verbosity 3 & sleep 0.1 && kill -2 $!
> ```

Here's a PowerShell script which does something similar on Windows:
```

 param (
    [Parameter(Mandatory=$false)][string]$iter = 5
 )

for ($i = 1; $i -le $iter; $i++) {
  Write-Host "Attempt" $i "of" $iter
  start-process -FilePath ".\Aleth-bootnode.exe" -ArgumentList "-v 4"
  $sleepTime = Get-Random -Maximum 30
  Write-Host "Sleeping for" $sleepTime "seconds"
  start-sleep -S $sleepTime
  stop-process -Name "aleth-bootnode"
}
```

libp2p/Host.cpp
```diff
@@ -839,6 +817,12 @@ void Host::disconnectLatePeers()
 
 bytes Host::saveNetwork() const
 {
+    if (haveNetwork())
+    {
+        cwarn << "Cannot save network configuration since network is still running.";
+        return RLPStream{}.out();
```
gumb0: Just `bytes{}` would be simpler; looking at this I thought it might return empty RLP-list or something

libp2p/Host.cpp
gumb0: Yeah, that's what I suggested, because I think it might be a good idea to tell it to not run the handlers' code even if they are still called

gumb0:
I don't get failures, let's merge it

PR #5457, opened 2019-01-29
Fix intermittently failing udpOnce test

gumb0:
This test was failing sometimes, I guess when the port is occupied.


gumb0:
@chfast it sends the datagram not to the port created in the test, but to the port selected by `TestUDPSocketHost` constructor (which takes care to find the free one)

PR #5463, opened 2019-02-01
Don't start TCP listener without registered capabilities

halfalicious:
Don't start the TCP listener when no capabilities are registered. The purpose of this work is to prevent aleth-bootnode from starting a TCP listener since it only runs discovery (which uses UDP) and doesn't support peer connections. These changes include a new unit test which validates this behavior.

This branch is currently based on my branch for the host shutdown PR - once that's merged I'll rebase. 

Note that these changes add the requirement that capabilities must be registered before the host is started - I think this makes sense since these changes do a capability check to determine if it needs to create a TCP listener.

### DIFF
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -426,8 +426,7 @@ void Host::runAcceptor()
 
     if (m_tcp4Acceptor.is_open() && !m_accepting)
     {
-        cnetdetails << "Listening on local port " << m_listenPort << " (public: " << m_tcpPublic
-                    << ")";
+        cnetdetails << "Listening on local port " << m_listenPort;
         m_accepting = true;
 
         auto socket = make_shared<RLPXSocket>(m_ioService);
@@ -771,18 +770,15 @@ void Host::startedWorking()
         if (port > 0)
         {
             m_listenPort = port;
-            determinePublic();
             runAcceptor();
         }
         else
             LOG(m_logger) << "p2p.start.notice id: " << id() << " TCP Listen port is invalid or unavailable.";
     }
     else
-    {
         m_listenPort = m_netConfig.listenPort;
-        determinePublic();
-    }
 
+    determinePublic();
     auto nodeTable = make_shared<NodeTable>(
         m_ioService,
         m_alias,
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -151,6 +151,11 @@ void Host::stop()
 
 void Host::doneWorking()
 {
+    // Return early if we have no capabilities since there's nothing to do. We've already stopped
+    // the io service and cleared the node table timers which means discovery is no longer running.
+    if (!haveCapabilities())
+        return;
+
     // reset ioservice (cancels all timers and allows manually polling network, below)
     m_ioService.reset();
 
@@ -494,6 +499,11 @@ void Host::registerCapability(std::shared_ptr<CapabilityFace> const& _cap)
 void Host::registerCapability(
     std::shared_ptr<CapabilityFace> const& _cap, std::string const& _name, u256 const& _version)
 {
+    if (haveNetwork())
+    {
+        cwarn << "Capabilities must be registered before the network is started";
+        return;
+    }
     m_capabilities[std::make_pair(_name, _version)] = _cap;
 }
 
@@ -529,6 +539,12 @@ void Host::requirePeer(NodeID const& _n, NodeIPEndpoint const& _endpoint)
     {
         cwarn << "Network not running so node (" << _n << ") with endpoint (" << _endpoint
               << ") cannot be added as a required peer";
+        return;
+    }
+    if (!haveCapabilities())
+    {
+        cwarn << "No capabilities registered so node (" << _n << ") with endpoint (" << _endpoint
+            << ") cannot be added as a required peer";
         return;
     }
 
@@ -590,7 +606,15 @@ void Host::relinquishPeer(NodeID const& _node)
 void Host::connect(std::shared_ptr<Peer> const& _p)
 {
     if (!m_run)
+    {
+        cwarn << "Network not running so cannot connect to peer " << _p->id << "@" << _p->address();
         return;
+    }
+    if (!haveCapabilities())
+    {
+        cwarn << "No capabilities registered so cannot connect to peer " << _p->id << "@" << _p->address();
+        return;
+    }
     
     if (havePeerSession(_p->id))
     {
@@ -740,16 +764,24 @@ void Host::startedWorking()
     for (auto const& h: m_capabilities)
         h.second->onStarting();
     
-    // try to open acceptor (todo: ipv6)
-    int port = Network::tcp4Listen(m_tcp4Acceptor, m_netConfig);
-    if (port > 0)
+    if (haveCapabilities())
     {
-        m_listenPort = port;
-        determinePublic();
-        runAcceptor();
+        // try to open acceptor (todo: ipv6)
+        int port = Network::tcp4Listen(m_tcp4Acceptor, m_netConfig);
+        if (port > 0)
+        {
+            m_listenPort = port;
+            determinePublic();
+            runAcceptor();
+        }
+        else
+            LOG(m_logger) << "p2p.start.notice id: " << id() << " TCP Listen port is invalid or unavailable.";
     }
     else
-        LOG(m_logger) << "p2p.start.notice id: " << id() << " TCP Listen port is invalid or unavailable.";
+    {
+        m_listenPort = m_netConfig.listenPort;
+        determinePublic();
+    }
 
     auto nodeTable = make_shared<NodeTable>(
         m_ioService,
@@ -763,11 +795,14 @@ void Host::startedWorking()
         m_nodeTable = nodeTable;
     restoreNetwork(&m_restoreNetwork);
 
-    LOG(m_logger) << "p2p.started id: " << id();
-
     m_run = true;
-
-    run(boost::system::error_code());
+    if (haveCapabilities())
+    {
+        LOG(m_logger) << "devp2p started. Node id: " << id();
+        run(boost::system::error_code());
+    }
+    else
+        LOG(m_logger) << "No registered capabilities, devp2p not started.";
 }
 
 void Host::doWork()
--- a/test/tools/libtestutils/Common.cpp
+++ b/test/tools/libtestutils/Common.cpp
@@ -61,7 +61,9 @@ int dev::test::randomNumber(int _min, int _max)
 
 unsigned short dev::test::randomPortNumber(unsigned short _min, unsigned short _max)
 {
-    return static_cast<unsigned short>(randomNumber(_min, _max));
+    static std::mt19937 randomGenerator(utcTime());
+    randomGenerator.seed(std::random_device()());
+    return std::uniform_int_distribution<unsigned short>(_min, _max)(randomGenerator);
 }
 
 Json::Value dev::test::loadJsonFromFile(fs::path const& _path)
--- a/test/tools/libtestutils/Common.cpp
+++ b/test/tools/libtestutils/Common.cpp
@@ -52,11 +52,16 @@ boost::filesystem::path dev::test::getTestPath()
     return boost::filesystem::path(testPath);
 }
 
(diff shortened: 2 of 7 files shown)


halfalicious:
Following tests are failing:
```
The following tests FAILED:
	480 - eip8/test_handshake_plain_auth (Failed)
	481 - eip8/test_handshake_eip8_auth1 (Failed)
	482 - eip8/test_handshake_eip8_auth2 (Failed)
	483 - eip8/test_handshake_eip8_ack_plain (Failed)
	484 - eip8/test_handshake_eip8_ack1 (Failed)
	485 - eip8/test_handshake_eip8_ack2 (Failed)
	511 - libp2p/p2p/host (Failed)
	512 - libp2p/p2p/networkConfig (Failed)
	513 - libp2p/p2p/saveNodes (Timeout)
	514 - libp2p/p2pPeer/requirePeer (Failed)
```

halfalicious:
> 
> 
> Following tests are failing:
> 
> ```
> The following tests FAILED:
> 	480 - eip8/test_handshake_plain_auth (Failed)
> 	481 - eip8/test_handshake_eip8_auth1 (Failed)
> 	482 - eip8/test_handshake_eip8_auth2 (Failed)
> 	483 - eip8/test_handshake_eip8_ack_plain (Failed)
> 	484 - eip8/test_handshake_eip8_ack1 (Failed)
> 	485 - eip8/test_handshake_eip8_ack2 (Failed)
> 	511 - libp2p/p2p/host (Failed)
> 	512 - libp2p/p2p/networkConfig (Failed)
> 	513 - libp2p/p2p/saveNodes (Timeout)
> 	514 - libp2p/p2pPeer/requirePeer (Failed)
> ```

The EIP8 tests hit a read AV on Windows, will debug on Linux

halfalicious:
The following test is failing in circleci (linux-gcc6): 
```

Running 1 test case...
Test Case "pingNotSentAfterPongForKnownNode": 
/home/builder/project/libdevcore/concurrent_queue.h(47): fatal error: in "_T dev::concurrent_queue<_T, _QueueT>::pop(const milliseconds&) [with _T = std::vector<unsigned char>; _QueueT = std::queue<std::vector<unsigned char>, std::deque<std::vector<unsigned char>, std::allocator<std::vector<unsigned char> > > >; std::chrono::milliseconds = std::chrono::duration<long int, std::ratio<1l, 1000l> >]": boost::exception_detail::clone_impl<dev::WaitTimeout>: /home/builder/project/libdevcore/concurrent_queue.h(47): Throw in function _T dev::concurrent_queue<_T, _QueueT>::pop(const milliseconds&) [with _T = std::vector<unsigned char>; _QueueT = std::queue<std::vector<unsigned char>, std::deque<std::vector<unsigned char>, std::allocator<std::vector<unsigned char> > > >; std::chrono::milliseconds = std::chrono::duration<long int, std::ratio<1l, 1000l> >]
Dynamic exception type: boost::exception_detail::clone_impl<dev::WaitTimeout>

/home/builder/project/test/unittests/libp2p/net.cpp(1063): last checkpoint

561/561 Test #210: GeneralStateTests/stSStoreTest ........................................................................................   Passed  290.24 sec

99% tests passed, 1 tests failed out of 561

Total Test time (real) = 337.34 sec

The following tests FAILED:
	508 - network/net/pingNotSentAfterPongForKnownNode (Failed)
Errors while running CTest
Exited with code 8
```

halfalicious:
I need to investigate the failing tests but I don't think we need to block starting the review on that.

libp2p/Host.cpp
```diff
@@ -488,6 +497,11 @@ std::unordered_map<Public, std::string> Host::pocHosts()
 
 void Host::registerCapability(std::shared_ptr<CapabilityFace> const& _cap)
 {
+    if (haveNetwork())
```
chfast: I believe this check should be in the other `registerCapability()` method which is called from this one.

libp2p/Host.cpp
```diff
@@ -665,25 +688,10 @@ size_t Host::peerCount() const
...
-        // stopping io service allows running manual network operations for shutdown
-        // and also stops blocking worker thread, allowing worker thread to exit
-        m_ioService.stop();
-
-        // resetting timer signals network that nothing else can be scheduled to run
-        DEV_GUARDED(x_runTimer)
-            m_timer.reset();
-
-        m_timerReset.notify_all();
+    if (!m_run || _ec)
```
chfast: In the matter of convention, maybe always check the error first?

libp2p/Host.cpp
```diff
@@ -109,6 +110,10 @@ Host::~Host()
 void Host::start()
 {
     DEV_TIMED_FUNCTION_ABOVE(500);
+    auto nodeTable = this->nodeTable();
```
chfast: Have you revisited this check? The `->nodeTable()` does a lock and returns a shared pointer.

halfalicious: yes I fixed it in the other PR so it will be fixed here too when I rebase after merging the other one

libp2p/Host.cpp
```diff
@@ -127,16 +132,21 @@ void Host::stop()
...
 
-    // ignore if already stopped/stopping, at the same time,
-    // signal run() to prepare for shutdown and reset m_timer
-    if (!m_run.exchange(false))
+    // ignore if already stopped/stopping, at the same time
+    // indicates that the network is shutting down
+    if (!haveNetwork())
         return;
+    
+    if (haveCapabilities())
```
gumb0: Do we need this check? Can't we make `m_run` have sensible values in both modes?

libp2p/Host.cpp
```diff
@@ -494,6 +503,12 @@ void Host::registerCapability(std::shared_ptr<CapabilityFace> const& _cap)
 void Host::registerCapability(
     std::shared_ptr<CapabilityFace> const& _cap, std::string const& _name, u256 const& _version)
 {
+    if (haveNetwork())
+    {
+        cwarn << "Capabilities must be registered before the network is started";
```
gumb0: Why this new requirement appeared?

libp2p/Host.cpp
halfalicious: Good point - I've reworked my changes so that m_run is now always set to true if the network thread has been started, I've added a `haveCapabilities()` call where appropriate.

libp2p/Host.cpp
halfalicious: Because we determine whether or not we start the TCP listener based on whether or not there are registered capabilities, and we start the TCP listener in network thread startup (`host::startedWorking`), meaning capabilities need to be registered before network start. I suppose I could add functionality which detects when capabilities are registered after network start and starts the TCP listener, but this would make the changes more complicated and I didn't think it was necessary because I went through the RPC functions and didn't see any which allow someone to register capabilities.

libp2p/Host.cpp
```diff
@@ -740,16 +764,24 @@ void Host::startedWorking()
...
     {
-        m_listenPort = port;
-        determinePublic();
-        runAcceptor();
+        // try to open acceptor (todo: ipv6)
+        int port = Network::tcp4Listen(m_tcp4Acceptor, m_netConfig);
+        if (port > 0)
+        {
+            m_listenPort = port;
+            determinePublic();
```
gumb0: Why do we do this only if acceptor successfully started? `NodeTable` might need it anyway

gumb0: `m_tcpPublic` is used in `runAcceptor` only for logging, so I think we can move `determinePublic()` to just before `NodeTable` construction

test/tools/libtestutils/Common.cpp
```diff
@@ -52,11 +52,16 @@ boost::filesystem::path dev::test::getTestPath()
...
 {
     static std::mt19937 randomGenerator(utcTime());
     randomGenerator.seed(std::random_device()());
-    return std::uniform_int_distribution<int>(1)(randomGenerator);
+    return std::uniform_int_distribution<int>(_min, _max)(randomGenerator);
+}
+
+unsigned short dev::test::randomPortNumber(unsigned short _min, unsigned short _max)
+{
+    return static_cast<unsigned short>(randomNumber(_min, _max));
```
gumb0: well it's a bit weird to use `uniform_int_distribution<int>` and then to cast it to `unsigned short`, you could just use `uniform_int_distribution<unsigned short>` (would be correctly uniformly distributed, too)

test/unittests/libp2p/net.cpp
```diff
@@ -439,6 +432,46 @@ BOOST_AUTO_TEST_CASE(kademlia)
...
+	{
+		this_thread::sleep_for(chrono::milliseconds(step));
+		if (host.haveNetwork())
+			break;
+	}
+
+    BOOST_REQUIRE(host.haveNetwork());
+    auto const hostPort = host.listenPort();
+    BOOST_REQUIRE(hostPort);
+	BOOST_REQUIRE(host.caps().empty());
```
gumb0: some tabs here

halfalicious: Whoops, forgot to run git clang-format before comitting

libp2p/Host.cpp
halfalicious: Don't we still want to log the correct public IP information in runAcceptor? Moving determinePublic to just before the NodeTable ctor will break this. I'd rather keep the correct logging at the expense of some slightly duplicated code.

test/tools/libtestutils/Common.cpp
```diff
@@ -61,7 +61,9 @@ int dev::test::randomNumber(int _min, int _max)
 
 unsigned short dev::test::randomPortNumber(unsigned short _min, unsigned short _max)
 {
-    return static_cast<unsigned short>(randomNumber(_min, _max));
+    static std::mt19937 randomGenerator(utcTime());
```
chfast: I believe you should reuse existing RNG, the one that is used by randomNumber().

test/tools/libtestutils/Common.cpp
```diff
@@ -61,7 +61,9 @@ int dev::test::randomNumber(int _min, int _max)
...
+    randomGenerator.seed(std::random_device()());
+    return std::uniform_int_distribution<unsigned short>(_min, _max)(randomGenerator);
```
chfast: Use `{}` to construct the object.
```
return std::uniform_int_distribution<unsigned short>{_min, _max}(randomGenerator);
```

test/tools/libtestutils/Common.cpp
halfalicious: Got it, will make it a global object in an anonymous namespace. Out of curiosity, why do we seed on each call in addition to seeding when we create the rng?

chfast: After your changes it is seeded only once as it should be. General rule is to seed a PRNG with a good source of entropy so that you will not need to reach out to this source any more later on. Using timestamps for entropy is bad practice, `std::random_device` would be much better, but it can stay for unit testing.

When security is needed, you probably should use secure PRNG which messner twister is not.

test/tools/libtestutils/Common.cpp
```diff
@@ -52,11 +40,16 @@ boost::filesystem::path dev::test::getTestPath()
     return boost::filesystem::path(testPath);
 }
 
-int dev::test::randomNumber()
+int dev::test::randomNumber(int _min, int _max)
+{
+    g_randomGenerator.seed(random_device{}());
```
chfast: Oh, this is not needed. Just do 
```
mt19937 g_randomGenerator(random_device{}());
```
once.

libp2p/Host.cpp
```diff
@@ -763,11 +791,14 @@ void Host::startedWorking()
         m_nodeTable = nodeTable;
     restoreNetwork(&m_restoreNetwork);
```
gumb0: I just realized that `restoreNetwork` requires `m_true` to be `true`, too, let's fix it in another PR

test/tools/libtestutils/Common.cpp
halfalicious: Yea I was wondering why the old code seeded twice (once on creation and once per call)

libp2p/Host.cpp
halfalicious: Ah thanks for the clarification

PR #5464, opened 2019-02-01
cpp-build-env: Upgrade OS and compilers

chfast:
- Upgrade Debian to testing.
- Upgrade default GCC to 8.
- Install GCC 7.
- Upgrade clang to 8.


halfalicious:
What do these upgrades get us functionally? Or is this just following best practices of running a newer compiler toolchain?

chfast:
This docker image is uses by CI services (also by other C++ projects). There are also versioned, so not upgrade is done automatically. In this version I upgraded some compilers and cmake.

PR #5468, opened 2019-02-04
Better output for endpoint

chfast:
I wanted to create an `operator<<` for `Node`, but ended up with cleaning up the `p2p/Common.cpp` file.

The only functional change is in `operator<<` of the endpoint. Now it looks like
```
DEBUG 02-04 11:57:25 p2p  discov Ping to ##e2bb9f88…@101.80.168.235:30303:udp59556
```
and the `:udp...` part is added only when the port is different from the TCP one.

Some other code uses `?discport=` convention. Still ok to me. I noticed it after the change was done.

I will send the `operator<<` for `Node` in the next PR.

### DIFF
--- a/libp2p/Common.cpp
+++ b/libp2p/Common.cpp
@@ -12,8 +12,6 @@ namespace p2p
 {
 using namespace std;
 
-const unsigned c_protocolVersion = 4;
-
 const NodeIPEndpoint UnspecifiedNodeIPEndpoint = NodeIPEndpoint{{}, 0, 0};
 const Node UnspecifiedNode = Node{{}, UnspecifiedNodeIPEndpoint};
 
--- a/libp2p/Common.cpp
+++ b/libp2p/Common.cpp
@@ -83,8 +83,8 @@ bool isLocalHostAddress(bi::address const& _addressToCheck)
         {bi::address_v6::from_string("::1")},
         {bi::address_v6::from_string("::")}
     };
-    
-    return find(c_rejectAddresses.begin(), c_rejectAddresses.end(), _addressToCheck) != c_rejectAddresses.end();
+
+    return c_rejectAddresses.find(_addressToCheck) != c_rejectAddresses.end();
 }
 
 bool isLocalHostAddress(std::string const& _addressToCheck)
@@ -144,7 +144,7 @@ void DeadlineOps::reap()
         return;
 
     Guard l(x_timers);
-    std::vector<DeadlineOp>::iterator t = m_timers.begin();
+    auto t = m_timers.begin();
     while (t != m_timers.end())
         if (t->expired())
         {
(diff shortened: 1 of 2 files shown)


libp2p/Common.cpp
```diff
@@ -144,7 +144,7 @@ void DeadlineOps::reap()
         return;
 
     Guard l(x_timers);
-    std::vector<DeadlineOp>::iterator t = m_timers.begin();
+    auto t = m_timers.begin();
```
halfalicious: Can this be const?

libp2p/Common.cpp
```diff
@@ -13,7 +13,6 @@ namespace p2p
 using namespace std;
 
 const unsigned c_protocolVersion = 4;
```
halfalicious: Constexpr?

libp2p/Common.cpp
```diff
@@ -13,7 +13,6 @@ namespace p2p
...
-static_assert(c_protocolVersion == 4, "Replace v3 compatbility with v4 compatibility before updating network version.");
 
 const NodeIPEndpoint UnspecifiedNodeIPEndpoint = NodeIPEndpoint{{}, 0, 0};
```
halfalicious: Constexpr?

libp2p/Common.cpp
```diff
@@ -13,7 +13,6 @@ namespace p2p
...
 const Node UnspecifiedNode = Node{{}, UnspecifiedNodeIPEndpoint};
```
halfalicious: Constexpr?

libp2p/Common.cpp
```diff
@@ -80,35 +85,50 @@ bool isLocalHostAddress(bi::address const& _addressToCheck)
         {bi::address_v4::from_string(c_localhostIp)},
         {bi::address_v4::from_string("0.0.0.0")},
         {bi::address_v6::from_string("::1")},
-        {bi::address_v6::from_string("::")}
+        {bi::address_v6::from_string("::")},
```
halfalicious: Unnecessary comma?

libp2p/Common.cpp
chfast: Not possible currently. It would require constexpr constructor of `NodeIPEndpoint` which would require constexpr constructor of `boost::address` which is not available (not even in the current development version). Good idea for Boost contribution.

libp2p/Common.cpp
chfast: This is the way to tell clang-format to list elements in own lines each. Otherwise it will pack it up to the line length limit.

libp2p/Common.cpp
chfast: No. 
It cannot be `const auto` - because of `t++` and `t = `.
It cannot be `.cbegin()` (const_iterator) because of `t->wait()` which cannot be a const method.

libp2p/Common.cpp
gumb0: wow I didn't know this trick

libp2p/Common.cpp
```diff
@@ -238,16 +237,14 @@ bool NodeSpec::isValid() const
...
-
-namespace dev
-{
-namespace p2p
-{
 std::ostream& operator<<(std::ostream& _out, NodeIPEndpoint const& _ep)
 {
-    _out << _ep.address() << " UDP " << _ep.udpPort() << " TCP " << _ep.tcpPort();
+    _out << _ep.address() << ':' << _ep.tcpPort();
+    // It rarely happens that TCP and UDP are different, so safe space
```
gumb0: I think you meant "save space"

PR #5469, opened 2019-02-05
Refactor capability

chfast:


libp2p/Host.cpp
```diff
@@ -486,15 +485,10 @@ std::unordered_map<Public, std::string> Host::pocHosts()
     };
 }
 
-void Host::registerCapability(std::shared_ptr<CapabilityFace> const& _cap)
+void Host::registerCapability(std::shared_ptr<CapabilityFace> _cap)
 {
-    registerCapability(_cap, _cap->name(), _cap->version());
-}
-
-void Host::registerCapability(
```
gumb0: I kept this on purpose.

The idea of having these two overloads is to allow one Capability implementation to handle two (or more) protocol versions. It was used previously to support both `eth/62` and `eth/63`, now I removed support for 62 for the sake of simplification, but it might be needed if at some point there's `eth/64` and we have to support it.
Also in general seems to be a useful feature of p2p library.
(though not properly tested I think)

gumb0: (that is, it's needed if there are clients in the network that support only older version, but we still want to talk to them)

libp2p/Session.cpp
```diff
@@ -44,8 +44,6 @@ Session::Session(Host* _h, unique_ptr<RLPXFrameCoder>&& _io, std::shared_ptr<RLP
 {
     m_peer->m_lastDisconnect = NoDisconnect;
     m_lastReceived = m_connect = chrono::steady_clock::now();
-    DEV_GUARDED(x_info)
-        m_info.socketId = m_socket->ref().native_handle();
```
gumb0: We could output it from `admin_peers` RPC as other `PeerSessionInfo` members. Not sure if it's useful though, maybe not

libp2p/Host.h
```diff
@@ -135,13 +135,7 @@ class Host: public Worker
     static std::unordered_map<Public, std::string> pocHosts();
 
     /// Register a host capability; all new peer connections will see this capability.
-    void registerCapability(std::shared_ptr<CapabilityFace> const& _cap);
-
-    /// Register a host capability with arbitrary name and version.
-    /// Might be useful when you want to handle several subprotocol versions with a single
```
gumb0: I even wrote this explaining comment :cry:

chfast: Restored.

PR #5472, opened 2019-02-06
Add log formatting for Node type

chfast:
I hoped there are more cases where it could be used.

### DIFF
--- a/libp2p/Common.h
+++ b/libp2p/Common.h
@@ -260,6 +260,8 @@ public:
     std::atomic<PeerType> peerType{PeerType::Optional};
 };
 
+boost::log::formatting_ostream& operator<<(boost::log::formatting_ostream& _log, const Node& _node);
+
 class DeadlineOps
 {
     // Boost deadline timer wrapper which provides thread-safety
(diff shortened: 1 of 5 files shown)


gumb0:
Needs rebase.

libp2p/Common.h
```diff
@@ -245,10 +245,9 @@ class Node
     Node(Public _publicKey, NodeIPEndpoint const& _ip, PeerType _peerType = PeerType::Optional): id(_publicKey), endpoint(_ip), peerType(_peerType) {}
     Node(NodeSpec const& _s, PeerType _peerType = PeerType::Optional);
 
-    virtual NodeID const& address() const { return id; }
-    virtual Public const& publicKey() const { return id; }
-    
-    virtual operator bool() const { return (bool)id; }
+    NodeID const& address() const { return id; }
+
+    explicit operator bool() const { return bool{id}; }
```
gumb0: Did the pattern `if (auto n = getNode())` not require change after this?

chfast: No because `explicit operator bool` can be used in `if` context.

PR #5473, opened 2019-02-07
Restore network after host marks network as initialized (m_run = true)

halfalicious:
Move the network restore call in `Host::startedWorking` (`Host::restoreNetwork`) to occur after the host marks the network as initialized (`m_run = true`) because the network restore code calls `Host::requirePeer() `which depends on `m_run = true`.


halfalicious:
@chfast / @gumb0 : Change is trivial but would one of you mind testing this real quick to be on the safe side since the restore network -> require peer code path isn't hit on Windows.

PR #5477, opened 2019-02-08
Build fixes for clang-8

chfast:


libp2p/RLPXFrameCoder.h
```diff
@@ -1,59 +1,38 @@
...
 
-struct RLPXFrameDecryptFailed: virtual dev::Exception {};
-
 /**
  * @brief Encapsulation of Frame
  * @todo coder integration; padding derived from coder
  */
 struct RLPXFrameInfo
 {
-	RLPXFrameInfo() = default;
```
chfast: The build fix is actually only here. You cannot create the default constructor while having `const` fields.

PR #5483, opened 2019-02-11
Fix multiple Ping and assertion failure in Discovery

gumb0:
Addresses https://github.com/ethereum/aleth/issues/5471 and probably some things from https://github.com/ethereum/aleth/issues/5484

Summary of changes:

- checking whether we've already sent one `Ping` (by looking at `m_sentPings`) before sending another one
- `ping` method is split into `ping` and `schedulePing` - when we are already in the network thread, we can just directly access `m_sentPings` without additional `m_timers.schedule(0, ...)`. So now when we need to ping from the network thread we just call `ping` method without scheduling
- `m_allNodes` now contains only the nodes from the node table buckets. 
- So the nodes that are still pending (`Pong` not received yet) and the nodes that didn't fit into the bucket, and wait for older node to be evicted, are not put into `m_allNodes`
- Nodes are added to `m_allNodes` only in `noteActiveNode` together with being put into the bucket.
This allows us not to care about erasing from `m_allNodes` the nodes that didn't get validated or are being thrown away when eviction ends with the old node answering.
(maybe `m_allNodes` should be renamed now)
- replacement nodes are remembered only in the `m_sentPings` items (as a `shared_ptr<NodeEntry> replacementNodeEntry` member). They are dropped when we erase from `m_sentPings` in `Pong` handler.
- we have to save node's `NodeIPEndpoint` in `m_sentPings` now, because this data is sent to us in `Ping` (or in `Neighbours`) and when we receive it from an unnknown node, we have to save it somewhere until the moment when we'll add it to the node table and to the `m_allNodes`.
(actually I think it's only TCP port number that is important to save, the rest is overwritten from the actual source of UDP packet anyway. But I expect this part to be changed somewhat when addressing the problem described in https://github.com/ethereum/aleth/issues/5455#issuecomment-460754175
I expect `m_sentPings` to have in the future the UDP endpoint as a map key instead of `NodeID`)

### DIFF
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -69,6 +69,8 @@ void NodeTable::processEvents()
 
 bool NodeTable::addNode(Node const& _node)
 {
+    LOG(m_logger) << "Adding node " << _node;
+
     shared_ptr<NodeEntry> entry;
     DEV_GUARDED(x_nodes)
     {
@@ -90,6 +92,8 @@ bool NodeTable::addNode(Node const& _node)
 bool NodeTable::addKnownNode(
     Node const& _node, uint32_t _lastPongReceivedTime, uint32_t _lastPongSentTime)
 {
+    LOG(m_logger) << "Adding known node " << _node;
+
     shared_ptr<NodeEntry> entry;
     DEV_GUARDED(x_nodes)
     {
@@ -204,6 +208,8 @@ void NodeTable::doDiscover(NodeID _node, unsigned _round, shared_ptr<set<shared_
             // This prevents being considered invalid node and FindNode being ignored.
             if (!node->hasValidEndpointProof())
             {
+                LOG(m_logger) << "Node " << static_cast<Node const&>(*node)
+                              << " endpoint proof expired.";
                 ping(*node);
                 continue;
             }
@@ -311,6 +317,9 @@ void NodeTable::ping(NodeEntry const& _nodeEntry, boost::optional<NodeID> const&
         if (_ec || m_timers.isStopped())
             return;
 
+        if (contains(m_sentPings, _nodeEntry.id))
+            return;
+
         NodeIPEndpoint src;
         src = m_hostNodeEndpoint;
         PingNode p(src, _nodeEntry.endpoint);
@@ -330,6 +339,7 @@ void NodeTable::evict(NodeEntry const& _leastSeen, NodeEntry const& _new)
     if (!m_socket->isOpen())
         return;
 
+    LOG(m_logger) << "Evicting node " << static_cast<Node const&>(_leastSeen);
     ping(_leastSeen, _new.id);
 }
 
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -317,8 +317,15 @@ void NodeTable::ping(NodeEntry const& _nodeEntry, boost::optional<NodeID> const&
         if (_ec || m_timers.isStopped())
             return;
 
-        if (contains(m_sentPings, _nodeEntry.id))
+        // don't sent Ping if one is already sent
+        auto sentPing = m_sentPings.find(_nodeEntry.id);
+        if (sentPing != m_sentPings.end())
+        {
+            // we don't need replacement if we're not going to ping
+            if (_replacementNodeID && sentPing->second.replacementNodeID != _replacementNodeID)
+                DEV_GUARDED(x_nodes) { m_allNodes.erase(*_replacementNodeID); }
             return;
+        }
 
         NodeIPEndpoint src;
         src = m_hostNodeEndpoint;
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -71,20 +71,21 @@ bool NodeTable::addNode(Node const& _node)
 {
     LOG(m_logger) << "Adding node " << _node;
 
-    shared_ptr<NodeEntry> entry;
+    if (!isValidNode(_node))
+        return false;
+
+    bool needToPing = false;
     DEV_GUARDED(x_nodes)
     {
         auto const it = m_allNodes.find(_node.id);
-        if (it == m_allNodes.end())
-            entry = createNodeEntry(_node, 0, 0);
-        else
-            entry = it->second;
+        needToPing = (it == m_allNodes.end() || !it->second->hasValidEndpointProof());
     }
-    if (!entry)
-        return false;
 
-    if (!entry->hasValidEndpointProof())
-        schedulePing(*entry);
+    if (needToPing)
+    {
+        LOG(m_logger) << "Pending " << _node;
+        schedulePing(_node);
+    }
 
     return true;
 }
@@ -94,58 +95,55 @@ bool NodeTable::addKnownNode(
 {
     LOG(m_logger) << "Adding known node " << _node;
 
-    shared_ptr<NodeEntry> entry;
-    DEV_GUARDED(x_nodes)
-    {
-        entry = createNodeEntry(_node, _lastPongReceivedTime, _lastPongSentTime);
-    }
-    if (!entry)
+    if (!isValidNode(_node))
         return false;
 
+    if (nodeEntry(_node.id))
+    {
+        LOG(m_logger) << "Node " << _node << " is already in the node table";
+        return true;
+    }
+
+    auto entry = make_shared<NodeEntry>(
+        m_hostNodeID, _node.id, _node.endpoint, _lastPongReceivedTime, _lastPongSentTime);
+
     if (entry->hasValidEndpointProof())
-        noteActiveNode(entry->id, entry->endpoint);
+    {
+        LOG(m_logger) << "Known " << _node;
+        noteActiveNode(move(entry), entry->endpoint);
+    }
     else
-        schedulePing(*entry);
+    {
+        LOG(m_logger) << "Pending " << _node;
+        schedulePing(_node);
+    }
 
     return true;
 }
 
-std::shared_ptr<NodeEntry> NodeTable::createNodeEntry(
-    Node const& _node, uint32_t _lastPongReceivedTime, uint32_t _lastPongSentTime)
+bool NodeTable::isValidNode(Node const& _node) const
 {
     if (!_node.endpoint || !_node.id)
     {
         LOG(m_logger) << "Supplied node " << _node
                       << " has an invalid endpoint or id. Skipping adding node to node table.";
-        return {};
+        return false;
     }
 
     if (!isAllowedEndpoint(_node.endpoint))
     {
         LOG(m_logger) << "Supplied node" << _node
                       << " doesn't have an allowed endpoint. Skipping adding node to node table";
-        return {};
+        return false;
     }
 
     if (m_hostNodeID == _node.id)
     {
         LOG(m_logger) << "Skip adding self to node table (" << _node.id << ")";
-        return {};
+        return false;
     }
 
-    if (m_allNodes.find(_node.id) != m_allNodes.end())
-    {
-        LOG(m_logger) << "Node " << _node << " is already in the node table";
-        return {};
-    }
-
-    auto nodeEntry = make_shared<NodeEntry>(
-        m_hostNodeID, _node.id, _node.endpoint, _lastPongReceivedTime, _lastPongSentTime);
-    m_allNodes.insert({_node.id, nodeEntry});
-
-    LOG(m_logger) << (_lastPongReceivedTime > 0 ? "Known " : "Pending ") << _node;
-
-    return nodeEntry;
+    return true;
 }
 
 list<NodeID> NodeTable::nodes() const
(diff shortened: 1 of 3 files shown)


chfast:
The assert failure still happens here.

chfast:
Still crashes :)

gumb0:
This still doesn't fix assertion failure, just improves handling of attempts to Ping multiple times.

But it gets quite confusing and difficult to keep in mind all the possible cases. I'm thinking now that `m_allNodes` shouldn't contain both the pending nodes and the nodes already in the node table. Maybe we'll create the `NodeEntry` only when it's actually added to the node table. Then there won't be the need to this assert.

libp2p/NodeTable.cpp
```diff
@@ -325,6 +331,9 @@ void NodeTable::ping(NodeEntry const& _nodeEntry, boost::optional<NodeID> const&
         if (_ec || m_timers.isStopped())
             return;
 
+        if (contains(m_sentPings, _nodeEntry.id))
```
halfalicious: Should we maybe log something here?

libp2p/NodeTable.cpp
```diff
@@ -320,22 +326,30 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID _target)
 
 void NodeTable::ping(NodeEntry const& _nodeEntry, boost::optional<NodeID> const& _replacementNodeID)
 {
-    m_timers.schedule(0, [this, _nodeEntry, _replacementNodeID](
-                             boost::system::error_code const& _ec) {
+    // don't sent Ping if one is already sent
+    if (contains(m_sentPings, _nodeEntry.id))
+        return;
+
+    NodeIPEndpoint src;
```
halfalicious: (Nit) Can be combined onto one line?

libp2p/NodeTable.cpp
```diff
@@ -320,22 +326,30 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID _target)
...
+    src = m_hostNodeEndpoint;
+    PingNode p(src, _nodeEntry.endpoint);
+    p.ts = nextRequestExpirationTime();
+    auto const pingHash = p.sign(m_secret);
+    LOG(m_logger) << p.typeName() << " to " << _nodeEntry.id << "@" << p.destination;
```
halfalicious: Can we just log the _nodeEntry here to get both the ID and the destination endpoint?

gumb0:
It turned out to be quite a lot of change, but I think it haven't got more complicated at least.

gumb0:
Rebased and addressed @halfalicious's comments

libp2p/NodeTable.h
```diff
@@ -126,14 +126,17 @@ class NodeTable : UDPSocketEvents
     /// Called by implementation which provided handler to process NodeEntryAdded/NodeEntryDropped events. Events are coalesced by type whereby old events are ignored.
     void processEvents();
 
-    /// Add node to the list of all nodes and ping it to trigger the endpoint proof.
+    /// Starts async node adding tot the node table by pinging it to trigger the endpoint proof.
```
halfalicious: Typo in comment (tot) and think it could be rephrased a little bit eg Starts async add of node to the node table...

libp2p/NodeTable.h
```diff
@@ -126,14 +126,17 @@ class NodeTable : UDPSocketEvents
...
+    /// In case the node is already in the node table, pings only if the endpoint proof expired.
     ///
-    /// @return True if the node has been added.
+    /// @return True if the node id valid.
```
halfalicious: Grammar: “If the node is is valid”

libp2p/NodeTable.cpp
```diff
@@ -69,79 +69,81 @@ void NodeTable::processEvents()
 
 bool NodeTable::addNode(Node const& _node)
 {
-    shared_ptr<NodeEntry> entry;
+    LOG(m_logger) << "Adding node " << _node;
+
+    if (!isValidNode(_node))
+        return false;
```
halfalicious: I think it would be good to have a log statement here

halfalicious:
What were the cases where we would send double pings?

gumb0:
Cases when we sent double pings:

- We receive `Neighbours` packet, it has some nodes that should go into the same bucket. But the bucket is full, when trying to add the first new node, we start eviction by pinging the oldest node of the bucket. Before `Ping` arrives, we add subsequent new nodes, and this leads to pinging the oldest node again.
(after the fix it ingores the attempts to add the subsequent new nodes)
- Sometimes we've already scheduled initiating a ping (with `m_timers.schedule(0, ...`), if another request to add the same node happens immediately in succession, we schedule another ping (i.e. we don't look into `io_service` to check whether we already plan to ping soon) - I beleive this sometimes still can happen after these fixes, but should be less often after the split to `ping/schedulePing` (regular `ping` now doesn't do scheduling with `m_timers`)

The assertion failure:
The reason was `m_allNodes` and `m_sentPings` getting out of sync, i.e. `m_sentPings` (in `Pong` handler) contained the ID of some node that already was erased from `m_allNodes`
I think this could happen when we ping the node that is considered already a "replacement node" (the one that we already tried to add to the bucket, but the bucket was full and therefore we pinged the oldest node instead), but then the `Pong` from the oldest node arrives and we erase replacement node from `m_allNodes` before `Pong` of replacement node arrives

libp2p/NodeTable.cpp
gumb0: I think there are already enough logs inside `isValidNode`

libp2p/NodeTable.cpp
```diff
@@ -304,101 +308,117 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID _target)
...
-    ping(_leastSeen, _new.id);
+    LOG(m_logger) << "Evicting node " << static_cast<Node const&>(_leastSeen);
+    ping(_leastSeen, std::move(_replacement));
+
+    if (m_nodeEventHandler)
+        m_nodeEventHandler->appendEvent(_leastSeen.id, NodeEntryScheduledForEviction);
 }
 
-void NodeTable::noteActiveNode(Public const& _pubk, bi::udp::endpoint const& _endpoint)
+void NodeTable::noteActiveNode(shared_ptr<NodeEntry> _nodeEntry, bi::udp::endpoint const& _endpoint)
```
gumb0: The changes in this method are only:
- `shared_ptr<NodeEntry>` instead of `Public` paramter.
- insert into `m_allNodes` together with inserting into the bucket

gumb0:
Tests seem to be fixed now, I think it's ready for review

libp2p/NodeTable.cpp
```diff
@@ -304,101 +308,117 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID _target)
     return ret;
 }
 
-void NodeTable::ping(NodeEntry const& _nodeEntry, boost::optional<NodeID> const& _replacementNodeID)
+void NodeTable::ping(Node const& _node, shared_ptr<NodeEntry> _replacementNodeEntry)
+{
+    // Don't sent Ping if one is already sent
+    if (contains(m_sentPings, _node.id))
+    {
+        LOG(m_logger) << "Ignoring request to ping " << _node << ", because it's already pinged";
```
gumb0: This is being logged quite a lot I should say

halfalicious: Yea, I think we benefit from more logging since it makes it easier to track what's going on and detect bugs but wading through the log spew can be challenging. It would be nice to eventually have some additional log levels so we could have more targeted logging.

libp2p/NodeTable.cpp
halfalicious: Ah yes, agreed

libp2p/NodeTable.cpp
```diff
@@ -69,79 +69,81 @@ void NodeTable::processEvents()
...
+
+    bool needToPing = false;
     DEV_GUARDED(x_nodes)
     {
         auto const it = m_allNodes.find(_node.id);
-        if (it == m_allNodes.end())
-            entry = createNodeEntry(_node, 0, 0);
-        else
-            entry = it->second;
+        needToPing = (it == m_allNodes.end() || !it->second->hasValidEndpointProof());
```
halfalicious: Should we also check m_sentPings? What if we sent a ping recently but just haven't received a response yet?

gumb0: yeah good point, it would improve things, but unfortunately it's not easy to do it here, because `m_sentPings` is accessed only from the network thread...

gumb0: We could either make `addNode` asynchronous; or we could create another private method for adding nodes from the packet handlers (but it would be very similar to `addNode`); or maybe we could check before sending Ping in `ping()` whether the node already happened to be validated, then skip additional `Ping`.

All this seems quite ugly for me for now, I'd suggest first to observe how often we get double pings because of this problem, then decide whether it make sense to complicate the code...

halfalicious:
>     * Sometimes we've already scheduled initiating a ping (with `m_timers.schedule(0, ...`), if another request to add the same node happens immediately in succession, we schedule another ping (i.e. we don't look into `io_service` to check whether we already plan to ping soon) - I beleive this sometimes still can happen after these fixes, but should be less often after the split to `ping/schedulePing` (regular `ping` now doesn't do scheduling with `m_timers`)

Can we address this by adding the ping information in schedulePing() before we actually schedule the ping? One danger here is that there's a delay of the ping not actually executing until after the UDP datagram time to live (1 minute) which means that the ping is already considered expired when it gets sent out over the wire. I don't know much about boost deadline timers but I don't think this can realistically happen, even if there are a lot of boost deadline timers expiring around the same time (since you'd have to execute a lot of handlers to consume 1 minute of wall clock time and I don't think we create that many deadline timers).

Another possible way to address this would be via something like m_queuedPings, and we could check both this and m_sentPings before deciding if we should queue a ping.

libp2p/NodeTable.cpp
chfast: Separating logs into INFO and DEBUG levels would be good. In INFO only actual changes (node added, node replaced by) and/or stats (X new nodes discovered).
The rest should go to DEBUG. Later we can consider splitting DEBUG into DEBUG and TRACE.

For aleth-bootnode the networking INFO level should be enabled by default.
For aleth all networking logs should be disabled.

libp2p/NodeTable.cpp
```diff
@@ -304,101 +308,117 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID _target)
...
     {
         LOG(m_logger) << "Skipping making self active.";
         return;
     }
     if (!isAllowedEndpoint(NodeIPEndpoint(_endpoint.address(), _endpoint.port(), _endpoint.port())))
     {
-        LOG(m_logger) << "Skipping making node with unallowed endpoint active. Node " << _pubk
-                      << "@" << _endpoint;
+        LOG(m_logger) << "Skipping making node with unallowed endpoint active. Node "
+                      << _nodeEntry->id << "@" << _endpoint;
```
chfast: You can log `_nodeEntry` here.

libp2p/NodeTable.cpp
```diff
@@ -304,101 +308,117 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID _target)
...
-    shared_ptr<NodeEntry> newNode = nodeEntry(_pubk);
-    if (newNode && newNode->hasValidEndpointProof())
-    {
-        LOG(m_logger) << "Active node " << _pubk << '@' << _endpoint;
-        newNode->endpoint.setAddress(_endpoint.address());
-        newNode->endpoint.setUdpPort(_endpoint.port());
+    if (!_nodeEntry->hasValidEndpointProof())
+        return;
 
+    LOG(m_logger) << "Active node " << _nodeEntry->id << '@' << _endpoint;
```
chfast: You can long `_nodeEntry` here.

libp2p/NodeTable.cpp
halfalicious: > For aleth-bootnode the networking INFO level should be enabled by default.

Agreed, filed #5499 to track this

halfalicious:
> 
> 
> > ```
> > * Sometimes we've already scheduled initiating a ping (with `m_timers.schedule(0, ...`), if another request to add the same node happens immediately in succession, we schedule another ping (i.e. we don't look into `io_service` to check whether we already plan to ping soon) - I beleive this sometimes still can happen after these fixes, but should be less often after the split to `ping/schedulePing` (regular `ping` now doesn't do scheduling with `m_timers`)
> > ```
> 
> Can we address this by adding the ping information in schedulePing() before we actually schedule the ping? One danger with doing this is that there's a delay of the ping not actually executing until after the UDP datagram time to live (1 minute) which means that the ping is already considered expired when it gets sent out over the wire. I don't know much about boost deadline timers but I don't think this can realistically happen, even if there are a lot of boost deadline timers expiring around the same time (since you'd have to execute a lot of handlers to consume 1 minute of wall clock time and I don't think we create that many deadline timers).
> 
> Another possible way to address this would be via something like m_queuedPings, and we could check both this and m_sentPings before deciding if we should queue a new ping.

Filed #5500 to track this

PR #5487, opened 2019-02-12
EIP-8 unittests

chfast:
@halfalicious Can you try enabling these tests on Windows?

### DIFF
--- a/test/unittests/libp2p/eip-8.cpp
+++ b/test/unittests/libp2p/eip-8.cpp
@@ -24,9 +24,9 @@ TEST(eip8, test_discovery_packets)
         "e9614ccfd9fc3e74360018522d30e1419a143407ffcce748de3e22116b7e8dc92ff74788c0b6663a"
         "aa3d67d641936511c8f8d6ad8698b820a7cf9e1be7155e9a241f556658c55428ec0563514365799a"
         "4be2be5a685a80971ddcfa80cb422cdd0101ec04cb847f000001820cfa8215a8d790000000000000"
-        "000000000000000000018208ae820d058443b9a3550102"
-    );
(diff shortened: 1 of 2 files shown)


halfalicious:
> 
> 
> @halfalicious Can you try enabling these tests on Windows?

I ran them a little while ago and they hit AVs, do we expect your changes to fix the AVs?

halfalicious:
> 
> 
> > @halfalicious Can you try enabling these tests on Windows?
> 
> I ran them a little while ago and they hit AVs, do we expect your changes to fix the AVs?

Actually scratch that, I built and ran your changes and the eip 8 test passes:

```
[----------] 1 test from eip8
[ RUN      ] eip8.test_discovery_packets
[       OK ] eip8.test_discovery_packets (2 ms)
[----------] 1 test from eip8 (2 ms total)
```

chfast:
> Actually scratch that, I built and ran your changes and the eip 8 test passes:

Most of the tests are still disabled on Windows.

test/unittests/libp2p/eip-8.cpp
```diff
@@ -1,40 +1,21 @@
...
 
 using namespace std;
 using namespace dev;
-using namespace dev::test;
 using namespace dev::p2p;
 
-BOOST_FIXTURE_TEST_SUITE(eip8, TestOutputHelperFixture)
-
-BOOST_AUTO_TEST_CASE(test_discovery_packets)
+TEST(eip8, test_discovery_packets)
```
gumb0: I think we wanted to use camelCase for test names?

chfast: Would be good at some point to convert names. And maybe add library prefixes.

PR #5490, opened 2019-02-13
Docker fix

chfast:


circle.yml
```diff
@@ -133,7 +133,7 @@ jobs:
       GENERATOR: Ninja
       BUILD_PARALLEL_JOBS: 8
       TEST_PARALLEL_JOBS: 8
-      CMAKE_OPTIONS: -DALETH_INTERPRETER_SHARED=ON
+      CMAKE_OPTIONS: -DALETH_INTERPRETER_SHARED=ON -DVMTRACE=ON
```
gumb0: why?

chfast: It's useful feature. If you download binaries you might want to use it. Otherwise you get error that you have to build project with this option.

I'd like to make it an runtime option only. We have to review what it changes.

yxliang01:
@gumb0 @chfast  I can see this updated dockerfile is included in v1.6.0-rc.1 . But, the published docker image for v1.6.0-rc.1 still wasn't built with `VMTRACE=On`. Is it that the official images are not built using the `aleth.dockerfile` to build?

PR #5485, opened 2019-02-12
Enable aleth-bootnode to connect to official Ethereum bootnodes

halfalicious:
Enable aleth-bootnode to connect to the official Ethereum bootnodes. This is now the behavior by default, but it can be disabled via the new "--no-boostrap" flag. 

I've included this functionality because I think it's useful for discovery testing purposes. Additionally, it now enables one to run a discovery-only node that can participate on a public network without one having to manually bootstrap it with another Ethereum node.

PR #5491, opened 2019-02-13
Update Aleth tools to return error codes

halfalicious:
Prior to these changes the various Aleth tools (e.g. aleth, aleth-bootnode, aleth-vm) would return -1 on error. These changes define a new error enum (AlethError) with error codes for the various error conditions and update the various tools to return these codes.

### DIFF
--- a/aleth-bootnode/main.cpp
+++ b/aleth-bootnode/main.cpp
@@ -17,6 +17,7 @@
 
 #include <libdevcore/FileSystem.h>
 #include <libdevcore/LoggingProgramOptions.h>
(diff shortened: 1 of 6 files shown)


halfalicious:
Need to fix bad cherry pick

halfalicious:
> 
> 
> Need to fix bad cherry pick

Fixed

halfalicious:
I'd like to merge this change before the aleth-bootnode one (#5485) so I can update aleth-bootnode to return the appropriate error code for the new error that it detects (network not started).

PR #5492, opened 2019-02-14
Fix unreliable discovery unit test (pingNotSentAfterPongForKnownNode)

### LINKED ISSUE #5488: Discovery unit test pingNotSentAfterPongForKnownNode fails intermittently
You can run the test via `testeth -t network/net/pingNotSentAfterPongForKnownNode`...the test fails intermittently in CircleCI: 

```
Running 1 test case...
Test Case "pingNotSentAfterPongForKnownNode": 
/home/builder/project/libdevcore/concurrent_queue.h(47): fatal error: in "_T dev::concurrent_queue<_T, _QueueT>::pop(const milliseconds&) [with _T = std::vector<unsigned char>; _QueueT = std::queue<std::vector<unsigned char>, std::deque<std::vector<unsigned char>, std::allocator<std::vector<unsigned char> > > >; std::chrono::milliseconds = std::chrono::duration<long int, std::ratio<1l, 1000l> >]": boost::exception_detail::clone_impl<dev::WaitTimeout>: /home/builder/project/libdevcore/concurrent_queue.h(47): Throw in function _T dev::concurrent_queue<_T, _QueueT>::pop(const milliseconds&) [with _T = std::vector<unsigned char>; _QueueT = std::queue<std::vector<unsigned char>, std::deque<std::vector<unsigned char>, std::allocator<std::vector<unsigned char> > > >; std::chrono::milliseconds = std::chrono::duration<long int, std::ratio<1l, 1000l> >]
Dynamic exception type: boost::exception_detail::clone_impl<dev::WaitTimeout>

/home/builder/project/test/unittests/libp2p/net.cpp(1069): last checkpoint
```

It looks like no packet is received after the pong (within 20 seconds) which is sent in response to the second ping...I'd expect a FindNode packet to be received:
https://github.com/ethereum/aleth/blob/0aff0181ba027d98e8612565d35855104be1024e/test/unittests/libp2p/net.cpp#L1060-L1075

If all we want to do is to verify that another ping isn't sent after we receive the pong, perhaps we can just listen for a short time period (e.g. 3 seconds) since if another ping were to be sent I'd expect it to be sent very quickly i.e. within a second or two? We could do something like:

`BOOST_REQUIRE_THROW(nodeSocketHost2.packetsReceived.pop(chrono::milliseconds(3000)), WaitTimeout)`

halfalicious:
Fix #5488 

Fix intermittent failure in pingNotSentAfterPongForKnownNode test by shortening wait time for ping packet to 3 seconds and expecting wait to fail. This is reasonable since any ping sent after the pong for a known node would be sent nearly immediately.

### DIFF
--- a/test/unittests/libp2p/net.cpp
+++ b/test/unittests/libp2p/net.cpp
@@ -1068,11 +1068,8 @@ BOOST_AUTO_TEST_CASE(pingNotSentAfterPongForKnownNode)
         DiscoveryDatagram::interpretUDP(bi::udp::endpoint{}, dev::ref(packetReceived4));
     BOOST_REQUIRE_EQUAL(datagram4->typeName(), "Pong");
 
-    // Verify that the next packet received is not a ping
-    auto packetReceived5 = nodeSocketHost2.packetsReceived.pop(chrono::milliseconds(20000));
-    auto datagram5 =
-        DiscoveryDatagram::interpretUDP(bi::udp::endpoint{}, dev::ref(packetReceived5));
-    BOOST_REQUIRE(datagram5->typeName() != "Ping");
+    // Verify that another ping isn't sent
+    BOOST_REQUIRE_THROW(nodeSocketHost2.packetsReceived.pop(chrono::seconds(3)), WaitTimeout);
 
     // Force the endpoint proof to be invalid
     auto newNode = nodeTable1->nodeEntry(nodeKeyPair2.pub());
--- a/test/unittests/libp2p/net.cpp
+++ b/test/unittests/libp2p/net.cpp
@@ -435,7 +435,7 @@ BOOST_AUTO_TEST_CASE(kademlia)
     nodes.sort();
     node.nodeTable->reset();
     node.populate(1);
-    this_thread::sleep_for(chrono::milliseconds(2000));
+    this_thread::sleep_for(chrono::seconds(2));
     BOOST_REQUIRE_EQUAL(node.nodeTable->count(), 8);
 }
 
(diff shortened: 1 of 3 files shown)


test/unittests/libp2p/net.cpp
```diff
@@ -1068,11 +1068,8 @@ BOOST_AUTO_TEST_CASE(pingNotSentAfterPongForKnownNode)
         DiscoveryDatagram::interpretUDP(bi::udp::endpoint{}, dev::ref(packetReceived4));
     BOOST_REQUIRE_EQUAL(datagram4->typeName(), "Pong");
 
-    // Verify that the next packet received is not a ping
-    auto packetReceived5 = nodeSocketHost2.packetsReceived.pop(chrono::milliseconds(20000));
-    auto datagram5 =
-        DiscoveryDatagram::interpretUDP(bi::udp::endpoint{}, dev::ref(packetReceived5));
-    BOOST_REQUIRE(datagram5->typeName() != "Ping");
+    // Verify that another ping isn't sent
+    BOOST_REQUIRE_THROW(nodeSocketHost2.packetsReceived.pop(chrono::milliseconds(3000)), WaitTimeout);
```
chfast: Why are you using `milliseconds(3000)` instead of `seconds(3)`.
Will this test wait 3 seconds all the time?

halfalicious: Will switch to seconds, this particular check should wait up to 3 seconds so chrono::seconds will be fine.

halfalicious:
cc @chfast

PR #5494, opened 2019-02-14
Return default bootnodes by vector

chfast:
This replaces usage of map with vector. It also removes the static duration of the container. It is not needed. This function is usually used once and will just have single vector allocation followed by deallocation when the items are processed.


halfalicious:
Just to confirm, the endpoint strings are literals so they end up being stored in the data section of the Aleth binary correct?

PR #5501, opened 2019-02-22
Rename network timers and log NodeEntry using ostream operator

halfalicious:
Rename timers in the node table and host for better code readability and log NodeEntry using the ostream operator. Also change the run timer interval from a static class var to a var in an unnnamed namespace since it's not accessed outside of Host.cpp.

### DIFF
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -163,7 +163,9 @@ void Host::doneWorking()
     m_ioService.reset();
 
     DEV_GUARDED(x_networkTimers)
-    m_networkTimers.clear();
+    {
+        m_networkTimers.clear();
+    }
 
     // shutdown acceptor
     m_tcp4Acceptor.cancel();
@@ -683,9 +685,11 @@ void Host::run(boost::system::error_code const& _ec)
     DEV_GUARDED(x_connecting)
         m_connecting.remove_if([](weak_ptr<RLPXHandshake> h){ return h.expired(); });
     DEV_GUARDED(x_networkTimers)
-    m_networkTimers.remove_if([](unique_ptr<io::deadline_timer> const& t) {
-        return t->expires_from_now().total_milliseconds() < 0;
-    });
+    {
+        m_networkTimers.remove_if([](unique_ptr<io::deadline_timer> const& t) {
+            return t->expires_from_now().total_milliseconds() < 0;
+        });
+    }
 
     keepAlivePeers();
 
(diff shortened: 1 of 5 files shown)


libp2p/Host.cpp
```diff
@@ -159,9 +162,9 @@ void Host::doneWorking()
     // reset ioservice (cancels all timers and allows manually polling network, below)
     m_ioService.reset();
 
-    DEV_GUARDED(x_timers)
-        m_timers.clear();
-    
+    DEV_GUARDED(x_networkTimers)
+    m_networkTimers.clear();
```
chfast: Enclose with `{}`.

libp2p/Host.cpp
```diff
@@ -672,37 +675,39 @@ void Host::run(boost::system::error_code const& _ec)
...
+    if (auto nodeTable = this->nodeTable())
         nodeTable->processEvents();
 
     // cleanup zombies
     DEV_GUARDED(x_connecting)
         m_connecting.remove_if([](weak_ptr<RLPXHandshake> h){ return h.expired(); });
-    DEV_GUARDED(x_timers)
-    m_timers.remove_if([](unique_ptr<io::deadline_timer> const& t) {
+    DEV_GUARDED(x_networkTimers)
+    m_networkTimers.remove_if([](unique_ptr<io::deadline_timer> const& t) {
```
chfast: Enclose with `{}`

libp2p/NodeTable.cpp
```diff
@@ -360,15 +360,14 @@ void NodeTable::noteActiveNode(shared_ptr<NodeEntry> _nodeEntry, bi::udp::endpoi
     }
     if (!isAllowedEndpoint(NodeIPEndpoint(_endpoint.address(), _endpoint.port(), _endpoint.port())))
     {
-        LOG(m_logger) << "Skipping making node with unallowed endpoint active. Node "
-                      << _nodeEntry->id << "@" << _endpoint;
+        LOG(m_logger) << "Skipping making node with unallowed endpoint active. Node " << dynamic_cast<Node const&>(*_nodeEntry);
```
chfast: It should be `static_cast<>`, but there is a better way of fixing that. We have to add another overload of `operator<<` for `Node&` (without `const`). It is a known problem as the same solution was applied for other types.

It's defined here: https://github.com/ethereum/aleth/blob/master/libp2p/Common.h#L247.

libp2p/Host.cpp
halfalicious: Ah, the clang-format messed things up. Will add braces.

libp2p/NodeTable.cpp
halfalicious: Ah, understood about the `static_cast` (static cast is preferred since we're casting to a non-virtual base class which means that it's a compile time operation). However, I don't understand why a better approach would be to overload the ostream operator for non-const `Node`& - why doesn't the existing overloaded `<< `work (for `const Node&`)?

a1love:
Merge

PR #5502, opened 2019-02-24
Fix deadline timer race condition which was causing premature discovery termination

### LINKED ISSUE #5495: Discovery timer stops unexpectedly
Sometimes after running `aleth` on Linux (on the mainnet) for some time (~5-10 mins), I observe this
```
DEBUG 02-15 15:09:34 p2p  discov Ping to ##719e0907…@167.99.102.6:50000
DEBUG 02-15 15:09:37 p2p  discov Ping from ##201dd24c…@34.248.64.28:21000
DEBUG 02-15 15:09:37 p2p  discov Adding node ##201dd24c…@34.248.64.28:21000
DEBUG 02-15 15:09:37 p2p  discov Pending ##201dd24c…@34.248.64.28:21000
DEBUG 02-15 15:09:37 p2p  discov Pong to ##201dd24c…@34.248.64.28:21000
DEBUG 02-15 15:09:41 p2p  discov Discovery timer was probably cancelled: 125 Operation canceled
```
Error is returned to the timer's callback, and after that discovery stops.
No idea why, we should look into which error is returned.
(the process doesn't stop, some Pings and Pongs continue to be received/sent)

I observe this on my branch from https://github.com/ethereum/aleth/pull/5483, not sure if the problem existed before those changes.

halfalicious:
Fix #5495 

Fix the boost deadline timer race condition which was causing the discovery process to terminate (in some cases extremely) prematurely. The race condition occurred because `DeadlineOps` would check for timer expiration (`deadline_timer::expires_from_now() < 0`) and delete expired timers (see `DeadlineOps::reap()`), but their handlers may not have executed yet which means that when they finally executed their boost error code would indicate operation aborted and the handler would return before executing any of its core logic.

To fix this, I created 2 dedicated deadline timers in `NodeTable` - 1 for the discovery process and 1 for the node eviction process. I think this makes sense since these are 2 algorithms which are supposed to be running for as long as the node table is running…additionally, having dedicated timers here makes the code easier to understand. I also only check the boost error code in these timer handlers since the timers are cancelled in `NodeTable::stop()` (which is executed on both network shutdown and on `NodeTable `destruction), meaning that the error code is guaranteed to indicate timer cancellation when a handler is executed after network shutdown or node table destruction.

I've also removed the `DeadlineOps `class since it's no longer needed.

### DIFF
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -17,6 +17,7 @@ BOOST_LOG_INLINE_GLOBAL_LOGGER_CTOR_ARGS(g_discoveryWarnLogger,
     (boost::log::keywords::severity = 0)(boost::log::keywords::channel = "discov"))
 
 constexpr unsigned c_processEvictionsIntervalMs = 5000;
+
 }  // namespace
 
 constexpr chrono::seconds DiscoveryDatagram::c_timeToLive;
@@ -55,31 +56,12 @@ NodeTable::NodeTable(ba::io_service& _io, KeyPair const& _alias, NodeIPEndpoint 
     try
     {
         m_socket->connect();
-    }
-    catch (exception const& _e)
-    {
-        cwarn << "Exception connecting NodeTable socket: " << _e.what();
-        cwarn << "Discovery disabled.";
-        return;
-    }
-}
-
-void NodeTable::start()
-{
-    if (!m_socket->isOpen())
-    {
-        cwarn << "NodeTable socket isn't open. Discovery is disabled";
-        return;
-    }
-
-    try
-    {
         doDiscovery();
         doProcessEvictions();
     }
     catch (exception const& _e)
     {
-        cwarn << "Exception starting the discovery process: " << _e.what();
+        cwarn << "Exception connecting NodeTable socket: " << _e.what();
         cwarn << "Discovery disabled.";
     }
 }
@@ -215,7 +197,8 @@ shared_ptr<NodeEntry> NodeTable::nodeEntry(NodeID const& _id)
 void NodeTable::doDiscoveryRound(
     NodeID _node, unsigned _round, shared_ptr<set<shared_ptr<NodeEntry>>> _tried)
 {
-    // NOTE: ONLY called by doDiscovery or via a lambda scheduled via deadline timer
+    // NOTE: ONLY called by doDiscovery or "recursively" via lambda scheduled via deadline timer at
+    // the end of this function
     if (!m_socket->isOpen())
         return;
 
@@ -257,13 +240,12 @@ void NodeTable::doDiscoveryRound(
 
     m_discoveryTimer->expires_from_now(
         boost::posix_time::milliseconds(c_discoveryRoundIntervalMs.count()));
-
-    auto self = shared_from_this();
-    shared_ptr<ba::deadline_timer> discoveryTimer{m_discoveryTimer};
+    auto discoveryTimer{m_discoveryTimer};
     m_discoveryTimer->async_wait(
-        [self, this, discoveryTimer, _node, _round, _tried](boost::system::error_code const& _ec) {
-            // We can't use m_logger yet since the node table might have already been destroyed
-            if (_ec == boost::asio::error::operation_aborted ||
+        [this, discoveryTimer, _node, _round, _tried](boost::system::error_code const& _ec) {
+            // We can't use m_logger here if there's an error because captured this might already be
+            // destroyed
+            if (_ec.value() == boost::asio::error::operation_aborted ||
                 discoveryTimer->expires_at() == boost::posix_time::min_date_time)
             {
                 clog(VerbosityDebug, "discov") << "Discovery timer was probably cancelled";
@@ -337,6 +319,9 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID _target)
 
 void NodeTable::ping(Node const& _node, shared_ptr<NodeEntry> _replacementNodeEntry)
 {
+    if (!m_socket->isOpen())
+        return;
+
     // Don't sent Ping if one is already sent
     if (contains(m_sentPings, _node.id))
     {
@@ -356,8 +341,7 @@ void NodeTable::ping(Node const& _node, shared_ptr<NodeEntry> _replacementNodeEn
 
 void NodeTable::schedulePing(Node const& _node)
 {
-    auto self = shared_from_this();
-    m_io.post([this, self, _node] { ping(_node, {}); });
+    m_io.post([this, _node] { ping(_node, {}); });
 }
 
 void NodeTable::evict(NodeEntry const& _leastSeen, shared_ptr<NodeEntry> _replacement)
@@ -649,50 +633,49 @@ void NodeTable::onPacketReceived(
 void NodeTable::doDiscovery()
 {
     m_discoveryTimer->expires_from_now(boost::posix_time::milliseconds(c_bucketRefresh.count()));
-
-    auto self = shared_from_this();
-    shared_ptr<ba::deadline_timer> discoveryTimer{m_discoveryTimer};
-    m_discoveryTimer->async_wait(
-        [self, this, discoveryTimer](boost::system::error_code const& _ec) {
-            // We can't use m_logger yet since the node table might have already been destroyed
-            if (_ec == boost::asio::error::operation_aborted ||
-                discoveryTimer->expires_at() == boost::posix_time::min_date_time)
-            {
-                clog(VerbosityDebug, "discov") << "Discovery timer was probably cancelled";
-                return;
-            }
-            else if (_ec)
-            {
-                clog(VerbosityDebug, "discov")
-                    << "Discovery timer error detected: " << _ec.value() << " " << _ec.message();
-                return;
-            }
-
-            NodeID randNodeId;
-            crypto::Nonce::get().ref().copyTo(randNodeId.ref().cropped(0, h256::size));
-            crypto::Nonce::get().ref().copyTo(randNodeId.ref().cropped(h256::size, h256::size));
-            LOG(m_logger) << "Starting discovery algorithm run for random node id: " << randNodeId;
-            doDiscoveryRound(randNodeId, 0 /* round */, make_shared<set<shared_ptr<NodeEntry>>>());
-        });
-}
-
-void NodeTable::doProcessEvictions()
-{
-    m_evictionTimer.expires_from_now(boost::posix_time::milliseconds(c_processEvictionsIntervalMs));
-    auto self = shared_from_this();
-    shared_ptr<ba::deadline_timer> evictionTimer{m_evictionTimer};
-    m_evictionTimer->async_wait([self, this, evictionTimer](boost::system::error_code const& _ec) {
-        // We can't use m_logger yet because the NodeTable might have already been destroyed
-        if (_ec == boost::asio::error::operation_aborted ||
-            evictionTimer->expires_at() == boost::posix_time::min_date_time)
+    auto discoveryTimer{m_discoveryTimer};
+    m_discoveryTimer->async_wait([this, discoveryTimer](boost::system::error_code const& _ec) {
+        // We can't use m_logger if an error occurred because captured this might be already
+        // destroyed
+        if (_ec.value() == boost::asio::error::operation_aborted ||
+            discoveryTimer->expires_at() == boost::posix_time::min_date_time)
         {
-            LOG(m_logger) << "evictions timer was probably cancelled";
+            clog(VerbosityDebug, "discov") << "Discovery timer was probably cancelled";
             return;
         }
         else if (_ec)
         {
-            LOG(m_logger) << "evictions timer encountered an error: " << _ec.value() << " "
-                          << _ec.message();
+            clog(VerbosityDebug, "discov")
+                << "Discovery timer error detected: " << _ec.value() << " " << _ec.message();
+            return;
+        }
+
+        NodeID randNodeId;
+        crypto::Nonce::get().ref().copyTo(randNodeId.ref().cropped(0, h256::size));
+        crypto::Nonce::get().ref().copyTo(randNodeId.ref().cropped(h256::size, h256::size));
+        LOG(m_logger) << "Starting discovery algorithm run for random node id: " << randNodeId;
+        doDiscoveryRound(randNodeId, 0 /* round */, make_shared<set<shared_ptr<NodeEntry>>>());
+    });
+}
+
+void NodeTable::doProcessEvictions()
+{
+    m_evictionTimer->expires_from_now(
+        boost::posix_time::milliseconds(c_processEvictionsIntervalMs));
+    auto evictionTimer{m_evictionTimer};
+    m_evictionTimer->async_wait([this, evictionTimer](boost::system::error_code const& _ec) {
+        // We can't use m_logger if an error occurred because captured this might be already
+        // destroyed
+        if (_ec.value() == boost::asio::error::operation_aborted ||
+            m_evictionTimer->expires_at() == boost::posix_time::min_date_time)
+        {
+            clog(VerbosityDebug, "discov") << "evictions timer was probably cancelled";
+            return;
+        }
+        else if (_ec)
+        {
+            clog(VerbosityDebug, "discov")
+                << "evictions timer encountered an error: " << _ec.value() << " " << _ec.message();
             return;
         }
 
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -16,13 +16,13 @@ BOOST_LOG_INLINE_GLOBAL_LOGGER_CTOR_ARGS(g_discoveryWarnLogger,
     boost::log::sources::severity_channel_logger_mt<>,
     (boost::log::keywords::severity = 0)(boost::log::keywords::channel = "discov"))
 
-constexpr unsigned c_handleTimeoutsIntervalMs = 5000;
-
+constexpr unsigned c_processEvictionsIntervalMs = 5000;
 }  // namespace
 
 constexpr chrono::seconds DiscoveryDatagram::c_timeToLive;
 constexpr chrono::milliseconds NodeTable::c_reqTimeout;
 constexpr chrono::milliseconds NodeTable::c_bucketRefresh;
+constexpr std::chrono::milliseconds NodeTable::c_discoveryRoundIntervalMs;
 constexpr chrono::milliseconds NodeTable::c_evictionCheckInterval;
 
 inline bool operator==(weak_ptr<NodeEntry> const& _weak, shared_ptr<NodeEntry> const& _shared)
@@ -75,10 +75,7 @@ void NodeTable::start()
     try
     {
         doDiscovery();
-        m_evictionTimer->expires_from_now(
-            boost::posix_time::milliseconds(c_handleTimeoutsIntervalMs));
-        m_evictionTimer->async_wait(bind(
-            &NodeTable::doHandleTimeouts, shared_from_this(), placeholders::_1, m_evictionTimer));
+        doProcessEvictions();
     }
     catch (exception const& _e)
     {
@@ -215,25 +212,10 @@ shared_ptr<NodeEntry> NodeTable::nodeEntry(NodeID const& _id)
     return it != m_allNodes.end() ? it->second : shared_ptr<NodeEntry>();
 }
 
-void NodeTable::doDiscover(boost::system::error_code _ec, NodeID _node, unsigned _round,
-    shared_ptr<set<shared_ptr<NodeEntry>>> _tried, shared_ptr<ba::deadline_timer> _timer)
+void NodeTable::doDiscoveryRound(
+    NodeID _node, unsigned _round, shared_ptr<set<shared_ptr<NodeEntry>>> _tried)
 {
-    // NOTE: ONLY called by doDiscovery!
-
-    // We can't use m_logger yet since the node table might have already been destroyed
-    if (_ec == boost::asio::error::operation_aborted ||
-        _timer->expires_at() == boost::posix_time::min_date_time)
-    {
-        clog(VerbosityDebug, "discov") << "Discovery timer was probably cancelled";
-        return;
-    }
-    else if (_ec)
-    {
-        clog(VerbosityDebug, "discov")
-            << "Discovery timer error detected: " << _ec.value() << " " << _ec.message();
-        return;
-    }
-
+    // NOTE: ONLY called by doDiscovery or via a lambda scheduled via deadline timer
     if (!m_socket->isOpen())
         return;
 
@@ -273,9 +255,29 @@ void NodeTable::doDiscover(boost::system::error_code _ec, NodeID _node, unsigned
         return;
     }
 
-    _timer->expires_from_now(boost::posix_time::milliseconds(c_reqTimeout.count() * 2));
-    _timer->async_wait(bind(&NodeTable::doDiscover, shared_from_this(), placeholders::_1, _node,
-        _round + 1, _tried, _timer));
+    m_discoveryTimer->expires_from_now(
+        boost::posix_time::milliseconds(c_discoveryRoundIntervalMs.count()));
+
+    auto self = shared_from_this();
+    shared_ptr<ba::deadline_timer> discoveryTimer{m_discoveryTimer};
+    m_discoveryTimer->async_wait(
+        [self, this, discoveryTimer, _node, _round, _tried](boost::system::error_code const& _ec) {
+            // We can't use m_logger yet since the node table might have already been destroyed
+            if (_ec == boost::asio::error::operation_aborted ||
+                discoveryTimer->expires_at() == boost::posix_time::min_date_time)
+            {
+                clog(VerbosityDebug, "discov") << "Discovery timer was probably cancelled";
+                return;
+            }
+            else if (_ec)
+            {
+                clog(VerbosityDebug, "discov")
+                    << "Discovery timer error detected: " << _ec.value() << " " << _ec.message();
+                return;
+            }
+
+            doDiscoveryRound(_node, _round + 1, _tried);
+        });
 }
 
 vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID _target)
@@ -646,60 +648,80 @@ void NodeTable::onPacketReceived(
 
 void NodeTable::doDiscovery()
 {
-    NodeID randNodeId;
-    crypto::Nonce::get().ref().copyTo(randNodeId.ref().cropped(0, h256::size));
-    crypto::Nonce::get().ref().copyTo(randNodeId.ref().cropped(h256::size, h256::size));
-    LOG(m_logger) << "Queueing discovery algorithm run for random node id: " << randNodeId;
-
     m_discoveryTimer->expires_from_now(boost::posix_time::milliseconds(c_bucketRefresh.count()));
-    m_discoveryTimer->async_wait(bind(&NodeTable::doDiscover, shared_from_this(), placeholders::_1,
-        randNodeId, 0 /* round */, make_shared<set<shared_ptr<NodeEntry>>>(), m_discoveryTimer));
-}
 
-void NodeTable::doHandleTimeouts(
-    boost::system::error_code const& _ec, shared_ptr<ba::deadline_timer> _timer)
-{
-    // We can't use m_logger yet because the NodeTable might have already been destroyed
-    if (_ec == boost::asio::error::operation_aborted ||
-        _timer->expires_at() == boost::posix_time::min_date_time)
-    {
-        clog(VerbosityDebug, "discov") << "handleTimeouts timer was probably cancelled";
-        return;
-    }
-    else if (_ec)
-    {
-        clog(VerbosityDebug, "discov")
-            << "handleTimeouts timer encountered an error: " << _ec.value() << " " << _ec.message();
-        return;
-    }
-
-    vector<shared_ptr<NodeEntry>> nodesToActivate;
-    for (auto it = m_sentPings.begin(); it != m_sentPings.end();)
-    {
-        if (chrono::steady_clock::now() > it->second.pingSendTime + m_requestTimeToLive)
-        {
-            if (auto node = nodeEntry(it->first))
+    auto self = shared_from_this();
+    shared_ptr<ba::deadline_timer> discoveryTimer{m_discoveryTimer};
+    m_discoveryTimer->async_wait(
+        [self, this, discoveryTimer](boost::system::error_code const& _ec) {
+            // We can't use m_logger yet since the node table might have already been destroyed
+            if (_ec == boost::asio::error::operation_aborted ||
+                discoveryTimer->expires_at() == boost::posix_time::min_date_time)
             {
-                dropNode(move(node));
-
-                // save the replacement node that should be activated
-                if (it->second.replacementNodeEntry)
-                    nodesToActivate.emplace_back(move(it->second.replacementNodeEntry));
+                clog(VerbosityDebug, "discov") << "Discovery timer was probably cancelled";
+                return;
+            }
+            else if (_ec)
+            {
+                clog(VerbosityDebug, "discov")
+                    << "Discovery timer error detected: " << _ec.value() << " " << _ec.message();
+                return;
             }
 
-            it = m_sentPings.erase(it);
+            NodeID randNodeId;
+            crypto::Nonce::get().ref().copyTo(randNodeId.ref().cropped(0, h256::size));
+            crypto::Nonce::get().ref().copyTo(randNodeId.ref().cropped(h256::size, h256::size));
+            LOG(m_logger) << "Starting discovery algorithm run for random node id: " << randNodeId;
+            doDiscoveryRound(randNodeId, 0 /* round */, make_shared<set<shared_ptr<NodeEntry>>>());
+        });
+}
+
+void NodeTable::doProcessEvictions()
+{
+    m_evictionTimer.expires_from_now(boost::posix_time::milliseconds(c_processEvictionsIntervalMs));
+    auto self = shared_from_this();
+    shared_ptr<ba::deadline_timer> evictionTimer{m_evictionTimer};
+    m_evictionTimer->async_wait([self, this, evictionTimer](boost::system::error_code const& _ec) {
+        // We can't use m_logger yet because the NodeTable might have already been destroyed
+        if (_ec == boost::asio::error::operation_aborted ||
+            evictionTimer->expires_at() == boost::posix_time::min_date_time)
+        {
+            LOG(m_logger) << "evictions timer was probably cancelled";
+            return;
         }
-        else
-            ++it;
-    }
+        else if (_ec)
+        {
+            LOG(m_logger) << "evictions timer encountered an error: " << _ec.value() << " "
+                          << _ec.message();
+            return;
+        }
 
-    // activate replacement nodes and put them into buckets
-    for (auto const& n : nodesToActivate)
-        noteActiveNode(n, n->endpoint);
+        vector<shared_ptr<NodeEntry>> nodesToActivate;
+        for (auto it = m_sentPings.begin(); it != m_sentPings.end();)
+        {
+            if (chrono::steady_clock::now() > it->second.pingSendTime + m_requestTimeToLive)
+            {
+                if (auto node = nodeEntry(it->first))
+                {
+                    dropNode(move(node));
 
-    _timer->expires_from_now(boost::posix_time::milliseconds(c_handleTimeoutsIntervalMs));
-    _timer->async_wait(
-        bind(&NodeTable::doHandleTimeouts, shared_from_this(), placeholders::_1, _timer));
+                    // save the replacement node that should be activated
+                    if (it->second.replacementNodeEntry)
+                        nodesToActivate.emplace_back(move(it->second.replacementNodeEntry));
+                }
+
+                it = m_sentPings.erase(it);
+            }
+            else
+                ++it;
+        }
+
+        // activate replacement nodes and put them into buckets
+        for (auto const& n : nodesToActivate)
+            noteActiveNode(n, n->endpoint);
+
+        doProcessEvictions();
+    });
 }
 
 unique_ptr<DiscoveryDatagram> DiscoveryDatagram::interpretUDP(bi::udp::endpoint const& _from, bytesConstRef _packet)
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -20,13 +20,12 @@ constexpr unsigned c_handleTimeoutsIntervalMs = 5000;
 
 }  // namespace
 
-constexpr std::chrono::seconds DiscoveryDatagram::c_timeToLive;
-constexpr std::chrono::milliseconds NodeTable::c_reqTimeout;
-constexpr std::chrono::milliseconds NodeTable::c_bucketRefresh;
-constexpr std::chrono::milliseconds NodeTable::c_evictionCheckInterval;
+constexpr chrono::seconds DiscoveryDatagram::c_timeToLive;
+constexpr chrono::milliseconds NodeTable::c_reqTimeout;
+constexpr chrono::milliseconds NodeTable::c_bucketRefresh;
+constexpr chrono::milliseconds NodeTable::c_evictionCheckInterval;
 
-inline bool operator==(
-    std::weak_ptr<NodeEntry> const& _weak, std::shared_ptr<NodeEntry> const& _shared)
+inline bool operator==(weak_ptr<NodeEntry> const& _weak, shared_ptr<NodeEntry> const& _shared)
 {
     return !_weak.owner_before(_shared) && !_shared.owner_before(_weak);
 }
@@ -40,8 +39,8 @@ NodeTable::NodeTable(ba::io_service& _io, KeyPair const& _alias, NodeIPEndpoint 
         _io, static_cast<UDPSocketEvents&>(*this), (bi::udp::endpoint)m_hostNodeEndpoint)},
     m_requestTimeToLive{DiscoveryDatagram::c_timeToLive},
     m_allowLocalDiscovery{_allowLocalDiscovery},
-    m_discoveryTimer{std::make_shared<ba::deadline_timer>(_io)},
-    m_evictionTimer{std::make_shared<ba::deadline_timer>(_io)},
+    m_discoveryTimer{make_shared<ba::deadline_timer>(_io)},
+    m_evictionTimer{make_shared<ba::deadline_timer>(_io)},
     m_io{_io}
 {
     for (unsigned i = 0; i < s_bins; i++)
@@ -78,10 +77,10 @@ void NodeTable::start()
         doDiscovery();
         m_evictionTimer->expires_from_now(
             boost::posix_time::milliseconds(c_handleTimeoutsIntervalMs));
-        m_evictionTimer->async_wait(std::bind(&NodeTable::doHandleTimeouts, shared_from_this(),
-            std::placeholders::_1, m_evictionTimer));
+        m_evictionTimer->async_wait(bind(
+            &NodeTable::doHandleTimeouts, shared_from_this(), placeholders::_1, m_evictionTimer));
     }
-    catch (std::exception const& _e)
+    catch (exception const& _e)
     {
         cwarn << "Exception starting the discovery process: " << _e.what();
         cwarn << "Discovery disabled.";
@@ -275,8 +274,8 @@ void NodeTable::doDiscover(boost::system::error_code _ec, NodeID _node, unsigned
     }
 
     _timer->expires_from_now(boost::posix_time::milliseconds(c_reqTimeout.count() * 2));
-    _timer->async_wait(std::bind(&NodeTable::doDiscover, shared_from_this(), std::placeholders::_1,
-        _node, _round + 1, _tried, _timer));
+    _timer->async_wait(bind(&NodeTable::doDiscover, shared_from_this(), placeholders::_1, _node,
+        _round + 1, _tried, _timer));
 }
 
 vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID _target)
@@ -365,7 +364,7 @@ void NodeTable::evict(NodeEntry const& _leastSeen, shared_ptr<NodeEntry> _replac
         return;
 
     LOG(m_logger) << "Evicting node " << _leastSeen;
-    ping(_leastSeen, std::move(_replacement));
+    ping(_leastSeen, move(_replacement));
 
     if (m_nodeEventHandler)
         m_nodeEventHandler->appendEvent(_leastSeen.id, NodeEntryScheduledForEviction);
@@ -403,7 +402,7 @@ void NodeTable::noteActiveNode(shared_ptr<NodeEntry> _nodeEntry, bi::udp::endpoi
         auto& nodes = s.nodes;
 
         // check if the node is already in the bucket
-        auto it = std::find(nodes.begin(), nodes.end(), _nodeEntry);
+        auto it = find(nodes.begin(), nodes.end(), _nodeEntry);
         if (it != nodes.end())
         {
             // if it was in the bucket, move it to the last position
@@ -633,7 +632,7 @@ void NodeTable::onPacketReceived(
         if (sourceNodeEntry)
             noteActiveNode(move(sourceNodeEntry), _from);
     }
-    catch (std::exception const& _e)
+    catch (exception const& _e)
     {
         LOG(m_logger) << "Exception processing message from " << _from.address().to_string() << ":"
                       << _from.port() << ": " << _e.what();
@@ -653,13 +652,12 @@ void NodeTable::doDiscovery()
     LOG(m_logger) << "Queueing discovery algorithm run for random node id: " << randNodeId;
 
     m_discoveryTimer->expires_from_now(boost::posix_time::milliseconds(c_bucketRefresh.count()));
-    m_discoveryTimer->async_wait(
-        std::bind(&NodeTable::doDiscover, shared_from_this(), std::placeholders::_1, randNodeId,
-            0 /* round */, make_shared<set<shared_ptr<NodeEntry>>>(), m_discoveryTimer));
+    m_discoveryTimer->async_wait(bind(&NodeTable::doDiscover, shared_from_this(), placeholders::_1,
+        randNodeId, 0 /* round */, make_shared<set<shared_ptr<NodeEntry>>>(), m_discoveryTimer));
 }
 
 void NodeTable::doHandleTimeouts(
-    boost::system::error_code const& _ec, std::shared_ptr<ba::deadline_timer> _timer)
+    boost::system::error_code const& _ec, shared_ptr<ba::deadline_timer> _timer)
 {
     // We can't use m_logger yet because the NodeTable might have already been destroyed
     if (_ec == boost::asio::error::operation_aborted ||
@@ -701,7 +699,7 @@ void NodeTable::doHandleTimeouts(
 
     _timer->expires_from_now(boost::posix_time::milliseconds(c_handleTimeoutsIntervalMs));
     _timer->async_wait(
-        std::bind(&NodeTable::doHandleTimeouts, shared_from_this(), std::placeholders::_1, _timer));
+        bind(&NodeTable::doHandleTimeouts, shared_from_this(), placeholders::_1, _timer));
 }
 
 unique_ptr<DiscoveryDatagram> DiscoveryDatagram::interpretUDP(bi::udp::endpoint const& _from, bytesConstRef _packet)
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -33,31 +33,57 @@ inline bool operator==(
 
 NodeTable::NodeTable(ba::io_service& _io, KeyPair const& _alias, NodeIPEndpoint const& _endpoint,
     bool _enabled, bool _allowLocalDiscovery)
-  : m_hostNodeID(_alias.pub()),
-    m_hostNodeEndpoint(_endpoint),
-    m_secret(_alias.secret()),
-    m_socket(make_shared<NodeSocket>(
-        _io, static_cast<UDPSocketEvents&>(*this), (bi::udp::endpoint)m_hostNodeEndpoint)),
-    m_requestTimeToLive(DiscoveryDatagram::c_timeToLive),
-    m_allowLocalDiscovery(_allowLocalDiscovery),
-    m_timers(_io),
-    m_io(_io)
+  : m_hostNodeID{_alias.pub()},
+    m_hostNodeEndpoint{_endpoint},
+    m_secret{_alias.secret()},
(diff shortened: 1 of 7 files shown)


halfalicious:
The following test is failing: 
`
 network/net/evictionWithOldNodeDropped (Failed)`
 
Details:
```
 501/564 Test #507: network/net/evictionWithOldNodeDropped ................................................................................***Failed    0.40 sec
Running 1 test case...
Test Case "evictionWithOldNodeDropped": 
unknown location(0): fatal error: in "network/net/evictionWithOldNodeDropped": signal: SIGSEGV, si_code: 0 (memory access violation at address: 0x00000080)
/home/builder/project/test/unittests/libp2p/net.cpp(881): last checkpoint

```

halfalicious:
Rebased

halfalicious:
The following 2 tests are failing due to near-null AVs: 

```
Test Case "noteActiveNodeEvictsTheNodeWhenBucketIsFull": 
unknown location(0): fatal error: in "network/net/noteActiveNodeEvictsTheNodeWhenBucketIsFull": signal: SIGSEGV, si_code: 0 (memory access violation at address: 0x00000080)
/home/builder/project/test/unittests/libp2p/net.cpp(516): last checkpoint

Test Case "evictionWithOldNodeDropped": 
unknown location(0): fatal error: in "network/net/evictionWithOldNodeDropped": signal: SIGSEGV, si_code: 0 (memory access violation at address: 0x00000080)
/home/builder/project/test/unittests/libp2p/net.cpp(881): last checkpoint
```

halfalicious:
> The following 2 tests are failing due to near-null AVs:
> 
> ```
> Test Case "noteActiveNodeEvictsTheNodeWhenBucketIsFull": 
> unknown location(0): fatal error: in "network/net/noteActiveNodeEvictsTheNodeWhenBucketIsFull": signal: SIGSEGV, si_code: 0 (memory access violation at address: 0x00000080)
> /home/builder/project/test/unittests/libp2p/net.cpp(516): last checkpoint
> 
> Test Case "evictionWithOldNodeDropped": 
> unknown location(0): fatal error: in "network/net/evictionWithOldNodeDropped": signal: SIGSEGV, si_code: 0 (memory access violation at address: 0x00000080)
> /home/builder/project/test/unittests/libp2p/net.cpp(881): last checkpoint
> ```

Issues seem specific to clang...can't repro locally on Windows with full page heap enabled and the Linux GCC CircleCI run passed.

chfast:
Bad news, man. It crashes when testing in parallel. I could not reproduce it at first, because the test must run in parallel to cause the crash. In the end I compiled with clang + ASan + libc++ on linux.

The report:
```
=================================================================
==19042==ERROR: AddressSanitizer: heap-use-after-free on address 0x625000027b50 at pc 0x0000013ede33 bp 0x7fc60c40a4a0 sp 0x7fc60c40a498
READ of size 1 at 0x625000027b50 thread T5 (test)
    #0 0x13ede32 in std::__1::__atomic_base<bool, false>::load(std::__1::memory_order) const /usr/include/c++/v1/atomic:926:17
    #1 0x13ede32 in std::__1::__atomic_base<bool, false>::operator bool() const /usr/include/c++/v1/atomic:930
    #2 0x13ede32 in dev::p2p::DeadlineOps::reap()::$_0::operator()(std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&) const /home/chfast/Projects/ethereum/aleth/libp2p/Common.cpp:177
    #3 0x13ef307 in std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)>::operator()(std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&) const /usr/include/c++/v1/functional:1916:12
    #4 0x13ef307 in decltype(std::__1::forward<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)>&>(fp)(std::__1::forward<std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>&>(fp0), std::__1::forward<boost::system::error_code const&>(fp0))) std::__1::__invoke<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)>&, std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>&, boost::system::error_code const&>(std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)>&, std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>&, boost::system::error_code const&) /usr/include/c++/v1/type_traits:4482
    #5 0x13ef307 in std::__1::__bind_return<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)>, std::__1::tuple<std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> >, std::__1::tuple<boost::system::error_code const&>, __is_valid_bind_return<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)>, std::__1::tuple<std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> >, std::__1::tuple<boost::system::error_code const&> >::value>::type std::__1::__apply_functor<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)>, std::__1::tuple<std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> >, 0ul, 1ul, std::__1::tuple<boost::system::error_code const&> >(std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)>&, std::__1::tuple<std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> >&, std::__1::__tuple_indices<0ul, 1ul>, std::__1::tuple<boost::system::error_code const&>&&) /usr/include/c++/v1/functional:2219
    #6 0x13ef307 in std::__1::__bind_return<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)>, std::__1::tuple<std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> >, std::__1::tuple<boost::system::error_code const&>, __is_valid_bind_return<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)>, std::__1::tuple<std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> >, std::__1::tuple<boost::system::error_code const&> >::value>::type std::__1::__bind<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)> const&, std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> const&>::operator()<boost::system::error_code const&>(boost::system::error_code const&) /usr/include/c++/v1/functional:2252
    #7 0x13ef307 in boost::asio::detail::binder1<std::__1::__bind<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)> const&, std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> const&>, boost::system::error_code>::operator()() /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/asio/detail/bind_handler.hpp:47
    #8 0x13ef307 in void boost::asio::asio_handler_invoke<boost::asio::detail::binder1<std::__1::__bind<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)> const&, std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> const&>, boost::system::error_code> >(boost::asio::detail::binder1<std::__1::__bind<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)> const&, std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> const&>, boost::system::error_code>&, ...) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/asio/handler_invoke_hook.hpp:69
    #9 0x13ef307 in void boost_asio_handler_invoke_helpers::invoke<boost::asio::detail::binder1<std::__1::__bind<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)> const&, std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> const&>, boost::system::error_code>, std::__1::__bind<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)> const&, std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> const&> >(boost::asio::detail::binder1<std::__1::__bind<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)> const&, std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> const&>, boost::system::error_code>&, std::__1::__bind<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)> const&, std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> const&>&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/asio/detail/handler_invoke_helpers.hpp:37
    #10 0x13ef307 in boost::asio::detail::wait_handler<std::__1::__bind<std::__1::function<void (std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp> const&, boost::system::error_code const&)> const&, std::__1::shared_ptr<dev::p2p::DeadlineOps::DeadlineOp>, std::__1::placeholders::__ph<1> const&> >::do_complete(boost::asio::detail::task_io_service*, boost::asio::detail::task_io_service_operation*, boost::system::error_code const&, unsigned long) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/asio/detail/wait_handler.hpp:70
    #11 0xe748f0 in boost::asio::detail::task_io_service_operation::complete(boost::asio::detail::task_io_service&, boost::system::error_code const&, unsigned long) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/asio/detail/task_io_service_operation.hpp:38:5
    #12 0xe748f0 in boost::asio::detail::task_io_service::do_run_one(boost::asio::detail::scoped_lock<boost::asio::detail::posix_mutex>&, boost::asio::detail::task_io_service_thread_info&, boost::system::error_code const&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/asio/detail/impl/task_io_service.ipp:372
    #13 0xe74480 in boost::asio::detail::task_io_service::run(boost::system::error_code&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/asio/detail/impl/task_io_service.ipp:149:10
    #14 0xe74339 in boost::asio::io_service::run() /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/asio/impl/io_service.ipp:59:25
    #15 0x1226cb9 in dev::Worker::workLoop() /home/chfast/Projects/ethereum/aleth/libdevcore/Worker.cpp:140:3
    #16 0x1227511 in dev::Worker::startWorking()::$_0::operator()() const /home/chfast/Projects/ethereum/aleth/libdevcore/Worker.cpp:62:6
    #17 0x1226dcb in decltype(std::__1::forward<dev::Worker::startWorking()::$_0>(fp)()) std::__1::__invoke<dev::Worker::startWorking()::$_0>(dev::Worker::startWorking()::$_0&&) /usr/include/c++/v1/type_traits:4482:1
    #18 0x1226dcb in void std::__1::__thread_execute<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, dev::Worker::startWorking()::$_0>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, dev::Worker::startWorking()::$_0>&, std::__1::__tuple_indices<>) /usr/include/c++/v1/thread:342
    #19 0x1226dcb in void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, dev::Worker::startWorking()::$_0> >(void*) /usr/include/c++/v1/thread:352
    #20 0x7fc6113286da in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x76da)
    #21 0x7fc6103a088e in clone /build/glibc-OTsEL5/glibc-2.27/misc/../sysdeps/unix/sysv/linux/x86_64/clone.S:95

0x625000027b50 is located 8784 bytes inside of 9016-byte region [0x625000025900,0x625000027c38)
freed by thread T0 here:
    #0 0x9147b2 in operator delete(void*) (/home/chfast/Projects/ethereum/aleth/build/asan-libcxx/test/testeth+0x9147b2)
    #1 0xe64ef8 in std::__1::__shared_count::__release_shared() /usr/include/c++/v1/memory:3490:9
    #2 0xe64ef8 in std::__1::__shared_weak_count::__release_shared() /usr/include/c++/v1/memory:3532
    #3 0xe64ef8 in std::__1::shared_ptr<network::net::TestNodeTable>::~shared_ptr() /usr/include/c++/v1/memory:4468
    #4 0xe64ef8 in std::__1::enable_if<is_convertible<network::net::TestNodeTable*, network::net::TestNodeTable*>::value, void>::type std::__1::shared_ptr<network::net::TestNodeTable>::reset<network::net::TestNodeTable>(network::net::TestNodeTable*) /usr/include/c++/v1/memory:4616
    #5 0xe64ef8 in network::net::TestNodeTableHost::TestNodeTableHost(unsigned int, unsigned short) /home/chfast/Projects/ethereum/aleth/test/unittests/libp2p/net.cpp:234
    #6 0xe3b931 in network::net::noteActiveNodeEvictsTheNodeWhenBucketIsFull::test_method() /home/chfast/Projects/ethereum/aleth/test/unittests/libp2p/net.cpp:514:23
    #7 0xe3b06c in network::net::noteActiveNodeEvictsTheNodeWhenBucketIsFull_invoker() /home/chfast/Projects/ethereum/aleth/test/unittests/libp2p/net.cpp:512:1
    #8 0xb8689b in boost::function0<void>::operator()() const /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/function/function_template.hpp:759:14
    #9 0xb8689b in boost::detail::forward::operator()() /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:1300
    #10 0xb8689b in boost::detail::function::function_obj_invoker0<boost::detail::forward, int>::invoke(boost::detail::function::function_buffer&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/function/function_template.hpp:138
    #11 0xb45699 in boost::function0<int>::operator()() const /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/function/function_template.hpp:759:14
    #12 0xb45699 in int boost::detail::do_invoke<boost::shared_ptr<boost::detail::translator_holder_base>, boost::function<int ()> >(boost::shared_ptr<boost::detail::translator_holder_base> const&, boost::function<int ()> const&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:281
    #13 0xb45699 in boost::execution_monitor::catch_signals(boost::function<int ()> const&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:864
    #14 0xb4589a in boost::execution_monitor::execute(boost::function<int ()> const&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:1203:16
    #15 0xb3a915 in boost::execution_monitor::vexecute(boost::function<void ()> const&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:1309:5
    #16 0xb418c8 in boost::unit_test::unit_test_monitor_t::execute_and_translate(boost::function<void ()> const&, unsigned int) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/unit_test_monitor.ipp:46:9
    #17 0xb64c64 in boost::unit_test::framework::state::execute_test_tree(unsigned long, unsigned int, boost::unit_test::framework::state::random_generator_helper const*) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:771:44
    #18 0xb650bc in boost::unit_test::framework::state::execute_test_tree(unsigned long, unsigned int, boost::unit_test::framework::state::random_generator_helper const*) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:717:54
    #19 0xb650bc in boost::unit_test::framework::state::execute_test_tree(unsigned long, unsigned int, boost::unit_test::framework::state::random_generator_helper const*) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:717:54
    #20 0xb650bc in boost::unit_test::framework::state::execute_test_tree(unsigned long, unsigned int, boost::unit_test::framework::state::random_generator_helper const*) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:717:54
    #21 0xb3f6f1 in boost::unit_test::framework::run(unsigned long, bool) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:1577:29
    #22 0xb5141c in boost::unit_test::unit_test_main(boost::unit_test::test_suite* (*)(int, char**), int, char**) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/unit_test_main.ipp:231:9
    #23 0xb5e0ff in main /home/chfast/Projects/ethereum/aleth/test/tools/libtesteth/boostTest.cpp:139:14
    #24 0x7fc6102a0b96 in __libc_start_main /build/glibc-OTsEL5/glibc-2.27/csu/../csu/libc-start.c:310

previously allocated by thread T0 here:
    #0 0x913b72 in operator new(unsigned long) (/home/chfast/Projects/ethereum/aleth/build/asan-libcxx/test/testeth+0x913b72)
    #1 0xe64da9 in network::net::TestNodeTableHost::TestNodeTableHost(unsigned int, unsigned short) /home/chfast/Projects/ethereum/aleth/test/unittests/libp2p/net.cpp:235:17
    #2 0xe3b931 in network::net::noteActiveNodeEvictsTheNodeWhenBucketIsFull::test_method() /home/chfast/Projects/ethereum/aleth/test/unittests/libp2p/net.cpp:514:23
    #3 0xe3b06c in network::net::noteActiveNodeEvictsTheNodeWhenBucketIsFull_invoker() /home/chfast/Projects/ethereum/aleth/test/unittests/libp2p/net.cpp:512:1
    #4 0xb8689b in boost::function0<void>::operator()() const /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/function/function_template.hpp:759:14
    #5 0xb8689b in boost::detail::forward::operator()() /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:1300
    #6 0xb8689b in boost::detail::function::function_obj_invoker0<boost::detail::forward, int>::invoke(boost::detail::function::function_buffer&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/function/function_template.hpp:138
    #7 0xb45699 in boost::function0<int>::operator()() const /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/function/function_template.hpp:759:14
    #8 0xb45699 in int boost::detail::do_invoke<boost::shared_ptr<boost::detail::translator_holder_base>, boost::function<int ()> >(boost::shared_ptr<boost::detail::translator_holder_base> const&, boost::function<int ()> const&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:281
    #9 0xb45699 in boost::execution_monitor::catch_signals(boost::function<int ()> const&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:864
    #10 0xb4589a in boost::execution_monitor::execute(boost::function<int ()> const&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:1203:16
    #11 0xb3a915 in boost::execution_monitor::vexecute(boost::function<void ()> const&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:1309:5
    #12 0xb418c8 in boost::unit_test::unit_test_monitor_t::execute_and_translate(boost::function<void ()> const&, unsigned int) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/unit_test_monitor.ipp:46:9
    #13 0xb64c64 in boost::unit_test::framework::state::execute_test_tree(unsigned long, unsigned int, boost::unit_test::framework::state::random_generator_helper const*) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:771:44
    #14 0xb650bc in boost::unit_test::framework::state::execute_test_tree(unsigned long, unsigned int, boost::unit_test::framework::state::random_generator_helper const*) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:717:54
    #15 0xb650bc in boost::unit_test::framework::state::execute_test_tree(unsigned long, unsigned int, boost::unit_test::framework::state::random_generator_helper const*) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:717:54
    #16 0xb650bc in boost::unit_test::framework::state::execute_test_tree(unsigned long, unsigned int, boost::unit_test::framework::state::random_generator_helper const*) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:717:54
    #17 0xb3f6f1 in boost::unit_test::framework::run(unsigned long, bool) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:1577:29
    #18 0xb5141c in boost::unit_test::unit_test_main(boost::unit_test::test_suite* (*)(int, char**), int, char**) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/unit_test_main.ipp:231:9
    #19 0xb5e0ff in main /home/chfast/Projects/ethereum/aleth/test/tools/libtesteth/boostTest.cpp:139:14
    #20 0x7fc6102a0b96 in __libc_start_main /build/glibc-OTsEL5/glibc-2.27/csu/../csu/libc-start.c:310

Thread T5 (test) created by T0 here:
    #0 0x8ccc2d in pthread_create (/home/chfast/Projects/ethereum/aleth/build/asan-libcxx/test/testeth+0x8ccc2d)
    #1 0x1225d3f in std::__1::__libcpp_thread_create(unsigned long*, void* (*)(void*), void*) /usr/include/c++/v1/__threading_support:327:10
    #2 0x1225d3f in std::__1::thread::thread<dev::Worker::startWorking()::$_0, void>(dev::Worker::startWorking()::$_0&&) /usr/include/c++/v1/thread:368
    #3 0x1225735 in dev::Worker::startWorking() /home/chfast/Projects/ethereum/aleth/libdevcore/Worker.cpp:44:20
    #4 0xe3be72 in network::net::TestHost::start() /home/chfast/Projects/ethereum/aleth/test/unittests/libp2p/net.cpp:37:20
    #5 0xe3be72 in network::net::noteActiveNodeEvictsTheNodeWhenBucketIsFull::test_method() /home/chfast/Projects/ethereum/aleth/test/unittests/libp2p/net.cpp:524
    #6 0xe3b06c in network::net::noteActiveNodeEvictsTheNodeWhenBucketIsFull_invoker() /home/chfast/Projects/ethereum/aleth/test/unittests/libp2p/net.cpp:512:1
    #7 0xb8689b in boost::function0<void>::operator()() const /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/function/function_template.hpp:759:14
    #8 0xb8689b in boost::detail::forward::operator()() /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:1300
    #9 0xb8689b in boost::detail::function::function_obj_invoker0<boost::detail::forward, int>::invoke(boost::detail::function::function_buffer&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/function/function_template.hpp:138
    #10 0xb45699 in boost::function0<int>::operator()() const /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/function/function_template.hpp:759:14
    #11 0xb45699 in int boost::detail::do_invoke<boost::shared_ptr<boost::detail::translator_holder_base>, boost::function<int ()> >(boost::shared_ptr<boost::detail::translator_holder_base> const&, boost::function<int ()> const&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:281
    #12 0xb45699 in boost::execution_monitor::catch_signals(boost::function<int ()> const&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:864
    #13 0xb4589a in boost::execution_monitor::execute(boost::function<int ()> const&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:1203:16
    #14 0xb3a915 in boost::execution_monitor::vexecute(boost::function<void ()> const&) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/execution_monitor.ipp:1309:5
    #15 0xb418c8 in boost::unit_test::unit_test_monitor_t::execute_and_translate(boost::function<void ()> const&, unsigned int) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/unit_test_monitor.ipp:46:9
    #16 0xb64c64 in boost::unit_test::framework::state::execute_test_tree(unsigned long, unsigned int, boost::unit_test::framework::state::random_generator_helper const*) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:771:44
    #17 0xb650bc in boost::unit_test::framework::state::execute_test_tree(unsigned long, unsigned int, boost::unit_test::framework::state::random_generator_helper const*) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:717:54
    #18 0xb650bc in boost::unit_test::framework::state::execute_test_tree(unsigned long, unsigned int, boost::unit_test::framework::state::random_generator_helper const*) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:717:54
    #19 0xb650bc in boost::unit_test::framework::state::execute_test_tree(unsigned long, unsigned int, boost::unit_test::framework::state::random_generator_helper const*) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:717:54
    #20 0xb3f6f1 in boost::unit_test::framework::run(unsigned long, bool) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/framework.ipp:1577:29
    #21 0xb5141c in boost::unit_test::unit_test_main(boost::unit_test::test_suite* (*)(int, char**), int, char**) /home/chfast/.hunter/_Base/c4cfcc0/59044ab/ef3c3eb/Install/include/boost/test/impl/unit_test_main.ipp:231:9
    #22 0xb5e0ff in main /home/chfast/Projects/ethereum/aleth/test/tools/libtesteth/boostTest.cpp:139:14
    #23 0x7fc6102a0b96 in __libc_start_main /build/glibc-OTsEL5/glibc-2.27/csu/../csu/libc-start.c:310

SUMMARY: AddressSanitizer: heap-use-after-free /usr/include/c++/v1/atomic:926:17 in std::__1::__atomic_base<bool, false>::load(std::__1::memory_order) const
Shadow bytes around the buggy address:
  0x0c4a7fffcf10: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c4a7fffcf20: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c4a7fffcf30: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c4a7fffcf40: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c4a7fffcf50: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
=>0x0c4a7fffcf60: fd fd fd fd fd fd fd fd fd fd[fd]fd fd fd fd fd
  0x0c4a7fffcf70: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c4a7fffcf80: fd fd fd fd fd fd fd fa fa fa fa fa fa fa fa fa
  0x0c4a7fffcf90: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c4a7fffcfa0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c4a7fffcfb0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==19042==ABORTING
```

chfast:
It looks like the shared pointer is deleted, but some still has the pointer to the object.

halfalicious:
> eadlineOps::DeadlineOp> const&, b

Ah, I think the problem is in the DeadlineOps::reap() handler - it checks the error code and m_stopped before deciding to proceed, but DeadlineLoops dtor doesn't cancel timers before deleting them which means that the error code won't be set and attempting to check m_stopped will trigger an AV.

chfast:
> Ah, I think the problem is in the DeadlineOps::reap() handler - it checks the error code and m_stopped before deciding to proceed, but DeadlineLoops dtor doesn't cancel timers before deleting them which means that the error code won't be set and attempting to check m_stopped will trigger an AV.

Can't you just check the error code first?

halfalicious:
> 
> 
> > Ah, I think the problem is in the DeadlineOps::reap() handler - it checks the error code and m_stopped before deciding to proceed, but DeadlineLoops dtor doesn't cancel timers before deleting them which means that the error code won't be set and attempting to check m_stopped will trigger an AV.
> 
> Can't you just check the error code first?

We do but it doesn't indicate operation aborted since the timer(s) haven't been cancelled + are still alive because of the shared_ptr accepted by the handler.

halfalicious:
I've rebased my changes after merging @chfast 's changes which queue pings via io_service::post and scoped them down to just adding the dedicated timers for the discovery and eviction processes. I've also removed DeadlineOps since we no longer need it.

I'm going to test my changes by running discovery overnight and verifying no cancellation but I don't envision any problems.

libp2p/NodeTable.cpp
```diff
@@ -233,22 +234,17 @@ void NodeTable::doDiscover(NodeID _node, unsigned _round, shared_ptr<set<shared_
         return;
     }
 
-    m_timers.schedule(c_reqTimeout.count() * 2, [this, _node, _round, _tried](boost::system::error_code const& _ec)
-    {
+    m_discoveryTimer.expires_from_now(boost::posix_time::milliseconds(c_reqTimeout.count() * 2));
+    m_discoveryTimer.async_wait([this, _node, _round, _tried](
+                                    boost::system::error_code const& _ec) {
         if (_ec)
+        {
```
chfast: Only checking `_ec` is not enough. This will handle only the case where the timer was cancelled before it expired. But there is also the case where the timer expired and handler in already in the completion queue. In this case there is not error code. So `doDiscover()` will be called again and new timer scheduled...

To handle this case there are two requirements:
- The NodeTable must still exist, what's why using shared_ptr is handy here.
- We have to check if NodeTable is in valid state before doing `doDiscover()`. We can use any `m_running` flag, but a smart way is to set the timer's `expires_at` to some arbitrary value like `time_point::min()`.

See https://stackoverflow.com/a/43169596/725174.

chfast: Quick fix would be to handle timers by shared_ptrs, i.e. 
```
std::shared_ptr<ba::deadline_timer> m_discoveryTimer;
```
and pass it also to the handler so it can access the timer safely and check if `expires_at` is not the `min` value.

halfalicious: > See https://stackoverflow.com/a/43169596/725174.

Thanks, that's an awesome link 😄 I didn't realize that calling `cancel `after a timer's handler has already been queued wouldn't affect its error code.

I'll make the `time_point::min` + `shared_ptr` change, though I don't think they're technically needed for a couple of reasons:
*  The node table lifetime (and therefore the timer lifetimes) is tied to the host's lifetime, and when the host is destroyed then there's no more ioservice work loop meaning that the node table is guaranteed to exist if the timer handlers are executing.
* I don't think `doDiscover` would be called if its handler is executed after timer cancellation because we cancel the timers in `NodeTable::stop()` and we also close the UDP socket, and the first thing that `doDiscover` does is check if the UDP socket is still open...so `doDiscover ` should immediately exit if it executes after the timers have been cancelled. 

The eviction timer handler however would still execute the eviction code...I took a quick look at it and I don't think we'd hit any major issues as long as the node table hasn't been destroyed but I'm not completely sure.

I do think it's good practice to keep the timers alive through `shared_ptr `rather than relying on the timers' owner's lifetime though (since we may refactor things later).

halfalicious: > 
> 
> This looks good to me, but there are some test failures that must be investigated.
> 
> Moreover, check out this tutorial about triggering timers periodically (we are close to this, but maybe you can spot some small differences): https://www.boost.org/doc/libs/1_69_0/doc/html/boost_asio/tutorial/tuttimer3.html
> 
> Also they are using `steady_timer` instead of `deadline_timer`. The difference is only in the clock used. We don't need high precision here, so maybe one is better than the other for our case.

steady_clock might be nice since we'd be able to switch to std::chrono for timers and it wouldn't be impacted like deadline timers if someone changes the system clock while Aleth is running. However, I need to read a little more about its precision before deciding if we should move over.

halfalicious: Steady-clock appears to provide 1/100 microsecond precision on both Windows and Linux so I think we can use it, I'd like to make this change in a future PR though.

halfalicious:
One test is failing in linux-clang6: 

`510 - network/net/findNodeIsSentAfterPong (Failed)`

I think it's just an unreliable test, I'll file a new issue for tracking purposes. 

No tests are failing in the appveyor config, but I'm seeing the following error which is causing the config to fail:

```
*** No errors detected
Collecting artifacts...
Found artifact 'build\aleth.zip' matching 'build\aleth.zip' path
Uploading artifacts...
Error uploading artifact the storage: Unable to connect to the remote server
```

I recall seeing this before... @chfast : Is this a known issue?

halfalicious:
> 
> 
> One test is failing in linux-clang6:
> 
> `510 - network/net/findNodeIsSentAfterPong (Failed)`
> 
> I think it's just an unreliable test, I'll file a new issue for tracking purposes.
> 
> No tests are failing in the appveyor config, but I'm seeing the following error which is causing the config to fail:
> 
> ```
> *** No errors detected
> Collecting artifacts...
> Found artifact 'build\aleth.zip' matching 'build\aleth.zip' path
> Uploading artifacts...
> Error uploading artifact the storage: Unable to connect to the remote server
> ```
> 
> I recall seeing this before... @chfast : Is this a known issue?

Looks like some more tests are failing now...I'll investigate but I suspect they are unreliable tests.

halfalicious:
> 
> 
> > One test is failing in linux-clang6:
> > `510 - network/net/findNodeIsSentAfterPong (Failed)`
> > I think it's just an unreliable test, I'll file a new issue for tracking purposes.
> > No tests are failing in the appveyor config, but I'm seeing the following error which is causing the config to fail:
> > ```
> > *** No errors detected
> > Collecting artifacts...
> > Found artifact 'build\aleth.zip' matching 'build\aleth.zip' path
> > Uploading artifacts...
> > Error uploading artifact the storage: Unable to connect to the remote server
> > ```
> > I recall seeing this before... @chfast : Is this a known issue?
> 
> Looks like some more tests are failing now...I'll investigate but I suspect they are unreliable tests.

I've fixed the `findNodeIsSentAfterPong `failure - I think the test was fundamentally unreliable but my changes increased the possibility of it failing because they decreased the amount of time between starting the network threads / discovery process in the 2 node table instances. Please see my commit notes for details.

I still need to investigate the other failure (`evictionWithOldNodeAnswering`) - I can't repro it locally or in CircleCI via SSH, even when running the tests in parallel via CTest. I seem to hit it pretty regularly in CircleCI runs kicked off after each PR update though, I think what I need to do is to create a commit with testeth's verbosity set to 4 in the CircleCI config and push those changes so I can see what's causing the failure. @chfast : Do you know how to do this? I had some problems doing this before due to escape issues.

halfalicious:
These network test failures still need to be investigated and addressed: 
* evictionWithOldNodeAnswering
* invalidPong

libp2p/NodeTable.cpp
```diff
@@ -20,44 +20,68 @@ constexpr unsigned c_handleTimeoutsIntervalMs = 5000;
...
+            m_socket->connect();
+        }
+        catch (exception const& _e)
+        {
+            cwarn << "Exception connecting NodeTable socket: " << _e.what();
+            _enabled = false;
+        }
+    }
+    else cwarn << "\"enabled\" parameter is false";
```
chfast: You have similar logs twice. Also use quick return pattern as previously.

halfalicious:
> > I think what I need to do is to create a commit with testeth's verbosity set to 4 in the CircleCI config and push those changes so I can see what's causing the failure.
> 
> If you want it for testeth directly try `testeth -- --verbosity 4`. If for ctest, use CMake option `"-DTESTETH_ARGS=--verbosity 4"` (you can do it also after the build is finished via SSH, it will only change the ctest config so no rebuild is required).

I think I need to create a new (temporary) commit with testeth’s verbosity set to 4 because I can’t repro some of the failures when running the tests manually via ctest or testeth. I remember having a problem last time with the escaping syntax in the circleci config, I’ll take another crack at it.

libp2p/NodeTable.cpp
halfalicious: Yea I wanted to avoid redundant logging using the _enabled flag I’ll switch back to early return.

halfalicious:
The AppVeyor failure appears to be due to a timeout...the build took ~47 minutes to complete. There were also a bunch of build warnings, I've filed a new issue to track them: #5515

halfalicious:
I don't see anymore failing tests in CircleCI...I also tried re-running the network tests a few times on a couple of different CircleCI configs when connected via SSH and wasn't able to hit any failures. It's possible that some intermittent failures might still pop up in the future but I think that we can address them then. 

I'm going to run a build of aleth-bootnode which has my changes overnight to verify that the discovery process doesn't get cancelled.

libp2p/NodeTable.h
```diff
@@ -306,7 +320,13 @@ class NodeTable : UDPSocketEvents
 
     bool m_allowLocalDiscovery;                                     ///< Allow nodes with local addresses to be included in the discovery process
 
-    DeadlineOps m_timers; ///< this should be the last member - it must be destroyed first
+    // timers are shared_ptr so we can keep them alive if handlers are executed after the node table
+    // has been destroyed so we can check for timer cancellation. Note that this is technically not
```
gumb0: Well yeah I think I got rid of this `enable_shared_from_this` approach a while ago, because we clearly control the life time of the `NodeTable` from the level above (`Host`) and should be able to cancel the timers when it's destructed.
This comeback doesn't seem to make the code simpler...

libp2p/NodeTable.h
```diff
@@ -114,10 +114,20 @@ class NodeTable : UDPSocketEvents
     /// Returns distance based on xor metric two node ids. Used by NodeEntry and NodeTable.
     static int distance(NodeID const& _a, NodeID const& _b) { u256 d = sha3(_a) ^ sha3(_b); unsigned ret; for (ret = 0; d >>= 1; ++ret) {}; return ret; }
 
+    /// Open the udp socket and schedule execution of the discovery algorithm
+    void start();
```
gumb0: Why is this needed, can't we do it in constructor as earlier?

halfalicious: We can’t call shared_from_this inside of the ctor

libp2p/NodeTable.h
```diff
@@ -275,6 +288,7 @@ class NodeTable : UDPSocketEvents
         return dev::p2p::isAllowedEndpoint(m_allowLocalDiscovery, _endpointToCheck);
     }
 
+    bool m_enabled;                                                 /// Is discovery enabled?
```
gumb0: this doesn't look used?

halfalicious: Ah yes artifact of an earlier change that I forgot to remove. Will remove it

libp2p/NodeTable.cpp
```diff
@@ -20,45 +20,70 @@ constexpr unsigned c_handleTimeoutsIntervalMs = 5000;
...
+    if (!m_socket->isOpen())
+    {
+        cwarn << "NodeTable socket isn't open. Discovery is disabled";
+        return;
+    }
+
+    try
+    {
+        doDiscovery();
+        m_evictionTimer->expires_from_now(
```
gumb0: Can we maybe make `doHandleTimeouts()` without parameters out of these two lines of initialization of `m_evictionTimer`? Then it will look more nicely symmetric

libp2p/NodeTable.cpp
```diff
@@ -20,45 +20,70 @@ constexpr unsigned c_handleTimeoutsIntervalMs = 5000;
...
+            boost::posix_time::milliseconds(c_handleTimeoutsIntervalMs));
+        m_evictionTimer->async_wait(bind(
```
gumb0: Better use lambdas instead of `bind`.
Lambda syntax is more common, and also `bind` can have some obscure problems

libp2p/NodeTable.cpp
```diff
@@ -190,10 +215,25 @@ shared_ptr<NodeEntry> NodeTable::nodeEntry(NodeID const& _id)
...
 
-void NodeTable::doDiscover(NodeID _node, unsigned _round, shared_ptr<set<shared_ptr<NodeEntry>>> _tried)
+void NodeTable::doDiscover(boost::system::error_code _ec, NodeID _node, unsigned _round,
+    shared_ptr<set<shared_ptr<NodeEntry>>> _tried, shared_ptr<ba::deadline_timer> _timer)
 {
     // NOTE: ONLY called by doDiscovery!
 
+    // We can't use m_logger yet since the node table might have already been destroyed
+    if (_ec == boost::asio::error::operation_aborted ||
+        _timer->expires_at() == boost::posix_time::min_date_time)
```
gumb0: Could you explain when this second condition can happen?

libp2p/NodeTable.cpp
```diff
@@ -233,25 +273,9 @@ void NodeTable::doDiscover(NodeID _node, unsigned _round, shared_ptr<set<shared_
...
-
-        // error::operation_aborted means that the timer was probably aborted.
-        // It usually happens when "this" object is deallocated, in which case
-        // subsequent call to doDiscover() would cause a crash. We can not rely on
-        // m_timers.isStopped(), because "this" pointer was captured by the lambda,
-        // and therefore, in case of deallocation m_timers object no longer exists.
-
-        doDiscover(_node, _round + 1, _tried);
-    });
+    _timer->expires_from_now(boost::posix_time::milliseconds(c_reqTimeout.count() * 2));
```
gumb0: Maybe make a constant out of `boost::posix_time::milliseconds(c_reqTimeout.count() * 2)`

libp2p/NodeTable.cpp
halfalicious: We set the min time instead of cancelling to get around the race condition, look in the node table destructor.

libp2p/NodeTable.cpp
```diff
@@ -233,25 +273,9 @@ void NodeTable::doDiscover(NodeID _node, unsigned _round, shared_ptr<set<shared_
...
+    _timer->async_wait(bind(&NodeTable::doDiscover, shared_from_this(), placeholders::_1, _node,
```
gumb0: lambda instead of `bind`

libp2p/NodeTable.cpp
halfalicious: Ah ok, went with bind since the boost examples i looked at used it and the syntax is a little more intuitive to me. For lambda should we just capture the shared_from_this by value?

halfalicious:
> 
> 
> I don't see anymore failing tests in CircleCI...I also tried re-running the network tests a few times on a couple of different CircleCI configs when connected via SSH and wasn't able to hit any failures. It's possible that some intermittent failures might still pop up in the future but I think that we can address them then.
> 
> I'm going to run a build of aleth-bootnode which has my changes overnight to verify that the discovery process doesn't get cancelled.

Discovery process is still running this morning after running all night so it's safe to say the bug is fixed.

libp2p/NodeTable.h
```diff
@@ -114,10 +114,20 @@ class NodeTable : UDPSocketEvents
...
+    /// Open the udp socket and schedule execution of the discovery algorithm
+    void start();
+
     void stop()
     {
-        m_socket->disconnect();
-        m_timers.stop();
+        if (m_socket->isOpen())
+        {
+            // We "cancel" the timers by setting min_date_time rather than calling cancel() because
```
gumb0: Is it necessary to use this magic value for timer expiration? Would it be clearer if we instead just used a special flag like `m_stopped` to tell handlers that we're cancelling everything?

(Again I'm still not completely sure that this complication is necessary - why couldn't `Host` make sure that no handlers are being called (i.e. not `io_service::run()` is called) after `NodeTable` and its timers are destroyed. I beleive this is `Host`'s job)

gumb0: Yeah I think the flag would make things cleaner here - it would be similar to what `DeadlineOps::m_stopped` flag did before

halfalicious: I actually like the magic value for timer expiration more because it's more self-contained (and I've commented its purpose so someone reading the code would know what it's used for) - we don't have to rely on an additional variable with potentially a different lifetime.

halfalicious:
I think I need to add a mutex for each timer since we access them on different threads e.g. NodeTable::stop() is called from the application thread and cancels the timers while timer handlers are executing on the network thread and checking the timer expiration time.

libp2p/NodeTable.cpp
```diff
@@ -621,36 +645,52 @@ void NodeTable::onPacketReceived(
...
-    m_timers.schedule(c_bucketRefresh.count(), [this](boost::system::error_code const& _ec)
-    {
-        if (_ec)
-            // we can't use m_logger here, because captured this might be already destroyed
-            clog(VerbosityDebug, "discov")
-                << "Discovery timer was probably cancelled: " << _ec.value() << " "
-                << _ec.message();
-
-        if (_ec.value() == boost::asio::error::operation_aborted || m_timers.isStopped())
+    m_discoveryTimer.expires_from_now(boost::posix_time::milliseconds(c_bucketRefresh.count()));
```
gumb0: I think you can define the constant now as `boost::posix_time::milliseconds`

libp2p/NodeTable.h
```diff
@@ -195,6 +211,8 @@ class NodeTable : UDPSocketEvents
     static constexpr std::chrono::milliseconds c_evictionCheckInterval{75};
     /// How long to wait for requests (evict, find iterations).
     static constexpr std::chrono::milliseconds c_reqTimeout{300};
+    /// How long to wait before starting a new discovery round
+    static constexpr std::chrono::milliseconds c_discoveryRoundIntervalMs{c_reqTimeout * 2};
```
gumb0: Better define it as `boost::posix_time::milliseconds`, because that's what's actually needed

chfast: I'm not paying attention to this PR any more, hopefully you will be able to sort it out.
In this case, there is a way to define boost timer with `std::chrono`. You need to use the templeted version `basic_deadline_timer`.

gumb0: If these intervals are used only for boost timers, maybe it's simpler just to use `boost::posix_times`

libp2p/NodeTable.cpp
halfalicious: I'm going to use steady_timer so we can use std::chrono time values

halfalicious:
Looks like the linux-clang failure is unrelated to my changes, all unit tests passed but the sstore test appears to have timed out.

halfalicious:
cc @chfast / @gumb0

libp2p/NodeTable.cpp
```diff
@@ -621,36 +630,56 @@ void NodeTable::onPacketReceived(
...
 void NodeTable::doDiscovery()
 {
-    m_timers.schedule(c_bucketRefresh.count(), [this](boost::system::error_code const& _ec)
-    {
-        if (_ec)
-            // we can't use m_logger here, because captured this might be already destroyed
+    m_discoveryTimer->expires_from_now(c_bucketRefreshMs);
+    auto discoveryTimer{m_discoveryTimer};
+    m_discoveryTimer->async_wait([this, discoveryTimer](boost::system::error_code const& _ec) {
+        // We can't use m_logger if an error occurred because captured this might be already
```
gumb0: I'd move this comment inside brackets, before `clog(...)<<`

libp2p/NodeTable.cpp
```diff
@@ -233,25 +237,27 @@ void NodeTable::doDiscover(NodeID _node, unsigned _round, shared_ptr<set<shared_
...
-        // error::operation_aborted means that the timer was probably aborted.
-        // It usually happens when "this" object is deallocated, in which case
-        // subsequent call to doDiscover() would cause a crash. We can not rely on
-        // m_timers.isStopped(), because "this" pointer was captured by the lambda,
-        // and therefore, in case of deallocation m_timers object no longer exists.
+    m_discoveryTimer->expires_from_now(c_discoveryRoundIntervalMs);
+    auto discoveryTimer{m_discoveryTimer};
+    m_discoveryTimer->async_wait(
+        [this, discoveryTimer, _node, _round, _tried](boost::system::error_code const& _ec) {
+            // We can't use m_logger here if there's an error because captured this might already be
```
gumb0: here too, move comment

libp2p/NodeTable.cpp
```diff
@@ -621,36 +630,56 @@ void NodeTable::onPacketReceived(
...
         crypto::Nonce::get().ref().copyTo(randNodeId.ref().cropped(h256::size, h256::size));
-        LOG(m_logger) << "Random discovery: " << randNodeId;
-        doDiscover(randNodeId, 0, make_shared<set<shared_ptr<NodeEntry>>>());
+        LOG(m_logger) << "Starting discovery algorithm run for random node id: " << randNodeId;
+        doDiscoveryRound(randNodeId, 0 /* round */, make_shared<set<shared_ptr<NodeEntry>>>());
     });
 }
 
-void NodeTable::doHandleTimeouts()
+void NodeTable::doProcessEvictions()
```
gumb0: Not sure about the naming, because it handles not only evictions from the buckets, but also just initial Pings timed out. I'd leave it with "timeouts" in the name

libp2p/NodeTable.cpp
```diff
@@ -621,36 +630,56 @@ void NodeTable::onPacketReceived(
...
 {
-    m_timers.schedule(c_handleTimeoutsIntervalMs, [this](boost::system::error_code const& _ec) {
-        if ((_ec && _ec.value() == boost::asio::error::operation_aborted) || m_timers.isStopped())
+    m_evictionTimer->expires_from_now(c_processEvictionsIntervalMs);
+    auto evictionTimer{m_evictionTimer};
+    m_evictionTimer->async_wait([this, evictionTimer](boost::system::error_code const& _ec) {
+        // We can't use m_logger if an error occurred because captured this might be already
```
gumb0: Also move comment

libp2p/NodeTable.cpp
halfalicious: So I purposely put this comment above the "if" case since it applies to both the "if" and "else if" and I didn't want to include the same comment in both cases.

libp2p/NodeTable.cpp
halfalicious: Ah, good point, I didn't realize that.

halfalicious:
Squashed a bunch of commits and addressed PR feedback

halfalicious:
@chfast : I'll merge tomorrow AM unless you have any objections.

PR #5503, opened 2019-02-25
Node logging

chfast:


libp2p/Common.cpp
```diff
@@ -248,9 +248,9 @@ ostream& operator<<(ostream& _out, NodeIPEndpoint const& _ep)
     return _out;
 }
 
-boost::log::formatting_ostream& operator<<(boost::log::formatting_ostream& _log, const Node& _node)
+ostream& operator<<(ostream& _out, const Node& _node)
```
halfalicious: Why the move from the boost logging ostream to std::ostream, don't we only log Node instances using boost loggers? Is the only benefit not having to do static_cast when logging NodeEntry instances?

libp2p/Common.h
```diff
@@ -254,26 +254,24 @@ class DeadlineOps
...
         DeadlineOp(ba::io_service& _io, unsigned _msInFuture, std::function<void(boost::system::error_code const&)> const& _f): m_timer(new ba::deadline_timer(_io)) { m_timer->expires_from_now(boost::posix_time::milliseconds(_msInFuture)); m_timer->async_wait(_f); }
         ~DeadlineOp() { if (m_timer) m_timer->cancel(); }
-        
-        DeadlineOp(DeadlineOp&& _s): m_timer(_s.m_timer.release()) {}
-        DeadlineOp& operator=(DeadlineOp&& _s)
-        {
-            assert(&_s != this);
 
-            m_timer.reset(_s.m_timer.release());
+        DeadlineOp(DeadlineOp&& _s) noexcept : m_timer(std::move(_s.m_timer)) {}
```
halfalicious: Why do we need noexcept here?

libp2p/Common.h
```diff
@@ -254,26 +254,24 @@ class DeadlineOps
...
+
     private:
         std::unique_ptr<ba::deadline_timer> m_timer;
         Mutex x_timer;
     };
-    
+
 public:
-    DeadlineOps(ba::io_service& _io, unsigned _reapIntervalMs = 100): m_io(_io), m_reapIntervalMs(_reapIntervalMs), m_stopped(false) { reap(); }
+    explicit DeadlineOps(ba::io_service& _io, unsigned _reapIntervalMs = 100): m_io(_io), m_reapIntervalMs(_reapIntervalMs), m_stopped(false) { reap(); }
```
halfalicious: Just to verify I understand this correctly, is the benefit of "explicit" here that we get a more informative compile-time error if the default constructor is invoked (since without explicit I think invoking the default constructor would trigger a compiler error anyway since DeadlineOps holds a reference to the IO service which needs to be initialized)

halfalicious:
Looks like this test is timing out, presumably it's just unreliable and not related to your changes since I didn't see anything in your changes which would cause this: 
`
500 - network/net/invalidPong (Timeout)
`
Can you open a new issue so one of us can investigate later?

libp2p/Common.cpp
halfalicious: > 
> 
> Why the move from the boost logging ostream to std::ostream, don't we only log Node instances using boost loggers? Is the only benefit not having to do static_cast when logging NodeEntry instances? And out of curiosity, why don't we have to do the static_cast when using std::ostream?

chfast: Overloading `boost::log::formatting_ostream` is seriously f*cked up. The Node derived types were not accepted and there are issues with `&` and `const&`. This one works for all 4 cases.

libp2p/Common.h
chfast: Move constructors are much more powerful with `noexcept`. Clang-tidy suggests this change. https://en.cppreference.com/w/cpp/utility/move_if_noexcept.

libp2p/Common.h
chfast: Yes, this blocks implicit conversion from `io_service`.

PR #5504, opened 2019-02-26
Use io_service::post()

chfast:


halfalicious:
Taking over for @chfast

halfalicious:
The test failures are intermittent and I don't see how they could be related to these changes - I'm chalking the failures up to flaky tests and am going to merge these changes, we can investigate if the test failures pop up again.

PR #5505, opened 2019-02-26
Allow building with libc++ on linux

chfast:


libdevcore/Common.cpp
```diff
@@ -113,10 +113,8 @@ that users do not need to install language packs for their OS.
 void setDefaultOrCLocale()
 {
 #if __unix__
-    if (!std::setlocale(LC_ALL, ""))
```
halfalicious: Why is this change needed?

cmake/EthCompilerSettings.cmake
```diff
@@ -41,15 +41,6 @@ if (("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MA
     set(CMAKE_CXX_FLAGS_RELEASE        "-O3 -DNDEBUG")
     set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O2 -g")
 
-    option(USE_LD_GOLD "Use GNU gold linker" ON)
```
halfalicious: Out of curiosity, which linker will we use for clang builds if we don't use the gold linker?

chfast: The GNU `ld`. Also `lld` from LLVM can be used (is much faster, I'm using it for building LLVM itself where linking time is huge). In the same way: `-fuse-ld=lld`. But only clang supports it (GCC not).

As I explained in the commit message, Google has switched to use LLD, and they are not interested in maintaining ld.gold. Fedora is considering moving ld.gold out of the main package repo. And I have linking error when using clang-8. There must be new "linking features" that clang-8 depends on.

libdevcore/Common.cpp
chfast: In libc++ the `setlocale` is not in the `std` namespace.

cmake/EthCompilerSettings.cmake
halfalicious: Thanks for the clarification!

PR #5512, opened 2019-03-04
Fix TransactionIsUnsigned on call with default from parameter

DimaStebaev:
As written in https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_call there is an optional parameter "from". But if you make call without "from" and there are no keys in local wallet TransactionIsUnsigned exception is thrown because default sender value (zero address) in transaction confuse with no value.

I fixed this issue and added test to reproduce it.


DimaStebaev:
Yes. Sure. I am sorry for that.

halfalicious:
Thank you very much for this work!

PR #5514, opened 2019-03-06
more debug info on BlockGasLimitReached exception log

winsvega:
when reading block gas limit reached log its unclear what are the expected values and why it is reached. 
more debug info should help with debugging such exceptions

### DIFF
--- a/libweb3jsonrpc/Eth.cpp
+++ b/libweb3jsonrpc/Eth.cpp
@@ -739,10 +739,12 @@ string dev::rpc::exceptionToErrorMessage()
 	{
 		ret = "Transaction gas amount is less than the intrinsic gas amount for this transaction type.";
 	}
-	catch (BlockGasLimitReached const&)
-	{
-		ret = "Block gas limit reached.";
-	}
+    catch (BlockGasLimitReached const& _ex)
+    {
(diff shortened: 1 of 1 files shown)


gumb0:
What kind of error does this return in the end?

In general, I think error messages of RPC methods should be more user-friendly, without messy internals like boost.exception dumps

winsvega:
then perhaps it could be extended by verbosity level of debug? 
like on first level it is just a notice message and on max level it is full exception info.

gumb0:
I don't think log verbosity should control RPC output.

gumb0:
Maybe output your message to log - and then look into aleth's log when you want to see debug info related to RPC call.

halfalicious:
> What kind of error does this return in the end?
> 
> In general, I think error messages of RPC methods should be more user-friendly, without messy internals like boost.exception dumps

Good point I agree with this since these errors will be displayed in response to functionally directly invoked by the end user.

winsvega:
```
Maybe output your message to log - and then look into aleth's log when you want to see debug info related to RPC call.
```
that is the issue. the log contain not enough info

gumb0:
I'm saying add a message to the log, but not to the RPC output

PR #5519, opened 2019-03-13
Handle receiving Discovery packets with changed endpoint

gumb0:
Fixes https://github.com/ethereum/aleth/issues/5455

### DIFF
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -502,6 +502,7 @@ void NodeTable::onPacketReceived(
                 // create or update nodeEntry with new Pong received time
                 DEV_GUARDED(x_nodes)
                 {
+                    auto const& sourceId = pong.sourceid;
                     auto it = m_allNodes.find(sourceId);
                     if (it == m_allNodes.end())
                         sourceNodeEntry = make_shared<NodeEntry>(m_hostNodeID, sourceId,
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -83,7 +83,7 @@ bool NodeTable::addNode(Node const& _node)
     DEV_GUARDED(x_nodes)
     {
         auto const it = m_allNodes.find(_node.id);
-        needToPing = (it == m_allNodes.end() || it->second->endpoint != _node.endpoint ||
+        needToPing = (it == m_allNodes.end() || it->second->endpoint() != _node.endpoint ||
                       !it->second->hasValidEndpointProof());
     }
 
@@ -365,7 +365,7 @@ void NodeTable::noteActiveNode(shared_ptr<NodeEntry> _nodeEntry)
         LOG(m_logger) << "Skipping making self active.";
         return;
     }
-    if (!isAllowedEndpoint(_nodeEntry->endpoint))
+    if (!isAllowedEndpoint(_nodeEntry->endpoint()))
     {
         LOG(m_logger) << "Skipping making node with unallowed endpoint active. Node "
                       << _nodeEntry->node;
@@ -375,7 +375,7 @@ void NodeTable::noteActiveNode(shared_ptr<NodeEntry> _nodeEntry)
     if (!_nodeEntry->hasValidEndpointProof())
         return;
 
-    LOG(m_logger) << "Active node " << *_nodeEntry;
+    LOG(m_logger) << "Active node " << _nodeEntry->node;
 
     shared_ptr<NodeEntry> nodeToEvict;
     {
@@ -510,8 +510,8 @@ void NodeTable::onPacketReceived(
                         sourceNodeEntry->lastPongReceivedTime =
                             RLPXDatagramFace::secondsSinceEpoch();
 
-                        if (sourceNodeEntry->endpoint != _from)
-                            sourceNodeEntry->endpoint = NodeIPEndpoint{
+                        if (sourceNodeEntry->endpoint() != _from)
+                            sourceNodeEntry->node.endpoint = NodeIPEndpoint{
                                 _from.address(), _from.port(), nodeValidation.tcpPort};
                     }
                 }
@@ -538,10 +538,10 @@ void NodeTable::onPacketReceived(
                                   << ") not found in node table. Ignoring Neighbours packet.";
                     return;
                 }
-                if (sourceNodeEntry->endpoint != _from)
+                if (sourceNodeEntry->endpoint() != _from)
                 {
                     LOG(m_logger) << "Neighbours packet from unexpected endpoint " << _from
-                                  << " instead of " << sourceNodeEntry->endpoint;
+                                  << " instead of " << sourceNodeEntry->endpoint();
                     return;
                 }
 
@@ -577,10 +577,10 @@ void NodeTable::onPacketReceived(
                                   << ") not found in node table. Ignoring FindNode request.";
                     return;
                 }
-                if (sourceNodeEntry->endpoint != _from)
+                if (sourceNodeEntry->endpoint() != _from)
                 {
                     LOG(m_logger) << "FindNode packet from unexpected endpoint " << _from
-                                  << " instead of " << sourceNodeEntry->endpoint;
+                                  << " instead of " << sourceNodeEntry->endpoint();
                     return;
                 }
                 if (!sourceNodeEntry->lastPongReceivedTime)
@@ -723,7 +723,7 @@ void NodeTable::doHandleTimeouts()
 
         // activate replacement nodes and put them into buckets
         for (auto const& n : nodesToActivate)
-
+            noteActiveNode(n);
 
         doHandleTimeouts();
     });
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -499,7 +499,6 @@ void NodeTable::onPacketReceived(
                 // create or update nodeEntry with new Pong received time
                 DEV_GUARDED(x_nodes)
                 {
-                    auto const& sourceId = pong.sourceid;
                     auto it = m_allNodes.find(sourceId);
                     if (it == m_allNodes.end())
                         sourceNodeEntry = make_shared<NodeEntry>(m_hostNodeID, sourceId,
@@ -541,7 +540,8 @@ void NodeTable::onPacketReceived(
                 }
                 if (sourceNodeEntry->endpoint != _from)
                 {
-                    LOG(m_logger) << "Neighbours packet from unexpected endpoint.";
+                    LOG(m_logger) << "Neighbours packet from unexpected endpoint " << _from
+                                  << " instead of " << sourceNodeEntry->endpoint;
                     return;
                 }
 
@@ -579,7 +579,8 @@ void NodeTable::onPacketReceived(
                 }
                 if (sourceNodeEntry->endpoint != _from)
                 {
-                    LOG(m_logger) << "FindNode packet from unexpected endpoint.";
+                    LOG(m_logger) << "FindNode packet from unexpected endpoint " << _from
+                                  << " instead of " << sourceNodeEntry->endpoint;
                     return;
                 }
                 if (!sourceNodeEntry->lastPongReceivedTime)
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -83,7 +83,8 @@ bool NodeTable::addNode(Node const& _node)
     DEV_GUARDED(x_nodes)
     {
         auto const it = m_allNodes.find(_node.id);
-        needToPing = (it == m_allNodes.end() || !it->second->hasValidEndpointProof());
+        needToPing = (it == m_allNodes.end() || it->second->endpoint != _node.endpoint ||
+                      !it->second->hasValidEndpointProof());
     }
 
     if (needToPing)
@@ -115,7 +116,7 @@ bool NodeTable::addKnownNode(
     if (entry->hasValidEndpointProof())
     {
         LOG(m_logger) << "Known " << _node;
-        noteActiveNode(move(entry), _node.endpoint);
+        noteActiveNode(move(entry));
     }
     else
     {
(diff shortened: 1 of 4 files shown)


libp2p/NodeTable.cpp
```diff
@@ -375,10 +376,6 @@ void NodeTable::noteActiveNode(shared_ptr<NodeEntry> _nodeEntry, bi::udp::endpoi
         return;
 
     LOG(m_logger) << "Active node " << *_nodeEntry;
-    // TODO: don't activate in case endpoint has changed
-    _nodeEntry->endpoint.setAddress(_endpoint.address());
```
gumb0: This part is basically moved to Pong handler (lines 512-516), because we should trust changed endpoint only after valid Pong

libp2p/NodeTable.cpp
```diff
@@ -512,6 +510,10 @@ void NodeTable::onPacketReceived(
                         sourceNodeEntry = it->second;
                         sourceNodeEntry->lastPongReceivedTime =
                             RLPXDatagramFace::secondsSinceEpoch();
+
```
halfalicious: I know this is good practice but do you know if we ever actually hit this case? It seems like we wouldn't given that pongs have to be received within 1 minute of the ping being sent to be considered valid?

libp2p/NodeTable.cpp
```diff
@@ -537,6 +539,11 @@ void NodeTable::onPacketReceived(
                                   << ") not found in node table. Ignoring Neighbours packet.";
                     return;
                 }
+                if (sourceNodeEntry->endpoint != _from)
+                {
+                    LOG(m_logger) << "Neighbours packet from unexpected endpoint.";
```
halfalicious: Should we log both endpoints here?

libp2p/NodeTable.cpp
```diff
@@ -570,6 +577,11 @@ void NodeTable::onPacketReceived(
                                   << ") not found in node table. Ignoring FindNode request.";
                     return;
                 }
+                if (sourceNodeEntry->endpoint != _from)
+                {
+                    LOG(m_logger) << "FindNode packet from unexpected endpoint.";
```
halfalicious: Should we log both endpoints here?

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+            NodeIPEndpoint{bi::address::from_string(c_localhostIp), nodePort2, nodePort2};
+
+        // add a node to node table, initiating PING
+        nodeTable->addNode(Node{nodePubKey, nodeEndpoint1});
+
+        // handle received PING
+        auto pingDataReceived = nodeSocketHost1.packetsReceived.pop();
+        auto pingDatagram =
+            DiscoveryDatagram::interpretUDP(bi::udp::endpoint{}, dev::ref(pingDataReceived));
+        auto ping = dynamic_cast<PingNode const&>(*pingDatagram);
```
halfalicious: Can we validate the name of the ping packet here via BOOST_REQUIRE or is that not something we can do in a test fixture class?

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+
+        // send PONG
+        Pong pong(nodeTable->m_hostNodeEndpoint);
```
halfalicious: (Nit) {} initializers

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+        nodeEndpoint1 =
+            NodeIPEndpoint{bi::address::from_string(c_localhostIp), nodePort1, nodePort1};
+        nodeEndpoint2 =
+            NodeIPEndpoint{bi::address::from_string(c_localhostIp), nodePort2, nodePort2};
+
+        // add a node to node table, initiating PING
+        nodeTable->addNode(Node{nodePubKey, nodeEndpoint1});
+
+        // handle received PING
+        auto pingDataReceived = nodeSocketHost1.packetsReceived.pop();
```
halfalicious: I think we should have a timeout here

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+BOOST_AUTO_TEST_CASE(addNode)
+{
+    // this should initiate Ping to endpoint 2
+    nodeTable->addNode(Node{nodePubKey, nodeEndpoint2});
+
+    // handle received PING
+    auto pingDataReceived = nodeSocketHost2.packetsReceived.pop();
+    auto pingDatagram =
+        DiscoveryDatagram::interpretUDP(bi::udp::endpoint{}, dev::ref(pingDataReceived));
+    auto ping = dynamic_cast<PingNode const&>(*pingDatagram);
```
halfalicious: Can we validate the packet name before the dynamic cast?

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+
+    // send PONG
+    Pong pong(nodeTable->m_hostNodeEndpoint);
```
halfalicious: (Nit) {} initializers

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    NodeIPEndpoint nodeEndpoint2;
+    KeyPair nodeKeyPair = KeyPair::create();
+    NodeID nodePubKey = nodeKeyPair.pub();
+
+    shared_ptr<NodeEntry> nodeEntry;
+};
+
+BOOST_FIXTURE_TEST_SUITE(packetsWithChangedEndpointSuite, PacketsWithChangedEndpointFixture)
+
+BOOST_AUTO_TEST_CASE(addNode)
```
halfalicious: I think a better name for this test would be something like "pongWithChangedEndpoint"

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    // wait for PONG to be received and handled
+    nodeTable->packetsReceived.pop(chrono::seconds(5));
+
+    BOOST_REQUIRE_EQUAL(nodeEntry->endpoint, nodeEndpoint2);
+}
+
+BOOST_AUTO_TEST_CASE(findNode)
+{
+    // Create and send the FindNode packet through endpoint 2
+    FindNode findNode(nodeTable->m_hostNodeEndpoint, KeyPair::create().pub() /* target */);
```
halfalicious: (Nit) {} initializers

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+        DiscoveryDatagram::interpretUDP(bi::udp::endpoint{}, dev::ref(pingDataReceived));
+    auto ping = dynamic_cast<PingNode const&>(*pingDatagram);
+
+    // send PONG
+    Pong pong(nodeTable->m_hostNodeEndpoint);
+    pong.echo = ping.echo;
+    pong.sign(nodeKeyPair.secret());
+    nodeSocketHost2.socket->send(pong);
+
+    // wait for PONG to be received and handled
```
halfalicious: It might make sense to validate the name of the received packet to make the test easier to quickly debug in the event of a failure

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+
+BOOST_AUTO_TEST_CASE(findNode)
+{
+    // Create and send the FindNode packet through endpoint 2
+    FindNode findNode(nodeTable->m_hostNodeEndpoint, KeyPair::create().pub() /* target */);
+    findNode.sign(nodeKeyPair.secret());
+    nodeSocketHost2.socket->send(findNode);
+
+    // Wait for FindNode to be received
+    nodeTable->packetsReceived.pop(chrono::seconds(5));
```
halfalicious: It might make sense to validate the name of the received packet to make the test easier to quickly debug in the event of a failure

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    // Verify that no neighbours response is received
+    BOOST_CHECK_THROW(nodeSocketHost2.packetsReceived.pop(chrono::seconds(5)), WaitTimeout);
+}
+
+BOOST_AUTO_TEST_CASE(neighbours)
+{
+    // Wait for FindNode arriving to endpoint 1
+    auto findNodeDataReceived = nodeSocketHost1.packetsReceived.pop(chrono::seconds(10));
+    auto findNodeDatagram =
+        DiscoveryDatagram::interpretUDP(bi::udp::endpoint{}, dev::ref(findNodeDataReceived));
```
halfalicious: Can we validate the packet name?

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    // Wait for FindNode to be received
+    nodeTable->packetsReceived.pop(chrono::seconds(5));
+
+    // Verify that no neighbours response is received
+    BOOST_CHECK_THROW(nodeSocketHost2.packetsReceived.pop(chrono::seconds(5)), WaitTimeout);
+}
+
+BOOST_AUTO_TEST_CASE(neighbours)
+{
+    // Wait for FindNode arriving to endpoint 1
```
halfalicious: 10 seconds is a long time to wait...is it possible to have a test hook in the node table which initiates a FindNode request? This would also make the test more reliable in the event that we change node table timings

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    auto findNodeDataReceived = nodeSocketHost1.packetsReceived.pop(chrono::seconds(10));
+    auto findNodeDatagram =
+        DiscoveryDatagram::interpretUDP(bi::udp::endpoint{}, dev::ref(findNodeDataReceived));
+    auto findNode = dynamic_cast<FindNode const&>(*findNodeDatagram);
+
+    // send Neighbours through endpoint 2
+    // TODO fill nearest with one node
```
halfalicious: TODO? I think you already do this?

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    NodeIPEndpoint neighbourEndpoint{boost::asio::ip::address::from_string("200.200.200.200"),
+        c_defaultListenPort, c_defaultListenPort};
+    vector<shared_ptr<NodeEntry>> nearest{make_shared<NodeEntry>(nodeTable->m_hostNodeID,
+        KeyPair::create().pub(), neighbourEndpoint, RLPXDatagramFace::secondsSinceEpoch(), 0)};
```
halfalicious: Would be nice to comment the 0, especially since there are a lot of ctor args

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    Neighbours neighbours(nodeTable->m_hostNodeEndpoint, nearest);
```
halfalicious: (Nit) {} initializers

test/unittests/libp2p/net.cpp
```diff
@@ -1173,6 +1147,128 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    neighbours.sign(nodeKeyPair.secret());
+    nodeSocketHost2.socket->send(neighbours);
+
+    // Wait for Neighbours to be received
+    nodeTable->packetsReceived.pop(chrono::seconds(5));
```
halfalicious: Should we validate the received packet?

test/unittests/libp2p/net.cpp
gumb0: It's inside another suite, so it's full name is `packetsWithChangedEndpointSuite/addNode`. I didn't want to repeat `changedEndpoint` in the name... The full names unfortunately are not displayed in windows test log, but maybe we convert this to Google.Test at some point, then it'll be better.

libp2p/NodeTable.cpp
gumb0: Not sure how it is related to Pong timeout, this is the case when there is a record in the bucket of the node table (maybe added long ago), but we receive a valid Pong with the same NodeID but from another endpoint.

test/unittests/libp2p/net.cpp
gumb0: I think we can.

test/unittests/libp2p/net.cpp
gumb0: It's 7.2 seconds till the first discovery round... Not sure how to optimize it easily right now, I think I'll leave it like this for now

gumb0:
Addresses review issues, rebased and added CHANGELOG entry.

PR #5520, opened 2019-03-13
testeth calculate hash with rewards for retesteth [WIP]

winsvega:
retesteth run state tests as blockchain tests. but the generated test files contain hash of a state filled without mining reward. therefore retesteth is unable to validate the hash of a post state. (retesteth does not have logic for state classes) 

this PR might not be valid. I need to check this change on every test.

### DIFF
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -667,9 +667,9 @@ bool ImportTest::checkGeneralTestSectionSearch(json_spirit::mObject const& _expe
     else
     {
         // Expect section in filled test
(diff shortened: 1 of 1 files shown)


winsvega:
tested this change with retesteth. all works. we could change hash calculation in GeneralStateTests
https://github.com/ethereum/tests/issues/591

but there is no reason to keep hash for a state without mining reward. no client support this hash via rpc -> thus rpc test tool could not fill a test with such a hash.

winsvega:
retesteth will focus on blockchain tests. having this change is unnecessary.
GStateTests will get deprecated when switch to RPC tool. (gStateTests fillers wont)

PR #5523, opened 2019-03-16
Fix use of timers when managing capability background work loops

halfalicious:
Fix #5502 

Fix timer race condition which can result in the capability background work loops being prematurely terminated and also do some light code cleanup in related areas and update license messages.

I've fixed the race condition by giving each capability their own dedicated `steady_timer `in the host (stored in the new `CapabilityRuntime ` struct stored in the `m_capabilities `map). I've removed `Host::m_networkTimers `and the associated `Host::scheduleExecution ` since there was no need for general-purpose work scheduling via the host since we only ended up doing it when scheduling the capability background work loops or for capability work that we needed to execute on the network thread - capabilities now call the new `Host::scheduleCapabilityBackgroundWork` function to schedule their next work loop iteration and can post work to the network thread via` Host::postCapabilityWork`. Both of these functions require a `CapDesc `be supplied and verify that the `CapDesc ` is in the registered capabilities map. 

Capability shutdown is now done via timer cancellation, which means that there's no longer a need for `CapabilityFace::onStopping `or the `m_backgroundWorkEnabled `member in each capability implementation - as such, I've removed them.

I've verified syncing still works on Windows and Linux (Ubuntu).

### DIFF
--- a/libethereum/EthereumCapability.cpp
+++ b/libethereum/EthereumCapability.cpp
@@ -407,13 +407,7 @@ chrono::milliseconds EthereumCapability::backgroundWorkInterval() const
 
 void EthereumCapability::onStarting()
 {
-    m_backgroundWorkEnabled = true;
     m_host->scheduleCapabilityBackgroundWork({name(), version()}, [this]() { doBackgroundWork(); });
-}
-
-void EthereumCapability::onStopping()
-{
-    m_backgroundWorkEnabled = false;
 }
 
 bool EthereumCapability::ensureInitialised()
@@ -480,9 +474,7 @@ void EthereumCapability::doBackgroundWork()
         }
     }
 
-    if (m_backgroundWorkEnabled)
-        m_host->scheduleCapabilityBackgroundWork(
-            {name(), version()}, [this]() { doBackgroundWork(); });
+    m_host->scheduleCapabilityBackgroundWork({name(), version()}, [this]() { doBackgroundWork(); });
 }
 
 void EthereumCapability::maintainTransactions()
--- a/libethereum/EthereumCapability.cpp
+++ b/libethereum/EthereumCapability.cpp
@@ -19,18 +19,18 @@ using namespace std;
 using namespace dev;
 using namespace dev::eth;
 
-char const* const EthereumCapability::s_stateNames[static_cast<int>(SyncState::Size)] = {
+char const* const EthereumCapability::c_stateNames[static_cast<int>(SyncState::Size)] = {
     "NotSynced", "Idle", "Waiting", "Blocks", "State"};
 
-chrono::milliseconds constexpr EthereumCapability::s_backgroundWorkInterval;
+chrono::milliseconds constexpr EthereumCapability::c_backgroundWorkInterval;
 
 namespace
 {
-static constexpr unsigned c_maxSendTransactions = 256;
-static constexpr unsigned c_maxHeadersToSend = 1024;
-static constexpr unsigned c_maxIncomingNewHashes = 1024;
-static constexpr unsigned c_peerTimeoutSeconds = 10;
-static constexpr int c_minBlockBroadcastPeers = 4;
+constexpr unsigned c_maxSendTransactions = 256;
+constexpr unsigned c_maxHeadersToSend = 1024;
+constexpr unsigned c_maxIncomingNewHashes = 1024;
+constexpr unsigned c_peerTimeoutSeconds = 10;
+constexpr int c_minBlockBroadcastPeers = 4;
 
 string toString(Asking _a)
 {
@@ -402,12 +402,7 @@ EthereumCapability::EthereumCapability(shared_ptr<p2p::CapabilityHostFace> _host
 
 chrono::milliseconds EthereumCapability::backgroundWorkInterval() const
 {
-    return s_backgroundWorkInterval;
-}
-
-void EthereumCapability::onStarting()
-{
-    m_host->scheduleCapabilityBackgroundWork({name(), version()}, [this]() { doBackgroundWork(); });
+    return c_backgroundWorkInterval;
 }
 
 bool EthereumCapability::ensureInitialised()
@@ -430,7 +425,7 @@ void EthereumCapability::reset()
 
     // reset() can be called from RPC handling thread,
     // but we access m_latestBlockSent and m_transactionsSent only from the network thread
-    m_host->postCapabilityWork({name(), version()}, [this]() {
+    m_host->postWork([this]() {
         m_latestBlockSent = h256();
         m_transactionsSent.clear();
     });
@@ -439,42 +434,6 @@ void EthereumCapability::reset()
 void EthereumCapability::completeSync()
 {
     m_sync->completeSync();
-}
-
-void EthereumCapability::doBackgroundWork()
-{
-    ensureInitialised();
-    auto h = m_chain.currentHash();
-    // If we've finished our initial sync (including getting all the blocks into the chain so as to reduce invalid transactions), start trading transactions & blocks
-    if (!isSyncing() && m_chain.isKnown(m_latestBlockSent))
-    {
-        if (m_newTransactions)
-        {
-            m_newTransactions = false;
-            maintainTransactions();
-        }
-        if (m_newBlocks)
-        {
-            m_newBlocks = false;
-            maintainBlocks(h);
-        }
-    }
-
-    time_t now = chrono::system_clock::to_time_t(chrono::system_clock::now());
-    if (now - m_lastTick >= 1)
-    {
-        m_lastTick = now;
-        for (auto const& peer : m_peers)
-        {
-            time_t now = chrono::system_clock::to_time_t(chrono::system_clock::now());
-
-            if (now - peer.second.lastAsk() > c_peerTimeoutSeconds && peer.second.isConversing())
-                // timeout
-                m_host->disconnect(peer.first, p2p::PingTimeout);
-        }
-    }
-
-    m_host->scheduleCapabilityBackgroundWork({name(), version()}, [this]() { doBackgroundWork(); });
 }
 
 void EthereumCapability::maintainTransactions()
@@ -637,7 +596,7 @@ SyncStatus EthereumCapability::status() const
 void EthereumCapability::onTransactionImported(
     ImportResult _ir, h256 const& _h, h512 const& _nodeId)
 {
-    m_host->postCapabilityWork({name(), version()}, [this, _ir, _h, _nodeId]() {
+    m_host->postWork([this, _ir, _h, _nodeId]() {
         auto itPeerStatus = m_peers.find(_nodeId);
         if (itPeerStatus == m_peers.end())
             return;
@@ -889,6 +848,41 @@ bool EthereumCapability::interpretCapabilityPacket(
     }
 
     return true;
+}
+
+void EthereumCapability::doBackgroundWork()
+{
+    ensureInitialised();
+    auto h = m_chain.currentHash();
+    // If we've finished our initial sync (including getting all the blocks into the chain so as to
+    // reduce invalid transactions), start trading transactions & blocks
+    if (!isSyncing() && m_chain.isKnown(m_latestBlockSent))
+    {
+        if (m_newTransactions)
+        {
+            m_newTransactions = false;
+            maintainTransactions();
+        }
+        if (m_newBlocks)
+        {
+            m_newBlocks = false;
+            maintainBlocks(h);
+        }
+    }
+
+    time_t now = chrono::system_clock::to_time_t(chrono::system_clock::now());
+    if (now - m_lastTick >= 1)
+    {
+        m_lastTick = now;
+        for (auto const& peer : m_peers)
+        {
+            time_t now = chrono::system_clock::to_time_t(chrono::system_clock::now());
+
+            if (now - peer.second.lastAsk() > c_peerTimeoutSeconds && peer.second.isConversing())
+                // timeout
+                m_host->disconnect(peer.first, p2p::PingTimeout);
+        }
+    }
 }
 
 void EthereumCapability::setIdle(NodeID const& _peerID)
--- a/libethereum/EthereumCapability.cpp
+++ b/libethereum/EthereumCapability.cpp
@@ -22,15 +22,15 @@ using namespace dev::eth;
 char const* const EthereumCapability::s_stateNames[static_cast<int>(SyncState::Size)] = {
     "NotSynced", "Idle", "Waiting", "Blocks", "State"};
 
-constexpr chrono::milliseconds c_backgroundWorkInterval{1000};
+chrono::milliseconds constexpr EthereumCapability::s_backgroundWorkInterval;
 
 namespace
 {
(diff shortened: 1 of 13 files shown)


halfalicious:
A possible "optimization" to these changes - currently capability background work is scheduled by the capability scheduling its `doBackgroundWork` function via `Host::scheduleCapabilityBackgroundWork()`...so the capability will call into the host who will schedule the capability function via the timer. I'm thinking this isn't needed and the entire thing can be driven by the host which results in a slightly more straightforward execution flow - the host can directly schedule capability work via the timer (probably by capturing the capability instance in a lambda passed to `async_wait ` and calling `doBackgroundWork`). This would also enable me to remove the `onStarting` function from each capability. Thoughts?

libp2p/Host.h
```diff
@@ -196,6 +196,9 @@ class Host: public Worker
     /// Resets acceptor, socket, and IO service. Called by deallocator.
     void stop();
 
+    /// Stop registered capabilities, typically done when the network is being shut down.
+    void stopCapabilities();
```
gumb0: This shouldn't be public, I think

libethereum/WarpCapability.cpp
```diff
@@ -1,35 +1,27 @@
...
 
 #include "WarpCapability.h"
 #include "BlockChain.h"
 #include "SnapshotStorage.h"
+#include "libp2p/Common.h"
 
 #include <boost/fiber/all.hpp>
 #include <chrono>
 
+using namespace std;
```
gumb0: I actually prefer to avoid this. My opinion is that it improves readability to explicitly show what comes from the standard library.
Also @chfast complained that it can lead to hard-to-track compilation errors in cases of conflict between names of different namespaces.

libethereum/WarpCapability.h
```diff
@@ -117,6 +104,8 @@ class WarpCapability : public p2p::CapabilityFace
     void disablePeer(NodeID const& _peerID, std::string const& _problem);
 
 private:
+    static constexpr std::chrono::milliseconds s_backgroundWorkInterval{1000};
```
gumb0: Constants have `c_` prefix

libethereum/EthereumCapability.h
```diff
@@ -144,16 +131,22 @@ class EthereumCapability : public p2p::CapabilityFace
 
 private:
     static char const* const s_stateNames[static_cast<int>(SyncState::Size)];
+    static constexpr std::chrono::milliseconds s_backgroundWorkInterval{1000};
```
gumb0: `c_backgroundWorkInterval`

libethereum/EthereumCapability.cpp
```diff
@@ -31,17 +19,19 @@ using namespace std;
...
-static int const c_minBlockBroadcastPeers = 4;
-
 char const* const EthereumCapability::s_stateNames[static_cast<int>(SyncState::Size)] = {
     "NotSynced", "Idle", "Waiting", "Blocks", "State"};
 
+chrono::milliseconds constexpr EthereumCapability::s_backgroundWorkInterval;
+
 namespace
 {
+static constexpr unsigned c_maxSendTransactions = 256;
```
gumb0: Having it in the unnamed namespace and making it `static` has the exact same effect, so having both is redundant. If you move it, remove `static`

gumb0:
Re optimization - I think it sounds ok in general.

A bit disappointing is that now the API of libp2p gets less flexible/less general - e.g. capabilities can't have more that one timer, if they would need it. The simple way to do this optimization would be to require them to always provide a function for background work - i.e. no option to not have one...

But I guess it's ok to get rid of this flexibility, as it actually is not needed at the moment, and we don't plan any new capabilities soon.

libp2p/CapabilityHost.h
```diff
@@ -86,9 +86,9 @@ class CapabilityHostFace
     virtual void foreachPeer(std::string const& _capabilityName,
         std::function<bool(NodeID const& _nodeID)> _f) const = 0;
 
-    /// Schedule callback to be called from the network handling thread
-    /// after at least @a _delayMs milliseconds
-    virtual void scheduleExecution(int _delayMs, std::function<void()> _f) = 0;
+    // TODO:
```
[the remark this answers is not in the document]
halfalicious: Ah, I thought I had commented that function but it turns out I did so in Host but not CapabilityHost. Thanks for calling this out!

libp2p/Host.cpp
```diff
@@ -1032,13 +1029,30 @@ void Host::forEachPeer(
...
+    auto timer = cap->second.backgroundWorkTimer;
+    timer->expires_from_now(cap->second.capability->backgroundWorkInterval());
+    timer->async_wait([timer, _f](boost::system::error_code _ec) {
+        if (timer->expires_at() != c_steadyClockMin && !_ec)
             _f();
     });
-    DEV_GUARDED(x_networkTimers) { m_networkTimers.emplace_back(move(t)); }
 }
+
+void Host::postCapabilityWork(CapDesc const& _capDesc, function<void()> _f)
```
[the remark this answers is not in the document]
halfalicious: We don't technically need it but I wanted to "lock down" this function to only accept work from capabilities. If this isn't technically needed I can remove the parameter (and rename the function)

libp2p/CapabilityHost.h
```diff
@@ -86,9 +86,9 @@ class CapabilityHostFace
     virtual void foreachPeer(std::string const& _capabilityName,
         std::function<bool(NodeID const& _nodeID)> _f) const = 0;
 
-    /// Schedule callback to be called from the network handling thread
-    /// after at least @a _delayMs milliseconds
-    virtual void scheduleExecution(int _delayMs, std::function<void()> _f) = 0;
+    // TODO: 
+    virtual void scheduleCapabilityBackgroundWork(CapDesc const& _capDesc, std::function<void()> _f) = 0;
```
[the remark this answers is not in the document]
halfalicious: Ah, good point!

libp2p/Host.h
halfalicious: Whoops, yes you're right, that was an accident 😄

libethereum/WarpCapability.cpp
halfalicious: Ah okay, we do this in many source files so I did so here in the interest of consistency. What you're saying makes sense...I'll undo this change.

libethereum/EthereumCapability.cpp
halfalicious: Ah you're right, objects defined in namespace scope have static storage duration. Thanks for calling this out!

libp2p/Capability.h
```diff
@@ -39,13 +28,12 @@ class CapabilityFace
     virtual unsigned version() const = 0;
     /// Number of messages supported by the subprotocol version.
     virtual unsigned messageCount() const = 0;
+    /// Time interval to run the background work loop at
+    virtual std::chrono::milliseconds backgroundWorkInterval() const = 0;
```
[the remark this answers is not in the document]
halfalicious: What's the benefit of this? I intentionally went with this approach because it made more conceptual sense to me to have the host "pull" this value because it's managing capability execution. I'd be happy to go with another approach if it makes more sense though 😄

libethereum/WarpCapability.h
halfalicious: I saw some cases where static overrode const e.g.:

https://github.com/ethereum/aleth/blob/ae042a32103e4c421f942952bd42afdbd1e6b96f/libethereum/EthereumCapability.h#L131

I'll switch back to using c_ (and update the above instance)

halfalicious:
> 
> 
> I suggest passing timer interval as a parameter of `scheduleBackgroundWork` and removing `CapDesc` parameter from `postWork`.
> 
> Also too many offtopic changes - I recommend not to do `using namespace std` at all.

Removing the std:: prefix added a lot of churn, I've undone those changes. Sorry about that!

halfalicious:
cc @gumb0

libp2p/Host.cpp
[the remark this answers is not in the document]
gumb0: I think it might be useful not only to capabilities

libp2p/Common.h
```diff
@@ -114,6 +114,8 @@ using CapDesc = std::pair<std::string, unsigned>;
 using CapDescSet = std::set<CapDesc>;
 using CapDescs = std::vector<CapDesc>;
 
+std::ostream& operator<<(std::ostream& _out, CapDesc const& _capDesc);
```
gumb0: Does logging not work without this?
We have a  boost.log output operator for `std::pair` (outputs it like "(first, second)"), but I'm not sure whether it's picked up for the alias of pair.

libp2p/Host.cpp
```diff
@@ -152,21 +152,60 @@ void Host::stop()
         stopWorking();
 }
 
+void Host::startCapabilities()
+{
+    for (auto const& itCap : m_capabilities)
+    {
+        auto timer = itCap.second.backgroundWorkTimer;
+        auto cap = itCap.second.capability;
+        m_ioService.post([this, timer, cap]() {
```
gumb0: Is this done to execute the first iteration without delay?
I think just calling `scheduleCapabilityWorkLoop` for each capability would be simpler and the delay wouldn't make any difference in practice.

libethereum/EthereumCapability.cpp
```diff
@@ -545,18 +494,18 @@ vector<NodeID> EthereumCapability::selectPeers(
...
-    vector<NodeID> part1(_peers);
-    vector<NodeID> part2;
+    std::vector<NodeID> part1(_peers);
+    std::vector<NodeID> part2;
 
     if (_number >= _peers.size())
         return std::make_pair(part1, part2);
 
-    std::shuffle(part1.begin(), part1.end(), m_urng);
+    shuffle(part1.begin(), part1.end(), m_urng);
```
gumb0: it seems in some random places you remove `std::` while in many others you add it

libp2p/Host.cpp
halfalicious: Yes, you are correct. I'll remove this.

libethereum/EthereumCapability.cpp
halfalicious: Looks like I missed some places when I was adding them back, sorry about that! This also means that one of the #included header files does a "using namespace std"...I'll leave that alone though to keep unnecessary churn to a minimum.

halfalicious: Changed unrolled!

halfalicious:
Squashed some commits

halfalicious:
Rebased

halfalicious:
This test is failing due to a UAF:
`
531 - JsonRpcSuite/jsonrpc_isListening (Failed)`

I've investigated and the problem is that there is still a capability handler in the run queue when the Host is destroyed, which results in the Capability(Face) instance staying alive (since the handler captures a shared_ptr to the capability) until the io service is destroyed...destroying the Capability instance releases the associated BlockchainSync (held via shared_ptr) which results in the BlockchainSync dtor executing what looks like sync cleanup code which calls into Host::foreachPeer which iterates over sessions but the sessions have already been destroyed (they are declared after the io service in Host.h). 

I've also found another potential issue with these changes - if there aren't any peers and you shut down the host, I think there will always be capability handlers in the run queue since we post capability timer cancellations via the IO service (so they will execute on the network thread) but at this point the IO service has already been stopped and is only polled if there are peers/incoming connections, which means that the timer cancellations may never execute. 

I can think of a few ways to address this:
* Don't capture shared_ptr to capabilities in the capability handlers. This isn't technically necessary anyway since we first check the timer status in the handler (which is captured via shared_ptr) and exit if the timer has been cancelled/aborted. This is the simplest approach.

halfalicious:
cc @gumb0  / @chfast

gumb0:
> We need to do this because we cancel capabilities after we've stopped the io service and we don't do any polling unless there are peers or incoming connections, which means that the cancellations wouldn't be processed and just sit in the handler queue.

I didn't quite get why this is needed. What happens if the cancellations aren't processed and sit in the handler queue? If we don't poll io_service at this point anymore, it shouldn't affect anything?

libp2p/Host.h
```diff
@@ -307,8 +307,7 @@ class Host: public Worker
     void startCapabilities();
 
     /// Schedule's a capability's work loop on the network thread
-    void scheduleCapabilityWorkLoop(
-        std::shared_ptr<CapabilityFace> _cap, std::shared_ptr<ba::steady_timer> _timer);
+    void scheduleCapabilityWorkLoop(CapabilityFace* _cap, std::shared_ptr<ba::steady_timer> _timer);
```
gumb0: Better to pass it by reference, than by raw pointer.

halfalicious:
> > We need to do this because we cancel capabilities after we've stopped the io service and we don't do any polling unless there are peers or incoming connections, which means that the cancellations wouldn't be processed and just sit in the handler queue.
> 
> I didn't quite get why this is needed. What happens if the cancellations aren't processed and sit in the handler queue? If we don't poll io_service at this point anymore, it shouldn't affect anything?

We don’t need to cancel the capability background work loops to address the UAF as long as we capture a raw capability pointer rather than a smart pointer in the background work loop handler. I don’t know if shutting down the host without canceling the work loops would cause other issues though (I don’t know enough about syncing and how the capabilities are involved). What are your thoughts, would this be safe to do?

gumb0:
But cancelling the work loops won't affect anything unless `m_ioService.poll()` or `m_ioService.run()` is called (i.e. unless the work loops themselves are called). So it seems that that change won't affect anything.

> I don’t know if shutting down the host without canceling the work loops would cause other issues though (I don’t know enough about syncing and how the capabilities are involved).

This cancelling is now happens inside `Host`, capabilities shouldn't be concerned with this "cancelling" event.
(Previously there was `onStopping` callback, but now you removed it, because it's not really useful)

halfalicious:
> 
> 
> But cancelling the work loops won't affect anything unless `m_ioService.poll()` or `m_ioService.run()` is called after that (i.e. unless the work loops themselves are called). So it seems that that change won't affect anything.

But the IO service can still be polled after we stop the capabilities if we are disconnecting peers or pending handshakes:

https://github.com/ethereum/aleth/blob/f427584da0e64de46e9d019d12461a82882c3afa/libp2p/Host.cpp#L220-L251

gumb0:
> But the IO service can still be polled after we stop the capabilities if we are disconnecting peers or pending handshakes

Right, in case it is polled - timers will be cancelled anyway by these `poll()` calls. In case it is not polled - it doesn't matter.

gumb0:
I'll merge it now, please add a changelog item and remove `m_ioService.poll();` call in a separate PR if you agree with my reasoning.

halfalicious:
> I'll merge it now, please add a changelog item and remove `m_ioService.poll();` call in a separate PR if you agree with my reasoning.

@gumb0 : Ah, I see what you’re saying - we only need to cancel the timers if we are going to schedule more capability work, and more capability work is only scheduled if poll is called while disconnecting pending handshakes / peers. If we post the timer cancellations before we disconnect peers / handshakes the polling will cancel the timers. So there’s no need to explicitly call poll after calling stopCapabities.

I’ll remove the poll call and update the change log in a new PR.

PR #5526, opened 2019-03-23
Log json exception details on chain config file load error

halfalicious:
Also return new error code `AlethErrors::ConfigFileInvalid`


gumb0:
I would prefer to keep `json_spirit` dependency inside `ChainParams.cpp`, i.e. handle  `js::Error_position` in `ChainParams::loadConfig` and re-throw as `dev::Exception`-derived exception.

(There is already some validation of json in `loadConfig`, that does this (see `validateConfigJson`). Also it logs directly to `std::error`, I think you may do the same)

halfalicious:
> at does this (see `validateConfigJson`). Also it logs directly to `std::error`, I think you may do the same)

Makes sense, thanks for the feedback!

halfalicious:
Rebased (to get changelog) and updated changelog

halfalicious:
Looks like there's a unit test that needs updating

halfalicious:
This test is failing on osx but it's a test reliability issue since my changes don't touch the discovery part of the code:

`510 - network/net/packetsWithChangedEndpointSuite/neighbours (Failed)`

I'll open a new issue so we can track this.

halfalicious:
Rebased to fix changelog

PR #5533, opened 2019-03-27
Don't manually poll the ioservice after stopping capabilities

halfalicious:
Manually polling the ioservice on shutdown is redundant. The motivation behind doing this was to ensure that the posted capability timer cancellations (see Host::stopCapabilities) were processed during host shutdown (see #5523 for details). However, it turns out that there's no need to do this because the capability timers only need to be cancelled if more capability work loops are being scheduled, but more capability work loops are only scheduled during shutdown if we are disconnecting pending handshakes and/or peers, each of which result in the io service being polled anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,5 +11,6 @@
 - Fixed: [#5502](https://github.com/ethereum/aleth/pull/5502) Fix Discovery terminating prematurely because of race condition.
 - Fixed: [#5452](https://github.com/ethereum/aleth/pull/5452) Correctly handle Discovery messages when the peer changes public key.
 - Fixed: [#5519](https://github.com/ethereum/aleth/pull/5519) Correctly handle Discovery messages with known public key but unknown endpoint.
+- Fixed: [#5523](https://github.com/ethereum/aleth/pull/5523) [#5533](https://github.com/ethereum/aleth/pull/5533) Fix syncing terminating prematurely because of race condition.
 
 [1.6.0]: https://github.com/ethereum/aleth/compare/v1.6.0-alpha.1...master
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -11,5 +11,7 @@
 - Fixed: [#5502](https://github.com/ethereum/aleth/pull/5502) Fix Discovery terminating prematurely because of race condition.
 - Fixed: [#5452](https://github.com/ethereum/aleth/pull/5452) Correctly handle Discovery messages when the peer changes public key.
 - Fixed: [#5519](https://github.com/ethereum/aleth/pull/5519) Correctly handle Discovery messages with known public key but unknown endpoint.
+- Fixed: [#5523](https://github.com/ethereum/aleth/pull/5523) Fix syncing terminating prematurely because of race condition.
+- Fixed: [#5533](https://github.com/ethereum/aleth/pull/5533) Remove unnecessary ioservice polling on network shutdown.
```
gumb0: I would skip this, too low-level for the user and it's something that we introduced and then quickly fixed inside one release.

chfast: Addressed.

CHANGELOG.md
```diff
@@ -11,5 +11,7 @@
 - Fixed: [#5502](https://github.com/ethereum/aleth/pull/5502) Fix Discovery terminating prematurely because of race condition.
 - Fixed: [#5452](https://github.com/ethereum/aleth/pull/5452) Correctly handle Discovery messages when the peer changes public key.
 - Fixed: [#5519](https://github.com/ethereum/aleth/pull/5519) Correctly handle Discovery messages with known public key but unknown endpoint.
+- Fixed: [#5523](https://github.com/ethereum/aleth/pull/5523) Fix syncing terminating prematurely because of race condition.
```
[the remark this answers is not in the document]
chfast: Addressed.

Issue #5517, opened 2019-03-12
SecureFixedHash wrong argument type for ConstructFromPointerType


etam:
This constructor:

https://github.com/ethereum/aleth/blob/7ef4da6f31214c5c5dcee664743e3f0c362f9c03/libdevcore/FixedHash.h#L245

It should take `byte const* _d`, not `bytes const* _d`.

PR #5536, opened 2019-03-28
SecureFixedHash argument fixed

twinstar26:
Constructor arguments of
`explicit SecureFixedHash(bytes const* _d, ConstructFromPointerType _t): FixedHash<T>(_d, _t) {}`
changed to
`explicit SecureFixedHash(byte const* _d, ConstructFromPointerType _t): FixedHash<T>(_d, _t)


chfast:
Fixes #5517

PR #5537, opened 2019-03-28
ENR record for host

gumb0:
https://eips.ethereum.org/EIPS/eip-778

This creates ENR record for the host at program start, filling values based on `--public-ip` and `--listen` CLI parameters.

For example `aleth --public-ip 13.74.189.147 --listen 30303` creates ENR like
```
[<signature>, 0, "id", "v4", "ip", 0xD4ABD93, "sec256k1", <compressed public key>, "tcp", 30303, "udp", 30303]
```
(sequence number is always 0 now, updating ENR will be added in later PRs)

At program exit ENR is serialized into `network.rlp` file and restored at the next program start.

cc @fjl

### DIFF
--- a/libp2p/ENR.cpp
+++ b/libp2p/ENR.cpp
@@ -35,13 +35,21 @@ ENR::ENR(RLP _rlp, VerifyFunction _verifyFunction)
     m_signature = _rlp[0].toBytes(RLP::VeryStrict);
 
     m_seq = _rlp[1].toInt<uint64_t>(RLP::VeryStrict);
-    // TODO check order
+
+    // read key-values into vector first, to check the order
+    std::vector<std::pair<std::string const, bytes>> keyValues;
     for (size_t i = 2; i < _rlp.itemCount(); i+= 2)
     {
-        auto key = _rlp[i].toString(RLP::VeryStrict);
-        auto value = _rlp[i + 1].data().toBytes();
-        m_map.insert({key, value});
+        auto const key = _rlp[i].toString(RLP::VeryStrict);
+        auto const value = _rlp[i + 1].data().toBytes();
+        keyValues.push_back({key, value});
     }
+
+    // transfer to map, this will order them
+    m_map.insert(keyValues.begin(), keyValues.end());
+
+    if (!std::equal(m_map.begin(), m_map.end(), keyValues.begin()))
+        BOOST_THROW_EXCEPTION(ENRKeysAreNotSorted());
 
     if (!_verifyFunction(m_map, dev::ref(m_signature), dev::ref(content())))
         BOOST_THROW_EXCEPTION(ENRSignatureIsInvalid());
--- a/libp2p/ENR.cpp
+++ b/libp2p/ENR.cpp
@@ -92,7 +92,7 @@ ENR createV4ENR(Secret const& _secret, boost::asio::ip::address const& _ip, uint
         // dev::sign returns 65 bytes signature containing r,s,v values
         Signature s = dev::sign(_secret, sha3(_data)); 
         // The resulting 64-byte signature is encoded as the concatenation of the r and s signature values.
-        return bytes(s.data(), s.data() + 64);
+        return bytes(&s[0], &s[64]);
     };
 
     PublicCompressed const publicKey = toPublicCompressed(_secret);
--- a/libp2p/ENR.cpp
+++ b/libp2p/ENR.cpp
@@ -17,7 +17,7 @@ constexpr char c_keyIP[] = "ip";
 constexpr char c_keyTCP[] = "tcp";
 constexpr char c_keyUDP[] = "udp";
 constexpr char c_IDV4[] = "v4";
-constexpr size_t c_ENRMaxSize = 300;
+constexpr size_t c_ENRMaxSizeBytes = 300;
 
 
 // Address can be either boost::asio::ip::address_v4 or boost::asio::ip::address_v6
@@ -29,9 +29,9 @@ bytes addressToBytes(Address const& _address)
 }
 }  // namespace
 
-ENR::ENR(RLP _rlp, VerifyFunction _verifyFunction)
+ENR::ENR(RLP _rlp, VerifyFunction const& _verifyFunction)
 {
-    if (_rlp.data().size() > c_ENRMaxSize)
+    if (_rlp.data().size() > c_ENRMaxSizeBytes)
         BOOST_THROW_EXCEPTION(ENRIsTooBig());
 
     m_signature = _rlp[0].toBytes(RLP::VeryStrict);
@@ -39,32 +39,33 @@ ENR::ENR(RLP _rlp, VerifyFunction _verifyFunction)
     m_seq = _rlp[1].toInt<uint64_t>(RLP::VeryStrict);
 
     // read key-values into vector first, to check the order
-    std::vector<std::pair<std::string const, bytes>> keyValues;
-    for (size_t i = 2; i < _rlp.itemCount(); i+= 2)
+    std::vector<std::pair<std::string const, bytes>> keyValuePairs;
+    for (size_t i = 2; i < _rlp.itemCount(); i += 2)
     {
         auto const key = _rlp[i].toString(RLP::VeryStrict);
         auto const value = _rlp[i + 1].data().toBytes();
-        keyValues.push_back({key, value});
+        keyValuePairs.push_back({key, value});
     }
 
     // transfer to map, this will order them
-    m_map.insert(keyValues.begin(), keyValues.end());
+    m_map.insert(keyValuePairs.begin(), keyValuePairs.end());
 
-    if (!std::equal(keyValues.begin(), keyValues.end(), m_map.begin()))
+    if (!std::equal(keyValuePairs.begin(), keyValuePairs.end(), m_map.begin()))
         BOOST_THROW_EXCEPTION(ENRKeysAreNotUniqueSorted());
 
     if (!_verifyFunction(m_map, dev::ref(m_signature), dev::ref(content())))
         BOOST_THROW_EXCEPTION(ENRSignatureIsInvalid());
 }
 
-ENR::ENR(uint64_t _seq, std::map<std::string, bytes> const& _keyValues, SignFunction _signFunction)
-  : m_seq(_seq), m_map(_keyValues), m_signature(_signFunction(dev::ref(content())))
+ENR::ENR(uint64_t _seq, std::map<std::string, bytes> const& _keyValuePairs,
+    SignFunction const& _signFunction)
+  : m_seq{_seq}, m_map{_keyValuePairs}, m_signature{_signFunction(dev::ref(content()))}
 {
 }
 
 bytes ENR::content() const
 {
-    RLPStream stream(contentListSize());
+    RLPStream stream{contentRlpListItemCount()};
     streamContent(stream);
     return stream.out();
 }
@@ -72,7 +73,7 @@ bytes ENR::content() const
 
 void ENR::streamRLP(RLPStream& _s) const
 {
-    _s.appendList(contentListSize() + 1);
+    _s.appendList(contentRlpListItemCount() + 1);
     _s << m_signature;
     streamContent(_s);
 }
@@ -80,7 +81,7 @@ void ENR::streamRLP(RLPStream& _s) const
 void ENR::streamContent(RLPStream& _s) const
 {
     _s << m_seq;
-    for (auto const keyValue : m_map)
+    for (auto const& keyValue : m_map)
     {
         _s << keyValue.first;
         _s.appendRaw(keyValue.second);
@@ -102,26 +103,26 @@ ENR createV4ENR(Secret const& _secret, boost::asio::ip::address const& _ip, uint
 
     // Values are of different types (string, bytes, uint16_t),
     // so we store them as RLP representation
-    std::map<std::string, bytes> const keyValues = {{c_keyID, rlp(c_IDV4)},
+    std::map<std::string, bytes> const keyValuePairs = {{c_keyID, rlp(c_IDV4)},
         {c_keySec256k1, rlp(publicKey.asBytes())}, {c_keyIP, rlp(address)},
         {c_keyTCP, rlp(_tcpPort)}, {c_keyUDP, rlp(_udpPort)}};
 
-    return ENR{0, keyValues, signFunction};
+    return ENR{0 /* sequence number */, keyValuePairs, signFunction};
 }
 
 ENR parseV4ENR(RLP _rlp)
 {
-    ENR::VerifyFunction verifyFunction = [](std::map<std::string, bytes> const& _keyValues,
+    ENR::VerifyFunction verifyFunction = [](std::map<std::string, bytes> const& _keyValuePairs,
                                              bytesConstRef _signature, bytesConstRef _data) {
-        auto itID = _keyValues.find(c_keyID);
-        if (itID == _keyValues.end())
+        auto itID = _keyValuePairs.find(c_keyID);
+        if (itID == _keyValuePairs.end())
             return false;
-        auto id = RLP(itID->second).toString(RLP::VeryStrict);
+        auto const id = RLP(itID->second).toString(RLP::VeryStrict);
         if (id != c_IDV4)
             return false;
 
-        auto itKey = _keyValues.find(c_keySec256k1);
-        if (itKey == _keyValues.end())
+        auto itKey = _keyValuePairs.find(c_keySec256k1);
+        if (itKey == _keyValuePairs.end())
             return false;
 
         auto const key = RLP(itKey->second).toHash<PublicCompressed>(RLP::VeryStrict);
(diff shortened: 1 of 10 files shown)


gumb0:
@chfast @halfalicious I think this is ready for review. 

It's based on https://github.com/ethereum/aleth/pull/5542 and should be rebased when it's merged.

Also I need to add a CHANGELOG item here, but maybe we merge this after 1.6.0 release?

halfalicious:
The failed tests might just be unreliable, see #5544

libp2p/ENR.cpp
```diff
@@ -0,0 +1,148 @@
+// Aleth: Ethereum C++ client, tools and libraries.
+// Copyright 2019 Aleth Authors.
+// Licensed under the GNU General Public License, Version 3.
+
+#include "ENR.h"
+#include <libdevcore/SHA3.h>
+
+static std::string const c_keyID = "id";
```
chfast: I don't think it is great to have global dynamically constructed objects.

https://godbolt.org/z/qZwVxl

libp2p/ENR.cpp
```diff
@@ -0,0 +1,148 @@
...
+    }
+}
+
+ENR createV4ENR(Secret const& _secret, boost::asio::ip::address const& _ip, uint16_t _tcpPort,  uint16_t _udpPort)
+{
+    ENR::SignFunction signFunction = [&_secret](bytesConstRef _data) {
+        // dev::sign returns 65 bytes signature containing r,s,v values
+        Signature s = dev::sign(_secret, sha3(_data)); 
+        // The resulting 64-byte signature is encoded as the concatenation of the r and s signature values.
+        return bytes(s.data(), s.data() + 64);
```
chfast: Avoid pointer arithmetic.
```suggestion
        return bytes(&s[0], &s[64]);
```

gumb0:
Rebased.

libp2p/ENR.cpp
gumb0: Because of run-time initialization cost? Is it better to have dynamic allocation each time this code is run? (`string` is probably short-string-optimized, but `vector` isn't)

chfast: Yes, to avoid dynamic initialization during startup.

This can land, but I'm leaving it to consider changes.

libp2p/ENR.cpp
```diff
@@ -0,0 +1,149 @@
...
+{
+namespace
+{
+constexpr char c_keyID[] = "id";
+constexpr char c_keySec256k1[] = "secp256k1";
+constexpr char c_keyIP[] = "ip";
+constexpr char c_keyTCP[] = "tcp";
+constexpr char c_keyUDP[] = "udp";
+constexpr char c_IDV4[] = "v4";
+constexpr size_t c_ENRMaxSize = 300;
```
halfalicious: Would be nice to call out that this is in bytes (preferably by renaming the variable)

libp2p/Host.h
```diff
@@ -286,8 +287,8 @@ class Host: public Worker
     /// Shutdown network. Not thread-safe; to be called only by worker.
     virtual void doneWorking();
 
-    /// Get or create host identifier (KeyPair).
-    static KeyPair networkAlias(bytesConstRef _b);
+    /// Get or create host Ethereum Node record.
```
halfalicious: (Nit) host -> host's

libp2p/Host.cpp
```diff
@@ -989,13 +997,35 @@ bool Host::peerSlotsAvailable(Host::PeerSlotType _type /*= Ingress*/)
...
+    Secret secret;
     if (r.itemCount() == 3 && r[0].isInt() && r[0].toInt<unsigned>() >= 3)
-        return KeyPair(Secret(r[1].toBytes()));
+    {
+        if (r[1].isList())
+        {
+            secret = Secret{r[1][0].toBytes()};
+            auto enrRlp = r[1][1];
+
+            return make_pair(secret, parseV4ENR(enrRlp));
```
halfalicious: std:: prefix (for make_pair)?

libp2p/Host.cpp
```diff
@@ -989,13 +997,35 @@ bool Host::peerSlotsAvailable(Host::PeerSlotType _type /*= Ingress*/)
...
+        secret = Secret{Secret(r[1].toBytes())};
+    }
     else
-        return KeyPair::create();
+    {
+        // no private key found, create new one
+        secret = KeyPair::create().secret();
+    }
+
+    // TODO(gumb0): update ENR in case new address given in config
```
halfalicious: Can we add an issue ID here for tracking purposes?

libp2p/Host.cpp
```diff
@@ -989,13 +997,35 @@ bool Host::peerSlotsAvailable(Host::PeerSlotType _type /*= Ingress*/)
...
+    auto const address = _netConfig.publicIPAddress.empty() ?
+                             bi::address{} :
+                             bi::address::from_string(_netConfig.publicIPAddress);
+    return make_pair(
```
halfalicious: std::make_pair?

libp2p/Host.cpp
```diff
@@ -82,7 +82,8 @@ bytes ReputationManager::data(SessionFace const& _s, string const& _sub) const
     return bytes();
 }
 
-Host::Host(string const& _clientVersion, KeyPair const& _alias, NetworkConfig const& _n)
+Host::Host(
+    string const& _clientVersion, pair<Secret, ENR> const& _secretAndENR, NetworkConfig const& _n)
```
halfalicious: std::pair?

libp2p/Host.cpp
```diff
@@ -91,15 +92,17 @@ Host::Host(string const& _clientVersion, KeyPair const& _alias, NetworkConfig co
                      // simultaneously
     m_tcp4Acceptor(m_ioService),
     m_runTimer(m_ioService),
-    m_alias(_alias),
+    m_alias(_secretAndENR.first),
```
halfalicious: m_alias is a keypair and we need the pubkey to build the node's URL - since we're just passing in a secret now to the KeyPair ctor and secret is the private key, does this mean that the generated node URL will include the private key rather than the public key?

libp2p/Host.cpp
```diff
@@ -91,15 +92,17 @@ Host::Host(string const& _clientVersion, KeyPair const& _alias, NetworkConfig co
...
+    m_enr(_secretAndENR.second),
     m_lastPing(chrono::steady_clock::time_point::min()),
     m_capabilityHost(createCapabilityHost(*this))
 {
     cnetnote << "Id: " << id();
+    cnetnote << "ENR: " << m_enr;
```
halfalicious: In the interest of consistency should we call enr() here? (since we call id() rather than m_alias.pub())

libp2p/Host.cpp
```diff
@@ -989,13 +997,35 @@ bool Host::peerSlotsAvailable(Host::PeerSlotType _type /*= Ingress*/)
...
+        if (r[1].isList())
+        {
+            secret = Secret{r[1][0].toBytes()};
+            auto enrRlp = r[1][1];
+
+            return make_pair(secret, parseV4ENR(enrRlp));
+        }
+
+        // Support for older format without ENR
+        secret = Secret{Secret(r[1].toBytes())};
```
halfalicious: {} for nested Secret ctor?

halfalicious:
I noticed you created the tests using Gtest...is the goal to switch everything over to Gtest eventually?

libp2p/ENR.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+
+// Address can be either boost::asio::ip::address_v4 or boost::asio::ip::address_v6
+template <class Address>
+bytes addressToBytes(Address const& _address)
+{
+    auto const addressBytes = _address.to_bytes();
+    return bytes(addressBytes.begin(), addressBytes.end());
```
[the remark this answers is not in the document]
gumb0: I didn't do it, because it's a vector, and {}-initializer for vector means different thing.

In this case it actually looks to be working correct with braces, too (probably because iterator is not convertible to `byte`), but I'll leave it with parenthesis to avoid confusion.

libp2p/ENR.cpp
```diff
@@ -0,0 +1,149 @@
...
+}
+}  // namespace
+
+ENR::ENR(RLP _rlp, VerifyFunction _verifyFunction)
```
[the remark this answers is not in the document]
gumb0: I think I should pass functions by const reference, as copying them might be expensive.

But RLP is a value-semantics class (immutable and light-weight) and I prefer to pass it by value.

libp2p/ENR.cpp
```diff
@@ -0,0 +1,149 @@
...
+    if (_rlp.data().size() > c_ENRMaxSize)
+        BOOST_THROW_EXCEPTION(ENRIsTooBig());
+
+    m_signature = _rlp[0].toBytes(RLP::VeryStrict);
+
+    m_seq = _rlp[1].toInt<uint64_t>(RLP::VeryStrict);
+
+    // read key-values into vector first, to check the order
+    std::vector<std::pair<std::string const, bytes>> keyValues;
+    for (size_t i = 2; i < _rlp.itemCount(); i+= 2)
```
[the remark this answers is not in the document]
gumb0: `count - 1` looks dangerous for unsigned type (well in this case it shouldn't underflow though)

Also `_rlp[i + 1]` in the body loop should throw for out of bounds and not lead to UB.

libp2p/ENR.cpp
```diff
@@ -0,0 +1,149 @@
...
+}
+
+ENR::ENR(uint64_t _seq, std::map<std::string, bytes> const& _keyValues, SignFunction _signFunction)
+  : m_seq(_seq), m_map(_keyValues), m_signature(_signFunction(dev::ref(content())))
+{
+}
+
+bytes ENR::content() const
+{
+    RLPStream stream(contentListSize());
```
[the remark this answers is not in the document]
gumb0: (our coding style actually doesn't require it, I would say both ways are acceptable)

libp2p/ENR.cpp
```diff
@@ -0,0 +1,149 @@
...
+{
+    _s.appendList(contentListSize() + 1);
+    _s << m_signature;
+    streamContent(_s);
+}
+
+void ENR::streamContent(RLPStream& _s) const
+{
+    _s << m_seq;
+    for (auto const keyValue : m_map)
```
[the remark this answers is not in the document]
gumb0: Right, good catch

libp2p/ENR.cpp
```diff
@@ -0,0 +1,149 @@
...
+    {
+        _s << keyValue.first;
+        _s.appendRaw(keyValue.second);
+    }
+}
+
+ENR createV4ENR(Secret const& _secret, boost::asio::ip::address const& _ip, uint16_t _tcpPort,  uint16_t _udpPort)
```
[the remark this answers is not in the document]
gumb0: Well maybe it doesn't - it creates one possible kind of ENR according to one specific Identity Scheme. Maybe it should be a part of some class representing Identity Scheme, but I decided not to complicate it for now.

It's a minor detail anyway, maybe I'll change it later.

libp2p/ENR.cpp
```diff
@@ -0,0 +1,149 @@
...
+        {c_keyTCP, rlp(_tcpPort)}, {c_keyUDP, rlp(_udpPort)}};
+
+    return ENR{0, keyValues, signFunction};
+}
+
+ENR parseV4ENR(RLP _rlp)
+{
+    ENR::VerifyFunction verifyFunction = [](std::map<std::string, bytes> const& _keyValues,
+                                             bytesConstRef _signature, bytesConstRef _data) {
+        auto itID = _keyValues.find(c_keyID);
```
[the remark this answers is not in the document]
gumb0: It's a `const_iterator` anyway, I prefer not to complicate iterator declarations with additional `const`s

libp2p/ENR.cpp
```diff
@@ -0,0 +1,149 @@
...
+            return false;
+
+        auto itKey = _keyValues.find(c_keySec256k1);
+        if (itKey == _keyValues.end())
+            return false;
+
+        auto const key = RLP(itKey->second).toHash<PublicCompressed>(RLP::VeryStrict);
+        h512 const signature{_signature};
+
+        return dev::verify(key, signature, sha3(_data));
```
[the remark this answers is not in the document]
gumb0: It's a function to compute hash

libp2p/ENR.h
```diff
@@ -0,0 +1,61 @@
...
+/// Class representing Ethereum Node Record - see EIP-778 for the spec
+class ENR
+{
+public:
+    // ENR class implementation is independent of Identity Scheme.
+    // Identity Scheme specifics are passed to ENR as functions.
+
+    // Sign function gets serialized ENR contents and signs it according to some Identity Scheme
+    using SignFunction = std::function<bytes(bytesConstRef)>;
+    // Verify function gets serialized ENR contents, signature and contents and validates the
```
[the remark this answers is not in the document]
gumb0: I'll reformulate it

libp2p/ENR.h
```diff
@@ -0,0 +1,61 @@
...
+    using VerifyFunction =
+        std::function<bool(std::map<std::string, bytes> const&, bytesConstRef, bytesConstRef)>;
+
+    // Parse from RLP with given signature verification function
+    ENR(RLP _rlp, VerifyFunction _verifyFunction);
+    // Create with given sign function
+    ENR(uint64_t _seq, std::map<std::string, bytes> const& _keyValues, SignFunction _signFunction);
+
+    uint64_t sequenceNumber() const { return m_seq; }
+    std::map<std::string, bytes> const& keyValues() const { return m_map; }
```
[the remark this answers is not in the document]
gumb0: Agree, I'll change it.

libp2p/ENR.h
```diff
@@ -0,0 +1,61 @@
...
+    // Serialize to given RLP stream
+    void streamRLP(RLPStream& _s) const;
+
+private:
+    uint64_t m_seq = 0;
+    std::map<std::string, bytes> m_map;
+    bytes m_signature;
+
+    bytes content() const;
+    size_t contentListSize() const { return m_map.size() * 2 + 1; }
```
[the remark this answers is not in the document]
gumb0: Ok, I'll call it `contentRlpListItemCount` as `RLP` has `itemCount()` method.

(Also note that "content" is a term from the spec, which is defined as RLP-list)

libp2p/Host.cpp
gumb0: This file has `using namespace std`, so I follow surrounding code's convention

libp2p/Host.cpp
gumb0: I've created an issue covering all ENR Extension tasks https://github.com/ethereum/aleth/issues/5551, will link it here

libp2p/Host.cpp
gumb0: No. Secret key and public key are different types (`Secret` and `Public`) and `KeyPair` constructor derives public key from secret key.

It was actually the same before, it happened in `Host::networkAlias` method

libp2p/Host.cpp
gumb0: Actually it should be just one

libp2p/ENR.cpp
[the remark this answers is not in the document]
halfalicious: I think _rlp[i + 1] would only throw if it happened to point to protected memory (unless you have something like[ full page heap enabled](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/gflags-and-pageheap)).

libp2p/ENR.cpp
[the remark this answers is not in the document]
halfalicious: Ah okay, I thought that was the preferred way of doing things since I remember it being called out in at least one of my previous pull requests 😄

libp2p/ENR.cpp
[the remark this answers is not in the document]
halfalicious: Ah, you're right, the map is const-qualified. Agreed, no need for an additional const.

libp2p/Host.cpp
halfalicious: Okay, I called this out since I saw some use of std:: in the file...but if the std:: namespace is being imported then ultimately it doesn't matter.

libp2p/Host.cpp
halfalicious: Makes sense, thanks for the clarification!

libp2p/ENR.cpp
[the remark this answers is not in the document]
halfalicious: Ah, right you are 😄

libp2p/ENR.cpp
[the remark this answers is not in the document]
gumb0: I think you confuse it with arrays/vectors.

Actually I was not right - `RLP::operator[]` will return empty `RLP` for out of bounds - I forget this all the time. So in this case it will end up with `key => ""` if last value is absent, which is not ideal, but it's not an UB, I think I'll merge it like this for now.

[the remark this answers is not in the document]
gumb0: I've added some comments to `RLP::operator[]` which is not very straight-forward https://github.com/ethereum/aleth/pull/5555/commits/a791c5b2427606da9bde536b86c74b36f4df5d29

Issue #5531, opened 2019-03-26
Remove --private option from aleth


halfalicious:
Aleth's `--private` option is unnecessary and should be removed - it sets some chain parameters to certain values, but in order to use a private chain one must also specify a chain configuration json file (via the `--config` parameter) and this config file must also include values for those chain parameters.

The private option's usage is logged here: https://github.com/ethereum/aleth/blob/523c965e9db1c9a1f6dde7ff7a042c28f6e5f13c/aleth/main.cpp#L211

The private option is parsed here: https://github.com/ethereum/aleth/blob/ad6f54f662fc7ddeac8bfd2c115362675e4800db/aleth/main.cpp#L626-L635

Chain parameters are then set to certain values (based on whether or not `--private` was supplied) here: https://github.com/ethereum/aleth/blob/ad6f54f662fc7ddeac8bfd2c115362675e4800db/aleth/main.cpp#L698-L703

The chain configuration file is processed here: https://github.com/ethereum/aleth/blob/ad6f54f662fc7ddeac8bfd2c115362675e4800db/aleth/main.cpp#L681-L694

twinstar26:
To resolve this issue,

I am planning to remove
https://github.com/ethereum/aleth/blob/523c965e9db1c9a1f6dde7ff7a042c28f6e5f13c/aleth/main.cpp#L211
so that "private" key-value pair is not created in map "vm".

Then I am going to move the contents of
https://github.com/ethereum/aleth/blob/523c965e9db1c9a1f6dde7ff7a042c28f6e5f13c/aleth/main.cpp#L626-L635
to
https://github.com/ethereum/aleth/blob/523c965e9db1c9a1f6dde7ff7a042c28f6e5f13c/aleth/main.cpp#L468-L486
with appropriate condition checking, but the string "privateChain" which used to contain the value of key "private" from map "vm" will now remain NULL. 
So is this information going to be included in "configJSON"? Or we can modify 
https://github.com/ethereum/aleth/blob/523c965e9db1c9a1f6dde7ff7a042c28f6e5f13c/aleth/main.cpp#L213
to "value<string>()->multitoken()" so as to obtain --config value as well as --private value (if it exists).

Please pardon my noviceness as this is my first open source issue.

twinstar26:
@halfalicious

halfalicious:
Thank you for your interest in fixing this issue @twinstar26 ! 

[...]

PR #5538, opened 2019-03-31
Removed --private option from aleth

twinstar26:
Issue Name: Remove --private option from aleth #5531 
--private option from aleth has been removed.

### DIFF
--- a/aleth/main.cpp
+++ b/aleth/main.cpp
@@ -148,7 +148,6 @@ int main(int argc, char** argv)
     std::map<p2p::NodeID, pair<NodeIPEndpoint, bool>> preferredNodes;
     bool bootstrap = true;
     bool disableDiscovery = false;
-    bool enableDiscovery = false;
     bool allowLocalDiscovery = false;
     static const unsigned NoNetworkID = (unsigned)-1;
     unsigned networkID = NoNetworkID;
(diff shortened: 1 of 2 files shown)


halfalicious:
Can you please update your PR description to mention the issue that you're fixing (and please be sure to include a "#" before the issue ID so that the issue is automatically updated with a link to this PR)?

halfalicious:
Can you also please remove the definition for `enableDiscovery` and remove this line in main where it's set to `false`:
https://github.com/ethereum/aleth/blob/4f638b0508e22a6c0ec0a9cd7e865548fc086e18/aleth/main.cpp#L378-L384
Also please replace the check for `enableDiscovery `here with a check for `!disableDiscovery`: https://github.com/ethereum/aleth/blob/4f638b0508e22a6c0ec0a9cd7e865548fc086e18/aleth/main.cpp#L926

aleth/main.cpp
```diff
@@ -756,7 +736,7 @@ int main(int argc, char** argv)
     };
 
     auto netPrefs = publicIP.empty() ? NetworkConfig(listenIP, listenPort, upnp) : NetworkConfig(publicIP, listenIP ,listenPort, upnp);
-    netPrefs.discovery = (privateChain.empty() && !disableDiscovery) || enableDiscovery;
+    netPrefs.discovery = !disableDiscovery;
```
chfast: How do the options for enable/disable discovery look like now?

PR #5539, opened 2019-04-01
Fix logic used to determine if Aleth should request dao fork block header from potential peer

### LINKED ISSUE #5271: Sync in private net tries to request the DAO fork block
Reported in https://github.com/ethereum/aleth/issues/5263

When `daoHardforkBlock` is missing from config, it has some default non-zero value (`c_infiniteBlockNumer`) and the code in `BlockChainSync` requests DAO fork block when starting to sync from the new peer always when DAO fork block number is non-zero.

I think we should check in `BlockChainSync` for both 0 and `c_infiniteBlockNumer`

halfalicious:
Fix #5271 

Update the logic which Aleth uses to determine if it should request the DAO fork block header from a peer before starting a sync based on the value of the `daoHardFork ` chain configuration parameter. Aleth needs to check for 0 or `c_infiniteBlockNumber ` (rather than just 0) because Aleth initializes the  `daoHardFork `configuration parameter to `c_infiniteBlockNumber ` by default.

The impact of these changes is that one doesn't have to explicitly specify `daoHardFork `= 0x0 in any `config.json` that they wish to use with Aleth.

### DIFF
--- a/libethcore/ChainOperationParams.h
+++ b/libethcore/ChainOperationParams.h
@@ -63,7 +63,7 @@ private:
     u256 m_startingBlock = 0;
 };
 
-static constexpr int64_t c_infiniteBlockNumber = std::numeric_limits<int64_t>::max();
+constexpr int64_t c_infiniteBlockNumber = std::numeric_limits<int64_t>::max();
 
 struct ChainOperationParams
 {
--- a/libethcore/ChainOperationParams.h
+++ b/libethcore/ChainOperationParams.h
@@ -63,7 +63,7 @@ private:
     u256 m_startingBlock = 0;
 };
 
-static constexpr int64_t c_infiniteBlockNumer = std::numeric_limits<int64_t>::max();
+static constexpr int64_t c_infiniteBlockNumber = std::numeric_limits<int64_t>::max();
 
 struct ChainOperationParams
 {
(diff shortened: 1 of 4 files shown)


libethcore/ChainOperationParams.h
```diff
@@ -63,7 +63,7 @@ class PrecompiledContract
     u256 m_startingBlock = 0;
 };
 
-static constexpr int64_t c_infiniteBlockNumer = std::numeric_limits<int64_t>::max();
+static constexpr int64_t c_infiniteBlockNumber = std::numeric_limits<int64_t>::max();
```
chfast: The `static` is redundant for `constexpr`.

libethereum/BlockChainSync.cpp
```diff
@@ -52,6 +30,10 @@ std::ostream& dev::eth::operator<<(std::ostream& _out, SyncStatus const& _sync)
 namespace  // Helper functions.
 {
 
+unsigned constexpr c_maxPeerUknownNewBlocks = 1024; /// Max number of unknown new blocks peer can give us
```
chfast: Pedantic mode: should be `constexpr unsigned`.

libethcore/ChainOperationParams.h
halfalicious: I thought that `static `and `constexpr `are orthogonal concepts? I agree that `static `doesn't make sense here (regardless of the constexpr) since this is a header file and `c_infiniteBlockNumber` is defined outside of a class - from what I understand, static when used this way limits the scope of the variable to the translation unit which means that each cpp file which includes the header will have a separate instance of c_infiniteBlockNumber (though each instance will have the same value) which seems unnecessary.

libethereum/BlockChainSync.cpp
halfalicious: Why do we put `constexpr `before the type but `const `after the type?

libethcore/ChainOperationParams.h
chfast: I'm also confused now. On the language level that might be true. But also `constexpr` implies `inline` and `inline` implies `static`. 

I check than in practice the `constexpr` symbols are not exported.

PR #5542, opened 2019-04-02
Verify signature with compressed public key

gumb0:
We'll need to verify ENR signature having only compressed public key (33 bytes) and `[r, s]` part of the signature (64 bytes)

For reference go-ethereum code doing the same: https://github.com/ethereum/go-ethereum/blob/c4109d790ffd26d67feb7745d4af8a8bc5090bd9/crypto/secp256k1/ext.h#L46-L63

### DIFF
--- a/libdevcrypto/Common.h
+++ b/libdevcrypto/Common.h
@@ -133,6 +133,9 @@ Signature sign(Secret const& _k, h256 const& _hash);
 /// Verify signature.
 bool verify(Public const& _k, Signature const& _s, h256 const& _hash);
 
+// Verify signature with compressed public key
+bool verify(PublicCompressed const& _key, h512 const& _signature, h256 const& _hash);
+
 /// Derive key via PBKDF2.
 bytesSec pbkdf2(std::string const& _pass, bytes const& _salt, unsigned _iterations, unsigned _dkLen = 32);
 
--- a/test/unittests/libdevcrypto/crypto.cpp
+++ b/test/unittests/libdevcrypto/crypto.cpp
@@ -1,18 +1,18 @@
 /*
-	This file is part of cpp-ethereum.
+    This file is part of cpp-ethereum.
 
-	cpp-ethereum is free software: you can redistribute it and/or modify
-	it under the terms of the GNU General Public License as published by
-	the Free Software Foundation, either version 3 of the License, or
-	(at your option) any later version.
+    cpp-ethereum is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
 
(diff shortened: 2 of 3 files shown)


libdevcrypto/Common.h
```diff
@@ -133,6 +133,9 @@ Signature sign(Secret const& _k, h256 const& _hash);
 /// Verify signature.
 bool verify(Public const& _k, Signature const& _s, h256 const& _hash);
 
+// Verify signature with compressed public key
+bool verify(PublicCompressed const& _key, h512 const& _signature, h256 const& _hash);
```
halfalicious: Should we name the args similar to the other verify function i.e. _k/_s?

test/unittests/libdevcrypto/crypto.cpp
```diff
@@ -181,6 +181,34 @@ BOOST_AUTO_TEST_CASE(verify_secert)
 	BOOST_REQUIRE(k.address());
 }
 
+BOOST_AUTO_TEST_CASE(verifyWithPublicCompressed)
+{
+    PublicCompressed pub{
```
halfalicious: Presumably you encrypted the hash into the hex value you're using as the signature value outside of the test case...can we include the encryption as part of the test?

test/unittests/libdevcrypto/crypto.cpp
```diff
@@ -181,6 +181,34 @@ BOOST_AUTO_TEST_CASE(verify_secert)
...
+    BOOST_REQUIRE(verify(pub, sig, msg));
+
+    h512 invalidSig{sig};
+    ++invalidSig[4];
+    BOOST_REQUIRE(!verify(pub, invalidSig, msg));
+}
+
+BOOST_AUTO_TEST_CASE(signAndVerify)
+{
+    KeyPair kp = KeyPair::create();
```
halfalicious: (Nit) can be auto

test/unittests/libdevcrypto/crypto.cpp
```diff
@@ -181,6 +181,34 @@ BOOST_AUTO_TEST_CASE(verify_secert)
...
+    auto msg = h256::random();
```
halfalicious: (Nit) Can msg and sig be const?

test/unittests/libdevcrypto/crypto.cpp
```diff
@@ -181,6 +181,34 @@ BOOST_AUTO_TEST_CASE(verify_secert)
...
+    auto sig = sign(kp.secret(), msg);
+
+    auto pubCompressed = toPublicCompressed(kp.secret());
```
halfalicious: (Nit) can pubCompressed be const?

test/unittests/libdevcrypto/crypto.cpp
gumb0: It's what `signAndVerify` test does

PR #5548, opened 2019-04-05
Fix rlp tool failing to parse long hex string

### LINKED ISSUE #5543: rlp tool fails to parse long hex string
```
> rlp/rlp 82abcd # short string is fine
"0xabcd"

> rlp/rlp  f884b8407098ad865b00a582051940cb9cf36836572411a47278783077011599ed5cd16b76f2635f4e234738f30813a89eb9137e3e3df5266e3a1f11df72ecf1145ccb9c01826964827634826970847f00000189736563703235366b31a103ca634cae0d49acb401d8a4c6b6fe8c55b70d115bf400769cc1400f3258cd31388375647082765f
terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
  what():  boost::filesystem::status: File name too long: "f884b8407098ad865b00a582051940cb9cf36836572411a47278783077011599ed5cd16b76f2635f4e234738f30813a89eb9137e3e3df5266e3a1f11df72ecf1145ccb9c01826964827634826970847f00000189736563703235366b31a103ca634cae0d49acb401d8a4c6b6fe8c55b70d115bf400769cc1400f3258cd31388375647082765f"
```

I think `boost::filesystem::is_regular_file` throws unhandled exception in this case. https://github.com/ethereum/aleth/blob/b09e9ad00b1f34259c0702ca521cd2e5585d526c/rlp/main.cpp#L262

twinstar26:
Fixes #5543

### DIFF
--- a/rlp/main.cpp
+++ b/rlp/main.cpp
@@ -12,6 +12,7 @@
 #include <libethcore/Common.h>
 #include <boost/algorithm/string.hpp>
 #include <boost/filesystem.hpp>
+#include <boost/system/error_code.hpp>
 #include <boost/program_options.hpp>
 #include <boost/program_options/options_description.hpp>
 #include <clocale>
@@ -259,10 +260,20 @@ int main(int argc, char** argv)
     if (inputFile == "--")
         for (int i = cin.get(); i != -1; i = cin.get())
             in.push_back((byte)i);
-    else if (boost::filesystem::is_regular_file(inputFile))
-        in = contents(inputFile);
     else
-        in = asBytes(inputFile);
+    {
+        try
+        {
+            if (boost::filesystem::is_regular_file(inputFile))
+                in = contents(inputFile);
+        }
+        catch (boost::filesystem::filesystem_error const&)
+        {
+        }
+
+        if (in.empty())
+            in = asBytes(inputFile);
+    }
 
     bytes b;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,5 +16,6 @@
 - Fixed: [#5523](https://github.com/ethereum/aleth/pull/5523) [#5533](https://github.com/ethereum/aleth/pull/5533) Fix syncing terminating prematurely because of race condition.
 - Fixed: [#5539](https://github.com/ethereum/aleth/pull/5539) Fix logic for determining if dao hard fork block header should be requested.
 - Fixed: [#5547](https://github.com/ethereum/aleth/pull/5547) Fix unnecessary slow-down of eth_flush RPC method.
+- Fixed: [#5548](https://github.com/ethereum/aleth/pull/5548) Fix rlp tool for long hexadecimal string inputs.
 
 [1.6.0]: https://github.com/ethereum/aleth/compare/v1.6.0-alpha.1...release/1.6


rlp/main.cpp
```diff
@@ -256,14 +256,22 @@ int main(int argc, char** argv)
...
+            for (int i = cin.get(); i != -1; i = cin.get())
+                in.push_back((byte)i);
+        else if (boost::filesystem::is_regular_file(inputFile))
+            in = contents(inputFile);
+    }
+    catch(const boost::filesystem::filesystem_error& _e)
+    {
+        //filesyatem_error contains lot of error flags but I couldn't find file name too long
+        //the closest I got was boost::system::errc -> enum errc_t ->  filename_too_long = ENAMETOOLONG
+        if(_e.what() == "File name too long") //don't know if this will work. Let's say temporary workaround.
```
jwasinger: Because `what()` returns a `char *`  you can't compare directly to a string literal here.  You need to wrap the result of `what` in a string object before doing the comparison.

twinstar26: Thanks for pointing out. My bad, sorry.
But the problem still remains unaddressed, if what I am doing (Directly comparing strings) will solve this issue.
I will correct my error soon.

rlp/main.cpp
```diff
@@ -256,14 +256,24 @@ int main(int argc, char** argv)
...
+        else if (boost::filesystem::is_regular_file(inputFile))
+            in = contents(inputFile);
+    }
+    catch(const boost::filesystem::filesystem_error& _e)
+    {
+        //temporary workaround by comparing exception message with inputFile
+        string error_e = _e.what();
+        string long_file_name = "boost::filesystem::status: File name too long: \"";
+        long_file_name += inputFile + "\"";
+        if(error_e == long_file_name)
```
jwasinger: You can probably assume that the input should be treated as bytecode if you get **any** error from `is_regular_file`.  Probably don't need to check the error message or type.

jwasinger: And checking for error types by comparing against hard coded strings is bad form because it will break if the error message is changed upstream in Boost.

twinstar26: Yeah I totally agree with you regarding the hard coded strings and that change in upstream boost will break code.
But,
I thought that in here
https://github.com/ethereum/aleth/blob/b09e9ad00b1f34259c0702ca521cd2e5585d526c/rlp/main.cpp#L259-L265
contents() internally call
https://github.com/ethereum/aleth/blob/ad7204c9ae7a13359520b7c21ffd8fc7484cf589/libdevcore/CommonIO.cpp#L86-L105
contentsGeneric() and here it's performing boost::filesystem operations so boost::filesystem::filesystem_error may occur and then catching every error and handling them as filename_too_long will create another bug.

Please correct me if I am wrong.
Also I have some ideas, but I need some time to select the best out of them.

twinstar26:
Hi,
I guess this latest commit will solve this issue.
Also if there are any exceptions to be handled please let me know.

rlp/main.cpp
```diff
@@ -256,14 +257,20 @@ int main(int argc, char** argv)
...
-
+    try
+    {
+        if (inputFile == "--")
+            for (int i = cin.get(); i != -1; i = cin.get())
+                in.push_back((byte)i);
+        else if (boost::filesystem::is_regular_file(inputFile))
+            in = contents(inputFile);
+    }
+    catch(const boost::filesystem::filesystem_error& _e)
```
gumb0:
```suggestion
    catch (const boost::filesystem::filesystem_error& e)
```
See https://github.com/ethereum/aleth/blob/master/CODING_STYLE.md

rlp/main.cpp
```diff
@@ -256,14 +257,20 @@ int main(int argc, char** argv)
...
+    {
+        if(_e.code() == boost::system::errc::errc_t::filename_too_long)
```
gumb0: It's ok to handle all file system errors the same way (errors from `contents`, too), not only `filename_too_long`

rlp/main.cpp
```diff
@@ -256,14 +257,20 @@ int main(int argc, char** argv)
...
     bytes in;
-    if (inputFile == "--")
-        for (int i = cin.get(); i != -1; i = cin.get())
-            in.push_back((byte)i);
-    else if (boost::filesystem::is_regular_file(inputFile))
-        in = contents(inputFile);
-    else
-        in = asBytes(inputFile);
-
+    try
```
gumb0: Don't make try-catch block wider than it needs to be - if you intend to handle errors from `is_regular_file` only (maybe together with `contents()`) - wrap only those functions in try-catch

twinstar26:
@gumb0 
So keeping in mind all the past comments,
`     try`
`     {`
`        if (inputFile == "--")`
`            for (int i = cin.get(); i != -1; i = cin.get())`
`                in.push_back((byte)i);`    
`        else if (boost::filesystem::is_regular_file(inputFile))`
`            in = contents(inputFile);`
`        else`
`        {`
`            in = asBytes(inputFile);`
`        }`        
`    }`
`    catch (const boost::filesystem::filesystem_error& e)`
`    {`
`            in = asBytes(inputFile);`
`    }`
    
is my modified code.
But I am not creating a pull request as during my tests,
`ubuntu18@ubuntu18-Inspiron-3268:~/opensource/aleth/build/rlp$ ./rlp f9037b04f8a8a00064e8a4f925e944185338c03a1b913c3bcc7e762f5ba5bde22fc17df59215c5f885b84017163d503ea69bca3d349654b21080e2431cf1102c8907841b43efa96e33b76012a3f8a1ec1851945d077a7e6b13d577f85892150a8fbd7579f539a06c53ee508082696476348269700d4abd93887365633235366b31036bffe5e6caa66422eccb705c6f1d18ec272b7e21d5648d2ca213be1888b3117083746370765f83756470765ff902cdf8578434e8f39882765f82765fb8406332792c4a00e3e4ee0926ed89e0d27ef985424d97b6a45bf0f23e51f0dcb5e66b875777506458aea7af6f9e4ffb69f43f3778ee73c81ed9d34c51c4b16b0b0f845ca24b9d845ca24b9df857843410bcb982765f82765fb840a979fb575495b8d6db44f750317d0f4622bf4c2aa3365d6af7c284339968eef29b69ad0dce72a4d8db5ebb4968de0e3bec910127f134779fbcb0cb6d3331163c845ca24b9c845ca24b9cf857840d4b9a8a82765f82765fb840158f8aab45f6d19c6cbf4a089c2670541a8da11978a2f90dbf6a502a4a3bab80d288afdbeb7ec0ef6d92de563767f3b1ea9e8e334ca711e9f8e2df5a0385e8e6845ca24b9c845ca24b9cf8578434b0644d82765f82765fb840865a63255b3bb68023b6bffd5095118fcc13e79dcf014fe4e47e065c350c7cc72af2e53eff895f11ba1bbb6a2b33271c1116ee870f266618eadfc2e78aa7349c845ca24b9c845ca24b9ff8578434b0070a82765f82765fb84015ac307a470b411745a6f10544ed54c0a14ad640b21f04f523e736e732bf709d4e28c2f06526ecabc03eed226b6d9bee8e433883cd20ab6cbd114bab77a8775d845ca24bb5845ca24b9cf857840d5dd35482765f82765fb840aa36fdf33dd030378a0168efe6ed7d5cc587fafa3cdd375854fe735a2e11ea3650ba29644e2db48368c46e1f60e716300ba49396cd63778bf8a818c09bded46f845ca24bb5845ca24b9cf85784344a397b82765f82765fb840669f45b66acf3b804c26ce13cfdd1f7e3d0ff4ed85060841b9af3af6dbfbacd05181e1c9363161446a307f3ca24e707856a01e4bf1eed5e1aefc14011a5c1c1c845ca24bb5845ca24b9cf85c8434b0644d82765f82765fb840865a63255b3bb68023b6bffd5095118fcc13e79dcf014fe4e47e065c350c7cc72af2e53eff895f11ba1bbb6a2b33271c1116ee870f266618eadfc2e78aa7349c01845ca24bb4845ca24bba04108080`
`[ "0x04", [ "0x0064e8a4f925e944185338c03a1b913c3bcc7e762f5ba5bde22fc17df59215c5", [ "0x17163d503ea69bca3d349654b21080e2431cf1102c8907841b43efa96e33b76012a3f8a1ec1851945d077a7e6b13d577f85892150a8fbd7579f539a06c53ee50", "0x", "0x6964", "0x76", "0x34", "0x6970", "0x0d", "0x4a"Error: Invalid format; bad RLP.`
After a lot of debugging i found that
https://github.com/ethereum/aleth/blob/a611afea888a44bd01acb958b14ef7a256e0373a/rlp/main.cpp#L412-L418
is getting executed and as a result its falling into this catch block
https://github.com/ethereum/aleth/blob/a611afea888a44bd01acb958b14ef7a256e0373a/rlp/main.cpp#L482-L486
Declaration can be found over here
https://github.com/ethereum/aleth/blob/a611afea888a44bd01acb958b14ef7a256e0373a/rlp/main.cpp#L74-L108

I am guessing this is not a normal situation...
Can you please provide me with more reference/insights regarding this new issue.

gumb0:
@twinstar26 That was incorrect RLP I gave in the issue description, so this output is actually expected, sorry for this confusion.

Try this instead:
```
rlp/rlp  f884b8407098ad865b00a582051940cb9cf36836572411a47278783077011599ed5cd16b76f2635f4e234738f30813a89eb9137e3e3df5266e3a1f11df72ecf1145ccb9c01826964827634826970847f00000189736563703235366b31a103ca634cae0d49acb401d8a4c6b6fe8c55b70d115bf400769cc1400f3258cd31388375647082765f
```

twinstar26:
@gumb0 
So i tried with the new RLP and this is my new output
`ubuntu18@ubuntu18-Inspiron-3268:~/opensource/aleth/build/rlp$ ./rlp f884b8407098ad865b00a582051940cb9cf36836572411a47278783077011599ed5cd16b76f2635f4e234738f30813a89eb9137e3e3df5266e3a1f11df72ecf1145ccb9c01826964827634826970847f00000189736563703235366b31a103ca634cae0d49acb401d8a4c6b6fe8c55b70d115bf400769cc1400f3258cd31388375647082765f
[ "0x7098ad865b00a582051940cb9cf36836572411a47278783077011599ed5cd16b76f2635f4e234738f30813a89eb9137e3e3df5266e3a1f11df72ecf1145ccb9c", "0x01", "0x6964", "0x7634", "0x6970", "0x7f000001", "0x736563703235366b31", "0x03ca634cae0d49acb401d8a4c6b6fe8c55b70d115bf400769cc1400f3258cd3138", "0x756470", "0x765f" ]
`
This time it did not enter catch block!!

twinstar26:
For this pull request,
Building failed on visual studio for commit "b695dee" due to " 'e': unreferenced local variable " (line 270)

Then for commit "1b2f222" Test Case "neighbours" failed (/Users/distiller/project/test/unittests/libp2p/net.cpp:1253: last checkpoint: "neighbours" test entry) on xcode

I failed to understand the connection between the two.

gumb0:
@twinstar26 there's no connection between these failures. Please ignore the second one (`neighbours`), this test is not stable lately (https://github.com/ethereum/aleth/issues/5544 tracks this)

twinstar26:
@gumb0 I reverted back. This was so embarrassing! :sweat_smile:

gumb0:
Changed base to `release/1.6` to include it in release.

PR #5558, opened 2019-04-10
testeth check for unused folders in test fillers

winsvega:
example on StateTests

I noticed that sometimes test suites are forgotten when copying boost test suite declarations from `GeneralStateTests` into `BCGeneralStateTests`. It seems to be unable to define the list of test cases at one place one time and copy, so then testeth could check if there are unused folders in test fillers. 

Also checking for unused test folders in other test suites would be nice.

### DIFF
--- a/test/tools/jsontests/StateTests.cpp
+++ b/test/tools/jsontests/StateTests.cpp
@@ -142,16 +142,23 @@ class GeneralTestFixture
 public:
 	GeneralTestFixture()
 	{
-		test::StateTestSuite suite;
-		string casename = boost::unit_test::framework::current_test_case().p_name;
-		if (casename == "stQuadraticComplexityTest" && !test::Options::get().all)
-		{
-			std::cout << "Skipping " << casename << " because --all option is not specified.\n";
-			return;
-		}
-		suite.runAllTestsInFolder(casename);
-	}
+        test::StateTestSuite suite;
+        string casename = boost::unit_test::framework::current_test_case().p_name;
+        boost::filesystem::path suiteFillerPath = suite.getFullPathFiller(casename).parent_path();
+
+        // Check specific test cases
+        static vector<string> timeConsumingTestSuites{string{"stQuadraticComplexityTest"}};
+        if (test::inArray(timeConsumingTestSuites, casename) && !test::Options::get().all)
+        {
+            std::cout << "Skipping " << casename << " because --all option is not specified.\n";
+            test::TestOutputHelper::get().markTestFolderAsFinished(suiteFillerPath, casename);
+            return;
+        }
+        suite.runAllTestsInFolder(casename);
+        test::TestOutputHelper::get().markTestFolderAsFinished(suiteFillerPath, casename);
+    }
 };
+
 
 BOOST_FIXTURE_TEST_SUITE(GeneralStateTests, GeneralTestFixture)
 
--- a/test/tools/jsontests/TransactionTests.cpp
+++ b/test/tools/jsontests/TransactionTests.cpp
@@ -219,9 +219,11 @@ class TransactionTestFixture
 public:
     TransactionTestFixture()
     {
-        string const& casename = boost::unit_test::framework::current_test_case().p_name;
         test::TransactionTestSuite suite;
+        string const casename = boost::unit_test::framework::current_test_case().p_name;
+        boost::filesystem::path suiteFillerPath = suite.getFullPathFiller(casename).parent_path();
         suite.runAllTestsInFolder(casename);
+        test::TestOutputHelper::get().markTestFolderAsFinished(suiteFillerPath, casename);
     }
 };
 
--- a/test/tools/libtesteth/TestOutputHelper.cpp
+++ b/test/tools/libtesteth/TestOutputHelper.cpp
@@ -18,12 +18,13 @@
  * Fixture class for boost output when running testeth
  */
 
-#include <boost/test/unit_test.hpp>
-#include <boost/io/ios_state.hpp>
 #include <libethashseal/Ethash.h>
 #include <libethcore/BasicAuthority.h>
-#include <test/tools/libtesteth/TestOutputHelper.h>
 #include <test/tools/libtesteth/Options.h>
+#include <test/tools/libtesteth/TestOutputHelper.h>
+#include <boost/filesystem.hpp>
+#include <boost/io/ios_state.hpp>
+#include <boost/test/unit_test.hpp>
 #include <numeric>
 
 using namespace std;
@@ -84,6 +85,7 @@ void TestOutputHelper::finishTest()
 
 void TestOutputHelper::printTestExecStats()
 {
+    checkUnfinishedTestFolders();
     if (Options::get().exectimelog)
     {
         boost::io::ios_flags_saver saver(cout);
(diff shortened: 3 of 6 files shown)


gumb0:
Windows build failed.

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -100,3 +108,32 @@ void TestOutputHelper::printTestExecStats()
...
+void TestOutputHelper::registerTestFolders(boost::filesystem::path _path)
+{
+    // -t SuiteName/caseName   parse caseName as filter
+    std::string filter;
+    std::string const& tSuiteString = Options::get().rCurrentTestSuite;
+    std::size_t pos = tSuiteString.find('/');
+    if (pos != string::npos)
+        filter = tSuiteString.substr(pos + 1, tSuiteString.length() - pos);
+
+    using fsIterator = boost::filesystem::directory_iterator;
```
gumb0: Not sure if it's a good idea to browse the file system in a class responsible for helping with output, but fine.

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -100,3 +108,32 @@ void TestOutputHelper::printTestExecStats()
...
 }
+
+void TestOutputHelper::registerTestFolders(boost::filesystem::path _path)
+{
+    // -t SuiteName/caseName   parse caseName as filter
+    std::string filter;
+    std::string const& tSuiteString = Options::get().rCurrentTestSuite;
+    std::size_t pos = tSuiteString.find('/');
+    if (pos != string::npos)
+        filter = tSuiteString.substr(pos + 1, tSuiteString.length() - pos);
```
gumb0: can be just `tSuiteString.substr(pos + 1)`

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -100,3 +108,32 @@ void TestOutputHelper::printTestExecStats()
...
+
+    using fsIterator = boost::filesystem::directory_iterator;
+    for (fsIterator it(_path); it != fsIterator(); ++it)
+    {
+        if (!filter.empty() && it->path().filename() == filter)
```
gumb0: You don't really need to iterate the directory if `filter` is set, right?
What about something like (pseudocode):
```
if (filter.empty())
{
  for (fsIterator it(_path); it != fsIterator(); ++it)
    if (is_directory(it))
      m_testSuiteStatus.emplace(std::make_pair(it->path().filename().c_str(), false));
}
else
{
  if (is_directory(_path + filter))
      m_testSuiteStatus.emplace(std::make_pair(filter, false));
}
    
```

test/tools/jsontests/StateTests.cpp
```diff
@@ -142,15 +142,28 @@ class GeneralTestFixture
...
-			std::cout << "Skipping " << casename << " because --all option is not specified.\n";
-			return;
-		}
-		suite.runAllTestsInFolder(casename);
-	}
+        test::StateTestSuite suite;
+        string casename = boost::unit_test::framework::current_test_case().p_name;
+
+        // Mark all test folders to run
+        static bool registerTestFolders = false;
```
gumb0: Why is this needed? Can you instead call `registerTestFolders` for a parent path once from the level above, like from the main function?

If not, better to make this check inside `registerTestFolders`

test/tools/libtesteth/TestOutputHelper.h
```diff
@@ -62,6 +66,7 @@ class TestOutputHelper
 	boost::filesystem::path m_currentTestFileName;
 	typedef std::pair<double, std::string> execTimeName;
 	std::vector<execTimeName> m_execTimeResults;
+    std::map<std::string, bool> m_testSuiteStatus;
```
gumb0: What's the "status"? I know now after reading the code, but it's not clear at all when you just look at this.

Alternative design suggestion: make this `std::vector<std::string> m_finishedTestSuites`; get rid of `registerTestFolders`; instead collect this list of test suit folders in the end when it's finished, put it into another `std::vector<std::string>`, use `std::set_difference` to find the folders present on disk but absent in `m_finishedTestSuites`

gumb0:
There's also `traverse_test_tree` thing, which should allow to collect all test suite names without hacks all over the code.
https://stackoverflow.com/a/36461218

winsvega:
Thanks will look at that option too

winsvega:
looks like `BOOST_GLOBAL_FIXTURE( GlobalFixture ) ` could not be combined with `BOOST_FIXTURE_TEST_SUITE(GeneralStateTests, GeneralTestFixture)`

test/tools/libtesteth/TestOutputHelper.h
winsvega: thanks!

winsvega:
added a small unrelated change: 
```
        static vector<string> timeConsumingTestSuites;
        if (timeConsumingTestSuites.size() == 0)
        {
            timeConsumingTestSuites.push_back("stStaticCall");
            timeConsumingTestSuites.push_back("stSStoreTest");
            timeConsumingTestSuites.push_back("stQuadraticComplexityTest");
        }
        if (test::inArray(timeConsumingTestSuites, casename) && !test::Options::get().all)
```

rafactored the logic to your suggestions. checking the folders at the execution end instead of registering it within the test fixture. 

added a code path work if `-t` argument not provided

gumb0:
Windows build still fails.

test/tools/libtesteth/TestOutputHelper.h
```diff
@@ -62,6 +67,8 @@ class TestOutputHelper
 	boost::filesystem::path m_currentTestFileName;
 	typedef std::pair<double, std::string> execTimeName;
 	std::vector<execTimeName> m_execTimeResults;
+    typedef std::set<std::string> finishedTestFoldersType;
```
gumb0: Type names start with a capital letter

gumb0: Also should finish with `...Set`. I'd say this is `FolderNameSet`

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -100,3 +102,50 @@ void TestOutputHelper::printTestExecStats()
...
+        else
+        {
+            if (boost::filesystem::is_directory(path / filter))
+                allFolders.emplace(filter);
+        }
+
+        std::vector<string> diff;
+        finishedTestFoldersType finishedFolders = allTestsIt.second;
+        std::set_difference(allFolders.begin(), allFolders.end(), finishedFolders.begin(),
+            finishedFolders.end(), std::inserter(diff, diff.begin()));
```
gumb0: Could be `std::back_inserter(diff)`

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -100,3 +102,50 @@ void TestOutputHelper::printTestExecStats()
...
+        {
+            for (fsIterator it(path); it != fsIterator(); ++it)
+            {
+                if (boost::filesystem::is_directory(*it))
+                    allFolders.emplace(it->path().filename().c_str());
+            }
+        }
+        else
+        {
+            if (boost::filesystem::is_directory(path / filter))
```
gumb0: Hm this is a bit weird - to do this inside the loop and then to do `set_difference` when we know it should be just one item.

Maybe better to move this special case with filter out of `for` loop and just check there whether `m_finishedTestFoldersMap` contains one expected element.

test/tools/libtesteth/TestOutputHelper.h
```diff
@@ -52,6 +52,11 @@ class TestOutputHelper
 	boost::filesystem::path const& testFile() { return m_currentTestFileName; }
 	void printTestExecStats();
 
+    // Checkup that all test folders are active during the test run
+    void checkUnfinishedTestFolders();
```
gumb0: This could be `private` I think

test/tools/jsontests/StateTests.cpp
```diff
@@ -142,17 +142,30 @@ class GeneralTestFixture
...
-			return;
-		}
-		suite.runAllTestsInFolder(casename);
-	}
+        test::StateTestSuite suite;
+        string casename = boost::unit_test::framework::current_test_case().p_name;
+        boost::filesystem::path suiteFillerPath = suite.getFullPathFiller(casename).parent_path();
+
+        // Check specific test cases
+        static vector<string> timeConsumingTestSuites;
```
gumb0: this should work without `push_back`s
```suggestion
        static vector<string> timeConsumingTestSuites{string{"stStaticCall"}, string{"stSStoreTest"}, string{"stQuadraticComplexityTest"}};
```

gumb0: Or maybe without `string{}`, not sure maybe Windows build would complain without

test/tools/libtesteth/TestOutputHelper.cpp
winsvega: if filter is set. then for loop works only once.

test/tools/jsontests/StateTests.cpp
winsvega: It does

gumb0:
Windows build still fails.

test/tools/libtesteth/TestOutputHelper.cpp
gumb0: I know, that's my point, going into the loop when you don't need it is just confusing

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -100,3 +102,50 @@ void TestOutputHelper::printTestExecStats()
...
+    {
+        boost::filesystem::path path = allTestsIt.first;
+        set<string> allFolders;
+        using fsIterator = boost::filesystem::directory_iterator;
+        if (filter.empty())
+        {
+            for (fsIterator it(path); it != fsIterator(); ++it)
+            {
+                if (boost::filesystem::is_directory(*it))
+                    allFolders.emplace(it->path().filename().c_str());
```
gumb0: The windows error is here

gumb0: Try to call `insert`

winsvega:
added a check for the folders without test files.

test/tools/libtesteth/TestOutputHelper.cpp
winsvega: worked

winsvega:
ready

test/tools/jsontests/TransactionTests.cpp
```diff
@@ -219,9 +219,11 @@ class TransactionTestFixture
 public:
     TransactionTestFixture()
     {
-        string const& casename = boost::unit_test::framework::current_test_case().p_name;
         test::TransactionTestSuite suite;
+        string const& casename = boost::unit_test::framework::current_test_case().p_name;
```
gumb0: I'm not sure if saving to reference is safe here, maybe save to `string const`

chfast:
> Note that the tests `stStaticCall` and `stSStore` won't be run for default testeth run after this, I don't know if that's ok.

Why is that?

winsvega:
let me move time consuming tests in a separate PR

PR #5563, opened 2019-04-11
add timeout to test_mineBlocks(_number, _timeout)

winsvega:
Some test blocks might take a lot of time to mine (heavy transactions) 
Thus it would be good to have the mining timeout option to be configurable in the client. 

Changing this method will affect 
* Solidity rpc tests
* Requrie retesteth test rpc change

### DIFF
--- a/libethereum/ClientTest.h
+++ b/libethereum/ClientTest.h
@@ -50,9 +50,10 @@ public:
     void rewindToBlock(unsigned _number);
     h256 importRawBlock(std::string const& _blockRLP);
     bool completeSync();
+    void setTestMiningBlockTimeout(unsigned _timeout) { m_singleBlockMaxMiningTimeInSeconds = _timeout; }
 
 protected:
-    unsigned const m_singleBlockMaxMiningTimeInSeconds = 5;
+    unsigned m_singleBlockMaxMiningTimeInSeconds = 5;
 };
 
 ClientTest& asClientTest(Interface& _c);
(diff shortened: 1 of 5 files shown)


libethereum/ClientTest.h
```diff
@@ -50,9 +50,10 @@ class ClientTest: public Client
     void rewindToBlock(unsigned _number);
     h256 importRawBlock(std::string const& _blockRLP);
     bool completeSync();
+    void setTestMiningBlockTimeout(unsigned _timeout) { m_singleBlockMaxMiningTimeInSeconds = _timeout; }
 
 protected:
-    unsigned const m_singleBlockMaxMiningTimeInSeconds = 5;
+    unsigned m_singleBlockMaxMiningTimeInSeconds = 5;
```
gumb0: I don't see how you're going to use it in mining. Forgot that part?

gumb0:
Do you intend to set it to larger value than default 5 seconds per block? Maybe we can just increase the default?

winsvega:
So increasing the default woukd be time consuming. Some transactions take long time to process and say 3 minute of execution time is ok. But other if take more then 5 seconds indicate that smth is wrong with the mining and its ok to mark the test as failed and go to the next test

gumb0:
But something wrong with the mining is not a typical scenario, it shouldn't happen at all (do you ever see it?). Not sure if you need to support tests running smoothly in case the client's behavior is clearly buggy.

@chfast Any thoughts about this PR?

winsvega:
Hm. The other solution would be to abort the test on retesteth side and close socket. Stopping execution of any other tests for this client.

gumb0:
> Hm. The other solution would be to abort the test on retesteth side and close socket. Stopping execution of any other tests for this client.

I guess it should also be possible to reopen the socket in case you want to continue the tests (but also not sure if it's worth the development effort)

PR #5566, opened 2019-04-13
build(scripts/docker/aleth.dockerfile): switchable build type

yxliang01:
Now, we can set build type (`Release` or `Debug`) via Docker build-time argument `BuildType`.

Related to https://github.com/ethereum/aleth/issues/5565 (a potential solution to it, might not be the best).

### DIFF
--- a/scripts/docker/aleth.dockerfile
+++ b/scripts/docker/aleth.dockerfile
@@ -7,6 +7,7 @@
 # Build stage
 
 FROM alpine:latest AS builder
+ARG BuildType=Release
 RUN apk add --no-cache \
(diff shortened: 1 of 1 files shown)


chfast:
@yxliang01 Thanks for the PR. This is good change but will be obsoleted by #5570, which just drops VMTRACE build option entirely.

yxliang01:
@chfast No, I think you are mistaken. This is not entirely related to `VMTRACE`. This is to switch the CMAKE build type. Being able to build the source in DEBUG mode allows the debug information being generated.

PR #5567, opened 2019-04-13
aleth.dockerfile: run testeth as non-root user

yxliang01:
For `testeth`, there is no need to run it as root. So, now making it run as non-root user.

### DIFF
--- a/scripts/docker/aleth.dockerfile
+++ b/scripts/docker/aleth.dockerfile
@@ -23,7 +23,9 @@ RUN make -j $(nproc) && make install
 #     docker build --target testeth -f scripts/docker/aleth.dockerfile .
 
 FROM alpine:latest AS testeth
+RUN adduser -D testeth
 RUN apk add --no-cache libstdc++
+USER testeth
 COPY --from=builder /build/test/testeth /usr/bin/
 ENTRYPOINT ["/usr/bin/testeth"]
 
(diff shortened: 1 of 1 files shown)


yxliang01:
@chfast "Add use `aleth`" I don't really get this sentence. Did you mean in the `builder` image, you build as the normal user `alteth` and both `aleth` image and `testeth` image are using that user created in `builder` image as  the runtime effective user?

chfast:
I meant "add user `aleth`". In general I'd like to apply the same change to both aleth and testeth images. I guess I was wrong that this can be done in the base image, because we only copy files from it. So, can you also add "aleth" user to the "aleth" image the same as you did for "testeth"?

yxliang01:
@chfast Right. Theoretically, `aleth` doesn't need `root` privilege and can run under normal user. However, while I can easily quickly do so, as I'm not using `aleth`, I might not be able to test, that's why it wasn't done at the beginning. Maybe I can add it, and you help me to test it to see if working properly?

yxliang01:
@chfast Might processing this PR? It has been hanging here for 4 months.

yxliang01:
Pinging again (2 more months have passed) @chfast

gumb0:
@yxliang01 Seems like you were requested some changes, could you submit them, then we'll check that they work?

chfast:
So I decided it will be simpler to build it as root (2 last commits are not necessary). This is the modified version: https://github.com/ethereum/aleth/pull/5817 (I was not able to push changes to this PR).

Let's either pull changes from my PR to this one or merge mine.

chfast:
Closed in favor of https://github.com/ethereum/aleth/pull/5817.

PR #5570, opened 2019-04-15
Remove VMTRACE option

### LINKED ISSUE #5565: testeth reports not built with -DVMTRACE=1 even when it was
This happens when I run testeth with a docker setup. But, I think it has nothing to do with the docker setup.

chfast:
Fixes #5565.

This removes CMake's VMTRACE options by making --vmtrace always available to testeth. Win-Win.

### DIFF
--- a/libdevcore/Log.cpp
+++ b/libdevcore/Log.cpp
@@ -89,6 +89,7 @@ void formatter(boost::log::record_view const& _rec, boost::log::formatting_ostre
         _strm << " " EthNavy << _rec[suffix] << EthReset;
 }
 
+std::atomic<bool> g_vmTraceEnabled{false};
 }  // namespace
 
 std::string getThreadName()
@@ -146,6 +147,13 @@ void setupLogging(LoggingOptions const& _options)
         boost::log::make_exception_handler<std::exception>([](std::exception const& _ex) {
         std::cerr << "Exception from the logging library: " << _ex.what() << '\n';
     }));
+
+    g_vmTraceEnabled = _options.vmTrace;
+}
+
+bool isVmTraceEnabled()
+{
+    return g_vmTraceEnabled;
 }
 
 }  // namespace dev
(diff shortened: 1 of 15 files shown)


winsvega:
Wait the option is still used in debug mode. I use it to trace step by step execution.

chfast:
> Wait the option is still used in debug mode. I use it to trace step by step execution.

Ok then. We will add `--vmtrace` option also for aleth. 

Anyway, can we merge it now and include in 1.6.0 release to make testeth there more useful? If too big or cannot land without the mentioned flag for aleth, I will do it for 1.7.

gumb0:
I would prefer not to break vmtrace in aleth in 1.6.0

chfast:
@winsvega Can you elaborate what "works"?
I though you need also tracing logs from running aleth, e.g. `aleth --dev`. Is that the case?

winsvega:
From retesteth. Evm debug from the client I propose for the future when retesteth gonna be adapted. 

I think @cdetrio Casey was using the trace from clients to check consensus also

winsvega:
*testeth works with this change

gumb0:
vmtrace in aleth is useful to debug issues with soltest, and also in case we have a consensus issue during sync.

chfast:
I don't have time to continue working on this currently, but it would be huge UX improvement to remove compile time flag.

gumb0:
I will try to finish this.

Not sure if we need additional `--vmtrace` flag for aleth, I think without it `-v 4 --log-channels vmtrace` will also work just fine.

gumb0:
I've added `--log-vmtrace` flag to aleth, it now controls the global flag for vmtrace, which is in all respects similar to the old compile-time flag. New flag lives in Log.cpp.

This should be ready, please review @chfast @winsvega

libdevcore/Log.cpp
```diff
@@ -89,6 +89,7 @@ void formatter(boost::log::record_view const& _rec, boost::log::formatting_ostre
         _strm << " " EthNavy << _rec[suffix] << EthReset;
 }
 
+std::atomic<bool> g_vmTraceEnabled{false};
```
chfast: There is also `std::atomic_flag` which is superior to `std::atomic<bool>` unless you need some integer-like features of the `bool` type.

chfast: Can you make it `static` for static linkage?

chfast:
Rebased. I will merge it after CI builds.

libdevcore/Log.cpp
gumb0: `atomic_flag` is lower level and superior in case you implement something low-level like new kind of concurrency primitive.

For program logic `atomic<bool>` is superior because it has more simple interface, including conversion to `bool`.

gumb0: It has internal linkage because it's in the unnamed namespace.

PR #5571, opened 2019-04-15
Discovery ENR extension - server part

gumb0:
Support ENRRequest / ENRResponse discovery packets.

This doesn't add any change in the Discovery algorithm itself yet, just adds the ability for the node to return its current ENR to peers.

I've decided not to add CLI option to enable this and have it always enabled, as it should be fully backwards-compatible, nodes not supporting ENR extension shouldn't notice any difference.

Also this adds a function to libdevcrypto for converting compressed public key to uncompressed (used only in test for now)

Also renames `DiscoveryDatagram::ts` => `DiscoveryDatagram::expiration`

cc @fjl

### DIFF
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -89,6 +89,26 @@ Public dev::toPublic(Secret const& _secret)
     return Public{&serializedPubkey[1], Public::ConstructFromPointer};
 }
 
+Public dev::toPublic(PublicCompressed const& _publicCompressed)
+{
+    auto* ctx = getCtx();
+
+    secp256k1_pubkey rawPubkey;
+    if (!secp256k1_ec_pubkey_parse(
+            ctx, &rawPubkey, _publicCompressed.data(), PublicCompressed::size))
+        return {};
+
+    std::array<byte, 65> serializedPubkey;
+    size_t serializedPubkeySize = serializedPubkey.size();
+    secp256k1_ec_pubkey_serialize(
+        ctx, serializedPubkey.data(), &serializedPubkeySize, &rawPubkey, SECP256K1_EC_UNCOMPRESSED);
+    assert(serializedPubkeySize == serializedPubkey.size());
+    // Expect single byte header of value 0x04 -- uncompressed public key.
+    assert(serializedPubkey[0] == 0x04);
+    // Create the Public skipping the header.
+    return Public{&serializedPubkey[1], Public::ConstructFromPointer};
+}
+
 PublicCompressed dev::toPublicCompressed(Secret const& _secret)
 {
     PublicCompressed serializedPubkey;
--- a/libdevcrypto/Common.cpp
+++ b/libdevcrypto/Common.cpp
@@ -99,7 +99,7 @@ Public dev::toPublic(PublicCompressed const& _publicCompressed)
         return {};
 
     std::array<byte, 65> serializedPubkey;
-    size_t serializedPubkeySize = serializedPubkey.size();
+    auto serializedPubkeySize = serializedPubkey.size();
     secp256k1_ec_pubkey_serialize(
         ctx, serializedPubkey.data(), &serializedPubkeySize, &rawPubkey, SECP256K1_EC_UNCOMPRESSED);
     assert(serializedPubkeySize == serializedPubkey.size());
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -687,7 +687,7 @@ std::shared_ptr<NodeEntry> NodeTable::handleENRRequest(
 
     auto const& in = dynamic_cast<ENRRequest const&>(_packet);
 
-    ENRResponse response(_from, m_hostENR);
+    ENRResponse response{_from, m_hostENR};
     LOG(m_logger) << response.typeName() << " to " << in.sourceid << "@" << _from;
     response.timestamp = nextRequestExpirationTime();
     response.echo = in.echo;
@@ -704,7 +704,7 @@ std::shared_ptr<NodeEntry> NodeTable::handleENRResponse(
     if (!sourceNodeEntry)
     {
         LOG(m_logger) << "Source node (" << _packet.sourceid << "@" << _from
-                      << ") not found in node table. Ignoring Neighbours packet.";
+                      << ") not found in node table. Ignoring ENRResponse packet.";
         return {};
     }
     if (sourceNodeEntry->endpoint() != _from)
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -221,7 +221,7 @@ void NodeTable::doDiscoveryRound(
             }
 
             FindNode p(nodeEntry->endpoint(), _node);
-            p.timestamp = nextRequestExpirationTime();
+            p.expiration = nextRequestExpirationTime();
             p.sign(m_secret);
             m_sentFindNodes.emplace_back(nodeEntry->id(), chrono::steady_clock::now());
             LOG(m_logger) << p.typeName() << " to " << nodeEntry->node << " (target: " << _node
@@ -303,7 +303,7 @@ void NodeTable::ping(Node const& _node, shared_ptr<NodeEntry> _replacementNodeEn
     }
 
     PingNode p{m_hostNodeEndpoint, _node.endpoint};
-    p.timestamp = nextRequestExpirationTime();
+    p.expiration = nextRequestExpirationTime();
     p.seq = m_hostENR.sequenceNumber();
     auto const pingHash = p.sign(m_secret);
     LOG(m_logger) << p.typeName() << " to " << _node;
@@ -618,7 +618,7 @@ std::shared_ptr<NodeEntry> NodeTable::handleFindNode(
     for (unsigned offset = 0; offset < nearest.size(); offset += nlimit)
     {
         Neighbours out(_from, nearest, offset, nlimit);
-        out.timestamp = nextRequestExpirationTime();
+        out.expiration = nextRequestExpirationTime();
         LOG(m_logger) << out.typeName() << " to " << in.sourceid << "@" << _from;
         out.sign(m_secret);
         if (out.data.size() > 1280)
@@ -641,7 +641,7 @@ std::shared_ptr<NodeEntry> NodeTable::handlePingNode(
     // Send PONG response.
     Pong p(sourceEndpoint);
     LOG(m_logger) << p.typeName() << " to " << in.sourceid << "@" << _from;
-    p.timestamp = nextRequestExpirationTime();
+    p.expiration = nextRequestExpirationTime();
     p.echo = in.echo;
     p.seq = m_hostENR.sequenceNumber();
     p.sign(m_secret);
@@ -689,7 +689,7 @@ std::shared_ptr<NodeEntry> NodeTable::handleENRRequest(
 
     ENRResponse response{_from, m_hostENR};
     LOG(m_logger) << response.typeName() << " to " << in.sourceid << "@" << _from;
-    response.timestamp = nextRequestExpirationTime();
+    response.expiration = nextRequestExpirationTime();
     response.echo = in.echo;
     response.sign(m_secret);
     m_socket->send(response);
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -32,9 +32,10 @@ inline bool operator==(weak_ptr<NodeEntry> const& _weak, shared_ptr<NodeEntry> c
 }
 
 NodeTable::NodeTable(ba::io_service& _io, KeyPair const& _alias, NodeIPEndpoint const& _endpoint,
-    bool _enabled, bool _allowLocalDiscovery)
+    ENR const& _enr, bool _enabled, bool _allowLocalDiscovery)
   : m_hostNodeID{_alias.pub()},
     m_hostNodeEndpoint{_endpoint},
+    m_hostENR{_enr},
     m_secret{_alias.secret()},
     m_socket{make_shared<NodeSocket>(
         _io, static_cast<UDPSocketEvents&>(*this), (bi::udp::endpoint)m_hostNodeEndpoint)},
@@ -220,7 +221,7 @@ void NodeTable::doDiscoveryRound(
             }
 
             FindNode p(nodeEntry->endpoint(), _node);
-            p.ts = nextRequestExpirationTime();
+            p.timestamp = nextRequestExpirationTime();
             p.sign(m_secret);
             m_sentFindNodes.emplace_back(nodeEntry->id(), chrono::steady_clock::now());
             LOG(m_logger) << p.typeName() << " to " << nodeEntry->node << " (target: " << _node
@@ -302,7 +303,8 @@ void NodeTable::ping(Node const& _node, shared_ptr<NodeEntry> _replacementNodeEn
     }
 
     PingNode p{m_hostNodeEndpoint, _node.endpoint};
-    p.ts = nextRequestExpirationTime();
+    p.timestamp = nextRequestExpirationTime();
+    p.seq = m_hostENR.sequenceNumber();
     auto const pingHash = p.sign(m_secret);
     LOG(m_logger) << p.typeName() << " to " << _node;
     m_socket->send(p);
@@ -453,6 +455,14 @@ void NodeTable::onPacketReceived(
 
         case PingNode::type:
             sourceNodeEntry = handlePingNode(_from, *packet);
+            break;
+
+        case ENRRequest::type:
+            sourceNodeEntry = handleENRRequest(_from, *packet);
+            break;
+
+        case ENRResponse::type:
+            sourceNodeEntry = handleENRResponse(_from, *packet);
             break;
         }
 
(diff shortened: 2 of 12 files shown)


libdevcrypto/Common.cpp
```diff
@@ -89,6 +89,26 @@ Public dev::toPublic(Secret const& _secret)
     return Public{&serializedPubkey[1], Public::ConstructFromPointer};
 }
 
+Public dev::toPublic(PublicCompressed const& _publicCompressed)
+{
+    auto* ctx = getCtx();
+
+    secp256k1_pubkey rawPubkey;
+    if (!secp256k1_ec_pubkey_parse(
```
gumb0: geth's similar code for reference: https://github.com/ethereum/go-ethereum/blob/c4109d790ffd26d67feb7745d4af8a8bc5090bd9/crypto/secp256k1/ext.h#L76-L90

libp2p/NodeTable.cpp
```diff
@@ -674,6 +685,69 @@ std::shared_ptr<NodeEntry> NodeTable::handlePingNode(
...
+    if (!sourceNodeEntry)
+    {
+        LOG(m_logger) << "Source node (" << _packet.sourceid << "@" << _from
+                      << ") not found in node table. Ignoring ENRRequest request.";
+        return {};
+    }
+    if (sourceNodeEntry->endpoint() != _from)
+    {
+        LOG(m_logger) << "ENRRequest packet from unexpected endpoint " << _from << " instead of "
+                      << sourceNodeEntry->endpoint();
```
halfalicious: Strange formatting, why is this expected endpoint on a separate line?

libp2p/NodeTable.cpp
```diff
@@ -674,6 +685,69 @@ std::shared_ptr<NodeEntry> NodeTable::handlePingNode(
...
+    }
+    if (!sourceNodeEntry->hasValidEndpointProof())
+    {
+        LOG(m_logger) << "Unexpected ENRRequest packet! Endpoint proof has expired.";
+        return {};
+    }
+
+    auto const& in = dynamic_cast<ENRRequest const&>(_packet);
+
+    ENRResponse response(_from, m_hostENR);
```
halfalicious: {} initializers?

libp2p/NodeTable.cpp
gumb0: That's clang-format wraps around max line width

halfalicious:
Not a part of your changes, but all discovery packet types can be constexpr (e.g. PingNode::type, PongNode::type)

halfalicious:
Should we also add a test for receiving an ENRRequest from a node for which we haven't completed the endpoint proof?

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+{
+    // NodeTable receiving ENRRequest
+    TestNodeTableHost nodeTableHost(0);
+    nodeTableHost.start();
+    auto& nodeTable = nodeTableHost.nodeTable;
+
+    // socket sending ENRRequest
+    TestUDPSocketHost nodeSocketHost;
+    nodeSocketHost.start();
+    uint16_t nodePort = nodeSocketHost.port;
```
halfalicious: (Nit) can be auto const

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+
+    // Exchange Ping/Pongs before sending ENRRequest
+
+    // add a node to node table, initiating PING
+    auto nodeEndpoint = NodeIPEndpoint{bi::address::from_string(c_localhostIp), nodePort, nodePort};
```
halfalicious: (Nit) Can be const

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    auto nodeKeyPair = KeyPair::create();
```
halfalicious: (Nit) can be const

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    auto nodePubKey = nodeKeyPair.pub();
```
halfalicious: (Nit) can be const

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    nodeTable->addNode(Node{nodePubKey, nodeEndpoint});
+
+    // handle received PING
+    auto pingDataReceived = nodeSocketHost.packetsReceived.pop();
+    auto pingDatagram =
+        DiscoveryDatagram::interpretUDP(bi::udp::endpoint{}, dev::ref(pingDataReceived));
+    BOOST_REQUIRE_EQUAL(pingDatagram->typeName(), "Ping");
+    auto ping = dynamic_cast<PingNode const&>(*pingDatagram);
```
halfalicious: (Nit) can be const

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    // Exchange Ping/Pongs before sending ENRRequest
+
+    // add a node to node table, initiating PING
+    auto nodeEndpoint = NodeIPEndpoint{bi::address::from_string(c_localhostIp), nodePort, nodePort};
+    auto nodeKeyPair = KeyPair::create();
+    auto nodePubKey = nodeKeyPair.pub();
+    nodeTable->addNode(Node{nodePubKey, nodeEndpoint});
+
+    // handle received PING
+    auto pingDataReceived = nodeSocketHost.packetsReceived.pop();
```
halfalicious: (Nit) can be const

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    auto pingDatagram =
```
halfalicious: (Nit) can be const

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    auto ping = dynamic_cast<PingNode const&>(*pingDatagram);
+
+    // send PONG
+    Pong pong(nodeTable->m_hostNodeEndpoint);
+    pong.echo = ping.echo;
+    pong.sign(nodeKeyPair.secret());
+    nodeSocketHost.socket->send(pong);
+
+    // wait for PONG to be received and handled
+    auto pongDataReceived = nodeTable->packetsReceived.pop(chrono::seconds(5));
```
halfalicious: (Nit) can be const

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    auto pongDatagram =
```
halfalicious: (Nit) can be const

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+        DiscoveryDatagram::interpretUDP(bi::udp::endpoint{}, dev::ref(pongDataReceived));
+    BOOST_REQUIRE_EQUAL(pongDatagram->typeName(), "Pong");
+
+    // send ENRRequest
+    ENRRequest enrRequest(nodeTable->m_hostNodeEndpoint);
+    enrRequest.sign(nodeKeyPair.secret());
+    nodeSocketHost.socket->send(enrRequest);
+
+    // wait for ENRRequest to be received and handled
+    nodeTable->packetsReceived.pop(chrono::seconds(5));
```
halfalicious: (Nit) Should we verify the packet type here to? Can be helpful for debugging if we get test failures due to different networking tests executing concurrently

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+
+    auto enrResponsePacket = nodeSocketHost.packetsReceived.pop(chrono::seconds(5));
```
halfalicious: (Nit) Can be const

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    auto datagram =
```
halfalicious: (Nit) Can be const

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+        DiscoveryDatagram::interpretUDP(bi::udp::endpoint{}, dev::ref(enrResponsePacket));
+    BOOST_REQUIRE_EQUAL(datagram->typeName(), "ENRResponse");
+
+    auto enrResponse = dynamic_cast<ENRResponse const&>(*datagram);
```
halfalicious: (Nit) Can be const

test/unittests/libp2p/net.cpp
```diff
@@ -1147,6 +1148,71 @@ BOOST_AUTO_TEST_CASE(addNodePingsNodeOnlyOnce)
...
+    auto const& keyValuePairs = enrResponse.enr->keyValuePairs();
+    BOOST_REQUIRE_EQUAL(RLP{keyValuePairs.at("id")}.toString(), "v4");
```
halfalicious: (Nit) Instead of hard-coding v4 here should we maybe pull the protocol version from the node table?

Should we also do the same for the key names?

libp2p/NodeTable.cpp
```diff
@@ -674,6 +685,69 @@ std::shared_ptr<NodeEntry> NodeTable::handlePingNode(
...
+    LOG(m_logger) << response.typeName() << " to " << in.sourceid << "@" << _from;
+    response.timestamp = nextRequestExpirationTime();
+    response.echo = in.echo;
+    response.sign(m_secret);
+    m_socket->send(response);
+
+    return sourceNodeEntry;
+}
+
+std::shared_ptr<NodeEntry> NodeTable::handleENRResponse(
```
halfalicious: We should verify the hash in the response against the request (or are you planning on doing that in another change)?

libp2p/NodeTable.h
```diff
@@ -404,18 +410,21 @@ struct DiscoveryDatagram: public RLPXDatagramFace
     NodeID sourceid; // sender public key (from signature)
     h256 echo;       // hash of encoded packet, for reply tracking
 
-    // All discovery packets carry a timestamp, which must be greater
+    // Most discovery packets carry a timestamp, which must be greater
     // than the current local time. This prevents replay attacks.
-    uint32_t ts = 0;
-    bool isExpired() const { return secondsSinceEpoch() > ts; }
+    boost::optional<uint32_t> timestamp;
```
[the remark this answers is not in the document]
gumb0: Because now not all of the packets contain timestamp - `ENRResponse` doesn't.
It won't open us up to replay attacks, because we still parse it and check it for all packets that should have it.

libp2p/NodeTable.h
```diff
@@ -434,22 +443,27 @@ struct PingNode: DiscoveryDatagram
...
     }
     void interpretRLP(bytesConstRef _bytes) override
     {
         RLP r(_bytes, RLP::AllowNonCanon|RLP::ThrowOnFail);
         version = r[0].toInt<unsigned>();
         source.interpretRLP(r[1]);
         destination.interpretRLP(r[2]);
-        ts = r[3].toInt<uint32_t>();
+        timestamp = r[3].toInt<uint32_t>();
+        if (r.itemCount() > 4 && r[4].isInt())
```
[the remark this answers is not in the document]
gumb0: This check for `isInt()` is actually kind of a hack, I've added it to work around failures in eip-8 tests here https://github.com/ethereum/aleth/blob/e80e19770e2ab60d2ba8fbace27a7f8b1d006f02/test/unittests/libp2p/eip-8.cpp#L18

EIP-8 prescribes to ignore any additional data after the last field of the packet, and the test vectors there check it by parsing packets with additional rlp-list in the end (so not an int).
Those test vectors are taken directly from EIP-8 spec, so I didn't want to modify them. I've asked whether we should change these test vectors in the EIP https://github.com/ethereum/devp2p/issues/44#issuecomment-483608954

In some sense the packet with 5 items, where the last one is not an int, is also a valid packet for pre-ENR-extension discovery protocol.

libp2p/NodeTable.h
```diff
@@ -556,18 +576,75 @@ struct Neighbours: DiscoveryDatagram
...
+        _s.appendList(1);
+        _s << *timestamp;
+    }
+    void interpretRLP(bytesConstRef _bytes) override
+    {
+        RLP r(_bytes, RLP::AllowNonCanon | RLP::ThrowOnFail);
+        timestamp = r[0].toInt<uint32_t>();
+    }
+
+    std::string typeName() const override { return "ENRRequest"; }
```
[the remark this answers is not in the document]
gumb0: Maybe there's no need, though it's defined in the base interface in `UDP.h`, so I'd prefer not to expand the scope of this PR too much.

Also I actually prefer `std::string`s for the sake of higher-level code and avoiding raw pointers (unless it causes some performance bottleneck)

libp2p/NodeTable.h
```diff
@@ -556,18 +576,75 @@ struct Neighbours: DiscoveryDatagram
...
+    {}
+    // Constructor for incoming packets
+    ENRResponse(bi::udp::endpoint const& _from, NodeID const& _fromid, h256 const& _echo)
+      : DiscoveryDatagram{_from, _fromid, _echo}
+    {}
+
+    static uint8_t const type = 6;
+    uint8_t packetType() const override { return type; }
+
+    boost::optional<ENR> enr;
```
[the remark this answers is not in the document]
gumb0: This is only because `ENR` doesn't have a default constructor (and I'd prefer to leave it like this - empty ENR wouldn't be valid), and it is created not in constructor, but in `interpretRLP` method.

Actually I think I'll change it to `unique_ptr`, it would better express the intent.

libp2p/NodeTable.cpp
```diff
@@ -674,6 +685,69 @@ std::shared_ptr<NodeEntry> NodeTable::handlePingNode(
...
+    std::shared_ptr<NodeEntry> sourceNodeEntry = nodeEntry(_packet.sourceid);
+    if (!sourceNodeEntry)
+    {
+        LOG(m_logger) << "Source node (" << _packet.sourceid << "@" << _from
+                      << ") not found in node table. Ignoring ENRRequest request.";
+        return {};
+    }
+    if (sourceNodeEntry->endpoint() != _from)
+    {
+        LOG(m_logger) << "ENRRequest packet from unexpected endpoint " << _from << " instead of "
```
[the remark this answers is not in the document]
gumb0: Well we don't for the same message of `FindNode` and `Neighbours`, and we log it just before in `onPacketReceived`

test/unittests/libp2p/net.cpp
gumb0: I'll make a helper function out of this wait + check, because it's used often.

test/unittests/libp2p/net.cpp
gumb0: Do you mean using the constants from the implementation instead of string literals here?

libp2p/NodeTable.cpp
gumb0: Yes, I planned to add the validation when implementing doing something useful with these responses (the last "ENR Client" part in https://github.com/ethereum/aleth/issues/5551)
But now I also think there's not much useful we can actually do with them in v4 discovery, so maybe I'll leave it just ignoring ENRResponses

gumb0:
> Should we also add tests for unexpected ENRRequests (e.g. we haven't completed the endpoint proof for the node, or the node isn't in our node table)?

Yeah would be good, but I'm a bit lazy for that :stuck_out_tongue_winking_eye:

libp2p/NodeTable.cpp
halfalicious: If we've enabled the functionality I think we should validate the responses so that way there's 1 less thing to do when we actually do something useful with them.

libp2p/NodeTable.h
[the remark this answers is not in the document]
halfalicious: Ah you're right, plus there's no danger of replay attacks with the response since ENRs have the sequence number and the hash of the request.

libp2p/NodeTable.h
```diff
@@ -394,7 +401,7 @@ struct DiscoveryDatagram: public RLPXDatagramFace
 
     /// Constructor used for sending.
     DiscoveryDatagram(bi::udp::endpoint const& _to)
-      : RLPXDatagramFace(_to), ts(futureFromEpoch(c_timeToLiveS))
+      : RLPXDatagramFace(_to), timestamp(futureFromEpoch(c_timeToLiveS))
```
halfalicious: Since timestamp is the packet expiration time, can we rename it to something like "expiration"?

halfalicious:
Also _rlp can be const ref here: https://github.com/ethereum/aleth/blob/332c02365245883ad19fe7fab43492c5ae07be26/libp2p/ENR.h#L32

and here: https://github.com/ethereum/aleth/blob/332c02365245883ad19fe7fab43492c5ae07be26/libp2p/ENR.h#L57

PR #5576, opened 2019-04-24
testeth: add stTimeConsuming

winsvega:
this tests take more then 3 minutes of test execution time. 
also PR updates the test submodule head

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [1.7.0] - Unreleased
 
+- Moved: [#5576](https://github.com/ethereum/aleth/pull/5576) Moved sstore_combinations tests and static_Call50000_sha256 to stTimeConsuming test suite. (require --all flag to run)
 - Added: [#5537](https://github.com/ethereum/aleth/pull/5537) Creating Ethereum Node Record (ENR) at program start.
 - Added: [#5557](https://github.com/ethereum/aleth/pull/5557) Improved debug logging of full sync.
 - Added: [#5564](https://github.com/ethereum/aleth/pull/5564) Improved help output of Aleth by adding list of channels.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,13 @@
 
 ## [1.7.0] - Unreleased
 
-- Moved: [#5576](https://github.com/ethereum/aleth/pull/5576) Moved sstore_combinations tests and static_Call50000_sha256 to stTimeConsuming test suite. (require --all flag to run)
 - Added: [#5537](https://github.com/ethereum/aleth/pull/5537) Creating Ethereum Node Record (ENR) at program start.
 - Added: [#5557](https://github.com/ethereum/aleth/pull/5557) Improved debug logging of full sync.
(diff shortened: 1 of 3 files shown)


test/tools/jsontests/StateTests.cpp
```diff
@@ -147,7 +147,8 @@ class GeneralTestFixture
         boost::filesystem::path suiteFillerPath = suite.getFullPathFiller(casename).parent_path();
 
         // Check specific test cases
-        static vector<string> timeConsumingTestSuites{string{"stQuadraticComplexityTest"}};
+        static vector<string> timeConsumingTestSuites{
```
gumb0: can be const

gumb0:
cc @chfast

chfast:
I don't understand the direction this is going. Is it safe to skip these tests on regular basis? I'd rather check why they take so much time, maybe we can reduce the number of them, etc. If you disable them, 99% of users (including us) will not run them. 

If you moved e.g. stSstore tests, does it mean we don't test (unless explicitly enabled) one of the features at all?

winsvega:
Only those tests I moved which take a lot of time to execute.

chfast:
> Only those tests I moved which take a lot of time to execute.

You mean test suites or test cases?

winsvega:
Test cases. One specific test is 50000 sha256 execution. 
And 5000 of sstore combinations tests

chfast:
Ok, that's fine then.

> One specific test is 50000 sha256 execution.

Are you sure such tests are useful? I believe you get the same effect when running sha256 10x instead of 50000.

winsvega:
> Are you sure such tests are useful? I believe you get the same effect when running sha256 10x instead of 50000.

don't remember why exactly Christoph added this test case. but I see its potential to crush some clients or make it timeout due to not optimized coding.

winsvega:
net tests failing randomly

CHANGELOG.md
```diff
@@ -6,8 +6,9 @@
 - Added: [#5557](https://github.com/ethereum/aleth/pull/5557) Improved debug logging of full sync.
 - Added: [#5564](https://github.com/ethereum/aleth/pull/5564) Improved help output of Aleth by adding list of channels.
 - Changed: [#5559](https://github.com/ethereum/aleth/pull/5559) Update peer validation error messages.
-- Fixed: [#5562](https://github.com/ethereum/aleth/pull/5562) Don't send header request messages to peers that haven't sent us Status yet.
 - Changed: [#5568](https://github.com/ethereum/aleth/pull/5568) Improve rlpx handshake log messages and create new rlpx log channel.
+- Changed: [#5576](https://github.com/ethereum/aleth/pull/5576) Moved sstore_combinations and static_Call50000_sha256 tests to stTimeConsuming test suite. (testeth runs them only with `--all` flag)
```
gumb0: (see https://keepachangelog.com/en/1.0.0/#how for changelog format)

gumb0:
> net tests failing randomly
We know, see https://github.com/ethereum/aleth/issues/5544 and https://github.com/ethereum/aleth/pull/5581

Issue #5544, opened 2019-04-03
Fix unreliable network/p2p tests


halfalicious:
I've seen the following tests fail recently in PRs which didn't touch network or p2p code (e.g. #5541)
* network/net/evictionWithOldNodeAnswering
* network/net/packetsWithChangedEndpointSuite/neighbours 
* libp2p/p2pPeer/requirePeer

Logs:

**evictionWithOldNodeAnswering**
```
495/563 Test #501: network/net/evictionWithOldNodeAnswering ..............................................................................***Failed    0.41 sec
Running tests using path: "/home/builder/project/test/jsontests"
Running 1 test case...
Test Case "evictionWithOldNodeAnswering": 
/home/builder/project/test/unittests/libp2p/net.cpp(854): fatal error: in "network/net/evictionWithOldNodeAnswering": critical check populated has failed
*** 1 failure is detected (5 failures are expected) in the test module "Master Test Suite"
```

**packetsWithChangedEndpointSuite/neighbours**
```
515/563 Test #510: network/net/packetsWithChangedEndpointSuite/neighbours ................................................................***Failed   10.37 sec
Running tests using path: "/home/builder/project/test/jsontests"
Running 1 test case...
Test Case "neighbours": 
/home/builder/project/libdevcore/concurrent_queue.h(47): fatal error: in "_T dev::concurrent_queue<_T, _QueueT>::pop(const milliseconds&) [with _T = std::vector<unsigned char>; _QueueT = std::queue<std::vector<unsigned char>, std::deque<std::vector<unsigned char>, std::allocator<std::vector<unsigned char> > > >; std::chrono::milliseconds = std::chrono::duration<long int, std::ratio<1l, 1000l> >]": boost::exception_detail::clone_impl<dev::WaitTimeout>: /home/builder/project/libdevcore/concurrent_queue.h(47): Throw in function _T dev::concurrent_queue<_T, _QueueT>::pop(const milliseconds&) [with _T = std::vector<unsigned char>; _QueueT = std::queue<std::vector<unsigned char>, std::deque<std::vector<unsigned char>, std::allocator<std::vector<unsigned char> > > >; std::chrono::milliseconds = std::chrono::duration<long int, std::ratio<1l, 1000l> >]
Dynamic exception type: boost::exception_detail::clone_impl<dev::WaitTimeout>
/home/builder/project/test/unittests/libp2p/net.cpp(1253): last checkpoint: "neighbours" test entry
[...]

PR #5580, opened 2019-04-28
Skip dao challenge if top block < hard fork block

### LINKED ISSUE #5579: Only perform DAO hard fork challenge if Aleth's top block is >= DAO hard fork block
Aleth contains a DAO hard fork challenge which it issues to potential peers in order to avoid wasting time syncing to ETC nodes. Additional details on the motivation for this challenge are described here: https://github.com/ethereum/aleth/issues/5540#issuecomment-479441130

The way the challenge works is that Aleth requests the DAO hard fork block from a potential peer and validates the extradata section of the block. If the peer is an ETC node, the hard fork block won't contain the expected extradata data and Aleth will avoid peering with the node. 

The block is requested here: https://github.com/ethereum/aleth/blob/6de1e889ca50dd3776892652af2802f5258193e1/libethereum/BlockChainSync.cpp#L204-L209

The challenge is verified in `BlockChainSync::onPeerBlockHeaders`: https://github.com/ethereum/aleth/blob/6de1e889ca50dd3776892652af2802f5258193e1/libethereum/BlockChainSync.cpp#L444-L453

However, this challenge prevents us from syncing blocks with ETC nodes from before the DAO hard fork block. An optimization would be to only perform the DAO hard fork challenge if our top block is >= the DAO hard fork block (1920000).

halfalicious:
Fix #5579 

As a syncing optimization, skip performing the dao hard fork challenge if our top block < dao hard fork block since we can still successfully sync these blocks from ETC nodes. Note that sync from these peers will fail starting with the dao hard fork block because we validate that the dao balance is refunded in `Block::performIrregularModifications` so we don't run the risk of syncing on an ETC chain after the dao fork block: https://github.com/ethereum/aleth/blob/60fbc2f6eba6b160824953b72f02a079364d02fe/libethereum/Block.cpp#L690-L701

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added: [#5557](https://github.com/ethereum/aleth/pull/5557) Improved debug logging of full sync.
 - Added: [#5564](https://github.com/ethereum/aleth/pull/5564) Improved help output of Aleth by adding list of channels.
 - Added: [#5575](https://github.com/ethereum/aleth/pull/5575) Log active peer count and peer list every 30 seconds.
+- Added: [#5580](https://github.com/ethereum/aleth/pull/5580) Enable syncing from ETC nodes for blocks < dao hard fork block.
 - Changed: [#5559](https://github.com/ethereum/aleth/pull/5559) Update peer validation error messages.
 - Fixed: [#5562](https://github.com/ethereum/aleth/pull/5562) Don't send header request messages to peers that haven't sent us Status yet.
 - Changed: [#5568](https://github.com/ethereum/aleth/pull/5568) Improve rlpx handshake log messages and create new rlpx log channel.
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -213,7 +213,7 @@ bool BlockChainSync::requestDaoForkBlockHeader(NodeID const& _peerID)
 {
     // DAO challenge
     u256 const daoHardfork = host().chain().sealEngine()->chainParams().daoHardforkBlock;
-    if (daoHardfork == 0 || daoHardfork == c_infiniteBlockNumber)
+    if (daoHardfork == 0 || daoHardfork == c_infiniteBlockNumber || host().chain().number() < daoHardfork)
         return false;
 
     m_daoChallengedPeers.insert(_peerID);


halfalicious:
These tests are failing on linux:

```
229 - GeneralStateTests/stSStoreTest (Timeout)
566 - JsonRpcSuite/adminEthVmTrace (Failed)
```
I suspect the timeout might just be something we hit from time-to-time based on how long the tests take to execute.

Here's the VM trace failure:
	
```
563/575 Test #566: JsonRpcSuite/adminEthVmTrace ..........................................................................................***Failed    0.33 sec
Running 1 test case...
Test Case "adminEthVmTrace": 
unknown location(0): fatal error: in "JsonRpcSuite/adminEthVmTrace": jsonrpc::JsonRpcException: Exception -32001 : The response is invalid: null

/home/builder/project/test/unittests/libweb3jsonrpc/jsonrpc.cpp(707): last checkpoint

ERROR: Expected a single test to be passed: adminEthVmTrace
```

I haven't seen this kind of failure before and I'm not familiar with this test, I'll investigate.

halfalicious:
Made the change slightly cleaner by performing the comparison in `BlockChainSync::requestDaoForkBlockheader` and comparing against the `daoHardforkBlock` retrieved from the chain config, rather than performing the comparison in `BlockChainSync::onPeerStatus` and defining a new constexpr for the dao hard fork block.

halfalicious:
> 
> 
> These tests are failing on linux:
> 
> ```
> 229 - GeneralStateTests/stSStoreTest (Timeout)
> 566 - JsonRpcSuite/adminEthVmTrace (Failed)
> ```
> 
> I suspect the timeout might just be something we hit from time-to-time based on how long the tests take to execute.
> 
> Here's the VM trace failure:
> 
> ```
> 563/575 Test #566: JsonRpcSuite/adminEthVmTrace ..........................................................................................***Failed    0.33 sec
> Running tests using path: "/home/builder/project/test/jsontests"
> Running 1 test case...
> Test Case "adminEthVmTrace": 
> unknown location(0): fatal error: in "JsonRpcSuite/adminEthVmTrace": jsonrpc::JsonRpcException: Exception -32001 : The response is invalid: null
> 
> /home/builder/project/test/unittests/libweb3jsonrpc/jsonrpc.cpp(707): last checkpoint
> 
> *** 1 failure is detected (5 failures are expected) in the test module "Master Test Suite"
> ERROR: Expected a single test to be passed: adminEthVmTrace
> ```
> 
> I haven't seen this kind of failure before and I'm not familiar with this test, I'll investigate.

Looks like these were one-off issues since they didn't repro on a subsequent run.

The new failing tests are known to be unreliable due to issues @gumb0 called out in #5544 

```
517 - network/net/pingNotSentAfterPongForKnownNode (Failed)
521 - network/net/packetsWithChangedEndpointSuite/neighbours (Failed)
```

gumb0:
Looks fine, but it would be nice to check how we pass DAO fork block after this (so sync ~2M mainnet blocks)

One potential complication is related to the parallel pipeline of the sync - at the moment when we check the highest block number (taken from the database) here, further blocks might be already downloaded, and around DAO fork block they might have been downloaded from the ETC nodes. 
In theory eventually we should reject all ETC blocks and proceed with syncing ETH chain, but we better check that it's not broken.

halfalicious:
> 
> 
> Looks fine, but it would be nice to check how we pass DAO fork block after this (so sync ~2M mainnet blocks)
> 
> One potential complication is related to the parallel pipeline of the sync - at the moment when we check the highest block number (taken from the database) here, further blocks might be already downloaded, and around DAO fork block they might have been downloaded from the ETC nodes.
> In theory eventually we should reject all ETC blocks and proceed with syncing ETH chain, but we better check that it's not broken.

Restarted my sync on Linux, hope to have hit 2M+ blocks by tomorrow morning (assuming I don't hit the "sync gets stuck" issue again).

halfalicious:
> 
> 
> > Looks fine, but it would be nice to check how we pass DAO fork block after this (so sync ~2M mainnet blocks)
> > One potential complication is related to the parallel pipeline of the sync - at the moment when we check the highest block number (taken from the database) here, further blocks might be already downloaded, and around DAO fork block they might have been downloaded from the ETC nodes.
> > In theory eventually we should reject all ETC blocks and proceed with syncing ETH chain, but we better check that it's not broken.
> 
> Restarted my sync on Linux, hope to have hit 2M+ blocks by tomorrow morning (assuming I don't hit the "sync gets stuck" issue again).

Sync'd my Linux machine to over 2M blocks but then hit a core dump a bit after 2M. Was able to resume syncing on relaunch but I'm going to try restarting syncing from genesis to see if I hit this again and if so debug.

gumb0:
It might be not a new issue, we have some reports about sync crashes like https://github.com/ethereum/aleth/issues/5046 and https://github.com/ethereum/aleth/issues/4584

halfalicious:
@gumb0 I resync'd past the DAO fork block and Aleth didn't crash. I did hit the sync stuck issue again despite Aleth consistently having between 5 and 10 peers and I'm seeing "stack offloading: depth 80" messages since block 1.7M, but I'll merge my changes and investigate those separately since they're probably unrelated. 

Here's the log: 
[dao_sync_logs_stack_offload.txt](https://github.com/ethereum/aleth/files/3138934/dao_sync_logs_stack_offload.txt)

PR #5581, opened 2019-04-29
Fix failing Discovery tests

gumb0:
Fixes `network/net` tests failures mentioned in https://github.com/ethereum/aleth/issues/5544

The reason for some of the network test failires was incorrect implementation of `NodeTable::nearestNodeEntries` method (finding the nodes from the node table closest to some target node)

The old implementation tried to utilize the ordering of the nodes in the node table (they are ordered in buckets by distance to the host node) to efficiently find the nodes closest to target. It first looked for the bucket where the target node would fall, and then assumed the nodes in that bucket and around would be close to the target. However I don't think the idea was completely correct, and the implementation definitely was buggy and difficult to understand.

I ended up rewriting it in another, I think simpler, manner using `std::multiset` ordered by the distance to target and keeping only 16 closest nodes.

Go-ethereum uses the similar apporach of iterating over all nodes and putting them into a data structure which keeps only 16 nodes closest to target. 
https://github.com/ethereum/go-ethereum/blob/4c90efdf57ce87edf0d855c8cec10525875a6ab1/p2p/discover/table.go#L521-L536
https://github.com/ethereum/go-ethereum/blob/4c90efdf57ce87edf0d855c8cec10525875a6ab1/p2p/discover/table.go#L757-L781

Parity uses another, much trickier (and presumably more efficient) approach, but I decided not to complicate it like this. Also it's very different from our original approach.
https://github.com/paritytech/parity-ethereum/blob/b30b54e446981dc57835b74d550b40b36385742a/util/network-devp2p/src/discovery.rs#L422-L471

### DIFF
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -270,7 +270,7 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID const& _targe
         return _node1.first < _node2.first;
     };
 
-    std::set<pair<int, shared_ptr<NodeEntry>>, decltype(distanceToTargetLess)>
+    std::multiset<pair<int, shared_ptr<NodeEntry>>, decltype(distanceToTargetLess)>
         nodesByDistanceToTarget(distanceToTargetLess);
     for (auto const& bucket : m_buckets)
         for (auto const& nodeWeakPtr : bucket.nodes)
@@ -279,7 +279,7 @@ vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID const& _targe
                 nodesByDistanceToTarget.emplace(distance(_target, node->id()), node);
 
                 if (nodesByDistanceToTarget.size() > s_bucketSize)
-                    nodesByDistanceToTarget.erase(nodesByDistanceToTarget.rbegin().base());
+                    nodesByDistanceToTarget.erase(--nodesByDistanceToTarget.end());
             }
 
     vector<shared_ptr<NodeEntry>> ret;
--- a/libp2p/NodeTable.cpp
+++ b/libp2p/NodeTable.cpp
@@ -265,30 +265,26 @@ void NodeTable::doDiscoveryRound(
 
 vector<shared_ptr<NodeEntry>> NodeTable::nearestNodeEntries(NodeID const& _target)
 {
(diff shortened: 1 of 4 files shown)


gumb0:
Also first I tried another approach using `std::partial_sort` https://github.com/ethereum/aleth/commit/85fe2a41815d601957b79431d28714d3a0a8ab22 but it required additional vector with all nodes, I decided it's too much of a memory footprint.

libp2p/NodeTable.cpp
```diff
@@ -262,57 +263,29 @@ void NodeTable::doDiscoveryRound(
...
+        return _node1.first < _node2.first;
+    };
+
+    std::multiset<pair<int, shared_ptr<NodeEntry>>, decltype(distanceToTargetLess)>
+        nodesByDistanceToTarget(distanceToTargetLess);
+    for (auto const& bucket : m_buckets)
+        for (auto const& nodeWeakPtr : bucket.nodes)
+            if (auto node = nodeWeakPtr.lock())
+            {
+                nodesByDistanceToTarget.emplace(distance(_target, node->id()), node);
```
gumb0: This is currently very inefficient, because it calculates `sha3(m_hostNodeID)` many times in the loop and `sha3(node->id())` every time `nearestNodeEntries` is called.

I will optimize it in the next PR by calculating hashes only once and making distance function get hashes as input.

test/unittests/libp2p/net.cpp
```diff
@@ -550,6 +551,93 @@ BOOST_AUTO_TEST_CASE(noteActiveNodeEvictsTheNodeWhenBucketIsFull)
...
+    nodeTableHost.populate(1);
+
+    auto& nodeTable = nodeTableHost.nodeTable;
+    vector<shared_ptr<NodeEntry>> const nearest = nodeTable->nearestNodeEntries(NodeID::random());
+
+    BOOST_REQUIRE_EQUAL(nearest.size(), 1);
+    BOOST_REQUIRE_EQUAL(nearest.front()->id(), nodeTableHost.testNodes.front().first);
+}
+
+unsigned xorDistance(h256 const& _h1, h256 const& _h2)
```
gumb0: I will move this function to `NodeTable` implementation in the next PR.

gumb0:
Rebased.

libp2p/NodeTable.cpp
gumb0: https://github.com/ethereum/aleth/pull/5586

PR #5586, opened 2019-04-30
Optimize NodeTable::nearestNodeEntries

gumb0:
Get rid of calculating hash many times in the loop.
Make `distance` function get hashes as input. Calculate hash for each `NodeEntry` only once and save it as a member of `NodeEntry`.

### DIFF
--- a/libp2p/NodeTable.h
+++ b/libp2p/NodeTable.h
@@ -112,7 +112,14 @@ public:
         bool _enabled = true, bool _allowLocalDiscovery = false);
 
     /// Returns distance based on xor metric two node ids. Used by NodeEntry and NodeTable.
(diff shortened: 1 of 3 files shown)


gumb0:
Needs rebase

gumb0:
Rebased.

libp2p/NodeTable.h
```diff
@@ -112,7 +112,14 @@ class NodeTable : UDPSocketEvents
         bool _enabled = true, bool _allowLocalDiscovery = false);
 
     /// Returns distance based on xor metric two node ids. Used by NodeEntry and NodeTable.
```
halfalicious: (Nit) Returns distance based on xor metric OF two node ids

PR #5589, opened 2019-05-02
Make log output always line-buffered

gumb0:
Grid team reports that when reading log output from a client process through a pipe,  other clients make it line-buffered (non-default behavior for stdout, which is line-buffered only when attached to tty)

Here I enable flushing after each line in boost.log.

The difference is noticeable when running commands like
```
>aleth/aleth >aleth.log &
>tail -f aleth.log
```

cc @evertonfraga


evertonfraga:
amazing, @gumb0. thanks a lot!

evertonfraga:
I just made a build of `aleth` and it worked as intended. waiting for the RC release ;)

PR #5591, opened 2019-05-04
Network logging bugfixes and improvements

halfalicious:
Various network logging bugfixes and changes. Bugfixes:
* Move "unsolicited neighbours packet" message from "net" to "discov" log channel since that's the channel that the rest of the discovery messages are logged to
* Fix log message that we write when we send a p2p capability pong (we incorrectly say we're sending a ping)

Other logging changes:
* Log messages related to p2p capability messaging to a new "p2pcap" log channel
* Log incoming capability packet type string
* Add node id and endpoint information to some network log messages
* Remove redundant "Packet" from p2p capability packet type strings
* Remove logging of RLP payloads in Session (this adds clutter to the log output and makes it harder to visually parse - I think it's safe to remove this information since I can't see us ever using it for debugging, if it turns out we need it we can always add it back)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Added: [#5564](https://github.com/ethereum/aleth/pull/5564) Improved help output of Aleth by adding list of channels.
 - Added: [#5575](https://github.com/ethereum/aleth/pull/5575) Log active peer count and peer list every 30 seconds.
 - Added: [#5580](https://github.com/ethereum/aleth/pull/5580) Enable syncing from ETC nodes for blocks < dao hard fork block.
+- Added: [#5591](https://github.com/ethereum/aleth/pull/5591) Network logging bugfixes and improvements.
 - Changed: [#5559](https://github.com/ethereum/aleth/pull/5559) Update peer validation error messages.
 - Changed: [#5568](https://github.com/ethereum/aleth/pull/5568) Improve rlpx handshake log messages and create new rlpx log channel.
 - Changed: [#5576](https://github.com/ethereum/aleth/pull/5576) Moved sstore_combinations and static_Call50000_sha256 tests to stTimeConsuming test suite. (testeth runs them only with `--all` flag)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Added: [#5564](https://github.com/ethereum/aleth/pull/5564) Improved help output of Aleth by adding list of channels.
 - Added: [#5575](https://github.com/ethereum/aleth/pull/5575) Log active peer count and peer list every 30 seconds.
 - Added: [#5580](https://github.com/ethereum/aleth/pull/5580) Enable syncing from ETC nodes for blocks < dao hard fork block.
-- Added: [#5591](https://github.com/ethereum/aleth/pull/5591) Network logging bugfixes and improvements.
+- Added: [#5591](https://github.com/ethereum/aleth/pull/5591) Network logging bugfixes and improvements and add p2pcap log channel.
 - Changed: [#5559](https://github.com/ethereum/aleth/pull/5559) Update peer validation error messages.
 - Changed: [#5568](https://github.com/ethereum/aleth/pull/5568) Improve rlpx handshake log messages and create new rlpx log channel.
 - Changed: [#5576](https://github.com/ethereum/aleth/pull/5576) Moved sstore_combinations and static_Call50000_sha256 tests to stTimeConsuming test suite. (testeth runs them only with `--all` flag)
--- a/libp2p/Session.cpp
+++ b/libp2p/Session.cpp
@@ -86,8 +86,8 @@ bool Session::readPacket(uint16_t _capId, unsigned _packetType, RLP const& _r)
 {
     m_lastReceived = chrono::steady_clock::now();
     clog(VerbosityTrace, "net") << "Received " << capabilityPacketTypeToString(_packetType) << "("
-                                << _packetType << ")"
-                                << " from " << m_info.id << "@" << m_socket->remoteEndpoint();
+                                << _packetType << ") from " << m_info.id << "@"
+                                << m_socket->remoteEndpoint();
     try // Generic try-catch block designed to capture RLP format errors - TODO: give decent diagnostics, make a bit more specific over what is caught.
     {
         // v4 frame headers are useless, offset packet type used
@@ -136,14 +136,15 @@ bool Session::interpret(P2pPacketType _t, RLP const& _r)
         else
         {
             reason = reasonOf(r);
-            cnetlog << "Disconnect (reason: " << reason << ")";
+            cnetlog << "Disconnect (reason: " << reason << ") from " << m_info.id << "@"
+                    << m_socket->remoteEndpoint();
             drop(DisconnectRequested);
         }
         break;
     }
     case PingPacket:
     {
-        cnetdetails << "Ping " << m_info.id;
+        cnetdetails << "Pong to " << m_info.id << "@" << m_socket->remoteEndpoint();
         RLPStream s;
         sealAndSend(prep(s, PongPacket));
         break;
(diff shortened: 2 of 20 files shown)


halfalicious:
Rebased + added changelog update

halfalicious:
Squashed some commits

libp2p/Session.cpp
```diff
@@ -139,43 +123,44 @@ bool Session::readPacket(uint16_t _capId, PacketType _packetType, RLP const& _r)
...
 }
 
-bool Session::interpret(PacketType _t, RLP const& _r)
+bool Session::interpret(P2pPacketType _t, RLP const& _r)
 {
     switch (_t)
     {
     case DisconnectPacket:
     {
+        clog(VerbosityTrace, "p2pcap") << p2pPacketTypeToString(DisconnectPacket) << " from " << m_info.id << "@" << m_socket->remoteEndpoint();
```
gumb0: Looks like this can be moved out of `switch` and done once for any packet

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Added: [#5564](https://github.com/ethereum/aleth/pull/5564) Improved help output of Aleth by adding list of channels.
 - Added: [#5575](https://github.com/ethereum/aleth/pull/5575) Log active peer count and peer list every 30 seconds.
 - Added: [#5580](https://github.com/ethereum/aleth/pull/5580) Enable syncing from ETC nodes for blocks < dao hard fork block.
+- Added: [#5591](https://github.com/ethereum/aleth/pull/5591) Network logging bugfixes and improvements.
```
gumb0: I think we should mention that `p2pcap` log channel is added.

libp2p/Session.cpp
```diff
@@ -213,7 +199,6 @@ bool Session::checkPacket(bytesConstRef _msg)
 void Session::send(bytes&& _msg)
 {
     bytesConstRef msg(&_msg);
-    clog(VerbosityTrace, "net") << "<- " << RLP(msg.cropped(1));
```
gumb0: I think after removing this we don't log anything for outgoing packets of capabilities. Maybe leave it just saying that there's some outgoing packet? Probably you could also look at the first byte to get the packet type

libp2p/Session.cpp
```diff
@@ -100,16 +82,18 @@ int Session::rating() const
     return m_peer->m_rating;
 }
 
-bool Session::readPacket(uint16_t _capId, PacketType _packetType, RLP const& _r)
+bool Session::readPacket(uint16_t _capId, unsigned _packetType, RLP const& _r)
 {
     m_lastReceived = chrono::steady_clock::now();
-    clog(VerbosityTrace, "net") << "-> " << _packetType << " " << _r;
+    clog(VerbosityTrace, "net") << "Received " << capabilityPacketTypeToString(_packetType) << "("
+                                << _packetType << ") from " << m_info.id << "@"
```
gumb0: I'm not sure about logging everywhere in Session class `<< m_info.id << "@" << m_socket->remoteEndpoint();`

Here `LOG_SCOPED_CONTEXT` is supposed to add a prefix of id+clientVersion for each log message inside log handlers. Isn't at least `m_info.id` duplicated in logs this way?

I suspect that sometimes IDs could be different, like when `BlockChainSync` code is called inside the handler of the message from one peer, and this code sends a message to another peer.
In this case, maybe it's better to remove `LOG_SCOPED_CONTEXT` stuff now? Though then clientVersion of the peers won't as visible in logs as now, it seems it's useful, too... What do you think?

libp2p/Session.cpp
halfalicious: Ah of course

libp2p/Session.cpp
halfalicious: Good idea - we appear to log messages in other parts of the code when we perform actions which result in capability packets being sent (e.g. EthereumPeer::requestBlockHeaders) but it could also be nice to be able to track exactly when the capability packets are being sent and exactly which packets are being sent.

libp2p/Session.cpp
halfalicious: I like having the client information in the log messages, though I don't have a concrete example of when it's helped me debug something yet. I think that prefixing each log message with the node id + client information makes the log messages a little counter-intuitive, for example, the following message could be logged when we receive a Pong p2p capability message from a node but the message reads as if 865a6325 received the pong:

`p2p p2pcap 865a6325…|Parity-Ethereum/v2.2.9-stable-5d5b372-20190203/x86_64-linux-gnu/rustc1.31.1 Receiving Pong`

An option is to remove the LOG_SCOPED_CONTEXT and to instead log the client information at the end of each message e.g. 

`TRACE 05-07 21:03:32 p2p  p2pcap Pong to ##865a6325…@52.176.100.77:30303 (Parity-Ethereum/v2.2.9-stable-5d5b372-20190203/x86_64-linux-gnu/rustc1.31.1)`

I think this is more readable, though it makes the log statements themselves more cumbersome (as you called out in my RLPXHandshake logging changes). Perhaps we could also create a logging wrapper function such as:

```
inline void logMessage(char const* _message, char const* _channel, Verbosity _verbosity)
{
      clog(_verbosity, _channel) << _message << " (" << m_info.clientVersion << "|" << m_info.id << "|" << m_socket->remoteEndpoint() << ")";
}
```

Thoughts?

gumb0: I think I dislike repeating clientVersion many times, it bloats the log. If we output it once, say for Hello message, it should be easy enough to find it in the log for particular NodeID, when we need it.

gumb0: Another thing: if we repeat this `<< m_info.id << "@" << m_socket->remoteEndpoint();` in every message and these values are constant, can we create a logger for this, that adds them as a Context attribute, similar to what you did in `RLPXHandshake` ?

This and removing `LOG_SCOPED_CONTEXT` can be done as a separate PR.

gumb0: The downside to removing `LOG_SCOPED_CONTEXT` would be that we won't see this context for the messages coming from `BlockChainSync`, not sure how useful it is. Maybe it's fine to remove, as all network handling happens sequentially in one thread and we should be able to find this context earlier in the log.

> 
> I think I dislike repeating clientVersion many times, it bloats the log. If we output it once, say for Hello message, it should be easy enough to find it in the log for particular NodeID, when we need it.

Yes, but the Hello might be long in the past so it might not be accessible in the console log output anymore. This might be okay though since we also periodically log connected peers, though only at trace level: https://github.com/ethereum/aleth/blob/7ecf42b4e4afd9fb5c323b2338c5ac60d6f41dea/libp2p/Host.cpp#L868

`Host::m_detailLogger` is defined here: https://github.com/ethereum/aleth/blob/7ecf42b4e4afd9fb5c323b2338c5ac60d6f41dea/libp2p/Host.h#L382

Maybe we should change the verbosity level of this log message to debug level so that it's more readily available for debugging purposes?

halfalicious:
>  logger in `Session` with Context that automatically adds NodeID and IP to the messages, I think it's still better to create a member variable logger(s) instead of using `clog(VerbosityTrace, "p2pcap")`

Agreed. For reference, clog(VerbosityTrace, "p2pcap") and LOG(m_logger) where m_logger is defined as Logger m_logger{createLogger(VerbosityTrace, "p2pcap")} are functionally equivalent?

libp2p/Session.cpp
halfalicious: > e repeat this `<< m_info.id << "@" << m_socket->remoteEndpoint();` in every message and these values are constant, can we create a logger for this, that adds them as a Context attribute...

@gumb0 Is there a way to add the context to the end of a log message? I think including remote endpoint information at the beginning of a log message makes the log a little counter-intuitive.

gumb0:
@halfalicious  Not exactly, `clog(...)` macro uses a global thread-safe logger, while `createLogger()` creates a non-thread-safe logger, so it doesn't have the overhead of thread-safety, and so should be more efficient.


Actually now I think that the reason global loggers like `cnetdetails` were used in `Session` class originally could be that some methods are called from different threads... Not sure if it's correct assumption, so if you change it to member variable logger, take care to use it only in network thread

libp2p/Session.cpp
gumb0: > Is there a way to add the context to the end of a log message?

Not right now, but I think it's simple enough to add this. We could rename `Context` attribute to `Prefix`, then add another `Suffix` one, and your logger would fill the `Suffix`

The changes will be need only here https://github.com/ethereum/aleth/blob/4e63cf289c2d1902ece63ce53545b25bd0257bae/libdevcore/Log.cpp#L51
and when we output it in formatter here
https://github.com/ethereum/aleth/blob/4e63cf289c2d1902ece63ce53545b25bd0257bae/libdevcore/Log.cpp#L99-L100

PR #5595, opened 2019-05-08
Log block queue status in BlockChainSync

halfalicious:
Log the block queue status when BlockChainSync cannot sync with a peer due to syncing being paused and when it cannot request blocks due to the block queue being full.

Also convert some BlockQueue constants to constexpr and wrap checking for a paused sync in a helper function (BlockChainSync::isSyncPaused) in the interest of consistency since pausing a sync is done via a helper function (BlockChainSync::pauseSync).

### DIFF
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -236,8 +236,13 @@ void BlockChainSync::syncPeer(NodeID const& _peerID, bool _force)
         return;
     }
 
-    if (m_state == SyncState::Waiting)
+    if (isSyncPaused())
+    {
+        LOG(m_loggerDetail) << "Can't sync with peer " << _peerID
+                            << " - sync state is paused. Block queue status: "
+                            << host().bq().status();
         return;
+    }
 
     u256 td = host().chain().details().totalDifficulty;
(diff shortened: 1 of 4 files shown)


halfalicious:
I don't think I need a changelog for this since it's just minor logging tweaks... @chfast  / @gumb0 : let me know if you disagree

halfalicious:
This is strange - this test is failing on osx:
```

567/580 Test #572: JsonRpcSuite/test_setChainParams ......................................................................................***Failed    0.87 sec
Running tests using path: "/Users/distiller/project/test/jsontests"
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: recursive_mutex lock failed: Invalid argument
Running 1 test case...
Test Case "test_setChainParams": 
unknown location:0: fatal error: in "JsonRpcSuite/test_setChainParams": signal: SIGABRT (application abort requested)
/Users/distiller/project/test/unittests/libweb3jsonrpc/jsonrpc.cpp:718: last checkpoint: "test_setChainParams" fixture dtor

```

This test doesn't do any syncing so it shouldn't be impacted by my changes...I'll restart the run and see if it's a transient issue.

PR #5597, opened 2019-05-09
Minor fix in RLP

gumb0:
`RLP::toArray()` should throw exception only when `ThrowOnFail` flag is passed, but currently it throws always from `itemCountStrict()` when called for non-list RLP.


gumb0:
@chfast fixed

PR #5599, opened 2019-05-12
Check for existing RLPX handshake when connecting to a node

### LINKED ISSUE #5600: Aleth tries to initiate another connection to node while RLPX handshake is still in progress
Aleth tries to initiate a connection to a node while an RLPX handshake is already in progress with that same node. For example, take bootnode `865a6325 `in the following log...Aleth opens the TCP socket to `865a6325  `and then starts the RLPX handshake: 

TRACE 05-12 20:42:35 p2p  net    Attempting connection to ##15ac307a…@52.176.7.10:30303 from ##a6abf591…
TRACE 05-12 20:42:35 p2p  net    Attempting connection to ##a979fb57…@52.16.188.185:30303 from ##a6abf591…
**TRACE 05-12 20:42:35 p2p  net    Attempting connection to ##865a6325…@52.176.100.77:30303 from ##a6abf591…**
TRACE 05-12 20:42:35 p2p  net    Attempting connection to ##3f1d1204…@13.93.211.84:30303 from ##a6abf591…
TRACE 05-12 20:42:35 p2p  net    Attempting connection to ##78de8a09…@191.235.84.50:30303 from ##a6abf591…

...
TRACE 05-12 20:42:35 p2p  net    Starting RLPX handshake with ##15ac307a…@52.176.7.10:30303
TRACE 05-12 20:42:35 p2p  rlpx   egress auth to ##15ac307a…@52.176.7.10:30303
TRACE 05-12 20:42:35 p2p  rlpx   egress ack from ##aa36fdf3…@13.93.211.84:30303
**TRACE 05-12 20:42:35 p2p  net    Starting RLPX handshake with ##865a6325…@52.176.100.77:30303**
TRACE 05-12 20:42:35 p2p  rlpx   egress auth to ##865a6325…@52.176.100.77:30303
TRACE 05-12 20:42:35 p2p  rlpx   egress ack from ##3f1d1204…@13.93.211.84:30303

The node then gets added to the node table via discovery (which means that a node table `NodeEntryAdded `event is generated which means that when the host processes the event - in `Host::onNodeTableEvent ` - it calls `Host::connect` on that node) and Aleth creates another TCP connection to the node and initiates another RLPX handshake:

TRACE 05-12 20:42:35 p2p  rlpx   egress ack from ##78de8a09…@191.235.84.50:30303
TRACE 05-12 20:42:35 p2p  rlpx   egress ack from ##1118980b…@52.74.57.123:30303
DEBUG 05-12 20:42:35 p2p  net    p2p.host.nodeTable.events.nodeEntryAdded ##865a6325…
**TRACE 05-12 20:42:35 p2p  net    Attempting connection to ##865a6325…@52.176.100.77:30303 from ##a6abf591…
TRACE 05-12 20:42:35 p2p  net    Starting RLPX handshake with ##865a6325…@52.176.100.77:30303**
TRACE 05-12 20:42:35 p2p  rlpx   egress auth to ##865a6325…@52.176.100.77:30303

The first handshake completes and we start a new session with that peer. The second handshake completes and we try to start another new session but detect that we already have a session with that peer so we disconnect:

TRACE 05-12 20:42:37 p2p  rlpx   egress Frame body from ##865a6325…@52.176.100.77:30303
TRACE 05-12 20:42:37 p2p  rlpx   egress Hello verified. Starting session with ##865a6325…@52.176.100.77:30303
DEBUG 05-12 20:42:37 p2p  net    Starting peer session with Parity-Ethereum/v2.2.9-stable-5d5b372-20190203/x86_64-linux-gnu/rustc1.31.1 V[5] ##865a6325… (eth,63) 30303
DEBUG 05-12 20:42:37 p2p  net    Session already exists for peer with id ##865a6325…
TRACE 05-12 20:42:37 p2p  net    Disconnecting (our reason: Peer was already connected.) from ##865a6325…@52.176.100.77:30303
TRACE 05-12 20:42:37 p2p  net    Sending Disconnect to ##865a6325…@52.176.100.77:30303
TRACE 05-12 20:42:37 p2p  net    Closing 52.176.100.77:30303 (Peer was already connected.)
DEBUG 05-12 20:42:37 p2p  net    865a6325…|Parity-Ethereum/v2.2.9-stable-5d5b372-20190203/x86_64-linux-gnu/rustc1.31.1 Closing peer session :-(

However, then the peer disconnects from us (I'm assuming because we sent a disconnect p2p capability packet to them to disconnect the duplicate session) so we presumably end up with no connection to the node:

TRACE 05-12 20:42:38 p2p  net    865a6325…|Parity-Ethereum/v2.2.9-stable-5d5b372-20190203/x86_64-linux-gnu/rustc1.31.1 Error reading: An existing connection was forcibly closed by the remote host
TRACE 05-12 20:42:38 p2p  net    865a6325…|Parity-Ethereum/v2.2.9-stable-5d5b372-20190203/x86_64-linux-gnu/rustc1.31.1 Closing 52.176.100.77:30303 (Low-level TCP communication error.)
DEBUG 05-12 20:42:38 p2p  net    865a6325…|Parity-Ethereum/v2.2.9-stable-5d5b372-20190203/x86_64-linux-gnu/rustc1.31.1 Closing peer session :-(

We eventually start trying to connect to the node again (since it's a bootnode and therefore a required peer):

TRACE 05-12 20:42:40 p2p  rlpx   egress ack from ##a979fb57…@52.16.188.185:30303
**TRACE 05-12 20:42:41 p2p  net    Attempting connection to ##865a6325…@52.176.100.77:30303 from ##a6abf591…**
TRACE 05-12 20:42:41 p2p  net    Attempting connection to ##669f45b6…@52.74.57.123:30303 from ##a6abf591…
**TRACE 05-12 20:42:41 p2p  net    Starting RLPX handshake with ##865a6325…@52.176.100.77:30303**

Full log: [duplicate_connection.txt](https://github.com/ethereum/aleth/files/3171084/duplicate_connection.txt)

halfalicious:
Fix #5600 

Check for existing RLPX handshake in `Host::connect` before attempting a new connection to a node - this is required because `Host::connect` only checks `m_pendingPeerConns` before starting a new connection, and entries are removed from `m_pendingPeerConns` after a new RLPX handshake has started rather than after the handshake has completed.

I've chosen to check for an existing RLPX handshake rather than garbage-collecting `m_pendingPeerConns` in `Host::run` since the former only adds cost during a new connection while the latter would add cost on every `Host::run` invocation.

### DIFF
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -421,7 +421,7 @@ void Host::onNodeTableEvent(NodeID const& _n, NodeTableEventType const& _e)
     }
 }
 
-bool Host::isConnecting(NodeID const& _id)
+bool Host::isHandshaking(NodeID const& _id) const
 {
     Guard l(x_connecting);
     for (auto const& cIter : m_connecting)
@@ -657,9 +657,9 @@ void Host::connect(shared_ptr<Peer> const& _p)
         return;
     }
 
-    if (isConnecting(_p->id))
+    if (isHandshaking(_p->id))
     {
-        cwarn << "Aborted connection. Connection to peer already in progress: " << _p->id << "@"
+        cwarn << "Aborted connection. RLPX handshake to peer already in progress: " << _p->id << "@"
               << _p->endpoint;
         return;
     }
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -421,6 +421,18 @@ void Host::onNodeTableEvent(NodeID const& _n, NodeTableEventType const& _e)
     }
 }
 
+bool Host::isConnecting(NodeID const& _id)
+{
+    Guard l(x_connecting);
+    for (auto const& cIter : m_connecting)
+    {
+        std::shared_ptr<RLPXHandshake> const connecting = cIter.lock();
+        if (connecting && connecting->remote() == _id)
+            return true;
+    }
+    return false;
+}
+
 void Host::determinePublic()
 {
     // set m_tcpPublic := listenIP (if public) > public > upnp > unspecified address.
@@ -638,15 +650,24 @@ void Host::connect(shared_ptr<Peer> const& _p)
         cwarn << "Network not running so cannot connect to peer " << _p->id << "@" << _p->address();
         return;
     }
+
     if (!haveCapabilities())
     {
         cwarn << "No capabilities registered so cannot connect to peer " << _p->id << "@" << _p->address();
         return;
     }
-    
+
+    if (isConnecting(_p->id))
+    {
+        cwarn << "Aborted connection. Connection to peer already in progress: " << _p->id << "@"
+              << _p->endpoint;
+        return;
+    }
+
     if (havePeerSession(_p->id))
     {
-        cnetdetails << "Aborted connect. Node already connected.";
+        cnetdetails << "Aborted connection. Peer already connected: " << _p->id << "@"
+                    << _p->endpoint;
         return;
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Changed: [#5589](https://github.com/ethereum/aleth/pull/5589) Make aleth output always line-buffered even when redirected to file or pipe.
 - Fixed: [#5562](https://github.com/ethereum/aleth/pull/5562) Don't send header request messages to peers that haven't sent us Status yet.
 - Fixed: [#5581](https://github.com/ethereum/aleth/pull/5581) Fixed finding neighbour nodes in Discovery.
+- Fixed: [#5599](https://github.com/ethereum/aleth/pull/5600) Prevent aleth from attempting concurrent connection to node which results in disconnect of original connection.
 
 ## [1.6.0] - 2019-04-16
 
--- a/libp2p/Host.h
+++ b/libp2p/Host.h
@@ -260,7 +260,7 @@ private:
     
     bool havePeerSession(NodeID const& _id) { return !!peerSession(_id); }
 
-    bool isConnecting(NodeID const& _id);
+    bool isHandshaking(NodeID const& _id) const;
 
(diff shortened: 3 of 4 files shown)


halfalicious:
Another approach would be to remove the peer entry from `Host::m_pendingPeerConns` once the RLPX handshake has been completed and a new session has been created in `RLPXHandshake::transition`: https://github.com/ethereum/aleth/blob/32f9c88f65fc30d7b682de651e2d40d8b7e26078/libp2p/RLPxHandshake.cpp#L454-L462

However, I think this would be a little messier since it would require adding 2 functions to Host - 1 which enables peer lookup by node ID and returns `std::shared_ptr<Peer>` and another which removes an entry from `m_pendingPeerConns` given a `std::shared_ptr<Peer>`. This would also open us up to synchronization issues unless we protected `m_pendingPeerConns ` and `m_peers` with synchronization primitives (which would add to the runtime cost of these functions).

gumb0:
Hm, what about removing from `Host::m_pendingPeerConns` inside `Host::startPeerSession` (perhaps after adding to `m_sessions`) instead of at the end of after handshake start in `Host::connect` ) ?

Might still be messier, considering that we'd also need to somehow remove them when handshake failed...

libp2p/Host.cpp
```diff
@@ -421,6 +421,18 @@ void Host::onNodeTableEvent(NodeID const& _n, NodeTableEventType const& _e)
     }
 }
 
+bool Host::isConnecting(NodeID const& _id)
```
gumb0: Can it be const method? (make mutex mutable if needed)

gumb0: Maybe it should be more precisely called `isHandshaking`, because connection itself actually already exists during handshake

halfalicious: Need to make mutex mutable, will do this and rename the method.

halfalicious: @gumb0 Addressed

PR #5602, opened 2019-05-15
Dynamic predicting of external IP and UDP port

gumb0:
This improves how we track what our peers see as our external IP address / UDP port.

Previously we updated external port every time we receive new port number in Pong packet, now we count which port is reported by most of the peers and update it when new value is reported from more than 10 peers.

Also external IP address is similarly updated from Pong packet, but only unless it was given as a command line argument (or detected by UPNP).

Here's go-ethereum code doing quite the same:
https://github.com/ethereum/go-ethereum/blob/HEAD@%7B2019-05-15T10:20:04Z%7D/p2p/netutil/iptrack.go
https://github.com/ethereum/go-ethereum/blob/97d3615612a49488d02807944696d001b88f9e0d/p2p/enode/localnode.go#L180-L202

### DIFF
--- a/libp2p/EndpointTracker.cpp
+++ b/libp2p/EndpointTracker.cpp
@@ -1,0 +1,76 @@
+// Aleth: Ethereum C++ client, tools and libraries.
+// Copyright 2019 Aleth Authors.
+// Licensed under the GNU General Public License, Version 3.
+
+#include "EndpointTracker.h"
+
+namespace dev
+{
+namespace p2p
+{
+namespace
+{
+// Interval during which each endpoint statement is kept
+constexpr std::chrono::minutes c_statementTimeToLiveMin{5};
+}  // namespace
+
+// Register the statement about endpoint from one othe peers.
+// Returns number of currently kept statements in favor of _externalEndpoint
+size_t EndpointTracker::addEndpointStatement(
+    bi::udp::endpoint const& _sourceEndpoint, bi::udp::endpoint const& _externalEndpoint)
+{
+    // remove previous statement by this peer
+    auto it = m_statementsMap.find(_sourceEndpoint);
+    if (it != m_statementsMap.end())
+        removeStatement(it);
+
+    return addStatement(_sourceEndpoint, _externalEndpoint);
+}
+
+// Find endpoint with max number of statemens
+bi::udp::endpoint EndpointTracker::bestEndpoint() const
+{
+    size_t maxCount = 0;
+    bi::udp::endpoint bestEndpoint;
+    for (auto const& endpointAndCount : m_endpointStatementCountMap)
+        if (endpointAndCount.second > maxCount)
+            std::tie(bestEndpoint, maxCount) = endpointAndCount;
+
+    return bestEndpoint;
+}
+
+// Remove old statements
+void EndpointTracker::garbageCollectStatements()
+{
+    auto const expiration = std::chrono::steady_clock::now() - c_statementTimeToLiveMin;
+    for (auto it = m_statementsMap.begin(); it != m_statementsMap.end();)
+    {
+        if (it->second.second > expiration)
+            it = removeStatement(it);
+        else
+            ++it;
+    }
+}
+
+size_t EndpointTracker::addStatement(
+    bi::udp::endpoint const& _sourceEndpoint, bi::udp::endpoint const& _externalEndpoint)
+{
+    EndpointAndTimePoint endpointAndTime{_externalEndpoint, std::chrono::steady_clock::now()};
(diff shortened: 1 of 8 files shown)


libp2p/EndpointTracker.h
```diff
@@ -0,0 +1,47 @@
...
+// Keeps track of what external endpoint is seen by every peer
+// and finds which endpoint is reported most often
+class EndpointTracker
+{
+public:
+    // Register the statement about endpoint from one othe peers.
+    // Returns number of currently kept statements in favor of _externalEndpoint
+    size_t addEndpointStatement(
+        bi::udp::endpoint const& _sourceEndpoint, bi::udp::endpoint const& _externalEndpoint);
+
```
halfalicious: (nit) typo (statemens)

libp2p/EndpointTracker.cpp
```diff
@@ -0,0 +1,71 @@
...
+// Licensed under the GNU General Public License, Version 3.
+
+#include "EndpointTracker.h"
+
+namespace dev
+{
+namespace p2p
+{
+
+// Register the statement about endpoint from one othe peers.
```
chfast: You should use `///` comments for documentation.

libp2p/EndpointTracker.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    return addStatement(_sourceEndpoint, _externalEndpoint);
+}
+
+// Find endpoint with max number of statemens
+bi::udp::endpoint EndpointTracker::bestEndpoint() const
+{
+    size_t maxCount = 0;
+    bi::udp::endpoint bestEndpoint;
+    for (auto const& endpointAndCount : m_endpointStatementCountMap)
```
chfast: `std::max_element`

libp2p/EndpointTracker.h
```diff
@@ -0,0 +1,47 @@
...
+    // Find endpoint with max number of statements
+    bi::udp::endpoint bestEndpoint() const;
+
+    // Remove statements older than _timeToLive
+    void garbageCollectStatements(std::chrono::seconds const& _timeToLive);
+
+private:
+    using EndpointAndTimePoint =
+        std::pair<bi::udp::endpoint, std::chrono::steady_clock::time_point>;
+    using SourceToStatementMap = std::map<bi::udp::endpoint, EndpointAndTimePoint>;
```
chfast: This can be unsorted map.

libp2p/EndpointTracker.h
```diff
@@ -0,0 +1,47 @@
...
+
+    size_t addStatement(
+        bi::udp::endpoint const& _sourceEndpoint, bi::udp::endpoint const& _externalEndpoint);
+
+    SourceToStatementMap::iterator removeStatement(SourceToStatementMap::iterator _it);
+
+    // Statements about our external endpoint, maps statement source peer => endpoint, timestamp
+    SourceToStatementMap m_statementsMap;
+    // map external endpoint => how many sources reported it
+    std::map<bi::udp::endpoint, size_t> m_endpointStatementCountMap;
```
chfast: This can be unsorted map.

libp2p/EndpointTracker.cpp
halfalicious: @chfast : I see a mix of both in our code base...should we always use ///? Presumably git clang-format won't set the right comment formatting for us?

chfast: The `///` or `/**` are doxygen comments so we could generate documentation out of it. Which we will probably never do so this is not critical. This should not matter for clang-format at all.

libp2p/EndpointTracker.h
gumb0: Boost doesn't have a hashing function defined for `udp::endpoint` (but has `operator<`), so I used `map` to avoid defining it myself

chfast: Can you check in the latest boost version? If not there we should report a bug.

gumb0: Not sure if it's a bug, maybe not every type should define hashing function?

chfast: Looks like good candidate for being a key in a hash map.

gumb0: Reported https://github.com/boostorg/asio/issues/245

nbougalis: Resurrecting this from the dead, just to say that I'm surprised that boostorg/asio#245 got no responses so far. We have been carrying custom hashing support for these types in `rippled` so it's been "out of sight, out of mind" for our codebase, but this really is something that boost should provide.

I'm happy to have Ripple sponsor a bounty to get the supported coded up, or to do it myself if nobody else is interested.

PR #5605, opened 2019-05-17
Various network-related logging changes

halfalicious:
* Move log messages from global thread-safe loggers to class-instance-loggers (where thread safety isn't needed because logging is done in functions called on a single thread)
* Add a log suffix to the `RLPXHandshake`, `Session`, and `EthereumPeer `log messages which is comprised of either the remote node ID and endpoint or just the remote node id (to remove the need of having to manually include this information in each log message). This suffix also removes the need for the `LOG_SCOPED_CONTEXT ` macro so I've removed it as well
* Manually log the peer ID where appropriate in the `BlockChainSync`, `EthereumCapability`, and `WarpCapability `classes (since these classes no longer automatically get peer client information in their log messages due to the removal of `LOG_SCOPED_CONTEXT ` in `Session`)
* Move `RLPXHandshake `log messages related to receiving specific messages from a remote node from before the read (which reads the message bytes from the socket) to after the read so identifying issues related to non-responsive nodes is more intuitive
* Remove redundant error messages from `RLPXHandshake`

Here's a snippet of the new logs:

```
DEBUG 05-23 21:07:41 p2p  net    Starting peer session with Geth/v1.7.1-stable-05101641/linux-amd64/go1.9 (protocol: V5) ##0546e7dd… capabilities: (eth,63) port: 0
DEBUG 05-23 21:07:41 p2p  net    New session for capability eth; idOffset: 16 with ##0546e7dd…@34.237.25.200:30303
TRACE 05-23 21:07:41 p2p  net    Status to (##0546e7dd…@34.237.25.200:30303)
TRACE 05-23 21:07:41 p2p  p2pcap Ping to (##0546e7dd…@34.237.25.200:30303)
TRACE 05-23 21:07:41 p2p  net    Ping to (##0546e7dd…@34.237.25.200:30303)
DEBUG 05-23 21:07:41 p2p  net    Peer connection successfully established with ##0546e7dd…@34.237.25.200:30303
TRACE 05-23 21:07:41 p2p  net    Starting RLPX handshake with ##374fd3c0…@35.198.202.233:30303
TRACE 05-23 21:07:41 p2p  rlpx   egress auth to (##374fd3c0…@35.198.202.233:30303)
TRACE 05-23 21:07:41 p2p  net    Starting RLPX handshake with ##348197fd…@47.91.239.54:30303
TRACE 05-23 21:07:41 p2p  rlpx   egress auth to (##348197fd…@47.91.239.54:30303)
DEBUG 05-23 21:07:41 p2p  net    p2p.host.nodeTable.events.nodeEntryAdded ##c8f41b59…
DEBUG 05-23 21:07:41 p2p  net    p2p.host.peers.events.peerAdded ##c8f41b59… 188.166.216.1:30303
TRACE 05-23 21:07:41 p2p  net    Attempting connection to ##c8f41b59…@188.166.216.1:30303 from ##5c1c3e78…
DEBUG 05-23 21:07:41 p2p  net    p2p.host.nodeTable.events.nodeEntryAdded ##fd6f0c38…
DEBUG 05-23 21:07:41 p2p  net    p2p.host.peers.events.peerAdded ##fd6f0c38… 31.3.23.102:30303
TRACE 05-23 21:07:41 p2p  net    Attempting connection to ##fd6f0c38…@31.3.23.102:30303 from ##5c1c3e78…
DEBUG 05-23 21:07:41 p2p  net    p2p.host.nodeTable.events.nodeEntryAdded ##c8e5b4b6…
DEBUG 05-23 21:07:41 p2p  net    p2p.host.peers.events.peerAdded ##c8e5b4b6… 101.207.224.48:40115
TRACE 05-23 21:07:41 p2p  net    Attempting connection to ##c8e5b4b6…@101.207.224.48:40115 from ##5c1c3e78…
TRACE 05-23 21:07:41 p2p  net    Starting RLPX handshake with ##9d676310…@18.197.177.150:30303
TRACE 05-23 21:07:41 p2p  rlpx   egress auth to (##9d676310…@18.197.177.150:30303)
TRACE 05-23 21:07:41 p2p  net    Received Status (16) from (##0546e7dd…@34.237.25.200:30303)
DEBUG 05-23 21:07:41 p2p  ethcap Status (from ##0546e7dd…): 63 / 4514 / #5e31ddf9…, TD: 3478195 = #f2df48a6…
DEBUG 05-23 21:07:41 p2p  sync   Peer ##0546e7dd… not suitable for sync: Network identifier mismatch. Host network id: 1, peer network id: 4514
TRACE 05-23 21:07:41 p2p  net    Disconnecting (our reason: Subprotocol reason.) from (##0546e7dd…@34.237.25.200:30303)
TRACE 05-23 21:07:41 p2p  net    Disconnect to (##0546e7dd…@34.237.25.200:30303)
```

### DIFF
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -196,7 +196,7 @@ void BlockChainSync::onPeerStatus(EthereumPeer const& _peer)
 
     if (!disconnectReason.empty())
     {
-        LOG(m_logger) << "Peer not suitable for sync: " << disconnectReason;
+        LOG(m_logger) << "Peer (" << _peer.id() << ") not suitable for sync: " << disconnectReason;
         m_host.capabilityHost().disconnect(_peer.id(), p2p::UserReason);
         return;
     }
@@ -255,13 +255,15 @@ void BlockChainSync::syncPeer(NodeID const& _peerID, bool _force)
     if (_force || peerTotalDifficulty > syncingDifficulty)
     {
         if (peerTotalDifficulty > syncingDifficulty)
-            LOG(m_logger) << "Discovered new highest difficulty";
+            LOG(m_logger) << "Discovered new highest difficulty (" << peerTotalDifficulty
+                          << ") from peer (" << peer.id() << ")";
 
         // start sync
         m_syncingTotalDifficulty = peerTotalDifficulty;
         if (m_state == SyncState::Idle || m_state == SyncState::NotSynced)
         {
             LOG(m_loggerInfo) << "Starting full sync";
+            LOG(m_logger) << "Syncing with peer " << peer.id();
             m_state = SyncState::Blocks;
         }
         peer.requestBlockHeaders(peer.latestHash(), 1, 0, false);
@@ -289,9 +291,8 @@ void BlockChainSync::requestBlocks(NodeID const& _peerID)
     clearPeerDownload(_peerID);
     if (host().bq().knownFull())
     {
-        LOG(m_loggerDetail)
-            << "Waiting for block queue before downloading blocks. Block queue status: "
-            << host().bq().status();
+        LOG(m_loggerDetail) << "Waiting for block queue before downloading blocks from " << _peerID
+                            << ". Block queue status: " << host().bq().status();
         pauseSync();
         return;
     }
@@ -379,7 +380,10 @@ void BlockChainSync::requestBlocks(NodeID const& _peerID)
             }
         }
         else
-            m_host.peer(_peerID).requestBlockHeaders(start, 1, 0, false);
+        {
+            LOG(m_loggerDetail) << "Requesting block headers from " << _peerID;
+            m_host.peer(_peerID).requestBlockHeaders(start, 1 /* count */, 0 /* skip */, false);
+        }
     }
 }
 
@@ -446,7 +450,7 @@ void BlockChainSync::onPeerBlockHeaders(NodeID const& _peerID, RLP const& _r)
     DEV_INVARIANT_CHECK;
     size_t itemCount = _r.itemCount();
     LOG(m_logger) << "BlocksHeaders (" << dec << itemCount << " entries) "
-                  << (itemCount ? "" : ": NoMoreHeaders");
+                  << (itemCount ? "" : ": NoMoreHeaders") << " from " << _peerID;
 
     if (m_daoChallengedPeers.find(_peerID) != m_daoChallengedPeers.end())
     {
@@ -462,17 +466,17 @@ void BlockChainSync::onPeerBlockHeaders(NodeID const& _peerID, RLP const& _r)
     clearPeerDownload(_peerID);
     if (m_state != SyncState::Blocks && m_state != SyncState::Waiting)
     {
-        LOG(m_logger) << "Ignoring unexpected blocks";
+        LOG(m_logger) << "Ignoring unexpected blocks from " << _peerID;
         return;
     }
     if (m_state == SyncState::Waiting)
     {
-        LOG(m_loggerDetail) << "Ignored blocks while waiting";
+        LOG(m_loggerDetail) << "Ignored blocks from " << _peerID << " while waiting";
         return;
     }
     if (itemCount == 0)
     {
-        LOG(m_loggerDetail) << "Peer does not have the blocks requested";
+        LOG(m_loggerDetail) << "Peer " << _peerID << " does not have the blocks requested";
         m_host.capabilityHost().updateRating(_peerID, -1);
     }
     for (unsigned i = 0; i < itemCount; i++)
@@ -481,17 +485,19 @@ void BlockChainSync::onPeerBlockHeaders(NodeID const& _peerID, RLP const& _r)
         unsigned blockNumber = static_cast<unsigned>(info.number());
         if (blockNumber < m_chainStartBlock)
         {
-            LOG(m_logger) << "Skipping too old header " << blockNumber;
+            LOG(m_logger) << "Skipping too old header " << blockNumber << " from " << _peerID;
             continue;
         }
         if (haveItem(m_headers, blockNumber))
         {
-            LOG(m_logger) << "Skipping header " << blockNumber << " (already downloaded)";
+            LOG(m_logger) << "Skipping header " << blockNumber << " (already downloaded) from "
+                          << _peerID;
             continue;
         }
         if (blockNumber <= m_lastImportedBlock && m_haveCommonHeader)
         {
-            LOG(m_logger) << "Skipping header " << blockNumber << " (already imported)";
+            LOG(m_logger) << "Skipping header " << blockNumber << " (already imported) from "
+                          << _peerID;
             continue;
         }
         if (blockNumber > m_highestBlock)
@@ -510,7 +516,7 @@ void BlockChainSync::onPeerBlockHeaders(NodeID const& _peerID, RLP const& _r)
                 // Start of the header chain in m_headers doesn't match our known chain,
                 // probably we've downloaded other fork
                 clog(VerbosityWarning, "sync")
-                    << "Unknown parent of the downloaded headers, restarting sync";
+                    << "Unknown parent of the downloaded headers, restarting sync with " << _peerID;
                 restartSync();
                 return;
             }
@@ -527,8 +533,9 @@ void BlockChainSync::onPeerBlockHeaders(NodeID const& _peerID, RLP const& _r)
                 if ((prevBlock && prevBlock->hash != info.parentHash()) || (blockNumber == m_lastImportedBlock + 1 && info.parentHash() != m_lastImportedBlockHash))
                 {
                     // mismatching parent id, delete the previous block and don't add this one
-                    clog(VerbosityWarning, "sync") << "Unknown block header " << blockNumber << " "
-                                                   << info.hash() << " (Restart syncing)";
+                    clog(VerbosityWarning, "sync")
+                        << "Unknown block header " << blockNumber << " " << info.hash()
+                        << " (Restart syncing with " << _peerID << ")";
                     m_host.capabilityHost().updateRating(_peerID, -1);
                     restartSync();
                     return ;
@@ -537,8 +544,8 @@ void BlockChainSync::onPeerBlockHeaders(NodeID const& _peerID, RLP const& _r)
                 Header const* nextBlock = findItem(m_headers, blockNumber + 1);
                 if (nextBlock && nextBlock->parent != info.hash())
                 {
-                    LOG(m_loggerDetail)
-                        << "Unknown block header " << blockNumber + 1 << " " << nextBlock->hash;
+                    LOG(m_loggerDetail) << "Unknown block header " << blockNumber + 1 << " "
+                                        << nextBlock->hash << " from " << _peerID;
                     // clear following headers
                     unsigned n = blockNumber + 1;
                     auto headers = m_headers.at(n);
@@ -590,20 +597,20 @@ void BlockChainSync::onPeerBlockBodies(NodeID const& _peerID, RLP const& _r)
     DEV_INVARIANT_CHECK;
     size_t itemCount = _r.itemCount();
     LOG(m_logger) << "BlocksBodies (" << dec << itemCount << " entries) "
-                  << (itemCount ? "" : ": NoMoreBodies");
+                  << (itemCount ? "" : ": NoMoreBodies") << " from " << _peerID;
     clearPeerDownload(_peerID);
     if (m_state != SyncState::Blocks && m_state != SyncState::Waiting) {
-        LOG(m_logger) << "Ignoring unexpected blocks";
+        LOG(m_logger) << "Ignoring unexpected blocks from " << _peerID;
         return;
     }
     if (m_state == SyncState::Waiting)
     {
-        LOG(m_loggerDetail) << "Ignored blocks while waiting";
+        LOG(m_loggerDetail) << "Ignored blocks from " << _peerID << " while waiting";
         return;
     }
     if (itemCount == 0)
     {
-        LOG(m_loggerDetail) << "Peer does not have the blocks requested";
+        LOG(m_loggerDetail) << "Peer (" << _peerID << ") does not have the blocks requested";
         m_host.capabilityHost().updateRating(_peerID, -1);
     }
     for (unsigned i = 0; i < itemCount; i++)
@@ -617,13 +624,14 @@ void BlockChainSync::onPeerBlockBodies(NodeID const& _peerID, RLP const& _r)
         auto iter = m_headerIdToNumber.find(id);
         if (iter == m_headerIdToNumber.end() || !haveItem(m_headers, iter->second))
         {
-            LOG(m_loggerDetail) << "Ignored unknown block body";
+            LOG(m_loggerDetail) << "Ignored unknown block body from " << _peerID;
             continue;
         }
         unsigned blockNumber = iter->second;
         if (haveItem(m_bodies, blockNumber))
         {
-            LOG(m_logger) << "Skipping already downloaded block body " << blockNumber;
+            LOG(m_logger) << "Skipping already downloaded block body " << blockNumber << " from "
+                          << _peerID;
             continue;
         }
         m_headerIdToNumber.erase(id);
@@ -755,7 +763,7 @@ void BlockChainSync::onPeerNewBlock(NodeID const& _peerID, RLP const& _r)
     unsigned blockNumber = static_cast<unsigned>(info.number());
     if (blockNumber > (m_lastImportedBlock + 1))
     {
-        LOG(m_loggerDetail) << "Received unknown new block";
+        LOG(m_loggerDetail) << "Received unknown new block from " << _peerID;
         // Update the hash of highest known block of the peer.
         // syncPeer will then request the highest block header to properly restart syncing
         peer.setLatestHash(h);
@@ -781,8 +789,9 @@ void BlockChainSync::onPeerNewBlock(NodeID const& _peerID, RLP const& _r)
         {
             if (!m_bodies.empty())
             {
-                LOG(m_loggerDetail)
-                    << "Block headers map is empty, but block bodies map is not. Force-clearing.";
+                LOG(m_loggerDetail) << "Block headers map is empty, but block bodies map is not. "
+                                       "Force-clearing (peer: "
+                                    << _peerID << ")";
                 m_bodies.clear();
             }
             completeSync();
@@ -815,7 +824,8 @@ void BlockChainSync::onPeerNewBlock(NodeID const& _peerID, RLP const& _r)
         u256 totalDifficulty = _r[1].toInt<u256>();
         if (totalDifficulty > peer.totalDifficulty())
         {
-            LOG(m_loggerDetail) << "Received block with no known parent. Peer needs syncing...";
+            LOG(m_loggerDetail) << "Received block with no known parent. Peer (" << _peerID
+                                << ") needs syncing...";
             syncPeer(_peerID, true);
         }
         break;
@@ -882,10 +892,12 @@ void BlockChainSync::onPeerNewHashes(
     auto& peer = m_host.peer(_peerID);
     if (peer.isConversing())
     {
-        LOG(m_loggerDetail) << "Ignoring new hashes since we're already downloading.";
+        LOG(m_loggerDetail) << "Ignoring new hashes since we're already downloading (peer: "
+                            << _peerID << ")";
         return;
     }
-    LOG(m_loggerDetail) << "Not syncing and new block hash discovered: syncing.";
+    LOG(m_loggerDetail) << "Not syncing and new block hash discovered: syncing (peer: " << _peerID
+                        << ")";
     unsigned knowns = 0;
     unsigned unknowns = 0;
     unsigned maxHeight = 0;
@@ -899,7 +911,7 @@ void BlockChainSync::onPeerNewHashes(
             knowns++;
         else if (status == QueueStatus::Bad)
         {
-            cwarn << "block hash bad!" << h << ". Bailing...";
+            cwarn << "block hash bad!" << h << ". Bailing... (peer: " << _peerID << ")";
             return;
         }
         else if (status == QueueStatus::Unknown)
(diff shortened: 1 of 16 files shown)


halfalicious:
@gumb0 : I've removed the `LOG_SCOPED_CONTEXT `from `Session`, which means we get no client or node ID information in `BlockChainSync `logs. I think it's a good idea to include node ID (and maybe endpoint) information in the `BlockChainSync `logs - I know that we can get these from other channels but I think we should try to make the logs for each channel as independent as possible from other channels so we have more freedom when it comes to enabling / disabling channels for debugging purposes. 

I don't think we can automatically include this information via the new boost logger `suffix `attribute since `BlockChainSync `is called to sync data from different peers, and I don't think we want to go back to using `LOG_SCOPED_CONTEXT ` in `Session ` either since it can result in information for the wrong node being logged (e.g.  `Session::~Session()` ->` EthereumCapability::onDisconnect()` -> `EthereumPeerObserver::onPeerAborting() `-> `BlockChainSync::onPeerAborting()` -> `BlockChainSync::continueSync()` where `continueSync() `calls` syncPeer()` for each of the connected peers).

The only other option I can think of is to manually include the node ID in each `BlockChainSync `log message but this is a bit cumbersome...can you think of a better approach?

gumb0:
No ideas how to make it more elegant, but still - almost all of the code of `BlockChainSync` is either handling eth messages or/and sending eth messages, so it's kind of tightly coupled with `EthereumPeer` and `EthereumCapability` and maybe it's enough to log peer ID only in the beginning of the handler in `EthereumCapability` class, or when we send a message in `EthereumPeer` class; and then log peer ID only in several special places in `BlockChainSync` like in a special case of `syncPeer` (special about it being that from a handler of one peer we access other peers)

gumb0:
Also in these higher-level classes (`BlockChainSync`, `EthereumCapability`, `EthereumPeer`) I'd probably log only Node IDs, endpoints are kind of less relevant here, than in lower-level networking.

libp2p/RLPxHandshake.h
```diff
@@ -90,8 +90,8 @@ class RLPXHandshake: public std::enable_shared_from_this<RLPXHandshake>
     void readAckEIP8();
     
     /// Closes connection and ends transitions.
-    void error();
-    
+    void error(boost::system::error_code _ech = boost::system::error_code());
```
chfast:
```suggestion
    void error(boost::system::error_code _ech = {});
```

libethereum/BlockChainSync.cpp
```diff
@@ -379,7 +380,10 @@ void BlockChainSync::requestBlocks(NodeID const& _peerID)
             }
         }
         else
-            m_host.peer(_peerID).requestBlockHeaders(start, 1, 0, false);
+        {
+            LOG(m_loggerDetail) << "Requesting block headers from " << _peerID;
```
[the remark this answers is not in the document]
halfalicious: Ah good point, will remove

halfalicious:
> 
> 
> Hm now I'm thinking `LOG_SCOPED_CONTEXT` or similar thing might have been useful, at least when there're many logging statements in one function, and instead of repeating `<< peerID` we could use `LOG_SCOPED_CONTEXT` once at the top of the function.
> 
> But maybe some messages would become less nice (the ones where now you output `<< peerID` in the middle of the message), so I don't know

@gumb0 : Yea, I thought of that but I'm not exactly sure how `LOG_SCOPED_CONTEXT` works so I wasn't sure if there were caveats to using it that I'm not aware of (e.g. perf overhead) which would make calling it in these functions inappropriate. Additionally, using it can introduce complications e.g. logging the wrong peer ID if you call `LOG_SCOPED_CONTEXT ` in a function which calls another function that initiates syncing with other peers. I also didn't want to spend too much time on these logging changes...I can take a look at using it after I've wrapped up some of the higher priority work on my plate e.g. our peering vs Geth and syncing getting stuck. What do you think?

halfalicious:
cc @gumb0

halfalicious:
I'm working on debugging another issue and I noticed that `BlockChainSync::syncPeer` can be called from both the client thread and the network thread - I think this means that both it and any function it calls (`EthereumPeer::requestBlockHeaders`, `BlockChainSync::requestBlocks`, `EthereumPeer::requestByHashes`) should use the global thread-safe loggers. Note that `BlockChainSync::syncPeer` used instance loggers before my changes e.g.: https://github.com/ethereum/aleth/blob/b094d0ccd58965571be0daad8074857216e1bbf7/libethereum/BlockChainSync.cpp#L227-L245

gumb0:
> I can take a look at using it after I've wrapped up some of the higher priority work on my plate e.g. our peering vs Geth and syncing getting stuck. What do you think?

Yeah it's fine to leave it like this for now.

gumb0:
> I'm working on debugging another issue and I noticed that BlockChainSync::syncPeer can be called from both the client thread and the network thread

Is it because of this single call?
https://github.com/ethereum/aleth/blob/b094d0ccd58965571be0daad8074857216e1bbf7/libethereum/BlockChainSync.cpp#L147-L152

If so, can we use here `m_host.capabilityHost()->postWork(...)` to make the call to `syncPeer` happen from network thread?

Then we'll avoid the mess with loggers, and it might be a good idea for thread-safety in general.

gumb0:
But if it's ever needed, it's also possible to create non-global but thread-safe loggers. For that you'd need another function like this
https://github.com/ethereum/aleth/blob/b094d0ccd58965571be0daad8074857216e1bbf7/libdevcore/Log.h#L124-L129
but with `severity_channel_logger_mt` (multi-threaded)

halfalicious:
> 
> 
> > I'm working on debugging another issue and I noticed that BlockChainSync::syncPeer can be called from both the client thread and the network thread
> 
> Is it because of this single call?
> 
> https://github.com/ethereum/aleth/blob/b094d0ccd58965571be0daad8074857216e1bbf7/libethereum/BlockChainSync.cpp#L147-L152
> 
> If so, can we use here `m_host.capabilityHost()->postWork(...)` to make the call to `syncPeer` happen from network thread?
> 
> Then we'll avoid the mess with loggers, and it might be a good idea for thread-safety in general.


> `ost.capabilityHost()->postWork(...)` to make the call to `syncPeer` happen from network thread?

@gumb0 : Yes, I think that's the only place where `syncPeer `is invoked on a non-network thread (`syncPeer `is called in peer data handlers i.e. `onPeer`* which from what I understand are always called on the network thread, or in `continueSync `and `continueSync `is called in `onPeerBlockHeaders `/ `onPeerBlockBodies `and in the `m_bqRoomAvailable ` handler)

Perhaps I should post the entire `m_bqRoomAvailable ` handler to the network thread since that would also fix the "sync stuck" race condition?

gumb0:
Yeah, good idea to move it all to network thread, though I think mutex will be still needed. I think other calls to `syncPeer` are under mutex, and I see at least `BlockChainSync::status()` is the function that might access from another thread some values modified inside `syncPeer`

halfalicious:
> ink mutex will be still needed. I think other calls to `syncPeer` are under mutex, and I see at least `BlockChainSync::status()` is the function that might access from another thread some values modified inside `syncPeer`

Ah I didn't realize that BlockChainSync::status() might call it from another thread but you're right, it can be called from the Client thread - agreed then, keeping the mutex makes sense.

halfalicious:
> > ink mutex will be still needed. I think other calls to `syncPeer` are under mutex, and I see at least `BlockChainSync::status()` is the function that might access from another thread some values modified inside `syncPeer`
> 
> @gumb0 : Ah I didn't realize that BlockChainSync::status() might call it from another thread but you're right, it can be called from the Client thread - agreed then, keeping the mutex makes sense.

@gumb0 Actually, since my changes don't impact the thread safety of `BlockChainSync ` logging (I don't convert any calls to thread-safe loggers to member loggers in functions that can be called from multiple threads) and posting the `m_bqRoomAvailable ` handler to the network thread would affect the "sync stuck" race condition, I think I'd prefer to make these changes as a part of my "sync stuck" fix.

libethereum/EthereumCapability.cpp
```diff
@@ -838,12 +842,13 @@ bool EthereumCapability::interpretCapabilityPacket(
     }
     catch (Exception const&)
     {
-        cnetlog << "Peer causing an Exception: "
-                << boost::current_exception_diagnostic_information() << " " << _r;
+        LOG(m_loggerError) << "Peer " << _peerID << " causing an exception: "
```
gumb0: I would use ERROR channel only for critical unrecoverable errors, this I think should better be warning at most.

libethereum/EthereumPeer.h
```diff
@@ -35,7 +21,12 @@ class EthereumPeer
     EthereumPeer(std::shared_ptr<p2p::CapabilityHostFace> _host, NodeID const& _peerID,
         u256 const& /*_capabilityVersion*/)
       : m_host(std::move(_host)), m_id(_peerID)
-    {}
+    {
+        std::stringstream remoteInfoStream;
+        remoteInfoStream << m_id;
```
gumb0: I think you could just have used `m_id.hex()` or `m_id.abridged()`

libp2p/Session.cpp
```diff
@@ -22,18 +22,30 @@ Session::Session(Host* _h, unique_ptr<RLPXFrameCoder>&& _io, std::shared_ptr<RLP
...
     m_info(_info),
-    m_ping(chrono::steady_clock::time_point::max()),
-    m_logContext(_info.id.abridged() + "|" + _info.clientVersion)
+    m_ping(chrono::steady_clock::time_point::max())
 {
+    stringstream remoteInfoStream;
+    remoteInfoStream << "(" << m_info.id << "@" << m_socket->remoteEndpoint() << ")";
+
+    m_netLogger.add_attribute(
+        "Suffix", boost::log::attributes::constant<std::string>(remoteInfoStream.str()));
```
gumb0: Maybe `auto const attr = boost::log::attributes::constant<std::string>(remoteInfoStream.str());` could help to have less repetition here

libp2p/Session.cpp
```diff
@@ -115,18 +126,18 @@ bool Session::readPacket(uint16_t _capId, unsigned _packetType, RLP const& _r)
...
+                              << ". PacketType: " << capabilityPacketTypeToString(_packetType)
+                              << " (" << _packetType << "). RLP: " << _r;
         disconnect(BadProtocol);
         return true;
     }
     return true;
 }
 
-bool Session::interpret(P2pPacketType _t, RLP const& _r)
+bool Session::interpretCapabilityPacket(P2pPacketType _t, RLP const& _r)
```
gumb0: Nit about the name: it actually interprets p2p packet, capability packets are interpreted by capability implementations.

libp2p/Session.cpp
```diff
@@ -115,18 +126,18 @@ bool Session::readPacket(uint16_t _capId, unsigned _packetType, RLP const& _r)
     }
     catch (std::exception const& _e)
     {
-        cnetlog << "Exception caught in p2p::Session::interpret(): " << _e.what()
-                << ". PacketType: " << _packetType << ". RLP: " << _r;
+        LOG(m_netLoggerError) << "Exception caught in p2p::Session::readPacket(): " << _e.what()
```
gumb0: Also not sure about ERROR channel here, I'd leave it in DEBUG maybe

libp2p/Session.cpp
```diff
@@ -320,13 +326,11 @@ void Session::doRead()
...
     ba::async_read(m_socket->ref(), boost::asio::buffer(m_data, h256::size),
         [this, self](boost::system::error_code ec, std::size_t length) {
-            LOG_SCOPED_CONTEXT(m_logContext);
-
             if (!checkRead(h256::size, ec, length))
                 return;
             else if (!m_io->authAndDecryptHeader(bytesRef(m_data.data(), length)))
             {
-                cnetlog << "header decrypt failed";
+                LOG(m_netLoggerError) << "Header decrypt failed";
```
gumb0: This I guess can happen if we get corrupted or incorrectly encrypted header, I would leave it at DEBUG level, it's not very interesting on default verbosity

libp2p/Session.cpp
halfalicious: Good idea, will fix

libp2p/Session.cpp
halfalicious: Well technically the spec refers to these packets as p2p capability packets 😄 From https://github.com/ethereum/devp2p/blob/master/rlpx.md: 

> The capabilities supported on either side of the connection are exchanged in the Hello message belonging to the 'p2p' capability which is required to be available on all connections.

I had originally named the function `interpretP2pCapabilityPacket` but decided that was a bit long and went with just `interpretCapabilityPacket `after seeing that the capability class implementations had functions named `interpretCapabilityPacket`.

I had also considered simply `interpretP2pPacket` but thought that p2p was a bit ambiguous.

halfalicious:
Rebased

libp2p/Session.cpp
gumb0: To me it seems confusing that method of capabilities and this one are named the same, because they interpret different classes of packets... In our code what this doc calls `p2p capability` is not a `Capability` but internal implementation of `Session`

libp2p/Session.cpp
```diff
@@ -22,18 +22,26 @@ Session::Session(Host* _h, unique_ptr<RLPXFrameCoder>&& _io, std::shared_ptr<RLP
     m_socket(_s),
     m_peer(_n),
     m_info(_info),
-    m_ping(chrono::steady_clock::time_point::max()),
-    m_logContext(_info.id.abridged() + "|" + _info.clientVersion)
+    m_ping(chrono::steady_clock::time_point::max())
 {
+    stringstream remoteInfoStream;
+    remoteInfoStream << "(" << m_info.id << "@" << m_socket->remoteEndpoint() << ")";
```
gumb0: abridged?

libp2p/Session.cpp
```diff
@@ -292,8 +295,7 @@ void Session::drop(DisconnectReason _reason)
 
 void Session::disconnect(DisconnectReason _reason)
 {
-    cnetdetails << "Disconnecting (our reason: " << reasonOf(_reason) << ") from " << m_info.id
-                << "@" << m_socket->remoteEndpoint();
+    LOG(m_netLoggerDetail) << "Disconnecting (our reason: " << reasonOf(_reason) << ") from";
```
gumb0: Also public method, so we should assume it may be called from any thread

libp2p/Session.cpp
```diff
@@ -355,22 +355,21 @@ void Session::doRead()
...
                         drop(BadProtocol);  // todo: better error
                         return;
                     }
 
                     bytesConstRef frame(m_data.data(), hLength);
                     if (!checkPacket(frame))
                     {
-                        cerr << "Received " << frame.size() << ": " << toHex(frame) << endl;
-                        cnetlog << "INVALID MESSAGE RECEIVED";
+                        LOG(m_netLoggerError) << "Received invalid message. Size: " << frame.size()
```
gumb0: We shouldn't log every invalid packet to ERROR

libp2p/Session.cpp
```diff
@@ -355,22 +355,21 @@ void Session::doRead()
...
                 [this, self, hLength, hProtocolId, tlen](
                     boost::system::error_code ec, std::size_t length) {
-                    LOG_SCOPED_CONTEXT(m_logContext);
 
                     if (!checkRead(tlen, ec, length))
                         return;
                     else if (!m_io->authAndDecryptFrame(bytesRef(m_data.data(), tlen)))
                     {
-                        cnetlog << "frame decrypt failed";
+                        LOG(m_netLoggerError) << "Frame decrypt failed";
```
gumb0: We shouldn't log every invalid packet to ERROR

libp2p/Session.cpp
```diff
@@ -380,7 +379,9 @@ void Session::doRead()
                         RLP r(frame.cropped(1));
                         bool ok = readPacket(hProtocolId, packetType, r);
                         if (!ok)
-                            cnetlog << "Couldn't interpret packet. " << RLP(r);
+                            LOG(m_netLoggerError)
+                                << "Couldn't interpret " << p2pPacketTypeToString(packetType)
```
gumb0: Here, too, I don't think it's for ERROR

libp2p/Session.cpp
```diff
@@ -406,7 +407,9 @@ bool Session::checkRead(std::size_t _expected, boost::system::error_code _ec, st
     {
         // with static m_data-sized buffer this shouldn't happen unless there's a regression
         // sec recommends checking anyways (instead of assert)
-        cnetlog << "Error reading - TCP read buffer length differs from expected frame size.";
+        LOG(m_netLoggerError)
+            << "Error reading - TCP read buffer length differs from expected frame size ("
```
gumb0: Also seems too low-level for ERROR

libp2p/Session.cpp
```diff
@@ -434,8 +437,8 @@ bool Session::canHandle(
 
 void Session::disableCapability(std::string const& _capabilityName, std::string const& _problem)
 {
-    cnetdetails << "DISABLE: Disabling capability '" << _capabilityName
-                << "'. Reason: " << _problem;
+    LOG(m_netLoggerDetail) << "DISABLE: Disabling capability '" << _capabilityName
```
gumb0: This could be useful on DEBUG

libethereum/EthereumPeer.h
```diff
@@ -35,7 +21,9 @@ class EthereumPeer
     EthereumPeer(std::shared_ptr<p2p::CapabilityHostFace> _host, NodeID const& _peerID,
         u256 const& /*_capabilityVersion*/)
       : m_host(std::move(_host)), m_id(_peerID)
-    {}
+    {
+        m_logger.add_attribute("Suffix", boost::log::attributes::constant<std::string>{m_id.hex()});
```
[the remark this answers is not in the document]
halfalicious: @gumb0 Definitely better abridged 😄

libethereum/WarpCapability.cpp
```diff
@@ -455,12 +457,13 @@ bool WarpCapability::interpretCapabilityPacket(NodeID const& _peerID, unsigned _
     }
     catch (Exception const&)
     {
-        cnetlog << "Warp Peer causing an Exception: "
-                << boost::current_exception_diagnostic_information() << " " << _r;
+        LOG(m_loggerError) << "Warp Peer " << _peerID << " causing an exception: "
```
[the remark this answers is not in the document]
halfalicious: Ah nice catch

libp2p/Session.cpp
halfalicious: Ah okay, I had converted all of these calls because I had done some code inspection and thought they were called from a single thread but you bring up a good point - any public function can be called from any thread....I'll change this log statement back.

gumb0: I think it's not the only change like this in `Session` class, please double-check this

libp2p/Session.cpp
halfalicious: @gumb0 The comment above the log statement gives me the impression that this should never happen which is why I chose to log this as error (since if it does happen we want to know about it).

libp2p/Session.cpp
halfalicious: Yes it's abridged: 

`TRACE 05-23 20:30:25 p2p  net    Pong to (##44eed8b3…@35.195.118.171:30303)`

libp2p/RLPxHandshake.cpp
```diff
@@ -33,16 +33,27 @@ RLPXHandshake::RLPXHandshake(
...
 {
     m_logger.add_attribute(
-        "Context", boost::log::attributes::constant<std::string>(connectionDirectionString()));
+        "Prefix", boost::log::attributes::constant<std::string>(connectionDirectionString()));
     m_errorLogger.add_attribute(
-        "Context", boost::log::attributes::constant<std::string>(connectionDirectionString()));
+        "Prefix", boost::log::attributes::constant<std::string>(connectionDirectionString()));
+
+    stringstream remoteInfoStream;
+    remoteInfoStream << "(" << _remote;
```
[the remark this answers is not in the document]
halfalicious: Yes: 

`TRACE 05-23 20:32:03 p2p  rlpx   egress Successfully decrypted frame header, validating contents... (##78de8a09…@191.235.84.50:30303)`

halfalicious:
> 
> 
> Several more comments. Sorry, but this PR is quite huge, I keep discovering new stuff in it.

No worries at all! I know these changes probably aren't very interesting so I really appreciate the in-depth review 😄

libp2p/RLPxHandshake.cpp
```diff
@@ -32,20 +32,22 @@ RLPXHandshake::RLPXHandshake(
     m_socket(_socket),
     m_idleTimer(m_socket->ref().get_io_service())
 {
+    auto const prefixAttr = boost::log::attributes::constant<std::string>{ connectionDirectionString() };
     m_logger.add_attribute(
-        "Prefix", boost::log::attributes::constant<std::string>(connectionDirectionString()));
+        "Prefix", prefixAttr);
```
gumb0: wrapping is odd, did you run clang-format here?

libp2p/Session.cpp
```diff
@@ -198,10 +207,10 @@ bool Session::checkPacket(bytesConstRef _msg)
 void Session::send(bytes&& _msg)
 {
     bytesConstRef msg(&_msg);
-    clog(VerbosityTrace, "net") << "Sending " << capabilityPacketTypeToString(_msg[0]) << " to "
-                                << m_info.id << "@" << m_socket->remoteEndpoint();
+    LOG(m_netLoggerDetail) << capabilityPacketTypeToString(_msg[0]) << " to";
```
gumb0: `send` is called from the public function `sealAndSend`, so I think it's better to leave global logger here

libp2p/RLPxHandshake.cpp
halfalicious: I think so but I can manually reformat this

gumb0: Better leave it like clang-format wishes, but it seems it shouldn't have done this here - the line above is 4 times longer, but only this one is wrapped

halfalicious: I manually reformatted it and ran it through clang-format again and it looks like this now:

```
 auto const prefixAttr =
        boost::log::attributes::constant<std::string>{connectionDirectionString()};
    m_logger.add_attribute("Prefix", prefixAttr);
    m_errorLogger.add_attribute("Prefix", prefixAttr);
```

So I suspect I forgot to run clang-format yesterday.

PR #5608, opened 2019-05-21
[WIP] Fix sync race condition which causes syncing to get stuck

### LINKED ISSUE #5312: Sync gets stuck
I saw this several times on Linux - sync doesn't proceed and no log output after with a last message about the bad hash in the discovery packet.
```
INFO  10-11 15:02:45 eth  client 2 blocks imported in 196 ms (10.1524 blocks/s) in #5340209
INFO  10-11 15:02:48 eth  client 3 blocks imported in 2784 ms (1.07733 blocks/s) in #5340212
INFO  10-11 15:02:49 eth  client 2 blocks imported in 781 ms (2.55817 blocks/s) in #5340214
INFO  10-11 15:02:49 eth  client 1 blocks imported in 375 ms (2.66216 blocks/s) in #5340215
ERROR 10-11 15:07:13 p2p  discov Invalid packet (bad hash) from 78.56.33.225:31313
```

halfalicious:
Fix #5312 

Fix race condition which can occur due to the block queue "room available" handler (`BlockChainSync::m_bqRoomAvailable`) executing on the client thread concurrently with `BlockChainSync::requestBlocks` executing on the network thread which results in syncing getting paused forever. 

The race condition occurs when the block queue gets full and `BlockChainSync::requestBlocks` executes, checks the block queue status, and since it's full pauses the sync:  https://github.com/ethereum/aleth/blob/b094d0ccd58965571be0daad8074857216e1bbf7/libethereum/BlockChainSync.cpp#L290-L297

However, if right after `requestBlocks ` checks the block queue status but before it pauses syncing, the `m_bqRoomAvailable `handler starts executing and changes the sync state to `SyncState::Blocks` and calls `continueSync`, we can get into a state where the sync state will be paused but `BlockChainSync `will attempt to continue syncing (i.e. request data from peers), but reject the incoming data (due to the paused sync state).

The fix for this is to acquire the `BlockChainSync::x_sync` mutex in `requestBlocks` so that the handler and requestBlocks can't access the sync state concurrently.

I've also reviewed the `BlockChainSync ` code and found several other places where I think it makes sense to acquire the mutex. Additionally, I've updated the `m_bqRoomAvailable `handler to post the lambda to the network thread to simplify the threading story of `BlockChainSync `a little bit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Fixed: [#5562](https://github.com/ethereum/aleth/pull/5562) Don't send header request messages to peers that haven't sent us Status yet.
 - Fixed: [#5581](https://github.com/ethereum/aleth/pull/5581) Fixed finding neighbour nodes in Discovery.
 - Fixed: [#5599](https://github.com/ethereum/aleth/pull/5600) Prevent aleth from attempting concurrent connection to node which results in disconnect of original connection.
+- Fixed: [#5608](https://github.com/ethereum/aleth/pull/5608) Fix sync race condition which causes syncing to get stuck.
 
 ## [1.6.0] - 2019-04-16
 
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -144,11 +144,12 @@ BlockChainSync::BlockChainSync(EthereumCapability& _host)
     m_lastImportedBlock(m_startingBlock),
     m_lastImportedBlockHash(_host.chain().currentHash())
 {
-    m_bqRoomAvailable = host().bq().onRoomAvailable([this]()
-    {
-        RecursiveGuard l(x_sync);
-        m_state = SyncState::Blocks;
-        continueSync();
+    m_bqRoomAvailable = host().bq().onRoomAvailable([this]() {
+        host().capabilityHost().postWork([this]() {
+            RecursiveGuard l(x_sync);
+            m_state = SyncState::Blocks;
+            continueSync();
+        });
     });
 }
 
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -154,7 +154,6 @@ BlockChainSync::BlockChainSync(EthereumCapability& _host)
 
 BlockChainSync::~BlockChainSync()
 {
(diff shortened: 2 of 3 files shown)


gumb0:
Actually I'm confused now, because it seems that race condition shouldn't have happened, because `requestBlocks` was already called only with acquired lock (and `onRoomAvailable` also acquired lock)

Possible paths up the stack from `requestBlocks` are:

- requestBlocks 
  - called only by syncPeer 
    - onPeerStatus (has guard)
    - continueSync
      - onRoomAvailable handler (has guard)
      - onPeerBlockHeaders (has guard)
      - onPeerBlockBodies (has guard)
      - onPeerAborting (has guard)
    - onPeerBlockHeaders (has guard)
    - onPeerNewBlock (has guard)
    - onPeerNewHashes (has guard)

Also looking at what calls what like this, all of the new locks you've added seem unnecessary (they are all in private methods, called with lock already acquired)

But if it's helping with the bug, then I don't know what's happening :cat: I think I'll try to run the sync with this branch, too

halfalicious:
> Actually I'm confused now, because it seems that race condition shouldn't have happened, because `requestBlocks` was already called only with acquired lock (and `onRoomAvailable` also acquired lock)
> 
> Possible paths up the stack from `requestBlocks` are:
> 
> * requestBlocks
>   
>   * called only by syncPeer
>     
>     * onPeerStatus (has guard)
>     * continueSync
>       
>       * onRoomAvailable handler (has guard)
>       * onPeerBlockHeaders (has guard)
>       * onPeerBlockBodies (has guard)
>       * onPeerAborting (has guard)
>     * onPeerBlockHeaders (has guard)
>     * onPeerNewBlock (has guard)
>     * onPeerNewHashes (has guard)
> 
> Also looking at what calls what like this, all of the new locks you've added seem unnecessary (they are all in private methods, called with lock already acquired)
> 
> But if it's helping with the bug, then I don't know what's happening 🐱 I think I'll try to run the sync with this branch, too

@gumb0 I haven’t verified that it fixes the bug yet, it sounds like it should have no impact. I will report back once I’ve done a couple of longer syncs.

halfalicious:
No longer needed - addressed by #5865

PR #5612, opened 2019-05-22
fix testeth threating boost `*` as filter folder

### LINKED ISSUE #5611: testeth crashes at the end
When using test filter with `*` the testeth will crash. E.g. 
```
terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
  what():  boost::filesystem::directory_iterator::construct: No such file or directory: "/home/chfast/Projects/ethereum/aleth/test/jsontests/src/GeneralStateTestsFiller/*"
[1]    24270 abort (core dumped)  testeth -t 'GeneralStateTests/*'
```

winsvega:
fixes #5611


winsvega:
in this case it would help. 
let me check how to workaround the other cases

PR #5613, opened 2019-05-27
IPv4 and IPv6 addresses are stored in separate fields in ENR

gumb0:
https://github.com/ethereum/EIPs/commit/9b1c22d2850cc8efb7aabb2aba4098873ed0d5a9


gumb0:
Not sure about failing test on Windows - it seems hard to reproduce it locally. It was reproduced only sometimes, and then stopped failing completely :confused:

PR #5616, opened 2019-05-29
Upgrade Boost to 1.70

### LINKED ISSUE #5614: Upgrade Boost to 1.70+
After Hunter is upgraded in #5506, I propose to use the latest Boost version. I noticed there are some API changes to Boost.ASIO. Someone will have to replace deprecated or remove API with a new one.

gumb0:
Fix #5614

Review commits separately.


gumb0:
Windows build runs out of time, I'll check it on a local machine.

libp2p/Host.cpp
```diff
@@ -508,21 +508,18 @@ void Host::runAcceptor()
         cnetdetails << "Listening on local port " << m_listenPort;
         m_accepting = true;
 
-        auto socket = make_shared<RLPXSocket>(m_ioContext);
-        m_tcp4Acceptor.async_accept(socket->ref(), [=](boost::system::error_code ec)
-        {
+        m_tcp4Acceptor.async_accept([=](boost::system::error_code _ec, bi::tcp::socket _socket) {
```
chfast: Does it pass it by value?

chfast: What other objects are captured? Can you be more specific than `=`?

gumb0: The documentation specifies the handler as `void handler(const boost::system::error_code& error, 
  typename Protocol::socket peer);`
https://www.boost.org/doc/libs/1_70_0/doc/html/boost_asio/reference/basic_socket_acceptor/async_accept/overload3.html

Looking in the debugger, it looks like it passes there `static_cast<socket&&>(peer)`

gumb0: Capturing `this` is enough, I'll change it

chfast: Ok, this looks like perfect forwarding so passing by reference will work as well. But I believe socket only has move constructor so we are fine here.

PR #5618, opened 2019-06-01
NoReward seal engine

winsvega:
working on generating the state tests via RPC
state tests does not have mining reward
there for introducing sealEngine: NoReward  
same as NoProof but wont calculate any mining rewards whatsoever

### DIFF
--- a/libethcore/SealEngine.h
+++ b/libethcore/SealEngine.h
@@ -161,11 +161,7 @@ class NoReward : public NoProof
 public:
     static std::string name() { return "NoReward"; }
     static void init();
-    u256 blockReward(u256 const& _blockNumber) const override
-    {
-        (void)_blockNumber;
-        return 0;
-    }
+    u256 blockReward(u256 const&) const override { return 0; }
 };
 }
 }
--- a/libethcore/SealEngine.h
+++ b/libethcore/SealEngine.h
@@ -154,5 +154,18 @@ u256 calculateEthashDifficulty(
 
 u256 calculateGasLimit(ChainOperationParams const& _chainParams, BlockHeader const& _bi,
     u256 const& _gasFloorTarget = Invalid256);
+
+
+class NoReward : public NoProof
+{
+public:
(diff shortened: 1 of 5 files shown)


gumb0:
Why can't you just set block reward to 0 in config?

winsvega:
Because of blockreward override at Byzantium makes config obsolete

libethcore/SealEngine.h
```diff
@@ -154,5 +154,18 @@ u256 calculateEthashDifficulty(
...
+
+
+class NoReward : public NoProof
+{
+public:
+    static std::string name() { return "NoReward"; }
+    static void init();
+    u256 blockReward(u256 const& _blockNumber) const override
+    {
+        (void)_blockNumber;
```
halfalicious: (Nit) Can you remove this if you don't include the arg name in the arg list?

winsvega: Then it wont override

gumb0: Of course this shouldn't affect overriding.
What Nils meant is
```
 u256 blockReward(u256 const&) const override
 {
    return 0;
 }
```

gumb0:
I suggest the name `NoProofZeroReward`

gumb0:
For reference: looks like this change was proposed by me a year ago https://github.com/ethereum/aleth/pull/4973#discussion_r184041880

chfast:
Ideally, I would fix the reward override mechanism in a way that would allow `NoProof` to be configured in a way that keeps reward 0 also after Byzantium.

But depending on timelines, I'm ok with this way. Although, going with `NoProofZeroReward` mean there is no way back as the `NoProofZeroReward` name will leak in some tests.

gumb0:
Apparently `NoReward` is already in some tests

winsvega:
NoReward is shorter name. Anyway it's meaning must be determined in the specs

gumb0:
@winsvega ready to merge this?

winsvega:
yes

PR #5621, opened 2019-06-05
 Implement EVMC Host using evmc::Host interface 

chfast:


libevm/ExtVMFace.h
```diff
@@ -193,23 +192,21 @@ struct CreateResult
...
-        : status{status}, output{std::move(output)}, address{address}
+      : status{status}, output{std::move(output)}, address{address}
     {}
 };
 
 /**
  * @brief Interface and null implementation of the class for specifying VM externalities.
  */
-class ExtVMFace: public evmc_context
+class ExtVMFace : public evmc::Host
```
gumb0: Well this was already a weird class, mixing pure virtual methods with some implementations and data members, and now it's even weirder having two sets of similar methods.

Would it be possible to split this into two, like maybe having `evmc::Host` implemented separately in terms of `ExtVMFace` (passing it to constructor), and then instantiating it inside `EVMC::exec` ?

libevm/ExtVMFace.cpp
```diff
@@ -214,43 +186,39 @@ evmc_result create(ExtVMFace& _env, evmc_message const* _msg) noexcept
...
 {
-    assert(_msg->gas >= 0 && "Invalid gas value");
-    auto& env = static_cast<ExtVMFace&>(*_context);
+    assert(_msg.gas >= 0 && "Invalid gas value");
 
     // Handle CREATE separately.
-    if (_msg->kind == EVMC_CREATE || _msg->kind == EVMC_CREATE2)
-        return create(env, _msg);
+    if (_msg.kind == EVMC_CREATE || _msg.kind == EVMC_CREATE2)
+        return eth::create(*this, _msg);
```
gumb0: Can we make `create` private member of the class instead of passing `*this`?

libevm/ExtVMFace.h
chfast: Probably can be done, but it would be better if the ExtVMFace for legacy VM is implemented as the wrapper for evmc::Host and evmc::Host implementation replaces ExtVM.

gumb0: I can try to do that

chfast: I can leave it to you. I'm currently busy with adding python bindings to ethash library. This PR is not really required, but reuses some code added to EVMC C++.

gumb0:
I tried first to go the way of replacing `ExtVMFace` with `evmc::Host` in `VMFace`, but it turned out to be a bit complicated. For example there's `bytes const& 
ExtVMFace::codeAt(Address)`, while `evmc::Host` has `size_t copy_code(const evmc_address&, size_t, uint8_t*, size_t)`
Implementing the first one in terms of the second one would be weird and not efficient, it's better to change `LegacyVM` to use `copy_code`-like interface, but then with all such changes `LegacyVM` will end up just not much different from `aleth-interpreter`, so what's the point then.

So I've made the change as I suggested, splitting out `evmc::Host` impementation into separate class, and it looks quite simple. Please review @chfast

libevm/ExtVMFace.h
```diff
@@ -271,13 +273,53 @@ class ExtVMFace: public evmc_context
...
     h256 codeHash;            ///< SHA3 hash of the executing code
-    u256 salt;                ///< Values used in new address construction by CREATE2 
+    u256 salt;                ///< Values used in new address construction by CREATE2
     SubState sub;             ///< Sub-band VM state (suicides, refund counter, logs).
     unsigned depth = 0;       ///< Depth of the present call.
     bool isCreate = false;    ///< Is this a CREATE call?
     bool staticCall = false;  ///< Throw on state changing.
 };
 
+class EvmCHost : public evmc::Host
```
gumb0: I would better call it `EvmcHost`, but I mimicked the functions `toEvmC` / `fromEvmC` below

chfast: This follows e.g. Java naming convention and it is fine. Following e.g. Go naming convention, this could be `EVMCHost`.

PR #5624, opened 2019-06-12
Garbage collect incompatible peers in Host::run()

halfalicious:
Garbage collect peers in `Host::run()` which are deemed incompatible - an incompatible peer is one with which we will never be able to successfully peer with (e.g. on a different chain or network, not running any capabilities).

### DIFF
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -200,23 +200,7 @@ std::shared_ptr<Peer> Host::peer(NodeID const& _n) const
 {
     RecursiveGuard l(x_sessions);
     auto it = m_peers.find(_n);
-    if (it == m_peers.end())
-    {
-        LOG(m_logger) << "Peer " << _n << " not found";
-        return nullptr;
-    }
-    return it->second;
-}
-
-void Host::handshakeFailed(NodeID const& _n, HandshakeFailureReason _r)
-{
-    std::shared_ptr<Peer> p = peer(_n);
-    if (!p)
-    {
-        cerror << "Peer " << _n << " not found";
-        return;
-    }
-    p->m_lastHandshakeFailure = _r;
+    return it != m_peers.end() ? it->second : nullptr;
 }
 
 void Host::doneWorking()
@@ -294,7 +278,10 @@ void Host::startPeerSession(Public const& _id, RLP const& _hello,
     {
         auto itPeer = m_peers.find(_id);
         if (itPeer != m_peers.end())
+        {
             peer = itPeer->second;
+            peer->m_lastHandshakeFailure = NoFailure;
+        }
         else
         {
             // peer doesn't exist, try to get port info from node table
@@ -307,7 +294,6 @@ void Host::startPeerSession(Public const& _id, RLP const& _hello,
             m_peers[_id] = peer;
         }
     }
-    peer->m_lastHandshakeFailure = NoFailure;
     if (peer->isOffline())
         peer->m_lastConnected = chrono::system_clock::now();
     peer->endpoint.setAddress(_s->remoteEndpoint().address());
@@ -408,6 +394,13 @@ void Host::startPeerSession(Public const& _id, RLP const& _hello,
 
     LOG(m_logger) << "Peer connection successfully established with " << _id << "@"
                   << _s->remoteEndpoint();
+}
+
+void Host::onHandshakeFailed(NodeID const& _n, HandshakeFailureReason _r)
+{
+    std::shared_ptr<Peer> p = peer(_n);
+    if (p)
+        p->m_lastHandshakeFailure = _r;
 }
 
 void Host::onNodeTableEvent(NodeID const& _n, NodeTableEventType const& _e)
@@ -807,21 +800,26 @@ void Host::run(boost::system::error_code const& _ec)
     unsigned reqConn = 0;
     {
         RecursiveGuard l(x_sessions);
+        auto p = m_peers.cbegin();
+        while (p != m_peers.cend())
         {
-            for (auto p = m_peers.cbegin(); p != m_peers.cend(); p++)
+            bool peerRemoved = false;
+            bool haveSession = havePeerSession(p->second->id);
+            bool required = p->second->peerType == PeerType::Required;
+            if (haveSession && required)
+                reqConn++;
+            else if (!haveSession)
             {
-                bool haveSession = havePeerSession(p->second->id);
-                bool required = p->second->peerType == PeerType::Required;
-                if (haveSession && required)
-                    reqConn++;
-                else if (!haveSession)
+                if (p->second->isUseless())
                 {
-                    if (p->second->uselessPeer())
-                        p = m_peers.erase(p);
-                    else if (p->second->shouldReconnect() && (!m_netConfig.pin || required))
-                        toConnect.push_back(p->second);
+                    peerRemoved = true;
+                    p = m_peers.erase(p);
                 }
+                else if (p->second->shouldReconnect() && (!m_netConfig.pin || required))
+                    toConnect.push_back(p->second);
             }
+            if (!peerRemoved)
+                p++;
         }
     }
 
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -211,7 +211,11 @@ std::shared_ptr<Peer> Host::peer(NodeID const& _n) const
 void Host::handshakeFailed(NodeID const& _n, HandshakeFailureReason _r)
 {
     std::shared_ptr<Peer> p = peer(_n);
-    assert(p);
+    if (!p)
+    {
+        cerror << "Peer " << _n << " not found";
+        return;
+    }
     p->m_lastHandshakeFailure = _r;
 }
 
@@ -812,7 +816,7 @@ void Host::run(boost::system::error_code const& _ec)
                     reqConn++;
                 else if (!haveSession)
                 {
-                    if (p->second->fallbackSeconds() == numeric_limits<unsigned>::max())
+                    if (p->second->uselessPeer())
                         p = m_peers.erase(p);
                     else if (p->second->shouldReconnect() && (!m_netConfig.pin || required))
                         toConnect.push_back(p->second);
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -196,6 +196,25 @@ void Host::stopCapabilities()
     }
 }
 
+std::shared_ptr<Peer> Host::peer(NodeID const& _n) const
+{
+    RecursiveGuard l(x_sessions);
+    auto it = m_peers.find(_n);
+    if (it == m_peers.end())
+    {
+        LOG(m_logger) << "Peer " << _n << " not found";
+        return nullptr;
+    }
+    return it->second;
+}
+
+void Host::handshakeFailed(NodeID const& _n, HandshakeFailureReason _r)
+{
+    std::shared_ptr<Peer> p = peer(_n);
+    assert(p);
+    p->m_lastHandshakeFailure = _r;
+}
+
 void Host::doneWorking()
 {
     // Return early if we have no capabilities since there's nothing to do. We've already stopped
@@ -284,6 +303,7 @@ void Host::startPeerSession(Public const& _id, RLP const& _hello,
             m_peers[_id] = peer;
         }
     }
+    peer->m_lastHandshakeFailure = NoFailure;
     if (peer->isOffline())
         peer->m_lastConnected = chrono::system_clock::now();
     peer->endpoint.setAddress(_s->remoteEndpoint().address());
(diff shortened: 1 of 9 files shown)


libp2p/Host.cpp
```diff
@@ -783,15 +803,21 @@ void Host::run(boost::system::error_code const& _ec)
...
-                toConnect.push_back(p.second);
+            for (auto p = m_peers.cbegin(); p != m_peers.cend(); p++)
+            {
+                bool haveSession = havePeerSession(p->second->id);
+                bool required = p->second->peerType == PeerType::Required;
+                if (haveSession && required)
+                    reqConn++;
+                else if (!haveSession)
+                {
+                    if (p->second->fallbackSeconds() == numeric_limits<unsigned>::max())
```
gumb0: Instead of using magic value of `fallbackSeconds()`, better add another method to `Peer` returning true if it had critical connection error

libp2p/Peer.cpp
```diff
@@ -45,18 +45,33 @@ Peer::Peer(Peer const& _original):
...
     {
     case BadProtocol:
-        return 30 * (m_failedAttempts + 1);
     case UselessPeer:
+    case IncompatibleProtocol:
+    case UnexpectedIdentity:
+    case UserReason:
+        return numeric_limits<unsigned>::max();
     case TooManyPeers:
         return 25 * (m_failedAttempts + 1);
```
gumb0: `m_failedAttempts` seems to affect only the value of fallbackSeconds currently. Maybe we should use it now to retry several times and then go to "critical error, disconnect" state.
(at least for some cases of failures)

libp2p/Host.cpp
```diff
@@ -196,6 +196,25 @@ void Host::stopCapabilities()
...
+        LOG(m_logger) << "Peer " << _n << " not found";
+        return nullptr;
+    }
+    return it->second;
+}
+
+void Host::handshakeFailed(NodeID const& _n, HandshakeFailureReason _r)
+{
+    std::shared_ptr<Peer> p = peer(_n);
+    assert(p);
```
gumb0: This I think will fail when it's the first handshake attempt that failed and `m_peers` don't have the peer yet (for incoming connections it's added only after successful handshake)
No need to set it to the peer in that case.

halfalicious: > 
> 
> This I think will fail when it's the first handshake attempt that failed and `m_peers` don't have the peer yet (for incoming connections it's added only after successful handshake)
> No need to set it to the peer in that case.

I don't think this case is possible - for outgoing connections we always decide which node to connect to using m_peers and the following check in RLPXHandshake protects against the failed incoming connection case: https://github.com/ethereum/aleth/blob/f1f5db331188cafe11db27f03b907b1475639742/libp2p/RLPxHandshake.cpp#L262-L263

libp2p/Host.cpp
halfalicious: @gumb0 : Good idea - I went with the magic number approach since I figured I'd have to return a large value for these cases in fallbackSeconds which would effectively prevent us from ever reconnecting, so I figured I didn't need the function. However, you're right that the code is more clear and magic numbers aren't good 😄 I'll add the function and also return value in fallbackSeconds which is large enough to prevent us from ever reconnecting but is small enough to prevent an overflow (I'm thinking 1 year in seconds)

libp2p/Host.cpp
halfalicious: @gumb0 : Actually, the EIP8 tests hit this so I've replaced the assert with a nullptr check and log an error now (with VerbosityError since we never expect to hit this when running Aleth, just during tests, so if we see the error message there's a bug and we should investigate)

libp2p/Peer.cpp
halfalicious: @gumb0 Good idea - what do you think of this being taken care of in another PR? I'd like to limit the amount of changes I make to the peer gc logic in this PR so if something ends up breaking it will be easier to debug.

libp2p/RLPxHandshake.cpp
```diff
@@ -112,10 +112,12 @@ void RLPXHandshake::writeAckEIP8()
     m_ackCipher.insert(m_ackCipher.begin(), prefix.begin(), prefix.end());
     
     auto self(shared_from_this());
-    ba::async_write(m_socket->ref(), ba::buffer(m_ackCipher), [this, self](boost::system::error_code ec, std::size_t)
-    {
-        transition(ec);
-    });
+    ba::async_write(m_socket->ref(), ba::buffer(m_ackCipher),
+        [this, self](boost::system::error_code ec, std::size_t) {
+            if (ec)
```
gumb0: It looks like we can put `if (ec) m_failureReason = TcpError;` inside `transition(ec)` and avoid some repetition.

libp2p/RLPxHandshake.cpp
```diff
@@ -246,6 +262,9 @@ void RLPXHandshake::cancel()
 
 void RLPXHandshake::error(boost::system::error_code _ech)
 {
+    if (m_originated)
```
gumb0: I think it's better to call this callback in any case (regardless of `m_originated`) and let Host decide whether it needs it.

(Because here it's not immediately clear why this condition exists)

libp2p/Host.h
```diff
@@ -343,6 +343,11 @@ class Host: public Worker
     /// Stop registered capabilities, typically done when the network is being shut down.
     void stopCapabilities();
 
+    std::shared_ptr<Peer> peer(NodeID const& _n) const;
+
+    /// Set a handshake failure reason for a peer
+    void handshakeFailed(NodeID const& _n, HandshakeFailureReason _r);
```
gumb0: Maybe better `onHandshakeFailed`

gumb0: Also I would make it public and declare close to `startPeerSession`

libp2p/Host.cpp
```diff
@@ -196,6 +196,29 @@ void Host::stopCapabilities()
     }
 }
 
+std::shared_ptr<Peer> Host::peer(NodeID const& _n) const
+{
+    RecursiveGuard l(x_sessions);
+    auto it = m_peers.find(_n);
+    if (it == m_peers.end())
+    {
+        LOG(m_logger) << "Peer " << _n << " not found";
```
gumb0: I think better not to put logging in the simple getter like this (better to log it on the calling side if needed, and you already do in `handshakeFailed`)

libp2p/Peer.cpp
```diff
@@ -45,18 +45,64 @@ Peer::Peer(Peer const& _original):
 
 bool Peer::shouldReconnect() const
 {
-    return id && endpoint && chrono::system_clock::now() > m_lastAttempted + chrono::seconds(fallbackSeconds());
+    return id && endpoint &&
+           fallbackSeconds() != numeric_limits<unsigned>::max() &&
```
gumb0: this is never the case now

gumb0: I guess you can replace with `&& !uselessPeer()`

libp2p/Peer.h
```diff
@@ -80,7 +84,8 @@ class Peer: public Node
     void noteSessionGood() { m_failedAttempts = 0; }
 
 private:
-    /// Returns number of seconds to wait until attempting connection, based on attempted connection history.
+    /// Returns number of seconds to wait until attempting connection, based on attempted connection history, or
+    /// numeric_limits<unsigned>::max() if a connection should never be attempted.
```
gumb0: comment not updated

libp2p/Peer.cpp
```diff
@@ -45,18 +45,64 @@ Peer::Peer(Peer const& _original):
...
+        break;
+    }
+
+    switch (m_lastDisconnect)
+    {
+    case BadProtocol:
+    case UselessPeer:
+    case IncompatibleProtocol:
+    case UnexpectedIdentity:
+    case UserReason:
```
gumb0: Not sure about `UserReason` - this could in theory be any kind of reason specific to subprotocol, including some temporary reasons, maybe it makes sense to try to reconnect

libp2p/RLPxHandshake.cpp
gumb0: And I think you already ignore it if peer is not already known to Host, so we can just remove `if` here

libp2p/Host.cpp
```diff
@@ -783,15 +807,21 @@ void Host::run(boost::system::error_code const& _ec)
     unsigned reqConn = 0;
     {
         RecursiveGuard l(x_sessions);
-        for (auto const& p : m_peers)
         {
```
gumb0: these braces are redundant

libp2p/Host.cpp
```diff
@@ -783,15 +807,21 @@ void Host::run(boost::system::error_code const& _ec)
...
-            bool haveSession = havePeerSession(p.second->id);
-            bool required = p.second->peerType == PeerType::Required;
-            if (haveSession && required)
-                reqConn++;
-            else if (!haveSession && p.second->shouldReconnect() &&
-                        (!m_netConfig.pin || required))
-                toConnect.push_back(p.second);
+            for (auto p = m_peers.cbegin(); p != m_peers.cend(); p++)
```
gumb0: You should do `++p` only unless you erased in this iteration. Now it'll skip the next one after erased one.

libp2p/Host.cpp
```diff
@@ -284,6 +307,7 @@ void Host::startPeerSession(Public const& _id, RLP const& _hello,
             m_peers[_id] = peer;
         }
     }
+    peer->m_lastHandshakeFailure = NoFailure;
```
gumb0: It's `NoFailure` when `Peer` is created, so better move this to `if (itPeer != m_peers.end())` case.

libp2p/Peer.cpp
gumb0: Ok for another PR, but it seems to be the matter of only adding "m_failedAttempts >= 20" condition to `uselessPeer` function.

gumb0: Additional thought is that we could make all this change more conservative if we leave `fallbackSeconds()` as it was before (so that we do the same reconnects as before) and have just this check for failed attempts count in `uselessPeer` (plus the new check for handshake failures)
This way it would reconnect with the same intervals for each case as before, but stop after limited number of attempts.

But I'm fine with it if you think it's better to immediately stop in some cases.

libp2p/Peer.h
```diff
@@ -70,6 +70,10 @@ class Peer: public Node
     /// Return true if connection attempt should be made to this peer or false if
     bool shouldReconnect() const;
 
+    /// A peer which should never be reconnected to - e.g. it's running on a different network, we
+    /// don't have any common capabilities
+    bool uselessPeer() const;
```
gumb0: Also maybe `isUselessPeer` is a better name or just `isUseless`

libp2p/Host.cpp
halfalicious: Ah good point

libp2p/Host.cpp
halfalicious: Good catch

libp2p/Host.cpp
halfalicious: Good point

libp2p/Host.h
halfalicious: Why make this public, does it make sense to expose the concept of a handshake to consumers of Host?

libp2p/RLPxHandshake.cpp
halfalicious: Good point 😄

libp2p/Peer.cpp
halfalicious: I chose to treat UserReason as terminal because we only use it in 2 cases - if status validation fails: https://github.com/ethereum/aleth/blob/505aeada27f396dd5cb11e69d5181ca84a914186/libethereum/BlockChainSync.cpp#L190-L202
Here's where we perform the actual validation in EthereumPeer::validate: https://github.com/ethereum/aleth/blob/505aeada27f396dd5cb11e69d5181ca84a914186/libethereum/EthereumPeer.cpp#L58-L68

* If there's a bug in our network code and Session read validation fails: https://github.com/ethereum/aleth/blob/505aeada27f396dd5cb11e69d5181ca84a914186/libp2p/Session.cpp#L409-L418

You bring up a good point though, technically it can be any reason specific to a subprotocol and other clients can also choose to send it to us for reasons specific to their implementation. I think that rather than treating it as immediately critical we should use it in some sort of reconnection count threshold.

libp2p/Peer.cpp
halfalicious: @gumb0 My hesitancy is that I don't know of a good value to use for the failed connection count threshold 😄 I'll ping the Geth and Parity folks to see if they have similar logic and if so, what their values are.

libp2p/Peer.cpp
halfalicious: @gumb0 I've decided to change `UserReason `-> `UselessPeer `when status message validation fails for a peer (since if that happens the peer is effectively useless to us i.e. we can't sync with it). I've also added in logic in `Peer::isUseless` to take the number of failed connection attempts into account when the last disconnect isn't critical.

libp2p/Host.cpp
halfalicious: Thanks, got confused due to `DEV_RECURSIVE_GUARDED `use in some places

libp2p/Host.h
gumb0: Well it looks to me like a callback similar to `startPeerSession`, the callback called by `RLPXHandshake` when handshake is finished. One is for success another one is for failure.

It works as private, because `RLPXHandshake` is a friend of `Host`, but idealluy should get rid of this friend declarations at some point.

Exposing it to the clients of `Host` is of course not great, but the proper way to deal with it could be to create a separate interface with these callbacks only, don't expose it to `Host` clients, but pass it only to `RLPXHandshake`. That's a bit complicated, at least it's not for this PR.

(In other words, we won't make it much worse, because `startPeerSession` is already public)

halfalicious: > 
> 
> Well it looks to me like a callback similar to `startPeerSession`, the callback called by `RLPXHandshake` when handshake is finished. One is for success another one is for failure.
> 
> It works as private, because `RLPXHandshake` is a friend of `Host`, but idealluy we should get rid of this friend declarations at some point.
> 
> Exposing it to the clients of `Host` is of course not great, but the proper way to deal with it could be to create a separate interface with these callbacks only, don't expose it to `Host` clients, but pass it only to `RLPXHandshake`. That's a bit complicated change, at least it's not for this PR.
> 
> (In other words, we won't make it much worse, because `startPeerSession` is already public)

Ah that makes sense, thank you for clarifying! 😄 I'll make the change before merging.

PR #5628, opened 2019-06-14
Don't add official Ethereum bootnodes as required peers

halfalicious:
Add them to the node table rather than as required peers - this results in us interacting with them via discovery (which is where their real value lies) and if we can validate the endpoint proof, adding them as optional peers and trying to connect to them via devp2p. However, we stop trying to connect to them if we get disconnected for a critical reason (e.g. they're on a different network or we don't have any common capabilities). If we get disconnected for a non-critical reason (e.g. a TCP error), we'll try to reconnect at increasing intervals. 

This makes more sense than adding them as required peers which results in us endlessly trying to connect to them via devp2p, regardless of the disconnect reason - note that many of them are on different Ethereum networks and also don't appear to be running devp2p so they don't support syncing, so we never end up being able to connect to them anyway. In practice, I believe we're only ever able to peer with 1 node out of our full list.


halfalicious:
We should wait until #5624 is merged to merge this (not strictly required but this work makes more sense after 5624 is in)

halfalicious:
Still need to test these changes and investigate the failing tests

chfast:
This is good change, because the predefined bootnodes are to be only used to bootstrap the p2p. Then we should rather depend on other peers.

PR #5640, opened 2019-06-26
Implement EIP-1702 Generalized Account Versioning Scheme

gumb0:
Closes https://github.com/ethereum/aleth/issues/5638
https://eips.ethereum.org/EIPS/eip-1702

### DIFF
--- a/aleth-vm/main.cpp
+++ b/aleth-vm/main.cpp
@@ -269,7 +269,7 @@ int main(int argc, char** argv)
     {
         // Deploy the code on some fake account to be called later.
         Account account(0, 0);
-        auto const latestVersion = se->evmSchedule(envInfo.number()).version;
+        auto const latestVersion = se->evmSchedule(envInfo.number()).accountVersion;
         account.setCode(bytes{code}, latestVersion);
         std::unordered_map<Address, Account> map;
         map[contractDestination] = account;
--- a/aleth-vm/main.cpp
+++ b/aleth-vm/main.cpp
@@ -259,13 +259,18 @@ int main(int argc, char** argv)
         }  // Ignore decoding errors.
     }
 
+    unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(networkName)).createSealEngine());
+    LastBlockHashes lastBlockHashes;
+    EnvInfo const envInfo(blockHeader, lastBlockHashes, 0);
+
     Transaction t;
     Address contractDestination("1122334455667788991011121314151617181920");
     if (!code.empty())
     {
         // Deploy the code on some fake account to be called later.
         Account account(0, 0);
-        account.setCode(bytes{code});
+        auto const latestVersion = se->evmSchedule(envInfo.number()).version;
+        account.setCode(bytes{code}, latestVersion);
         std::unordered_map<Address, Account> map;
         map[contractDestination] = account;
         state.populateFrom(map);
@@ -278,9 +283,6 @@ int main(int argc, char** argv)
 
     state.addBalance(sender, value);
 
-    unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(networkName)).createSealEngine());
-    LastBlockHashes lastBlockHashes;
-    EnvInfo const envInfo(blockHeader, lastBlockHashes, 0);
     Executive executive(state, envInfo, *se);
     ExecutionResult res;
     executive.setResultRecipient(res);
--- a/libethcore/EVMSchedule.h
+++ b/libethcore/EVMSchedule.h
@@ -29,7 +29,7 @@ struct EVMSchedule
 {
     EVMSchedule(): tierStepGas(std::array<unsigned, 8>{{0, 2, 3, 5, 8, 10, 20, 0}}) {}
     EVMSchedule(bool _efcd, bool _hdc, unsigned const& _txCreateGas): exceptionalFailedCodeDeposit(_efcd), haveDelegateCall(_hdc), tierStepGas(std::array<unsigned, 8>{{0, 2, 3, 5, 8, 10, 20, 0}}), txCreateGas(_txCreateGas) {}
-    unsigned version = 0;
+    unsigned accountVersion = 0;
     bool exceptionalFailedCodeDeposit = true;
     bool haveDelegateCall = true;
     bool eip150Mode = false;
@@ -149,7 +149,7 @@ static const EVMSchedule ConstantinopleFixSchedule = [] {
 
 static const EVMSchedule IstanbulSchedule = [] {
     EVMSchedule schedule = ConstantinopleFixSchedule;
-    schedule.version = 1;
+    schedule.accountVersion = 1;
     return schedule;
 }();
 
@@ -163,10 +163,15 @@ inline EVMSchedule const& evmScheduleForAccountVersion(u256 const& _version)
 {
     if (_version == 0)
         return ConstantinopleFixSchedule;
-    else if (_version == IstanbulSchedule.version)
+    else if (_version == IstanbulSchedule.accountVersion)
         return IstanbulSchedule;
     else
+    {
+        // This should not happen, as all existing accounts
+        // are created either with version 0 or with one of fork's versions
+        assert(false);
         return DefaultSchedule;
+    }
 }
 }
 }
--- a/libethcore/EVMSchedule.h
+++ b/libethcore/EVMSchedule.h
@@ -29,6 +29,7 @@ struct EVMSchedule
 {
     EVMSchedule(): tierStepGas(std::array<unsigned, 8>{{0, 2, 3, 5, 8, 10, 20, 0}}) {}
     EVMSchedule(bool _efcd, bool _hdc, unsigned const& _txCreateGas): exceptionalFailedCodeDeposit(_efcd), haveDelegateCall(_hdc), tierStepGas(std::array<unsigned, 8>{{0, 2, 3, 5, 8, 10, 20, 0}}), txCreateGas(_txCreateGas) {}
+    int version = 0;
     bool exceptionalFailedCodeDeposit = true;
     bool haveDelegateCall = true;
     bool eip150Mode = false;
@@ -146,10 +147,24 @@ static const EVMSchedule ConstantinopleFixSchedule = [] {
     return schedule;
 }();
 
+static const EVMSchedule IstanbulSchedule = [] {
+    EVMSchedule schedule = ConstantinopleFixSchedule;
+    schedule.version = 1;
+    return schedule;
+}();
+
 static const EVMSchedule ExperimentalSchedule = [] {
-    EVMSchedule schedule = ConstantinopleSchedule;
+    EVMSchedule schedule = IstanbulSchedule;
     schedule.blockhashGas = 800;
     return schedule;
 }();
+
+inline EVMSchedule const& evmScheduleForAccountVersion(u256 const& _version)
+{
+    if (_version == IstanbulSchedule.version)
+        return IstanbulSchedule;
+    else
+        return DefaultSchedule;
+}
 }
 }
--- a/libethcore/EVMSchedule.h
+++ b/libethcore/EVMSchedule.h
@@ -29,7 +29,7 @@ struct EVMSchedule
 {
     EVMSchedule(): tierStepGas(std::array<unsigned, 8>{{0, 2, 3, 5, 8, 10, 20, 0}}) {}
     EVMSchedule(bool _efcd, bool _hdc, unsigned const& _txCreateGas): exceptionalFailedCodeDeposit(_efcd), haveDelegateCall(_hdc), tierStepGas(std::array<unsigned, 8>{{0, 2, 3, 5, 8, 10, 20, 0}}), txCreateGas(_txCreateGas) {}
-    int version = 0;
+    unsigned version = 0;
     bool exceptionalFailedCodeDeposit = true;
     bool haveDelegateCall = true;
     bool eip150Mode = false;
--- a/libethcore/EVMSchedule.h
+++ b/libethcore/EVMSchedule.h
@@ -161,7 +161,9 @@ static const EVMSchedule ExperimentalSchedule = [] {
 
 inline EVMSchedule const& evmScheduleForAccountVersion(u256 const& _version)
 {
-    if (_version == IstanbulSchedule.version)
+    if (_version == 0)
+        return ConstantinopleFixSchedule;
+    else if (_version == IstanbulSchedule.version)
         return IstanbulSchedule;
     else
         return DefaultSchedule;
--- a/libethcore/EVMSchedule.h
+++ b/libethcore/EVMSchedule.h
@@ -159,7 +159,7 @@ static const EVMSchedule ExperimentalSchedule = [] {
     return schedule;
 }();
 
-inline EVMSchedule const& evmScheduleForAccountVersion(u256 const& _version)
+inline EVMSchedule const& latestScheduleForAccountVersion(u256 const& _version)
 {
     if (_version == 0)
         return ConstantinopleFixSchedule;
--- a/libethereum/Account.cpp
+++ b/libethereum/Account.cpp
@@ -34,7 +34,7 @@ using namespace dev::eth::validation;
 
 namespace fs = boost::filesystem;
 
-void Account::setCode(bytes&& _code)
+void Account::setCode(bytes&& _code, u256 const& _version)
 {
     auto const newHash = sha3(_code);
     if (newHash != m_codeHash)
@@ -43,6 +43,7 @@ void Account::setCode(bytes&& _code)
         m_hasNewCode = true;
         m_codeHash = newHash;
     }
+    m_version = _version;
 }
 
 void Account::resetCode()
@@ -166,7 +167,7 @@ AccountMap dev::eth::jsonToAccountMap(std::string const& _json, u256 const& _def
                         cerr << "Error importing code of account " << a
                              << "! Code needs to be hex bytecode prefixed by \"0x\".";
                     else
-                        ret[a].setCode(fromHex(codeStr));
+                        ret[a].setCode(fromHex(codeStr), 0);
                 }
                 else
                     cerr << "Error importing code of account " << a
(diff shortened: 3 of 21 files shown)


libethereum/State.h
```diff
@@ -136,8 +136,8 @@ struct Change
     {}
 
     /// Helper constructor especially for new code change log.
-    Change(Address const& _addr, bytes const& _oldCode):
-            kind(Code), address(_addr), oldCode(_oldCode)
+    Change(Address const& _addr, bytes const& _oldCode, u256 const& _oldVersion)
```
gumb0: I forgot what's the exact reason to save the old code, I'll try to remember, but if it's needed, probably we need to save the version, too, and restore during rollback.

libethereum/Account.cpp
```diff
@@ -155,7 +156,7 @@ AccountMap dev::eth::jsonToAccountMap(std::string const& _json, u256 const& _def
                         cerr << "Error importing code of account " << a
                              << "! Code needs to be hex bytecode prefixed by \"0x\".";
                     else
-                        ret[a].setCode(fromHex(codeStr));
+                        ret[a].setCode(fromHex(codeStr), 0);
```
gumb0: I'll create an issue to handle this properly. We should have the ability to set contract's version in config file.

libethereum/State.h
gumb0: Apparently this is not needed anymore, I think I'll remove saving the old code in https://github.com/ethereum/aleth/pull/5641, then will rebase this.

libethereum/Account.cpp
gumb0: https://github.com/ethereum/aleth/issues/5645

libethereum/Account.cpp
```diff
@@ -34,18 +34,24 @@ using namespace dev::eth::validation;
...
 namespace fs = boost::filesystem;
 
-void Account::setCode(bytes&& _code)
+void Account::setCode(bytes&& _code, u256 const& _version)
 {
-    m_codeCache = std::move(_code);
-    m_hasNewCode = true;
-    m_codeHash = sha3(m_codeCache);
+    auto const newHash = sha3(_code);
+    if (newHash != m_codeHash)
```
gumb0: I think we should have done this check before, too, because without it in case init code returns empty code, `setCode({})` raises `m_hasNewCode` flag, and this makes `commit` function to save into database the pair `EmptySHA3 => ""`. This should better not happen.

Here it also helps with the same problem in case of empty init code. In this case `setCode` is still called to set the version.

libethcore/EVMSchedule.h
```diff
@@ -146,10 +147,26 @@ static const EVMSchedule ConstantinopleFixSchedule = [] {
     return schedule;
 }();
 
+static const EVMSchedule IstanbulSchedule = [] {
+    EVMSchedule schedule = ConstantinopleFixSchedule;
+    schedule.version = 1;
```
gumb0: This value is not finalized yet, see https://github.com/ethereum/EIPs/pull/2156

gumb0:
I think this is ready for review. I've also added some relevant unit tests for `Executive`, which we never had before.

libethashseal/genesis/test/istanbulTransitionTest.cpp
```diff
@@ -0,0 +1,66 @@
+/*
```
halfalicious: Shouldn't we have the new license message?

libethcore/EVMSchedule.h
```diff
@@ -29,6 +29,7 @@ struct EVMSchedule
 {
     EVMSchedule(): tierStepGas(std::array<unsigned, 8>{{0, 2, 3, 5, 8, 10, 20, 0}}) {}
     EVMSchedule(bool _efcd, bool _hdc, unsigned const& _txCreateGas): exceptionalFailedCodeDeposit(_efcd), haveDelegateCall(_hdc), tierStepGas(std::array<unsigned, 8>{{0, 2, 3, 5, 8, 10, 20, 0}}), txCreateGas(_txCreateGas) {}
+    int version = 0;
```
halfalicious: Should this be unsigned since values are only >= 0?

libethereum/Account.cpp
```diff
@@ -34,18 +34,24 @@ using namespace dev::eth::validation;
 
 namespace fs = boost::filesystem;
 
-void Account::setCode(bytes&& _code)
+void Account::setCode(bytes&& _code, u256 const& _version)
```
halfalicious: Not related to your changes, but why do we have a setCode function rather than only allowing code to be set in the account object in a ctor since from what I understand smart contracts can't be redeployed at the same address?

halfalicious:
Should we also reset the version in `Account::kill`: https://github.com/ethereum/aleth/blob/23b85b576223b44310fa95cda0a62f7e1e8a25ee/libethereum/Account.h#L92

halfalicious:
Where do we retrieve the code version during syncing pass it to the VM?

libethcore/EVMSchedule.h
gumb0: Well in general it's not a good idea to try to enforce non-negative values with `unsigned` (because it doesn't really enforce it)
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es106-dont-try-to-avoid-negative-values-by-using-unsigned
But now I think that I'll change it since other values in this struct are `unsigned` (gas prices), and in other parts of the code I made it `u256`

Maybe makes sense to make it `u256` here as well, but for now `unsigned` definitely will work, too.

libethereum/Account.cpp
gumb0: I think because contract creation has init code execution phase. First init code (provided in the contract deployment transaction or in CREATE/CREATE2 params) is executed, and it returns in the end another code, which becomes the code of new contract, saved into the state (with `setCode`).

So init code execution is performed in the context of the newly created contract, and so this new contract should already exist, but its code is empty until init code is finished.

gumb0:
@halfalicious Good catch about `Account::kill`. Doesn't look like it would affect consensus, because killed accounts are not executed / are removed from DB anyway, but it's better to set version to 0 there, too, I think.

aleth-vm/main.cpp
```diff
@@ -259,13 +259,18 @@ int main(int argc, char** argv)
...
+    EnvInfo const envInfo(blockHeader, lastBlockHashes, 0);
+
     Transaction t;
     Address contractDestination("1122334455667788991011121314151617181920");
     if (!code.empty())
     {
         // Deploy the code on some fake account to be called later.
         Account account(0, 0);
-        account.setCode(bytes{code});
+        auto const latestVersion = se->evmSchedule(envInfo.number()).version;
```
chfast: Can you run this change in separate PR?

libethashseal/genesis/test/istanbulTransitionTest.cpp
```diff
@@ -0,0 +1,54 @@
+// Aleth: Ethereum C++ client, tools and libraries.
+// Copyright 2019 Aleth Authors.
+// Licensed under the GNU General Public License, Version 3.
+
+#include "../../GenesisInfo.h"
+
+static std::string const c_genesisInfoIstanbulTransitionTest = std::string() +
```
chfast: Can you add Istanbul configs being a copy of Petersburg in separate PR?

libethcore/EVMSchedule.h
```diff
@@ -29,6 +29,7 @@ struct EVMSchedule
 {
     EVMSchedule(): tierStepGas(std::array<unsigned, 8>{{0, 2, 3, 5, 8, 10, 20, 0}}) {}
     EVMSchedule(bool _efcd, bool _hdc, unsigned const& _txCreateGas): exceptionalFailedCodeDeposit(_efcd), haveDelegateCall(_hdc), tierStepGas(std::array<unsigned, 8>{{0, 2, 3, 5, 8, 10, 20, 0}}), txCreateGas(_txCreateGas) {}
+    unsigned version = 0;
```
chfast: Maybe it should be named `accountVersion` for precision?

libethcore/EVMSchedule.h
```diff
@@ -146,10 +147,26 @@ static const EVMSchedule ConstantinopleFixSchedule = [] {
...
 }();
+
+inline EVMSchedule const& evmScheduleForAccountVersion(u256 const& _version)
+{
+    if (_version == 0)
+        return ConstantinopleFixSchedule;
+    else if (_version == IstanbulSchedule.version)
+        return IstanbulSchedule;
+    else
+        return DefaultSchedule;
```
chfast: Does the EIP specifies the rules here?

libethereum/Account.cpp
chfast: Separate PR?

libethereum/Account.cpp
```diff
@@ -34,18 +34,24 @@ using namespace dev::eth::validation;
...
+    }
+    m_version = _version;
 }
 
 void Account::resetCode()
 {
     m_codeCache.clear();
     m_hasNewCode = false;
     m_codeHash = EmptySHA3;
+    m_version = 0;
```
chfast: Add a comment why.

libethereum/ExtVM.h
```diff
@@ -97,19 +98,28 @@ class ExtVM : public ExtVMFace
...
 
 private:
+    EVMSchedule const& initEvmSchedule(int64_t _blockNumber, u256 const& _version) const
+    {
+        // Run with a schedule according to hard fork if before Istanbul
+        // Run with a schedule according to account version if Istanbul and later
+        if (_blockNumber < m_sealEngine.chainParams().istanbulForkBlock)
+            return m_sealEngine.evmSchedule(_blockNumber);
+        else
+            return evmScheduleForAccountVersion(_version);
```
chfast: Wouldn't this work all the time because we have version 0 before Istanbul?

libethashseal/genesis/test/istanbulTransitionTest.cpp
gumb0: Yes

libethcore/EVMSchedule.h
gumb0: Ok

libethereum/Account.cpp
gumb0: Yeah can be separate

libethcore/EVMSchedule.h
gumb0: No, but this should never happen, as all new accounts are created either with version 0 or with the version of the current fork.
We could add an assertion here.

aleth-vm/main.cpp
gumb0: Not sure what you mean, `setCode` now has two parameters, so we should pass something anyway to compile it.

chfast: I misread it. I thought it takes the latest schedule by default.

libethcore/EVMSchedule.h
chfast: Go with assert.

libethereum/ExtVM.h
gumb0: Not really, `EvmSchedule` used changes each fork before Istanbul, so this function to work correctly would need to return different ones for version 0 based on block number.
If I understood correctly your question.

chfast: Yes, you are right.

But I think the more future proof strategy is to select schedule group by version number and then select the schedule within the group by the block number. As I understand the EIP-1702 and as I want to be, we will not bump the version on each HF but only if some breaking changes are introduced.

aleth-vm/main.cpp
gumb0: It takes the schedule according to `--network` CLI argument (though there's no Istanbul yet, that would go into separate PR)

libethashseal/GenesisInfo.h
```diff
@@ -67,6 +67,8 @@ enum class Network
     ConstantinopleNoProofTest = 82,
     /// Byzantium + Constantinople + ConstantinopleFix active from block 0
     ConstantinopleFixTest = 83,
+    /// ConstantinopleFixTest + Istanbul active from block 2
```
halfalicious: Why is this active from block 2 rather than block 0?

libethashseal/GenesisInfo.cpp
```diff
@@ -77,6 +78,8 @@ std::string const& dev::eth::genesisInfo(Network _n)
         return c_genesisInfoExperimentalTransitionTest;
     case Network::ConstantinopleFixTest:
         return c_genesisInfoConstantinopleFixTest;
+    case Network::IstanbulTransitionTest:
```
halfalicious: Do we need to add configuration information for Instanbul as a part of this PR?

libethcore/EVMSchedule.h
halfalicious: Is this the best place to store the version? EVMSchedule appears to only contain opcode / gas cost info, is there another place where we store other VM configuration information?

libethereum/Account.cpp
halfalicious: Ah okay, thanks for the clarification! Still not clear on this:

>  and it returns in the end another code, which becomes the code of new contract, saved into the state (with setCode).

What is this other code that is returned by init code execution, how does it differ from the code supplied in the contract creation tx?

libethcore/EVMSchedule.h
halfalicious: I was thinking less of enforcing non-negative values and more of code clarity, i.e. someone reading the code will be able to tell that version should be non-negative.

libethashseal/GenesisInfo.h
gumb0: This is a special kind of config for the tests to be able to check the behavior before Istanbul (in block 1) as well as after Istanbul. We have similar configs for other forks, too.

libethashseal/GenesisInfo.cpp
gumb0: I'll split it into separate PR. (note that this is a config only for tests, though)

libethcore/EVMSchedule.h
gumb0: There's also at least block reward value, and some other things. I think it's an ok place, I don't think there is any other currently.

libethereum/Account.cpp
gumb0: The code supplied in the contract creation tx (or in `CREATE`/`CREATE2` instructions) can return any data (via `RETURN` instruction). The rules are that the data returned from this code becomes the code of the newly created account.

In other words the code supplied in the contract creation tx is only constructor code (in solidity terms) and it _returns_ the code that becomes the actual contract code.

libethereum/ExtVM.h
gumb0: @chfast I see, it makes sense, but also makes the logic of choosing schedule more complicated, as there's not one-to-one relation between schedules and versions.

I changed this just slightly now. It's not exactly as you suggest, but I think more in line with this strategy.
So basically we care about choosing by block number only within the latest version, while all past versions choose the fixed schedule "just before the version bump".
What do you think?

gumb0:
Rebased and issues addressed.

halfalicious:
@gumb0 Feel free to check in once Pawel signs off, I don't know enough about the related areas to be able to sign off on this PR with a sufficient level of confidence

libethereum/Executive.cpp
```diff
@@ -411,7 +427,11 @@ bool Executive::executeCreate(Address const& _sender, u256 const& _endowment, u2
     // Schedule _init execution if not empty.
     if (!_init.empty())
         m_ext = make_shared<ExtVM>(m_s, m_envInfo, m_sealEngine, m_newAddress, _sender, _origin,
-            _endowment, _gasPrice, bytesConstRef(), _init, sha3(_init), m_depth, true, false);
+            _endowment, _gasPrice, bytesConstRef(), _init, sha3(_init), _version, m_depth, true,
+            false);
+    else
+        // code stays empty, but we set the version
+        m_s.setCode(m_newAddress, {}, _version);
```
halfalicious: What's the point of setting the version if there's no code?

libethereum/State.cpp
```diff
@@ -168,11 +168,15 @@ Account* State::account(Address const& _addr)
...
-        std::forward_as_tuple(state[0].toInt<u256>(), state[1].toInt<u256>(), state[2].toHash<h256>(), state[3].toHash<h256>(), Account::Unchanged)
-    );
+    auto const nonce = state[0].toInt<u256>();
+    auto const balance = state[1].toInt<u256>();
+    auto const storageRoot = state[2].toHash<h256>();
+    auto const codeHash = state[3].toHash<h256>();
+    // version is 0 if absent from RLP
+    auto const version = state[4] ? state[4].toInt<u256>() : 0;
+
+    auto i = m_cache.emplace(std::piecewise_construct, std::forward_as_tuple(_addr),
```
halfalicious: Don't need `std::` prefix since there's `using namespace std`

libethereum/State.cpp
```diff
@@ -516,13 +520,13 @@ bytes const& State::code(Address const& _addr) const
...
 
-void State::setCode(Address const& _address, bytes&& _code)
+void State::setCode(Address const& _address, bytes&& _code, u256 const& _version)
 {
     // rollback assumes that overwriting of the code never happens
     // (not allowed in contract creation logic in Executive)
     assert(!addressHasCode(_address));
     m_changeLog.emplace_back(Change::Code, _address);
-    m_cache[_address].setCode(std::move(_code));
+    m_cache[_address].setCode(std::move(_code), _version);
```
halfalicious: Don't need `std::` prefix

libethereum/Executive.cpp
chfast: This question might be delegated to the EIP.

gumb0: I discussed this with EIP author here https://github.com/sorpaas/EIPs/issues/2#issuecomment-506363868

gumb0:
@halfalicious Please approve or dismiss your review

halfalicious:
> 
> 
> @halfalicious Please approve or dismiss your review

@gumb0 : Done!

gumb0:
Rebased.

PR #5641, opened 2019-06-26
Reset code to empty in rollback

gumb0:
I think we don't need to save the old code for rollback anymore, because overwriting the code is not allowed according to address collision rules (https://github.com/ethereum/EIPs/issues/684 and https://eips.ethereum.org/EIPS/eip-689) and we never get to rolling back the overwritten code.

A bit of history of this code:
It was written as a fix of incorrect rollback of code overwrite in case of address collision (back then it was possible to overwrite the code)
https://github.com/ethereum/aleth/issues/4130
https://github.com/ethereum/aleth/pull/4151

Then in preparation for Byzantium new collision rules were introduced, which applied retroactively to all previous forks, too.
https://github.com/ethereum/aleth/pull/4384

To quote EIP-689:

> Regarding testing, this EIP relieves clients from supporting reversion of code overwriting.

This will simplify a little bit the implementation of EIP-1702 Generalized Account Versioning Scheme, because if we don't need to save the old code in journal, then we don't need to save the version either.

### DIFF
--- a/libethereum/Account.cpp
+++ b/libethereum/Account.cpp
@@ -41,6 +41,13 @@ void Account::setCode(bytes&& _code)
     m_codeHash = sha3(m_codeCache);
 }
 
+void Account::resetCode()
+{
+    m_codeCache.clear();
+    m_hasNewCode = false;
+    m_codeHash = EmptySHA3;
+}
+
 u256 Account::originalStorageValue(u256 const& _key, OverlayDB const& _db) const
 {
     auto it = m_storageOriginal.find(_key);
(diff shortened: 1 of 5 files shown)


chfast:
Why is the EIP-689 still a draft?

gumb0:
@chfast Probably no one cared to accept it yet. There was some discussion recently on All Core Devs call, where @holiman mentioned that it's already implemented, see discussion of "Action 60.6"  in https://github.com/ethereum/pm/blob/master/All%20Core%20Devs%20Meetings/Meeting%2061.md

halfalicious:
@gumb0 : When would we need to rollback to old code before EIP-689? When there's a chain reorg?

gumb0:
@halfalicious No, rollback happens any time after exception during transaction execution (for example Out-of-gas or invalid instruction)

PR #5647, opened 2019-07-03
Handle import failures in test_importRawBlock

### LINKED ISSUE #5646: test_importRawBlock doesn't handle errors properly
It succeeds even if the block import failed.

@winsvega reports that this is what happens when importing invalid block:
```
WARN  06-30 14:53:01 aleth warn   
                                                                                
                                                                                
    Import Failure     /home/wins/Ethereum/cpp-ethereum/libethereum/Block.cpp(565): Throw in function dev::u256 dev::eth::Block::enact(const dev::eth::VerifiedBlockRef&, const dev::eth::BlockChain&)
Dynamic exception type: boost::exception_detail::clone_impl<dev::eth::UncleInChain>
[dev::tag_comment*] = Uncle in block already mentioned
[dev::eth::tag_uncleIndex*] = 1
[dev::eth::tag_unclesExcluded*] = { 9e38ab1792276d10df19067001684545b2c905a44423f6e50b1835b12b3ac3ef, 3874ace79c930a81d6c8e77c2eaec5c2d474678c120ce35c2aa3c5bf25bf8012, 250c192ac598dcc6d79142a983be39358eeba2547d689b91cab661da74aa5cfa, 5a39ed1020c04d4d84539975b893a4e7c53eab6c2965db8bc3468093a31bc5ae, 2b6e0badc6ea187cc798918e1515dcbe3f0e1cecec42dcd51e63a63cdf2f1e66 }
[dev::eth::tag_now*] = 1561895581
[dev::eth::tag_block*] = [ type: std::vector<unsigned char, std::allocator<unsigned char> >, size: 24, dump: d0 3b 05 b8 64 7f 00 00 2c 42 05 b8 64 7f 00 00 ]
[dev::tag_hash*] = 250c192ac598dcc6d79142a983be39358eeba2547d689b91cab661da74aa5cfa
    
                                                                                
                       Bad Block #00000003.2b6e0bad…                      
                                                                                
                                                                                
TRACE 06-30 14:53:01 aleth rpc    {"id":118,"jsonrpc":"2.0","result":"0x2b6e0badc6ea187cc798918e1515dcbe3f0e1cecec42dcd51e63a63cdf2f1e66"}
```
https://github.com/ethereum/aleth/blob/be6ce38cfac2710b1710d951cc93c441dbc80fe2/libethereum/ClientTest.cpp#L157-L163

The simple fix to try would be to look at fresh blocks returned as the first value from `syncQueue`. If `blockHash` was not returned there, it means it was rejected as bad block, and then we should throw exception in this case.

gumb0:
Fixes #5646


test/unittests/libweb3jsonrpc/jsonrpc.cpp
```diff
@@ -729,7 +729,33 @@ BOOST_AUTO_TEST_CASE(test_importRawBlock)
     Json::Value ret;
     Json::Reader().parse(c_genesisConfigString, ret);
     rpcClient->test_setChainParams(ret);
+
     string blockHash = rpcClient->test_importRawBlock(
```
gumb0: This test actually failed to import the block (difficulty was wrong in that block after changing difficulty rules for NoProof), but `test_importRawBlock` didn't report it.

winsvega: difficulty rules are still checked for NoProof. 
nonce and mixHash are not checked.

gumb0: > difficulty rules are still checked for NoProof.
that's the reason why this test worked incorrectly (block was invalid, I guess you created it when there were no difficulty check in NoProof yet)

I changed it now to have correct difficulty.

PR #5648, opened 2019-07-03
split BlockchainTests into Valid and Invalid

winsvega:
split BlockchainTests into BlockchainTests/ValidBlocks and BlockchainTests/InvalidBlocks
move TransitionTests under BlockchainTests

ValidBlocks contain only tests with valid rlps 
InvalidBlocks contain tests where some block rlp's are invalid and expect and exception upon import

Update test subfolder to this PR https://github.com/ethereum/tests/pull/605

### DIFF
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -59,7 +59,8 @@ namespace test {
 eth::Network ChainBranch::s_tempBlockchainNetwork = eth::Network::MainNetwork;
 eth::Network TestBlockChain::s_sealEngineNetwork = eth::Network::FrontierTest;
 
-json_spirit::mValue BlockchainTestSuite::doTests(json_spirit::mValue const& _input, bool _fillin) const
+json_spirit::mValue doBCTest(
+    json_spirit::mValue const& _input, bool _fillin, bool _allowInvalidBlocks)
 {
     json_spirit::mObject tests;
     for (auto const& i : _input.get_obj())
@@ -122,7 +123,7 @@ json_spirit::mValue BlockchainTestSuite::doTests(json_spirit::mValue const& _inp
                     std::cout << "Filling " << newtestname << std::endl;
 
                 TestOutputHelper::get().setCurrentTestName(newtestname);
-                jObjOutput = fillBCTest(jObjOutput);
+                jObjOutput = fillBCTest(jObjOutput, _allowInvalidBlocks);
                 jObjOutput["network"] = test::netIdToString(network);
                 if (inputTest.count("_info"))
                     jObjOutput["_info"] = inputTest.at("_info");
@@ -161,13 +162,34 @@ json_spirit::mValue BlockchainTestSuite::doTests(json_spirit::mValue const& _inp
 
     return tests;
 }
-fs::path BlockchainTestSuite::suiteFolder() const
+
+json_spirit::mValue BlockchainValidTestSuite::doTests(
+    json_spirit::mValue const& _input, bool _fillin) const
 {
-    return "BlockchainTests";
+    return doBCTest(_input, _fillin, false);
 }
-fs::path BlockchainTestSuite::suiteFillerFolder() const
+
+json_spirit::mValue BlockchainInvalidTestSuite::doTests(
+    json_spirit::mValue const& _input, bool _fillin) const
 {
-    return "BlockchainTestsFiller";
+    return doBCTest(_input, _fillin, true);
+}
+
+fs::path BlockchainInvalidTestSuite::suiteFolder() const
+{
+    return fs::path("BlockchainTests") / "InvalidBlocks";
+}
+fs::path BlockchainInvalidTestSuite::suiteFillerFolder() const
+{
+    return fs::path("BlockchainTestsFiller") / "InvalidBlocks";
+}
+fs::path BlockchainValidTestSuite::suiteFolder() const
+{
+    return fs::path("BlockchainTests") / "ValidBlocks";
+}
+fs::path BlockchainValidTestSuite::suiteFillerFolder() const
+{
+    return fs::path("BlockchainTestsFiller") / "ValidBlocks";
 }
 fs::path BCGeneralStateTestsSuite::suiteFolder() const
 {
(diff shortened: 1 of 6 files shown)


gumb0:
We should first fix failing tests where we disagree with geth, I try to work on that in https://github.com/ethereum/aleth/pull/5652

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -124,14 +124,49 @@ bool pathHasTests(boost::filesystem::path const& _path)
     return false;
 }
 
-void TestOutputHelper::checkUnfinishedTestFolders()
+string TestOutputHelper::detectFilterForMinusTArgument()
```
gumb0: Can you summarize what are the changes in this function? This seems unrelated to the rest

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -1024,6 +1054,20 @@ class bcTestFixture {
     }
 };
 
+class bcInvalidTestFixture
```
gumb0: This class is almost identical to the one above, it looks like you could have a base class common for both to avoid code repetition. Just a suggestion.

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -416,19 +439,25 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input)
...
+            blockImportExceptionHappen = true;
         }
         catch (...)
         {
             cout << testName + "block import thrown unknown exeption\n";
             eraseJsonSectionForInvalidBlock(blObj);
+            blockImportExceptionHappen = true;
         }
 
+        if (!_allowInvalidBlocks && blockImportExceptionHappen)
```
gumb0: What if invalid was expected, but exception was not thrown, will that be caught somewhere?

winsvega: yes it is being checked at function `checkExpectedException(blObj, _e);` in catch block above

test/tools/jsontests/BlockChainTests.cpp
winsvega: There is a plan for malicious block blockchain testsing, when the test tool will produce invlid block rlps and try to import. 
Thats for the future. for now just separate tests with invalid blocks that we have.

test/tools/libtesteth/TestOutputHelper.cpp
winsvega: thats for afterrun checkup of unused and empty folders. 
it will detect -t filters properly for case    <TestSuite>/<SubtestSuite>/<TestCase>
where need to look for unused and empty folders with <TestCase> names.

test/tools/jsontests/BlockChainTests.cpp
gumb0: That looks like something else, because it's in `catch` block, so is checked only when excption is thrown.

I'm talking about the case when `_allowInvalidBlocks && !blockImportExceptionHappen`

test/tools/libtesteth/TestOutputHelper.cpp
gumb0: I would prefer this to go into separate PR

gumb0: The main part would be merged faster then, I expect

test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -124,14 +124,49 @@ bool pathHasTests(boost::filesystem::path const& _path)
...
+    // if (SuiteName/subSuiteName == -t argument) then filter is empty
+    bool thereisPathEqualToSuite = false;
+    for (auto const& allTestsIt : m_finishedTestFoldersMap)
+    {
+        // Remove Filler from folder name so it equal to -t argument
+        string parent = allTestsIt.first.parent_path().stem().string();
+        size_t index = parent.find("Filler");
+        if (index == std::string::npos)
+            break;
+        parent = parent.replace(index, 6, "");
```
gumb0:
```suggestion
        parent = parent.substr(0, index);
```

test/tools/jsontests/BlockChainTests.cpp
winsvega: in case when `_allowInvalidBlocks && !blockImportExceptionHappen`  blocks without exception are allowed.

gumb0: Aren't those blocks invalid?
Do you mean that not all blocks in Invalid suite generate exceptions? Why are they in Invalid suite then?

winsvega: Sometimes for a block to be invalid chain must contain some valid blocks first

test/tools/libtesteth/TestOutputHelper.cpp
winsvega: hmm. I think it wont work. `BlockchainTestsFiller/sometestsFoder1`
I need to get  `BlockchainTestsFiller/sometestsFoder1`
but will get `BlockchainTests`

winsvega: hm. another hardcode on folder names.

test/tools/libtesteth/TestOutputHelper.h
```diff
@@ -60,6 +60,7 @@ class TestOutputHelper
 	TestOutputHelper() {}
     void checkUnfinishedTestFolders();  // Checkup that all test folders are active during the test
                                         // run
+    std::string detectFilterForMinusTArgument();
```
gumb0: this is not needed anymore?

test/unittests/libethereum/GasPricer.cpp
```diff
@@ -95,95 +95,95 @@ BOOST_AUTO_TEST_CASE(basicGasPricer_RPC_API_Test_Frontier)
 {
 	u256 _expectedAsk = 16056883295;
 	u256 _expectedBid = 1;
-	dev::test::executeGasPricerTest("RPC_API_Test_Frontier", 30.679, 15.0, "/BlockchainTests/bcGasPricerTest/RPC_API_Test.json", TransactionPriority::Medium, _expectedAsk, _expectedBid, eth::Network::FrontierTest);
+    dev::test::executeGasPricerTest("RPC_API_Test_Frontier", 30.679, 15.0, "/BlockchainTests/ValidBlocks/bcGasPricerTest/RPC_API_Test.json", TransactionPriority::Medium, _expectedAsk, _expectedBid, eth::Network::FrontierTest);
```
gumb0: Would be better to get rid of repetition here, moving `"/BlockchainTests/ValidBlocks/"` to a constant

gumb0: or by moving it inside `executeGasPricerTest` function

PR #5652, opened 2019-07-04
Coming to agreement with geth on RIPEMD touch revert

gumb0:
> figured cpp does a very complicated thing. It's so complicated that I don't want to describe it.
> 
> -- Yoichi Hirai

This is the old issue, originated that time when both Geth and Parity had incorrect revert behavior, and then it was decided to introduce a hack into consensus rules instead of doing huge chain reorg.

The buggy behavior that is now a rule is basically: when precompiled contract with address `0x03` (RIPEMD160) is touched, then exception happens and state changes are reverted, this address should be still considered touched in the end, so that "remove empty touched accounts" rule (EIP-158) deletes it from the state at the end of transaction.

There are several related tests created after this incident:
- [`GeneralStateTests/stSpecialTest/tx_xcf416c53`](https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stSpecialTest/failed_tx_xcf416c53Filler.json) - emulates the transaction where this special behavior took place on the main net. Here the gas given to `CALL` to RIPEMD is not enough, so the `CALL` itself fails, but then transaction proceeds further till success. RIPEMD stays touched and then is deleted. All clients pass this test (otherwise they wouldn't be able to do full sync)
- [`GeneralStateTests/stRevertTest/RevertPrecompiledTouch`](https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stRevertTest/RevertPrecompiledTouchFiller.json) and  [`GeneralStateTests/stRevertTest/RevertPrecompiledTouch_storage`](https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stRevertTest/RevertPrecompiledTouch_storageFiller.json) - here `CALL` to RIPEMD succeeds, but later execution runs out of gas. Changes to the state are reverted, but RIPEMD stays touched and then is deleted. In these tests we have different results than geth.

Dimitriy now has regenerated these latter tests via geth, and now aleth doesn't pass them. 

In geth all of these cases look to be handled by just one hack https://github.com/ethereum/go-ethereum/blob/7527215a68eaac8a880be83f3f1ded0ed82f6650/core/state/state_object.go#L145-L149

Our current hack covers only the first `tx_xcf416c53` case, because for us failing `CALL` to precompile is a special case, that doesn't invoke any rollback https://github.com/ethereum/aleth/blob/007d89a712209ca16198be41a5d9c70498138eb6/libethereum/Executive.cpp#L313-L318

Now for us to cover the case of OOG in caller's frame, it looks like we unfortunately have to add another similar hack in revert logic.

From the point of view of our code it feels like an unnecessary hack, not related to original problem, but I guess from geth's perspective to replicate our behavior they would need to do some "unnecessary" hacks in their code. Oh well.

More references:
- Original geth's fix https://github.com/ethereum/go-ethereum/pull/3341/files#diff-2433aa143ee4772026454b8abd76b9dd
- More detailed description https://github.com/ethereum/EIPs/issues/716
- Recent mention in AllCoreDevs chat https://gitter.im/ethereum/AllCoreDevs?at=5ce3bd9863dea422b4bc9b91
- Yellow Paper mentions this anomaly in Appendix K https://ethereum.github.io/yellowpaper/paper.pdf

cc @holiman @winsvega

### DIFF
--- a/libethereum/State.cpp
+++ b/libethereum/State.cpp
@@ -36,6 +36,11 @@ using namespace dev;
 using namespace dev::eth;
 namespace fs = boost::filesystem;
 
+namespace
+{
+Address const c_RipemdPrecompiledAddress{0x03};
+}
+
 State::State(u256 const& _accountStartNonce, OverlayDB const& _db, BaseState _bs):
     m_db(_db),
     m_state(&m_db),
@@ -589,8 +594,16 @@ void State::rollback(size_t _savepoint)
             account.resetCode();
             break;
         case Change::Touch:
-            account.untouch();
-            m_unchangedCacheEntries.emplace_back(change.address);
+            // Empty RIPEMD contract needs to be deleted even in case of exception.
+            // This doesn't affect main net, but is needed for some consensus tests.
+            // https://github.com/ethereum/go-ethereum/pull/3341/files#diff-2433aa143ee4772026454b8abd76b9dd
+            // https://github.com/ethereum/EIPs/issues/716
+            // https://github.com/ethereum/aleth/pull/5652
+            if (change.address != c_RipemdPrecompiledAddress)
+            {
+                account.untouch();
+                m_unchangedCacheEntries.emplace_back(change.address);
+            }
             break;
         }
         m_changeLog.pop_back();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 - Fixed: [#5627](https://github.com/ethereum/aleth/pull/5627) Correct testeth --help log output indentation.
 - Fixed: [#5644](https://github.com/ethereum/aleth/pull/5644) Avoid attempting to sync with disconnected peers.
 - Fixed: [#5647](https://github.com/ethereum/aleth/pull/5647) test_importRawBlock RPC method correctly fails in case of import failure.
+- Fixed: [#5652](https://github.com/ethereum/aleth/pull/5652) Behavior in corner case tests about touching empty Precompiles now agrees with geth's results.
 
 
 ## [1.6.0] - 2019-04-16


gumb0:
Looks now like the revert tests generated with old testeth are failing, and the tests generated with geth are passing.

gumb0:
Note that I don't see any difference in test results, if I change our original hack ([the one covering failed `CALL`](https://github.com/ethereum/aleth/blob/007d89a712209ca16198be41a5d9c70498138eb6/libethereum/Executive.cpp#L317)) to handle only RIPEMD and not other precompiles.

This suggests that tests might not cover the cases of `CALL` to other empty precompiles running OOG and leaving them touched/untouched.
(so similar to `tx_xcf416c53`, but with other precompiles that are empty and not enough gas for `CALL`)
Not sure if we should care about this coverage.

@holiman @winsvega

libethereum/State.cpp
```diff
@@ -589,8 +594,16 @@ void State::rollback(size_t _savepoint)
...
             break;
         case Change::Touch:
-            account.untouch();
-            m_unchangedCacheEntries.emplace_back(change.address);
+            // Empty RIPEMD contract needs to be deleted even in case of exception.
+            // This doesn't affect main net, but is needed for some consensus tests.
+            // https://github.com/ethereum/go-ethereum/pull/3341/files#diff-2433aa143ee4772026454b8abd76b9dd
+            // https://github.com/ethereum/EIPs/issues/716
+            // https://github.com/ethereum/aleth/pull/5652
+            if (change.address != c_RipemdPrecompiledAddress)
```
chfast: Uff, this check is actually placed in some hot part of the code...

gumb0: Maybe we could try to introduce some new kind of touch, that is un-revertable and called only for precompiles...

gumb0: @chfast Ok, here's the alternative solution https://github.com/ethereum/aleth/pull/5663
No hacks in `rollback()` at the expense of additional flag in `Account` class.

I think I like it better, at least it allows to have one central piece of weirdness only in `Executive`.

gumb0: Actually now I have a third idea, to keep the list of unrevertably touched accounts in `State`, instead of putting the flag in each `Account`.
Third PR is coming...

gumb0: https://github.com/ethereum/aleth/pull/5664

winsvega:
> tx_xcf416c53

added more tests around touch on precompiled addresses and OOG / NO OOG within the same call, not within the same transaction

https://github.com/ethereum/tests/pull/610

winsvega:
have a new test case for this 
https://github.com/ethereum/tests/pull/610

gumb0:
> have aleth argue with geth on Byzantium no OOG touch to 0x0000....008 empty account

> have a new test case for this
> ethereum/tests#610

This branch and `untouch-ripemd-alternative2` branch pass it, did you try with aleth's `master`?

winsvega:
> did you try with aleth's master
no

> This branch and untouch-ripemd-alternative2 branch pass it, 
good. merging the tests then

gumb0:
Closing in favor of https://github.com/ethereum/aleth/pull/5664

PR #5656, opened 2019-07-08
Don't save empty code into database

gumb0:
Extracted from #5640.

When init code returns empty code, account is still created, and setting `m_hasNewCode` to `true` in `setCode` was leading to `commit` function pushing into the database the pair `EmptySHA3 => ""`, which is not necessary.

This doesn't affect consensus.

### DIFF
--- a/libethereum/Account.cpp
+++ b/libethereum/Account.cpp
@@ -36,9 +36,13 @@ namespace fs = boost::filesystem;
 
 void Account::setCode(bytes&& _code)
 {
-    m_codeCache = std::move(_code);
-    m_hasNewCode = true;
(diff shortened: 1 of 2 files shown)


halfalicious:
> 
> 
> When init code returns empty code, account is still created, and setting `m_hasNewCode` to `true` in `setCode` was leading to `commit` function pushing into the database the pair `EmptySHA3 => ""`, which is not necessary.

Out of curiosity, what is the function which performs init code?

gumb0:
@halfalicious It's prepared by `Executive::create*` functions, then execution is started by `Executive::go`

PR #5658, opened 2019-07-08
[WIP] Aleth now waits for 2 secs after sending disconnect packet #5650

### LINKED ISSUE #5650: Aleth should wait 2 seconds to close socket after sending disconnect to peer
According to the [RLPX handshake spec](https://github.com/ethereum/devp2p/blob/master/rlpx.md), one should give clients 2 seconds to disconnect after sending a disconnect packet:

> Inform the peer that a disconnection is imminent; if received, a peer should disconnect immediately. When sending, well-behaved hosts give their peers a fighting chance (read: wait 2 seconds) to disconnect to before disconnecting themselves.

However, Aleth closes the socket immediately (via `Session::drop`). This also means that the disconnect packet may not be sent - this is because `Session::disconnect` (and all of the functions it calls) is executed on the network thread, whereas the `async_write` may not execute immediately in which case I suspect it will execute after `Session::disconnect` / `Session::drop`:  https://github.com/ethereum/aleth/blob/c630593658da5044f3498e67697d658ab6f12ee3/libp2p/Session.cpp#L299-L310

Here's where the actual socket write takes place (in `Session::write`): https://github.com/ethereum/aleth/blob/c630593658da5044f3498e67697d658ab6f12ee3/libp2p/Session.cpp#L237-L256

twinstar26:
Fixes #5650

### DIFF
--- a/libp2p/Session.cpp
+++ b/libp2p/Session.cpp
@@ -305,8 +305,20 @@ void Session::disconnect(DisconnectReason _reason)
         RLPStream s;
         prep(s, DisconnectPacket, 1) << (int)_reason;
         sealAndSend(s);
+        auto disconnectTimer = m_server->createTimer();
+        auto self(shared_from_this());
+        disconnectTimer->expires_from_now(std::chrono::seconds(2));
+        disconnectTimer->async_wait([self, this, _reason](boost::system::error_code _ec) {
+            if(_ec)
+                LOG(m_netLoggerDetail)
+                    << "Error encountered while waiting after disconnect: " << _ec.message();
+            else
+                drop(_reason);
+        });
+        m_server->pollAsyncTimerHandler();
     }
-    drop(_reason);
+    else
+        drop(_reason);
 }
 
 void Session::start()
--- a/libp2p/Host.h
+++ b/libp2p/Host.h
@@ -258,6 +258,16 @@ public:
 
     std::shared_ptr<CapabilityHostFace> capabilityHost() const { return m_capabilityHost; }
 
+    std::shared_ptr<ba::steady_timer> createTimer()
+    {
+        return std::make_shared<ba::steady_timer>(m_ioContext);
+    }
+
+    void pollAsyncTimerHandler()
+    {
+        m_ioContext.poll();
+    }
+
 protected:
     /*
      * Used by the host to run a capability's background work loop


halfalicious:
You're getting a build error:

```
[100/253] Building CXX object libp2p/CMakeFiles/p2p.dir/Session.cpp.o
FAILED: libp2p/CMakeFiles/p2p.dir/Session.cpp.o 
/usr/bin/ccache /usr/bin/clang++-6.0  -DETH_FATDB -I/home/builder/project -isystem /home/builder/.hunter/_Base/3511e5f/e2d20c6/0d35112/Install/include -I. -isystem deps/include -O3 -DNDEBUG -fPIC   -fstack-protector-strong -Wimplicit-fallthrough -Wall -Wextra -Werror -Wno-unknown-pragmas -fvisibility=hidden -std=c++11 -MD -MT libp2p/CMakeFiles/p2p.dir/Session.cpp.o -MF libp2p/CMakeFiles/p2p.dir/Session.cpp.o.d -o libp2p/CMakeFiles/p2p.dir/Session.cpp.o -c /home/builder/project/libp2p/Session.cpp
/home/builder/project/libp2p/Session.cpp:263:49: error: lambda capture 'this' is not used [-Werror,-Wunused-lambda-capture]
                    disconnectTimer.async_wait([this](boost::system::error_code const& ) {});
                                                ^
1 error generated.
```

halfalicious:
@twinstar26 Can you please add a changelog entry? Please see item #2 here for details: https://github.com/ethereum/aleth/blob/master/CONTRIBUTING.md

halfalicious:
@twinstar26 Can you please update the PR description to reference the issue you're fixing (so that the issue will be automatically updated with a link to the PR and closed when the PR is closed)?

libp2p/Session.cpp
```diff
@@ -206,6 +206,8 @@ bool Session::checkPacket(bytesConstRef _msg)
 
 void Session::send(bytes&& _msg)
 {
+    if(m_dropped)
```
halfalicious: (Nit) Formatting...can you please format your changes using `git clang-format` as described in item 3 here: https://github.com/ethereum/aleth/blob/master/CONTRIBUTING.md

libp2p/Session.cpp
```diff
@@ -229,14 +231,19 @@ void Session::send(bytes&& _msg)
 void Session::write()
 {
     bytes const* out = nullptr;
+
+    bool disconnectImminent = false;
```
halfalicious: This can be simplified to:

bool disconnectImminent = (writeQueue[0][0] == DisconnectPacket);

Also, since the type of the disconnectImminent value is obvious, we can make disconnectImmiment auto (and const since we're not going to modify its value after initially setting it). 

However, you can remove the disconnectImmiment check completely since @gumb0 brought up a good way to simplify the changes - start the timer in Session::disconnect after the sealAndSend call. The resulting wait won't be exactly 2 seconds after the disconnect packet is actually sent out but it should be close enough.

libp2p/Session.cpp
```diff
@@ -248,6 +255,24 @@ void Session::write()
             DEV_GUARDED(x_framing)
             {
                 m_writeQueue.pop_front();
+                if(m_writeQueue.empty() && disconnectImminent)
+                {
+                    ba::io_context ioc;
```
halfalicious: I think you should re-use the Host's io_context (`Host::m_ioContext`) - one way to do this would be to create the timer in `Host::startPeerSession` and pass it into the Session ctor. @gumb0 - thoughts?

libp2p/Session.cpp
```diff
@@ -248,6 +255,24 @@ void Session::write()
...
+                    ba::steady_timer disconnectTimer(ioc);
+                    disconnectTimer.expires_from_now(chrono::seconds(2));
+                    disconnectTimer.async_wait([this](boost::system::error_code const& ) {});
+                    ioc.run();
```
halfalicious: You don't need to explicitly call `run`, Host already does this (see `Host::doWork`)

libp2p/Session.cpp
```diff
@@ -248,6 +255,24 @@ void Session::write()
             DEV_GUARDED(x_framing)
             {
                 m_writeQueue.pop_front();
+                if(m_writeQueue.empty() && disconnectImminent)
+                {
+                    ba::io_context ioc;
+                    ba::steady_timer disconnectTimer(ioc);
+                    disconnectTimer.expires_from_now(chrono::seconds(2));
+                    disconnectTimer.async_wait([this](boost::system::error_code const& ) {});
```
halfalicious: You'll want to close the socket in the timer handler (since we want to close the socket after the timer has fired)...your code currently kicks off the timer (async) wait but then immediately closes the socket (if it's already open).

Note that the timer wait is asynchronous, not synchronous, which means that async_wait call returns immediately (the wait operation is inserted into a boost queue and the handler is inserted into the boost handler queue, and both queues are processed in response to calls of io_context::run on whatever thread run() is called on which in this case is the network thread

twinstar26: Oh that's what the anomaly was in my testing before and after changes. So stupid of me.

twinstar26:
Okay so I will be shifting the timer into Session::Disconnect. And we will wait for 2 secs. We woundn't care if the packet has been sent or not. Just wait for 2 secs and let the async writing happen (may happen in worst case that timer runs out before writing happens and we just close the socket in the middle of the operation). But I think we should make some kinda provision to let us know for sure that the operation has been completed. I mean the async is really cool but before closing we should make sure everything is fine so that we will have a goodnight's sleep. :-) 

Also you mentioned that we should use Host::io_context. I did think about that at first but then I realized that we will be sharing it with something and if so will it cause problems like some code is using it for its async ops and we are using it in the middle of it??

twinstar26:
> Also you mentioned that we should use Host::io_context. I did think about that at first but then I realized that we will be sharing it with something and if so will it cause problems like some code is using it for its async ops and we are using it in the middle of it??

Distinct objects: Safe.
Shared objects: Safe, with the specific exceptions of the restart() and notify_fork() functions. Calling restart() while there are unfinished run(), run_one(), run_for(), run_until(), poll() or poll_one() calls results in undefined behaviour. The notify_fork() function should not be called while any io_context function, or any function on an I/O object that is associated with the io_context, is being called in another thread. 
From official docs.
And 
https://github.com/ethereum/aleth/blob/007d89a712209ca16198be41a5d9c70498138eb6/libp2p/Host.cpp#L206-L214
Host do call restart(). So I think we are better off with new io_context. 
Let me know.

libp2p/Session.cpp
gumb0: Yes, timer should better use `io_context` of `Host`.

Not sure about passing timer to constructor. `Session` has a pointer to `Host`, maybe better to introduce a public method in `Host` that will create a timer? This way we won't create it until it's really needed in `disconnect`.

gumb0:
@twinstar26  This `restart()` is called only when `run()` call is finished, so it shouldn't be of any concern for your solution.

We should reuse `Host::m_context` because this way we can be sure that all handlers 
(including your new timer handler) are called from the same thread, and thread-safety problems are avoided.

libp2p/Session.cpp
twinstar26: Alright!!

libp2p/Session.cpp
halfalicious: > Oh that's what the anomaly was in my testing before and after changes. So stupid of me.

@twinstar26 No worries accidents happen 😄

libp2p/Session.cpp
> 
> Yes, timer should better use `io_context` of `Host`.
> 
> Not sure about passing timer to constructor. `Session` has a pointer to `Host`, maybe better to introduce a public method in `Host` that will create a timer? This way we won't create it until it's really needed in `disconnect`.

Good idea, this way we can also easily create timers in other classes if needed.

twinstar26:
@gumb0 
So I have created Host::disconnectTimer() which is public and I want to call Session::drop() inside handler of timer.async_wait(). Session::drop() is private so I cannot call it using object of Session.
So options maybe to convert drop() to public, convert drop() to protected and inherit, use friend, other concept unknown to me. 
So which one to go by??

halfalicious:
> 
> 
> @gumb0
> So I have created Host::disconnectTimer() which is public and I want to call Session::drop() inside handler of timer.async_wait(). Session::drop() is private so I cannot call it using object of Session.
> So options maybe to convert drop() to public, convert drop() to protected and inherit, use friend, other concept unknown to me.
> So which one to go by??

@twinstar26 I would name the Host function which returns a new timer something more general e.g. `Host::createTimer()` since the returned timer doesn't have to be used for managing disconnects.

Also, `Session::drop() `being private shouldn't be an issue, you should be able to call it from within the timer handler inside of `Session::disconnect` e.g. (note that this is just example code and might not compile but you should get the general idea): 

```
auto self = shared_from_this();
auto disconnectTimer = m_host->createTimer();
disconnectTimer->expires_from_now(2000);
disconnectTimer->async_wait([self, this, disconnectTimer](boost::system::error_code -ec){
this->drop();
});
```

You can also take a look at `Host::scheduleCapabilityWorkLook` for an example of how we pass handlers to timers.

twinstar26:
@gumb0 @halfalicious 
When I am using `m_ioContext` as io_context, I am not getting a 2 sec wait and socket is getting closed. But if I use other io_context object with `run()` in `Host::Disconnect()` I am getting desired result.
As @halfalicious pointed out that `Host::dowork()` is where `m_ioContext.run()` is present and it should work out but for some reason it's not.
Also for every timer start, socket closes twice and i still haven't found a reason for this.

This is my log when using m_ioContext except socket closes immediately and doesn't wait for 2 secs.
`Test Case "host": `
`timer starting`
`//////////socket is closed due to reason 8 //////////`
`//////////socket is closed due to reason 0 //////////`
`Test Case "saveNodes": `
`timer starting`
`timer starting`
`timer starting`
`timer starting`
`//////////socket is closed due to reason 0 //////////`
`timer starting`
`//////////socket is closed due to reason 0 //////////`
`//////////socket is closed due to reason 8 //////////`
`//////////socket is closed due to reason 0 //////////`
`//////////socket is closed due to reason 0 //////////`
`//////////socket is closed due to reason 8 //////////`
`//////////socket is closed due to reason 0 //////////`
`//////////socket is closed due to reason 8 //////////`
`//////////socket is closed due to reason 8 //////////`
`//////////socket is closed due to reason 8 //////////`
`timer starting`
`//////////socket is closed due to reason 8 //////////`
`//////////socket is closed due to reason 0 //////////`
`timer starting`
`//////////socket is closed due to reason 8 //////////`
`//////////socket is closed due to reason 0 //////////`
`timer starting`
`//////////socket is closed due to reason 8 //////////`
`//////////socket is closed due to reason 0 //////////`
`timer starting`
`//////////socket is closed due to reason 8 //////////`
`//////////socket is closed due to reason 0 //////////`

halfalicious:
> 
> 
> @gumb0 @halfalicious
> When I am using `m_ioContext` as io_context, I am not getting a 2 sec wait and socket is getting closed. But if I use other io_context object with `run()` in `Host::Disconnect()` I am getting desired result.
> As @halfalicious pointed out that `Host::dowork()` is where `m_ioContext.run()` is present and it should work out but for some reason it's not.
> Also for every timer start, socket closes twice and i still haven't found a reason for this.
> 
> This is my log when using m_ioContext except socket closes immediately and doesn't wait for 2 secs.
> `Test Case "host": `
> `timer starting`
> `//////////socket is closed due to reason 8 //////////`
> `//////////socket is closed due to reason 0 //////////`
> `Test Case "saveNodes": `
> `timer starting`
> `timer starting`
> `timer starting`
> `timer starting`
> `//////////socket is closed due to reason 0 //////////`
> `timer starting`
> `//////////socket is closed due to reason 0 //////////`
> `//////////socket is closed due to reason 8 //////////`
> `//////////socket is closed due to reason 0 //////////`
> `//////////socket is closed due to reason 0 //////////`
> `//////////socket is closed due to reason 8 //////////`
> `//////////socket is closed due to reason 0 //////////`
> `//////////socket is closed due to reason 8 //////////`
> `//////////socket is closed due to reason 8 //////////`
> `//////////socket is closed due to reason 8 //////////`
> `timer starting`
> `//////////socket is closed due to reason 8 //////////`
> `//////////socket is closed due to reason 0 //////////`
> `timer starting`
> `//////////socket is closed due to reason 8 //////////`
> `//////////socket is closed due to reason 0 //////////`
> `timer starting`
> `//////////socket is closed due to reason 8 //////////`
> `//////////socket is closed due to reason 0 //////////`
> `timer starting`
> `//////////socket is closed due to reason 8 //////////`
> `//////////socket is closed due to reason 0 //////////`

Can you please push your latest changes so one of us can take a look and try to repro your issue locally? (I took a look at your branch and it doesn't have the changes you're referring to)

twinstar26:
Oh sorry!
https://github.com/twinstar26/aleth/tree/aleth-test

twinstar26:
@halfalicious  
So I confirmed that our logic (your prototype with corrections) was indeed correct. I also confirmed that `async_wait's` handler was never executed during unit-tests (Actually the handler was exiting with exception stating `Operation failed` which is `boost::asio::error::operation_aborted` and is due to timer being cancelled). Now it makes sense why we thought our solution was not working. `Host::doWork()` has `run()` but due to `m_ioContext` object being `restarted` / `stopped` in a certain way that `async_wait's` handler which was queued was never executed. 

Possible solution might be to create another method inside `Host` which will `run` `m_ioContext's` handlers without interfering with `Host::doWork()`.
But we have a problem as `Run network. Not thread-safe; to be called only by worker.` this is the comment left by creator of `doWork()`. 
@halfalicious @gumb0 please confirm my findings and possible solution that I have suggested. Also if you have better approach, let me know.

twinstar26:
[log_wait_handler_not _executed.txt](https://github.com/ethereum/aleth/files/3406670/log_wait_handler_not._executed.txt)

Log

libp2p/Session.cpp
```diff
@@ -305,8 +305,20 @@ void Session::disconnect(DisconnectReason _reason)
...
+        auto self(shared_from_this());
+        disconnectTimer->expires_from_now(std::chrono::seconds(2));
+        disconnectTimer->async_wait([self, this, _reason](boost::system::error_code _ec) {
+            if(_ec)
+                LOG(m_netLoggerDetail)
+                    << "Error encountered while waiting after disconnect: " << _ec.message();
+            else
+                drop(_reason);
+        });
+        m_server->pollAsyncTimerHandler();
```
gumb0: What happens if you don't do this?

twinstar26: @gumb0 
Generally we poll when we are calling the function to which the handler belongs multiple times (in a recursion or a loop) so that any handlers ready to run will start its execution.
The original motivation was when in a way I knew unit-tests were calling disconnect multiple times which later you confirmed that they create and destroy host multiple times.

But I realized that there are various parts of code that do call disconnect in a loop for eg.
https://github.com/ethereum/aleth/blob/4946ed615b11033db61ec47b2a97ef36bd3c047f/libp2p/Host.cpp#L247-L263
So I kept `pollAsyncTimerHandler` public in `Host` definition of which is here
https://github.com/twinstar26/aleth/blob/e83913941ace68e2b41597ddae7fa5ba8e6c3b6c/libp2p/Host.h#L266-L269

twinstar26: Also I realize that there is a bug that we should be calling `_drop(DisconnectReason)` in here too right?
https://github.com/twinstar26/aleth/blob/e83913941ace68e2b41597ddae7fa5ba8e6c3b6c/libp2p/Session.cpp#L312-L314
So that if timer fails for some reason we will issue a drop. What do you think @gumb0 ?

twinstar26:
ToDo list
1) halfalicious suggested to replace socket->isConnected() with m_dropped (as to prevent wasting resources e.g. continuing to sync with a client that's disconnecting from it) but that's causing multiple unit-tests to fail. They state that `check host.peerCount() == c_peers has failed` (need to research)
2) Need to add changelog entry
3) Logical bug over here (https://github.com/ethereum/aleth/pull/5658#discussion_r305450240)
4) Deprecated steady_timer function
5) Create new unit-test (maybe)
6) Modify existing unit-tests (maybe) (need to research)

libp2p/Session.cpp
halfalicious: I actually don't think you need to call drop() there...I think that the handler will only ever called with a boost error code if the timer is cancelled but you never call cancel on the timer.

halfalicious:
> 
> 
> ToDo list
> 
>     1. halfalicious suggested to replace socket->isConnected() with m_dropped (as to prevent wasting resources e.g. continuing to sync with a client that's disconnecting from it) but that's causing multiple unit-tests to fail. They state that `check host.peerCount() == c_peers has failed` (need to research)
> 

This should be` !m_dropped` - I tested this locally with my prototype and all of the libp2p tests pass (testeth -t libp2p/*)

halfalicious:
I don't think you need to check the error code in the handler, you can just check if the socket is open and if so, close the socket. There's no risk of use-after-free since the handler only accesses Session state and the Session instance is guaranteed to still be available since the handler takes a shared_from_this.

halfalicious:
Good call on the timer handlers not being executed when the timers are cancelled (on shutdown) - I don't think we care about this when Aleth is shutting down since I think that the sockets which are left open will be reclaimed when the process exits, but I do think we care about this when running tests since I think that the same process is used to run all tests.

I think the best way to ensure that the handlers are executed is to modify `Host::doneWorking` rather than adding polling the io context from within `Session` since the role of `doneWorking `is to handle cleaning up Host resources on network shutdown. I think we can just move the `n++` above the "`if (p->isConnected())`" check here: https://github.com/twinstar26/aleth/blob/e83913941ace68e2b41597ddae7fa5ba8e6c3b6c/libp2p/Host.cpp#L247-L257

This would prevent the disconnect peers loop from exiting until all of the `Session ` weak pointers expired, which would happen after all of the disconnect timer handlers execute (the timer handler takes a Session shared_ptr which is what keeps a Session alive - all other Session pointers e.g. the ones in Host are weak_ptr so they don't count towards the Session reference counting).

I've tested this change locally and the handlers are executed in the relevant network tests. The down side of this change is that some of the network tests will take a little longer to run since they wait 2 seconds for the disconnect timers to expire but there aren't many tests which do this so  think that's okay (@gumb0 : please correct me if I'm wrong here)

twinstar26:
> > ToDo list
> > ```
> > 1. halfalicious suggested to replace socket->isConnected() with m_dropped (as to prevent wasting resources e.g. continuing to sync with a client that's disconnecting from it) but that's causing multiple unit-tests to fail. They state that `check host.peerCount() == c_peers has failed` (need to research)
> > ```
> 
> This should be` !m_dropped` - I tested this locally with my prototype and all of the libp2p tests pass (testeth -t libp2p/*)

When we change https://github.com/ethereum/aleth/blob/ece965791c7d5c5ae60d062e62b0bee4ed5d8e5a/libp2p/Session.h#L87 to `    bool isConnected() const override { return m_dropped; }` then the issue is coming.
I initially thought we are using m_dropped check elsewhere and here we are using m_socket check in isConnected so i reverted everything.

gumb0:
@halfalicious Would this result in 2 seconds (or more) wait at aleth exit?

I don't think it's very necessary to wait at exit. And sockets will be anyway closed in `Session` destructor, so in tests it should be fine already.

halfalicious:
> 
> 
> @halfalicious Would this result in 2 seconds (or more) wait at aleth exit?
> 
> I don't think it's very necessary to wait at exit. And sockets will be anyway closed in `Session` destructor, so in tests it should be fine already.

@gumb0 Ah you're right, the socket is closed here: https://github.com/ethereum/aleth/blob/5bf79e54751d36af4f60a6a3d7b6270aae40cab2/libp2p/Session.cpp#L44

In that case I agree, we don't care about the handlers not being executed on Aleth shutdown. No need to modify `Host::doneWorking`.

halfalicious:
> 
> 
> @halfalicious Would this result in 2 seconds (or more) wait at aleth exit?
> 
> I don't think it's very necessary to wait at exit. And sockets will be anyway closed in `Session` destructor, so in tests it should be fine already.

@gumb0 Yes, it would. Thanks for calling out that the sockets are closed in the dtor, @twinstar26 they are closed here: https://github.com/ethereum/aleth/blob/5bf79e54751d36af4f60a6a3d7b6270aae40cab2/libp2p/Session.cpp#L44

We also poll after calling disconnect in `Host::doneWorking` which will ensure that the disconnect packets are actually sent out on shutdown. As such, I don't think any modifications to `Host::doneWorking` are necessary.

twinstar26:
Thanks for the reviews and suggestions!! @gumb0 and @halfalicious 

So to conclude:
1) No need of `m_server->pollAsyncTimerHandler()` polling function
2) No need of modifying `Host::doneWorking` and calling `Host::doneWorking` inside disconnect.
3) No need of error code inside time handler.
4) Implement this todo list (https://github.com/ethereum/aleth/pull/5658#issuecomment-513431898)

After these changes aleth will not wait for 2 secs during tests but actual working of aleth is quite different and tests don't actually simulate that.

halfalicious:
@twinstar26 Just checking in to see if you're still working on this?

twinstar26:
> @twinstar26 Just checking in to see if you're still working on this?

@halfalicious . Sorry! Yes I am working. I was busy with my college stuff. I will push a commit in 3-4 hours from now.

halfalicious:
> 
> 
> > @twinstar26 Just checking in to see if you're still working on this?
> 
> @halfalicious . Sorry! Yes I am working. I was busy with my college stuff. I will push a commit in 3-4 hours from now.

No worries, just wanted to see if you were still interested in wrapping this up 😄 I've reviewed your new changes, left comments in the new PR.

gumb0:
Replaced by https://github.com/ethereum/aleth/pull/5707

PR #5659, opened 2019-07-09
Remove devcore from aleth-interpreter dependencies

chfast:
This should make building aleth-interpreter only much easier. We will not need to build devcore library but also probably we will not need to link boost_system.

### DIFF
--- a/libaleth-interpreter/CMakeLists.txt
+++ b/libaleth-interpreter/CMakeLists.txt
@@ -11,13 +11,23 @@ set(
     VMConfig.h
     VMOpt.cpp
 )
+
+set(
+    dependencies
+    aleth-buildinfo
+    ethash::ethash
+    evmc::evmc
+    evmc::instructions
+    Boost::boost
+)
+
 add_library(aleth-interpreter STATIC ${sources})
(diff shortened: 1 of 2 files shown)


libaleth-interpreter/CMakeLists.txt
```diff
@@ -11,13 +11,23 @@ set(
...
     VMOpt.cpp
 )
+
+set(
+    dependencies
+    aleth-buildinfo
+    ethash::ethash
+    evmc::evmc
+    evmc::instructions
+    Boost::boost
```
gumb0: We link all of the boost libraries?

chfast: The `Boost::boost` represents header-only Boost libs. We need it for u256, but mostly for Windows build.

gumb0: How come header-only libs need to be linked?

chfast: This is modern CMake magic. The CMake targets contains other information that is going to be propagated to the targets linking them, e.g. compile flags, include dir, etc.

In this case this will set `-DBOOST_ALL_NO_LIB=1` for MSVC builds. Otherwise we will have linking issues in testeth :/

PR #5660, opened 2019-07-09
Upgrade Cable to 0.4.2

chfast:


gumb0:
There's also https://github.com/ethereum/aleth/blob/007d89a712209ca16198be41a5d9c70498138eb6/test/tools/libtesteth/TestHelper.cpp#L30
Should it be updated?

chfast:
> There's also
> 
> https://github.com/ethereum/aleth/blob/007d89a712209ca16198be41a5d9c70498138eb6/test/tools/libtesteth/TestHelper.cpp#L30
>
> Should it be updated?

Rather not. I added the `version.h` to make aleth-interpreter "standalone". Now it do not need to be shipped with aleth-buildinfo.a. For other cases it is better to link the aleth-buildinfo - you don't have to rebuild if the version changes.

PR #5661, opened 2019-07-09
detect folders with no tests when subfolders depth is more then 1

winsvega:


test/tools/libtesteth/TestOutputHelper.cpp
```diff
@@ -129,30 +129,40 @@ void TestOutputHelper::checkUnfinishedTestFolders()
...
             return;
 
         std::map<boost::filesystem::path, FolderNameSet>::const_iterator it =
             m_finishedTestFoldersMap.begin();
-        if (!pathHasTests(it->first / filter))
-            std::cerr << "WARNING: Test folder " << it->first / filter
-                      << " appears to have no tests!"
-                      << "\n";
+        if (!boost::filesystem::exists(it->first / filter))
+            notAFilter = true;
```
gumb0: Is it the case when `-t` points to a unit test?

gumb0: Ah no, for unit tests it should return from line 141.
When this is the case then?

winsvega: it is the case when you select  `-t TestSuite / SubTestSuite`

gumb0: It looks like you could move this check to the outer `if` (line 137)
Kind of like
```
if (!filter.empty() || boost::filesystem::exists(m_finishedTestFoldersMap.begin()->first / filter))
{
...
}
else
{
...
}
```
(but check that `m_finishedTestFoldersMap` is not empty first)

Would it be simpler?

gumb0: Also maybe it would easier/more efficient to look at `c_allTestNames` to check whether it's single folder or subsuite, but I'm not sure

winsvega: if you want to call c_allTestNames more then from one place, will have to intoduce c_allTestNames.cpp

PR #5662, opened 2019-07-09
interpreter: Fix call depth handling in CALLs

chfast:
Fixes an issue in aleth-interpreter where for call the depth is not increased.
The additional assert now checks if the depth is the same as in ExtVM (will cause a lot of state test failures if the fix is not applied).

### DIFF
--- a/libaleth-interpreter/VMCalls.cpp
+++ b/libaleth-interpreter/VMCalls.cpp
@@ -258,6 +258,7 @@ bool VM::caseCallSetup(evmc_message& o_msg, bytesRef& o_output)
     m_runGas = o_msg.gas;
     updateIOGas();
 
+    o_msg.depth = m_message->depth + 1;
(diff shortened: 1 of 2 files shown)


gumb0:
I'm a bit confused about this
https://github.com/ethereum/aleth/blob/d169557f3eff908d4d77d27519413f4564ed132e/libethereum/ExtVM.cpp#L128
`EvmCHost::call` delegates to `ExtVM::call` in the end, and there depth is incremented. Wouldn't it be then incremented twice with your fix?

chfast:
`ExtVM` has it's own depth counter. We just check if both agree.

gumb0:
Ok, I think I see. I guess it was working fine, because `ExtVM` has was correctly counting anyway.

chfast:
>  I guess it was working fine, because `ExtVM` was correctly counting anyway.

Exactly.

PR #5663, opened 2019-07-10
Coming to agreement with geth on RIPEMD touch revert - alternative solution

gumb0:
Alternative solution to https://github.com/ethereum/aleth/pull/5652

PR #5664, opened 2019-07-10
Coming to agreement with geth on RIPEMD touch revert - another alternative solution

gumb0:
One more alternative to #5652 and #5663


gumb0:
Updated tests and added changelog item.

PR #5666, opened 2019-07-11
Distinguish between INVALID opcode and unknown opcodes in aleth-interpreter

gumb0:
This fixes the failure of `evm.invalid` test, which expects `EVMC_INVALID_INSTRUCTION` to be returned for `INVALID` (`0xfe`) opcode and `EVMC_UNKNOWN_INSTRUCTION` for all not defined opcodes.
```
[ RUN      ] evm.invalid
/home/andrei/dev/evmone/test/unittests/evm_test.cpp:668: Failure
Expected equality of these values:
  result.status_code
    Which is: 5
  EVMC_INVALID_INSTRUCTION
    Which is: 4
[  FAILED  ] evm.invalid (0 ms)
```

### DIFF
--- a/libaleth-interpreter/VM.cpp
+++ b/libaleth-interpreter/VM.cpp
@@ -59,6 +59,10 @@ evmc_result execute(evmc_instance* _instance, evmc_context* _context, evmc_revis
         result.gas_left = vm->m_io_gas;
         output = ex.output();  // This moves the output from the exception!
     }
+    catch (dev::eth::InvalidInstruction const&)
+    {
+        result.status_code = EVMC_INVALID_INSTRUCTION;
+    }
     catch (dev::eth::BadInstruction const&)
     {
         result.status_code = EVMC_UNDEFINED_INSTRUCTION;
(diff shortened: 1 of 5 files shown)


libaleth-interpreter/VM.cpp
```diff
@@ -1416,7 +1420,10 @@ void VM::interpretCases()
         CASE(INVALID)
         DEFAULT
         {
-            throwBadInstruction();
+            if (m_OP == Instruction::INVALID)
```
chfast: Shouln't it be just

```
CASE(INVALID)
  throwInvalidInstruction();
DEFAULT
  throwUndefinedInstruction();
```

gumb0: No, this doesn't work on gcc/clang, because the jump table is populated with `INVALID` opcode for all unknown ones https://github.com/ethereum/aleth/blob/089e7ddaa741825fdf62974ad66b077c715b8f25/libaleth-interpreter/VMConfig.h#L154

chfast: Can't we change that by having `&&UNDEFINED` everywhere and `&&INVALID` in https://github.com/ethereum/aleth/blob/089e7ddaa741825fdf62974ad66b077c715b8f25/libaleth-interpreter/VMConfig.h#L409?

gumb0: There's no such thing as `UNDEFINED`, and this enum is 8 bit-wide https://github.com/ethereum/aleth/blob/2b11ce5c5aaf00f54ce9a301f8c103148d00abdc/libevm/Instruction.h#L28

So we can't really add `UNDEFINED` unless we change the width

PR #5667, opened 2019-07-11
fix suggestion output on failing tests

winsvega:


test/tools/libtesteth/boostTest.cpp
```diff
@@ -147,14 +147,19 @@ int main(int argc, const char* argv[])
         }
     }
 
+    // Print suggestions of a test case if test suite not found
+    if (!sMinusTArg.empty() && !dev::test::inArray(c_allTestNames, sMinusTArg))
+    {
+        std::cerr << "Error: '" + sMinusTArg + "' suite not found! \n";
+        printTestSuiteSuggestions(sMinusTArg);
+        return 0;
```
gumb0: It should better return non-zero in this case (maybe 1 would do)

PR #5668, opened 2019-07-12
Add changelog item for BlockChainTests split

gumb0:


gumb0:
We have there already something about moving some tests into "time consuming" suite, so I've decided this is similar.
Anyway I think it's visible to testeth users, so should be mentioned.

PR #5672, opened 2019-07-15
testeth: compress large state data into hash

winsvega:
do not print a megabyte long states to the tests. use hashing instead.

### DIFF
--- a/test/tools/jsontests/BlockChainTests.cpp
+++ b/test/tools/jsontests/BlockChainTests.cpp
@@ -275,6 +275,21 @@ void ChainBranch::resetBlockchain()
     dev::test::TestBlockChain::s_sealEngineNetwork = s_tempBlockchainNetwork;
 }
 
+bool isSmallState(State const& _state)
+{
+    if (_state.addresses().size() > 20)
+        return false;
+    size_t totalSize = 0;
+    for (auto const& accAddr : _state.addresses())
+    {
+        totalSize += _state.codeSize(accAddr.first) * 2;         // each byte represented by 2 chars
+        totalSize += _state.storage(accAddr.first).size() * 64;  // potentially expensive operation
+        if (totalSize > 1024000)                                 // 1MB
+            return false;
+    }
+    return true;
+}
+
 json_spirit::mObject fillBCTest(json_spirit::mObject const& _input, bool _allowInvalidBlocks)
 {
     json_spirit::mObject output;
@@ -471,7 +486,11 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input, bool _allowI
     }
 
     output["blocks"] = blArray;
-    output["postState"] = fillJsonWithState(testChain.topBlock().state());
+
+    if (Options::get().fullstate || isSmallState(testChain.topBlock().state()))
+        output["postState"] = fillJsonWithState(testChain.topBlock().state());
+    else
+        output["postState"] = toHexPrefixed(testChain.topBlock().blockHeader().stateRoot());
     output["lastblockhash"] = toHexPrefixed(testChain.topBlock().blockHeader().hash(WithSeal));
 
     //make all values hex in pre section
@@ -628,9 +647,24 @@ void testBCTest(json_spirit::mObject const& _o)
 
     State postState(State::Null); //Compare post states
     BOOST_REQUIRE((_o.count("postState") > 0));
-    ImportTest::importState(_o.at("postState").get_obj(), postState);
-    ImportTest::compareStates(postState, testChain.topBlock().state());
-    ImportTest::compareStates(postState, blockchain.topBlock().state());
+    if (_o.at("postState").type() == json_spirit::Value_type::obj_type)
+    {
+        ImportTest::importState(_o.at("postState").get_obj(), postState);
+        ImportTest::compareStates(postState, testChain.topBlock().state());
+        ImportTest::compareStates(postState, blockchain.topBlock().state());
+    }
+    else
+    {
+        // Attempt to read hash instead of the full state
+        BOOST_REQUIRE(_o.at("postState").type() == json_spirit::Value_type::str_type);
+        auto postHash = h256{_o.at("postState").get_str()};
+        string message = testName +
+                         "Final StateRoot is different! (Look at expect section in the test /src "
+                         "Filler file, or run the test generation with --fullstate flag to get the "
+                         "full post state.";
+        BOOST_CHECK_MESSAGE(postHash == testChain.topBlock().state().rootHash(), message);
+        BOOST_CHECK_MESSAGE(postHash == blockchain.topBlock().state().rootHash(), message);
+    }
 }
 
 bigint calculateMiningReward(u256 const& _blNumber, u256 const& _unNumber1, u256 const& _unNumber2, SealEngineFace const& _sealEngine)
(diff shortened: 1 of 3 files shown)


winsvega:
hm actually might be better to calculate accounts codesize and accounts count in the states

gumb0:
Maybe it makes sense to put the hash into different field (`postStateHash`) and have it always present? So that it's either `postState` and `postStateHash` or only `postStateHash`

test/tools/libtesteth/Options.h
```diff
@@ -44,7 +44,7 @@ class Options
     bool exectimelog = false; ///< Print execution time for each test suite
     std::string rCurrentTestSuite; ///< Remember test suite before boost overwrite (for random tests)
     bool statediff = false;///< Fill full post state in General tests
-    bool fulloutput = false;///< Replace large output to just it's length
+    bool fullstate = false;///< Replace large state output to just it's length or hash
```
gumb0: it's never length, right?

winsvega: right. it used to be. but perhaps hash is better

gumb0: If block has incorrect uncles, it could be consensus issue with equal state roots.

Anyway, here delete the mention of length, if it's not true anymore.

gumb0:
Actually, maybe it's better to use the block hash instead of just state root hash? It would cover other things like uncles and receipts

winsvega:
Then it is already implemented in lastblockhash

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -471,7 +471,23 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input, bool _allowI
...
     output["blocks"] = blArray;
-    output["postState"] = fillJsonWithState(testChain.topBlock().state());
+
+    size_t totalCodeSize = 0;
+    bool isSmallState = testChain.topBlock().state().addresses().size() < 20;
+    if (isSmallState)
+    {
+        for (auto const& accAddr : testChain.topBlock().state().addresses())
+        {
+            totalCodeSize += testChain.topBlock().state().codeSize(accAddr.first);
```
gumb0: Why only code? Storage could be huge, too (I don't know if it's worth to consider in tests)

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -471,7 +471,23 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input, bool _allowI
...
+            if (totalCodeSize > 1024 * 1000)  // if totalCodeSize > 1Mb
+                isSmallState = false;
```
gumb0: You can `break` here

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -471,7 +471,23 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input, bool _allowI
     }
 
     output["blocks"] = blArray;
-    output["postState"] = fillJsonWithState(testChain.topBlock().state());
+
+    size_t totalCodeSize = 0;
+    bool isSmallState = testChain.topBlock().state().addresses().size() < 20;
```
gumb0: Cleaner way than setting the flag to true, then possibly setting it to false would be to move it all to separate function `isSmallState(testChain.topBlock().state())`
(with early return like
```
if (testChain.topBlock().state().addresses().size() >= 20)
  return false;

for (...)
{
  ...
  if (totalCodeSize > 1024 * 1000)
    return false;
}
return true;
```

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -471,7 +471,23 @@ json_spirit::mObject fillBCTest(json_spirit::mObject const& _input, bool _allowI
...
+    {
+        for (auto const& accAddr : testChain.topBlock().state().addresses())
+        {
+            totalCodeSize += testChain.topBlock().state().codeSize(accAddr.first);
+            if (totalCodeSize > 1024 * 1000)  // if totalCodeSize > 1Mb
+                isSmallState = false;
+        }
+    }
+
+    if (isSmallState || Options::get().fullstate)
```
gumb0: Check `fullstate` first, if it's true, it won't need to iterate over 20 addresses

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -628,9 +644,21 @@ void testBCTest(json_spirit::mObject const& _o)
...
+        ImportTest::importState(_o.at("postState").get_obj(), postState);
+        ImportTest::compareStates(postState, testChain.topBlock().state());
+        ImportTest::compareStates(postState, blockchain.topBlock().state());
+    }
+    else
+    {
+        // Attempt to read hash instead of the full state
+        BOOST_REQUIRE(_o.at("postState").type() == json_spirit::Value_type::str_type);
+        string postHash = _o.at("postState").get_str();
+        string message = testName + "Final StateRoot is different! (Look at expect section in the test /src Filler file, or run the test generation with --fullstate flag to geth the full post state.";
```
gumb0:
```suggestion
        string message = testName + "Final StateRoot is different! (Look at expect section in the test /src Filler file, or run the test generation with --fullstate flag to get the full post state.";
```

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -628,9 +644,21 @@ void testBCTest(json_spirit::mObject const& _o)
...
+        BOOST_CHECK_MESSAGE(postHash == toHexPrefixed(testChain.topBlock().state().rootHash()), message);
```
gumb0: Probably better the other way around `h256{postHash} == testChain.topBlock().state().rootHash()`
(and do conversion to `h256` only once for these two checks)

test/tools/jsontests/BlockChainTests.cpp
winsvega: hm. you right. although it is likely that a huge code is deployed rather then storage. but that one needs to be checked too.

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -275,6 +275,21 @@ void ChainBranch::resetBlockchain()
     dev::test::TestBlockChain::s_sealEngineNetwork = s_tempBlockchainNetwork;
 }
 
+bool isSmallState(State const& _state)
+{
+    if (_state.addresses().size() > 20)
+        return false;
+    size_t totalSize = 0;
+    for(auto const& accAddr: _state.addresses())
```
gumb0: not clang-formatted

test/tools/jsontests/BlockChainTests.cpp
```diff
@@ -628,9 +647,21 @@ void testBCTest(json_spirit::mObject const& _o)
...
+    {
+        ImportTest::importState(_o.at("postState").get_obj(), postState);
+        ImportTest::compareStates(postState, testChain.topBlock().state());
+        ImportTest::compareStates(postState, blockchain.topBlock().state());
+    }
+    else
+    {
+        // Attempt to read hash instead of the full state
+        BOOST_REQUIRE(_o.at("postState").type() == json_spirit::Value_type::str_type);
+        h256 postHash = h256{_o.at("postState").get_str()};
```
gumb0: could be `auto`, don't repeat the type

PR #5675, opened 2019-07-17
Initiate p2p disconnect when all capabilities are disabled

### LINKED ISSUE #5674: Aleth doesn't disconnect peer session when all capabilities are disabled
Aleth disables the "eth" capability for a peer when certain conditions occur e.g. Aleth receives a new block from a peer which is malformed. However, the network session is kept alive despite the capability being disabled (note that Aleth only supports the eth capability so disabling it means the connected peer is useless) since `Session`'s `doRead `loop (which performs an async read from a socket) passes a `shared_ptr ` instance of the `Session `to the `async_read `handler, which means that the Session will be kept alive until the peer disconnects. Note that `Host `also pings peer sessions periodically to keep them alive (see `Host::keepAlivePeers`), so it's possible that a session could remain connected a long time after all capabilities have been disabled on it.

halfalicious:
Fix #5674 

Send out a p2p capability disconnect packet when all capabilities are disabled. Also log the outgoing Disconnect p2p capability packet to the `p2pcap `logger rather than the `net ` logger.


halfalicious:
Rebased to fix changelog conflict

PR #5679, opened 2019-07-17
Use metrics of the requested revision in aleth-interpreter

gumb0:
Fixes the problem described in https://github.com/ethereum/aleth/issues/5653#issuecomment-510605376, makes this change in test not needed https://github.com/ethereum/evmone/pull/91

~~I came up here with (maybe a bit tricky) way to lazy-initialize metrics array for needed revision using `std::once_flag`. But I think it looks not so complicated.~~ Settled upon a simpler non-lazy approach.

### DIFF
--- a/libaleth-interpreter/VM.cpp
+++ b/libaleth-interpreter/VM.cpp
@@ -244,7 +244,7 @@ void VM::logGasMem()
 void VM::fetchInstruction()
 {
     m_OP = Instruction(m_code[m_PC]);
-    auto const metric = c_metrics[static_cast<size_t>(m_OP)];
+    auto const metric = metrics(m_rev)[static_cast<size_t>(m_OP)];
     adjustStack(metric.num_stack_arguments, metric.num_stack_returned_items);
 
     // FEES...
--- a/libaleth-interpreter/VM.cpp
+++ b/libaleth-interpreter/VM.cpp
@@ -247,7 +247,7 @@ void VM::logGasMem()
 void VM::fetchInstruction()
 {
     m_OP = Instruction(m_code[m_PC]);
-    auto const metric = metrics(m_rev)[static_cast<size_t>(m_OP)];
+    auto const metric = (*m_metrics)[static_cast<size_t>(m_OP)];
     adjustStack(metric.num_stack_arguments, metric.num_stack_returned_items);
 
     // FEES...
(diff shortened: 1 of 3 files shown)


libaleth-interpreter/VMOpt.cpp
```diff
@@ -21,22 +21,23 @@ namespace dev
...
+std::array<std::once_flag, EVMC_MAX_REVISION> VM::s_metricsInitialized;
+
+void VM::initMetrics(evmc_revision _revision)
 {
-    static bool done = []() noexcept
-    {
-        // Copy the metrics of the top EVM revision.
-        std::memcpy(&c_metrics[0], evmc_get_instruction_metrics_table(EVMC_MAX_REVISION),
-            c_metrics.size() * sizeof(c_metrics[0]));
+    std::call_once(s_metricsInitialized[_revision], [_revision]() {
```
gumb0: `call_once` is needed to make initialization thread-safe, and we can't use the trick of static initialized by lambda as before, because now we have array of metric arrays (and we want to lazily initialize only the requested revision)

chfast: I would do it once in https://github.com/ethereum/aleth/blob/master/libaleth-interpreter/VM.cpp#L117.

gumb0: Yeah, it would be simpler if it's for all possible revisions at one. We could also do it here as before, in static initialized by lambda call. But just wanted to avoid this unnecessary overhead.
I think I'll try that if I don't solve the GCC issue here.

gumb0:
Now GCC issue disappeared, so I think this is ready for review.

libaleth-interpreter/VM.cpp
```diff
@@ -244,7 +244,7 @@ void VM::logGasMem()
 void VM::fetchInstruction()
 {
     m_OP = Instruction(m_code[m_PC]);
-    auto const metric = c_metrics[static_cast<size_t>(m_OP)];
+    auto const metric = metrics(m_rev)[static_cast<size_t>(m_OP)];
```
chfast: Because `m_rev` never changes during execution, you should get the right table during execution initialization.

libaleth-interpreter/VMOpt.cpp
chfast: I think this is nice and generic design, but if done not lazily in `evmc_create_interpreter` we would avoid dependency on atomics and locks. I expect this to be also additional runtime dependency so you will need to link with more symbols from standard library.

I was also thinking if any of the tables are the same, but this is unlikely and we should probably not optimize for this case.

gumb0:
Rebased.

PR #5683, opened 2019-07-18
Updated deprecated functions of steady_timer

### LINKED ISSUE #5669: Boost::steady_timer deprecated functions.
Boost::steady_timer has deprecated functions like `expires_from_now()` and specific overloads of `expires_at()` and `cancel()`.
According to boost official docs,
1) `expires_from_now()` should be changed to `expires_after()`
2) if `time_point expires_at() const` and `std::size_t expires_at(const time_point & expiry_time, boost::system::error_code & ec)`  exist, should be changed to `std::size_t expires_at(const time_point & expiry_time)` i.e. non-error_code overload.
3) if `std::size_t cancel(boost::system::error_code & ec)` exist, should be changed to `std::size_t cancel()` i.e. non-error_code overload.

For more reference:-
[https://www.boost.org/doc/libs/1_70_0/doc/html/boost_asio/reference/steady_timer.html
](url)

twinstar26:
Fixes #5669

### DIFF
--- a/libp2p/Host.cpp
+++ b/libp2p/Host.cpp
@@ -169,7 +169,7 @@ void Host::startCapabilities()
 
 void Host::scheduleCapabilityWorkLoop(CapabilityFace& _cap, shared_ptr<ba::steady_timer> _timer)
 {
-    _timer->expires_from_now(_cap.backgroundWorkInterval());
+    _timer->expires_after(_cap.backgroundWorkInterval());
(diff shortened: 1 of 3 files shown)


twinstar26:
One more signature of deprecated function of steady_timer exist in #5658 which I will update in next force push of 5658.

halfalicious:
@twinstar26 Thanks for your contribution! 😀

PR #5687, opened 2019-07-22
Only allow up to 1024 dropped transactions in the transaction queue

### LINKED ISSUE #5688: Aleth never clears transaction queue's "dropped transactions" list
Transactions which are dropped during transaction queue import are stored in `TransactionQueue::m_dropped`. A max size isn't enforced on `m_dropped `and `m_dropped `is only cleared when the entire Transaction Queue is cleared (`TransactionQueue::clear`), which doesn't happen during normal execution.

halfalicious:
Fix #5688

### DIFF
--- a/libdevcore/LruCache.h
+++ b/libdevcore/LruCache.h
@@ -12,45 +12,13 @@ namespace dev
 template <class Key, class Value>
 class LruCache
 {
-    typedef Key key_type;
-    typedef Value value_type;
-    typedef std::list<std::pair<key_type, value_type>> list_type;
-    typedef std::unordered_map<key_type, typename list_type::const_iterator> map_type;
+    using key_type = Key;
+    using value_type = Value;
+    using list_type = std::list<std::pair<key_type, value_type>>;
+    using map_type = std::unordered_map<key_type, typename list_type::const_iterator>;
 
 public:
     explicit LruCache(size_t _capacity) : m_capacity(_capacity) {}
-
-    LruCache(LruCache<key_type, value_type> const& _l)
-    {
-        m_capacity = _l.m_capacity;
-        m_data = _l.m_data;
-        m_index = _l.m_index;
-    }
-
-    LruCache(LruCache<key_type, value_type>&& _l)
-    {
-        m_capacity = _l.m_capacity;
-        m_data = std::move(_l.m_data);
-        m_index = std::move(_l.m_index);
-    }
-
-    LruCache<key_type, value_type>& operator=(LruCache const& _l)
-    {
-        m_capacity = _l.m_capacity;
-        m_data = _l.m_data;
-        m_index = _l.m_index;
-
-        return *this;
-    }
-
-    LruCache<key_type, value_type>& operator=(LruCache&& _l)
-    {
-        m_capacity = _l.m_capacity;
-        m_data = std::move(_l.m_data);
-        m_index = std::move(_l.m_index);
-
-        return *this;
-    }
 
     size_t insert(key_type const& _key, value_type const& _val)
     {
@@ -66,9 +34,7 @@ public:
             m_index[_key] = m_data.begin();
         }
         else
-        {
             m_data.splice(m_data.begin(), m_data, cIter->second);
-        }
 
         return m_index.size();
     }
--- a/libdevcore/LruCache.h
+++ b/libdevcore/LruCache.h
@@ -1,0 +1,130 @@
+// Aleth: Ethereum C++ client, tools and libraries.
+// Copyright 2019 Aleth Authors.
+// Licensed under the GNU General Public License, Version 3.
+
+#pragma once
+
+#include <list>
+#include <unordered_map>
+
+namespace dev
+{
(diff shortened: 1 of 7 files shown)


halfalicious:
clang6 failures aren't due to my changes (fairly certain I've seen these failures before), will file a new issue:

```
Test Case "saveNodes": 
/home/builder/project/test/unittests/libp2p/peer.cpp(188): error: in "libp2p/p2p/saveNodes": check host.peerCount() == c_peers has failed [0 != 5]
/home/builder/project/test/unittests/libp2p/peer.cpp(189): error: in "libp2p/p2p/saveNodes": check host2.peerCount() == c_peers has failed [4 != 5]
/home/builder/project/test/unittests/libp2p/peer.cpp(215): fatal error: in "libp2p/p2p/saveNodes": critical check r[2].itemCount() >= c_nodes has failed

*** 3 failures are detected (5 failures are expected) in the test module "Master Test Suite"
```

libethereum/TransactionQueue.cpp
```diff
@@ -328,6 +312,14 @@ void TransactionQueue::drop(h256 const& _txHash)
         return;
 
     UpgradeGuard ul(l);
+    if (m_dropped.size() == c_maxDroppedTransactionCount)
+    {
+        LOG(m_loggerDetail) << "Dropped transaction list is at capacity ("
+                            << c_maxDroppedTransactionCount
+                            << "), removing dropped transaction from list (txhash: "
+                            << *m_dropped.begin() << ")";
+        m_dropped.erase(m_dropped.begin());
```
gumb0: This way it removes random transaction. As @chfast pointed out in the issue, it would be nicer to have some kind of LRU cache here, what do you think?

halfalicious: > 
> 
> This way it removes random transaction. As @chfast pointed out in the issue, it would be nicer to have some kind of LRU cache here, what do you think?

Agreed, didn't go with that in the first place since it increases the storage requirements but I'll go ahead and add that in.

chfast: I don't care so much any more. I think it is ok to merge it as it is and fill additional issue.

gumb0: I'm fine if we merge this as is, too

halfalicious:
cc @chfast  / @gumb0 : Ready for review

libethereum/TransactionQueue.cpp
```diff
@@ -70,7 +58,7 @@ ImportResult TransactionQueue::check_WITH_LOCK(h256 const& _h, IfDropped _ik)
     if (m_known.count(_h))
         return ImportResult::AlreadyKnown;
 
-    if (m_dropped.count(_h) && _ik == IfDropped::Ignore)
+    if (m_dropped.contains(_h) && _ik == IfDropped::Ignore)
```
gumb0: I think tx should also be touched here, if it's found.

libdevcore/LruCache.h
```diff
@@ -0,0 +1,128 @@
...
+
+#include <list>
+#include <unordered_map>
+
+namespace dev
+{
+template <class Key, class Value>
+class LruCache
+{
+    typedef Key key_type;
```
gumb0: Better to use `using key_type = Key;` instead of `typedef`

libdevcore/LruCache.h
```diff
@@ -0,0 +1,128 @@
...
+    typedef Value value_type;
+    typedef std::list<std::pair<key_type, value_type>> list_type;
+    typedef std::unordered_map<key_type, typename list_type::const_iterator> map_type;
+
+public:
+    explicit LruCache(size_t _capacity) : m_capacity(_capacity) {}
+
+    LruCache(LruCache<key_type, value_type> const& _l)
```
gumb0: I think default-generated copy and move constructors and assignment operators should be enough.

libdevcore/LruCache.h
halfalicious: Agreed, much more intuitive / readable

halfalicious:
Rebased to address CHANGELOG merge conflict

halfalicious:
Squashed some commits

test/unittests/libdevcore/LruCache.cpp
```diff
@@ -0,0 +1,109 @@
...
+}
+
+void VerifyEquals(LRU& _lruCache, VEC& _data)
+{
+    EXPECT_EQ(_lruCache.size(), _data.size());
+    size_t i = 0;
+    auto iter = _lruCache.begin();
+    while (iter != _lruCache.cend() && i < _data.size())
+    {
+        EXPECT_EQ(iter->first, _data[i].first);
```
gumb0: Probably `EXPECT_EQ(*iter, _data[i])` should work

halfalicious:
> 
> 
> Looks good to me, one minor thing: you could make `empty()`, `size()`, `capacity()`, `clear()` and iterator getters `noexcept`.

@gumb0: Thanks for calling this out

PR #5691, opened 2019-07-22
EIP-2028: Set tx data nonzero byte cost to 16 in Istanbul

gumb0:
Closes https://github.com/ethereum/aleth/issues/5690
https://eips.ethereum.org/EIPS/eip-2028

### DIFF
--- a/libethcore/EVMSchedule.h
+++ b/libethcore/EVMSchedule.h
@@ -150,6 +150,7 @@ static const EVMSchedule ConstantinopleFixSchedule = [] {
 static const EVMSchedule IstanbulSchedule = [] {
     EVMSchedule schedule = ConstantinopleFixSchedule;
     schedule.accountVersion = 1;
+    schedule.txDataNonZeroGas = 16;
     return schedule;
 }();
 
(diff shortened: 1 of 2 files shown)


chfast:
Please reference this change in the EIP.

gumb0:
Posted a link on ethmagicians discussion thread

holiman:
Regarding implementing this in geth, we see a minor snag with transaction pools -- if they start accepting istanbul-priced txs too early, that's a DoS vector. So we're probably making the pools stateful, and when istanbul hits we flip the switch to use another formula to calculate the intrinsic gas (and thus validity of) a transaction. 
I see you did not touch anything in that area -- is that not a concern for aleth ?

gumb0:
It looks like we don't check intrinsic gas at all when adding to the transaction pool.

@holiman Could you point me to the the area of geth code where this happens?

holiman:
See https://github.com/ethereum/go-ethereum/blob/b80c840af33aae1ff99ce9879fecc6161a0c1620/core/tx_pool.go#L543, which uses the method here: https://github.com/ethereum/go-ethereum/blob/b80c840af33aae1ff99ce9879fecc6161a0c1620/core/state_transition.go#L79

PR #5694, opened 2019-07-22
fix allTestNames construction for test suites

winsvega:
correctly put test suites to AllTestNames.h


test/CMakeLists.txt
```diff
@@ -97,13 +95,13 @@ foreach(file ${sources})
                     add_test(NAME ${TestSuitePathFixed}/${TestCase} WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/test COMMAND testeth ${TestEthArgs})
                     set_tests_properties(${TestSuitePathFixed}/${TestCase} PROPERTIES TIMEOUT ${timeout})
                     set(allTests "${allTests}   \"${TestSuitePathFixed}/${TestCase}\",\n")
-                    if (NOT ";${allSuites};" MATCHES ";${TestSuitePathFixed};")
```
gumb0: So `allTests` now don't include individual tests, only suites?

winsvega: at line 97 it has individual test case added
line 100 - 103 was a wrong way of adding suites to the allTests

PR #5697, opened 2019-07-23
Output ENR text representation in admin.nodeInfo RPC

### LINKED ISSUE #5622: Output ENR text representation in admin.nodeInfo RPC
See https://github.com/ethereum/EIPs/blob/master/EIPS/eip-778.md#text-encoding for the relevant part of ENR spec.

1. Adapt base64 encoding/decoding functions to support URL-safe encoding (basically use `-` instead of `+`, `_` instead of `/` and don't pad with `=`)
https://github.com/ethereum/aleth/blob/master/libdevcore/Base64.h

2. Add method to `ENR` class returning string with text form of ENR (serilize to RLP and base64-encode it)

3. Add `enr` field in the output of `admin_net_nodeInfo` https://github.com/ethereum/aleth/blob/6890d9a8548008419dd4bc054ba7908eac3a933f/libweb3jsonrpc/AdminNet.cpp#L27
You'll need to extend `dev::p2p::NodeInfo` struct for that and fill it in the method `Host::nodeInfo()` with `Host::enr()`.

twinstar26:
Fixes #5622

### DIFF
--- a/libdevcore/Base64.h
+++ b/libdevcore/Base64.h
@@ -35,6 +35,7 @@ namespace dev
 {
 
 std::string toBase64(bytesConstRef _in);
+std::string toBase64url(bytesConstRef _in);
 bytes fromBase64(std::string const& _in);
 
 }
--- a/libp2p/ENR.h
+++ b/libp2p/ENR.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <libdevcore/Base64.h>
 #include "Common.h"
 
 namespace dev
@@ -36,6 +37,7 @@ public:
     /// Create with given sign function
     ENR(uint64_t _seq, std::map<std::string, bytes> const& _keyValuePairs,
         SignFunction const& _signFunction);
+    ENR() = default;
 
     uint64_t sequenceNumber() const { return m_seq; }
     bytes const& signature() const { return m_signature; }
@@ -56,6 +58,8 @@ public:
     ENR update(
         std::map<std::string, bytes> const& _keyValuePair, SignFunction const& _signFunction) const;
 
+    std::string encodeEnr();
+
 private:
     uint64_t m_seq = 0;
     std::map<std::string, bytes> m_keyValuePairs;
@@ -64,6 +68,7 @@ private:
     bytes content() const;
     size_t contentRlpListItemCount() const { return m_keyValuePairs.size() * 2 + 1; }
     void streamContent(RLPStream& _s) const;
+    Host* h;
 };
 
 class IdentitySchemeV4
--- a/libp2p/Host.h
+++ b/libp2p/Host.h
@@ -89,8 +89,8 @@ private:
 struct NodeInfo
 {
     NodeInfo() = default;
-    NodeInfo(NodeID const& _id, std::string const& _address, unsigned _port, std::string const& _version):
-        id(_id), address(_address), port(_port), version(_version) {}
+    NodeInfo(NodeID const& _id, std::string const& _address, unsigned _port, std::string const& _version, std::string const& _enr):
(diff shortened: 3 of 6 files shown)


twinstar26:
Found a bug where they are using "==" as pads and I am removing only one of them, actually I should be removing both.

libp2p/ENR.h
```diff
@@ -64,6 +68,7 @@ class ENR
     bytes content() const;
     size_t contentRlpListItemCount() const { return m_keyValuePairs.size() * 2 + 1; }
     void streamContent(RLPStream& _s) const;
+    Host* h;
```
gumb0: This is not correct (`ENR` can belong to some peer, not to `Host`) and not needed, just use `streamRLP` of this object in `encodeEnr`

libdevcore/Base64.h
```diff
@@ -35,6 +35,7 @@ namespace dev
 {
 
 std::string toBase64(bytesConstRef _in);
+std::string toBase64url(bytesConstRef _in);
```
gumb0: Call it `toBase64URLSafe`

libp2p/ENR.h
```diff
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <libdevcore/Base64.h>
```
gumb0: Move this to `ENR.cpp`

libp2p/ENR.h
```diff
@@ -56,6 +58,8 @@ class ENR
     ENR update(
         std::map<std::string, bytes> const& _keyValuePair, SignFunction const& _signFunction) const;
 
+    std::string encodeEnr();
```
gumb0: Call it `textEncoding()`

gumb0: Also make it `const` method

libp2p/ENR.h
```diff
@@ -36,6 +37,7 @@ class ENR
     /// Create with given sign function
     ENR(uint64_t _seq, std::map<std::string, bytes> const& _keyValuePairs,
         SignFunction const& _signFunction);
+    ENR() = default;
```
gumb0: You shouldn't need this.

libp2p/Host.h
```diff
@@ -240,7 +241,12 @@ class Host: public Worker
     }
 
     /// Get the node information.
-    p2p::NodeInfo nodeInfo() const { return NodeInfo(id(), (networkConfig().publicIPAddress.empty() ? m_tcpPublic.address().to_string() : networkConfig().publicIPAddress), m_tcpPublic.port(), m_clientVersion); }
+    p2p::NodeInfo nodeInfo() const {
+        ENR e;
```
gumb0: You will return empty ENR this way, not useful to anyone.
Get it from `enr()` method

gumb0: Also use clang-format for you changes as described in https://github.com/ethereum/aleth/blob/master/CONTRIBUTING.md

twinstar26: Thanks for the reviews!!

twinstar26:
> Instead of processing the result of `toBase64` it's better to create another similar function, then make both functions implemented in terms of the third internal function, taking `bytesConstRef _in` and the chars that vary as parameters.
> It could be for example `string toBase64(bytesConstRef _in, char const* _base64_chars, bool _pad)`
> (then define two variants of `base64_chars` array to pass there; `_pad` flags enables padding with `=`)

Oh some sort of wrapper functionality??

gumb0:
There will be one internal function, and then two public function will wrap internal one, passing it different arguments.

gumb0:
Replaced by https://github.com/ethereum/aleth/pull/5701

PR #5699, opened 2019-07-24
Implement EIP-2046 Reduced gas cost for static calls made to precompiles in LegacyVM

gumb0:
https://eips.ethereum.org/EIPS/eip-2046

~~Note that it puts repricing behind account versioning, i.e. reduced cost will be in effect only in contracts deployed with external transaction after Istanbul fork.~~

~~If it's decided that it should be in effect for version 0, we'll need to introduce another `EVMSchedule` ("between" current `ConstantinopleFixSchedule` and `IstanbulSchedule`)~~
Account versioning is disabled in Istanbul.

### DIFF
--- a/test/unittests/libevm/VMTest.cpp
+++ b/test/unittests/libevm/VMTest.cpp
@@ -782,13 +782,13 @@ public:
         BOOST_REQUIRE_EQUAL(gasBefore - gasAfter, 700 + 15);
     }
 
-    void testStaticCallCostEqualToCallBeforeIstanbul()
+    void testStaticCallCostEqualToCallBeforeBerlin()
     {
         // let r := staticcall(10000, 0x4, 0, 0, 0, 0)
         bytes code = fromHex("60006000600060006004612710fa50");
 
-        se.reset(ChainParams(genesisInfo(Network::ConstantinopleFixTest)).createSealEngine());
-        version = ConstantinopleFixSchedule.accountVersion;
+        se.reset(ChainParams(genesisInfo(Network::IstanbulTest)).createSealEngine());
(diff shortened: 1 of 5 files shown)


gumb0:
This is also not accepted yet, so shouldn't be merged.

gumb0:
This is rebased and ready for review.
The current state of the EIP is "tentatively accepted"

gumb0:
Rebased once again and updated changelog.

test/unittests/libevm/VMTest.cpp
```diff
@@ -570,6 +570,115 @@ class LegacyVMChainIDTestFixture : public ChainIDTestFixture
...
 };
 
+class PrecompileCallFixture : public TestOutputHelperFixture
+{
+public:
+    explicit PrecompileCallFixture(VMFace* _vm) : vm{_vm} { state.addBalance(address, 1 * ether); }
+
+    void testCallHasCorrectCost()
+    {
+        // let r := call(10000, 0x4, 0, 0, 0, 0, 0)
```
halfalicious: (Nit) can we document these args e.g.:

call(10000 /* gas  */, 0x4 /* address */, 0 /* value */, ...)

Makes the test code a little more readable 😀

gumb0:
@halfalicious Thanks for the review, this shouldn't be merged yet, as this EIP was not accepted for Istanbul, but probably will go into the next hard fork.

I will either put it into Experimental fork or create a new EVMSchedule for Berlin, next hard fork.

gumb0:
Rebased.

gumb0:
Moved to Berlin fork.

PR #5700, opened 2019-07-25
Implement EIP-1884: Repricing for trie-size-dependent opcodes in LegacyVM

gumb0:
https://eips.ethereum.org/EIPS/eip-1884

~~Note that it puts repricing and `SELFBALANCE` behind account versioning, i.e. reduced costs and the new opcode will be active only in contracts deployed with external transaction after Istanbul fork.~~

~~If it's decided that it should be in effect for version 0, we'll need to introduce another `EVMSchedule` ("between" current `ConstantinopleFixSchedule` and `IstanbulSchedule`)~~
Account versioning is disabled in Istanbul.

### DIFF
--- a/test/unittests/libevm/VMTest.cpp
+++ b/test/unittests/libevm/VMTest.cpp
@@ -577,6 +577,84 @@ public:
     {}
 };
 
+class SelfBalanceFixture : public TestOutputHelperFixture
+{
+public:
+    explicit SelfBalanceFixture(VMFace* _vm) : vm{_vm} { state.addBalance(address, 1 * ether); }
+
+    void testSelfBalanceWorksInIstanbul()
+    {
+        ExtVM extVm(state, envInfo, *se, address, address, address, value, gasPrice, {}, ref(code),
+            sha3(code), version, depth, isCreate, staticCall);
+
+        owning_bytes_ref ret = vm->exec(gas, extVm, OnOpFunc{});
+
+        BOOST_REQUIRE_EQUAL(fromBigEndian<u256>(ret), 1 * ether);
+    }
+
+    void testSelfBalanceHasCorrectCost()
+    {
+        ExtVM extVm(state, envInfo, *se, address, address, address, value, gasPrice, {}, ref(code),
+            sha3(code), version, depth, isCreate, staticCall);
+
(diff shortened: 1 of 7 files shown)


gumb0:
cc @holiman @winsvega

gumb0:
This is rebased and ready for review.
The current state of the EIP is "tentatively accepted".

gumb0:
Rebased.

EIP is Accepted now, please review @chfast @halfalicious

halfalicious:
@gumb0 For reference, why was the account versioning EIP bumped from Istanbul?

test/unittests/libevm/VMTest.cpp
```diff
@@ -577,6 +577,84 @@ class AlethInterpreterChainIDTestFixture : public ChainIDTestFixture
...
+
+        vm->exec(gas, extVm, onOp);
+
+        BOOST_REQUIRE_EQUAL(gasBefore - gasAfter, 5);
+    }
+
+    void testSelfBalanceisInvalidBeforeIstanbul()
+    {
+        se.reset(ChainParams(genesisInfo(Network::ConstantinopleFixTest)).createSealEngine());
+        version = ConstantinopleFixSchedule.accountVersion;
```
halfalicious: How come the account version is available in the Constantinople Fix schedule..do we need to reference it in this test since from what I understand it's been punted to post-Istanbul?

halfalicious:
Should we also add the following tests from [the EIP](https://eips.ethereum.org/EIPS/eip-1884):
* balance(this) and selfbalance return the same value (first test case)
* balance(this) costs 700 gas (second test case)
* selfbalance doesn't pop from stack (third test case)

gumb0:
@halfalicious Re account versioning: for now it's decided that no changes accepted for Istanbul are radical enough to put them under versioning. Actually I'm not sure anymore it will be ever needed.

test/unittests/libevm/VMTest.cpp
gumb0: Versioning still exists in the code for all forks, with version being 0 in all forks before we activate version 1 (currently it's `Experimental` fork)

This line is actually redundant.

PR #5703, opened 2019-07-29
Disable account versioning in Istanbul

gumb0:
Now it feel like most (if not all) of the changes in Istanbul are not going to use account versioning, so I disable it here.

### DIFF
--- a/libevm/EVMC.cpp
+++ b/libevm/EVMC.cpp
@@ -14,7 +14,7 @@ namespace
 {
 evmc_revision toRevision(EVMSchedule const& _schedule) noexcept
 {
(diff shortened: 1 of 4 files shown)


libevm/EVMC.cpp
```diff
@@ -14,7 +14,7 @@ namespace
 {
 evmc_revision toRevision(EVMSchedule const& _schedule) noexcept
 {
-    if (_schedule.accountVersion == IstanbulSchedule.accountVersion)
+    if (_schedule.txDataNonZeroGas == 16)
```
gumb0: Probably better to change this to `.haveChainID` after https://github.com/ethereum/aleth/pull/5696 is merged

PR #5707, opened 2019-07-30
Aleth waits for 2 seconds after sending disconnect packet

### LINKED ISSUE #5650: Aleth should wait 2 seconds to close socket after sending disconnect to peer

twinstar26:
Fixes #5650 
Old PR #5658

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Added: [#5691](https://github.com/ethereum/aleth/pull/5691) Istanbul support: EIP-2028 Transaction data gas cost reduction.
 - Added: [#5696](https://github.com/ethereum/aleth/pull/5696) Istanbul support: EIP-1344 ChainID opcode.
 - Added: [#5701](https://github.com/ethereum/aleth/issues/5701) Outputs ENR text representation in admin.nodeInfo RPC.
+- Added: [#5707](https://github.com/ethereum/aleth/pull/5707) Aleth waits for 2 seconds after sending disconnect to peer before closing socket.
 - Changed: [#5532](https://github.com/ethereum/aleth/pull/5532) The leveldb is upgraded to 1.22. This is breaking change on Windows and the old databases are not compatible.
 - Changed: [#5559](https://github.com/ethereum/aleth/pull/5559) Update peer validation error messages.
 - Changed: [#5568](https://github.com/ethereum/aleth/pull/5568) Improve rlpx handshake log messages and create new rlpx log channel.
--- a/libp2p/Host.h
+++ b/libp2p/Host.h
@@ -266,9 +266,9 @@ public:
 
     std::shared_ptr<CapabilityHostFace> capabilityHost() const { return m_capabilityHost; }
 
-    /// Execute work on the network thread after @a _expiryTime delay.
+    /// Execute work on the network thread after an @a _expiryDelay delay.
     /// Returned timer should be kept alive until delay is over.
-    std::unique_ptr<ba::steady_timer> createTimer(std::chrono::seconds const& _expiryTime,
+    std::unique_ptr<ba::steady_timer> createTimer(std::chrono::seconds const& _expiryDelay,
         std::function<void(const boost::system::error_code& error)>&& _f);
 
 protected:
--- a/libp2p/Host.h
+++ b/libp2p/Host.h
@@ -266,10 +266,10 @@ public:
 
     std::shared_ptr<CapabilityHostFace> capabilityHost() const { return m_capabilityHost; }
 
-    std::shared_ptr<ba::steady_timer> createTimer()
-    {
-        return std::make_shared<ba::steady_timer>(m_ioContext);
-    }
+    /// Execute work on the network thread after @a _expiryTime delay.
+    /// Returned timer should be kept alive until delay is over.
+    std::unique_ptr<ba::steady_timer> createTimer(std::chrono::seconds const& _expiryTime,
+        std::function<void(const boost::system::error_code& error)>&& _f);
 
 protected:
     /*
--- a/libp2p/Host.h
+++ b/libp2p/Host.h
@@ -266,6 +266,11 @@ public:
 
     std::shared_ptr<CapabilityHostFace> capabilityHost() const { return m_capabilityHost; }
 
+    std::shared_ptr<ba::steady_timer> createTimer()
+    {
+        return std::make_shared<ba::steady_timer>(m_ioContext);
+    }
+
 protected:
     /*
      * Used by the host to run a capability's background work loop
--- a/libp2p/Session.cpp
+++ b/libp2p/Session.cpp
@@ -299,23 +299,24 @@ void Session::disconnect(DisconnectReason _reason)
     RLPStream s;
     prep(s, DisconnectPacket, 1) << (int)_reason;
     sealAndSend(s);
-    auto disconnectTimer = m_server->createTimer();
+
     auto self(shared_from_this());
-    disconnectTimer->expires_after(std::chrono::seconds(2));
-    disconnectTimer->async_wait([self, this, _reason](boost::system::error_code) {
-        bi::tcp::socket& socket = m_socket->ref();
-        if (socket.is_open())
-            try
-            {
-                boost::system::error_code ec;
-                LOG(m_netLoggerDetail) << "Closing (" << reasonOf(_reason) << ") connection with";
-                socket.shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec);
-                socket.close();
-            }
-            catch (...)
-            {
-            }
-    });
+    m_disconnectTimer = m_server->createTimer(
+        std::chrono::seconds(2), [self, this, _reason](boost::system::error_code) {
+            bi::tcp::socket& socket = m_socket->ref();
+            if (socket.is_open())
+                try
+                {
+                    boost::system::error_code ec;
+                    LOG(m_netLoggerDetail)
+                        << "Closing (" << reasonOf(_reason) << ") connection with";
+                    socket.shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec);
+                    socket.close();
+                }
+                catch (...)
+                {
+                }
+        });
 
     drop(_reason);
 }
--- a/libp2p/Session.cpp
+++ b/libp2p/Session.cpp
@@ -301,22 +301,11 @@ void Session::disconnect(DisconnectReason _reason)
     sealAndSend(s);
 
     auto self(shared_from_this());
-    m_disconnectTimer = m_server->createTimer(
-        std::chrono::seconds(2), [self, this, _reason](boost::system::error_code) {
-            bi::tcp::socket& socket = m_socket->ref();
-            if (socket.is_open())
-                try
-                {
-                    boost::system::error_code ec;
-                    LOG(m_netLoggerDetail)
-                        << "Closing (" << reasonOf(_reason) << ") connection with";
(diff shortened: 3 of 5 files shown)


halfalicious:
@twinstar26 Is this ready for review? If so, can you please request a review from myself / gumb0 / chfast?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Added: [#5634](https://github.com/ethereum/aleth/pull/5634) Bootnodes for Rinkeby and Goerli.
 - Added: [#5640](https://github.com/ethereum/aleth/pull/5640) Istanbul support: EIP-1702 Generalized Account Versioning Scheme.
 - Added: [#5690](https://github.com/ethereum/aleth/issues/5690) Istanbul support: EIP-2028 transaction data gas cost reduction.
+- Added: [#5650](https://github.com/ethereum/aleth/issues/5650) Aleth should wait for 2 seconds to close socket after sending disconnect to peer.
```
halfalicious: Changelog should describe the behavior with your fix applied (e.g. "Aleth now waits for 2 seconds after sending disconnect to peer before closing socket")

libp2p/Session.cpp
```diff
@@ -300,13 +303,23 @@ void Session::disconnect(DisconnectReason _reason)
 {
     clog(VerbosityTrace, "p2pcap") << "Disconnecting (our reason: " << reasonOf(_reason) << ") from " << m_logSuffix;
 
-    if (m_socket->ref().is_open())
+    if (!m_dropped)
     {
         RLPStream s;
         prep(s, DisconnectPacket, 1) << (int)_reason;
         sealAndSend(s);
+        auto disconnectTimer = m_server->createTimer();
```
halfalicious: This code will still close the socket immediately (`Session::drop` closes the socket and you're calling it in both your timer handler and outside the handler if it hasn't already been called i.e. !m_dropped). Since what we want to do is delay the socket closure for 2 seconds, you should move the code which closes the socket out of Session::drop and into the timer handler and also not call drop inside of the timer handler.

twinstar26: @halfalicious  
Thanks for the reviews!
Okay so I have to copy required logic of `drop` into `disconnect` and never call `drop` from inside of `disconnect` and its handler. Am I right??

halfalicious: You should move the socket closure code out of drop and into the handler. Don’t call drop inside of the handler but still call it in your else statement in disconnect (since we want to prevent further session operations after we have sent the disconnect, you can see that we check for m_drop in various places in session and return early)

twinstar26: @halfalicious 
So pushed changes according to your suggested changes. But was curious what will happen to code thats calling `drop()` (apart from `disconnect`) separately??

gumb0: This still doesn't look right:
1. Changing `drop()` now breaks all the cases where it is called from other methods of `Session`. I guess those are reasonable cases where we wanted to just disconnect immediately without sending `Disconnect` packet (with current changes we won't disconnect)

2. Calling `disconnect` now doesn't ever set `m_dropped` to `true`. @halfalicious Maybe you meant calling current form of `drop` from inside `if (!m_dropped)` clause?

twinstar26: @gumb0 @halfalicious 
So what we can do is create a overloaded copy of drop() since functionality of drop is now slightly different in 2 cases. (one where normal code calls it. two where disconnect calls it). We can place second argument of drop() as bool or number of seconds(which will be 2). 

Will this work out??

halfalicious: @gumb0 You're right we need to also call drop() inside of the !m_dropped clause. 

I don't think this new form of drop breaks anything though...even if drop doesn't close the socket all shared_ptr Session instances will still go out of scope and therefore Session will be garbage collected by Host and the socket will be closed when the dtor is invoked. Note that functions in which we pass new shared_ptr instances (shared_from_this) to handlers check `if (m_dropped)` and then prematurely terminate before creating new handlers (e.g. `doRead`).  In fact, after taking a closer look at the Session code, I don't think we need to explicitly close the socket at all in any place other than the dtor (unless there's some benefit to closing the socket sooner rather than later that I'm not aware of)? If we just call drop() without closing the socket the Session instance will be destroyed once all queued handlers which contain a shared_ptr instance are executed and the Host garbage collects the Session.

So, to summarize:
* I think we should call drop() in both the if (!m_dropped) and else cases...in fact, we can simplify this by removing the "else" case and always calling drop() after  the `if (!m_dropped)` code block since `drop `returns early if `m_dropped` is true.
* I think we can remove all cases where we close the socket except for in the dtor. 

@gumb0 : Thoughts?

gumb0: >  I don't think we need to explicitly close the socket at all in any place other than the dtor (unless there's some benefit to closing the socket sooner rather than later that I'm not aware of)? 

@halfalicious The benefit I see is that we immediately stop answering any further requests on this connection, not waiting for all handlers to terminate, so maybe it's more efficient...

Also, what if some malicious peer keeps DOSing us with some invalid requests on this connection - it will be kept in the handler queue (becouse the new data keeps coming) and never disconnected, it seems so.

libp2p/Session.cpp
```diff
@@ -298,14 +291,32 @@ void Session::drop(DisconnectReason _reason)
...
-    if (m_socket->ref().is_open())
-    {
-        RLPStream s;
-        prep(s, DisconnectPacket, 1) << (int)_reason;
-        sealAndSend(s);
-    }
+    RLPStream s;
+    prep(s, DisconnectPacket, 1) << (int)_reason;
+    sealAndSend(s);
+    auto disconnectTimer = m_server->createTimer();
```
gumb0: I think this timer will be destroyed right after this function ends, and outstanding wait operation will be cancelled. (Probably should be made a member?)

gumb0:
I moved the calls to timer's member inside `createTimer()` function to hide these lower-level details from `Host`'s users and for better symmetry between `postWork` and `createTimer`

gumb0:
Rebased.

libp2p/Session.cpp
gumb0: @halfalicious pointed out to me:

> I think that we will still stop reading packets from a peer if we call drop without closing the socket since we check for m_dropped at the beginning of Session::doRead before proceeding with the read: https://github.com/ethereum/aleth/blob/d8f4498f5b2bfa218c40a0cc760b6a30a4d946c1/libp2p/Session.cpp#L318-L322

I think I agree with this reasoning.

gumb0: One corner case to check: spec requires that we disconnet immediately upon receiving Disconnect packet ourselves. My experiment confirms that this looks to be working correctly with current changes in this PR:
```
TRACE 08-08 11:54:24 p2p  p2pcap Disconnect from (##a8768590…@51.141.78.53:30303)
DEBUG 08-08 11:54:24 p2p  p2pcap Disconnect (reason: Peer had too many connections.) from (##a8768590…@51.141.78.53:30303)
DEBUG 08-08 11:54:24 p2p  net    Closing peer session with (##a8768590…@51.141.78.53:30303)
```

libp2p/Session.cpp
```diff
@@ -298,14 +291,33 @@ void Session::drop(DisconnectReason _reason)
...
-    }
+    RLPStream s;
+    prep(s, DisconnectPacket, 1) << (int)_reason;
+    sealAndSend(s);
+
+    auto self(shared_from_this());
+    m_disconnectTimer = m_server->createTimer(
+        std::chrono::seconds(2), [self, this, _reason](boost::system::error_code) {
+            bi::tcp::socket& socket = m_socket->ref();
+            if (socket.is_open())
```
gumb0: One thing I still don't like here is that we have the same code (closing the socket) here in the timer handler and then in destructor. And destructor is executed anyway, so perhaps we could avoid this duplication, maybe just posting empty handler? @halfalicious what do you think?

halfalicious: @gumb0 Yes I agree with this, in fact [I suggested this earlier in this PR](https://github.com/ethereum/aleth/pull/5707#discussion_r309971423) 😄

libp2p/Session.cpp
> 
> One corner case to check: spec requires that we disconnet immediately upon receiving Disconnect packet ourselves. My experiment confirms that this looks to be working correctly with current changes in this PR:
> 
> ```
> TRACE 08-08 11:54:24 p2p  p2pcap Disconnect from (##a8768590…@51.141.78.53:30303)
> DEBUG 08-08 11:54:24 p2p  p2pcap Disconnect (reason: Peer had too many connections.) from (##a8768590…@51.141.78.53:30303)
> DEBUG 08-08 11:54:24 p2p  net    Closing peer session with (##a8768590…@51.141.78.53:30303)
> ```

Confirmed by looking at the code - when we receive a disconnect packet we call `drop ` immediately in `Session::interpretP2pPacket`: https://github.com/ethereum/aleth/blob/78355a28da55f1c0e952e3cf1a17891cea981cb4/libp2p/Session.cpp#L142-L154

libp2p/Session.cpp
halfalicious: @gumb0 Agreed, in my prototype I had captured the timer in the handler but making it a member works as well 😄

halfalicious:
Confirmed that we wait for 2 seconds now:

```
TRACE 08-12 21:33:11 p2p  p2pcap Disconnecting (our reason: Peer had no use for this node.) from (##03957ecc…@51.75.127.66:30303)
TRACE 08-12 21:33:11 p2p  net    Disconnect to (##03957ecc…@51.75.127.66:30303)
...
DEBUG 08-12 21:33:13 p2p  net    Closing peer session with (##03957ecc…@51.75.127.66:30303)
```

halfalicious:
@gumb0 I think this is ready to be merged, please give it a quick look when you get a chance

libp2p/Host.h
```diff
@@ -266,6 +266,11 @@ class Host: public Worker
 
     std::shared_ptr<CapabilityHostFace> capabilityHost() const { return m_capabilityHost; }
 
+    /// Execute work on the network thread after an @_expiryDelay delay.
```
gumb0: That was doxygen syntax to refer to the parameter inside description
(http://www.doxygen.nl/manual/commands.html#cmda)

```suggestion
    /// Execute work on the network thread after an @a _expiryDelay delay.
```

halfalicious: @gumb0 That's what I suspected but I saw different syntax used in other parts of the project e.g.: https://github.com/ethereum/aleth/blob/78355a28da55f1c0e952e3cf1a17891cea981cb4/libethereum/Account.h#L150-L152

PR #5709, opened 2019-07-31
Implement EIP-2200: Structured Definitions for Net Gas Metering in LegacyVM

gumb0:
https://github.com/ethereum/EIPs/pull/2200

I refactored SSTORE pricing unit tests a bit, to make them more table-driven.

### DIFF
--- a/libethcore/EVMSchedule.h
+++ b/libethcore/EVMSchedule.h
@@ -35,6 +35,7 @@ struct EVMSchedule
     bool eip150Mode = false;
     bool eip158Mode = false;
     bool eip1283Mode = false;
+    bool eip2200Mode = false;
     bool haveBitwiseShifting = false;
     bool haveRevert = false;
     bool haveReturnData = false;
@@ -85,6 +86,8 @@ struct EVMSchedule
     bool staticCallDepthLimit() const { return !eip150Mode; }
     bool emptinessIsNonexistence() const { return eip158Mode; }
     bool zeroValueTransferChargesNewAccountGas() const { return !eip158Mode; }
+    bool sstoreNetGasMetering() const { return eip1283Mode; }
+    bool sstoreThrowsIfGasBelowCallStipend() const { return eip2200Mode; }
 };
 
 static const EVMSchedule DefaultSchedule = EVMSchedule();
@@ -156,6 +159,7 @@ static const EVMSchedule IstanbulSchedule = [] {
     schedule.haveChainID = true;
     schedule.haveSelfbalance = true;
     schedule.eip1283Mode = true;
+    schedule.eip2200Mode = true;
     schedule.sstoreUnchangedGas = 800;
     return schedule;
 }();
--- a/libethcore/EVMSchedule.h
+++ b/libethcore/EVMSchedule.h
@@ -53,7 +53,6 @@ struct EVMSchedule
     unsigned sstoreResetGas = 5000;
     unsigned sstoreUnchangedGas = 200;
     unsigned sstoreRefundGas = 15000;
-    unsigned sstoreRefundNonzeroGas = 4800;
     unsigned jumpdestGas = 1;
     unsigned logGas = 375;
     unsigned logDataGas = 8;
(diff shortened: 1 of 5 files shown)


gumb0:
Rebased.

halfalicious:
You're getting an internal compiler error on the gcc config for some reason:
```
[195/251] Building CXX object aleth/CMakeFiles/aleth.dir/main.cpp.o
FAILED: aleth/CMakeFiles/aleth.dir/main.cpp.o 
/usr/bin/ccache /usr/bin/g++-6  -DETH_FATDB -I/home/builder/project -I/home/builder/project/aleth/../utils -isystem /home/builder/.hunter/_Base/c022f0c/0fe0970/6d1bd54/Install/include -I. -isystem deps/include -I/home/builder/project/evmc/include -O2 -g -fPIE   -fstack-protector-strong -Wall -Wextra -Werror -Wno-unknown-pragmas -fvisibility=hidden -g --coverage -std=c++11 -MD -MT aleth/CMakeFiles/aleth.dir/main.cpp.o -MF aleth/CMakeFiles/aleth.dir/main.cpp.o.d -o aleth/CMakeFiles/aleth.dir/main.cpp.o -c /home/builder/project/aleth/main.cpp
g++-6: internal compiler error: Killed (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-6/README.Bugs> for instructions.
[196/251] Building CXX object test/CMakeFiles/testeth.dir/tools/libtesteth/ImportTest.cpp.o
[197/251] Building CXX object test/CMakeFiles/testeth.dir/tools/libtesteth/boostTest.cpp.o
ninja: build stopped: subcommand failed.
Exited with code 1
```

chfast:
> You're getting an internal compiler error on the gcc config for some reason:

Not enough memory. Try to lower the number of parallel builds.

gumb0:
@chfast @halfalicious lowered `BUILD_PARALLEL_JOBS` to 2, now it passes

libethcore/EVMSchedule.h
```diff
@@ -156,6 +158,9 @@ static const EVMSchedule IstanbulSchedule = [] {
     schedule.extcodehashGas = 700;
     schedule.haveChainID = true;
     schedule.haveSelfbalance = true;
+    schedule.eip1283Mode = true;
```
halfalicious: Do we still need this var? eip1283 was rejected from Istanbul in favor of eip2200: https://eips.ethereum.org/EIPS/eip-1679

Can we control both the sstore and sload net gas metering via eip2200Mode?

libevm/LegacyVM.cpp
```diff
@@ -97,10 +97,13 @@ void LegacyVM::adjustStack(unsigned _removed, unsigned _added)
 
 void LegacyVM::updateSSGas()
 {
+    if (m_schedule->sstoreThrowsIfGasBelowCallStipend() && m_io_gas <= m_schedule->callStipend)
```
halfalicious: Why do we only throw OOG if EIP2200 is enabled, I'd expect us to always throw OOG if we don't have enough gas to cover the call stipend?

libethcore/EVMSchedule.h
gumb0: Well it's basically the same rules... 
(These two EIPs differ only in some price constants and in additional "throw below 2300 gas" rule)

But yeah, I think it would make sense to have `eip1283Mode = true` only in Constantinople, and then I'll change  `sstoreNetGasMetering()` to return `true` for both `eip1283Mode` and `eip2200Mode`

libevm/LegacyVM.cpp
gumb0: This rule is introduced only in EIP-2200.
(without it we normally throw in case of OOG from the `updateIOGas` call here https://github.com/ethereum/aleth/blob/0faec9d18e09459462f392d710b1c0d7f5008d5f/libevm/LegacyVM.cpp#L1678)

libethcore/EVMSchedule.h
gumb0: changed

gumb0:
Rebased.

PR #5712, opened 2019-08-02
[WIP] fix testeth stateTests transaction execution on expectSections

winsvega:
Depends on: 
https://github.com/ethereum/aleth/pull/5618
https://github.com/ethereum/aleth/pull/5714

fix testeth statetests transaction execution on expect sections

This is the aleth version I use with retesteth. 
I've noticed that testeth generate transactions on expect sections that are not described. The sorting was done only on networks. So if transaction has gasIndex  0,  but there was no expect section with gasIndex0 it would still generate the test with it, leading to unwanted transaction being executed in some tests.

### DIFF
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -229,6 +229,44 @@ vector<int> parseJsonIntValueIntoVector(json_spirit::mValue const& _json)
     return out;
 }
 
+bool ImportTest::findExpectSectionForTransaction(
+    transactionToExecute const& _tr, eth::Network const& _net, bool _isFilling)
+{
+    bool found = false;
+    json_spirit::mArray const& expects =
+        _isFilling ? m_testInputObject.at("expect").get_array() :
+                     m_testInputObject.at("post").get_obj().at(netIdToString(_net)).get_array();
+    for (auto const& exp : expects)
+    {
+        json_spirit::mObject const& expSection = exp.get_obj();
+
+        if (_isFilling)
+        {
+            set<string> networkSet;
+            parseJsonStrValueIntoSet(expSection.at("network"), networkSet);
+            networkSet = test::translateNetworks(networkSet);
+            if (!networkSet.count(netIdToString(_net)))
+                continue;
+        }
+
+        BOOST_REQUIRE_MESSAGE(expSection.at("indexes").type() == jsonVType::obj_type,
+            "indexes field expected to be json Object!");
+        json_spirit::mObject const& indexes = expSection.at("indexes").get_obj();
+        vector<int> d = parseJsonIntValueIntoVector(indexes.at("data"));
+        vector<int> g = parseJsonIntValueIntoVector(indexes.at("gas"));
+        vector<int> v = parseJsonIntValueIntoVector(indexes.at("value"));
+        BOOST_CHECK_MESSAGE(d.size() > 0 && g.size() > 0 && v.size() > 0,
+            TestOutputHelper::get().testName() + " Indexes arrays not set!");
+        if ((inArray(d, _tr.dataInd) || inArray(d, -1)) &&
+            (inArray(g, _tr.gasInd) || inArray(g, -1)) &&
+            (inArray(v, _tr.valInd) || inArray(v, -1)))
+            found = true;
+        if (found)
+            break;
+    }
+    return found;
+}
+
 bytes ImportTest::executeTest(bool _isFilling)
 {
     assert(m_envInfo);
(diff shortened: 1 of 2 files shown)


gumb0:
The changes outside of libtesteth should better go to another PR, so remove them from here

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -205,6 +205,28 @@ set<eth::Network> ImportTest::getAllNetworksFromExpectSections(
     return networkSet;
 }
 
+void ImportTest::parseJsonStrValueIntoSet(json_spirit::mValue const& _json, set<string>& _out)
```
gumb0: Why is this a member of the class and the function below is not?
Make them both free functions in unnamed namespace. Make them return set/vector instead of output params.

winsvega: unnamed no. it is used outside of this file

winsvega: the other one is not used outside

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -241,10 +263,62 @@ bytes ImportTest::executeTest(bool _isFilling)
...
                 continue;
 
-            std::tie(tr.postState, tr.output, tr.changeLog) =
-                    executeTransaction(net, *m_envInfo, m_statePre, tr.transaction);
-            tr.netId = net;
-            transactionResults.push_back(tr);
+            // Skip transaction if there is no expect section for this transaction (noindex data,
+            // gas, val, network)
+            bool found = false;
+            json_spirit::mArray expects;
```
gumb0: Can be const if you use ternary operator
```
json_spirit::mArray const expects = _isFilling ? m_testInputObject.at("expect").get_array() : m_testInputObject.at("post").get_obj().at(netIdToString(net)).get_array();
```

gumb0: Also maybe should be reference?

gumb0: Can this be done outside of for loops?

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -241,10 +263,62 @@ bytes ImportTest::executeTest(bool _isFilling)
...
+                expects = m_testInputObject.at("post").get_obj().at(netIdToString(net)).get_array();
+
+            for (auto const& exp : expects)
+            {
+                json_spirit::mObject const& expSection = exp.get_obj();
+
+                if (_isFilling)
+                {
+                    set<string> networkSet;
+                    parseJsonStrValueIntoSet(expSection.at("network"), networkSet);
```
gumb0: Can this and the next line be done outside of for loops?

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -241,10 +263,62 @@ bytes ImportTest::executeTest(bool _isFilling)
...
+                    (inArray(v, tr.valInd) || inArray(v, -1)))
+                    found = true;
+                if (found)
+                    break;
+            }
+            if (!found)
+                continue;
+
+
+            transactionToExecute tr2 = tr;
```
gumb0: call it `transactuinResult`
(weird that the same type represents both the transaction and transaction result)

gumb0:
Needs rebase

test/tools/libtesteth/ImportTest.cpp
winsvega: no this is done for each expect section

winsvega: but you right this information could be gathered once. before the loops. it's just the plain way to do this. more logical objects to store the information about expect section required. I don't want to change the code very much. I hope it is the last change. and this code will soon become obsolite / replaced by retesteth

test/tools/libtesteth/ImportTest.cpp
winsvega: looking at this code after one year I am surprised how tangled it is and how it is hard to make any changes. I really hope to drop testeth support soon.

winsvega: you're welcome to review retesteth code to prevent this entanglement from happening. 
that's voluntary

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -205,6 +205,68 @@ set<eth::Network> ImportTest::getAllNetworksFromExpectSections(
...
+    if (_json.type() == json_spirit::array_type)
+    {
+        for (auto const& val : _json.get_array())
+            _out.emplace(val.get_str());
+    }
+    else
+        _out.emplace(_json.get_str());
+}
+
+vector<int> parseJsonIntValueIntoVector(json_spirit::mValue const& _json)
```
gumb0: Either put it into unnamed namespace or make it private static member of the class

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -205,6 +205,68 @@ set<eth::Network> ImportTest::getAllNetworksFromExpectSections(
...
+        for (auto const& val : _json.get_array())
+            out.push_back(val.get_int());
+    }
+    else
+        out.push_back(_json.get_int());
+    return out;
+}
+
+bool ImportTest::findExpectSectionForTransaction(
+    transactionToExecute const& _tr, eth::Network const& _net, bool _isFilling)
```
gumb0: Can be const method?

test/tools/libtesteth/ImportTest.h
```diff
@@ -69,10 +69,13 @@ class ImportTest
 	eth::State m_statePost;
 
 private:
-	using ExecOutput = std::pair<eth::ExecutionResult, eth::TransactionReceipt>;
+    struct transactionToExecute;
```
gumb0: This is not needed, it's defined in the scope of the same class.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -205,6 +205,68 @@ set<eth::Network> ImportTest::getAllNetworksFromExpectSections(
...
+        json_spirit::mObject const& indexes = expSection.at("indexes").get_obj();
+        vector<int> d = parseJsonIntValueIntoVector(indexes.at("data"));
+        vector<int> g = parseJsonIntValueIntoVector(indexes.at("gas"));
+        vector<int> v = parseJsonIntValueIntoVector(indexes.at("value"));
+        BOOST_CHECK_MESSAGE(d.size() > 0 && g.size() > 0 && v.size() > 0,
+            TestOutputHelper::get().testName() + " Indexes arrays not set!");
+        if ((inArray(d, _tr.dataInd) || inArray(d, -1)) &&
+            (inArray(g, _tr.gasInd) || inArray(g, -1)) &&
+            (inArray(v, _tr.valInd) || inArray(v, -1)))
+            found = true;
```
gumb0: You can get rid of `found` variable and just `return true`

test/tools/libtesteth/ImportTest.h
winsvega: compiler errors on declaration of this method 
```
bool findExpectSectionForTransaction(
        transactionToExecute const& _tr, eth::Network const& _net, bool _isFilling);

```

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -205,6 +205,68 @@ set<eth::Network> ImportTest::getAllNetworksFromExpectSections(
...
+{
+    bool found = false;
+    json_spirit::mArray const& expects =
+        _isFilling ? m_testInputObject.at("expect").get_array() :
+                     m_testInputObject.at("post").get_obj().at(netIdToString(_net)).get_array();
+    for (auto const& exp : expects)
+    {
+        json_spirit::mObject const& expSection = exp.get_obj();
+
+        if (_isFilling)
```
gumb0: I don't get why we do this part only for filling

winsvega: because when filling one expect section defines many networks
when running the test each expect section is already under single network

test/tools/libtesteth/ImportTest.h
gumb0: Oh ok, I was wrong.
I would move then this struct declaration to here, above all methods.

test/tools/libtesteth/ImportTest.cpp
gumb0: Do we then need to check during running that this found expected section is for the needed network?
(now `_net` parameter is not used during running, is this correct?)

gumb0: Ah no, I see now it's used

gumb0: Ok I kind of got it
(I would create 2 different `findExpectSection` function for these two very different file formats, just saying)

winsvega: Pawel suggest to remove test generation first from testeth

winsvega:
looks like with this PR I could detect a couple of uncovered with expect sections transactions at StateTests

PR #5713, opened 2019-08-03
Propagate blocks after PoW validation

### LINKED ISSUE #5371: Propagate new blocks after PoW check
There's a consensus now that it's more efficient for the network as a whole to propagate the blocks as soon as the PoW check is complete. go-ethereum and Parity already do that, we still propagate only after the blocks are fully validated and imported into the chain.

Relevant go-ethereum codes 
https://github.com/ethereum/go-ethereum/blob/edc39aaedf526bfb7757a445c1f9dd42f45dc8d4/eth/handler.go#L694-L732
https://github.com/ethereum/go-ethereum/blob/edc39aaedf526bfb7757a445c1f9dd42f45dc8d4/eth/fetcher/fetcher.go#L635-L682

Note that the hashes of the new blocks area announced (with `NewBlockHashes` message) by go-ethereum only after sucessfull full validation.

Relevant Parity PR https://github.com/paritytech/parity-ethereum/pull/9954

halfalicious:
Fix #5371 

Propagate new blocks after they've been imported into the block queue and validated. This results in the blocks being sent out faster compared to waiting until they've been imported into the block chain which reduces the uncle rate.

### DIFF
--- a/libdevcore/Common.h
+++ b/libdevcore/Common.h
@@ -317,4 +317,6 @@ private:
 
 bool isTrue(std::string const& _m);
 bool isFalse(std::string const& _m);
+
+int randomNumber(int _min = INT_MIN, int _max = INT_MAX);
 }  // namespace dev
--- a/libdevcore/Common.h
+++ b/libdevcore/Common.h
@@ -317,6 +317,4 @@ private:
 
 bool isTrue(std::string const& _m);
 bool isFalse(std::string const& _m);
-
-int randomNumber(int _min = INT_MIN, int _max = INT_MAX);
 }  // namespace dev
--- a/libethereum/BlockChain.h
+++ b/libethereum/BlockChain.h
@@ -101,10 +101,15 @@ public:
     void process();
 
     /// Sync the chain with any incoming blocks. All blocks should, if processed in order.
-    /// @returns fresh blocks, dead blocks and true iff there are additional blocks to be processed
-    /// waiting.
+    /// @returns tuple with three members - the first (ImportRoute) contains hashes of fresh blocks
+    /// and dead blocks as well as a list of imported transactions. The second is a bool which is
+    /// true iff there are additional blocks to be processed. The third is the imported block count.
     std::tuple<ImportRoute, bool, unsigned> sync(
         BlockQueue& _bq, OverlayDB const& _stateDB, unsigned _max);
+
+    /// Import the supplied blocks into the chain. Blocks should be processed in order.
+    /// @returns a tuple with two members - the first (ImportRoute) contains fresh blocks, dead
+    /// blocks and imported transactions. The second contains the imported block count.
     std::tuple<ImportRoute, unsigned> sync(
         VerifiedBlocks const& _blocks, h256s& _badBlockHashes, OverlayDB const& _stateDB);
 
--- a/libethereum/BlockChain.h
+++ b/libethereum/BlockChain.h
@@ -111,7 +111,7 @@ public:
     /// @returns a tuple with two members - the first (ImportRoute) contains fresh blocks, dead
     /// blocks and imported transactions. The second contains the imported block count.
     std::tuple<ImportRoute, unsigned> sync(
-        VerifiedBlocks const& _blocks, h256s& _badBlockHashes, OverlayDB const& _stateDB);
+        VerifiedBlocks const& _blocks, h256s& o_badBlockHashes, OverlayDB const& _stateDB);
 
     /// Attempt to import the given block directly into the BlockChain and sync with the state DB.
     /// @returns the block hashes of any blocks that came into/went out of the canonical block chain.
--- a/libethereum/BlockChain.h
+++ b/libethereum/BlockChain.h
@@ -1,23 +1,6 @@
-/*
-    This file is part of cpp-ethereum.
-
-    cpp-ethereum is free software: you can redistribute it and/or modify
-    it under the terms of the GNU General Public License as published by
-    the Free Software Foundation, either version 3 of the License, or
-    (at your option) any later version.
-
-    cpp-ethereum is distributed in the hope that it will be useful,
-    but WITHOUT ANY WARRANTY; without even the implied warranty of
-    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-    GNU General Public License for more details.
-
-    You should have received a copy of the GNU General Public License
-    along with cpp-ethereum.  If not, see <http://www.gnu.org/licenses/>.
-*/
-/** @file BlockChain.h
- * @author Gav Wood <i@gavwood.com>
- * @date 2014
- */
+// Aleth: Ethereum C++ client, tools and libraries.
+// Copyright 2019 Aleth Authors.
+// Licensed under the GNU General Public License, Version 3.
 
 #pragma once
 
--- a/libethereum/BlockChain.h
+++ b/libethereum/BlockChain.h
@@ -101,8 +101,12 @@ public:
     void process();
 
     /// Sync the chain with any incoming blocks. All blocks should, if processed in order.
-    /// @returns fresh blocks, dead blocks and true iff there are additional blocks to be processed waiting.
-    std::tuple<ImportRoute, bool, unsigned> sync(BlockQueue& _bq, OverlayDB const& _stateDB, unsigned _max);
+    /// @returns fresh blocks, dead blocks and true iff there are additional blocks to be processed
+    /// waiting.
+    std::tuple<ImportRoute, bool, unsigned> sync(
+        BlockQueue& _bq, OverlayDB const& _stateDB, unsigned _max);
+    std::tuple<ImportRoute, unsigned> sync(
+        VerifiedBlocks const& _blocks, h256s& _badBlockHashes, OverlayDB const& _stateDB);
 
     /// Attempt to import the given block directly into the BlockChain and sync with the state DB.
     /// @returns the block hashes of any blocks that came into/went out of the canonical block chain.
--- a/libethereum/BlockChain.h
+++ b/libethereum/BlockChain.h
@@ -108,10 +108,11 @@ public:
         BlockQueue& _bq, OverlayDB const& _stateDB, unsigned _max);
 
     /// Import the supplied blocks into the chain. Blocks should be processed in order.
-    /// @returns a tuple with two members - the first (ImportRoute) contains fresh blocks, dead
-    /// blocks and imported transactions. The second contains the imported block count.
-    std::tuple<ImportRoute, unsigned> sync(
-        VerifiedBlocks const& _blocks, h256s& o_badBlockHashes, OverlayDB const& _stateDB);
+    /// @returns a tuple with three members - the first (ImportRoute) contains fresh blocks, dead
+    /// blocks and imported transactions. The second contains hashes of bad blocks. The third
+    /// contains the imported block count.
+    std::tuple<ImportRoute, h256s, unsigned> sync(
+        VerifiedBlocks const& _blocks, OverlayDB const& _stateDB);
 
     /// Attempt to import the given block directly into the BlockChain and sync with the state DB.
     /// @returns the block hashes of any blocks that came into/went out of the canonical block chain.
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -388,7 +388,24 @@ void Client::syncBlockQueue()
     ImportRoute ir;
     unsigned count;
     Timer t;
-    tie(ir, m_syncBlockQueue, count) = bc().sync(m_bq, m_stateDB, m_syncAmount);
+
+    std::shared_ptr<VerifiedBlocks> verifiedBlocks = std::make_shared<VerifiedBlocks>();
+    m_bq.drain(*verifiedBlocks, m_syncAmount);
+
+    // Propagate new blocks to peers before importing them into the chain
+    auto h = m_host.lock();
+    if (!h)
+    {
+        // TODO: Can we ever hit this?
+        LOG(m_logger) << "Host unavailable??";
+        return;
+    }
+    h->propagateBlocks(verifiedBlocks);
+
+    h256s badBlockHashes;
+    tie(ir, count) = bc().sync(*verifiedBlocks, badBlockHashes, m_stateDB);
+    m_syncBlockQueue = m_bq.doneDrain(badBlockHashes);
+
     double elapsed = t.elapsed();
 
     if (count)
@@ -557,8 +574,8 @@ void Client::onChainChanged(ImportRoute const& _ir)
     h256Hash changeds;
     onDeadBlocks(_ir.deadBlocks, changeds);
     vector<h256> goodTransactions;
-    goodTransactions.reserve(_ir.goodTranactions.size());
-    for (auto const& t: _ir.goodTranactions)
+    goodTransactions.reserve(_ir.goodTransactions.size());
+    for (auto const& t: _ir.goodTransactions)
     {
         LOG(m_loggerDetail) << "Safely dropping transaction " << t.sha3();
         m_tq.dropGood(t);
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -43,6 +43,10 @@ static_assert(BOOST_VERSION >= 106400, "Wrong boost headers version");
 
 namespace
 {
+unsigned constexpr c_syncMin = 1;
+unsigned constexpr c_syncMax = 1000;
+double constexpr c_targetDuration = 1;
+
 std::string filtersToString(h256Hash const& _fs)
 {
     std::stringstream str;
@@ -377,10 +381,6 @@ void Client::appendFromBlock(h256 const& _block, BlockPolarity _polarity, h256Ha
     }
 }
 
-unsigned static const c_syncMin = 1;
-unsigned static const c_syncMax = 1000;
-double static const c_targetDuration = 1;
-
 void Client::syncBlockQueue()
 {
 //  cdebug << "syncBlockQueue()";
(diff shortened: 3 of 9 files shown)


halfalicious:
Currently propagate all blocks which are imported into the block queue and which pass validation, should only propagate new blocks

halfalicious:
> 
> 
> Currently propagate all blocks which are imported into the block queue and which pass validation, should only propagate new blocks

At the very least we should check if we are currently syncing in `EthereumCapability::propagateBlocks `and return early if we are here (the presumption being that if we are syncing we are still trying to get to chain tip and haven't started being sent new blocks or mining yet): https://github.com/ethereum/aleth/blob/e390d3c778abb94930423e6cb328f876ea4497e2/libethereum/EthereumCapability.cpp#L955-L958

Note that we have similar logic in `EthereumCapability::doBackgroundWork`: https://github.com/ethereum/aleth/blob/e390d3c778abb94930423e6cb328f876ea4497e2/libethereum/EthereumCapability.cpp#L857-L864

halfalicious:
Had to do some manual reformatting of the code since I don't have git clang-format working again via the command line after reinstalling my OS so apologies if there are some unnecessary formatting changes in this PR.

libethereum/Client.cpp
```diff
@@ -377,18 +381,27 @@ void Client::appendFromBlock(h256 const& _block, BlockPolarity _polarity, h256Ha
...
 void Client::syncBlockQueue()
 {
 //  cdebug << "syncBlockQueue()";
 
     ImportRoute ir;
     unsigned count;
     Timer t;
-    tie(ir, m_syncBlockQueue, count) = bc().sync(m_bq, m_stateDB, m_syncAmount);
+
+    shared_ptr<VerifiedBlocks> verifiedBlocks = make_shared<VerifiedBlocks>();
```
halfalicious: shared_ptr is required here since we propagate the blocks on another thread so we need to ensure that they are kept around

halfalicious:
Received some internal RLPX handshake errors during local testing:

```
aleth, a C++ Ethereum client
INFO  08-12 20:41:00 main net    Id: ##90a0431e…
INFO  08-12 20:41:00 main net    ENR: [ seq=1 id=v4 key=0290a043… tcp=30303 udp=30303 ]
aleth 1.7.0-alpha.1-20+commit.b4e4fb1a.dirty
INFO  08-12 20:41:41 p2p  info   UPnP device not found.
INFO  08-12 20:41:41 p2p  net    Active peer count: 0
INFO  08-12 20:41:41 p2p  net    Looking for peers...
Node ID: enode://90a0431e893c8fa9a0697b51e5876668f099e0fd7fd268bd0021e8d9125c1805bfac7d0e44613c7fce8275406b8126a4f299990ec8df9b1859e6b7f4e123352a@127.0.0.1:30303
INFO  08-12 20:41:41 main rpc    JSON-RPC socket path: \\.\pipe\\geth.ipc
JSONRPC Admin Session Key: hUlrVuzBFoo=
INFO  08-12 20:41:41 p2p  net    ENR updated: [ seq=2 id=v4 key=0290a043… ip=71.227.153.208 tcp=30303 udp=30303 ]
INFO  08-12 20:41:42 p2p  sync   Starting full sync
ERROR 08-12 20:41:54 p2p  rlpx   egress Internal error in handshake: RLPXFrameCoder disappeared (##5a85cb59…@35.181.105.86:30303)
ERROR 08-12 20:41:54 p2p  rlpx   egress Internal error in handshake: RLPXFrameCoder disappeared (##e31b2dbe…@40.115.25.33:30303)
ERROR 08-12 20:41:54 p2p  rlpx   egress Internal error in handshake: RLPXFrameCoder disappeared (##5fb5c839…@188.214.30.221:30415)
                                                                                                                                           
```

These appear to be transient, will dig deeper to see if they are related to my changes (I don't think they are since I didn't touch anything related to handshakes but I haven't seen these errors before)

halfalicious:
Rebased

halfalicious:
> 
> 
> Received some internal RLPX handshake errors during local testing:
> 
> ```
> aleth, a C++ Ethereum client
> INFO  08-12 20:41:00 main net    Id: ##90a0431e…
> INFO  08-12 20:41:00 main net    ENR: [ seq=1 id=v4 key=0290a043… tcp=30303 udp=30303 ]
> aleth 1.7.0-alpha.1-20+commit.b4e4fb1a.dirty
> INFO  08-12 20:41:41 p2p  info   UPnP device not found.
> INFO  08-12 20:41:41 p2p  net    Active peer count: 0
> INFO  08-12 20:41:41 p2p  net    Looking for peers...
> Node ID: enode://90a0431e893c8fa9a0697b51e5876668f099e0fd7fd268bd0021e8d9125c1805bfac7d0e44613c7fce8275406b8126a4f299990ec8df9b1859e6b7f4e123352a@127.0.0.1:30303
> INFO  08-12 20:41:41 main rpc    JSON-RPC socket path: \\.\pipe\\geth.ipc
> JSONRPC Admin Session Key: hUlrVuzBFoo=
> INFO  08-12 20:41:41 p2p  net    ENR updated: [ seq=2 id=v4 key=0290a043… ip=71.227.153.208 tcp=30303 udp=30303 ]
> INFO  08-12 20:41:42 p2p  sync   Starting full sync
> ERROR 08-12 20:41:54 p2p  rlpx   egress Internal error in handshake: RLPXFrameCoder disappeared (##5a85cb59…@35.181.105.86:30303)
> ERROR 08-12 20:41:54 p2p  rlpx   egress Internal error in handshake: RLPXFrameCoder disappeared (##e31b2dbe…@40.115.25.33:30303)
> ERROR 08-12 20:41:54 p2p  rlpx   egress Internal error in handshake: RLPXFrameCoder disappeared (##5fb5c839…@188.214.30.221:30415)                                                                                                            
> ```
> 
> These appear to be transient, will dig deeper to see if they are related to my changes (I don't think they are since I didn't touch anything related to handshakes but I haven't seen these errors before)
> 
> [Edit] I haven't been able to replicate this - @gumb0 : Have you seen this before?

I investigated and I'm fairly confident that my changes aren't causing this issue. It looks like this error can happen if writing the hello packet during the RLPX handshake fails - we use the `RLPXFrameCoder `to read/write frames after exchanging the auths/acks with the key material, and we start using it to queue a write hello packet, followed by queueing a read of an incoming hello packet. If the write fails then we will call `transition `with a boost error code which will call into `error`, which in turn will call into `cancel `which will destroy the `RLPXFrameCoder `instance. When the read handler is finally executed - assuming that it has no boost error code - then the `if (!m_io) check` will be true which will result in the error message being logged: https://github.com/ethereum/aleth/blob/78355a28da55f1c0e952e3cf1a17891cea981cb4/libp2p/RLPxHandshake.cpp#L354-L370

halfalicious:
Potential optimization - EthereumCapability::propagateNewBlocks assumes that if a peer doesn't know about the most recent block in the set of new blocks, it doesn't know about any of them which results in the whole set of blocks being sent to the peer (this is existing logic): https://github.com/ethereum/aleth/blob/0775de0d43dbb6a5b9f124c37fde5f293b9b0eef/libethereum/EthereumCapability.cpp#L977-L978

I think this assumption is incorrect - we should check if a peer knows about a block on a block-by-block basis and only send new blocks to peers who don't know about them.

libethereum/BlockChain.h
```diff
@@ -118,8 +101,17 @@ class BlockChain
...
+    /// and dead blocks as well as a list of imported transactions. The second is a bool which is
+    /// true iff there are additional blocks to be processed. The third is the imported block count.
+    std::tuple<ImportRoute, bool, unsigned> sync(
+        BlockQueue& _bq, OverlayDB const& _stateDB, unsigned _max);
+
+    /// Import the supplied blocks into the chain. Blocks should be processed in order.
+    /// @returns a tuple with two members - the first (ImportRoute) contains fresh blocks, dead
+    /// blocks and imported transactions. The second contains the imported block count.
+    std::tuple<ImportRoute, unsigned> sync(
+        VerifiedBlocks const& _blocks, h256s& o_badBlockHashes, OverlayDB const& _stateDB);
```
gumb0: I think instead output param we should better return bad blocks as another member of the returned tuple

libethereum/EthereumCapability.cpp
```diff
@@ -206,13 +212,14 @@ class EthereumHostData: public EthereumHostDataFace
                         top = n + step * (numHeadersToSend - 1);
                     }
                     assert(top <= lastBlock && "invalid top block calculated");
-                    blockHash = m_chain.numberHash(static_cast<unsigned>(top)); // override start block hash with the hash of the top block we have
+                    blockHash = m_chain.numberHash(static_cast<unsigned>(
+                        top));  // override start block hash with the hash of the top block we have
```
gumb0: I don't know why you reformatted this and some lines below, but then it's better to move the comment to a separate line above, to make it look nicer
(applies to some cases below in this file, too)

libethereum/EthereumCapability.cpp
```diff
@@ -509,76 +540,48 @@ std::pair<std::vector<NodeID>, std::vector<NodeID>> EthereumCapability::randomPa
...
-            auto const peersWithoutBlock = selectPeers(
-                [&](EthereumPeer const& _peer) { return !_peer.isBlockKnown(_currentHash); });
-
-            auto const peersToSendNumber =
-                std::max<std::size_t>(c_minBlockBroadcastPeers, std::sqrt(m_peers.size()));
+    // Send any new block hashes
+    auto const detailsFrom = m_chain.details(m_latestBlockHashSent);
+    auto const detailsTo = m_chain.details(_currentHash);
+    if (detailsFrom.totalDifficulty >= detailsTo.totalDifficulty ||
+        // don't be sending more than c_maxSendNewBlocksCount "new" block hashes. if there are any
```
gumb0: Formatting: it would look nicer as two ifs with early returns
```
if (...)
  return;

// comment
if (...)
  return;
```

libethereum/BlockChain.h
```diff
@@ -118,8 +101,17 @@ class BlockChain
     void process();
 
     /// Sync the chain with any incoming blocks. All blocks should, if processed in order.
-    /// @returns fresh blocks, dead blocks and true iff there are additional blocks to be processed waiting.
-    std::tuple<ImportRoute, bool, unsigned> sync(BlockQueue& _bq, OverlayDB const& _stateDB, unsigned _max);
+    /// @returns tuple with three members - the first (ImportRoute) contains hashes of fresh blocks
+    /// and dead blocks as well as a list of imported transactions. The second is a bool which is
+    /// true iff there are additional blocks to be processed. The third is the imported block count.
+    std::tuple<ImportRoute, bool, unsigned> sync(
```
gumb0: Is this still used?

libethereum/EthereumCapability.cpp
halfalicious: @gumb0 😀 https://github.com/ethereum/aleth/pull/5713#issuecomment-518047367

libethereum/EthereumCapability.cpp
```diff
@@ -375,7 +386,23 @@ class EthereumHostData: public EthereumHostDataFace
...
+    if (_peers.size() <= _count || _peers.empty())
+        return _peers;
+
+    std::vector<NodeID> peers{_peers};
+    std::vector<NodeID> randomPeers;
+    while (_count-- && !peers.empty())
+    {
+        auto const cIter = peers.begin() + randomNumber(0, peers.size() - 1);
+        randomPeers.push_back(*cIter);
+        peers.erase(cIter);
```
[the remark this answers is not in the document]
halfalicious: Ah, good point, thanks for calling this out!

libdevcore/Common.h
```diff
@@ -317,4 +317,6 @@ class ExitHandler
 
 bool isTrue(std::string const& _m);
 bool isFalse(std::string const& _m);
+
+int randomNumber(int _min = INT_MIN, int _max = INT_MAX);
```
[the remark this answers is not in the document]
halfalicious: Actually it's no longer needed since I can just shuffle the peers in random peers

libethereum/BlockChain.h
halfalicious: @gumb0 : Yes, this overload is still called in a couple of places (which is why I left it in 😀 )...I forget the exact locations, I can dig them up if you're interested.

libethereum/BlockChain.h
halfalicious: Good idea, I'll do that

halfalicious:
Rebased (to address CHANGELOG conflict) and addressed PR feedback

halfalicious:
cc @gumb0

libethereum/EthereumCapability.cpp
```diff
@@ -951,6 +940,58 @@ void EthereumCapability::removeSentTransactions(std::vector<h256> const& _txHash
...
+                m_host->prep(peerID, name(), ts, NewBlockPacket, 2)
+                    .appendRaw(b.blockData, 1)
+                    .append(b.verified.info.difficulty());
+
+                auto itPeer = m_peers.find(peerID);
+                if (itPeer != m_peers.end())
+                {
+                    m_host->sealAndSend(peerID, ts);
+                    // We don't want to send new block hashes to these same peers
+                    itPeer->second.markBlockAsKnown(b.verified.info.hash());
```
gumb0: There was `clearKnownBlocks()` instead of this in the old code, do you have explanation for this change?

libethereum/Client.cpp
```diff
@@ -377,18 +381,27 @@ void Client::appendFromBlock(h256 const& _block, BlockPolarity _polarity, h256Ha
...
 {
 //  cdebug << "syncBlockQueue()";
 
     ImportRoute ir;
+    h256s badBlockHashes;
     unsigned count;
     Timer t;
-    tie(ir, m_syncBlockQueue, count) = bc().sync(m_bq, m_stateDB, m_syncAmount);
+
+    std::shared_ptr<VerifiedBlocks> verifiedBlocks = std::make_shared<VerifiedBlocks>();
```
gumb0: Could you add an explanation comment why we need a shared_ptr (we need to keep these blocks around until they're later processed in the network thread)

libethereum/EthereumCapability.cpp
```diff
@@ -492,93 +507,65 @@ vector<NodeID> EthereumCapability::selectPeers(
...
-    part1.erase(part1.begin() + _number, part1.end());
-    return std::make_pair(move(part1), move(part2));
+    std::vector<NodeID> peers{_peers};
+    if (peers.empty())
+        return peers;
+
+    if (_count > peers.size())
+        _count = peers.size();
+    auto itEnd = peers.begin() + _count;
+    std::shuffle(peers.begin(), itEnd, m_urng);
```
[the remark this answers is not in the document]
halfalicious: Ah, good idea, more entropy that way

libethereum/Client.cpp
halfalicious: Ah good suggestion, I had already left a comment in this PR but it would be nice to also have a comment in the code.

libethereum/EthereumCapability.cpp
```diff
@@ -492,93 +507,65 @@ vector<NodeID> EthereumCapability::selectPeers(
...
-            std::vector<NodeID> peersToAnnounce;
-            std::tie(peersToSend, peersToAnnounce) =
-                randomPartitionPeers(peersWithoutBlock, peersToSendNumber);
+    if (detailsFrom.totalDifficulty >= detailsTo.totalDifficulty)
+        return;
+    
+    // don't be sending more than c_maxSendNewBlocksCount "new" block hashes. if there are any
+    // more we were probably waaaay behind.
+    if (diff(detailsFrom.number, detailsTo.number) > c_maxSendNewBlocksCount)
+        return;
```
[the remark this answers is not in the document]
halfalicious: Thank you for calling this out, I had missed that in the original code but that makes total sense - if we don't update m_latestBlockSent the "gap" between our latest block sent record and chain head will continue to grow and we will never send out new blocks.

libethereum/EthereumCapability.cpp
halfalicious: Yes, this change was intentional - in the old code (`EthereumCapability::maintainBlocks`), we used the "known block" list in each peer to avoid sending new blocks and hashes to peers who already know about the block i.e. peers from whom we received the new block or block hash. We would clear this list after each iteration of sending out new blocks / hashes because we no longer need to track this information for that batch of blocks (since they've already been propagated and will not be re-propagated). 

In the new code, we can't clear this information after propagating new blocks since we still need to use it when we send out new block hashes (in `EthereumCapability::maintainBlockHashes`) to avoid sending hashes to peers who already know about the block.

One thing I'm not sure about is why - in the original code in `EthereumCapability::maintainBlocks` - we cleared a peer's know blocks list after sending them each block, rather than waiting until we've propagated all of the blocks to them:https://github.com/ethereum/aleth/blob/7e6095dc3d6c9cefac4b0867f60d06f955ab7c7e/libethereum/EthereumCapability.cpp#L539-L553 

 I don't see any benefit to doing this or any downside from not doing this...we only use the known blocks list to determine which peers to send blocks to in `maintainBlocks `so it doesn't have anything to do with synchronizing actions across different threads.

libethereum/EthereumCapability.cpp
```diff
@@ -492,93 +507,65 @@ vector<NodeID> EthereumCapability::selectPeers(
...
-
-    // Remove elements from the end of the shuffled part1 vector and move them to part2.
-    std::move(part1.begin() + _number, part1.end(), std::back_inserter(part2));
-    part1.erase(part1.begin() + _number, part1.end());
-    return std::make_pair(move(part1), move(part2));
+    std::vector<NodeID> peers{_peers};
+    if (peers.empty())
+        return peers;
+
+    if (_count > peers.size())
```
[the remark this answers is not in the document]
halfalicious: I was thinking that there could be some benefit to this, but after thinking about it some more there isn't any - the order in which we send blocks to peers doesn't matter since they get all of the new blocks anyway.

halfalicious:
Rebased

gumb0:
> I think this assumption is incorrect - we should check if a peer knows about a block on a block-by-block basis and only send new blocks to peers who don't know about them.

Yeah I think it would be reasonable, but for now I'd avoid further complicating this area of code. It's difficult to test, I'm not even sure whether the current logic works as expected.

PR #5714, opened 2019-08-05
increase test block mining time limit

winsvega:
allow preCompiled configs with code,nonce,storage in genesis config


libethereum/ValidationSchemes.cpp
```diff
@@ -118,6 +118,9 @@ void validateAccountObj(js::mObject const& _obj)
         requireJsonFields(_obj, "validateAccountObj",
             {{c_precompiled, {{js::obj_type}, JsonFieldPresence::Required}},
                 {c_wei, {{js::str_type}, JsonFieldPresence::Optional}},
+                {c_code, {{js::str_type}, JsonFieldPresence::Optional}},
```
gumb0: Is code really possible?

winsvega: I found out that vitalik added some tests where all fields are set for all precompiles in the state test pre state. that's to cover the ambiguous scenario where precompile is empty

PR #5718, opened 2019-08-16
Selfdestruct optimizations for Frontier and Homestead

### LINKED ISSUE #5682: Aleth-interpreter accesses states 2 times in simple SELFDESTRUCT
This is similar issue as with checking account existence in calls.

For Frontier, the SELFDESTRUCT is accessing state 2 unnecessary times - one to check balance, one to check destination account existence.

Fix also the Legacy VM.

halfalicious:
Fix #5682 

Since self-destruct operations have no gas cost on Frontier and Homestead, avoid checking the contract balance and for account existence.


libaleth-interpreter/VM.cpp
```diff
@@ -380,17 +380,21 @@ void VM::interpretCases()
...
             {
-                // After EIP150 hard fork charge additional cost of sending
-                // ethers to non-existing account.
-                int destinationExists =
-                    m_context->host->account_exists(m_context, &destination);
-                if (m_rev >= EVMC_TANGERINE_WHISTLE && !destinationExists)
-                    m_runGas += VMSchedule::callNewAccount;
+                // After EIP158 zero-value suicides do not have to pay account creation gas.
+                u256 const balance =
+                    fromEvmC(m_context->host->get_balance(m_context, &m_message->destination));
```
chfast: This is still "wrong", because `get_balance()` will be executed in case of `m_rev == EVMC_TANGERINE_WHISTLE` where its value does not matter.

halfalicious: @chfast Ah thought that the changes should apply to Homestead / Frontier, I'll take another look at which hard forks the EIPs were enabled on. According to this they were both enabled on Tangerine Whistle but that's not right if we don't care about the balance on Tangerine Whistle: https://ethereum.stackexchange.com/questions/13014/please-provide-a-summary-of-the-ethereum-hard-forks

chfast: No need to check external sources. The logic here is correct. Just if you check the `if` at https://github.com/ethereum/aleth/pull/5718/files#diff-4a2862de6e660dc5045835d8879be719R389 you will see that:
1. you can swap the checks in `||`.
2. The balance has be checked only if the first condition is false.

halfalicious: @chfast I'm confused...the line that you're referring to comes after balance retrieval so changing the checks will have no impact on whether or not we check the balance on Tangerine Whistle? 

Line in question: https://github.com/ethereum/aleth/blob/fcb1f3d4d953896a4114a6cd9e77f98826437926/libaleth-interpreter/VM.cpp#L389

halfalicious: @chfast I'm still a little confused about this part of the aleth-interpreter changes: https://github.com/ethereum/aleth/blob/29ed24eddb9e40cc450259916c7892f0685fa8e0/libaleth-interpreter/VM.cpp#L383-L388

The check for EVMC_TANGERINE_WHISTLE on line 387 implies that on that hard fork, all suicides (0-value or otherwise) have an account-creation charge. This implies that EIP158 (which removes the account-creation charge for 0-value selfdestructs) was included in Spurious Dragon, but the following says that it was included in Tangerine Whistle: https://ethereum.stackexchange.com/questions/13014/please-provide-a-summary-of-the-ethereum-hard-forks

So to summarize, EIP158 was included in Spurious Dragon and not Tangerine Whistle?

libaleth-interpreter/VM.cpp
```diff
@@ -380,17 +380,20 @@ void VM::interpretCases()
...
-                    m_context->host->account_exists(m_context, &destination);
-                if (m_rev >= EVMC_TANGERINE_WHISTLE && !destinationExists)
-                    m_runGas += VMSchedule::callNewAccount;
+                // After EIP158 zero-value suicides do not have to pay account creation gas.
+                if (m_rev == EVMC_TANGERINE_WHISTLE ||
+                    fromEvmC(m_context->host->get_balance(m_context, &m_message->destination)) > 0)
+                {
+                    // After EIP150 hard fork charge additional cost of sending
+                    // ethers to non-existing account.
+                    int const destinationExists =
```
gumb0: This should be `bool` now, maybe it was `int` because of some older version of EVMC

gumb0: or could be just inside `if (...)` because now it has only this condition and so is simpler

libaleth-interpreter/VM.cpp
gumb0: EIP-158 was activated at block 2'675'000 which is Spurious Dragon https://github.com/ethereum/aleth/blob/7e6095dc3d6c9cefac4b0867f60d06f955ab7c7e/libethashseal/genesis/mainNetwork.cpp#L29

in EIPs actually EIP-161 supercedes EIP-158 and is listed in meta-EIP https://eips.ethereum.org/EIPS/eip-607

halfalicious: @gumb0 Thanks for the info! Is there a reliable list that you can recommend of hard forks + the block numbers that they occurred at?

halfalicious:
cc @chfast  / @gumb0

libaleth-interpreter/VM.cpp
gumb0: I think meta-EIPs should correctly reflect the reality of mainnet https://eips.ethereum.org/meta

libevm/LegacyVM.cpp
```diff
@@ -331,15 +331,19 @@ void LegacyVM::interpretCases()
...
-                // After EIP150 hard fork charge additional cost of sending
-                // ethers to non-existing account.
-                if (m_schedule->suicideChargesNewAccountGas() && !m_ext->exists(dest))
-                    m_runGas += m_schedule->callNewAccountGas;
+
+            Address const dest = asAddress(m_SP[0]);
+            // Starting with EIP150, self-destructs need to pay both gas cost and account creation
+            // gas cost. Starting with EIP158, 0-value self-destructs don't need to pay this account
+            // creation cost.
+            if (m_schedule->zeroValueTransferChargesNewAccountGas() ||
```
gumb0: aleth-interpreter part is quite clear, but I'm not the fan of the change in LegacyVM.

I think we can assume that eip158 is always activated not earlier than eip150, and then simplify it to very similar to interpterter's version:
```
if (m_schedule->suicideChargesNewAccountGas())
{
  if (m_schedule->zeroValueTransferChargesNewAccountGas() || m_ext->balance(m_ext->myAddress) > 0)
  {
     if (!m_ext->exists(dest))
       m_runGas += m_schedule->callNewAccountGas;
  }
}
```

gumb0: Maybe add `assert(m_schedule->zeroValueTransferChargesNewAccountGas() || m_schedule->suicideChargesNewAccountGas())` - this was your first condition, and it should be always true when eip158 comes after eip150

libaleth-interpreter/VM.cpp
```diff
@@ -377,21 +377,19 @@ void VM::interpretCases()
             if (m_message->flags & EVMC_STATIC)
                 throwDisallowedStateChange();
 
+            // Self-destructs only have gas cost starting with Tangerine Whistle
             m_runGas = m_rev >= EVMC_TANGERINE_WHISTLE ? 5000 : 0;
```
chfast: This line is gone. Please rebase.

libevm/LegacyVM.cpp
gumb0: Agree that using eips would be simpler, but 
> This assumes that m_schedule->eip150Mode will not be true for the Spurious Dragon schedule.

this assumption is wrong, because `eip150Mode` is true for Tangerine Whistle and further, so including Spurious Dragon.

So I think it should be 
```
            if (m_schedule->eip150Mode && (!m_schedule->eip158Mode || m_ext->balance(m_ext->myAddress) > 0))
            {
                if (!m_ext->exists(dest))
                    m_runGas += m_schedule->callNewAccountGas;
            }
```

or leave it as 3 nested ifs as I suggested above similar to interpreter version.

PR #5721, opened 2019-08-21
docs(usingtesteth): can provide absolute path via --singletest

yxliang01:
As per https://github.com/ethereum/aleth/issues/5420 , suggested by @winsvega , we can provide absolute path as an argument to `--singletest` option.

### DIFF
--- a/doc/usingtesteth.rst
+++ b/doc/usingtesteth.rst
@@ -42,9 +42,12 @@ Or just the test suite:
 
    ``./testeth -t <TEST_SUITE>``
    
-To run a specific test from the test case:
+To run a specific test from a test case (reference by name):
 
    ``./testeth -t <TEST_SUITE>/<TEST_CASE> -- --singletest <TEST_NAME>``
+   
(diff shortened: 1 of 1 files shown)


winsvega:
Could you please add this example line to the manual too

'''
./testeth -t GeneralStateTests -- --singletest /home/wins/Ethereum/tests/GeneralStateTests/stCreate2/create2noCash.json create2noCash --verbosity 5
'''

yxliang01:
@winsvega How's the example just added? I added a command with placeholder, instead of your suggested concrete example because it would be rather weird if all examples are with placeholders but only this one is not? Also `--verbosity` option is removed as it's irrelevant.

winsvega:
It has to be --singletest testname 
Or --singletest filepath testname

yxliang01:
@winsvega Yes. Version at latest commit c235c3d agrees with your last comment?

winsvega:
Ah. Looks like so. Yes.

yxliang01:
Great. Looks like I'm done for this PR then. :+1:

PR #5723, opened 2019-08-23
Updated Visual Studio notes in README

halfalicious:
Added generator command for each supported VS release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,14 +89,15 @@ cmake ..               # Configure the project.
 cmake --build .        # Build all default targets.
 ```
 
-On **Windows** we support up to Visual Studio 2017. You should generate a Visual Studio solution file (.sln) for the 64-bit architecture by adding
-`-G "Visual Studio 15 2017 Win64"` argument to the CMake configuration command.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,14 +89,20 @@ cmake ..               # Configure the project.
 cmake --build .        # Build all default targets.
 ```
 
-On **Windows** we support up to Visual Studio 2017. You should generate a Visual Studio solution file (.sln) for the 64-bit architecture by adding
-`-G "Visual Studio 15 2017 Win64"` argument to the CMake configuration command.
-After configuration is completed, the `aleth.sln` can be found in the
-`build` directory.
+On **Windows** we support up to Visual Studio 2019. You should generate a Visual Studio solution file (.sln) for the 64-bit architecture by adding the following argument to the CMake configuration command:
 
+* **Visual Studio 2015**: `-G "Visual Studio 14 2015 Win64"`
```
gumb0: I'd better just listed full commands (with `cmake ..`) for all three cases

halfalicious:
cc @gumb0

PR #5728, opened 2019-08-28
Implement EIP-2200: Structured Definitions for Net Gas Metering in aleth-interpreter

gumb0:
https://github.com/ethereum/EIPs/pull/2200

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
 - Added: [#5701](https://github.com/ethereum/aleth/issues/5701) Outputs ENR text representation in admin.nodeInfo RPC.
 - Added: [#5705](https://github.com/ethereum/aleth/pull/5705) Istanbul support: EIP 1108 Reduce alt_bn128 precompile gas costs.
 - Added: [#5707](https://github.com/ethereum/aleth/pull/5707) Aleth waits for 2 seconds after sending disconnect to peer before closing socket.
-- Added: [#5709](https://github.com/ethereum/aleth/pull/5709) Istanbul support: EIP-2200 Structured Definitions for Net Gas Metering.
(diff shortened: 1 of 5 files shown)


winsvega:
waiting this PR to update the blockchain tests and check aleth consensus with geth and besu

payvint:
Hi guys, could you please explain me where is this constants are located?
https://gist.github.com/karalabe/adc43c07db9f03be82093cd5466562b0#specification
I have a little bit got lost in code

gumb0:
@payvint there might be not exact one-to-one correspondence between the constants in the spec and constants defined in our code, but for LegacyVM they are in https://github.com/ethereum/aleth/blob/master/libethcore/EVMSchedule.h
and for aleth-interpreter they are in https://github.com/ethereum/aleth/blob/master/libaleth-interpreter/VM.h

gumb0:
@payvint Our implementation I think is closer to the spec definition in the current version of the EIP at https://github.com/ethereum/EIPs/pull/2200, than to the one at your link (but I guess they are effectively equivalent)

payvint:
Thanks @gumb0, I will take a look

payvint:
So, @gumb0 , thank you for links very much. I 've just figured it out. But I have one question:
https://gist.github.com/karalabe/adc43c07db9f03be82093cd5466562b0#specification
ethereum/EIPs#2200
In last version - constants are changed and optimized from 8 new to 1 changes in existing implementation, but you changed a constant named SLOAD_GAS - does it related to existence one?
SLOAD_GAS was 200 and became 800 - does it mean that opcode sload - costs 4 times more?
if yes - it is not mentioned in EIP, as I read. if no - please show me where could I find proofs?:)
Thanks

gumb0:
@payvint that's right, SLOAD cost has changed, and it's the subject of another Istanbul EIP https://eips.ethereum.org/EIPS/eip-1884

payvint:
@gumb0 Thank you very much. Figured out everything!!!

PR #5729, opened 2019-08-31
Enable Istanbule on TransactionTests and Update the tests

winsvega:
* Enables Istanbul for transaction tests
* Update the test repo
    - All StateTests from <=constantinople are disabled (moved to LegacyTests folder)
    - The BCGeneralStateTests outdated and removed
    - Recent tests and updates to GeneralStateTests supported by geth + retesteth

### DIFF
--- a/test/tools/jsontests/TransactionTests.cpp
+++ b/test/tools/jsontests/TransactionTests.cpp
@@ -59,6 +59,9 @@ mObject getExpectSection(mValue const& _expect, eth::Network _network)
         if (networks.count(test::netIdToString(_network)))
             objVector.push_back(obj);
     }
+    if (objVector.size() == 0)
+        BOOST_ERROR("Expect network '" + test::netIdToString(_network) +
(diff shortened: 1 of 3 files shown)


test/tools/jsontests/TransactionTests.cpp
```diff
@@ -59,6 +59,9 @@ mObject getExpectSection(mValue const& _expect, eth::Network _network)
         if (networks.count(test::netIdToString(_network)))
             objVector.push_back(obj);
     }
+    if (objVector.size() == 0)
```
gumb0: Why not `BOOST_REQUIRE_MESSAGE(objVector.size() == 0, msg)`, as below?

gumb0:
@chfast The tests submodule update here disables testing all forks before Constantinople, and because of this I don't want to merge it like this.

I suggested to @winsvega to try running old forks tests via retesteth+aleth on CI.
(this is going to make it slower, but might be a better way forward in case testeth deprecation is planned)

PR #5735, opened 2019-09-04
support LegacyTests in testeth

winsvega:
update the test subrepo to latest develop
support testeth to run LegacyTests (all tests for forks <= Constantinople. inc. GeneralStateTests and its blockchain form at the moment)

```
./testeth -t LegacyTests -- --all
```

### DIFF
--- a/test/tools/jsontests/BlockChainTests.h
+++ b/test/tools/jsontests/BlockChainTests.h
@@ -53,6 +53,13 @@ class BCGeneralStateTestsSuite : public BlockchainValidTestSuite
     boost::filesystem::path suiteFillerFolder() const override;
 };
 
+class BCGeneralStateTestsSuiteLegacyConstantinople : public BCGeneralStateTestsSuite
+{
+    boost::filesystem::path suiteFolder() const override;
+    boost::filesystem::path suiteFillerFolder() const override;
+};
+
+
 class TransitionTestsSuite: public TestSuite
 {
     json_spirit::mValue doTests(json_spirit::mValue const& _input, bool _fillin) const override;
--- a/test/tools/jsontests/LegacyTestsBoost.cpp
+++ b/test/tools/jsontests/LegacyTestsBoost.cpp
@@ -1,0 +1,220 @@
+#include <test/tools/jsontests/StateTests.h>
+#include <test/tools/jsontests/BlockChainTests.h>
+#include <test/tools/libtesteth/TestHelper.h>
+
+#include <boost/test/unit_test.hpp>
+
+using namespace dev::test;
+namespace fs = boost::filesystem;
+
+fs::path StateTestSuiteLegacyConstantinople::suiteFolder() const
+{
+    if (Options::get().fillchain || Options::get().filltests)
+        BOOST_FAIL("Legacy tests are sealed and not refillable!");
+    return "LegacyTests/Constantinople/GeneralStateTests";
+}
+
+fs::path StateTestSuiteLegacyConstantinople::suiteFillerFolder() const
+{
+    return "LegacyTests/Constantinople/GeneralStateTestsFiller";
+}
+
+fs::path BCGeneralStateTestsSuiteLegacyConstantinople::suiteFolder() const
+{
+    if (Options::get().filltests)
+        BOOST_FAIL("Legacy tests are sealed and not refillable!");
+    return "LegacyTests/Constantinople/BlockchainTests/GeneralStateTests";
+}
+
+fs::path BCGeneralStateTestsSuiteLegacyConstantinople::suiteFillerFolder() const
+{
+    return "LegacyTests/Constantinople/BlockchainTestsFiller/GeneralStateTests";
+}
+
+
+class LegacyConstantinopleGeneralStateTestFixture
+{
+public:
+    LegacyConstantinopleGeneralStateTestFixture()
+    {
+        StateTestSuiteLegacyConstantinople suite;
+        string casename = boost::unit_test::framework::current_test_case().p_name;
+        boost::filesystem::path suiteFillerPath = suite.getFullPathFiller(casename).parent_path();
+        if (!test::Options::get().all)
+        {
+            std::cout << "Skipping " << casename << " because --all option is not specified.\n";
+            test::TestOutputHelper::get().markTestFolderAsFinished(suiteFillerPath, casename);
+            return;
+        }
+        suite.runAllTestsInFolder(casename);
+        test::TestOutputHelper::get().markTestFolderAsFinished(suiteFillerPath, casename);
+    }
+};
+
+class LegacyConstantinopleBCGeneralStateTestFixture
+{
+public:
+    LegacyConstantinopleBCGeneralStateTestFixture()
+    {
+        BCGeneralStateTestsSuiteLegacyConstantinople suite;
+        string const casename = boost::unit_test::framework::current_test_case().p_name;
+        boost::filesystem::path suiteFillerPath = suite.getFullPathFiller(casename).parent_path();
+
+        //skip this test suite if not run with --all flag (cases are already tested in state tests)
+        if (!test::Options::get().all)
+        {
+            cnote << "Skipping hive test " << casename << ". Use --all to run it.\n";
+            test::TestOutputHelper::get().markTestFolderAsFinished(suiteFillerPath, casename);
+            return;
+        }
+        suite.runAllTestsInFolder(casename);
+        test::TestOutputHelper::get().markTestFolderAsFinished(suiteFillerPath, casename);
+    }
+};
+
+
+BOOST_AUTO_TEST_SUITE(LegacyTests)
+    BOOST_AUTO_TEST_SUITE(Constantinople)
+        BOOST_FIXTURE_TEST_SUITE(GeneralStateTests, LegacyConstantinopleGeneralStateTestFixture)
+            // Frontier Tests
+            BOOST_AUTO_TEST_CASE(stCallCodes) {}
+            BOOST_AUTO_TEST_CASE(stCallCreateCallCodeTest) {}
+            BOOST_AUTO_TEST_CASE(stExample) {}
+            BOOST_AUTO_TEST_CASE(stInitCodeTest) {}
+            BOOST_AUTO_TEST_CASE(stLogTests) {}
+            BOOST_AUTO_TEST_CASE(stMemoryTest) {}
+            BOOST_AUTO_TEST_CASE(stPreCompiledContracts) {}
+            BOOST_AUTO_TEST_CASE(stPreCompiledContracts2) {}
+            BOOST_AUTO_TEST_CASE(stRandom) {}
+            BOOST_AUTO_TEST_CASE(stRandom2) {}
+            BOOST_AUTO_TEST_CASE(stRecursiveCreate) {}
+            BOOST_AUTO_TEST_CASE(stRefundTest) {}
+            BOOST_AUTO_TEST_CASE(stSolidityTest) {}
+            BOOST_AUTO_TEST_CASE(stSpecialTest) {}
+            BOOST_AUTO_TEST_CASE(stSystemOperationsTest) {}
+            BOOST_AUTO_TEST_CASE(stTransactionTest) {}
+            BOOST_AUTO_TEST_CASE(stTransitionTest) {}
+            BOOST_AUTO_TEST_CASE(stWalletTest) {}
+
+            // Homestead Tests
+            BOOST_AUTO_TEST_CASE(stCallDelegateCodesCallCodeHomestead) {}
+            BOOST_AUTO_TEST_CASE(stCallDelegateCodesHomestead) {}
+            BOOST_AUTO_TEST_CASE(stHomesteadSpecific) {}
+            BOOST_AUTO_TEST_CASE(stDelegatecallTestHomestead) {}
+
+            // EIP150 Tests
+            BOOST_AUTO_TEST_CASE(stChangedEIP150) {}
+            BOOST_AUTO_TEST_CASE(stEIP150singleCodeGasPrices) {}
+            BOOST_AUTO_TEST_CASE(stMemExpandingEIP150Calls) {}
+            BOOST_AUTO_TEST_CASE(stEIP150Specific) {}
+
+            // EIP158 Tests
+            BOOST_AUTO_TEST_CASE(stEIP158Specific) {}
+            BOOST_AUTO_TEST_CASE(stNonZeroCallsTest) {}
+            BOOST_AUTO_TEST_CASE(stZeroCallsTest) {}
+            BOOST_AUTO_TEST_CASE(stZeroCallsRevert) {}
+            BOOST_AUTO_TEST_CASE(stCodeSizeLimit) {}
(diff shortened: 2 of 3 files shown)


test/tools/jsontests/LegacyTestsBoost.cpp
```diff
@@ -0,0 +1,220 @@
...
+            return;
+        }
+        suite.runAllTestsInFolder(casename);
+        test::TestOutputHelper::get().markTestFolderAsFinished(suiteFillerPath, casename);
+    }
+};
+
+
+BOOST_AUTO_TEST_SUITE(LegacyTests)
+    BOOST_AUTO_TEST_SUITE(Constantinople)
```
gumb0: Are these testing only Constantinople?

gumb0: Ah I see it repeats tests repo structure. Why is the subdir called Constantinople then?

test/tools/jsontests/LegacyTestsBoost.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+#include <boost/test/unit_test.hpp>
+
+using namespace dev::test;
+namespace fs = boost::filesystem;
+
+fs::path StateTestSuiteLegacyConstantinople::suiteFolder() const
+{
+    if (Options::get().fillchain || Options::get().filltests)
+        BOOST_FAIL("Legacy tests are sealed and not refillable!");
```
gumb0: That's not a great idea to throw from a const getter method. If you really need this, you should check this on the calling side.

test/tools/jsontests/LegacyTestsBoost.cpp
winsvega: Because it's Constantinople tests snapshot. Meaning all the tests we had at the moment of Constantinople. It actually includes ConstantinopleFix also which I count as Constantinople as well.

test/tools/jsontests/LegacyTestsBoost.cpp
winsvega: which would be a test fixture. no template then

test/tools/jsontests/BlockChainTests.h
```diff
@@ -53,6 +53,13 @@ class BCGeneralStateTestsSuite : public BlockchainValidTestSuite
     boost::filesystem::path suiteFillerFolder() const override;
 };
 
+class BCGeneralStateTestsSuiteLegacyConstantinople : public BCGeneralStateTestsSuite
```
gumb0: Why not put this in a separate header for Legacy tests?

It's difficult now to find what is where: usual state tests are in StateTests.h + StateTests.cpp
usual blockchain tests are in BlockChainTests.h + BlockChainTests.cpp

and now legacy state tests are in StateTests.h + LegacyTestBoost.cpp
legacy blockchain tests in BlockChainTests.h + LegacyTestBoost.cpp

I'd create a header for LegacyBoostTests and also maybe split them into separate files for state and blockchain ones.

test/tools/jsontests/LegacyTestsBoost.cpp
```diff
@@ -0,0 +1,221 @@
...
+public:
+    LegacyTestFixtureBase(bool _onlyRunIfOptionsAllSet)
+    {
+        T suite;
+        if (Options::get().fillchain || Options::get().filltests)
+            BOOST_FAIL("Legacy tests are sealed and not refillable!");
+
+        string casename = boost::unit_test::framework::current_test_case().p_name;
+        boost::filesystem::path suiteFillerPath = suite.getFullPathFiller(casename).parent_path();
+        static vector<string> const timeConsumingTestSuites{
```
gumb0: Move this into a single common place, so that it's not duplicated with regular state tests.

It's a maintenance pain to remember that it exists in several places, when there's need to modify it.

test/tools/jsontests/LegacyTestsBoost.cpp
```diff
@@ -0,0 +1,221 @@
...
+            string{"stTimeConsuming"}, string{"stQuadraticComplexityTest"}};
+
+        bool skipTheTest = false;
+        if (test::inArray(timeConsumingTestSuites, casename) && !test::Options::get().all)
+            skipTheTest = true;
+        if (_onlyRunIfOptionsAllSet && !Options::get().all)
```
gumb0: I suggest to simplify this by calling the parameter flag just `_skip` and checking `Options::get().all` on the calling side
(that is you'll have `LegacyConstantinopleGeneralStateTestFixture() : LegacyTestFixtureBase(false) {}` and `LegacyConstantinopleBCGeneralStateTestFixture() : LegacyTestFixtureBase(!test::Options::get().all) {}`

test/tools/jsontests/LegacyTestsBoost.cpp
```diff
@@ -0,0 +1,221 @@
...
+            skipTheTest = true;
+
+        if (skipTheTest)
+        {
+            std::cout << "Skipping " << casename << " because --all option is not specified.\n";
+            test::TestOutputHelper::get().markTestFolderAsFinished(suiteFillerPath, casename);
+            return;
+        }
+        suite.runAllTestsInFolder(casename);
```
gumb0: this is just a rant, we won't change it now, but this should have been called from the test body instead of weirdly from fixture constructor; for that you could have created a macro wrapping `BOOST_AUTO_TEST_CASE` macro always calling it from the body
(maybe do that in retesteth if it's any similar there)

gumb0:
I don't see legacy tests running on AppVeyor, do they run?

gumb0:
Add item to changelog

test/tools/jsontests/LegacyTestsBoost.cpp
winsvega: it is not global. StateTests has their own list which might change. LegacyStateTests has their current one. Currently Legacy tests and State tests has the same. But this might change.

test/tools/jsontests/LegacyTestsBoost.cpp
winsvega: I would not say it simplify. Sometimes it better be readable then optimized. 
I would replace bool with Enum

test/tools/jsontests/LegacyTestsBoost.cpp
winsvega: Please explain better the idea.

winsvega:
In AppVeyor it just runs `testeth` that output just the casenames without the tree structure.

test/tools/jsontests/LegacyTestsBoost.cpp
gumb0: Don't run the tests from the constructor, run them from the test body.
```
#define RETESTETH_AUTO_TEST_CASE(casename) BOOST_AUTO_TEST_CASE(casename) { m_suite.runAllTestsInFolder(casename); }
```

...anyway it's better to get rid of boost.test altogether

winsvega: I think it will get more confusing if to continue refactoring it that way

test/tools/jsontests/Common.h
```diff
@@ -0,0 +1,55 @@
+#pragma once
```
gumb0: License info missing at the top of the file

test/tools/jsontests/Common.h
```diff
@@ -0,0 +1,55 @@
...
+#include <test/tools/libtesteth/TestHelper.h>
+using namespace std;
```
gumb0: Never do `using namespace` in the headers

test/tools/jsontests/Common.h
```diff
@@ -0,0 +1,55 @@
...
+
+namespace dev
+{
+namespace test
+{
+enum class TestExecution
+{
+    REQUIRE_OPTION_ALL,
```
gumb0: enum members shouldn't be all caps, should CamelCase with first letter uppercase https://github.com/ethereum/aleth/blob/master/CODING_STYLE.md#capitalization

test/tools/jsontests/LegacyTestsBoost.cpp
```diff
@@ -0,0 +1,150 @@
+#include <test/tools/jsontests/LegacyTestsBoost.h>
```
gumb0: License info missing at the top of the file

test/tools/jsontests/Common.h
```diff
@@ -0,0 +1,55 @@
...
+    NOT_REFILLABLE,
+    STANDARD
+};
+
+
+template <class T>
+class StateTestFixtureBase
+{
+public:
+    StateTestFixtureBase(TestExecution const _execFlag)
```
gumb0: Usually we don't declare parameters passed by value as `const`

test/tools/jsontests/Common.h
```diff
@@ -0,0 +1,55 @@
...
+    {
+        T suite;
+        if ((_execFlag == TestExecution::REQUIRE_OPTION_ALL_NOT_REFILLABLE ||
+                _execFlag == TestExecution::NOT_REFILLABLE) &&
+            (Options::get().fillchain || Options::get().filltests))
+            BOOST_FAIL("Tests are sealed and not refillable!");
+
+        string casename = boost::unit_test::framework::current_test_case().p_name;
```
gumb0: can be `const`

test/tools/jsontests/Common.h
```diff
@@ -0,0 +1,55 @@
...
+        boost::filesystem::path suiteFillerPath = suite.getFullPathFiller(casename).parent_path();
+        static vector<string> const timeConsumingTestSuites{
```
gumb0: Still it would be easier to maintain, if these constants are somewhere at the top of the file

test/tools/jsontests/Common.h
gumb0: Why is the file called "Common", it should be StateTestFixtureBase.h

test/tools/jsontests/Common.h
```diff
@@ -0,0 +1,55 @@
...
+            string{"stTimeConsuming"}, string{"stQuadraticComplexityTest"}};
+
+        bool skipTheTest = false;
+        if (test::inArray(timeConsumingTestSuites, casename) && !test::Options::get().all)
+            skipTheTest = true;
+        if ((_execFlag == TestExecution::REQUIRE_OPTION_ALL ||
```
gumb0: Better could be `else if`

test/tools/jsontests/Common.h
gumb0: This enum seems to conflate two concepts together - whether the tests are run with `--all` and whether it's allowed to refill. This makes it too complicated, you should split it into two enums.

On another note, why do you even need to forbid refilling? I wouldn't be so sure that you'll never ever refill legacy tests (for example what if you decide to change the test format in non-backwards-compatible way?) You control anyway what's being merged into the tests repo, so I don't see the point of complicating this.

winsvega: This is so testeth wont accedentially corrupt the LegacyTests
using too enums looks too heavy for just this options.

gumb0: If you corrupt them locally, you can easily check them out of git again.

This is a mess, these options, I don't understand what they mean, without carefully investigating the code. Compare with:

```
enum class RefillOption
{
    Allowed,
	Forbidden
};

enum class AllOption
{
    Required,
	NotRequired
};
```
```
 if (_refill == RefillOption::Forbidden && (Options::get().fillchain || Options::get().filltests))
    fail
```

```
  if (!Options::get().all && (_all == AllOption::Required || test::inArray(c_timeConsumingTestSuites, casename))
     skip;
```

test/tools/libtesteth/TestHelper.cpp
```diff
@@ -151,25 +151,9 @@ eth::Network stringToNetId(string const& _netname)
     return eth::Network::FrontierTest;
 }
 
-bool isDisabledNetwork(eth::Network _net)
+bool isDisabledNetwork(eth::Network)
```
gumb0: I think better to delete this function.

test/tools/jsontests/StateTestFixtureBase.h
```diff
@@ -0,0 +1,71 @@
...
+public:
+    explicit StateTestFixtureBase(std::set<TestExecution> const& _execFlags)
+    {
+        T suite;
+        if (_execFlags.count(TestExecution::NotRefillable) &&
+            (Options::get().fillchain || Options::get().filltests))
+            BOOST_FAIL("Tests are sealed and not refillable!");
+
+        std::string const casename = boost::unit_test::framework::current_test_case().p_name;
+        boost::filesystem::path suiteFillerPath = suite.getFullPathFiller(casename).parent_path();
```
gumb0:
```suggestion
        boost::filesystem::path const suiteFillerPath = suite.getFullPathFiller(casename).parent_path();
```

test/tools/jsontests/StateTestFixtureBase.h
```diff
@@ -0,0 +1,71 @@
+/*
```
gumb0: replace this with 
```
// Aleth: Ethereum C++ client, tools and libraries.
// Copyright 2019 Aleth Authors.
// Licensed under the GNU General Public License, Version 3.
```

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 - Added: [#5701](https://github.com/ethereum/aleth/issues/5701) Outputs ENR text representation in admin.nodeInfo RPC.
 - Added: [#5705](https://github.com/ethereum/aleth/pull/5705) Istanbul support: EIP 1108 Reduce alt_bn128 precompile gas costs.
 - Added: [#5707](https://github.com/ethereum/aleth/pull/5707) Aleth waits for 2 seconds after sending disconnect to peer before closing socket.
+- Changed: [#5735](https://github.com/ethereum/aleth/pull/5735) Tests <= ConstantinopleFix moved to `-t LegacyTests suite` in the testeth
```
gumb0:
```suggestion
- Changed: [#5735](https://github.com/ethereum/aleth/pull/5735) `testeth -t GeneralStateTests` and `testeth -t BlockchainTests` commands now don't run the tests for the forks before Istanbul. They are moved to a separate `LegacyTests` suite, to run them use the commands `testeth -t LegacyTests/Constantinople/GeneralStateTests` and ``testeth -t LegacyTests/Constantinople/BlockchainTests -- --all`
```
please check the commands there ^

also better move it to be the last of `Changed` (to keep them sorted by id)

gumb0:
`LegacyTests/Constantinople/BCGeneralStateTests` are reported as passed by ctest, but they are actually skipped, aren't they?
But it was similar before for blockchain tests generated from state tests, right?

winsvega:
`LegacyTests/Constantinople/BCGeneralStateTests` get executed if ctest run testeth with `--all` flag. otherwise it is skipped same as stTimeConsuming tests

winsvega:
is there a command to squash commits in the history not touching the recent ones?

gumb0:
depends on what you mean by not touching
`git rebase -i HEAD~9` and set the first ones to `squash` and the last ones to `pick`
(all commits will be rewritten anyway, resulting in new commit hashes)

winsvega:
looks like I can't sqush the first unnamed commit. you could only squash following commits into previous.

PR #5738, opened 2019-09-06
Rename suicide to selfdestruct

chfast:


libethereum/Executive.cpp
```diff
@@ -561,7 +561,8 @@ bool Executive::finalize()
     if (m_ext)
     {
         // Accumulate refunds for suicides.
-        m_ext->sub.refunds += m_ext->evmSchedule().suicideRefundGas * m_ext->sub.suicides.size();
+        m_ext->sub.refunds +=
+            m_ext->evmSchedule().suicideRefundGas * m_ext->sub.selfdestructs.size();
```
gumb0: `selfdestructRefundGas`?

libevm/ExtVMFace.h
```diff
@@ -239,8 +239,15 @@ class ExtVMFace
     /// Does the account exist?
     virtual bool exists(Address) { return false; }
 
-    /// Suicide the associated contract and give proceeds to the given address.
-    virtual void suicide(Address) { sub.suicides.insert(myAddress); }
+    /// Selfdestruct the associated contract and give proceeds to the given address.
+    ///
+    /// @param beneficiary  The address of the account which will receive ETH
+    ///                     from the selfdestructed account.
+    virtual void selfdestruct(Address beneficiary)
```
halfalicious: Should we leave this arg unnamed? It would enable us to skip adding the (void)beneficiary...is there a reason to name it?

halfalicious:
Should we also change the opcode names in aleth-interpreter (e.g. in libaleth-interpreter/VM.cpp)

halfalicious:
Should we update the comments in Account.h?

libethereum/Executive.cpp
```diff
@@ -561,7 +561,8 @@ bool Executive::finalize()
     if (m_ext)
     {
         // Accumulate refunds for suicides.
```
halfalicious: What about this comment?

chfast:
I have run out of time on Friday. To be continued... I plan to also change Instructions and comments. The only "suicide" to be left are references to existing consensus tests.

PR #5740, opened 2019-09-09
Change Chain ID width to 64 bit

gumb0:
It's decided to specify ChainID as 64-bit value.
https://github.com/ethereum/EIPs/pull/2263

### DIFF
--- a/libaleth-interpreter/VM.cpp
+++ b/libaleth-interpreter/VM.cpp
@@ -1160,7 +1160,7 @@ void VM::interpretCases()
 
             updateIOGas();
 
-            m_SPP[0] = fromEvmC(getTxContext().chain_id);
+            m_SPP[0] = getTxContext().chain_id;
         }
         NEXT
 
--- a/libethereum/ChainParams.cpp
+++ b/libethereum/ChainParams.cpp
@@ -115,7 +115,7 @@ ChainParams ChainParams::loadConfig(
     setOptionalU256Parameter(cp.durationLimit, c_durationLimit);
 
     if (params.count(c_chainID))
-        cp.chainID = int(fromBigEndian<u256>(fromHex(params.at(c_chainID).get_str())));
+        cp.chainID = uint64_t(fromBigEndian<u256>(fromHex(params.at(c_chainID).get_str())));
     if (params.count(c_networkID))
         cp.networkID = int(fromBigEndian<u256>(fromHex(params.at(c_networkID).get_str())));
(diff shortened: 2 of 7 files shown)


libethereum/ChainParams.cpp
```diff
@@ -115,7 +115,7 @@ ChainParams ChainParams::loadConfig(
     setOptionalU256Parameter(cp.durationLimit, c_durationLimit);
 
     if (params.count(c_chainID))
-        cp.chainID = int(fromBigEndian<u256>(fromHex(params.at(c_chainID).get_str())));
+        cp.chainID = uint64_t(fromBigEndian<u256>(fromHex(params.at(c_chainID).get_str())));
```
halfalicious: Should we maybe do a static_cast instead (since static_casts are both easier to visually identify in code and to search for)?

libaleth-interpreter/VM.cpp
```diff
@@ -1160,7 +1160,7 @@ void VM::interpretCases()
 
             updateIOGas();
 
-            m_SPP[0] = fromEvmC(getTxContext().chain_id);
+            m_SPP[0] = getTxContext().chain_id;
```
halfalicious: How come we don't have to do the big endian conversion anymore?

halfalicious:
Should we also update ETHash e.g. ETHash::verifyTransaction: https://github.com/ethereum/aleth/blob/0faec9d18e09459462f392d710b1c0d7f5008d5f/libethashseal/Ethash.cpp#L135

libethereum/ChainParams.cpp
gumb0: Yes, I just did the same as surrounding code, but I'll change it.

libaleth-interpreter/VM.cpp
gumb0: `chain_id` changed in evmc, now it's `uint64_t` (in native system endianness) and it can be converted implicitly to `u256`

gumb0:
@halfalicious  The logic around Chain ID in the transaction is out of scope of EIP-1344, it's specced in the older EIP-155, that's why I didn't touch those parts. 

EIP-155 doesn't specify the type though, and obviously it should be the same everywhere, so now I think it's a good idea to change it in `TransactionBase`, too.

gumb0:
@halfalicious changes in `TransactionBase` require a bit of refactoring, so I'll better submit them in a separate PR

PR #5741, opened 2019-09-10
EIP-2046 independent activation

gumb0:
Addresses https://github.com/ethereum/aleth/issues/5789

It introduces support for individual EIP activation without fork, then moves EIP-2046 implementation from Berlin fork definition to this new individual model of activation.

#### Implementing new EIP
- Add new member to `AdditionalEIPs` struct https://github.com/ethereum/aleth/pull/5741/files#diff-c93cb246fbbc22300c737cb7ce2a2741R57-R60
- If EIP affects anything outside EVM, the helper `ChainOperationParams::isEIPXXXXEnabled` would be helpful https://github.com/ethereum/aleth/pull/5741/files#diff-c93cb246fbbc22300c737cb7ce2a2741R72-R75
- If EIP affects EVM, affected prices/flags should be set to `EVMSchedule` in the function `addEIPsToSchedule` https://github.com/ethereum/aleth/pull/5741/files#diff-363b4e07ecbf6ed5a7eb650e5cb69bcbR12-R19

#### Including EIP into the new fork definition, after it's accepted
- Remove EIP member from `AdditionalEIPs` struct
- If EIP affects anything outside EVM, modify `ChainOperationParams::isEIPXXXXEnabled` helper https://github.com/ethereum/aleth/pull/5741/files#diff-c93cb246fbbc22300c737cb7ce2a2741R76-R80
- If EIP affects EVM remove it from `addEIPsToSchedule` and add to the fork's `EVMSchedule` instance

#### Writing a unit test
- Instantiate `ChainParams` with additional EIPs activated using `ChainParams(string, AdditionalEIPs)` constructor https://github.com/ethereum/aleth/pull/5741/files#diff-bca8e2c5f9ec26d6b11f9fd86abd6338R815-R817

#### Writing consensus state test
- Use `Forkname+EIPXXXX` instead of regular fork name.
See example test for EIP2046 that I created https://github.com/ethereum/tests/commit/8bbbd484227684339ec64ed6a0083d0b137436c8
(based on existing `CallIdentitiy_1` test)

### Limitations / future improvements (if needed)
- Not supported in blockchain tests, transaction tests
- Not supported in generating blockchain tests from state tests
- Not supported in chain config json, only in testeth. So it's not possible to create a chain with individual EIPs activated, that also means it won't work for retesteth.

These should be fairly easy to support though.

@winsvega please review the changes in testeth. Basically it operates now with a network name string (which can possibly contain additional EIP numbers) in the places where it used to use `eth::Network`. Then creates `ChainParams` based on this string, when it needs to execute the transaction.

### DIFF
--- a/libethcore/ChainOperationParams.cpp
+++ b/libethcore/ChainOperationParams.cpp
@@ -6,7 +6,6 @@
 #include "ChainOperationParams.h"
 #include <libdevcore/CommonData.h>
 #include <libdevcore/Log.h>
-#include <libethcore/EVMSchedule.h>
 
 using namespace std;
 using namespace dev;
@@ -38,17 +37,13 @@ ChainOperationParams::ChainOperationParams():
 
 EVMSchedule const& ChainOperationParams::scheduleForBlockNumber(u256 const& _blockNumber) const
 {
-    EVMSchedule const& fixedForkSchedule = constScheduleForBlockNumber(_blockNumber);
-
-    if (_blockNumber < lastForkBlock)
-        return fixedForkSchedule;
-
-    static EVMSchedule const lastForkWithAdditionalEIPsSchedule =
-        addEIPs(fixedForkSchedule, lastForkAdditionalEIPs);
-    return lastForkWithAdditionalEIPsSchedule;
+    if (_blockNumber >= lastForkBlock)
+        return lastForkWithAdditionalEIPsSchedule;
+    else
+        return fixedScheduleForBlockNumber(_blockNumber);
 }
 
-EVMSchedule const& ChainOperationParams::constScheduleForBlockNumber(u256 const& _blockNumber) const
+EVMSchedule const& ChainOperationParams::fixedScheduleForBlockNumber(u256 const& _blockNumber) const
 {
     if (_blockNumber >= experimentalForkBlock)
         return ExperimentalSchedule;
--- a/libethcore/ChainOperationParams.cpp
+++ b/libethcore/ChainOperationParams.cpp
@@ -4,8 +4,10 @@
 
 
 #include "ChainOperationParams.h"
-#include <libdevcore/Log.h>
 #include <libdevcore/CommonData.h>
+#include <libdevcore/Log.h>
+#include <libethcore/EVMSchedule.h>
+
 using namespace std;
 using namespace dev;
 using namespace eth;
@@ -35,6 +37,18 @@ ChainOperationParams::ChainOperationParams():
 }
 
 EVMSchedule const& ChainOperationParams::scheduleForBlockNumber(u256 const& _blockNumber) const
+{
+    EVMSchedule const& fixedForkSchedule = constScheduleForBlockNumber(_blockNumber);
+
+    if (_blockNumber < lastForkBlock)
+        return fixedForkSchedule;
+
+    static EVMSchedule const lastForkWithAdditionalEIPsSchedule =
+        addEIPs(fixedForkSchedule, lastForkAdditionalEIPs);
+    return lastForkWithAdditionalEIPsSchedule;
+}
+
+EVMSchedule const& ChainOperationParams::constScheduleForBlockNumber(u256 const& _blockNumber) const
 {
     if (_blockNumber >= experimentalForkBlock)
         return ExperimentalSchedule;
--- a/libethcore/ChainOperationParams.cpp
+++ b/libethcore/ChainOperationParams.cpp
@@ -40,10 +40,10 @@ EVMSchedule const& ChainOperationParams::scheduleForBlockNumber(u256 const& _blo
     if (_blockNumber >= lastForkBlock)
         return lastForkWithAdditionalEIPsSchedule;
     else
-        return fixedScheduleForBlockNumber(_blockNumber);
+        return forkScheduleForBlockNumber(_blockNumber);
 }
 
-EVMSchedule const& ChainOperationParams::fixedScheduleForBlockNumber(u256 const& _blockNumber) const
+EVMSchedule const& ChainOperationParams::forkScheduleForBlockNumber(u256 const& _blockNumber) const
 {
     if (_blockNumber >= experimentalForkBlock)
         return ExperimentalSchedule;
--- a/libethcore/ChainOperationParams.h
+++ b/libethcore/ChainOperationParams.h
@@ -9,12 +9,12 @@
 #include <libethcore/Precompiled.h>
 
 #include "Common.h"
-#include "EVMSchedule.h"
 
 namespace dev
 {
 namespace eth
 {
+struct EVMSchedule;
 
 class PrecompiledContract
 {
@@ -53,6 +53,11 @@ private:
 
 constexpr int64_t c_infiniteBlockNumber = std::numeric_limits<int64_t>::max();
 
+struct AdditionalEIPs
+{
+    bool eip2046 = false;
+};
+
 struct ChainOperationParams
 {
     ChainOperationParams();
@@ -62,9 +67,23 @@ struct ChainOperationParams
     /// The chain sealer name: e.g. Ethash, NoProof, BasicAuthority
     std::string sealEngineName = "NoProof";
 
+    // Example of how to check EIP activation from outside of EVM:
+    // bool isEIP2046Enabled(u256 const& _blockNumber) const
+    // {
+    //     return _blockNumber >= lastForkBlock && lastForkAdditionalEIPs.eip2046;
+    // }
+    // After hard fork finalization this is changed to:
+    // bool isEIP2046Enabled(u256 const& _blockNumber) const
+    // {
+    //     return _blockNumber >= berlinForkBlock;
+    // }
+
     /// General chain params.
 private:
     u256 m_blockReward;
+
+    EVMSchedule const& constScheduleForBlockNumber(u256 const& _blockNumber) const;
+
 public:
     EVMSchedule const& scheduleForBlockNumber(u256 const& _blockNumber) const;
     u256 blockReward(EVMSchedule const& _schedule) const;
@@ -86,6 +105,8 @@ public:
     u256 experimentalForkBlock = c_infiniteBlockNumber;
     u256 istanbulForkBlock = c_infiniteBlockNumber;
     u256 berlinForkBlock = c_infiniteBlockNumber;
+    u256 lastForkBlock = c_infiniteBlockNumber;
+    AdditionalEIPs lastForkAdditionalEIPs;
     int chainID = 0;    // Distinguishes different chains (mainnet, Ropsten, etc).
     int networkID = 0;  // Distinguishes different sub protocols.
 
--- a/libethcore/ChainOperationParams.h
+++ b/libethcore/ChainOperationParams.h
@@ -6,6 +6,7 @@
 #pragma once
 
 #include <libdevcore/Common.h>
+#include <libethcore/EVMSchedule.h>
 #include <libethcore/Precompiled.h>
 
 #include "Common.h"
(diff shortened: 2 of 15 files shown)


gumb0:
Rebased.

libethcore/ChainOperationParams.h
```diff
@@ -62,11 +68,24 @@ struct ChainOperationParams
     /// The chain sealer name: e.g. Ethash, NoProof, BasicAuthority
     std::string sealEngineName = "NoProof";
 
+    // Example of how to check EIP activation from outside of EVM:
```
halfalicious: Can we simplify this: 

    // Example of how to check EIP activation from outside of EVM:
    // bool isEIP2046Enabled(u256 const& _blockNumber) const
    // {
    //     return _blockNumber >= lastForkBlock && lastForkAdditionalEIPs.eip2046;
    // }
    // After hard fork finalization this is changed to:
    // bool isEIP2046Enabled(u256 const& _blockNumber) const
    // {
    //     return _blockNumber >= berlinForkBlock;
    // }

Instead of having a separate "is enabled" function for each EIP could we instead have a "isEipEnabled" function which takes both a block number and eip number and does the check? You'd need to add a switch statement in the function but overall I think it would result in cleaner code since you'd have fewer function signatures

libethcore/ChainOperationParams.cpp
```diff
@@ -35,6 +36,14 @@ ChainOperationParams::ChainOperationParams():
 }
 
 EVMSchedule const& ChainOperationParams::scheduleForBlockNumber(u256 const& _blockNumber) const
+{
+    if (_blockNumber >= lastForkBlock)
+        return lastForkWithAdditionalEIPsSchedule;
+    else
+        return fixedScheduleForBlockNumber(_blockNumber);
```
halfalicious: (Nit) Would a better name for this function be something like "forkScheduleforBlockNumber"? As someone who's unfamiliar with this part of the code I don't know what a "fixed" schedule is 😀

libethcore/EVMSchedule.cpp
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the GNU General Public License, Version 3.
+
+#include "EVMSchedule.h"
+#include <libethcore/ChainOperationParams.h>
+
+namespace dev
+{
+namespace eth
+{
+EVMSchedule addEIPsToSchedule(EVMSchedule const& _schedule, AdditionalEIPs const& _eips)
```
halfalicious: Why don't we modify `_schedule` instead of creating a new `EVMSchedule `instance?

halfalicious:
You mentioned in the issue that the syntax for specifying at runtime which EIPs to enable is a delimited list of EIPs - is there a chance that different EIPs can change the same settings so the ordering of the EIPs in the list matters? Or is this generally not a concern?

halfalicious:
What happens if someone specifies an eip not in the "AdditionalEIPs" struct? Presumably we get some sort of error?

libethereum/ChainParams.cpp
```diff
@@ -147,6 +180,18 @@ ChainParams ChainParams::loadGenesis(string const& _json, h256 const& _stateRoot
     return cp;
 }
 
+ChainParams ChainParams::addEIPs(AdditionalEIPs const& _eips) const
+{
+    ChainParams cp(*this);
```
halfalicious: {}?

libethereum/ChainParams.h
```diff
@@ -51,6 +59,9 @@ struct ChainParams: public ChainOperationParams
     ChainParams loadConfig(std::string const& _json, h256 const& _stateRoot = {},
         const boost::filesystem::path& _configPath = {}) const;
 
+    /// Activatie additional EIPs on top of the last fork block
```
halfalicious: Typo (Activatie -> Activate)

libethereum/ChainParams.cpp
```diff
@@ -147,6 +180,18 @@ ChainParams ChainParams::loadGenesis(string const& _json, h256 const& _stateRoot
     return cp;
 }
 
+ChainParams ChainParams::addEIPs(AdditionalEIPs const& _eips) const
```
halfalicious: If we can't update the existing ChainParams instance, should we make this a static function instead?

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -50,6 +50,25 @@ int mainnetChainID()
         ChainParams(genesisInfo(eth::Network::MainNetworkTest)).chainID;
     return c_mainnetChainID;
 }
+
+ChainParams createChainParamsForNetwork(string const& _networkName)
+{
+    // network name may be network+EIPs
+    vector<string> networkAndEIPs;
+    boost::algorithm::split(networkAndEIPs, _networkName, boost::is_any_of("+"));
```
halfalicious: There's no appropriate STL function that we can use?

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -855,11 +896,11 @@ void ImportTest::traceStateDiff()
 
     for(auto const& tr : m_transactions)
     {
-        if (network == netIdToString(tr.netId) || network == "ALL")
+        if (network == tr.netId || network == "ALL")
```
halfalicious: (Nit) Perhaps we should make this comparison case-insensitive?

libethereum/ChainParams.cpp
```diff
@@ -37,6 +59,12 @@ ChainParams::ChainParams(string const& _json, h256 const& _stateRoot)
     *this = loadConfig(_json, _stateRoot);
 }
 
+ChainParams::ChainParams(std::string const& _s, AdditionalEIPs const& _additionalEIPs)
```
halfalicious: (Nit) Can we choose a name for the `_s `arg which reflects the semantics of the data it contains?

libethereum/ChainParams.cpp
```diff
@@ -37,6 +59,12 @@ ChainParams::ChainParams(string const& _json, h256 const& _stateRoot)
...
+  : ChainParams(_s)
+{
+    *this = addEIPs(_additionalEIPs);
```
halfalicious: I think this approach is a little confusing (calling another ctor which sets some state and then overwriting the current instance) - I'd prefer if the addEIPs function either updates the current instances state (i.e. it's non-const) or we turn this ctor into a static function which calls `addEIPs ` and returns the result.

test/tools/libtesteth/ImportTest.cpp
```diff
@@ -307,7 +343,7 @@ bytes ImportTest::executeTest(bool _isFilling)
             if (Options::get().verbosity > 2)
             {
                 std::cout << "Executed transaction: \n";
```
halfalicious: Should we use boost loggers here? Or can this code be called before boost loggers are initialized?

gumb0:
> You mentioned in the issue that the syntax for specifying at runtime which EIPs to enable is a delimited list of EIPs - is there a chance that different EIPs can change the same settings so the ordering of the EIPs in the list matters? Or is this generally not a concern?

@halfalicious In theory it's possible, even if not about the same settings, some rules might be interrelated so that the order of applying matters... But I think the complexity of handling such flexibility is not worth the trouble for now (that is, EIPs in close areas will be applied in hard-coded order)

> What happens if someone specifies an eip not in the "AdditionalEIPs" struct? Presumably we get some sort of error?

Not sure if I understand the question, but if someone implements an EIP, there will be some kind of check against chainParams to activate it - like without this individual activation the check looked like `if (blockNumber >= chainParams.EIP158ForkBlock)` - and this way EIP is included in some fork definition.

libethcore/ChainOperationParams.h
gumb0: I wanted these functions to be very efficient (constant time), because they might be called in any hot path of consensus-critical code (only outside of EVM though). So minimizing the overhead of these helpers might be important for benchmarking EIP implementations and in general.

Implementing this with a switch would mean O(n) linear search for each such check, where n would grow with each new implemented EIP.

libethcore/ChainOperationParams.cpp
gumb0: yes, will change

libethcore/EVMSchedule.cpp
gumb0: Can't modify becasue fork schedules are constants. I can make it a static method or maybe a new constrctor of `EVMSchedule`

libethereum/ChainParams.cpp
gumb0: Agree, this is complicated. I originally added only `addEIPs` and made it similar to `loadConfig` and `loadGenesis` (which create new instance out of the current one)

But now I think I should remove `addEIPs` and leave only this constructor.

test/tools/libtesteth/ImportTest.cpp
gumb0: Not really, string functions are pretty low-level and/or verbose (like you could of cource `string::find` in a loop)

I like boost string functions because they create higher-level code, e.g. similar to split/join functions of other languages.

(We're probably going to get split via ranges in C++20 https://stackoverflow.com/questions/48402558/how-to-split-a-stdstring-into-a-range-v3-of-stdstring-views)

test/tools/libtesteth/ImportTest.cpp
gumb0: It's case-sensitive everywhere in testeth, I'd leave it like this.

(these strings are actually field names test JSON files, so proper behaviour depends on whether field names are case-sensitive in JSON, are they?)

test/tools/libtesteth/ImportTest.cpp
gumb0: I think we should, however testeth has its own hacking around logging and handling verbosities, so I'm not going to touch it here.

test/tools/libtesteth/ImportTest.cpp
halfalicious: Makes sense, if it's already case sensitive let's leave it alone.

test/tools/libtesteth/ImportTest.cpp
halfalicious: Makes sense, let's leave it alone

libethereum/ChainParams.cpp
```diff
@@ -37,6 +59,16 @@ ChainParams::ChainParams(string const& _json, h256 const& _stateRoot)
     *this = loadConfig(_json, _stateRoot);
 }
 
+ChainParams::ChainParams(std::string const& _configJson, AdditionalEIPs const& _additionalEIPs)
+  : ChainParams(_configJson)
+{
+    lastForkAdditionalEIPs = _additionalEIPs;
```
halfalicious: (Nit) Can we initialize this as a part of the initializers list?

libethcore/EVMSchedule.h
```diff
@@ -19,6 +19,8 @@ struct EVMSchedule
 {
     EVMSchedule(): tierStepGas(std::array<unsigned, 8>{{0, 2, 3, 5, 8, 10, 20, 0}}) {}
     EVMSchedule(bool _efcd, bool _hdc, unsigned const& _txCreateGas): exceptionalFailedCodeDeposit(_efcd), haveDelegateCall(_hdc), tierStepGas(std::array<unsigned, 8>{{0, 2, 3, 5, 8, 10, 20, 0}}), txCreateGas(_txCreateGas) {}
+    // consturct schedule with additional EIPs on top
```
halfalicious: (Nit) Spelling typo

libethereum/ChainParams.cpp
gumb0: No, because these are members of the base class

gumb0:
@winsvega @chfast I'm going to merge this today, if no reviews from you

PR #5742, opened 2019-09-11
Refactor transaction replay protection implementation

gumb0:
I got rid of the magic value `-4` which was used to represent transactions without replay protection (`-4` was used because inserting it into formulas `CHAIN_ID * 2 + 35` / `CHAIN_ID * 2 + 35` conveniently produced valid values `27` / `28` for transactions without replay protection)
Instead of this I represent `chainID` with `optional<uint64_t>`

For reference replay protection spec is in https://eips.ethereum.org/EIPS/eip-155

This is quite independent of the change to 64 bit in EVMC and CHAINID implementation https://github.com/ethereum/aleth/pull/5740 (works and passes the tests without it), but I'd merge this after the Chain ID size decision is finalized.

### DIFF
--- a/libethcore/TransactionBase.cpp
+++ b/libethcore/TransactionBase.cpp
@@ -51,25 +51,26 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
 
         m_data = rlp[5].toBytes();
 
-        int const v = rlp[6].toInt<int>();
+        u256 const v = rlp[6].toInt<u256>();
         h256 const r = rlp[7].toInt<u256>();
         h256 const s = rlp[8].toInt<u256>();
 
         if (isZeroSignature(r, s))
         {
-            m_chainId = v;
+            m_chainId = static_cast<uint64_t>(v);
             m_vrs = SignatureStruct{r, s, 0};
         }
         else
         {
             if (v > 36)
-                m_chainId = (v - 35) / 2; 
-            else if (v == 27 || v == 28)
-                m_chainId = -4;
-            else
+                m_chainId = (v - 35) / 2;
+            // only values 27 and 28 are allowed for non-replay protected transactions
+            else if (v != 27 && v != 28)
                 BOOST_THROW_EXCEPTION(InvalidSignature());
 
-            m_vrs = SignatureStruct{r, s, static_cast<byte>(v - (m_chainId * 2 + 35))};
+            auto const recoveryID =
+                m_chainId.has_value() ? byte{v - (*m_chainId * 2 + 35)} : byte{v - 27};
+            m_vrs = SignatureStruct{r, s, recoveryID};
 
             if (_checkSig >= CheckTransaction::Cheap && !m_vrs->isValid())
                 BOOST_THROW_EXCEPTION(InvalidSignature());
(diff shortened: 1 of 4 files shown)


libethcore/TransactionBase.cpp
```diff
@@ -51,25 +51,26 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
...
-                m_chainId = -4;
-            else
+                m_chainId = (v - 35) / 2;
+            // only values 27 and 28 are allowed for non-replay protected transactions
+            else if (v != 27 && v != 28)
                 BOOST_THROW_EXCEPTION(InvalidSignature());
 
-            m_vrs = SignatureStruct{r, s, static_cast<byte>(v - (m_chainId * 2 + 35))};
+            auto const recoveryID =
+                m_chainId.has_value() ? byte{v - (*m_chainId * 2 + 35)} : byte{v - 27};
```
chfast: This silently truncates a `uint256` to `byte`.

gumb0: I've added a check above to cap `m_chainId` at 64 bit. Given how chainID is calculated there, expression here inside `byte{...}` can only be either 0 or 1

libethcore/TransactionBase.cpp
```diff
@@ -63,13 +63,18 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
         else
         {
             if (v > 36)
-                m_chainId = (v - 35) / 2;
+            {
+                auto const chainId = (v - 35) / 2;
+                if (chainId > std::numeric_limits<uint64_t>::max())
+                    BOOST_THROW_EXCEPTION(InvalidSignature());
+                m_chainId = static_cast<byte>(chainId);
```
chfast: Why casted to byte?

gumb0: oh should be `unit64_t`, sorry

libethcore/TransactionBase.h
```diff
@@ -168,7 +167,9 @@ class TransactionBase
     u256 m_gas;							///< The total gas to convert, paid for from sender's account. Any unused gas gets refunded once the contract is ended.
     bytes m_data;						///< The data associated with the transaction, or the initialiser if it's a creation transaction.
     boost::optional<SignatureStruct> m_vrs;	///< The signature of the transaction. Encodes the sender.
-    int m_chainId = -4;					///< EIP155 value for calculating transaction hash https://github.com/ethereum/EIPs/issues/155
+    /// EIP155 value for calculating transaction hash
+    /// https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md
+    boost::optional<uint64_t> m_chainId;
```
halfalicious: Should we use std::optional now that we can build C++17 on Windows? Or should we just convert all boost::optional to std::optional in one go at some point in the future?

test/unittests/libethereum/Transaction.cpp
```diff
@@ -41,19 +41,48 @@ BOOST_AUTO_TEST_CASE(TransactionGasRequired)
 
 BOOST_AUTO_TEST_CASE(TransactionWithEmptyRecepient)
 {
-    // recipient RLP is 0x80 (empty array)
+    // recepient RLP is 0x80 (empty array)
```
halfalicious: (Nit) The old spelling is the correct one? https://www.merriam-webster.com/dictionary/recipient

gumb0: oh thanks, my mistake

libethcore/TransactionBase.h
gumb0: We still build with C++11 standard at the moment (it should be a separate PR when we want to switch; I'm not sure whether AppVeyor supports it)

chfast: Should be fine in VS 2017.

gumb0:
@fubuloubu The test vectors here might be useful in case you plan to submit a separate proposal to fix the size of ChainID to 64 bit in EIP-155 signing rules
https://github.com/ethereum/aleth/pull/5742/commits/48a46c1bc40f996079ad864c732a0f64ce0a9400#diff-fb674f25a4fbb780be5f264a0c9a9d76R71-R105

(I wanted to create reference Transaction tests out of them, but it doesn't look possible with current test format)

fubuloubu:
Thanks!

It's looking clear that the restriction to `uint64` might not be implemented for Istanbul (preferring EIP-155's potentially unlimited size for chain ID), so it is good that this is separated out.

chfast:
Merge?

gumb0:
Yeah I think it's safe to merge this regardless of the decisions around chain ID width

PR #5746, opened 2019-09-13
Improve transaction RLP decoding

gumb0:
cc @fjl @holiman


libethcore/TransactionBase.cpp
```diff
@@ -38,6 +38,9 @@ TransactionBase::TransactionBase(bytesConstRef _rlpData, CheckTransaction _check
         m_nonce = rlp[0].toInt<u256>();
         m_gasPrice = rlp[1].toInt<u256>();
         m_gas = rlp[2].toInt<u256>();
+        if (!rlp[3].isData())
```
chfast: Why is an array denoted as `isData()`?

gumb0: Somehow this is terminology of `RLP` class. RLP types are either "list" or "data"

gumb0: (error message should better say "byte array", then it's less confusing I think)

PR #5748, opened 2019-09-17
fix std::out of range in testeth when using --singlenet

### LINKED ISSUE #5736: testeth crashes with singlenet
```
testeth -- --singlenet Byzantium
```

The exception:
```
unknown location(0): fatal error: in "GeneralStateTests/stSStoreTest": std::out_of_range: map::at
/home/chfast/Projects/ethereum/aleth/test/tools/libtesteth/TestHelper.cpp(622): last checkpoint
```

What is https://github.com/ethereum/aleth/blob/0faec9d18e09459462f392d710b1c0d7f5008d5f/test/tools/libtesteth/TestHelper.cpp#L622

winsvega:
fix out of range in --singlenet in testeth

fixes #5736

### DIFF
--- a/test/tools/libtesteth/ImportTest.cpp
+++ b/test/tools/libtesteth/ImportTest.cpp
@@ -239,6 +239,9 @@ void ImportTest::parseJsonStrValueIntoSet(json_spirit::mValue const& _json, set<
 bool ImportTest::findExpectSectionForTransaction(
     transactionToExecute const& _tr, eth::Network const& _net, bool _isFilling) const
 {
+    if (!_isFilling && !m_testInputObject.at("post").get_obj().count(netIdToString(_net)))
+        return false;
+
     json_spirit::mArray const& expects =
         _isFilling ? m_testInputObject.at("expect").get_array() :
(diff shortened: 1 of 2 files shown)


test/tools/libtesteth/ImportTest.cpp
```diff
@@ -239,6 +239,9 @@ void ImportTest::parseJsonStrValueIntoSet(json_spirit::mValue const& _json, set<
 bool ImportTest::findExpectSectionForTransaction(
     transactionToExecute const& _tr, eth::Network const& _net, bool _isFilling) const
 {
+    if (!_isFilling && !m_testInputObject.at("post").get_obj().count(netIdToString(_net)))
```
chfast: I have no idea what it does :)

winsvega: so the issue was that this function was assuming that filled tests have all the networks, while tests could have only one or a set of networks. out of range happend on json object. 
now it return that post section not found for provided singlenet if there is no network in test post section. still results in an error but not in exception

chfast: I think that deserves a comment in the code.

PR #5750, opened 2019-09-17
fix testeth -t <TestSuite> -- --singletest <path> <testname> running a test from file

winsvega:
Closes https://github.com/ethereum/aleth/issues/5743

Add support for `testeth -t <TestSuite> -- --testfile <path>`

### DIFF
--- a/test/tools/libtesteth/Options.cpp
+++ b/test/tools/libtesteth/Options.cpp
@@ -34,7 +34,8 @@ void printHelp()
     cout << setw(35) << "-g <index>" << setw(25) << "Set the transaction gas array index when running GeneralStateTests\n";
     cout << setw(35) << "-v <index>" << setw(25) << "Set the transaction value array index when running GeneralStateTests\n";
     cout << setw(35) << "--singletest <TestName>" << "Run on a single test\n";
-    cout << setw(35) << "--singletest <TestFile> <TestName>" << "Run on a single test from file\n";
+    cout << setw(35) << "--singletest <TestFile>" << "Run tests from file. Require -t <TestSuite>\n";
+    cout << setw(35) << "--singletest <TestFile> <TestName>" << "Run on a single test from file. Require -t <TestSuite>\n";
     cout << setw(35) << "--singlenet <networkId>" << setw(25) << "Run tests for a specific network (Frontier|Homestead|EIP150|EIP158|Byzantium|Constantinople|ConstantinopleFix)\n";
     cout << setw(35) << "--verbosity <level>" << setw(25) << "Set logs verbosity. 0 - silent, 1 - only errors, 2 - informative, >2 - detailed\n";
     cout << setw(35) << "--vm <name|path> (=legacy)" << setw(25) << "Set VM type for VMTests suite. Available options are: interpreter, legacy.\n";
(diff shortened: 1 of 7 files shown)


test/tools/libtesteth/Options.cpp
```diff
@@ -48,7 +48,8 @@ void printHelp()
     cout << setw(35) << "-g <index>" << setw(25) << "Set the transaction gas array index when running GeneralStateTests\n";
     cout << setw(35) << "-v <index>" << setw(25) << "Set the transaction value array index when running GeneralStateTests\n";
     cout << setw(35) << "--singletest <TestName>" << "Run on a single test\n";
-    cout << setw(35) << "--singletest <TestFile> <TestName>" << "Run on a single test from file\n";
+    cout << setw(35) << "--singletest <TestFile>" << "Run tests from file. Require -t <TestSuite>\n";
```
gumb0: I would strongly prefer this to be another, separate option for pointing it to a file

test/tools/libtesteth/Options.cpp
```diff
@@ -48,7 +48,8 @@ void printHelp()
...
+    cout << setw(35) << "--singletest <TestFile> <TestName>" << "Run on a single test from file. Require -t <TestSuite>\n";
```
gumb0: And this should probably better be `--testfile <TestFile> --singletest <TestName>`

test/tools/libtesteth/Options.cpp
winsvega: actually I met quite amount of people who always try to pass file to this optio instead of testname

gumb0: right, because there's no other option that clearly looks like it should point to the file

test/tools/libtesteth/Options.cpp
gumb0: Actually, doesn't each file contain exactly one test always?

test/tools/libtesteth/Options.cpp
```diff
@@ -229,7 +230,15 @@ Options::Options(int argc, const char** argv)
                 }
             }
             else
-                singleTestName = std::move(name1);
+            {
+                if (boost::filesystem::exists(name1))
```
gumb0: This could confusingly try to run a file in case there's a subdirectory or a file in the current dir with the same name as the test - the downside of using one CLI flag for several purposes

test/tools/libtesteth/boostTest.cpp
```diff
@@ -148,7 +148,8 @@ int main(int argc, const char* argv[])
     }
 
     // Print suggestions of a test case if test suite not found
-    if (!sMinusTArg.empty() && !dev::test::inArray(c_allTestNames, sMinusTArg))
+    if (!sMinusTArg.empty() && !dev::test::inArray(c_allTestNames, sMinusTArg) &&
+        sMinusTArg != "customTestSuite")
```
gumb0: what does this change do?

test/tools/libtesteth/Options.cpp
yxliang01: +1 with @gumb0 It's really confusing that the same argument option takes very different types of values.

test/tools/libtesteth/Options.cpp
chfast: It does now in the result files. I'm not sure if source files are relevant here. Having `--testfile` (or `--file`) and dropping `--singletest` would be great.

gumb0: source files not really relevant, the option should just run the filled file

gumb0:
@winsvega please review my changes, hopefully I didn't break anything.

I removed `Options::singleTest` flag, now if `singleTestName` is non-empty it means `--singletest` is enabled.
And I changed `optional<string> singleTestFile` to just `string`, empty string means it was not set.

test/tools/libtesteth/boostTest.cpp
winsvega: disables name suggestion check for customTestSuite which is created dynamicly

PR #5751, opened 2019-09-19
Implement EIP-152: Add Blake2 compression function F precompile

gumb0:
https://eips.ethereum.org/EIPS/eip-152


libdevcrypto/Blake2.cpp
```diff
@@ -0,0 +1,189 @@
...
+};
+// clang-format on
+
+inline uint64_t load64(const void* src) noexcept
+{
+#if defined(NATIVE_LITTLE_ENDIAN)
+    uint64_t w;
+    memcpy(&w, src, sizeof w);
+    return w;
+#else
```
gumb0: @chfast Should we care about big endian platforms?

libdevcrypto/Blake2.cpp
```diff
@@ -166,11 +142,10 @@ bytes blake2FCompression(uint32_t _rounds, bytesConstRef _stateVector, bytesCons
         BOOST_THROW_EXCEPTION(InvalidInputSize());
 
     blake2b_state s{};
-    for (size_t i = 0; i < 8; ++i)
-        s.h[i] = load64(&_stateVector[i * sizeof(uint64_t)]);
+    std::memcpy(&s.h, _stateVector.data(), _stateVector.size());
 
-    s.t[0] = load64(_t0.data());
-    s.t[1] = load64(_t1.data());
+    s.t[0] = *reinterpret_cast<uint64_t const*>(_t0.data());
```
chfast: These should stay `load64` because otherwise this is undefined behavior if the memory is not aligned to 8 bytes.

libdevcrypto/Blake2.cpp
```diff
@@ -0,0 +1,163 @@
...
+namespace
+{
+DEV_SIMPLE_EXCEPTION(InvalidInputSize);
+
+enum blake2b_constant
+{
+    BLAKE2B_BLOCKBYTES = 128,
+};
+
+typedef struct blake2b_state__
```
chfast: Drop `typedef`.

libdevcrypto/Blake2.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+namespace crypto
+{
+namespace
+{
+DEV_SIMPLE_EXCEPTION(InvalidInputSize);
+
+enum blake2b_constant
+{
+    BLAKE2B_BLOCKBYTES = 128,
```
chfast: Convert to `constexpr`.

libdevcrypto/Blake2.cpp
```diff
@@ -0,0 +1,163 @@
...
+    uint64_t t[2];
+    uint64_t f[2];
+    uint8_t buf[BLAKE2B_BLOCKBYTES];
+    size_t buflen;
+    size_t outlen;
+    uint8_t last_node;
+} blake2b_state;
+
+// clang-format off
+const uint64_t blake2b_IV[8] =
```
chfast: `constexpr`.

libdevcrypto/Blake2.cpp
```diff
@@ -0,0 +1,163 @@
...
+// clang-format on
+
+inline uint64_t load64(const void* src) noexcept
+{
+    uint64_t w;
+    memcpy(&w, src, sizeof w);
+    return w;
+}
+
+inline uint64_t rotr64(const uint64_t w, const unsigned c) noexcept
```
chfast:
```suggestion
inline constexpr uint64_t rotr64(uint64_t w, unsigned c) noexcept
```

libdevcrypto/Blake2.cpp
```diff
@@ -0,0 +1,163 @@
...
+    G(r, 7, v[3], v[4], v[9], v[14], m);
+}
+
+
+void blake2b_compress(
+    uint32_t rounds, blake2b_state* S, const uint8_t block[BLAKE2B_BLOCKBYTES]) noexcept
+{
+    uint64_t m[16];
+    uint64_t v[16];
+    size_t i;
```
chfast: Move `i` to loops.

libdevcrypto/Blake2.cpp
chfast: I don't think so.

libdevcrypto/Blake2.cpp
chfast: I don't think so.

libdevcrypto/Blake2.cpp
```diff
@@ -0,0 +1,160 @@
+/// Aleth: Ethereum C++ client, tools and libraries.
+// Copyright 2019 Aleth Authors.
+// Licensed under the GNU General Public License, Version 3.
+
+#include "Blake2.h"
+
+#include <libdevcore/Exceptions.h>
+
+// The Blake 2 F compression function implemenation is based on the reference implementation,
```
gumb0: @chfast do we need any kind of license note here?

chfast: It is not required because the reference implementation is in public domain, but I think it is nice to leave it here.

gumb0:
benchmarking test vectors from EIP (+ another one for 1200 rounds) currently looks like this to me
```
Running 1 test case...
Test Case "bench_blake2compression": 
bench_blake2compression/test4: 67 ns
bench_blake2compression/test5: 343 ns
bench_blake2compression/test6: 257 ns
bench_blake2compression/test7: 57 ns
bench_blake2compression/test_1200rounds: 18502 ns
```

```
Running 1 test case...
Test Case "bench_blake2compression_maxrounds": 
bench_blake2compression_maxrounds/test8: 66246045792 ns
```

gumb0:
This passes state tests from https://github.com/ethereum/tests/pull/619, so I'm going to merge soon

PR #5752, opened 2019-09-20
Implement eip1380 (call-to-self) in LegacyVM

halfalicious:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1380.md

Charge a reduced gas fee (40 vs 700) for calls (call/callcode/delegatecall/staticcall) to functions within the same contract.

Add unit tests for both call-to-self and non-call-to-self cases (for call/callcode/delegatecall/staticcall)

### DIFF
--- a/libethcore/EVMSchedule.h
+++ b/libethcore/EVMSchedule.h
@@ -23,6 +23,7 @@ struct EVMSchedule
     bool eip158Mode = false;
     bool eip1283Mode = false;
     bool eip2200Mode = false;
+    bool eip1380Mode = false;
     bool haveBitwiseShifting = false;
     bool haveRevert = false;
     bool haveReturnData = false;
@@ -51,6 +52,7 @@ struct EVMSchedule
     unsigned callStipend = 2300;
     unsigned callValueTransferGas = 9000;
     unsigned callNewAccountGas = 25000;
+    unsigned callSelfGas = 40;
     unsigned selfdestructRefundGas = 24000;
     unsigned memoryGas = 3;
     unsigned quadCoeffDiv = 512;
@@ -74,8 +76,12 @@ struct EVMSchedule
     bool staticCallDepthLimit() const { return !eip150Mode; }
     bool emptinessIsNonexistence() const { return eip158Mode; }
     bool zeroValueTransferChargesNewAccountGas() const { return !eip158Mode; }
+<<<<<<< HEAD
     bool sstoreNetGasMetering() const { return eip1283Mode || eip2200Mode; }
     bool sstoreThrowsIfGasBelowCallStipend() const { return eip2200Mode; }
+=======
+    bool reducedCallToSelfGas() const { return eip1380Mode; }
+>>>>>>> Implement eip1380 in LegacyVM
 };
 
 static const EVMSchedule DefaultSchedule = EVMSchedule();
@@ -161,6 +167,7 @@ static const EVMSchedule ExperimentalSchedule = [] {
     EVMSchedule schedule = BerlinSchedule;
     schedule.accountVersion = 1;
     schedule.blockhashGas = 800;
+    schedule.eip1380Mode = true;
     return schedule;
 }();
 
--- a/libethcore/EVMSchedule.h
+++ b/libethcore/EVMSchedule.h
@@ -23,7 +23,6 @@ struct EVMSchedule
     bool eip158Mode = false;
     bool eip1283Mode = false;
     bool eip2200Mode = false;
-    bool eip1380Mode = false;
     bool haveBitwiseShifting = false;
     bool haveRevert = false;
     bool haveReturnData = false;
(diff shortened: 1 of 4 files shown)


halfalicious:
Still need to create tests

halfalicious:
Need to move changes to Berlin schedule once #5754 is merged

libethcore/EVMSchedule.h
```diff
@@ -35,6 +35,7 @@ struct EVMSchedule
     bool eip150Mode = false;
     bool eip158Mode = false;
     bool eip1283Mode = false;
+    bool eip1380Mode = false;
```
gumb0: I would rather simplify this by removing this flag, defining `callSelfGas` as equal to `callGas` in all schedules before Berlin, then setting it to 40 in Berlin only.

Then EVM will need to check only the dest address and no additional flags.

Make sure to change it together with `callGas` in `EIP150Schedule`.

halfalicious: @gumb0 Ah good idea!

libevm/LegacyVMCalls.cpp
```diff
@@ -219,6 +219,9 @@ bool LegacyVM::caseCallSetup(CallParameters *callParams, bytesRef& o_output)
     if (haveValueArg && m_SP[2] > 0)
         m_runGas += toInt63(m_schedule->callValueTransferGas);
 
+    if (m_ext->myAddress == destinationAddr)
+        m_runGas -= (m_schedule->callGas - m_schedule->callSelfGas);
```
gumb0: Let's better do this without subtraction, putting the condition to the initial assignment at line 207.
It's kind of simpler, no calculations, and wouldn't need to assume `callGas >= callSelfGas`

test/unittests/libevm/VMTest.cpp
```diff
@@ -694,6 +695,88 @@ class AlethInterpreterBalanceFixture : public BalanceFixture
...
+
+class CallSelfFixture : public TestOutputHelperFixture
+{
+public:
+    explicit CallSelfFixture(VMFace* _vm) : vm{_vm} { state.addBalance(address, 1 * ether); }
+
+    void testCallSelf()
+    {
+        constexpr size_t codeElementCount = 4;
+        bytes codeArray[] = {
```
gumb0: can be const

test/unittests/libevm/VMTest.cpp
```diff
@@ -694,6 +695,88 @@ class AlethInterpreterBalanceFixture : public BalanceFixture
...
+            // staticcall must come first (we need to pass if the code performs a
```
gumb0: well you could also just use `i == staticCallCaseIndex` as a condition there in the last argument of `ExtVM` constructor

halfalicious:
cc @gumb0

halfalicious:
Marking as in-progress since I'm adding call tests (non-call-to-self cases) as a part of this PR

halfalicious:
Rebased + squashed some commits

halfalicious:
@gumb0 / @chfast These changes are ready for review again. Some things to call out:
* Calls-to-self which include a value transfer are charged the value-transfer gas cost (currently 9000) - the call stipend (currently 2300). This seems wrong to me since I'd expect this to be a NOOP at the VM level and therefore I don't think there should be any gas cost for this but this is existing behavior so I'm not going to change it (since that would cause consensus issues).
* There's a bit of code duplication in the call tests - I don't think addressing this is high priority, especially since I've spent a good chunk of time on the tests already.  I could potentially address this in the future if we decide cleaning up the test code is important.

halfalicious:
Rebased + moved EIP to Berlin hardfork

test/unittests/libevm/VMTest.cpp
```diff
@@ -860,6 +859,191 @@ class LegacyVMPrecompileCallFixture : public PrecompileCallFixture
...
+                        VMFace const*, ExtVMFace const*) {
+            if (CallFixture::IsCallOp(_instr))
+                gasBefore = _gas;
+            else if (gasBefore != 0 && gasAfter == 0)
+                gasAfter = _gas;
+        };
+        vm->exec(gas, extVm, onOp);
+        // Subtract the call stipend gas since there's no code at the address being called
+        unsigned int gasExpected =
+            extVm.evmSchedule().callValueTransferGas - extVm.evmSchedule().callStipend;
```
gumb0: I'm a bit lost about this. Can you explain how it can cost less that transfer gas?

halfalicious: @gumb0 Yes - according to the definition of the call stipend (gcallstipend) in the YP, the call stipend is subtracted from the value transfer gas (for a non-zero value transfer). Additionally, the call stipend is added to the gas being passed to the call and its purpose is to ensure that there's just enough gas for the contract's fallback function to perform some basic steps (e.g. log something), but not enough for it to do something nefarious. Also, gas not used in a contract call is refunded to the caller. So, since there's no code at the addresses being called in my test cases, the stipend is refunded.

libevm/LegacyVMCalls.cpp
```diff
@@ -191,12 +191,12 @@ bool LegacyVM::caseCallSetup(CallParameters *callParams, bytesRef& o_output)
     assert(callParams->apparentValue == 0);
 
     callParams->staticCall = (m_OP == Instruction::STATICCALL || m_ext->staticCall);
-    Address const destinationAddr = asAddress(m_SP[1]);
-
+    auto const destinationAddr = asAddress(m_SP[1]);
     if (callParams->staticCall && isPrecompiledContract(destinationAddr))
-        m_runGas = toInt63(m_schedule->precompileStaticCallGas);
+        m_runGas += toInt63(m_schedule->precompileStaticCallGas);
```
gumb0: @halfalicious I just noticed that this became `+=`, why?

PR #5753, opened 2019-09-21
Implement EIP1380 (call-to-self) in aleth-interpreter

halfalicious:
No tests since @gumb0 mentioned that there's no way to measure gas cost in tests which use aleth-interpreter.

### DIFF
--- a/libaleth-interpreter/VMCalls.cpp
+++ b/libaleth-interpreter/VMCalls.cpp
@@ -197,8 +197,6 @@ bool VM::caseCallSetup(evmc_message& o_msg, bytesRef& o_output)
     // Check for call-to-self (eip1380) and adjust gas accordingly
     if (m_rev >= EVMC_BERLIN && m_message->destination == destination)
         m_runGas = VMSchedule::callSelfGas;
-    else
-        m_runGas = evmc_get_instruction_metrics_table(m_rev)[static_cast<size_t>(m_OP)].gas_cost;
 
     switch (m_OP)
     {
--- a/libaleth-interpreter/VMCalls.cpp
+++ b/libaleth-interpreter/VMCalls.cpp
@@ -192,7 +192,13 @@ void VM::caseCall()
 
 bool VM::caseCallSetup(evmc_message& o_msg, bytesRef& o_output)
 {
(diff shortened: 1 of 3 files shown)


halfalicious:
Still need to create tests

halfalicious:
Rebased

libaleth-interpreter/VMCalls.cpp
```diff
@@ -190,7 +190,13 @@ void VM::caseCall()
 
 bool VM::caseCallSetup(evmc_message& o_msg, bytesRef& o_output)
 {
-    m_runGas = m_rev >= EVMC_TANGERINE_WHISTLE ? 700 : 40;
+    evmc_address const destination = toEvmC(asAddress(m_SP[1]));
+    
+    // Check for call-to-self (eip1380) and adjust gas accordingly
+    if (fromEvmC(m_message->destination) == fromEvmC(destination) && m_rev >= EVMC_BERLIN)
```
chfast: You need to have the cheap `m_rev >= EVMC_BERLIN` check first.

halfalicious: @chfast Ah good point, comparing addresses is expensive since you're comparing 32 byte boost multiprecision values which are also first converted from big endian, instead of an EVMC_REVISION which is an enum (4 bytes).

halfalicious:
@chfast I removed the `fromEvmc `calls here: https://github.com/ethereum/aleth/blob/d843b9b2dd1308d951932fc52cc4698057e1752e/libaleth-interpreter/VMCalls.cpp#L197-L198

I think that's okay since these are both `evmc_address` objects so they can be compared (well technically `destination `is `evmc::address` which is a subclass of `evmc_address `but I don't think the subclass implements anything which would affect the comparison).

gumb0:
@halfalicious I removed the assignment to `m_runGas` from the metrics table, because it's actually already done for each opcode in `fetchInstruction()`

halfalicious:
> @halfalicious I removed the assignment to `m_runGas` from the metrics table, because it's actually already done for each opcode in `fetchInstruction()`

@gumb0 Ah I see, it’s done here: https://github.com/ethereum/aleth/blob/06dad45bf8be0de6e3560806607ebaea977c68e0/libaleth-interpreter/VM.cpp#L220

Thanks for the heads up!

PR #5732, opened 2019-09-03
Introduce Berlin hard fork

gumb0:

PR #5754, opened 2019-09-22
Add Berlin hardfork support to Aleth

halfalicious:
Added the LegacyVM schedule and made other edits so that the Berlin hardfork option is available.


halfalicious:
@gumb0 / @chfast Do I also need to add Berlin test entries to the Network enum: https://github.com/ethereum/aleth/blob/7d82c605ffeca6fae096fea70db0baf30efe9767/libethashseal/GenesisInfo.h#L29-L83

gumb0:
What's the difference from https://github.com/ethereum/aleth/pull/5732 ?

halfalicious:
> What's the difference from #5732 ?

Ah I forgot about that PR, I’ll close this one

PR #5756, opened 2019-09-23
cmake: Use C++14 by default

chfast:
This seems to be working with GCC6 and Visual Studio 2015, but maybe some features are missing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ include(CableBuildType)
 include(CableToolchains)
 include(GNUInstallDirs)
 
-cable_configure_toolchain(DEFAULT cxx11-pic)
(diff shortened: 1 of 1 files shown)


chfast:
It looks like VS 2015 do not have full C++14 `constexpr` support. It may still be not enough to use intx in #5757

halfalicious:
Why don't we move to C++17? How much work would that be?

chfast:
> Why don't we move to C++17? How much work would that be?

I don't know. Probably not much, but I want to do it in steps.

gumb0:
> It looks like VS 2015 do not have full C++14 constexpr support. It may still be not enough to use intx in #5757
Maybe we should drop VS 2015 support then?

gumb0:
Let's merge this?

chfast:
Rebased and I will merge this.

PR #5760, opened 2019-09-24
Drop Visual Studio 2015 support

chfast:


gumb0:
This deserves changelog item I think.

gumb0:
Why are there now two AppVeyor tasks, what's the difference?

chfast:
> Why are there now two AppVeyor tasks, what's the difference?

The `appveyor` branch is on the whitelist. I test it this way before creating a PR to save other CIs from running.

PR #5761, opened 2019-09-25
Update EVMC

gumb0:


gumb0:
@chfast updated to current EVMC master, waiting for release

chfast:
I believe this should now target `release/1.7` branch, because the `1.7.0` just waits for the EVMC release.

libaleth-interpreter/VM.cpp
```diff
@@ -234,7 +228,8 @@ void VM::fetchInstruction()
 {
     m_OP = Instruction(m_code[m_PC]);
     auto const metric = (*m_metrics)[static_cast<size_t>(m_OP)];
-    adjustStack(metric.num_stack_arguments, metric.num_stack_returned_items);
+    adjustStack(
```
chfast: This might be handled here: if you check `adjustStack()` it actually reverts the addition.

gumb0: @chfast Not sure I understand, do you mean changing `adjustStack` to something like this?
```
void VM::adjustStack(int _required, int _change)
{
    m_SP = m_SPP;

    // adjust stack and check bounds
    if (m_stackEnd < m_SP + _required)
        throwBadStack(_required, _change);
    m_SPP += _change;
    if (m_SPP < m_stack)
        throwBadStack(_removed, _added);
}
```

chfast: Yes.

PR #5774, opened 2019-10-09
fixedTypo

kailwallin:
'a' is the wrong article in this context, 'the' is a better option

### DIFF
--- a/doc/dependencies.rst
+++ b/doc/dependencies.rst
@@ -12,7 +12,7 @@ Security aspects
 ----------------
 
 A project that uses Hunter first downloads the hunter package itself. This is
-done with a help of HunterGate CMake module. A copy of the HunterGate module
+done with the help of HunterGate CMake module. A copy of the HunterGate module
 is included in the main project. The HunterGate is ordered to download specified
 release of Hunter and to check whenever the downloaded sources match the
 checksum provided to the HunterGate invocation.
--- a/doc/dependencies.rst
+++ b/doc/dependencies.rst
@@ -14,7 +14,7 @@ Security aspects
 A project which uses Hunter first downloads the Hunter client. This is
 done with the help of the HunterGate CMake module. A copy of the HunterGate module
 is included in the main project. The module is instructed to download a specified
-Hunter release and verify the downloaded source matches the checksum provided to the HunterGate invocation, after which the downloaded source is compiled.
+Hunter release and verify the downloaded source matches the checksum provided to the HunterGate invocation.
 
 A Hunter release contains a full list of (a specific version) of packages and checksums. Each downloaded package source checksum is compared against the corresponding checksum in the Hunter release.
 
--- a/doc/dependencies.rst
+++ b/doc/dependencies.rst
@@ -11,25 +11,16 @@ See documentation at https://docs.hunter.sh.
 Security aspects
 ----------------
 
-A project that uses Hunter first downloads the hunter package itself. This is
-done with the help of HunterGate CMake module. A copy of the HunterGate module
-is included in the main project. The HunterGate is ordered to download specified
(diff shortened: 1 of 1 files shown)


doc/dependencies.rst
```diff
@@ -12,7 +12,7 @@ Security aspects
 ----------------
 
 A project that uses Hunter first downloads the hunter package itself. This is
```
halfalicious: * Can you please capitalize the second instance of Hunter? 
* Can you remove "itself"?

Sentence should read: 

"A project that uses Hunter first downloads the Hunter package."

doc/dependencies.rst
```diff
@@ -12,7 +12,7 @@ Security aspects
...
-done with a help of HunterGate CMake module. A copy of the HunterGate module
+done with the help of HunterGate CMake module. A copy of the HunterGate module
```
halfalicious: Can you add "the" between "of HunterGate" i.e. "done with the help of the HunterGate CMake module."

doc/dependencies.rst
```diff
@@ -12,7 +12,7 @@ Security aspects
...
 is included in the main project. The HunterGate is ordered to download specified
```
halfalicious: Can you add "module" after "HunterGate" and "the" between "download" and "specified" i.e.:

"The HunterGate module is ordered to download the specified"

doc/dependencies.rst
```diff
@@ -12,7 +12,7 @@ Security aspects
...
 release of Hunter and to check whenever the downloaded sources match the
```
halfalicious: This sentence can be rewritten as:

"release of Hunter and to verify that the downloaded sources match the checksum provided to the HunterGate invocation."

doc/dependencies.rst
gumb0: This sounds a bit ambiguous, "hunter package" usually refers to some library packaged for Hunter, while this sentence I think refers to downloading some core Hunter stuff. Any ideas how to better reformulate?

gumb0:
@halfalicious please make the edits yourself here

doc/dependencies.rst
halfalicious: Perhaps "Hunter client"? The term is used here: https://docs.hunter.sh/en/latest/

halfalicious:
cc @gumb0

doc/dependencies.rst
```diff
@@ -11,25 +11,16 @@ See documentation at https://docs.hunter.sh.
...
-Hunter has an option to download cached binaries from specified cache server
-instead of building packages from source. This option is enabled by default.
-Because checksums of binaries are not know up front the main project can be at
-risk in case the cache server is compromised. In cpp-ethereum we decided to
-use only the cache server controlled by Ethereum C++ team located in
-https://github.com/ethereum/hunter-cache.
+A project which uses Hunter first downloads the Hunter client. This is
+done with the help of the HunterGate CMake module. A copy of the HunterGate module
+is included in the main project. The module is instructed to download a specified
+Hunter release and verify the downloaded source matches the checksum provided to the HunterGate invocation, after which the downloaded source is compiled.
```
gumb0: "verify that"?

gumb0: Not sure about `after which the downloaded source is compiled`, Hunter itself is not compiled, it's just a collection of cmake files

halfalicious: Was referring to the package source there (hunter release in my mind = collection of packages with specific version numbers...can be source or cached binaries). After re reading this I don’t think I know enough about how hunter works, do you want to update this text so that it’s more accurate?

gumb0: I think this paragraph refers to the cmake sources of the Hunter itself and their checksum. I'll just remove the last part.

PR #5801, opened 2019-10-28
testeth legacy blockchain tests suite

winsvega:
test repo update is in progress

### DIFF
--- a/test/tools/jsontests/BlockChainTests.h
+++ b/test/tools/jsontests/BlockChainTests.h
@@ -45,10 +45,10 @@ public:
 };
 
 template <class T>
-class bcValidTestFixture
+class bcTestFixture
 {
 public:
-    bcValidTestFixture(std::set<TestExecution> const& _execFlags = {})
+    bcTestFixture(std::set<TestExecution> const& _execFlags = {})
     {
         T suite;
         if (_execFlags.count(TestExecution::NotRefillable) &&
@@ -56,7 +56,7 @@ public:
             BOOST_FAIL("Tests are sealed and not refillable!");
 
         string const casename = boost::unit_test::framework::current_test_case().p_name;
-        boost::filesystem::path suiteFillerPath = suite.getFullPathFiller(casename).parent_path();
+        boost::filesystem::path const suiteFillerPath = suite.getFullPathFiller(casename).parent_path();
 
         // skip wallet test as it takes too much time (250 blocks) run it with --all flag
         if (casename == "bcWalletTest" && !test::Options::get().all)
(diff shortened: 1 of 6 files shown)


test/tools/jsontests/BlockChainTests.h
```diff
@@ -44,6 +44,52 @@ class bcGeneralTestsFixture : public StateTestFixtureBase<BCGeneralStateTestsSui
     bcGeneralTestsFixture() : StateTestFixtureBase({TestExecution::RequireOptionAll}) {}
 };
 
+template <class T>
```
gumb0: I think you can easy get rid of duplication here like this
```
BlockChainTestFixture
class BlockChainTestFixture
{
public:
    BlockChainTestFixture(std::set<TestExecution> const& _execFlags = {}, std::vector<std::string> const& _timeConsumingCases)
    {
        T suite;
        if (_execFlags.count(TestExecution::NotRefillable) &&
            (Options::get().fillchain || Options::get().filltests))
            BOOST_FAIL("Tests are sealed and not refillable!");

        string const casename = boost::unit_test::framework::current_test_case().p_name;
        boost::filesystem::path const suiteFillerPath = suite.getFullPathFiller(casename).parent_path();

        // skip time consuming tests, unless --all option is given
        if (!test::Options::get().all && contains(_timeConsumingCases, casename))
        {
            cnote << "Skipping " << casename << " because --all option is not specified.\n";
            test::TestOutputHelper::get().markTestFolderAsFinished(suiteFillerPath, casename);
            return;
        }

        suite.runAllTestsInFolder(casename);
        test::TestOutputHelper::get().markTestFolderAsFinished(suiteFillerPath, casename);
    }
};
```
```
class BlockChainValidBlocksTestFixture
  : public BlockChainTestFixture<BlockchainValidTestSuite>
{
public:
    BlockChainValidBlocksTestFixture()
      : BlockChainValidBlocksTestFixture({TestExecution::NotRefillable}, {"bcWalletTest"})
    {}
};

```

gumb0:
Does this change anything regarding how to run the tests? (command line)

test/tools/jsontests/BlockChainTests.h
gumb0: Actually, if `stWallet` exists in only in either valid or invalid suites, you can use the exact same code with `stWallet` check for both suites (it won't skip `stWallet` if it doesn't exist there)

test/tools/jsontests/LegacyTests.h
```diff
@@ -37,6 +37,30 @@ class BCGeneralStateTestsSuiteLegacyConstantinople : public BCGeneralStateTestsS
     }
 };
 
+class BlockchainInvalidTestSuiteLegacyConstantinople : public BlockchainInvalidTestSuite
```
gumb0: (I would name these classes rather like `LegacyConstantinopleBlockchainInvalidTestSuite` - it's easier to see what it is derived from by the suffix, it's a common convention like `class Thing  <- class DerivedThing`

winsvega:
> Does this change anything regarding how to run the tests? (command line)

to run the old blockchain tests now use command: 
```./testeth -t LegacyTests/Constantinople/BlockchainTests```

File: test/tools/jsontests/BlockChainTests.h (hunk 1 above)
Comment (winsvega) [reply test/tools/jsontests/BlockChainTests.h]: I think it is ok to use just one fixture class in this case

File: test/tools/jsontests/BlockChainTests.h [hunk 3]
```diff
@@ -44,6 +44,33 @@ class bcGeneralTestsFixture : public StateTestFixtureBase<BCGeneralStateTestsSui
...
+public:
+    bcTestFixture(std::set<TestExecution> const& _execFlags = {})
+    {
+        T suite;
+        if (_execFlags.count(TestExecution::NotRefillable) &&
+            (Options::get().fillchain || Options::get().filltests))
+            BOOST_FAIL("Tests are sealed and not refillable!");
+
+        string const casename = boost::unit_test::framework::current_test_case().p_name;
+        boost::filesystem::path suiteFillerPath = suite.getFullPathFiller(casename).parent_path();
```
gumb0: can be const

winsvega:
updated the change log.

winsvega:
the errors seems to be in libp2p test again. merge?

PR #5810, opened 2019-11-06
Upgrade EVMC to version 7.0.0

chfast:
Questions:
- Is aleth-interpreter still being tested?
- Are we running any CHAINID tests?

### DIFF
--- a/libaleth-interpreter/VM.cpp
+++ b/libaleth-interpreter/VM.cpp
@@ -24,8 +24,9 @@ void delete_output(const evmc_result* result)
     delete[] result->output_data;
 }
 
-evmc_result execute(evmc_vm* _instance, evmc_host_context* _context, evmc_revision _rev,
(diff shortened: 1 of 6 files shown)


gumb0:
> Is aleth-interpreter still being tested?

Yes, but here it fails to load
```
#!/bin/bash -eo pipefail
ETHEREUM_TEST_PATH=~/project/test/jsontests \
test/testeth -t GeneralStateTests -- --vm libaleth-interpreter/libaleth-interpreter.so
terminate called after throwing an instance of 'boost::wrapexcept<std::system_error>'
  what():  loading lib*****-interpreter/lib*****-interpreter.so failed: EVMC create function not found: Illegal seek
Received 'aborted' signal
```

> Are we running any CHAINID tests?

Yes, there're ` GeneralStateTests/stChainId` and `BCGeneralStateTests/stChainId`
Also a couple of unit tests.

chfast:
> Yes, but here it fails to load

This is related to the change: https://github.com/ethereum/evmc/pull/440

PR #5812, opened 2019-11-09
Update mainNetwork.cpp

Abeseba:
Add account


gumb0:
This would break compatibility with the mainnet, what did you try to achieve?

Issue #5813, opened 2019-11-10
Aleth never persists database minor version


halfalicious:
Aleth checks the minor version when opening the various databases - however, Aleth only persists the minor version to a file if the file already exists and contains a different value than Aleth expects (`c_minorProtocolVersion`). Aleth should obviously also persist the minor protocol version if the file doesn't already exist, since otherwise the file will never be created in the first place.

Relevant code is here (`BlockChain::open`): https://github.com/ethereum/aleth/blob/39ee6165d30d6027a0360a399fe516e14837a197/libethereum/BlockChain.cpp#L209-L222

halfalicious:
Impact of this is that Aleth won’t attempt a rebuild when the extras dB format changes since it never detects a minor version change.

PR #5827, opened 2019-11-15
Write database minor version to disk and fix rebuild on minor version mismatch bug

### LINKED ISSUE #5822: Aleth uses genesis block as latest block when performing rebuild in response to database minor version mismatch
Even if we address #5813 and persist the minor version to disk, `BlockChain::open()` renames the extras database path on minor version mismatch: https://github.com/ethereum/aleth/blob/cd25d2f56b90a1e8fb6441343e8cd94a184ef019/libethereum/BlockChain.cpp#L213-L222

It then opens the extras DB from the old path and populates m_lastBlockHash / m_lastBlockNumber to the genesis block since the old path is empty.

Therefore the last block hash used by `BlockChain::rebuild()` to identify the range of blocks which need to be reimported will be the genesis block, and the extras (and state) databases won't actually be rebuilt.

halfalicious:
Fix #5822, #5813 

Built ontop of #5826 so will wait to merge this until that has been merged.

This PR fixes two bugs:
* It writes the database minor version to disk (if it's not already present such as when you launch Aleth with a new database path) - this is required to detect database minor version changes which occur during extras database upgrades (5822)
* It doesn't remove the extras or state databases on minor version mismatch (5813) - removing these databases breaks the rebuild functionality because the rebuild functionality uses the last hash to determine chain tip (and therefore the range of blocks which must be reimported) - however, `BlockChain::open` populates this after renaming extras -> extras.old so since it can't find the "best" block in the extras DB it just uses the genesis hash.

### DIFF
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -192,35 +192,18 @@ void BlockChain::init(ChainParams const& _p)
     genesis();
 }
 
-unsigned BlockChain::open(fs::path const& _path, WithExisting _we)
+bool BlockChain::open(fs::path const& _path, WithExisting _we)
 {
     auto const path = _path.empty() ? db::databasePath() : _path;
     auto const chainPath = path / fs::path(toHex(m_genesisHash.ref().cropped(0, 4)));
     auto const chainSubPathBlocks = chainPath / fs::path("blocks");
     auto const extrasPath = chainPath / fs::path(toString(c_databaseVersion));
     auto const extrasSubPathExtras = extrasPath / fs::path("extras");
-    unsigned lastMinor = c_minorProtocolVersion;
+    unsigned lastMinor = c_databaseMinorVersion;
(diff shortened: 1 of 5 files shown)


libethereum/BlockChain.cpp
```diff
@@ -268,7 +276,7 @@ unsigned BlockChain::open(fs::path const& _path, WithExisting _we)
         throw;
     }
 
-    if (_we != WithExisting::Verify && !details(m_genesisHash))
+    if (_we != WithExisting::Verify && !rebuildNeeded && !details(m_genesisHash))
```
gumb0: If do this insertion here in case `rebuildNeeded == true`, maybe then we don't need to duplicate this code in `rebuild` ?

gumb0: Ah I see, rebuild will first rename this database anyway

halfalicious:
Rebased

Issue #5814, opened 2019-11-11
Misleading log in MemoryDB mode


gumb0:
```
> aleth/aleth --db memorydb -v 4
aleth, a C++ Ethereum client
INFO  11-11 17:30:37 aleth net    Id: ##77047807…
INFO  11-11 17:30:37 aleth net    ENR: [ seq=6 id=v4 key=02770478… ip=178.19.221.38 tcp=30303 udp=30303 ]
TRACE 11-11 17:30:37 aleth trace  Opened blockchain DB. Latest: #d4e56740…(rebuild not needed)
TRACE 11-11 17:30:37 aleth statedb Opening state database: "/home/andrei/.ethereum/d4e56740/12041/state"
```

Does StateDB still use disk in `--db memorydb` mode? I think it doesn't, and then the log message should be fixed, this path doesn't tell anything about reality.

halfalicious:
Agreed, will fix

halfalicious:
Fixed by #5829

PR #5829, opened 2019-11-17
Return size of block (in bytes) from web3.eth.getBlock RPC function

### LINKED ISSUE #5797: Aleth's block JSON (retrieved via web3.eth.getBlock) has different "size" value than Geth's
When one retrieves blocks from Aleth via `web3.eth.getBlock(<block number>)`, the block "size" field differs from Geth's. For example, take the Genesis block and an arbitrary block with some transactions ([65003](https://etherscan.io/block/65003)) - in both cases Aleth's block size is 0 whereas Geth's varies:

### Genesis Block
Aleth size = 0, Geth size = 540
**Aleth**
```
  author: "0x0000000000000000000000000000000000000000",
  boundary: "0x0000000040000000000000000000000000000000000000000000000000000000",
  difficulty: 17179869184,
  extraData: "0x11bbe8db4e347b4e8c937c1c8370e4b5ed33adb3db69cbdb7a38e1e50b1b82fa",
  gasLimit: 5000,
  gasUsed: 0,
  hash: "0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3",
  logsBloom: "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  miner: "0x0000000000000000000000000000000000000000",
  mixHash: "0x0000000000000000000000000000000000000000000000000000000000000000",
  nonce: "0x0000000000000042",
  number: 0,
  parentHash: "0x0000000000000000000000000000000000000000000000000000000000000000",
  receiptsRoot: "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
  seedHash: "0x0000000000000000000000000000000000000000000000000000000000000000",
  sha3Uncles: "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
  size: 0,
  stateRoot: "0xd7f8974fb5ac78d9ac099b9ad5018bedc2ce0a72dad1827a1709da30580f0544",
  timestamp: 0,
  totalDifficulty: 17179869184,
  transactions: [],
  transactionsRoot: "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
  uncles: []
```

**Geth**
```
  difficulty: 17179869184,
  extraData: "0x11bbe8db4e347b4e8c937c1c8370e4b5ed33adb3db69cbdb7a38e1e50b1b82fa",
  gasLimit: 5000,
  gasUsed: 0,
  hash: "0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3",
  logsBloom: "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  miner: "0x0000000000000000000000000000000000000000",
  mixHash: "0x0000000000000000000000000000000000000000000000000000000000000000",
  nonce: "0x0000000000000042",
  number: 0,
  parentHash: "0x0000000000000000000000000000000000000000000000000000000000000000",
  receiptsRoot: "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
  sha3Uncles: "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
  size: 540,
  stateRoot: "0xd7f8974fb5ac78d9ac099b9ad5018bedc2ce0a72dad1827a1709da30580f0544",
  timestamp: 0,
  totalDifficulty: 17179869184,
  transactions: [],
  transactionsRoot: "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
  uncles: []
```
### Block 65003
Aleth size = 0, Geth size = 846
**Aleth**
```
  author: "0xf927a40c8b7f6e07c5af7fa2155b4864a4112b13",
  boundary: "0x0000000000932bb9e4d20f331c0fd1eda100ae03dd36d33a9b2a6a0e9cd3915e",
  difficulty: 1912573463224,
  extraData: "",
  gasLimit: 3141592,
  gasUsed: 118496,
  hash: "0x650ec59d844817aff623efdc7a976e2f7005b17035e5ac5e7025747fbc4091ad",
  logsBloom: "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  miner: "0xf927a40c8b7f6e07c5af7fa2155b4864a4112b13",
  mixHash: "0xdfe2a94512f29e4363b48d0e8661a78f11d41e6fd36d9f2cd91e0c77ba0d8082",
  nonce: "0x5f7c213366edf1ef",
  number: 64999,
  parentHash: "0x23ee2f4a420e1ce4beb6d5902fafa232c0d5718146e53a31b179c86949191b85",
  receiptsRoot: "0x61b8587e6a557ecefb87f0091be5b894c197a5b43d3960093bd051589351bc32",
  seedHash: "0x510e4e770828ddbf7f7b00ab00a9f6adaf81c0dc9cc85f1f8249c256942d61d9",
  sha3Uncles: "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
  size: 0,
  stateRoot: "0x4db358e5d7eca7ead868e622e560b2c4d4588147aee681905ad7acc17e8fd8ec",
  timestamp: 1439229736,
  totalDifficulty: 73588533123492323,
  transactions: ["0xa047af96f2342ff7bc1a061b6d83a9f00af0c7986bb24924391bf0e83de22f85", "0x49be91905241a1bac6e9cfd2568be07aa4c2c13aec5c2ce4f63f27068a5836b1", "0x9acb4b6573444ed7ed12cd8920fdcded1beadacb1872228ce09b34ccfce8e008"],
  transactionsRoot: "0xac021c3d56771233aebe36ebb0442373812b0489c3b478309339991d6dd51a71",
  uncles: []
```

**Geth**
```
  difficulty: 1912573463224,
  extraData: "0x",
  gasLimit: 3141592,
  gasUsed: 118496,
  hash: "0x650ec59d844817aff623efdc7a976e2f7005b17035e5ac5e7025747fbc4091ad",
  logsBloom: "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  miner: "0xf927a40c8b7f6e07c5af7fa2155b4864a4112b13",
  mixHash: "0xdfe2a94512f29e4363b48d0e8661a78f11d41e6fd36d9f2cd91e0c77ba0d8082",
  nonce: "0x5f7c213366edf1ef",
  number: 64999,
  parentHash: "0x23ee2f4a420e1ce4beb6d5902fafa232c0d5718146e53a31b179c86949191b85",
  receiptsRoot: "0x61b8587e6a557ecefb87f0091be5b894c197a5b43d3960093bd051589351bc32",
  sha3Uncles: "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
  size: 846,
  stateRoot: "0x4db358e5d7eca7ead868e622e560b2c4d4588147aee681905ad7acc17e8fd8ec",
  timestamp: 1439229736,
  totalDifficulty: 73588533123492323,
  transactions: ["0xa047af96f2342ff7bc1a061b6d83a9f00af0c7986bb24924391bf0e83de22f85", "0x49be91905241a1bac6e9cfd2568be07aa4c2c13aec5c2ce4f63f27068a5836b1", "0x9acb4b6573444ed7ed12cd8920fdcded1beadacb1872228ce09b34ccfce8e008"],
  transactionsRoot: "0xac021c3d56771233aebe36ebb0442373812b0489c3b478309339991d6dd51a71",
  uncles: []
```

halfalicious:
Fix #5797, #5814

Return the size of a block in bytes to calls to `web3.eth.getBlock`. This requires extending the `BlockDetails `structure (by adding a new `blockSizeBytes `field) which in turn requires incrementing the db minor version number (since `BlockDetails ` are stored in the Extras database) which triggers an automatic database rebuild (which can take a while depending on how many blocks are in your local chain).

I've also made some minor changes to the database rebuild code to handle edge cases:
* Detect a previously prematurely terminated rebuild and log an error / throw exception if detected
* Throw an exception on rebuild failure (previously only logged an error message)
* Only update the database minor version file after a successful rebuild
* Additional logging during database rebuild

Tested changes by calling web3.eth.getBlock with the following block numbers and comparing the results against Etherscan: 0, 1, 65001 (3 txs), 101146 (2 txs), 109424 (1 tx), 115367 (7 txs)

### DIFF
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -26,7 +26,6 @@ using namespace std;
 using namespace dev;
 using namespace dev::eth;
 namespace fs = boost::filesystem;
-namespace db_paths = dev::eth::database_paths;
 
 #define ETH_TIMED_IMPORTS 1
 
@@ -200,17 +199,19 @@ bool BlockChain::open(fs::path const& _path, WithExisting _we)
 
     if (db::isDiskDatabase())
     {
-        db_paths::setDatabasePaths(_path, m_genesisHash);
+        if (!m_dbPaths || m_dbPaths->rootPath() != _path)
+            m_dbPaths = make_unique<DatabasePaths>(_path, m_genesisHash);
+
         if (_we == WithExisting::Kill)
         {
             LOG(m_loggerInfo)
                 << "Deleting chain databases. This will require a resync from genesis.";
-            fs::remove_all(db_paths::blocksDatabasePath());
-            fs::remove_all(db_paths::extrasDatabasePath());
-            fs::remove_all(db_paths::extrasDatabaseTemporaryPath());
+            fs::remove_all(m_dbPaths->blocksPath());
+            fs::remove_all(m_dbPaths->extrasPath());
+            fs::remove_all(m_dbPaths->extrasTemporaryPath());
         }
 
-        bytes const minorVersionBytes = contents(db_paths::extrasDatabaseMinorVersionPath());
+        bytes const minorVersionBytes = contents(m_dbPaths->extrasMinorVersionPath());
         if (!minorVersionBytes.empty())
         {
             DEV_IGNORE_EXCEPTIONS(lastMinor = (unsigned)RLP(minorVersionBytes));
@@ -219,16 +220,16 @@ bool BlockChain::open(fs::path const& _path, WithExisting _we)
                 rebuildNeeded = true;
                 LOG(m_loggerInfo) << "Database minor version change detected, the extras and state "
                                      "databases will be rebuilt.";
-                LOG(m_loggerInfo) << "Version from " << db_paths::extrasDatabaseMinorVersionPath()
+                LOG(m_loggerInfo) << "Version from " << m_dbPaths->extrasMinorVersionPath()
                                   << " (" << lastMinor << ") != Aleth's version ("
                                   << c_databaseMinorVersion << ")";
             }
         }
-        else if (fs::exists(db_paths::extrasDatabasePath()))
+        else if (fs::exists(m_dbPaths->extrasPath()))
         {
             LOG(m_loggerInfo) << "Database minor version file not found ("
-                              << db_paths::extrasDatabaseMinorVersionPath()
-                              << ") but extras database exists (" << db_paths::extrasDatabasePath()
+                              << m_dbPaths->extrasMinorVersionPath()
+                              << ") but extras database exists (" << m_dbPaths->extrasPath()
                               << "), assuming extras database needs to be upgraded.";
             rebuildNeeded = true;
         }
@@ -236,27 +237,27 @@ bool BlockChain::open(fs::path const& _path, WithExisting _we)
         {
             // First launch with new database
             LOG(m_loggerDetail) << "Creating database minor version file: "
-                                << db_paths::extrasDatabaseMinorVersionPath()
+                                << m_dbPaths->extrasMinorVersionPath()
                                 << " (minor version: " << c_databaseMinorVersion << ")";
-            writeFile(db_paths::extrasDatabaseMinorVersionPath(), rlp(c_databaseMinorVersion));
+            writeFile(m_dbPaths->extrasMinorVersionPath(), rlp(c_databaseMinorVersion));
         }
     }
 
     try
     {
-        m_blocksDB = db::DBFactory::create(db_paths::blocksDatabasePath());
-        m_extrasDB = db::DBFactory::create(db_paths::extrasDatabasePath());
+        m_blocksDB = db::DBFactory::create(m_dbPaths->blocksPath());
+        m_extrasDB = db::DBFactory::create(m_dbPaths->extrasPath());
     }
     catch (db::DatabaseError const& ex)
     {
         // Determine which database open call failed
         auto const dbPath =
-            !m_blocksDB.get() ? db_paths::blocksDatabasePath() : db_paths::extrasDatabasePath();
+            !m_blocksDB.get() ? m_dbPaths->blocksPath() : m_dbPaths->extrasPath();
         if (db::isDiskDatabase())
         {
             LOG(m_loggerError) << "Error occurred when opening database: " << dbPath;
             db::DatabaseStatus const dbStatus = *boost::get_error_info<db::errinfo_dbStatusCode>(ex);
-            if (fs::space(db_paths::rootPath()).available < 1024)
+            if (fs::space(m_dbPaths->rootPath()).available < 1024)
             {
                 LOG(m_loggerError)
                     << "Not enough available space found on hard drive. Please free some up and "
@@ -287,13 +288,14 @@ bool BlockChain::open(fs::path const& _path, WithExisting _we)
     if (_we != WithExisting::Verify && !rebuildNeeded && !details(m_genesisHash))
     {
         bytes const genesisBlockBytes = m_params.genesisBlock();
-        BlockHeader gb{genesisBlockBytes};
+        BlockHeader const genesisHeader{genesisBlockBytes};
         // Insert details of genesis block.
-        m_details[m_genesisHash] = BlockDetails{0 /* number */, gb.difficulty(),
+        m_details[m_genesisHash] = BlockDetails{0 /* number */, genesisHeader.difficulty(),
             h256{} /* parent */, {} /* children */, genesisBlockBytes.size()};
-        auto r = m_details[m_genesisHash].rlp();
-        m_extrasDB->insert(toSlice(m_genesisHash, ExtraDetails), (db::Slice)dev::ref(r));
-        assert(isKnown(gb.hash()));
+        auto const genesisDetailsRlp = m_details[m_genesisHash].rlp();
+        m_extrasDB->insert(
+            toSlice(m_genesisHash, ExtraDetails), (db::Slice)dev::ref(genesisDetailsRlp));
+        assert(isKnown(genesisHeader.hash()));
     }
 
     // TODO: Implement ability to rebuild details map from DB.
@@ -318,7 +320,7 @@ void BlockChain::reopen(ChainParams const& _p, WithExisting _we, ProgressCallbac
 {
     close();
     init(_p);
-    open(db_paths::rootPath(), _we, _pc);
+    open(m_dbPaths->rootPath(), _we, _pc);
 }
 
 void BlockChain::close()
@@ -354,7 +356,8 @@ void BlockChain::rebuild(
         return;
     }
 
-    db_paths::setDatabasePaths(_path, m_genesisHash);
+    if (!m_dbPaths || m_dbPaths->rootPath() != _path)
+        m_dbPaths = make_unique<DatabasePaths>(_path, m_genesisHash);
 
     unsigned const originalNumber = m_lastBlockNumber;
 
@@ -365,25 +368,25 @@ void BlockChain::rebuild(
 
     // Keep extras DB around, but under a temp name
     m_extrasDB.reset();
-    LOG(m_loggerInfo) << "Renaming extras path " << db_paths::extrasDatabasePath() << " to "
-                      << db_paths::extrasDatabaseTemporaryPath();
-    if (fs::exists(db_paths::extrasDatabaseTemporaryPath()))
+    LOG(m_loggerInfo) << "Renaming extras path " << m_dbPaths->extrasPath() << " to "
+                      << m_dbPaths->extrasTemporaryPath();
+    if (fs::exists(m_dbPaths->extrasTemporaryPath()))
     {
         LOG(m_loggerError)
-            << "Temporary extras path " << db_paths::extrasDatabaseTemporaryPath()
+            << "Temporary extras path " << m_dbPaths->extrasTemporaryPath()
             << " already exists (this usually happens because an in-progress rebuild was "
                "prematurely terminated).";
         LOG(m_loggerError) << "Please re-run Aleth the --kill option to delete all databases. This "
                               "will remove all chain data and require you to resync from genesis.";
         BOOST_THROW_EXCEPTION(DatabaseExists());
     }
-    fs::rename(db_paths::extrasDatabasePath(), db_paths::extrasDatabaseTemporaryPath());
-    std::unique_ptr<db::DatabaseFace> oldExtrasDB(
-        db::DBFactory::create(db_paths::extrasDatabaseTemporaryPath()));
-    m_extrasDB = db::DBFactory::create(db_paths::extrasDatabasePath());
+    fs::rename(m_dbPaths->extrasPath(), m_dbPaths->extrasTemporaryPath());
+    std::unique_ptr<db::DatabaseFace> oldExtrasDB{
+        db::DBFactory::create(m_dbPaths->extrasTemporaryPath())};
+    m_extrasDB = db::DBFactory::create(m_dbPaths->extrasPath());
 
     // Open a fresh state DB
-    Block s = genesisBlock(State::openDB(db_paths::rootPath(), m_genesisHash, WithExisting::Kill));
+    Block s = genesisBlock(State::openDB(m_dbPaths->rootPath(), m_genesisHash, WithExisting::Kill));
 
     // Clear all memos ready for replay.
     m_details.clear();
@@ -452,14 +455,13 @@ void BlockChain::rebuild(
         if (_progress)
             _progress(d, originalNumber);
     }
-    LOG(m_loggerInfo) << "Removing old extras database: "
-                      << db_paths::extrasDatabaseTemporaryPath();
+    LOG(m_loggerInfo) << "Removing old extras database: " << m_dbPaths->extrasTemporaryPath();
     oldExtrasDB.reset();
-    fs::remove_all(db_paths::extrasDatabaseTemporaryPath());
+    fs::remove_all(m_dbPaths->extrasTemporaryPath());
     if (!rebuildFailed)
     {
         LOG(m_loggerInfo) << "Rebuild complete! Reimported " << originalNumber << " blocks!";
-        writeFile(db_paths::extrasDatabaseMinorVersionPath(), rlp(c_databaseMinorVersion));
+        writeFile(m_dbPaths->extrasMinorVersionPath(), rlp(c_databaseMinorVersion));
     }
     else
     {
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -206,10 +206,8 @@ bool BlockChain::open(fs::path const& _path, WithExisting _we)
     {
         if (_we == WithExisting::Kill)
         {
-            cnote << "Killing blockchain (" << chainSubPathBlocks << ") & extras ("
-                  << extrasSubPathExtras << ") databases (WithExisting::Kill).";
-            fs::remove_all(chainSubPathBlocks);
-            fs::remove_all(extrasSubPathExtras);
+            LOG(m_loggerInfo) << "Deleting all databases. This will require a resync from genesis.";
+            fs::remove_all(chainPath);
         }
 
         fs::create_directories(extrasPath);
@@ -217,7 +215,6 @@ bool BlockChain::open(fs::path const& _path, WithExisting _we)
 
         auto const extrasSubPathMinor = extrasPath / fs::path("minor");
         bytes const minorVersionBytes = contents(extrasSubPathMinor);
-        bool writeMinorVersion = false;
         if (!minorVersionBytes.empty())
         {
             DEV_IGNORE_EXCEPTIONS(lastMinor = (unsigned)RLP(minorVersionBytes));
@@ -228,13 +225,15 @@ bool BlockChain::open(fs::path const& _path, WithExisting _we)
                                      "databases will be rebuilt.";
                 LOG(m_loggerInfo) << "Version from " << extrasSubPathMinor << " (" << lastMinor
                                   << ") != Aleth's version (" << c_databaseMinorVersion << ")";
-                writeMinorVersion = true;
             }
         }
         else
-            writeMinorVersion = true;
-        if (writeMinorVersion)
+        {
+            LOG(m_loggerDetail) << "Creating database minor version file: " << extrasSubPathMinor
+                                << " (current database minor version: " << c_databaseMinorVersion
+                                << ")";
             writeFile(extrasSubPathMinor, rlp(c_databaseMinorVersion));
+        }
     }
 
     try
@@ -246,32 +245,35 @@ bool BlockChain::open(fs::path const& _path, WithExisting _we)
     {
         // Determine which database open call failed
         auto const dbPath = !m_blocksDB.get() ? chainSubPathBlocks : extrasSubPathExtras;
-        cerror << "Error opening database: " << dbPath;
+        LOG(m_loggerError) << "Error opening database: " << dbPath;
 
         if (db::isDiskDatabase())
         {
             db::DatabaseStatus const dbStatus = *boost::get_error_info<db::errinfo_dbStatusCode>(ex);
             if (fs::space(path).available < 1024)
             {
-                cerror << "Not enough available space found on hard drive. Please free some up and "
-                        "re-run.";
+                LOG(m_loggerError)
+                    << "Not enough available space found on hard drive. Please free some up and "
+                       "re-run.";
                 BOOST_THROW_EXCEPTION(NotEnoughAvailableSpace());
             }
             else if (dbStatus == db::DatabaseStatus::Corruption)
             {
-                cerror << "Database corruption detected. Please see the exception for corruption "
-                        "details. Exception: "
+                LOG(m_loggerError)
+                    << "Database corruption detected. Please see the exception for corruption "
+                       "details. Exception: "
                     << ex.what();
                 BOOST_THROW_EXCEPTION(DatabaseCorruption());
             }
             else if (dbStatus == db::DatabaseStatus::IOError)
             {
-                cerror << "Database already open. You appear to have another instance of Aleth running.";
+                LOG(m_loggerError) << "Database already open. You appear to have another instance "
+                                      "of Aleth running.";
                 BOOST_THROW_EXCEPTION(DatabaseAlreadyOpen());
             }
         }
-        
-        cerror << "Unknown error occurred. Exception details: " << ex.what();
+
+        LOG(m_loggerError) << "Unknown error occurred. Exception details: " << ex.what();
         throw;
     }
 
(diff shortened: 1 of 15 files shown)


halfalicious:
Size returned for genesis block is still 0 - I suspect that the genesis block isn't actually inserted into the db

halfalicious:
Size returned for genesis block is still 0 bytes, need to investigate.

libethereum/BlockDetails.cpp
```diff
@@ -14,14 +14,13 @@ BlockDetails::BlockDetails(RLP const& _r)
 {
 	number = _r[0].toInt<unsigned>();
```
gumb0: Could you also convert tabs to spaces in BlockDetails.cpp and BlockDetails.h?

libethereum/BlockDetails.h
```diff
@@ -17,27 +17,39 @@ namespace eth
...
 	u256 totalDifficulty = Invalid256;
-	h256 parent;
-	h256s children;
+	h256 parentHash;
+	h256s childHashes;
 
-	mutable unsigned size;
+    // Size of the BlockDetails RLP (in bytes). Used for computing blockchain memory usage
+    // statistics. Field name must be 'size' as BlockChain::getHashSize depends on this
+    mutable unsigned size;
```
gumb0: Is it still useful for anything?

libweb3jsonrpc/Eth.cpp
```diff
@@ -345,17 +345,28 @@ Json::Value Eth::eth_getBlockByHash(string const& _blockHash, bool _includeTrans
...
 
-		if (_includeTransactions)
-			return toJson(client()->blockInfo(h), client()->blockDetails(h), client()->uncleHashes(h), client()->transactions(h), client()->sealEngine());
-		else
-			return toJson(client()->blockInfo(h), client()->blockDetails(h), client()->uncleHashes(h), client()->transactionHashes(h), client()->sealEngine());
-	}
-	catch (...)
-	{
+        Json::Value ret;
+        auto const blockDetails = client()->blockDetails(h);
```
gumb0: Could you then also extract here the calls ro `blockInfo`, `uncleHashes`, `sealEngine`, not to duplicate them below

libweb3jsonrpc/Eth.cpp
```diff
@@ -345,17 +345,28 @@ Json::Value Eth::eth_getBlockByHash(string const& _blockHash, bool _includeTrans
...
+            ret = toJson(client()->blockInfo(h), blockDetails,
+                client()->uncleHashes(h), client()->transactions(h), client()->sealEngine());
+        else
+            ret = toJson(client()->blockInfo(h), blockDetails,
+                client()->uncleHashes(h), client()->transactionHashes(h), client()->sealEngine());
+
+        // We need to explicitly set the "size" field to the block size in bytes since the
+        // BlockDetails "size" field refers to something else (size of BlockDetails RLP) and cannot
+        // be changed
+        ret["size"] = toJS(blockDetails.blockSizeBytes);
```
gumb0: Why not do this inside `toJson(BlockHeader, BlockDetails, UncleHashes, Transactions, SealEngineFace*)` ?

libethereum/BlockChain.cpp
```diff
@@ -206,18 +206,15 @@ bool BlockChain::open(fs::path const& _path, WithExisting _we)
     {
         if (_we == WithExisting::Kill)
         {
-            cnote << "Killing blockchain (" << chainSubPathBlocks << ") & extras ("
-                  << extrasSubPathExtras << ") databases (WithExisting::Kill).";
-            fs::remove_all(chainSubPathBlocks);
-            fs::remove_all(extrasSubPathExtras);
+            LOG(m_loggerInfo) << "Deleting all databases. This will require a resync from genesis.";
+            fs::remove_all(chainPath);
```
gumb0: I think here it was removing only blocks and extras because `BlockChain` class is responsible on only for these DB, and `State` is responsible for state.

I'd leave it without change for better separation of concertns.
(with this change there's at least a problem when it gets to `State::openDB` and logs `Killing state database` - that's not true anymore)

libethereum/BlockChain.cpp
```diff
@@ -206,18 +206,15 @@ bool BlockChain::open(fs::path const& _path, WithExisting _we)
...
         }
 
         fs::create_directories(extrasPath);
         DEV_IGNORE_EXCEPTIONS(fs::permissions(extrasPath, fs::owner_all));
 
         auto const extrasSubPathMinor = extrasPath / fs::path("minor");
```
gumb0: Perhaps make a constant out of `"minor"` as it's used in more that one place now

libethereum/BlockChain.cpp
gumb0: Ah I see below that you did it to remove also extras.old, if it's there... Maybe better to remove existing extras.old in that case?

libethereum/BlockChain.cpp
```diff
@@ -384,16 +398,19 @@ void BlockChain::rebuild(fs::path const& _path, std::function<void(unsigned, uns
...
     // first block.
-    auto const genesisDetails = BlockDetails{0, s.info().difficulty(), h256(), {}};
+    auto const genesisDetails =
+        BlockDetails{0 /* block number */, s.info().difficulty(), h256{} /* parent */,
+            {} /* children */, static_cast<unsigned>(m_params.genesisBlock().size())};
     m_details[m_genesisHash] = genesisDetails;
-    auto const genesisDetailsRlp = genesisDetails.rlp();
     m_extrasDB->insert(
-        toSlice(m_genesisHash, ExtraDetails), (db::Slice)dev::ref(genesisDetailsRlp));
+        toSlice(m_genesisHash, ExtraDetails), (db::Slice)dev::ref(genesisDetails.rlp()));
```
gumb0: I think this is dangerous and might crash. `genesisDetails.rlp()` creates temporary `bytes`, then `dev::ref` creates `vector_ref` that stores the pointer to the data of that temporary, but temporary is not guaranteed to exist after `vector_ref(bytes const&)` is finished

libethereum/ClientBase.cpp
```diff
@@ -417,9 +417,11 @@ BlockHeader ClientBase::pendingInfo() const
...
 BlockDetails ClientBase::pendingDetails() const
 {
-    auto pm = postSeal().info();
-    auto li = Interface::blockDetails(LatestBlock);
-    return BlockDetails((unsigned)pm.number(), li.totalDifficulty + pm.difficulty(), pm.parentHash(), h256s{});
+    auto pendingHeader = postSeal().info();
+    auto latestDetails = Interface::blockDetails(LatestBlock);
+    return BlockDetails{(unsigned)pendingHeader.number(),
+        latestDetails.totalDifficulty + pendingHeader.difficulty(), pendingHeader.parentHash(),
+        h256s{} /* children */, static_cast<unsigned>(postSeal().blockData().size())};
```
gumb0: (`static_cast` will not be needed if you change the type of `blockSizeBytes` to `size_t`)

halfalicious:
Need to also handle case of the minor version file not existing and the DB format changing

libethereum/ClientBase.cpp
halfalicious: @gumb0: had originally used size_t but the returned block sized were zero for some reason..will try again and root cause why.

libweb3jsonrpc/Eth.cpp
halfalicious: Great suggestion, will do that

libethereum/BlockDetails.h
halfalicious: @gumb0 Not sure, will see if it can be removed

libethereum/BlockChain.cpp
halfalicious: I think we should leave extras.old (unless one explicitly specifies --kill) since there's the possibility of using it to restore your extras DB (if you know what you're doing), keeping it around gives the user the chance to do that.

libethereum/BlockChain.cpp
halfalicious: @gumb0 I don't see where it's being used multiple times in a function?

libethereum/ClientBase.cpp
halfalicious: Looks like that was a different issue that I've since addressed, size_t works

halfalicious: Actually looks like that was an issue that I've since addressed, it works now

libethereum/BlockChain.cpp
gumb0: In another function (`rebuild`)

halfalicious: Ah, yes, most of the paths are used in both places. I can move them to anonymous namespace vars.

libethereum/BlockDetails.h
halfalicious: It's used to compute BlockChain memory usage stats which can be used to garbage collect data from the BlockChain cache:

Memory usage stats computation: https://github.com/ethereum/aleth/blob/43780e1cfaad9d887b5cd730a5df6abcea81f302/libethereum/BlockChain.cpp#L1184-L1205

Garbage collection: https://github.com/ethereum/aleth/blob/43780e1cfaad9d887b5cd730a5df6abcea81f302/libethereum/BlockChain.cpp#L1207-L1271

gumb0: Then I think you should still set it in ` BlockDetails::rlp()` function

halfalicious: @gumb0 Oops yes I definitely should 😀

halfalicious:
Will squash some changes to clean up the commit history before merging

halfalicious:
Rebased to fix changelog merge conflict

halfalicious:
Also did a successful rebuild of ~2M blocks.

libethereum/ClientBase.cpp
gumb0: So can `static_cast` me removed?

libethereum/DatabasePaths.h
```diff
@@ -0,0 +1,27 @@
...
+
+namespace dev
+{
+namespace eth
+{
+namespace database_paths
+{
+bool databasePathsSet();
+void setDatabasePaths(boost::filesystem::path const& _rootPath, h256 const& _genesisHash);
+boost::filesystem::path rootPath();
```
gumb0: I think I would return const references from these, to avoid copying

libethereum/DatabasePaths.cpp
```diff
@@ -0,0 +1,82 @@
+// Aleth: Ethereum C++ client, tools and libraries.
+// Copyright 2014-2019 Aleth Authors.
+// Licensed under the GNU General Public License, Version 3.
+
+#include "DatabasePaths.h"
+#include "libdevcore/CommonIO.h"
+#include "libdevcore/DBFactory.h"
+#include "libethcore/Exceptions.h"
```
gumb0: Is this needed?

libethereum/DatabasePaths.h
gumb0: Somewhat confusing that we have `db::databasePath()` and now these. Should all these maybe better put into `DBFactory.h`?

halfalicious: @gumb0 I considered that but DBFactory strikes me as more general purpose..it doesn’t include anything related to chains and such. Should I maybe rename the new namespace and files to something like BlockchainDatabasePaths to make the differences more clear?

libethereum/DatabasePaths.cpp
halfalicious: Ah nice catch - it used to be since I would throw if an empty genesis hash was supplied, but I pulled that out because tests use empty hashes. I’ll remove the header.

libethereum/DatabasePaths.h
gumb0: The namespace and files are fine, but then maybe it makes sense to move `db::setDatabasePath` to this file and/or merge with `setDatabasePaths`?

Isn't `rootPath` here basically the same as `db::databasePath` ?

halfalicious: @gumb0 Yes, rootPath is the same as db::databasePath. After looking at this a little more I like your original suggestion of merging these together...I'll move the DatabasePaths functions out of that namespace and into the dev::db namespace.

halfalicious:
Still need to merge DatabasePaths with dev::db

libethereum/DatabasePaths.h
```diff
@@ -0,0 +1,27 @@
...
+
+namespace dev
+{
+namespace eth
+{
+namespace database_paths
+{
+bool databasePathsSet();
+void setDatabasePaths(boost::filesystem::path const& _rootPath, h256 const& _genesisHash);
+boost::filesystem::path const& rootPath();
```
gumb0: Also make them `noexcept`

libethereum/ClientBase.cpp
```diff
@@ -417,9 +417,12 @@ BlockHeader ClientBase::pendingInfo() const
 
 BlockDetails ClientBase::pendingDetails() const
 {
-    auto pm = postSeal().info();
-    auto li = Interface::blockDetails(LatestBlock);
-    return BlockDetails((unsigned)pm.number(), li.totalDifficulty + pm.difficulty(), pm.parentHash(), h256s{});
+    auto pendingHeader = postSeal().info();
```
gumb0: Can be `const`

libethereum/ClientBase.cpp
```diff
@@ -417,9 +417,12 @@ BlockHeader ClientBase::pendingInfo() const
...
+    auto latestDetails = Interface::blockDetails(LatestBlock);
+    return BlockDetails
+    {static_cast<unsigned>(pendingHeader.number()),
```
gumb0: weird indentation, is this clang-format?

libethereum/State.cpp
```diff
@@ -42,40 +44,39 @@ State::State(State const& _s):
...
 OverlayDB State::openDB(fs::path const& _basePath, h256 const& _genesisHash, WithExisting _we)
 {
-    auto const path = _basePath.empty() ? db::databasePath() : _basePath;
-    auto const chainPath = path / fs::path(toHex(_genesisHash.ref().cropped(0, 4))) /
-                           fs::path(toString(c_databaseVersion));
-    auto const statePath = chainPath / fs::path("state");
-
     if (db::isDiskDatabase())
     {
+        db_paths::setDatabasePaths(_basePath, _genesisHash);
```
gumb0: You call this from here and in `BlockChain` class, so it will be called at least twice, won't it?
Also it makes it impossible to have two instantces of `State` pointing to different databases.

It should rather be called once somewhere at the higher level (like main function or `Client` initialization).

libethereum/State.cpp
```diff
@@ -42,40 +44,39 @@ State::State(State const& _s):
...
+        clog(VerbosityDebug, "statedb")
+            << "Ensuring permissions are set for path: " << db_paths::chainPath();
+        DEV_IGNORE_EXCEPTIONS(fs::permissions(db_paths::chainPath(), fs::owner_all));
     }
 
     try
     {
-        clog(VerbosityTrace, "statedb") << "Opening state database: " << statePath;
-        std::unique_ptr<db::DatabaseFace> db = db::DBFactory::create(statePath);
+        clog(VerbosityTrace, "statedb") << "Opening state database";
```
gumb0: Why did you remove the path here?

libethereum/BlockChain.cpp
```diff
@@ -226,61 +219,78 @@ bool BlockChain::open(fs::path const& _path, WithExisting _we)
...
+        LOG(m_loggerError) << "Unknown error occurred when opening database. Exception details: "
+                           << ex.what();
         throw;
     }
 
     if (_we != WithExisting::Verify && !rebuildNeeded && !details(m_genesisHash))
     {
-        BlockHeader gb(m_params.genesisBlock());
+        bytes const genesisBlockBytes = m_params.genesisBlock();
+        BlockHeader gb{genesisBlockBytes};
```
gumb0: can be `const`

libethereum/ClientBase.cpp
halfalicious: @gumb0 Yes, clang did that...i can manually format it a little nicer

libethereum/State.cpp
halfalicious: @gumb0 In case we're running memoryDB

libethereum/BlockChain.cpp
halfalicious: Ah nice catch, thanks!

libethereum/State.cpp
halfalicious: @gumb0 For context, when would we ever have two `State `instances pointing to different databases?

libethereum/ClientBase.cpp
halfalicious: Nice catch, thanks!

libethereum/State.cpp
gumb0: Maybe in tests, maybe nowhere, but it's one thing to set the global data once and then only read it in various places, but writing global data from various places just doesn't seem good practice - both from thread-safety perspective and from general complexity management perspective

halfalicious:
Rebased to address CHANGELOG merge conflict

halfalicious:
Need to fix OpenDB test error

libethereum/BlockChain.cpp
```diff
@@ -194,30 +194,26 @@ void BlockChain::init(ChainParams const& _p)
...
-    auto const chainPath = path / fs::path(toHex(m_genesisHash.ref().cropped(0, 4)));
-    auto const chainSubPathBlocks = chainPath / fs::path("blocks");
-    auto const extrasPath = chainPath / fs::path(toString(c_databaseVersion));
-    auto const extrasSubPathExtras = extrasPath / fs::path("extras");
     unsigned lastMinor = c_databaseMinorVersion;
     bool rebuildNeeded = false;
 
     if (db::isDiskDatabase())
     {
+        if (!_path.empty())
```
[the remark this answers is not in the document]
halfalicious: @gumb0: We can, but that seems unnecessary if m_dbPaths is already set?

libethereum/BlockChain.cpp
```diff
@@ -226,88 +222,107 @@ bool BlockChain::open(fs::path const& _path, WithExisting _we)
...
+    LOG(m_loggerInfo) << "Opened blockchain database. Latest block hash: " << currentHash()
+                      << (!rebuildNeeded ? "(rebuild not needed)" : "*** REBUILD NEEDED ***");
     return rebuildNeeded;
 }
 
 void BlockChain::open(fs::path const& _path, WithExisting _we, ProgressCallback const& _pc)
 {
     if (open(_path, _we) || _we == WithExisting::Verify)
-        rebuild(_path, _pc);
+        rebuild(fs::path{}, _pc);
```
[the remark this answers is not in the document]
halfalicious: @gumb0: Wanted a way to avoid unnecessarily reconstructing the DatabasePaths object...I don’t think we need to recreate it in rebuild if we already have built m_dbPaths

libethereum/BlockChain.cpp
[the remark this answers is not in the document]
gumb0: You could then just 
```
if (!m_dbPaths.pathsSet()) 
  m_dbPaths.setPaths(_path, m_genesisHash);
```

But I would probably make it `unique_ptr` and
```
if (!m_dbPaths) 
  m_dbPaths = std::make_unique<DatabasePaths>(_path, m_genesisHash);
```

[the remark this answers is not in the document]
halfalicious: @gumb0 unique_ptr makes sense, but I like the current behavior of only setting the path if the supplied path arg is empty since I think that's more intuitive, i.e. if I call a function and pass it an arg I expect that arg to be used and not silently dropped. What do you think?

libethereum/DatabasePaths.h
```diff
@@ -0,0 +1,39 @@
...
+namespace dev
+{
+namespace eth
+{
+class DatabasePaths
+{
+public:
+    DatabasePaths(boost::filesystem::path const& _rootPath, h256 const& _genesisHash) noexcept;
+    DatabasePaths() = default;
+    void setPaths(boost::filesystem::path const& _rootPath, h256 const& _genesisHash) noexcept;
```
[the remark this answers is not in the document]
halfalicious: @gumb0 I like the immutable idea, I went with the current approach since I was hesitant to use dynamic allocation for something trivial like paths. I can use unique_ptr and remove the default ctor + setter method which will effectively make the class immutable.

libethereum/DatabasePaths.h
```diff
@@ -0,0 +1,39 @@
...
+    bool pathsSet() const noexcept;
+    boost::filesystem::path const& rootPath() noexcept;
```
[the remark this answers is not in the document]
halfalicious: Ah of course 😀

libethereum/BlockChain.cpp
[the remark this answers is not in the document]
gumb0: Makes sense, but I don't like the hidden features of the function interface, like magic value "empty path" meaning "use previous path".
How about instead of empty path calling it with the current path (as before) and then here:
```
if (!m_dbPaths || m_dbPaths.rootPath() != _path) 
  m_dbPaths = std::make_unique<DatabasePaths>(_path, m_genesisHash);
```

then it's kind of an optimization to not reset the paths when not needed. And in practice re-creation will never happen, because it's called with the same path only.

libethereum/BlockChain.cpp
```diff
@@ -226,74 +220,93 @@ bool BlockChain::open(fs::path const& _path, WithExisting _we)
...
         }
-        
-        cerror << "Unknown error occurred. Exception details: " << ex.what();
+
+        LOG(m_loggerError) << "Unknown error occurred when opening database. Exception details: "
+                           << ex.what();
         throw;
     }
 
     if (_we != WithExisting::Verify && !rebuildNeeded && !details(m_genesisHash))
```
halfalicious: I can slightly streamline things by initializing rebuildNeeded to (_we == WithExisting::Verify) and then checking !rebuildNeeded here rather than also checking _we. This would also enable me to skip the minor version checks if rebuildNeeded == true. However, this file already has a ton of churn and this streamlining is pretty minor so I can take care of it in a future change.

halfalicious:
Squashed a bunch of commits and rebased

PR #5830, opened 2019-11-18
Fix mainnet and ropsten configs for correct EC precompiles cost

gumb0:
Without this change cost of ECADD and ECADD stays the same in Istanbul.

See also EIP-1108 implementation https://github.com/ethereum/aleth/pull/5705

### DIFF
--- a/libethashseal/genesis/mainNetwork.cpp
+++ b/libethashseal/genesis/mainNetwork.cpp
@@ -47,10 +47,10 @@ R"E(
         "0000000000000000000000000000000000000003": { "precompiled": { "name": "ripemd160", "linear": { "base": 600, "word": 120 } } },
         "0000000000000000000000000000000000000004": { "precompiled": { "name": "identity", "linear": { "base": 15, "word": 3 } } },
         "0000000000000000000000000000000000000005": { "precompiled": { "name": "modexp", "startingBlock": "0x42ae50" } },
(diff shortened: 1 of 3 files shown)


libethashseal/genesis/mainNetwork.cpp
```diff
@@ -47,10 +47,10 @@ R"E(
         "0000000000000000000000000000000000000003": { "precompiled": { "name": "ripemd160", "linear": { "base": 600, "word": 120 } } },
         "0000000000000000000000000000000000000004": { "precompiled": { "name": "identity", "linear": { "base": 15, "word": 3 } } },
         "0000000000000000000000000000000000000005": { "precompiled": { "name": "modexp", "startingBlock": "0x42ae50" } },
-        "0000000000000000000000000000000000000006": { "precompiled": { "name": "alt_bn128_G1_add", "startingBlock": "0x42ae50", "linear": { "base": 500, "word": 0 } } },
```
chfast: So where is the new cost encoded then?

gumb0: It's now in the code, because it needs the current block number to figure out the precompile cost https://github.com/ethereum/aleth/blob/403dc9067a453420c43c974f919586285dc30ab7/libethcore/Precompiled.cpp#L181

chfast:
Oh, this should be still `release/1.7` branch.

gumb0:
@chfast Same patch for `release/1.7` branch https://github.com/ethereum/aleth/pull/5833

PR #5831, opened 2019-11-18
Disable all compiler optimizations in Debug build

gumb0:
This really makes it slower, on my machine the full `testeth` run (without `--all`) takes ~17 minutes with `-Og` vs 23-25 minutes with `-O0`.

But I think it's worth to try it, as Debug is not built on CI, and I don't really run all of the tests locally often.


gumb0:
Note to @winsvega: please never measure any performance with Debug build of aleth/testeth, it's going to be really slow now.

PR #5840, opened 2019-11-21
Get rid of precompile list in configs

gumb0:
Closes https://github.com/ethereum/aleth/issues/5825

1. Instead of filling `ChainOperationParams::precompiled` from config JSON, fill it with hard-coded list of 9 currently existing precompiles.
2. `PrecompiledContract` constructor changed to take precompile name string and starting block.
3. Price functions for linear cost precompiles created, to keep all pricers consistently in one place (instead of keeping price constants somewhere else, as before)
4. I've decided to consider `"precompile"` records in configs deprecated for now, so they will not generate validation error, but be ignored.

We could also get rid of `PrecompiledRegistrar` at some point, I think it's a useless complication.

cc @winsvega


gumb0:
Rebased

Issue #5835, opened 2019-11-19
Exception thrown during syncing: Throw in function class dev::eth::EthereumPeer &__cdecl dev::eth::EthereumCapability::peer(const class dev::FixedHash<64> &)


halfalicious:
Exceptions can be thrown during syncing if a peer reconnects while there's a pending disconnect - there will be a live `Session` (entry in `Host::m_sessions` which is still alive and `isConnected`) but no entry for the node in `EthereumCapability::m_peers`, so if the peer sends us an eth capability packet, or if something happens which triggers us to continue syncing (`BlockChainSync::continueSync`) e.g. a node (other than the one which reconnected during the pending disconnect) initiates a disconnect, we will try to lookup the peer in `m_peers ` and not find it and throw here: https://github.com/ethereum/aleth/blob/3396b487a7c5b7f1c47dcecd47b0260793931466/libethereum/EthereumCapability.cpp#L1004-L1011

Please see [here ](https://github.com/ethereum/aleth/issues/5633#issuecomment-550934074)for a detailed description of this scenario. 

The exception that we will see when a node disconnects (other than the one which reconnected during a pending disconnect) is:
```
WARN  11-05 09:55:39 p2p  warn   Exception on peer destruction: C:\Users\nilse\Documents\Code\aleth_ref\libethereum\EthereumCapability.cpp(1008): Throw in function class dev::eth::EthereumPeer &__cdecl dev::eth::EthereumCapability::peer(const class dev::FixedHash<64> &)
Dynamic exception type: struct boost::wrapexcept<struct dev::eth::PeerDisconnected>
[struct dev::tag_nodeID * __ptr64] = e5a7df63cde56b0a33b14de356dcae10cc8a7e3ec18a3749bc51c3566fbad09467980ebd4e4fc8610152781a7832a61539a2267da09d781871cb33dc54af65da
```

The exception that we will see when we receive a capability packet from the node which reconnected during a pending disconnect is:

```
WARN  11-05 09:55:35 p2p  ethcap Peer ##9b2af51d… causing an exception: C:\Users\nilse\Documents\Code\aleth_ref\libethereum\EthereumCapability.cpp(1008): Throw in function class dev::eth::EthereumPeer &__cdecl dev::eth::EthereumCapability::peer(const class dev::FixedHash<64> &)
Dynamic exception type: struct boost::wrapexcept<struct dev::eth::PeerDisconnected>
[...]

PR #5841, opened 2019-11-22
Revert "wait for 2 seconds on disconnect" changes

halfalicious:
Changes were causing frequent exceptions during syncing (see #5835, #5828) due to `Host::m_session`s and `EthereumPeer::m_peers` getting out-of-sync (if a peer reconnects during a pending disconnect), so we are reverting the changes in the 1.7 release until we can properly address the sync issues.


gumb0:
I think we shouldn't modify 1.7.0 changelog, as it's already released, so I added an entry about revert to 1.7.2 instead

PR #5844, opened 2019-11-22
docker: Add all tools to aleth docker image

### LINKED ISSUE #5842: Add aleth-vm to docker image
Needed by #4613.

chfast:
This will additionally add: aleth-bootnode, aleth-key and aleth-vm. The aleth-key is probably useless but for simplicity it may stay.

The default entry point is still aleth.py, but you can use other tools as

```
docker run -it --rm --entrypoint /usr/bin/aleth-vm ethereum/aleth --version
aleth-vm 1.8.0-alpha.1-12+commit.43780e1c.dirty
Build: linux/release
```

Closes #5842.


chfast:
The smallest PR ever.

PR #5849, opened 2019-11-25
Remove ChainParams default constructor

gumb0:


halfalicious:
@gumb0 Why is this needed?

gumb0:
@halfalicious My reasoning was that it's weird to have the notion of default chain params, initialized with some magical values (e.g. there are 4 frontier precompiles hard-coded in default constructor), and in practice chain params are always created from some config JSON.

But now it seems that it's still useful for some unit tests that don't need any particular config and just want some chain params quickly constructed, without parsing JSON.

So I think I'll close this.
Any thoughts?

halfalicious:
@gumb0 That sounds good to me. I'm running into a similar issue building database directories using the genesis hash

PR #5851, opened 2019-11-27
Post body of onRoomAvailable handler to network thread

halfalicious:
The `onRoomAvailable `handler calls `BlockChainSync::syncPeer` on each peer to request more blocks - this handler is currently called on the client thread (since the handler is called in `BlockQueue::drain` and `drain `is called on the client thread) and we want network-related syncing work to occur on the network thread to reduce lock contention and reduce the impact of network-related work on other threads.


gumb0:
Does this improve anything with exceptions etc.?

halfalicious:
it will prevent network related exceptions that occur during syncing triggered by draining a full block queue from stopping the client thread (which effectively stops syncing)

PR #5852, opened 2019-11-27
Output original opcode instead of PUSHC/JUMPC/JUMPCI in VM trace

gumb0:
Fixes https://github.com/ethereum/aleth/issues/5847

cc @holiman

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 - Fixed: [#5821](https://github.com/ethereum/aleth/pull/5821) `test_setChainParams` correctly initializes custom configuration of precompiled contracts.
 - Fixed: [#5826](https://github.com/ethereum/aleth/pull/5826) Fix blocking bug in database rebuild functionality - users can now rebuild their databases via Aleth's '-R' switch.
 - Fixed: [#5827](https://github.com/ethereum/aleth/pull/5827) Detect database upgrades and automatically rebuild the database when they occur.
-- Fixed: [#5834](https://github.com/ethereum/aleth/pull/5834) Fix segmentation fault during sync.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -18,7 +18,7 @@
 - Fixed: [#5821](https://github.com/ethereum/aleth/pull/5821) `test_setChainParams` correctly initializes custom configuration of precompiled contracts.
 - Fixed: [#5826](https://github.com/ethereum/aleth/pull/5826) Fix blocking bug in database rebuild functionality - users can now rebuild their databases via Aleth's '-R' switch.
 - Fixed: [#5827](https://github.com/ethereum/aleth/pull/5827) Detect database upgrades and automatically rebuild the database when they occur.
-- Fixed: [#5834](https://github.com/ethereum/aleth/pull/5834) Fix segmentation fault during sync.
```
gumb0: Removed this entry, as it's already mentioned in 1.7.2

libevm/LegacyVM.cpp
```diff
@@ -1582,7 +1583,7 @@ void LegacyVM::interpretCases()
         CASE(JUMPCI)
         {
 #if EVM_REPLACE_CONST_JUMP
-            ON_OP();
+            onOperation(Instruction::JUMPC);
```
chfast:
```suggestion
            onOperation(Instruction::JUMPI);
```

gumb0:
> Does this resolve to no-op in case there is no tracing callback?

@chfast Yes, in case callback is not set `opOperation` does nothing https://github.com/ethereum/aleth/pull/5852/files#diff-214bff4e48034ccf16329f486716aefbR63-R66

There's other definition for `ON_OP` when `EVM_TRACE` compile-time flag is set, outputting to stderr instead. These changes don't affect that, but I suspect it's not useful anymore.
https://github.com/ethereum/aleth/blob/bdc76746c2950c1981deb50c38ded27f6ad46e6a/libevm/LegacyVMConfig.h#L68-L73

PR #5857, opened 2019-12-03
Fix Windows build error in libscrypt

halfalicious:
Build error is due to `strerror `usage. Ideally we should fix this in the libscrypt source and then in the corresponding Hunter package.

### DIFF
--- a/cmake/Hunter/config.cmake
+++ b/cmake/Hunter/config.cmake
@@ -33,3 +33,11 @@ hunter_config(
     SHA1 152b849610d91a9dfa1401293f43230c2e0c33f8
     CMAKE_ARGS BUILD_GMOCK=OFF gtest_force_shared_crt=ON
 )
+
+if (MSVC)
+    hunter_config(
+        libscrypt
+        VERSION ${HUNTER_libscrypt_VERSION}
+        CMAKE_ARGS CMAKE_C_FLAGS=-D_CRT_SECURE_NO_WARNINGS
+    )
+endif()


cmake/Hunter/config.cmake
```diff
@@ -33,3 +33,9 @@ hunter_config(
     SHA1 152b849610d91a9dfa1401293f43230c2e0c33f8
     CMAKE_ARGS BUILD_GMOCK=OFF gtest_force_shared_crt=ON
 )
+
+hunter_config(
+    libscrypt
+    VERSION 1.21-p1
+    CMAKE_ARGS CMAKE_C_FLAGS=-D_CRT_SECURE_NO_WARNINGS
```
gumb0: @chfast should it be better under `if (MSVC)` only?

chfast: Yes.

cmake/Hunter/config.cmake
```diff
@@ -33,3 +33,9 @@ hunter_config(
     SHA1 152b849610d91a9dfa1401293f43230c2e0c33f8
     CMAKE_ARGS BUILD_GMOCK=OFF gtest_force_shared_crt=ON
 )
+
+hunter_config(
+    libscrypt
+    VERSION 1.21-p1
```
chfast:
```suggestion
    VERSION ${HUNTER_libscrypt_VERSION}
```
Not to override the default version number.

cmake/Hunter/config.cmake
chfast: The whole `hunter_config()` can be inside `if(MSVC)` then.

PR #5864, opened 2019-12-06
Allow submitting several transaction from the same sender

### LINKED ISSUE #5863: Invalid transaction nonce when import transaction of the same origin
When importing multiple transactions from one address into the mining block

```
trace  Processed 1 transactions in0.285(false)
Nonce: 0x01
Request: {"jsonrpc":"2.0","method":"eth_sendRawTransaction","params":["0xf85f010a82520894bbbf5374fce5edbc8e2a8697c15331677e6ebf0b0a801ba04fa966bf34b93abc1bcd665554b7f316b50f928477b50be0f3285ead29d18c5ba017bba0eeec1625ab433746955e125d46d80b7fdc97386c51266f842d8e02192e"],"id":5}
TRACE 12-05 14:50:28 aleth rpc    {"jsonrpc":"2.0","method":"eth_sendRawTransaction","params":["0xf85f010a82520894bbbf5374fce5edbc8e2a8697c15331677e6ebf0b0a801ba04fa966bf34b93abc1bcd665554b7f316b50f928477b50be0f3285ead29d18c5ba017bba0eeec1625ab433746955e125d46d80b7fdc97386c51266f842d8e02192e"],"id":5}
DEBUG 12-05 14:50:28 aleth overlaydb Closing state DB
DEBUG 12-05 14:50:28 aleth exec   Sender: a94f5374fce5edbc8e2a8697c15331677e6ebf0b Invalid Nonce: Require 0 Got 1
TRACE 12-05 14:50:28 aleth rpc    {"error":{"code":0,"data":null,"message":"Invalid transaction nonce."},"id":5,"jsonrpc":"2.0"}

Reply: {"error":{"code":0,"data":null,"message":"Invalid transaction nonce."},"id":5,"jsonrpc":"2.0"}
```

the second (valid) transaction fails due to incorrect nonce exception. 
making aleth unable to construct a pending block with multiple transactions from one address (for blockchain test generation in retesteth)

halfalicious:
Fix #5863 

The old check was restrictive - it prohibited the same account from sending multiple transactions before a new block was mined, because the check required that the new tx nonce match the nonce retrieved from the state in the most recent block in the local chain. The new check is more flexible - it requires that new tx nonces be greater than or equal to the required nonce.

These changes also add a test which verifies the new behavior.

### DIFF
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -109,11 +109,10 @@ void Executive::initialize(Transaction const& _transaction)
             m_excepted = TransactionException::InvalidSignature;
             throw;
         }
-        if (m_t.nonce() < nonceReq)
+        if (m_t.nonce() != nonceReq)
         {
-            LOG(m_execLogger) << "Sender: " << m_t.sender().hex()
-                              << " Invalid Nonce: Required nonce >= " << nonceReq << ", received "
-                              << m_t.nonce();
+            LOG(m_execLogger) << "Sender: " << m_t.sender().hex() << " Invalid Nonce: Required "
+                              << nonceReq << ", received " << m_t.nonce();
             m_excepted = TransactionException::InvalidNonce;
             BOOST_THROW_EXCEPTION(
                 InvalidNonce() << RequirementError((bigint)nonceReq, (bigint)m_t.nonce()));
--- a/libethereum/Executive.cpp
+++ b/libethereum/Executive.cpp
@@ -109,12 +109,14 @@ void Executive::initialize(Transaction const& _transaction)
             m_excepted = TransactionException::InvalidSignature;
             throw;
         }
-        if (m_t.nonce() != nonceReq)
+        if (m_t.nonce() < nonceReq)
         {
-            LOG(m_execLogger) << "Sender: " << m_t.sender().hex() << " Invalid Nonce: Require "
-                              << nonceReq << " Got " << m_t.nonce();
+            LOG(m_execLogger) << "Sender: " << m_t.sender().hex()
+                              << " Invalid Nonce: Required nonce >= " << nonceReq << ", received "
+                              << m_t.nonce();
             m_excepted = TransactionException::InvalidNonce;
-            BOOST_THROW_EXCEPTION(InvalidNonce() << RequirementError((bigint)nonceReq, (bigint)m_t.nonce()));
+            BOOST_THROW_EXCEPTION(
+                InvalidNonce() << RequirementError((bigint)nonceReq, (bigint)m_t.nonce()));
         }
 
         // Avoid unaffordable transactions.
--- a/test/unittests/libweb3jsonrpc/jsonrpc.cpp
+++ b/test/unittests/libweb3jsonrpc/jsonrpc.cpp
@@ -263,6 +263,7 @@ BOOST_AUTO_TEST_CASE(eth_sendTransaction)
     auto address = coinbase.address();
     auto countAt = jsToU256(rpcClient->eth_getTransactionCount(toJS(address), "latest"));
 
+    // Verify initial account state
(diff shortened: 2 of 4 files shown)


libethereum/Executive.cpp
```diff
@@ -109,12 +109,14 @@ void Executive::initialize(Transaction const& _transaction)
             m_excepted = TransactionException::InvalidSignature;
             throw;
         }
-        if (m_t.nonce() != nonceReq)
+        if (m_t.nonce() < nonceReq)
```
gumb0: I believe this is going to break consensus, allowing the blocks with transactions in incorrect order, so this won't do.

gumb0: @winsvega do we have any blockchain tests with blocks containing transactions with invalid nonces?

gumb0:
A couple of unit tests still fails

test/unittests/libweb3jsonrpc/jsonrpc.cpp
```diff
@@ -377,8 +437,9 @@ BOOST_AUTO_TEST_CASE(eth_sendRawTransaction_errorInvalidNonce)
     auto txHash = rpcClient->eth_sendRawTransaction(signedTx["raw"].asString());
     BOOST_REQUIRE(!txHash.empty());
 
-    auto invalidNonce =
-        jsToU256(rpcClient->eth_getTransactionCount(toJS(senderAddress), "latest")) - 1;
```
gumb0: without mining previously sent transaction, `getTransactionCount` here returned 0, then u256 underflowed, and `invalidNonce` was max<u256>

I replaced it with just using 0 as invalid nonce

test/unittests/libweb3jsonrpc/jsonrpc.cpp
```diff
@@ -415,39 +476,6 @@ BOOST_AUTO_TEST_CASE(eth_sendRawTransaction_errorInsufficientGas)
...
-    BOOST_REQUIRE(!txHash.empty());
-
-    auto txNonce = jsToU256(rpcClient->eth_getTransactionCount(toJS(senderAddress), "latest"));
-    t["nonce"] = jsToDecimal(toJS(txNonce));
-
-    signedTx = rpcClient->eth_signTransaction(t);
-    BOOST_REQUIRE(!signedTx["raw"].empty());
-
-    BOOST_CHECK_EQUAL(sendingRawShouldFail(signedTx["raw"].asString()),
-        "Same transaction already exists in the pending transaction queue.");
```
gumb0: It's really difficult now to hit this error, because we check the nonce now against pending block, and the first transaction is added to the pending block faster than we manage to send the second, duplicate one.
So sending the duplicate transaction still fails, but with "Invalid nonce" error.

I haven't found anything better to do, than remove this test.
(changing the expected error would work, but then it would be kind of the same test as `eth_sendRawTransaction_errorInvalidNonce` above)

winsvega:
With this PR I am able to import multiple transactions from same address for blockchain test generation. 
Thanks!

libethereum/Executive.cpp
halfalicious: @gumb0 Ah you're correct, I had thought that the `Executive `was only invoked when submitting txs but it looks like it's invoked when executing txs as a part of importing txs into a block: `Block::sync->Block::execute->State::execute->State::executeTransaction->Executive::initialize`

...and when importing blocks into the chain: 

`BlockChain::sync -> BlockChain::import -> Block::enactOn -> Block::enact -> Block::execute -> State::execute -> State::executeTransaction -> Executive::initialize`

We definitely don't want to change the nonce check in the Executive since that's consensus-critical code. Thanks for catching this.

test/unittests/libweb3jsonrpc/jsonrpc.cpp
halfalicious: Ah, nice catch!

PR #5865, opened 2019-12-06
Fix race condition which permanently pauses sync

### LINKED ISSUE #5312: Sync gets stuck

halfalicious:
Fix #5312 

The race condition occurs when the block queue is full so syncing is paused, and the function which detects when the block queue has room (`BlockQueue::drain`) does the first of the two-part check when a block queue verifier thread has temporarily removed block data from the verification queues.
As such, the "room available" detection fails and `drain `removes more blocks, which means that the block queue enters a permanently not full state and syncing is paused indefinitely.

This is fixed by also performing the "room available" detection check in the body of the block queue verifier threads.


halfalicious:
Yes kicked it off last night, will let it run for a bit before merging.

halfalicious:
Ran sync for ~12 hours and sync'd from genesis to block 1.4M without issues.

PR #5868, opened 2019-12-10
return importBlockFailed ImportResult in test_importRawBlock

winsvega:
Export exception information on test_improtRawBlock

### DIFF
--- a/libweb3jsonrpc/Test.cpp
+++ b/libweb3jsonrpc/Test.cpp
@@ -7,6 +7,7 @@
 #include <jsonrpccpp/common/errors.h>
 #include <jsonrpccpp/common/exception.h>
 #include <libdevcore/CommonJS.h>
+#include <libethcore/Exceptions.h>
 #include <libethereum/ChainParams.h>
 #include <libethereum/ClientTest.h>
 
@@ -115,6 +116,35 @@ bool Test::test_rewindToBlock(int _number)
     return true;
 }
 
+string ImportResultToString(ImportResult const& _res)
+{
+    switch (_res)
+    {
+    case ImportResult::AlreadyInChain:
+        return "AlreadyInChain";
+    case ImportResult::AlreadyKnown:
+        return "AlreadyKnown";
+    case ImportResult::BadChain:
+        return "BadChain";
+    case ImportResult::FutureTimeKnown:
+        return "FutureTimeKnown";
+    case ImportResult::FutureTimeUnknown:
+        return "FutureTimeUnknown";
+    case ImportResult::Malformed:
+        return "Malformed";
+    case ImportResult::OverbidGasPrice:
+        return "OverbidGasPrice";
+    case ImportResult::Success:
+        return "Success";
+    case ImportResult::UnknownParent:
+        return "UnknownParent";
+    case ImportResult::ZeroSignature:
+        return "ZeroSignature";
+    default:
+        return "unknown";
+    }
+    return "unknown";
+}
 std::string Test::test_importRawBlock(string const& _blockRLP)
 {
     try
@@ -125,7 +155,8 @@ std::string Test::test_importRawBlock(string const& _blockRLP)
     catch (ImportBlockFailed const& e)
     {
         cwarn << diagnostic_information(e);
-        throw JsonRpcException("Block import failed.");
+        ImportResult const& res = *boost::get_error_info<dev::eth::errinfo_importResult>(e);
+        throw JsonRpcException("Block import failed: " + ImportResultToString(res));
     }
     catch (std::exception const& e)
     {
--- a/libweb3jsonrpc/Test.cpp
+++ b/libweb3jsonrpc/Test.cpp
@@ -21,6 +21,114 @@ Test::Test(eth::Client& _eth): m_eth(_eth) {}
 
 namespace
 {
+string importResultToErrorMessage(ImportResult _res)
+{
+    switch (_res)
+    {
+    case ImportResult::AlreadyInChain:
+        return "AlreadyInChain";
+    case ImportResult::AlreadyKnown:
+        return "AlreadyKnown";
+    case ImportResult::BadChain:
+        return "BadChain";
+    case ImportResult::FutureTimeKnown:
+        return "FutureTimeKnown";
+    case ImportResult::FutureTimeUnknown:
+        return "FutureTimeUnknown";
+    case ImportResult::Malformed:
+        return "Malformed";
+    case ImportResult::OverbidGasPrice:
+        return "OverbidGasPrice";
+    case ImportResult::Success:
+        return "Success";
+    case ImportResult::UnknownParent:
+        return "UnknownParent";
+    case ImportResult::ZeroSignature:
+        return "ZeroSignature";
+    default:
+        return "ImportResult unhandled case";
+    }
+}
+
+string exceptionToErrorMessage(boost::exception_ptr _e)
+{
+    string ret;
+    try
+    {
+        boost::rethrow_exception(_e);
+    }
+    catch (ExtraDataTooBig const&)
+    {
+        ret = "ExtraData too big.";
+    }
+    catch (InvalidDifficulty const&)
+    {
+        ret = "Invalid Difficulty.";
+    }
+    catch (InvalidGasLimit const&)
+    {
+        ret = "Invalid Block GasLimit.";
+    }
+    catch (BlockGasLimitReached const&)
+    {
+        ret = "Block GasLimit reached.";
+    }
+    catch (TooMuchGasUsed const&)
+    {
+        ret = "Too much gas used.";
+    }
+    catch (InvalidNumber const&)
+    {
+        ret = "Invalid number.";
+    }
+    catch (InvalidLogBloom const&)
+    {
+        ret = "Invalid log bloom.";
+    }
+    catch (InvalidTimestamp const&)
+    {
+        ret = "Invalid timestamp.";
+    }
+    catch (InvalidBlockNonce const&)
+    {
+        ret = "Invalid block nonce.";
+    }
+    catch (UnknownParent const&)
+    {
+        ret = "Unknown parent.";
+    }
+    catch (InvalidUnclesHash const&)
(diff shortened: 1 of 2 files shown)


libweb3jsonrpc/Test.cpp
```diff
@@ -115,6 +116,35 @@ bool Test::test_rewindToBlock(int _number)
     return true;
 }
 
+string ImportResultToString(ImportResult const& _res)
```
gumb0: Put this into unnamed namespace above

halfalicious:
Unit test failures:
```

760/769 Test #761: JsonRpcSuite/test_importRawBlockFailsForInvalidHeader .................................................................***Failed    0.22 sec
Running 1 test case...
Test Case "test_importRawBlockFailsForInvalidHeader": 
unknown location(0): fatal error: in "JsonRpcSuite/test_importRawBlockFailsForInvalidHeader": memory access violation at address: 0x00000000: no mapping at fault address
/home/builder/project/test/unittests/libweb3jsonrpc/jsonrpc.cpp(804): last checkpoint

*** 1 failure is detected (6 failures are expected) in the test module "Master Test Suite"


761/769 Test #762: JsonRpcSuite/test_importRawBlockFailsForInvalidBlockBody ..............................................................***Failed    0.22 sec
Running 1 test case...
Test Case "test_importRawBlockFailsForInvalidBlockBody": 
unknown location(0): fatal error: in "JsonRpcSuite/test_importRawBlockFailsForInvalidBlockBody": memory access violation at address: 0x00000000: no mapping at fault address
/home/builder/project/test/unittests/libweb3jsonrpc/jsonrpc.cpp(828): last checkpoint

*** 1 failure is detected (6 failures are expected) in the test module "Master Test Suite"


762/769 Test #760: JsonRpcSuite/test_importRawBlock ......................................................................................   Passed    0.32 sec
        Start 765: ClientBase.1
        Start 766: ClientBase.2
        Start 767: ClientBase.3
        Start 768: ClientBase.4
        Start 769: ClientBase.5
763/769 Test #768: ClientBase.4 ..........................................................................................................   Passed    0.10 sec
764/769 Test #769: ClientBase.5 ..........................................................................................................   Passed    0.10 sec
765/769 Test #767: ClientBase.3 ..........................................................................................................   Passed    0.10 sec
766/769 Test #766: ClientBase.2 ..........................................................................................................   Passed    0.11 sec
767/769 Test #765: ClientBase.1 ..........................................................................................................   Passed    0.21 sec
768/769 Test #735: ClientTests/Personal ..................................................................................................   Passed    8.36 sec
769/769 Test #313: LegacyTests/Constantinople/BlockchainTests/ValidBlocks/bcExploitTest ..................................................   Passed  145.04 sec

99% tests passed, 2 tests failed out of 769

Total Test time (real) = 180.73 sec

The following tests FAILED:
	761 - JsonRpcSuite/test_importRawBlockFailsForInvalidHeader (Failed)
	762 - JsonRpcSuite/test_importRawBlockFailsForInvalidBlockBody (Failed)
Errors while running CTest
```

winsvega:
The tests are failing because of  this line

```
// check that it was really imported and not rejected as invalid
    if (!bc().isKnown(blockHash))
        BOOST_THROW_EXCEPTION(ImportBlockFailed() << errinfo_importResult(ImportResult::BadChain));
```

the block was successfully imported into blockqueue  but  then dropped from the chain. 
this is the case I was talking about. in this particular tests there should be an error code  Invalid Difficulty.

gumb0:
Your exception handler should better check `boost::get_error_info<dev::eth::errinfo_importResult>(e) != nullptr` before dereferencing it

gumb0:
As for Invalid Difficulty - I'm not sure, it seems to me from the code, that it should be checked on the very first stage during adding to Block Queue.

gumb0:
But as a general comment I'm not sure that the tests should require the exact error message to be returned from the client.

The way to check that invalid difficulty is rejected is just to construct the block that has all other fields right except for the difficulty, and check whether   the client rejects it. Without relying on error messages, error messages are not part of the consensus.

winsvega:
"The way to check that invalid difficulty is rejected is just to construct the block that has all other fields right except for the difficulty, and check whether the client rejects it. Without relying on error messages, error messages are not part of the consensus."  

that's how the tests do. But when I generate the test I must verify the exception to be safe that the exact exception is checked.

gumb0:
> that's how the tests do. 

Well the tests contain things like
```
                "expectExceptionALL" : "InvalidGasLimit",
```

So it looks they try to check the exact exception.

gumb0:
@winsvega Could you provide me the chain config and block data with invalid difficulty? I'll try to check what's happening with it

winsvega:
try running 
```./retesteth -t BlockchainTests/InvalidBlocks/bcInvalidHeaderTest -- --clients "aleth"  --verbosity 6 --singletest wrongDifficulty```

```
   "expectExceptionALL" : "InvalidGasLimit",
```
this is old format. I am customizing this so that clients could report any kind of message.

**Comment by gumb0** [comment]
Yeah this block doesn't pass the first checks in BlockQueue, but `BlockQueue::import` returns `ImportResult::Malformed` for all such checks.

**Comment by gumb0** [comment]
> this is old format. I am customizing this so that clients could report any kind of message.

I still don't get, how do you change it? If the test won't contain different errors, then what's the point of checking the error during generation? (the result test is the same anyway?)

File: libweb3jsonrpc/Test.cpp [hunk 2]
```diff
@@ -125,7 +155,8 @@ std::string Test::test_importRawBlock(string const& _blockRLP)
     catch (ImportBlockFailed const& e)
     {
         cwarn << diagnostic_information(e);
-        throw JsonRpcException("Block import failed.");
+        ImportResult const& res = *boost::get_error_info<dev::eth::errinfo_importResult>(e);
```
gumb0: So to fix the unit tests, just check here that the pointer returned from `get_error_info` is not null before dereferencing, don't add additional string if it's null.

Then this PR is fine, but I'm not sure if it helps you at all.

winsvega:
The generated test is the same but I need a confirmation from the client that a specific exception was hit to be sure.

gumb0:
To be sure of what?
You create this invalid header yourself from the valid one, then you create a test saying "this block should be rejected", what's the purpose of these additional checks, to verify that your own invalid header generation code works correctly?


So the test fillers still contain different exception names? Why not get rid of them there, too, if you remove them from filled test?

winsvega:
To be sure that a client that generates the test reject the block with required exception.

gumb0:
why do you need that?

winsvega:
This is my standart of quality for the tests

gumb0:
Ok I have some ideas how to return the exact error

winsvega:
Btw geth return exact exception

gumb0:
In https://github.com/ethereum/aleth/pull/5871 I'm adding a nested exception with the exact header validation error to `ImportBlockFailed` thrown from `ClientTest::importRawBlock`, see exact spec here https://github.com/ethereum/aleth/pull/5871/files#diff-4c08ffb035d5be4408f58a82ddff192bR35-R39

When it's merged, you would need to
1. rebase this
2. Add function to Test.cpp to convert exceptions to strings like this
```
string exceptionToErrorMessage(boost::exception_ptr _e)
{
    string ret;
    try
    {
        boost::rethrow_exception(_e);
    }
    catch (InvalidDifficulty const&)
    {
        ret = "Invalid Difficulty.";
    }
    catch (...)
    {
        ret = "Unknown error.";
    }
    return ret;
}
```
3. Handle `ImportBlockFailed` like this
```
    catch (ImportBlockFailed const& e)
    {
        // ...
        auto importResult = boost::get_error_info<errinfo_importResult>(e);
        std::string detailedError;
        if (importResult && *importResult == ImportResult::Malformed)
        {
            auto nested = boost::get_error_info<errinfo_nestedException>(e);
            if (nested)
                detailedError = exceptionToErrorMessage(*nested);
        }
        // ...
    }
```

winsvega:
rebased on top of https://github.com/ethereum/aleth/pull/5871
it works! thanks!

chfast:
Will you be able to finish it today so I can include it in the 1.8.0 release?

winsvega:
No. I checked yesterday this hack doesn't detect some of the block validation exceptions. Such as ```Invalid Timestamp, Invalid Number```

gumb0:
Please rebase on master

libethereum/ClientTest.cpp
```diff
@@ -170,7 +170,7 @@ h256 ClientTest::importRawBlock(const string& _blockRLP)
     // check that it was really imported and not rejected as invalid
     if (!bc().isKnown(blockHash))
     {
-        auto ex = ImportBlockFailed{};
+        auto ex = ImportBlockFailed{} << errinfo_importResult(ImportResult::Malformed);
```
gumb0: I think this is not necessary correct (it could fail for some other reason)

libweb3jsonrpc/Test.cpp
```diff
@@ -125,7 +177,15 @@ std::string Test::test_importRawBlock(string const& _blockRLP)
     catch (ImportBlockFailed const& e)
     {
         cwarn << diagnostic_information(e);
-        throw JsonRpcException("Block import failed.");
+        auto importResult = boost::get_error_info<errinfo_importResult>(e);
+        std::string detailedError = "empty errorInfo_inportResult";
+        if (importResult)
```
gumb0: You don't really need this check, so just check for the nested exception only

gumb0: Or you could additionaly check import result and handle its values, too https://github.com/ethereum/aleth/blob/255a75ec4fc8ed7fce3449404595d7589d5d907f/libethcore/Common.h#L95-L107

with something like
```
if (auto nested = boost::get_error_info<errinfo_nestedException>(e))
    detailedError = exceptionToErrorMessage(*nested);
else if (auto importResult = boost::get_error_info<errinfo_importResult>(e))
    detailedError = importResultToErrorMessage(*importResult);
```

